PR #8583, opened 2024-07-02
feat(continuous-test): Add histograms to measure latency of requests

bubu11e:
#### What this PR does

Add histograms metrics to measure latency of read and write requests.

This will allow us to better appreciate the performances of the underlying mimir cluster/instance

Read request latency metrics have a label to indicate if results cache is in use or not as this could have a huge impact on the performances.

This will allow us to setup SLO on an end-to-end write and read pipeline reflecting the normal behavior of our users.

#### Checklist

- [X] Tests updated.
- [X] Documentation added.

### DIFF
--- a/pkg/continuoustest/metrics.go
+++ b/pkg/continuoustest/metrics.go
@@ -3,6 +3,8 @@
 package continuoustest
 
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
@@ -12,8 +14,10 @@ import (
 type TestMetrics struct {
 	writesTotal                  *prometheus.CounterVec
 	writesFailedTotal            *prometheus.CounterVec
+	writesLatency                *prometheus.HistogramVec
 	queriesTotal                 *prometheus.CounterVec
 	queriesFailedTotal           *prometheus.CounterVec
+	queriesLatency               *prometheus.HistogramVec
 	queryResultChecksTotal       *prometheus.CounterVec
 	queryResultChecksFailedTotal *prometheus.CounterVec
 }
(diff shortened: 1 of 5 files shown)


pkg/continuoustest/metrics.go
```diff
@@ -30,6 +32,12 @@ func NewTestMetrics(testName string, reg prometheus.Registerer) *TestMetrics {
 			Help:        "Total number of failed write requests.",
 			ConstLabels: map[string]string{"test": testName},
 		}, []string{"status_code", "type"}),
+		writesLatency: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Name:        "mimir_continuous_test_writes_request_duration_seconds",
+			Help:        "Duration of the write requests.",
+			Buckets:     prometheus.ExponentialBuckets(0.001, 4, 6),
+			ConstLabels: map[string]string{"test": testName},
+		}, []string{"type"}),
```
narqo, line 40: @krajorama could you help us here, would we want all new histograms to be native-histograms?

tacole02 (approved):

pkg/continuoustest/metrics.go
bubu11e: Hi, I did not implements native histograms are this feature is still experimental. But changes are minor and resolutions is better so if you prefer just let me know and i would implement this change.

narqo: Yes, please. I think keeping the new metrics to be native histogram-ready, at least, doesn't increase the amount of metrics to [migrate][1] in the future.

[1]: https://grafana.com/docs/mimir/latest/send/native-histograms/#migrate-from-classic-histograms

bubu11e: Hi, I have updated the code to use native-histograms.

narqo (approved):
Works for me, thank you 🔥

PR #8586, opened 2024-07-02
Alertmanager: Add static headers to Grafana email receiver

santihernandezc:
This PR passes static headers to the Grafana email receiver at creation time. These headers come from Grafana.

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -358,7 +358,7 @@ func clusterWait(position func() int, timeout time.Duration) func() time.Duratio
 }
 
 // ApplyConfig applies a new configuration to an Alertmanager.
-func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, tmpls []io.Reader, rawCfg string, tmplExternalURL *url.URL) error {
+func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, tmpls []io.Reader, rawCfg string, tmplExternalURL *url.URL, staticHeaders map[string]string) error {
 	tmpl, err := loadTemplates(tmpls, WithCustomFunctions(am.cfg.UserID))
 	if err != nil {
 		return err
@@ -389,7 +389,7 @@ func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, 
 		return d + waitFunc()
 	}
 
-	integrationsMap, err := am.buildIntegrationsMap(cfg.Global, conf.Receivers, tmpl, tmpls)
+	integrationsMap, err := am.buildIntegrationsMap(cfg.Global, conf.Receivers, tmpl, tmpls, staticHeaders)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/alertmanager/alertmanager.go
```diff
@@ -505,6 +505,21 @@ func (am *Alertmanager) buildIntegrationsMap(gCfg *config.GlobalConfig, nc []*de
...
+		ContentTypes:  []string{"text/html"},
+		EhloIdentity:  gCfg.SMTPHello,
+		ExternalURL:   tmpl.ExternalURL.String(),
+		FromAddress:   gCfg.SMTPFrom,
+		FromName:      "Grafana",
+		Host:          gCfg.SMTPSmarthost.String(),
+		KeyFile:       gCfg.HTTPConfig.TLSConfig.KeyFile,
+		SkipVerify:    !gCfg.SMTPRequireTLS,
+		StaticHeaders: staticHeaders,
+	}
```
santihernandezc, line 521: This struct was previously created in `buildgrafanaReceiverIntegrations`. To avoid having to pass both the global config and the external headers as arguments, I'm creating it here and passing the final struct.

pkg/alertmanager/config.go
```diff
@@ -8,33 +8,43 @@ package alertmanager
...
 
 	"github.com/grafana/mimir/pkg/alertmanager/alertspb"
 )
 
-// createUsableGrafanaConfig creates an AlertConfigDesc from a GrafanaAlertConfigDesc.
+// createUsableGrafanaConfig creates an amConfig from a GrafanaAlertConfigDesc.
 // If provided, it assigns the global section from the Mimir config to the Grafana config.
 // The SMTP and HTTP settings in this section can be used to configure Grafana receivers.
-func createUsableGrafanaConfig(gCfg alertspb.GrafanaAlertConfigDesc, mCfg *alertspb.AlertConfigDesc) (alertspb.AlertConfigDesc, error) {
+func createUsableGrafanaConfig(gCfg alertspb.GrafanaAlertConfigDesc, mCfg *alertspb.AlertConfigDesc) (amConfig, error) {
```
santihernandezc, line 21: Every time we call `createUsableGrafanaConfig` we end up using the resulting `AlertConfigDesc` within an `amConfig` struct. Creating the final `amConfig` here makes sense and reduces the duplicated lines in the calling function.

pkg/alertmanager/multitenant.go
```diff
@@ -688,52 +684,47 @@ func (am *MultitenantAlertmanager) syncConfigs(cfgMap map[string]alertspb.AlertC
 
 // computeConfig takes an AlertConfigDescs struct containing Mimir and Grafana configurations.
 // It returns the final configuration and external URL the Alertmanager will use.
-func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs) (alertspb.AlertConfigDesc, *url.URL, error) {
-	var cfg alertspb.AlertConfigDesc
-	var externalURL *url.URL
-	var err error
+func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs) (amConfig, error) {
```
santihernandezc, line 687: Similar to `createUsableGrafanaConfig`'s case, we can return `amConfig` here.

pkg/alertmanager/multitenant_test.go
```diff
@@ -2462,11 +2462,12 @@ func TestComputeConfig(t *testing.T) {
...
-		cfg    alertspb.AlertConfigDescs
-		expErr string
-		expCfg alertspb.AlertConfigDesc
-		expURL string
+		name       string
+		cfg        alertspb.AlertConfigDescs
+		expErr     string
+		expCfg     alertspb.AlertConfigDesc
+		expURL     string
+		expHeaders map[string]string
```
santihernandezc, line 2470: This is the only new field in this test. I'm ensuring that we add the expected headers to the `amConfig` in each case.

alexander-akhmetov (approved):

PR #8591, opened 2024-07-03
blockbuilder: improve how group handles rebalancing

narqo:
~~I'm still in the process of testing the corned cases against the changes, but so far they look promising~~

The primary idea of the changeset is to replace how a cycle works with the assigned partitions. Since now on, we look at the assignment as a queue, that we traverse start to end. Inside the partition processing, we now allow a rebalance between the subsequent calls to `kafka.PolRrecords`. If Kafka broker calls for rebalance, the partition processing bails, and the cycle continues with the next partition in the newly assigned queue. The "trick" is to make sure the cycle skips the partitions that had been already processed before the rebalance (refer to the implementation of `partitions.update()`).

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -290,7 +290,6 @@ func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			} else {
 				level.Info(b.logger).Log(
 					"msg", "nothing to consume in partition",
-					"err", pl.Err,
 					"part", part,
 					"offset", pl.Commit.At,
 					"end_offset", pl.End.Offset,
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -26,6 +26,9 @@ import (
 	"github.com/grafana/mimir/pkg/util/validation"
 )
 
+// errPartitionLost signals the consumer that its active partition was lost due to group rebalance.
+var errPartitionLost = errors.New("partition lost")
+
 type BlockBuilder struct {
 	services.Service
 
@@ -41,8 +44,10 @@ type BlockBuilder struct {
 	// for testing
 	tsdbBuilder func() builder
 
-	assignmentMu sync.Mutex
-	assignment   map[string][]int32
+	// mu protects the following fields
+	mu               sync.Mutex
+	parts            partitions
+	cancelActivePart context.CancelCauseFunc
 }
 
 func New(
@@ -78,6 +83,7 @@ func New(
 
 func (b *BlockBuilder) starting(ctx context.Context) (err error) {
 	const fetchMaxBytes = 100_000_000
+
 	// Empty any previous artifacts.
 	if err := os.RemoveAll(b.cfg.BlocksStorageConfig.TSDB.Dir); err != nil {
 		return fmt.Errorf("removing tsdb dir: %w", err)
@@ -94,7 +100,7 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
 		kgo.ConsumeTopics(b.cfg.Kafka.Topic),
 		kgo.ConsumerGroup(b.cfg.Kafka.ConsumerGroup),
 		kgo.DisableAutoCommit(),
-		kgo.Balancers(kgo.RoundRobinBalancer()), // TODO(v): figure out the best strategy to assign partitions and handle rebalancing
+		kgo.Balancers(kgo.RoundRobinBalancer()),
 		kgo.BlockRebalanceOnPoll(),
 		kgo.OnPartitionsAssigned(b.handlePartitionsAssigned),
 		kgo.OnPartitionsRevoked(b.handlePartitionsLost),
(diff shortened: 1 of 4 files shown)


pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -275,10 +276,6 @@ func TestBlockBuilder(t *testing.T) {
 		)
 	})
 
-	t.Run("restart scenario when there is a kafka commit", func(t *testing.T) {
-		// TODO(codesome): likely requires a new test because startup uses wall clock
-	})
```
narqo, line 280: We already have a separate test for this one

pkg/blockbuilder/blockbuilder.go
```diff
@@ -242,142 +258,147 @@ func cycleEndAtStartup(interval, buffer time.Duration) time.Time {
...
-	parts := make(map[int32]partitionLag, len(assignmentParts))
-	groupLag := lags[b.cfg.Kafka.ConsumerGroup].Lag
-	for _, part := range assignmentParts {
-		pl, ok := groupLag.Lookup(b.cfg.Kafka.Topic, part)
+		if b.cancelActivePart != nil {
+			b.cancelActivePart(context.Canceled)
+			b.cancelActivePart = nil
+		}
+
+		b.parts.reset()
```
codesome, line 280: I prefer this reset to be done just before the below loop. Feels safer and easier to read.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -242,142 +258,147 @@ func cycleEndAtStartup(interval, buffer time.Duration) time.Time {
...
 
 			CommitRecTs:  commitRecTs,
 			SeenTillTs:   seenTillTs,
 			LastBlockEnd: lastBlockEnd,
 		}
+		if err := b.nextConsumeCycle(ctx, pl, cycleEnd); err != nil {
+			level.Error(b.logger).Log("msg", "failed to consume partition", "err", err, "part", part)
+		}
 	}
+	return nil
```
codesome, line 349: I think we should return `ctx.Err()` here. And if it was the partition lost error, then the `running()` should retry `NextConsumeCycle()` with the new partitions. Else there might be some partitions that will lag for a cycle. (can we test this somehow?)

codesome (commented):
And initial review. I plan to do a deeper review tomorrow.

pkg/blockbuilder/blockbuilder.go
narqo: Caution: at this point, `ctx.Err()` only happens when that service is stopping. Note that the `partition lost` ONLY happens inside the loop above this line.

I think we need a better name for the `ctx` on L285 so it was less confusing.

codesome: Yup. Explicitly different names would be good. Even if it is working correctly now, it will be easy to introduce bugs without noticing.

pkg/blockbuilder/partitions.go
```diff
@@ -0,0 +1,73 @@
...
+	i := p.cur
+	p.cur++
+
+	return p.list[i]
+}
+
+// update replaces the queue with a new list of assigned partitions.
+// It doesn't reset the queue's cursor. That is, the later call to next will return the last active item
+// if the item exists in the new list.
+func (p *partitions) update(list []int32) {
```
codesome, line 30: So, if I am understanding it right, when update() happens during an on-going consumption, say: if the BB was consuming parts 1 3 5, and it was at 3 now. part 1 and 5 were removed and part 2 and 4 were added to this BB. So the partitions will get updated as 3 2 4 with cursor at 3. So the BB will continue consumption and always consume the newly added partitions (2 and 4 here). Right?

I have one concern here. A race is possible: if there is a BB2 that just got done with consumption cycle and gets assigned part 5 from the above BB, now part 5 will lag for a cycle because the above BB had not processed it yet.

I think we should check if a partition is up to date with the BB's consumption cycle by checking its last commit, and do some consumption if it is lagging, when a partition is newly assigned to a BB (this can be in a follow up PR)

pkg/blockbuilder/partitions_test.go
```diff
@@ -0,0 +1,124 @@
...
+	})
+}
+
+func TestPartitions_reset(t *testing.T) {
+	var ps partitions
+
+	ps.update([]int32{1, 2, 3})
+	for i := 0; i < 3; i++ {
+		got := ps.next()
+		require.Positive(t, got)
```
codesome, line 113: Because we know what we expect here, let's compare it with i+1? Even below.

pkg/blockbuilder/partitions.go
narqo: I think you're right about the particular case. But I'd see that as an optimization to explore later.

There is a similar scenario possible, when a BB instance is assigned partitions 1, 3, 5. There is a chance that 3 and 5 won't have anything to consume because the ingesters that own these partitions were downscaled.

If BB finishes its cycle before ingesters are scaled up again, the partitions 3 and 5 will end up lagging for one cycle.

pkg/blockbuilder/partitions_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		for _, want := range []int32{1, 2} {
+			got := ps.next()
+			require.Equal(t, want, got)
+		}
+
+		ps.update([]int32{1, 3, 5})
+		require.Equal(t, []int32{1, 3, 5}, ps.collect())
+
+		for _, want := range []int32{1, 3} {
```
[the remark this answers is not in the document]
narqo: This is a great point: the logic was wrong here.

This wouldn't create troubles in reality, because a cycle knows the timestamp where it ends. I.e. reprocessing the same partition inside the cycle is a noop. Still, awesome you've noticed it 💯

codesome (approved):

PR #8595, opened 2024-07-03
Ensure read path errors use correct HTTP status codes

### LINKED ISSUE #8139: Mimir returns HTTP status 422 in cases where 5xx makes more sense
We often receive a `422` HTTP response from Mimir with the following error message:

```
{
  "status": "error",
  "errorType": "execution",
  "error": "attempted to read series at index 0 from store-gateway chunks stream, but the stream has failed: rpc error: code = Internal desc = loading chunks: block 01HXN90CNDFJGRBZWY5HK5FMV8: get range reader: fetching range [27712000, 27728000]: Get \"<bucket_url>/<tenant_id>%2F01HXN90CNDFJGRBZWY5HK5FMV8%2Fchunks%2F000011\": http2: client connection force closed via ClientConn.Close"
}
```

In this case it seems like a 500-level error is more appropriate given this appears to be a server error, and retrying the same query usually works. Many HTTP clients have the ability to retry 5xx-level error codes automatically, so returning the proper status code prevents additional workarounds.

56quarters:
#### What this PR does

Extend the "error translate" `Queryable` implementation to translate errors returned by any part of the `Queryable` type hierarchy. This ensures that all errors returned by queriers are API errors.

Fixes #8139

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## main / unreleased
 * [ENHANCEMENT] Query-scheduler: Introduce `query-scheduler.use-multi-algorithm-query-queue`, which allows use of an experimental queue structure, with no change in external queue behavior. #7873
+* [BUGFIX] Query-frontend: Ensure that internal errors result in an HTTP 500 response code instead of 422. #8595 
 
 ### Grafana Mimir
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## main / unreleased
 * [ENHANCEMENT] Query-scheduler: Introduce `query-scheduler.use-multi-algorithm-query-queue`, which allows use of an experimental queue structure, with no change in external queue behavior. #7873
-* [BUGFIX] Query-frontend: Ensure that internal errors result in an HTTP 500 response code instead of 422. #8595 
+* [BUGFIX] Query-frontend: Ensure that internal errors result in an HTTP 500 response code instead of 422. #8595
 
 ### Grafana Mimir
 
--- a/integration/querier_test.go
+++ b/integration/querier_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/grafana/e2e"
 	e2ecache "github.com/grafana/e2e/cache"
 	e2edb "github.com/grafana/e2e/db"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/prompb"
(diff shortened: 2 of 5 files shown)


56quarters:
> changes make sense, but I see `TestPrometheusCodec_DecodeResponse_ContentTypeHandling` is failing. A missing content type was treated as an internal error previously. 

Right. I wasn't sure about the correct behavior yet.

> Shouldn't the json `apierror` errors the querier returns be part of the json codec as opposed to no codec?

I'm not sure what you mean. When the querier returns JSON encoded `apierror` instances, it sets the `Content-Type` header and they are handled correctly. However, when querier worker [sends a bare `err.Error()`](https://github.com/grafana/mimir/blob/a9e1d80fdff609edf639be394d17ed114000c24a/pkg/querier/worker/scheduler_processor.go#L273-L282) value back as the HTTP body, it doesn't set the `Content-Type` header. This is why the previous code short-circuited the content type encoders. Maybe it would make sense to have a "fallback" codec in this case when there's no content-type header? Or we fix scheduler/frontend workers to send correctly encoded things.

56quarters:
> > Shouldn't the json `apierror` errors the querier returns be part of the json codec as opposed to no codec?
> 
> I'm not sure what you mean. When the querier returns JSON encoded `apierror` instances, it sets the `Content-Type` header and they are handled correctly. However, when querier worker [sends a bare `err.Error()`](https://github.com/grafana/mimir/blob/a9e1d80fdff609edf639be394d17ed114000c24a/pkg/querier/worker/scheduler_processor.go#L273-L282) value back as the HTTP body, it doesn't set the `Content-Type` header. This is why the previous code short-circuited the content type encoders. Maybe it would make sense to have a "fallback" codec in this case when there's no content-type header? Or we fix scheduler/frontend workers to send correctly encoded things.

Actually, I think it'll be a pretty simple change to always have the scheduler/frontend workers send back errors in a format that can be correctly decoded. I'll pursue that which will allow us to keep "unknown encoding" as an internal error.

dimitarvdimitrov:
> > Or we fix scheduler/frontend workers to send correctly encoded things.
> 
> Actually, I think it'll be a pretty simple change to always have the scheduler/frontend workers send back errors in a format that can be correctly decoded. I'll pursue that which will allow us to keep "unknown encoding" as an internal error.

I agree this looks like the least surprising option

56quarters:
> Actually, I think it'll be a pretty simple change to always have the scheduler/frontend workers send back errors in a format that can be correctly decoded. I'll pursue that which will allow us to keep "unknown encoding" as an internal error.

This is actually _not_ pretty simple. The dichotomy between gRPC errors used by the frontend `GrpcRoundTripper` and responses that represent errors makes it difficult to tell what the correct behavior actually is. I'm going to keep the changes in this PR isolated to the querier error translation `Queryable` implementation and leave the other cleanup for another time.

dimitarvdimitrov (approved):
LGTM

dimitarvdimitrov:
> > Actually, I think it'll be a pretty simple change to always have the scheduler/frontend workers send back errors in a format that can be correctly decoded. I'll pursue that which will allow us to keep "unknown encoding" as an internal error.
> 
> This is actually _not_ pretty simple. The dichotomy between gRPC errors used by the frontend `GrpcRoundTripper` and responses that represent errors makes it difficult to tell what the correct behavior actually is. I'm going to keep the changes in this PR isolated to the querier error translation `Queryable` implementation and leave the other cleanup for another time.

yeah makes sense. The PR is still an improvement

PR #8598, opened 2024-07-04
`grpcurl-query-ingesters` improvements

charleskorn:
#### What this PR does

This PR fixes a number of small issues with `grpcurl-query-ingesters`:

* it is now possible to run the script from any directory
* the script is marked as executable
* I've fixed the incorrect path to `download-chunks-from-ingesters-query.json`
* I've extracted the output directory to a variable
* I've added @pstibrany's improvement to examine all responses in the given file, rather than just the first one

One thing I have not yet solved is running the script without making the following change to `pkg/ingester/client/ingester.proto`:

```diff
-import "github.com/grafana/mimir/pkg/mimirpb/mimir.proto";
+import "pkg/mimirpb/mimir.proto";
```

Without this change, `grpcurl` complains that it cannot resolve `mimir.proto`. @pracucci, you added this tool in https://github.com/grafana/mimir/pull/6523 - does the script work for you without this change? 

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
+++ b/tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
@@ -7,7 +7,7 @@ K8S_NAMESPACE=""
 MIMIR_TENANT_ID=""
 # End of configuration.
 
-SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+SCRIPT_DIR=$(realpath "$(dirname "${0}")")
 OUTPUT_DIR="chunks-dump"
 
 mkdir -p "$OUTPUT_DIR"
--- a/tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
+++ b/tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
@@ -7,7 +7,10 @@ K8S_NAMESPACE=""
 MIMIR_TENANT_ID=""
 # End of configuration.
 
-mkdir -p chunks-dump/
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+OUTPUT_DIR="chunks-dump"
+
+mkdir -p "$OUTPUT_DIR"
 
 # Get list of pods.
 PODS=$(kubectl --context "$K8S_CONTEXT" -n "$K8S_NAMESPACE" get pods --no-headers | grep ingester | awk '{print $1}')
@@ -22,14 +25,14 @@ for POD in $PODS; do
   # Wait some time
   sleep 5
 
-  cat query.json | grpcurl \
+  cat "$SCRIPT_DIR/download-chunks-from-ingesters-query.json" | grpcurl \
     -d @ \
     -H "X-Scope-OrgID: $MIMIR_TENANT_ID" \
   	-proto pkg/ingester/client/ingester.proto \
-  	-import-path . \
-  	-import-path ./vendor \
+  	-import-path "$SCRIPT_DIR" \
+  	-import-path "$SCRIPT_DIR/vendor" \
   	-plaintext \
-  	localhost:9095 "cortex.Ingester/QueryStream" > "chunks-dump/$POD"
+  	localhost:9095 "cortex.Ingester/QueryStream" > "$OUTPUT_DIR/$POD"
 
   kill $KUBECTL_PID
   wait $KUBECTL_PID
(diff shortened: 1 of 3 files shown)


tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
```diff
@@ -22,14 +25,14 @@ for POD in $PODS; do
   # Wait some time
   sleep 5
 
-  cat query.json | grpcurl \
+  cat "$SCRIPT_DIR/download-chunks-from-ingesters-query.json" | grpcurl \
     -d @ \
     -H "X-Scope-OrgID: $MIMIR_TENANT_ID" \
   	-proto pkg/ingester/client/ingester.proto \
```
aknuds1, line 31: Tabs -> spaces
```suggestion
    -proto pkg/ingester/client/ingester.proto \
```

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
```diff
@@ -22,14 +25,14 @@ for POD in $PODS; do
...
-  	-import-path . \
-  	-import-path ./vendor \
+  	-import-path "$SCRIPT_DIR" \
+  	-import-path "$SCRIPT_DIR/vendor" \
   	-plaintext \
-  	localhost:9095 "cortex.Ingester/QueryStream" > "chunks-dump/$POD"
+  	localhost:9095 "cortex.Ingester/QueryStream" > "$OUTPUT_DIR/$POD"
```
aknuds1, line 35:
```suggestion
    localhost:9095 "cortex.Ingester/QueryStream" > "$OUTPUT_DIR/$POD"
```

aknuds1 (commented):
Please see comments.

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
```diff
@@ -22,14 +25,14 @@ for POD in $PODS; do
...
 
-  cat query.json | grpcurl \
+  cat "$SCRIPT_DIR/download-chunks-from-ingesters-query.json" | grpcurl \
     -d @ \
     -H "X-Scope-OrgID: $MIMIR_TENANT_ID" \
   	-proto pkg/ingester/client/ingester.proto \
-  	-import-path . \
-  	-import-path ./vendor \
+  	-import-path "$SCRIPT_DIR" \
+  	-import-path "$SCRIPT_DIR/vendor" \
```
aknuds1, line 33: Tabs -> spaces (was missed during my first review, not sure why)
```suggestion
    -import-path "$SCRIPT_DIR" \
    -import-path "$SCRIPT_DIR/vendor" \
```

aknuds1 (approved):
LGTM modulo nits

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
charleskorn: Good catch, fixed in f60bc81

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
charleskorn: Good catch, fixed in f60bc81

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
charleskorn: Good catch, fixed in f60bc81

tools/grpcurl-query-ingesters/download-chunks-from-ingesters.sh
```diff
@@ -7,7 +7,10 @@ K8S_NAMESPACE=""
 MIMIR_TENANT_ID=""
 # End of configuration.
 
-mkdir -p chunks-dump/
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
```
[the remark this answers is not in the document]
charleskorn: What I have here is what I usually use (for reasons I've long since forgotten... I think there was some edge case to do with symlinks or something like that).

If we use that pattern in other places, I'll just follow that for consistency.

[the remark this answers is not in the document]
charleskorn: Resolved in 3abf2ce

pracucci:
> Without this change, grpcurl complains that it cannot resolve mimir.proto. @pracucci, you added this tool in https://github.com/grafana/mimir/pull/6523 - does the script work for you without this change?

I'm not running this tool since when I introduced it. I barely remember I was doing a manual change in the code, so that's probably the exact thing you're mentioning.

charleskorn:
> > Without this change, grpcurl complains that it cannot resolve mimir.proto. @pracucci, you added this tool in #6523 - does the script work for you without this change?
> 
> I'm not running this tool since when I introduced it. I barely remember I was doing a manual change in the code when using grpcurl but I don't remember if it was related to this tool (maybe it is).

Alright, for the time being I'll add a comment to the script and the README and we can come back and fix this later if it really annoys us (or replace the script with a Go app and remove the problem entirely).

aknuds1 (approved):

PR #8599, opened 2024-07-04
Log chunks information if mimir_continuous_test_sine_wave is queried

krajorama:
#### What this PR does

Detailed logs of what querier does when querying mimir_continuous_test_sine_wave

Fixes N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (commented):
is this doable on some HTTP header instead of only the queried metric?

pkg/querier/block_streaming.go
```diff
@@ -78,19 +85,31 @@ func (bqss *blockStreamingQuerierSeriesSet) Warnings() annotations.Annotations {
 }
 
 // newBlockStreamingQuerierSeries makes a new blockQuerierSeries. Input labels must be already sorted by name.
-func newBlockStreamingQuerierSeries(lbls labels.Labels, seriesIdxStart, seriesIdxEnd int, streamReader chunkStreamReader) *blockStreamingQuerierSeries {
+func newBlockStreamingQuerierSeries(lbls labels.Labels, seriesIdxStart, seriesIdxEnd int, streamReader chunkStreamReader, logger log.Logger, chunkInfo *chunkreplyformatter.ChunkReplyFormatter, lastOne bool, traceId, remoteAddress string) *blockStreamingQuerierSeries {
```
charleskorn, line 88: To make this less of an invasive change, what if `chunkreplyformatter.ChunkReplyFormatter` held the logger and trace ID? Then we wouldn't need to pass these around everywhere.

pkg/querier/blocks_store_queryable.go
```diff
@@ -749,6 +752,8 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
 		streams       []storegatewaypb.StoreGateway_SeriesClient
 	)
 
+	debugContinuousTest := queryLimiter.DebugContinuousTest()
```
charleskorn, line 755: Having this on the `QueryLimiter` seems out of place.

What if we had a "enable debug mode" flag that was added to the context instead?

pkg/querier/blocks_store_queryable.go
```diff
@@ -879,6 +884,18 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
...
+			// debug
+			var chunkInfo *chunkreplyformatter.ChunkReplyFormatter
+			if debugContinuousTest {
+				for _, cm := range convertedMatchers {
+					if cm.Name == "__name__" && cm.Value == "mimir_continuous_test_sine_wave_v2" {
+						chunkInfo = chunkreplyformatter.NewChunkReplyFormatter()
+						break
+					}
+				}
+			}
```
charleskorn, line 896: Rather than specifically checking for the continuous-test series, what if we relied on continuous-test setting a HTTP header (eg. `Mimir-Enable-Query-Debugging: true`) when it makes the query request that enables this logging?

Then we could reuse the same logging for debugging other queries as well if need be.

pkg/querier/distributor_queryable_streaming.go
```diff
@@ -67,6 +82,13 @@ func (s *streamingChunkSeries) Iterator(it chunkenc.Iterator) chunkenc.Iterator
 		chunkBytes += c.Size()
 	}
 
+	if s.chunkInfo != nil {
+		needPrint := s.chunkInfo.EndSeries()
+		if s.lastOne || needPrint {
```
charleskorn, line 87: Rather than having these checks, could we just log a separate line per series? Or would that generate too much output?

pkg/util/chunkreplyformatter/chunkreplyformatter_test.go
```diff
@@ -0,0 +1,123 @@
...
+		{
+			StartTimestampMs: 2000,
+			EndTimestampMs:   4000,
+			Data:             []byte("data2"),
+		},
+	})
+	formatter.EndSeries()
+	expected := `{"series1":{"source1":["1:1:5:57ca2ca6","0:2:5:cec37d1c"]},"series2":{"source1":["1:1:5:57ca2ca6"],"source2":["2:2:5:cec37d1c"]}}`
+	require.Equal(t, expected, formatter.GetChunkInfo())
+	require.True(t, json.Valid([]byte(expected)))
```
charleskorn, line 48: [nit] This could be replaced with a single call to `require.JSONEq()`

pkg/querier/distributor_queryable_streaming.go
krajorama: too much output as both of us experienced

pkg/querier/blocks_store_queryable.go
krajorama: yep, I started from what we had now. We can add it incrementally . I had some wild ideas, but it's a scope creep :)

pkg/querier/blocks_store_queryable.go
```diff
@@ -896,6 +908,14 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
 					"fetched index bytes", indexBytesFetched,
 					"requested blocks", strings.Join(convertULIDsToString(blockIDs), " "),
 					"queried blocks", strings.Join(convertULIDsToString(myQueriedBlocks), " "))
+				if chunkInfo != nil {
+					for i, s := range mySeries {
+						ls := mimirpb.FromLabelAdaptersToLabels(s.Labels)
+						chunkInfo.StartSeries(ls.Get("series_id"))
+						chunkInfo.FormatStoreGatewayChunkInfo(c.RemoteAddress(), s.Chunks)
```
charleskorn, line 915: Given we create a separate `ChunkInfoLogger` for each store-gateway, what do you think about doing something similar per ingester? Then we could set the store-gateway or ingester name when creating the `ChunkInfoLogger`, and wouldn't need to pass the store-gateway or ingester name here. 

This could also simplify things for the ingester logging: we could instead do the logging in `SeriesChunksStreamReader.GetChunks`, which knows when we've received the last series and so we wouldn't need to pass `lastOne` to `streamingChunkSeries`.

pkg/util/chunkinfologger/chunkinfologger.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (c *ChunkInfoLogger) FormatStoreGatewayChunkInfo(sourceID string, chunks []storepb.AggrChunk) {
+	c.formatChunkInfo(sourceID, len(chunks), func(i int, prevTime *int64) string {
+		chunk := chunks[i]
+		startTime := chunk.MinTime/1000 - *prevTime
+		*prevTime = chunk.MaxTime / 1000
+		return fmt.Sprintf("%v:%v:%v:%x", startTime, chunk.MaxTime/1000-chunk.MinTime/1000, len(chunk.Raw.Data), crc32.ChecksumIEEE(chunk.Raw.Data))
+	})
+}
```
charleskorn, line 97: These could call a shared method that does the formatting, passing in the raw chunk bytes, rather than duplicating very similar logic in both methods:

```suggestion
func (c *ChunkInfoLogger) FormatIngesterChunkInfo(sourceID string, chunks []client.Chunk) {
	c.formatChunkInfo(sourceID, len(chunks), func(i int, prevTime *int64) string {
		chunk := chunks[i]
		return formatChunk(chunk.StartTimestampMs, chunk.EndTimestampMs, chunk.Data)
	})
}

func (c *ChunkInfoLogger) FormatStoreGatewayChunkInfo(sourceID string, chunks []storepb.AggrChunk) {
	c.formatChunkInfo(sourceID, len(chunks), func(i int, prevTime *int64) string {
		chunk := chunks[i]
		return formatChunk(chunk.MinTime, chunk.MaxTime, chunk.Raw.Data)
	})
}
```

charleskorn, line 97: Also, I think it'd be clearer if we logged the start timestamp as-is, rather than making it relative to the previous end timestamp - I realise this will increase the size of the messages, but this will be clearer for someone to understand without doing extra processing of the messages logged.

Or we should include a script / tool that formats these messages nicely for human consumption, in which case this is less of an issue.

pkg/util/chunkinfologger/chunkinfologger.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (c *ChunkInfoLogger) SetMsg(msg string) {
+	c.msg = msg
+}
+
+func (c *ChunkInfoLogger) StartSeries(seriesID string) {
+	c.sourceCount = 0
+	if c.chunkInfo.Len() > 0 {
+		c.chunkInfo.WriteString(",\"") // next series
```
charleskorn, line 62: [nit] Might be clearer to use raw strings in cases like this in this file

```suggestion
		c.chunkInfo.WriteString(`,"`) // next series
```

charleskorn (commented):
There are quite a few references to continuous-test in comments, variable names (eg. `debugContinuousTest`) and the messages logged (eg. `CT: chunk series from store-gateway`). Given this is no longer specific to continuous-test, what do you think about renaming them to something like "query debugging"?

pkg/querier/distributor_queryable_streaming.go
```diff
@@ -45,13 +50,20 @@ func (s *streamingChunkSeries) Iterator(it chunkenc.Iterator) chunkenc.Iterator
 	var uniqueChunks []client.Chunk
 	totalChunks := 0
 
+	if s.chunkInfo != nil {
+		s.chunkInfo.StartSeries(s.labels.Get("series_id"))
```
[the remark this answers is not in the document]
krajorama: We're writing a log for each series and continuous test only has one label, so I think in this case we should just dump all labels to be simpler? wdyt?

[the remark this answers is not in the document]
krajorama: Although the series __name__ might be overcall, especially in a known case like continuous test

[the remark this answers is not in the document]
krajorama: change turned out to be not so bad, testing

pkg/querier/blocks_store_queryable.go
krajorama: I think in the case of ingesters we can potentially get data from hundreds, so the current method feels more efficient, less log message clutter.

pkg/util/chunkinfologger/chunkinfologger.go
krajorama: switched to logging the start timestamp as-is

krajorama:
> There are quite a few references to continuous-test in comments, variable names (eg. `debugContinuousTest`) and the messages logged (eg. `CT: chunk series from store-gateway`). Given this is no longer specific to continuous-test, what do you think about renaming them to something like "query debugging"?

Added a k-v called chunkinfo=true to be searchable specifically:
```
mimir-1-1        | ts=2024-07-15T08:20:14.975418287Z caller=distributor_queryable_streaming.go:83 chunkinfo=true 
msg="ingester streaming" traceId=3fd016db223a6426 info="{\"0\":{\"b57b31acb77c\":
[\"1721031280:320:143:45d0081a\"]},\"1\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},\"2\":
{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},\"3\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},
\"4\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},\"5\":{\"b57b31acb77c\":
[\"1721031280:320:143:45d0081a\"]},\"6\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},\"7\":
{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},\"8\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]},
\"9\":{\"b57b31acb77c\":[\"1721031280:320:143:45d0081a\"]}}"
```

tacole02 (approved):

pkg/continuoustest/client.go
```diff
@@ -67,7 +70,7 @@ func (cfg *ClientConfig) RegisterFlags(f *flag.FlagSet) {
 
 	f.Var(&cfg.ReadBaseEndpoint, "tests.read-endpoint", "The base endpoint on the read path. The URL should have no trailing slash. The specific API path is appended by the tool to the URL, for example /api/v1/query_range for range query API, so the configured URL must not include it.")
 	f.DurationVar(&cfg.ReadTimeout, "tests.read-timeout", 60*time.Second, "The timeout for a single read request.")
-
+	f.BoolVar(&cfg.RequestDebug, "tests.request-debug", false, "Request debugging on the server side via header.")
```
charleskorn, line 73: [nit] How about something like `-tests.request-chunks-debugging`? Then it's very clear exactly what this enables.

As it stands, from the name alone, it sounds like continuous-test will log more information about the requests it sends.

pkg/querier/block_streaming.go
```diff
@@ -37,6 +38,10 @@ type blockStreamingQuerierSeriesSet struct {
 	nextSeriesIndex int
 
 	currSeries storage.Series
+
+	// Debug logging for continuous test queries.
```
charleskorn, line 42: This comment still refers to continuous-test.

(same applies to line 103 below)

pkg/querier/distributor_queryable.go
```diff
@@ -129,15 +132,28 @@ func (q *distributorQuerier) streamingSelect(ctx context.Context, minT, maxT int
 		sets = append(sets, newTimeSeriesSeriesSet(results.Timeseries))
 	}
 
+	traceID, _ := tracing.ExtractTraceID(ctx)
```
charleskorn, line 135: Could we move this inside the `if` two lines below? I don't think `traceID` is used anywhere else.

pkg/querier/distributor_queryable.go
```diff
@@ -129,15 +132,28 @@ func (q *distributorQuerier) streamingSelect(ctx context.Context, minT, maxT int
...
+	var chunkInfo *chunkinfologger.ChunkInfoLogger
+	if chunkinfologger.IsChunkInfoLoggingEnabled(ctx) {
+		chunkInfo = chunkinfologger.NewChunkInfoLogger("ingester message", traceID, q.logger, chunkinfologger.ChunkInfoLoggingFromContext(ctx))
+		q.logger.Log("msg", "ingester streamSelect: start", "traceid", traceID, "mint", minT, "maxt", maxT)
```
charleskorn, line 139: Should we move this into `ChunkInfoLogger` and log similar information for store-gateway requests too?

pkg/util/chunkinfologger/chunkinfologger.go
```diff
@@ -0,0 +1,119 @@
...
+		c.chunkInfo.WriteRune('"')
+	}
+	c.endSourceInfo()
+}
+
+func (c *ChunkInfoLogger) startSourceInfo(sourceID string) {
+	if c.sourceCount > 0 {
+		c.chunkInfo.WriteRune(',')
+	}
+	c.sourceCount++
```
charleskorn, line 111: Given we only ever check if `sourceCount` is greater than 0, what if we use a bool instead?

pkg/util/chunkinfologger/debugheader_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+
+	"github.com/grafana/dskit/middleware"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChunkInfoLoggerMiddleware(t *testing.T) {
+	testCases := map[string]struct {
+		headers       http.Header
+		expectedOk    bool
```
charleskorn, line 18: [nit] `expectedLoggingEnabled` might be clearer

(similar thing applies to `recordedOk` below)

pkg/util/chunkinfologger/debugheader_test.go
```diff
@@ -0,0 +1,71 @@
...
+		expectedValue []string
```
charleskorn, line 19: [nit] `expectedLoggingSeriesLabels` might be clearer

(similar thing applies to `recordedValue` below)

pkg/util/chunkinfologger/debugheader.go
```diff
@@ -0,0 +1,47 @@
...
+const ChunkInfoLoggingHeader = "X-Mimir-Chunk-Info-Logger"
+
+func Middleware() middleware.Interface {
+	return middleware.Func(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+			if value := r.Header.Get(ChunkInfoLoggingHeader); value != "" {
+				// Split along commas
+				labels := strings.Split(value, ",")
+				ctx = ContextWithChunkInfoLogging(ctx, labels)
```
charleskorn, line 26: Would it be worthwhile having a wildcard option as well, for logging all the series labels without having to list all of them?

krajorama: sure, in a next PR ;)

pkg/continuoustest/client.go
krajorama: maybe be more explicit and have `-tests.add-debug-header` ?

pkg/util/chunkinfologger/debugheader_test.go
krajorama: suggest: expectedEnabled, expectedLabels and actualEnabled, actualLabels for short

pkg/util/chunkinfologger/debugheader_test.go
krajorama: resolving to keep one discussion for the 4 variables

pkg/continuoustest/client.go
charleskorn: How about `-tests.send-chunks-debugging-header`?

pkg/querier/distributor_queryable.go
krajorama: added

pkg/util/chunkinfologger/debugheader_test.go
charleskorn: Sounds good to me.

pkg/util/chunkinfologger/chunkinfologger.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/grafana/mimir/pkg/storegateway/storepb"
+)
+
+// Allow for 10% overhead from string quotes when logger dumps the JSON.
+const maxSize = (64 * 1024 * 90) / 100
+
+type ChunkInfoLogger struct {
+	chunkInfo strings.Builder
+	msg       string
+	traceID   string
```
charleskorn, line 24: Should we include the span ID as well? In a query with multiple selectors (or a query that gets sharded), the span should help us work out which log lines belong to a single set of ingester or store-gateway requests.

krajorama: I haven't run into that yet, but a quick look at the logs tells me that some continuous test queries are indeed sharded. I'll add it.

krajorama: Added via dskit

pkg/util/chunkinfologger/chunkinfologger_test.go
```diff
@@ -0,0 +1,194 @@
...
+	})
+	formatter.EndSeries(true)
+	expected := `{"series1":{"source1":["1:1:5:57ca2ca6","2:2:5:cec37d1c"]},"series2":{"source1":["1:1:5:57ca2ca6"],"source2":["2:2:5:cec37d1c"]}}`
+	require.Len(t, logger.logs, 1)
+	require.Contains(t, logger.logs[0], "msg")
+	require.Equal(t, "test", logger.logs[0]["msg"])
+	require.Contains(t, logger.logs[0], "traceId")
+	require.Equal(t, "123", logger.logs[0]["traceId"])
+	require.Contains(t, logger.logs[0], "spandId")
+	require.Equal(t, "456", logger.logs[0]["spandId"])
```
charleskorn, line 62:
```suggestion
	require.Contains(t, logger.logs[0], "spanId")
	require.Equal(t, "456", logger.logs[0]["spanId"])
```

pkg/util/chunkinfologger/chunkinfologger_test.go
```diff
@@ -0,0 +1,194 @@
...
+	logger := &testLogger{}
+	formatter := NewChunkInfoLogger("test", "123", "456", logger, []string{"series_id"})
+	formatter.LogSelect("test", 1000, 2000)
+	require.Len(t, logger.logs, 1)
+	require.Contains(t, logger.logs[0], "msg")
+	require.Equal(t, "test", logger.logs[0]["msg"])
+	require.Contains(t, logger.logs[0], "traceId")
+	require.Equal(t, "123", logger.logs[0]["traceId"])
+	require.Contains(t, logger.logs[0], "spandId")
+	require.Equal(t, "456", logger.logs[0]["spandId"])
```
charleskorn, line 172:
```suggestion
	require.Contains(t, logger.logs[0], "spanId")
	require.Equal(t, "456", logger.logs[0]["spanId"])
```

charleskorn (approved):
LGTM modulo test failure due to typo

PR #8601, opened 2024-07-04
Fix the flag names for BlockBuilder

codesome:
The flags were missing a `block-builder.` prefix.

I also fix the lint errors in this PR.


codesome:
Merging for now. Will investigate the test failure separately.

Issue #8389, opened 2024-06-17
store-gateway: refactor BucketStore to have proper module lifecycle


dimitarvdimitrov:
The BucketStore and the ReaderPool it uses need to start goroutines. It's hard to keep track of those and hard to clean them up. #8281 introduced another background goroutine. See https://github.com/grafana/mimir/blob/36f96f19bde5d6e5552600b25893244971b007c8/pkg/util/test/leak.go#L27-L30

The store-gateway would be easier to navigate and less prone to bugs if the BucketStore has proper lifecycle functions like `start`, `run`, and `stop`. Like a lot of other internal services.

This issue is a follow-up for this comment thread https://github.com/grafana/mimir/pull/8281#discussion_r1638174462

dimitarvdimitrov:
related comment from @pracucci https://github.com/grafana/mimir/pull/8281#discussion_r1642887792

dimitarvdimitrov:
The changes became slightly long and hard to follow. I plan to open a few PRs for this


- [ ] Make BucketStore and BucketStores services
- [ ] Make Snapshotter a service 
- [ ] Move eager loading from LazyBinaryReader into BucketStore
- [ ] Make ReaderPool a service
- [ ] Properly close LazyBinaryReader when we are done with it

PR #8602, opened 2024-07-04
store-gateway: Move eager loading to BucketStore

dimitarvdimitrov:
This is the third PR for https://github.com/grafana/mimir/issues/8389

#### What this PR does

This PR does one major change: Move eager loading from `ReaderPool` to `BucketStore`. 

The problem with keeping eager loading in the `ReaderPool` is that it needs to have the list of previously loaded blocks injected. This is hard if we want to move the snapshotter and index reader pool to being services.

The PR also slightly changes the `Snapshotter` to receive its dependency (`BlocksLoader` which has the list of currently loaded blocks) via the constructor as opposed to via each snapshot iteration. This is now possible because there is no circular dependency between the `Snapshotter` and the `ReaderPool` (the snapshotter needed to pool to know what's loaded; the pool needed the snapshotter to know what to eager load).

related to https://github.com/grafana/mimir/issues/8389

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket.go
+++ b/pkg/storegateway/bucket.go
@@ -239,17 +239,7 @@ func NewBucketStore(
 		option(s)
 	}
 
-	lazyLoadedBlocks, err := indexheader.RestoreLoadedBlocks(dir)
-	if err != nil {
-		level.Warn(s.logger).Log(
-			"msg", "loading the list of index-headers from snapshot file failed; not eagerly loading index-headers for tenant",
-			"dir", dir,
-			"err", err,
-		)
-		// Don't fail initialization. If eager loading doesn't happen, then we will load index-headers lazily.
-		// Lazy loading which is slower, but not worth failing startup for.
-	}
-	s.indexReaderPool = indexheader.NewReaderPool(s.logger, bucketStoreConfig.IndexHeader, s.lazyLoadingGate, metrics.indexHeaderReaderMetrics, lazyLoadedBlocks)
+	s.indexReaderPool = indexheader.NewReaderPool(s.logger, bucketStoreConfig.IndexHeader, s.lazyLoadingGate, metrics.indexHeaderReaderMetrics)
 
 	if bucketStoreConfig.IndexHeader.EagerLoadingStartupEnabled {
 		snapConfig := indexheader.SnapshotterConfig{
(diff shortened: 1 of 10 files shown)


dimitarvdimitrov:
this is now ready for a review

pkg/storegateway/bucket.go
```diff
@@ -385,10 +372,44 @@ func (s *BucketStore) syncBlocks(ctx context.Context, initialSync bool) error {
...
+
+func (s *BucketStore) loadBlocks(ctx context.Context, blocks map[ulid.ULID]int64) {
+	// We ignore the time the block was used because it can only be in the map if it was still loaded before the shutdown
+	s.blockSet.forEach(func(b *bucketBlock) {
+		if _, ok := blocks[b.meta.ULID]; !ok {
+			return
+		}
+		if lazyReader, ok := b.indexHeaderReader.(*indexheader.LazyBinaryReader); ok {
+			lazyReader.EagerLoad(ctx)
+		}
```
narqo, line 408: It seems that now we can use `b.ensureIndexHeaderLoaded` to avoid branching on `(*indexheader.LazyBinaryReader)`? This could make the `BucketStore` internals a bit more abstract and simpler to follow.

pkg/storegateway/bucket.go
```diff
@@ -385,10 +372,44 @@ func (s *BucketStore) syncBlocks(ctx context.Context, initialSync bool) error {
...
+	previouslyLoadedBlocks, err := indexheader.RestoreLoadedBlocks(s.dir)
+	if err != nil {
+		level.Warn(s.logger).Log(
+			"msg", "loading the list of index-headers from snapshot file failed; not eagerly loading index-headers for tenant",
+			"dir", s.dir,
+			"err", err,
+		)
+		// Don't fail initialization. If eager loading doesn't happen, then we will load index-headers lazily.
+		// Lazy loading which is slower, but not worth failing startup for.
+	}
```
narqo, line 384: How about we move this chunk inside `s.loadBlocks` (and maybe rename this method to `s.restorePreviousBlocks`)? We don't need to do it if `EagerLoadingStartupEnabled` is disabled. And also this will make the piece about "not failing the initialization" less awkward (IMHO).

pkg/storegateway/gateway.go
```diff
@@ -253,6 +253,7 @@ func (g *StoreGateway) starting(ctx context.Context) (err error) {
 	// At this point, the instance is registered with some tokens
 	// and we can start the bucket stores.
 	g.bucketSync.WithLabelValues(syncReasonInitial).Inc()
+	g.logger.Log("msg", "starting bucket initial sync")
```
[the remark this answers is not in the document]
dimitarvdimitrov: haha, yes, this was me debugging the code

pkg/storegateway/bucket.go
dimitarvdimitrov: good idea, did that

pkg/storegateway/bucket.go
dimitarvdimitrov: after we run syncBlocks the snapshotter will be running, so there's a theoretical race between the snapshotter and reading the snapshot from before the shutdown. It's not obvious, so I now added a comment explaining this and moved this whole thing into a function.

But if you have ideas for how to make this better I can do it

narqo: It makes more sense now, thanks for explaining. I'm ok to stick to the current version. I think, the comments you've added makes it clear 👍

narqo (approved):
It's even better now 🔥🔥

PR #8607, opened 2024-07-04
Commit properly when we have to retain the old commit

codesome:
When commit record stays same as the previous cycle, we were handling it wrongly. Instead of keeping the old commit, we were taking the `lastRec` as the `commitRec` because of how `commitRec` was set. We identify that the first record was not fully processed, so the `commitRec` is set to the `lastRec` which is nil. So `commitRec` got wrongly overridden by the `lastRec` in this case.

In this PR what I am doing is passing around the last commit information. And when the first record is the one that is not fully processed, I am setting `commitRec` with the old commit info. That way comitting it correctly. We do need to re-commit in order to update the "last seen" timestamp in the metadata.

This was why the out-of-order test was failing (test was checking the right stuff apparently).

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -339,10 +339,10 @@ func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			level.Warn(b.logger).Log("msg", "error unmarshalling commit metadata", "err", err, "part", part, "offset", offset.At, "metadata", offset.Metadata)
 		}
 
-		pl := partitionLag{
-			Partition: part,
-			Lag:       lag.Lag,
-
+		pl := partitionInfo{
+			Partition:    part,
+			Lag:          lag.Lag,
+			Commit:       offset,
 			CommitRecTs:  commitRecTs,
 			SeenTillTs:   seenTillTs,
 			LastBlockEnd: lastBlockEnd,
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -516,6 +521,19 @@ func (b *BlockBuilder) consumePartition(
...
+				// this is the commit point.
+				if commitRec == nil {
+					// When the commit record does not change from the previous commit, we get nil here.
+					commitRec = &kgo.Record{
+						Timestamp:   time.UnixMilli(pl.CommitRecTs),
+						Topic:       lastCommit.Topic,
+						Partition:   lastCommit.Partition,
+						LeaderEpoch: lastCommit.LeaderEpoch,
+						Offset:      lastCommit.At - 1,
+					}
```
narqo, line 535: Nice catch. Just as an alternative, instead of building an artificial commit, can we `continue` here, leaving the `commitRec` and `lastRec` to be `nil`? In this case, the code below will skip doing any commit, and just rewind the offset to the first record (ref [code][1]).

[1]: https://github.com/grafana/mimir/blob/558ff24f6d4946aca7e94f64f2641ec32cbdb5b6/pkg/blockbuilder/blockbuilder.go#L553-L561

codesome: From the PR description
> We do need to re-commit in order to update the "last seen" timestamp in the metadata.

narqo (approved):
🔥

PR #8603, opened 2024-07-04
store-gateway: Make index reader pool a service

dimitarvdimitrov:
This is the fourth PR for https://github.com/grafana/mimir/issues/8389

#### What this PR does

Makes `ReaderPool` a `service.Service`. Its lifecycle is controlled by the `BucketStore`. The `BucketStore` will stop the `ReaderPool` whenever it is shutting down. This prevents leaking goroutines.

#### Which issue(s) this PR fixes or relates to

related to https://github.com/grafana/mimir/issues/8389

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8610, opened 2024-07-04
store-gateway: properly close LazyBinaryReader on shutdown

dimitarvdimitrov:
This is the fifth and final PR for https://github.com/grafana/mimir/issues/8389. The base branch is that of https://github.com/grafana/mimir/pull/8603 and the idea is that we merge this PR after https://github.com/grafana/mimir/pull/8603.

This is a small PR, but I decided to not combine it with https://github.com/grafana/mimir/pull/8603 as in theory does more than refactor. See below.

#### What this PR does

Properly closes `LazyBinaryReader` and cleans up the control loop goroutine.

As part of that I realized that `LazyBinaryReader`'s `Close` claims that calling `Close` and later using the reader again is valid. However, the other implementation of `indexheader.Reader` - `StreamBinaryReader` does not satisfy the same contract - calling `Close` makes `StreamBinaryReader` unusable. Moreover, there is no code which actually uses this feature of `LazyBinaryReader` - an `indexheader.Reader` is never `Close`d in a couple of places.

Here are all the places which end up calling `indexheader.Reader.Close()`:
1.  [when the lazy reader is unloaded](https://github.com/grafana/mimir/blob/05dee32601c3f1832307b15f10e78212ebac43f6/pkg/storegateway/indexheader/lazy_binary_reader.go#L387-L393) - this calls `StreamBinaryReader.Close()` and discards it, so the reader cannot be used again. There are also no pending readers when this is happening.
2. [when the block is being deleted from disk](https://github.com/grafana/mimir/blob/05dee32601c3f1832307b15f10e78212ebac43f6/pkg/storegateway/bucket.go#L549-L579) - the index header is deleted from disk and we also lose the reference to the `bucketBlock`, so no chance that we need the `LazyBinaryReader` again
3. [when the store-gateway (and BucketStore) is shutting down](https://github.com/grafana/mimir/blob/05dee32601c3f1832307b15f10e78212ebac43f6/pkg/storegateway/bucket.go#L589-L591) - in this case the BucketStore doesn't need to query the block again because there will be no more queries
4. [when the BucketStore no longer owns the given tenant and is deleting all blocks for the tenant](https://github.com/grafana/mimir/blob/6bf07eb1330ce5b3b7f29c00c4dcaf871a106d73/pkg/storegateway/bucket_stores.go#L418-L442) - similar reasoning to case 2. - disk is being cleaned up too - the reader won't be used.
 

closes https://github.com/grafana/mimir/issues/8389

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket.go
+++ b/pkg/storegateway/bucket.go
@@ -260,25 +260,17 @@ func NewBucketStore(
 	return s, nil
 }
 
-func (s *BucketStore) subservices() []services.Service {
-	return []services.Service{s.snapshotter, s.indexReaderPool}
-}
-
 func (s *BucketStore) start(context.Context) error {
 	// Use context.Background() so that we stop the index reader pool ourselves and do it after closing all blocks.
 	return services.StartAndAwaitRunning(context.Background(), s.indexReaderPool)
 }
 
 func (s *BucketStore) stop(err error) error {
-	subservices := s.subservices()
-
 	errs := multierror.New(err)
-	for _, svc := range subservices {
-		if err := services.StopAndAwaitTerminated(context.Background(), svc); err != nil {
-			errs.Add(fmt.Errorf("stop %T: %w", svc, err))
-		}
-	}
-
+	errs.Add(s.closeAllBlocks())
+	// The snapshotter depends on the reader pool, so we close the snapshotter first.
+	errs.Add(services.StopAndAwaitTerminated(context.Background(), s.snapshotter))
+	errs.Add(services.StopAndAwaitTerminated(context.Background(), s.indexReaderPool))
 	return errs.Err()
 }
 
@@ -349,7 +341,7 @@ func (s *BucketStore) syncBlocks(ctx context.Context) error {
 		return metaFetchErr
 	}
 
-	blockIDs := s.blockSet.blockULIDs()
+	blockIDs := s.blockSet.openBlocksULIDs()
 	for _, id := range blockIDs {
 		if _, ok := metas[id]; ok {
 			continue
@@ -534,8 +526,12 @@ func (s *BucketStore) removeBlock(id ulid.ULID) (returnErr error) {
 	return nil
 }
 
+func (s *BucketStore) closeAllBlocks() error {
+	return s.blockSet.closeAll()
+}
+
 func (s *BucketStore) removeAllBlocks() error {
-	blockIDs := s.blockSet.blockULIDs()
+	blockIDs := s.blockSet.allBlockULIDs()
 
 	errs := multierror.New()
 	for _, id := range blockIDs {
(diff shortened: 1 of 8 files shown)


dimitarvdimitrov:
this is now ready for a review

dimitarvdimitrov:
@narqo will you have time to give it a review? I think this is the trickiest of the 5 PRs for this refactor, so it might take a bit longer

pkg/storegateway/bucket.go
```diff
@@ -1903,14 +1899,38 @@ func (s *bucketBlockSet) forEach(fn func(b *bucketBlock)) {
 	})
 }
 
-func (s *bucketBlockSet) blockULIDs() []ulid.ULID {
+// closeAll closes all blocks in the set and returns all encountered errors after trying all blocks.
+func (s *bucketBlockSet) closeAll() error {
+	errs := multierror.New()
+	s.blockSet.Range(func(_, val any) bool {
+		errs.Add(val.(*bucketBlock).Close())
```
narqo, line 1906: To double-check, I got the point in the description right: `bucketBlock.Close()` is expected to be idempotent, right? 

It seems that we can (potentially) call it from several places. Also, looks like the change in `*LazyBinaryReader.Close()` below (about `close(done)` channel) introduces a potential panic, if multiple goroutines tries to close the `indexHeaderReader`. Is this not a problem here?

dimitarvdimitrov: sorry for the delay @narqo 

> To double-check that I got the points from the description right: `bucketBlock.Close()` is expected to be idempotent, right?

I didn't check about this. So I had to review it again. I don't think `bucketBlock.Close()` is expected to be idempotent. There are multiple code paths that can close the block, but I believe they are mutually exclusive:

1. [when the block is being removed from disk via `BucketStore.removeBlock()`](https://github.com/grafana/mimir/blob/a54091cedb143c0f6758c2df087ad2b564699d7f/pkg/storegateway/bucket.go#L510-L522) (scenario 2. and 4. from the description). In this case the block is also removed from the block set - `s.blockSet.remove(id)`, so we can't call `bucketBlock.Close()` multiple times even if  `BucketStore.removeBlock()` is called multiple times
2. [when the BucketStore is being shut down, so all blocks need to be closed via `bucketBlockSet.closeAll()`](https://github.com/grafana/mimir/blob/a54091cedb143c0f6758c2df087ad2b564699d7f/pkg/storegateway/bucket.go#L1906) (scenario 3. in the description). This runs once in the process' lifetime.

1\. can never run after 2. and 2. cannot discover the block if 1. has previously ran. 1. and 2. can also never run concurrently. 1. can only happen as a result of the periodic `StoreGateway.running > BucketStores.SyncBlocks > BucketStore.SyncBlocks`. 2. can only happen as a result of `StoreGateway.stopping > BucketStores.stopBucketStores > BucketStore.stop`. `StoreGateway.running` and `StoreGateway.stopping` can't be running concurrently.


It's a long-winded answer, but I hope it was clear. Happy to jump on a call or chat on slack

dimitarvdimitrov: > 1. ...
> 2. ...

the way I discovered this is by renaming `bucketBlock.Close()` to `bucketBlock.CloseBlock()` in my IDE and finding all usages. This prevents the name conflict with `io.Closer` and shows these two usages only.

narqo (approved):
Thank you, for explaining the code flow 🔥

PR #8630, opened 2024-07-05
continuous-test improvements

charleskorn:
#### What this PR does

This PR makes a number of small improvements to continuous-test:

* It changes the test series to not pass through 0. This makes it easier to tell if a series is missing or incorrect at timesteps that we would otherwise expect to have a 0 value.
* It initialises most of the test-specific metrics to 0 at startup. This ensures that the `MimirContinuousTestFailed` alert fires for the first failure after continuous-test starts.
* It adds human-friendly timestamps to the diffs included in failure messages.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@
 
 ### Mimir Continuous Test
 
+* [CHANGE] Use test metrics that do not pass through 0 to make identifying incorrect results easier. #8630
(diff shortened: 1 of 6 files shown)


aknuds1 (approved):
LGTM, thanks!

PR #8632, opened 2024-07-05
Update dskit and mimir-prometheus

pracucci:
#### What this PR does

In this PR I'm updating dskit and mimir-prometheus to get:
- https://github.com/grafana/dskit/pull/538
- https://github.com/grafana/mimir-prometheus/pull/658 changes

In this PR we're also introducing the new experimental PromQL functions `limitk()` and `limit_ratio()` done by @jjo. I've marked them as non shardable, just because I didn't want to spend an effort right now to brainstorm whether they're shardable or not and then add tests on them (they're disabled by default). We can revisit this later.

N/A

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.


pkg/frontend/querymiddleware/astmapper/parallel.go
```diff
@@ -28,6 +28,8 @@ var NonParallelFuncs = []string{
 	"absent",
 	"absent_over_time",
 	"histogram_quantile",
+	"limitk",
```
colega, line 31: I _think_ `limitk` is safe to parallelize, just like topk/bottomk.

Maybe it's worth creating a new `good-first-issue` issue and let someone send a PR adding some tests for it?

colega (approved):

pkg/frontend/querymiddleware/astmapper/parallel.go
pracucci: Topk and bottok are not parallelizeable as well ([see details in this internal ticket](https://github.com/grafana/mimir-squad/issues/270#issuecomment-918920983)).

Why they're not mentioned here? Right now I don't remember all implementation details of how these functions are guaranteed not to be parallelized. I will investigate it further, but will not block this PR on this.

jjo:
> #### What this PR does
> In this PR I'm updating dskit and mimir-prometheus to get:
> 
> * [Add Backoff.ErrCause() dskit#538](https://github.com/grafana/dskit/pull/538)
> * [Sync upstream prometheus mimir-prometheus#658](https://github.com/grafana/mimir-prometheus/pull/658) changes
> 
> In this PR we're also introducing the new experimental PromQL functions `limitk()` and `limit_ratio()` done by @jjo. I've marked them as non shardable, just because I didn't want to spend an effort right now to brainstorm whether they're shardable or not and then add tests on them (they're disabled by default). We can revisit this later.
> 
>[...]

woOT!, thanks @pracucci 🎉

PR #8633, opened 2024-07-05
Add prefix field to images.libsonnet to allow adding a prefix to all image names to point to a custom registry

liam-howe-maersk:
#### What this PR does

Adds a `prefix` field to the `_images` object so that we can prefix all of the images with a custom registry URL. We want to fetch all of our images from a private container registry, this change will allow us to do this.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
thanks. LGTM

pstibrany:
I would expect that this is pretty simple to customize already, without needing `prefix` field. (With object comprehension, updating all fields)

Admittedly, when I tried to put the code together quickly to do such change, I failed. I'm sure @colega would do it in few minutes though :)

narqo:
I agree with @pstibrany. One should be able to amend the images with the following:

```
{
  local prefix = 'registry.internal/',
  _images+: std.mapWithKey(function(k, v) { [k]: prefix + v }, super._images, )
}
```

colega:
Not only _can_ fix it with prefix, but actually the version that @narqo suggests is more robust as it would add prefix to all previously overridden images as well, while the proposed solution in the PR only applies to these specific settings, and doesn't apply to future overrides.

liam-howe-maersk:
I also tried to find a way to do this without needing an update to the jsonnet in this repo but admittedly my jsonnet knowledge is fairly basic and I couldn't figure it out. Appreciate the help with this and happy to find an approach I can take on our end, however, applying the suggested change results in a map where the key is repeated, nested in the original key. To demonstrate what I mean, the resulting `_images` object looks something like

```
{
  "_images": {
    "alertmanager": {
      "alertmanager": "registry.internal/grafana/mimir:2.11.0"
    },
    "compactor": {
      "compactor": "registry.internal/grafana/mimir:2.11.0"
    },
    "continuous_test": {
      "continuous_test": "registry.internal/grafana/mimir-continuous-test:main-8a8648e81"
    },
    "distributor": {
      "distributor": "registry.internal/grafana/mimir:2.11.0"
    }
...
```

Which means that the image in the k8s resources looks like

```
image:
     compactor: harbor.maersk.io/dockerhub-proxy/grafana/mimir:2.11.0
```

Is there a way to ensure the keys of the object are not duplicated like this when applying the mapping function?

liam-howe-maersk:
Ok I figured it out, updated to 

```
{
  local prefix = 'registry.internal/',
  _images+: std.mapWithKey(function(k, v) prefix + v, super._images, )
}
```

This looks like it should work for us so happy with this approach and to close this PR as well.

PR #8634, opened 2024-07-05
query-tee: always return response from preferred backend if one is configured

charleskorn:
#### What this PR does

This PR changes the behaviour of query-tee to always return the response from the preferred backend, if one is configured.

Previously, query-tee would only return the response from the preferred backend if it succeeded, or if all other requests failed and the response from the preferred backend was the first response received. This behaviour was somewhat surprising, and doesn't fit with the scenarios we usually use query-tee for, such as safely comparing a change to the query path.

Now, even if the response from the preferred backend is not successful, it will still be returned.

(none)

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@
 
 ### Query-tee
 
-* [CHANGE] If a preferred backend is configured, then query-tee will now always return its response, regardless of the response status code. Previously, query-tee would only return the response from the preferred backend if it did not have a 5xx status code. #8634
+* [CHANGE] If a preferred backend is configured, then query-tee always returns its response, regardless of the response status code. Previously, query-tee would only return the response from the preferred backend if it did not have a 5xx status code. #8634
 * [ENHANCEMENT] Emit trace spans from query-tee. #8419
 * [ENHANCEMENT] Log trace ID (if present) with all log messages written while processing a request. #8419
 * [ENHANCEMENT] Log user agent when processing a request. #8419
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@
 
 ### Query-tee
 
+* [CHANGE] If a preferred backend is configured, then query-tee will now always return its response, regardless of the response status code. Previously, query-tee would only return the response from the preferred backend if it did not have a 5xx status code. #8634
 * [ENHANCEMENT] Emit trace spans from query-tee. #8419
 * [ENHANCEMENT] Log trace ID (if present) with all log messages written while processing a request. #8419
 * [ENHANCEMENT] Log user agent when processing a request. #8419
--- a/docs/sources/mimir/manage/tools/query-tee.md
+++ b/docs/sources/mimir/manage/tools/query-tee.md
@@ -112,7 +112,7 @@ When a preferred backend is not configured, the query-tee uses the following alg
 1. If no backend response status code is 2xx or 4xx, the query-tee selects the first received response regardless of the status code.
 
 {{< admonition type="note" >}}
-The query-tee considers a 4xx response as a valid response to select because a 4xx status code generally an invalid request and not a server-side issue.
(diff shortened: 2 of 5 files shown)


aknuds1:
Could you provide context for why this change in behaviour is desired?

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@
 
 ### Query-tee
 
+* [CHANGE] If a preferred backend is configured, then query-tee will now always return its response, regardless of the response status code. Previously, query-tee would only return the response from the preferred backend if it did not have a 5xx status code. #8634
```
tacole02, line 73:
```suggestion
* [CHANGE] If a preferred backend is configured, then query-tee always returns its response, regardless of the response status code. Previously, query-tee would only return the response from the preferred backend if it did not have a 5xx status code. #8634
```

docs/sources/mimir/manage/tools/query-tee.md
```diff
@@ -104,19 +104,15 @@ The query-tee returns the `Content-Type` header, HTTP status code, and body of t
 The preferred backend can be configured via `-backend.preferred=<hostname>`.
 The value of the preferred backend configuration option must be the hostname of one of the configured backends.
 
-When a preferred backend is configured, the query-tee uses the following algorithm to select the backend response to send back to the client:
-
-1. If the preferred backend response status code is 2xx or 4xx, the query-tee selects the response from the preferred backend.
-1. If at least one backend response status code is 2xx or 4xx, the query-tee selects the first received response whose status code is 2xx or 4xx.
-1. If no backend response status code is 2xx or 4xx, the query-tee selects the first received response regardless of the status code.
+When a preferred backend is configured, the query-tee always returns the response from the preferred backend.
```
tacole02, line 107:
```suggestion
When a preferred backend is configured, the query-tee always returns a response from the preferred backend.
```

docs/sources/mimir/manage/tools/query-tee.md
```diff
@@ -104,19 +104,15 @@ The query-tee returns the `Content-Type` header, HTTP status code, and body of t
...
 
 When a preferred backend is not configured, the query-tee uses the following algorithm to select the backend response to send back to the client:
 
 1. If at least one backend response status code is 2xx or 4xx, the query-tee selects the first received response whose status code is 2xx or 4xx.
 1. If no backend response status code is 2xx or 4xx, the query-tee selects the first received response regardless of the status code.
 
 {{< admonition type="note" >}}
-The query-tee considers a 4xx response as a valid response to select because a 4xx status code generally an invalid request and not a server side issue.
+The query-tee considers a 4xx response as a valid response to select because a 4xx status code generally an invalid request and not a server-side issue.
```
tacole02, line 115:
```suggestion
The query-tee considers a 4xx response as a valid response to select because a 4xx status code is generally an invalid request and not a server-side issue.
```

tacole02 (approved):
Thank you!

docs/sources/mimir/manage/tools/query-tee.md
aknuds1: I think "returns the response" is more technically correct here.

charleskorn:
@aknuds1:

> Could you provide context for why this change in behaviour is desired?
> 
> It sounds sensible, but I don't know enough about the usage of query-tee to tell.

I've added some extra explanation to the PR description:

> This behaviour was somewhat surprising, and doesn't fit with the scenarios we usually use query-tee for, such as safely comparing a change to the query path.

aknuds1 (approved):
LGTM

docs/sources/mimir/manage/tools/query-tee.md
charleskorn: Agree with Arve - there is only one response from the preferred backend, so "returns the response" makes more sense here.

Issue #8390, opened 2024-06-17
Release 2.13


dimitarvdimitrov:
### Publish the release candidate (rc.0)

- [x] Begin drafting the [release notes](https://github.com/grafana/mimir/blob/main/RELEASE.md#write-release-notes-document) https://github.com/grafana/mimir/pull/8445
  - Create the release notes PR targeting the main branch
  - This step shouldn't block from publishing release candidate
  - After the release notes PR is merged (which usually happen after RC is published), cherry-pick them into the release branch
- [x] Wait for any open PR we want to get merged before cutting the release candidate
  - We shouldn't wait for the open PRs beyond the scheduled release date
  - [x] Eventually open a PR for every experimental feature we want to promote to stable https://github.com/grafana/mimir/pull/8400
  - [x] Eventually open a PR to remove any deprecated feature or configuration option that should be removed in this release   https://github.com/grafana/mimir/pull/8395
- [x] Update `CHANGELOG.md` https://github.com/grafana/mimir/pull/8459
  - [x] Run `./tools/release/check-changelog.sh LAST-RELEASE-TAG...main` and add missing PRs to CHANGELOG
  - [x] Ensure CHANGELOG entries are [sorted by type](https://github.com/grafana/mimir/blob/main/docs/internal/contributing/README.md#changelog)
  - [x] Add a new section for the new release so that `## main / unreleased` is blank and at the top. The new section should say `## x.y.0-rc.0`.
- [x] Run `./tools/release/notify-changelog-cut.sh`
- [x] Run `make mixin-screenshots`
  - Before opening the PR, review all updated screenshots and ensure no sensitive data is disclosed https://github.com/grafana/mimir/pull/8470
- [x] Create new release branch
  - [x] Create the branch
    ```bash
    git checkout r<xxx> # xxx is the latest weekly release
    git checkout -b release-<version>
    git push -u origin release-<version>
    ```
  - [x] Remove "main / unreleased" section from the CHANGELOG https://github.com/grafana/mimir/pull/8477
  - [x] If a new minor or major version is being released, adjust the settings in the `renovate.json` configuration on the `main` branch by adding the new version.
        This way we ensure that dependency updates maintain the new version, as well as the latest two minor versions.
        For instance, if versions 3.0 and 2.10 are configured in `renovate.json`, and version 3.1 is being released,
[...]

PR #8641, opened 2024-07-05
helm 5.4.0: cut final release

dimitarvdimitrov:

### DIFF
--- a/docs/sources/helm-charts/mimir-distributed/_index.md
+++ b/docs/sources/helm-charts/mimir-distributed/_index.md
@@ -8,7 +8,7 @@ keywords:
   - Grafana Enterprise Metrics
   - Grafana metrics
 cascade:
-  MIMIR_DOCS_VERSION: "next"
+  MIMIR_DOCS_VERSION: "2.13.x"
   gem_docs_version: "v2.12.x"
 refs:
   grafana-mimir:
--- a/docs/sources/helm-charts/mimir-distributed/_index.md
+++ b/docs/sources/helm-charts/mimir-distributed/_index.md
@@ -8,7 +8,7 @@ keywords:
   - Grafana Enterprise Metrics
   - Grafana metrics
 cascade:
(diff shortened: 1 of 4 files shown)


docs/sources/helm-charts/mimir-distributed/_index.md
```diff
@@ -8,7 +8,7 @@ keywords:
   - Grafana Enterprise Metrics
   - Grafana metrics
 cascade:
-  MIMIR_DOCS_VERSION: "next"
+  MIMIR_DOCS_VERSION: "2.13.x"
```
dimitarvdimitrov, line 11:
```suggestion
  MIMIR_DOCS_VERSION: "v2.13.x"
```

dimitarvdimitrov:
hm, actually maybe we want to wait for GEM 2.13. Keep in draft until then

tacole02 (approved):

QuentinBisson:
Hey @dimitarvdimitrov do you know when we can expect the release of 5.5.0 to be out?

dimitarvdimitrov:
probably in the next couple of weeks; you can follow #9382

QuentinBisson:
Thank you. Do you think it would be possible in the future to have the helm chart releases decoupled from the mimir releases without having to use weekly charts as it's done with Loki ? (I'm not sure where I should bring such topics, sorry about that)

dimitarvdimitrov:
there's nothing stopping us from cutting new helm releases out of sync with Mimir releases. The community call is tomorrow, you can drop by and discuss it there. Otherwise a new issue might be a suitable place to start discussion. 

My 2c is that we're not super strict about when we make helm releases, so if there's a need for an earlier stable release, we can decide on a case-by-base basis.

QuentinBisson:
Ah alright then I will ask next time for the new release. Sadly, I cannot join the community calls because I have kids duty at this time, or I would join every month :D

PR #8642, opened 2024-07-05
Remove 2.13 release from renovate config

dimitarvdimitrov:

we keep only 2 releases updated


narqo (approved):
🔥

PR #8643, opened 2024-07-05
fix(helm): only set blocks_storage.backend=s3 when using minio

### LINKED ISSUE #6538: HELM: Trying to deploy on local DEV Openshift (no S3) fails
Trying to do a simple install on my local DEV Openshift (running with CRC). I have no S3 

The only value I set is 

global:
  openshift:
    enabled: true

#### Describe the bug
I'm getting crashloopback on multiple components with 'No s3 endpoint in config file'

#### To Reproduce
helm -n mimir install mimir grafana/mimir-distributed -f local.yaml

Where the local.yaml has only 
global:
  openshift:
    enabled: true

#### Expected behavior
Successful deployment

#### Environment
Openshift 4.13.6 local env VM running on CRC

echozio:
#### What this PR does
Changes the config template in `values.yaml` to only set `blocks_storage.backend=s3` when using MinIO, allowing it to fall through to `common.storage.backend`. Currently you have to set both to be able to use a different backend (e.g. gcs).

Fixes #6538

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov:
thanks for the change, but unfortunately that's a breaking change. I think we tried to introduce the same at some point, but realized that this leaves the Mimir default of `filesystem` as the backed. I suspect quite a lot of people are relying on the explicit value in the default values.yaml.

PR #8652, opened 2024-07-08
Updated metrics and logs for blockbuilder

codesome:
- Updating the lag metric more frequently
    - During partition consumption
    - Every 5 mins when consumption is not going on
- Add a metric to track time taken for compaction and upload
- Add logs for compaction and upload
- Made the processPartitionDuration metric to be per partition
- More info in some existing logs

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -78,8 +78,6 @@ func New(
 
 	b.Service = services.NewBasicService(b.starting, b.running, b.stopping)
 
-	// TODO(codesome): add a shipping subservice responsible for shipping the blocks to the storage.
-
 	return b, nil
 }
 
@@ -224,8 +222,6 @@ func (b *BlockBuilder) stopping(_ error) error {
 		b.kafkaClient.Close()
 	}
 
-	// TODO(codesome): wait for any active consume cycle or abort it.
-
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -492,6 +543,12 @@ func (b *BlockBuilder) consumePartition(
 
 		recIter := fetches.RecordIter()
 		for !recIter.Done() {
+			metricUpdate++
+			if metricUpdate%10000 == 0 {
+				consumerLagMetric.Set(float64(lag))
+				metricUpdate = 0
+			}
```
narqo, line 550: I'm also no sure there is a need to track the lag from inside the cycle. The lag isn't static, i.e. while cycle is going on Kafka producer keep producing messages, thus I think this is misleading.

What we care about from the inside of the block builder's cycle is how fast we consume + process records.

Also note that we can monitor per-consumer group lag from Kafka itself, as part of how we observe Kafka cluster.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -541,9 +598,13 @@ func (b *BlockBuilder) consumePartition(
 		b.kafkaClient.AllowRebalance()
 	}
 
+	consumerLagMetric.Set(float64(lag))
+	start := time.Now()
 	if err := builder.compactAndUpload(ctx, b.blockUploaderForUser); err != nil {
```
narqo, line 603: I'd move this inside the `builder` (and also I would add more metrics to the builder itself). I.e. I'd love to track the builder's processing latency.

codesome: I tried to move this metric inside builder but it is getting convoluted. Logically does not make sense to pass the topic, partition, to the builder just for the metric. I think we should keep the metric here for now.

pkg/blockbuilder/tsdb.go
```diff
@@ -304,6 +306,7 @@ func (b *tsdbBuilder) compactAndUpload(ctx context.Context, blockUploaderForUser
 		for _, b := range db.db.Blocks() {
 			blockNames = append(blockNames, b.Meta().ULID.String())
 		}
+		numBlocks += len(blockNames)
```
narqo, line 309: Did you mean to use this `numBlocks` somewhere?

narqo (approved):
Works for me, although I think there is some left-over that needs clean-up. It won't require another look, though 🔥

PR #8657, opened 2024-07-08
Refactoring introduce partitionOffsetClient

pracucci:
#### What this PR does

I'm working to move the fetch of the last produced offset from ingesters to query-frontend. In this PR I'm proposing a preliminary refactoring which moves the logic used to fetch partition offsets to a new component called `partitionOffsetClient`.

After this refactoring we have:
- `partitionOffsetClient`: responsible to fetch partition offsets issuing Kafka requests. This is **not** a `services.Service`.
- `partitionOffsetReader`: responsible to periodically fetch partition last produced offset and expose functions to conveniently get the offset from the next request that will be issued. This **is** a `services.Service`.

Why this refactoring? Because in a follow up PR I will introduce a component similar to `partitionOffsetReader` but fetching the last produced offset for **all** partitions and not just 1 specifically (like `partitionOffsetReader` does). We'll need both: the ingester will keep fetching the last produced offset only for their own partition, and query-frontend will fetch the last produced offset for all partitions.

The logic shouldn't have changed in this refactoring PR (if it does, then it's a bug).

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/partition_offset_client.go
```diff
@@ -0,0 +1,194 @@
...
+
+		if returnErr != nil {
+			failuresTotal.Inc()
+		}
+	}()
+
+	return p.fetchPartitionOffset(ctx, partitionID, kafkaOffsetStart)
+}
+
+func (p *partitionOffsetClient) fetchPartitionOffset(ctx context.Context, partitionID int32, position int64) (int64, error) {
```
narqo, line 143: Could you add a comment on why didn't [`kadm.Client.ListStartOffsets`][1] and [`kadm.Client.ListEndOffsets`][2] work here, and we had to construct the low-level messages manually? We use the variant of those helper functions in the `Reader`, so maybe there is a room to simplify it here to.

[1]: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListStartOffsets
[2]: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListEndOffsets

pracucci: Yes I will. The reason is very simple: [kadm.Client.ListStartOffsets](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListStartOffsets) and [kadm.Client.ListEndOffsets](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListEndOffsets) fetch offsets for all partitions, not a single one. It's just more expensive to fetch for all partitions, but ingesters only need the offset of their own one.

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -0,0 +1,194 @@
...
+			Help: "Total number of failed requests to get the partition start offset.",
+		}, []string{"partition"}),
+		partitionStartOffsetLatency: promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+			Name:                            "cortex_ingest_storage_reader_partition_start_offset_request_duration_seconds",
+			Help:                            "The duration of requests to fetch the start offset of a given partition.",
+			NativeHistogramBucketFactor:     1.1,
+			NativeHistogramMaxBucketNumber:  100,
+			NativeHistogramMinResetDuration: 1 * time.Hour,
+			Buckets:                         prometheus.DefBuckets,
+		}, []string{"partition"}),
```
dimitarvdimitrov, line 76: here we'd get one histogram per query-frontend per partition. Previously it would be only one per partition. Do we need this granularity? I'm not sure why we added it in the first place

the same applies to the rest of the metrics. Can we just remove the `partition` label?

dimitarvdimitrov (approved):
easy, i was afraid of the 600 LOC diff, but it was all the same. I'm not sure about the cardinality of those metrics, but otherwise the refactor makes sense

pkg/storage/ingest/partition_offset_client.go
pracucci: >here we'd get one histogram per query-frontend per partition

No, we don't. The query-frontends will request "all partitions" and my plan is to to use a special label value for that (`all`).

dimitarvdimitrov: ok, sounds good

pkg/storage/ingest/partition_offset_client.go
pracucci: Will comment on it in the next PR.

PR #8577, opened 2024-07-02
MQE: Add support for rate on native histograms

jhesketh:
#### What this PR does

Add support for `rate` function on native histograms.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8660, opened 2024-07-09
Mimir query engine: annotations

charleskorn:
#### What this PR does

This PR adds support for annotations (aka warnings) to the Mimir query engine.

It adds support for a single annotation that Prometheus' engine emits: `sum()` with both float and histogram samples at the same timestep for the same output series

(This PR previously also included adding an annotation for `rate()` over a suspected non-counter, but that will be added in a separate PR.)

The latency impact of these changes is low (<1% for our benchmarks), and the impact on peak memory utilisation is within the margin of error for our benchmarks.

This PR also adds support for comparing warnings when comparing results in query-tee. 

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/benchmarks/comparison_test.go
```diff
@@ -189,12 +189,9 @@ func TestBenchmarkSetup(t *testing.T) {
...
 // Why do we do this rather than require.Equal(t, expected, actual)?
 // It's possible that floating point values are slightly different due to imprecision, but require.Equal doesn't allow us to set an allowable difference.
-func requireEqualResults(t testing.TB, expected, actual *promql.Result) {
+func requireEqualResults(t testing.TB, expr string, expected, actual *promql.Result) {
 	require.Equal(t, expected.Err, actual.Err)
-
-	// Ignore warnings until they're supported by the streaming engine.
-	// require.Equal(t, expected.Warnings, actual.Warnings)
-
+	require.ElementsMatch(t, expected.Warnings.AsStrings(expr, 0), actual.Warnings.AsStrings(expr, 0))
```
jhesketh, line 194: Why not check the position of the warnings as well? (ditto in some of the other tests)

jhesketh (commented):
Just curious if we need to handle info annotations, or if only warnings come through the query engine?

pkg/streamingpromql/engine_test.go
```diff
@@ -1216,3 +1216,127 @@ func (t *timeoutTestingQueryTracker) Insert(ctx context.Context, _ string) (int,
...
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			store := promqltest.LoadedStorage(t, "load 1m\n"+strings.TrimSpace(testCase.data))
+			t.Cleanup(func() { _ = store.Close() })
+
+			for engineName, engine := range engines {
+				t.Run(engineName, func(t *testing.T) {
+					for queryType, generator := range queryTypes {
+						t.Run(queryType, func(t *testing.T) {
```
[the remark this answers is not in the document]
charleskorn: It's a good idea - I'll have a think and maybe try to pull this out as a separate PR.

pkg/streamingpromql/operators/function_over_instant_vector.go
```diff
@@ -24,10 +26,34 @@ type FunctionOverInstantVector struct {
...
 }
 
 var _ types.InstantVectorOperator = &FunctionOverInstantVector{}
 
+func NewFunctionOverInstantVector(
+	inner types.InstantVectorOperator,
+	pool *pooling.LimitingPool,
+	metadataFunc functions.SeriesMetadataFunction,
+	seriesDataFunc functions.InstantVectorFunction,
+	expressionPosition posrange.PositionRange,
```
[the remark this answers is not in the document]
charleskorn: >Fair enough though if you want to omit it until we implement the dependent functions.

Yeah, this is what I'm thinking.

The other idea I had was to pass something (the `Query`? something with just the pool and annotations?) to each operator, so we don't have to pass all this stuff around. My only hesitation with this was I worry about passing the `Query` around because it'd be tempting to access things like the start and end timestamps from there, but that will cause issues for subqueries. Maybe I'm being too paranoid though 🙂

charleskorn:
> Just curious if we need to handle info annotations, or if only warnings come through the query engine?

Both are handled as annotations in the same way, but both are emitted together in one (confusingly named) `Warnings` slice on the result. 

The only difference between warnings and info annotations are the message: warnings start with `PromQL warning` and info annotations start with `PromQL info`. See [here](https://github.com/grafana/mimir/blob/a1aff84406ce0ad3f611b38736a1cee77cebb544/pkg/streamingpromql/engine_test.go#L1240) for an example of a warning, and [here](https://github.com/grafana/mimir/blob/a1aff84406ce0ad3f611b38736a1cee77cebb544/pkg/streamingpromql/engine_test.go#L1261) for an example of an info annotation.

pkg/streamingpromql/operators/function_over_instant_vector.go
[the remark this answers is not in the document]
jhesketh: I think it's fine passing each thing for now :+1:

jhesketh (approved):
lgtm, but if possible would love to move this behind #8577

charleskorn:
I've merged in the latest changes from `main` which includes handling info and warning annotations separately (https://github.com/prometheus/prometheus/pull/14327).

jhesketh (approved):
lgtm

charleskorn:
Putting this back to a draft - this needs to be reworked slightly for the annotation emitted by `rate()` when native histograms are involved

charleskorn:
This is ready to go again.

jhesketh (commented):
Looks fine, but I'm not a huge fan of the `haveFloats`. I'm not sure that'll make sense when we make the functions generic in #8747.

If possible, can we resolve (as in come to agreement on) the open discussions on there and then decide which one we merge first and which one we re-work?

pkg/streamingpromql/engine_test.go
```diff
@@ -1215,3 +1215,88 @@ func (t *timeoutTestingQueryTracker) Insert(ctx context.Context, _ string) (int,
...
+		metric{type="float", series="2"} 1+1x3
+		metric{type="histogram", series="1"} {{schema:0 sum:0 count:0}}+{{schema:0 sum:5 count:4 buckets:[1 2 1]}}x3
+		metric{type="histogram", series="2"} {{schema:0 sum:1 count:1 buckets:[1]}}+{{schema:0 sum:5 count:4 buckets:[1 2 1]}}x3
+	`
+
+	testCases := map[string]struct {
+		data                       string
+		expr                       string
+		expectedWarningAnnotations []string
+		expectedInfoAnnotations    []string
```
jhesketh, line 1235: No test cases for Info annotations?

Also a test case for multiple annotations would also be good to ensure we are appending correctly.

pkg/streamingpromql/operators/aggregation.go
```diff
@@ -330,10 +346,19 @@ func (g *group) reconcileAndCountFloatPoints() int {
 		for idx, present := range g.floatPresent {
 			if present {
 				if g.histogramSums[idx] != nil {
-					// If a mix of histogram samples and float samples, the corresponding vector element is removed from the output vector entirely.
+					// If a mix of histogram samples and float samples, the corresponding vector element is removed from the output vector entirely
+					// and a warning is emitted.
```
jhesketh, line 350: (nit)
```suggestion
					// and a warning annotation is emitted.
```

pkg/streamingpromql/query.go
```diff
@@ -320,6 +330,8 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
 		return &promql.Result{Err: compat.NewNotSupportedError(fmt.Sprintf("unsupported result type %s", parser.DocumentedType(q.statement.Expr.Type())))}
 	}
 
+	q.result.Warnings = *q.annotations
```
jhesketh, line 333: I thought Info was added upstream. Why not add support for it now?

charleskorn: tl;dr: this is all we need to do here to cover both info and warning annotations

The API here is a bit confusing, but we're stuck with what Prometheus has done: `promql.Result` has a single field called `Warnings` of type `annotations.Annotations`, which contains both warning and info annotations.

When the result is serialized as JSON, that single field becomes two separate `warnings` and `infos` fields.

jhesketh: Ah, okay, thanks!

pkg/streamingpromql/engine_test.go
charleskorn: >No test cases for Info annotations?

Not yet - we're not yet emitting any info annotations. That will come in a separate PR (eg. the various annotations for `rate`).

>Also a test case for multiple annotations would also be good to ensure we are appending correctly.

Because we only support a single kind of annotation, and that annotation always has the same message, there's no way to construct a test case for this yet.

For example, the expression `sum by (series) (metric) + sum by (series) (metric)` only produces a single annotation even though both `sum`s would emit a warning about a mix of floats and histograms. This is because the deduplication that `annotations.Annotations` does does not consider the position attached to the annotation, and the message is the same in both cases. (Both Prometheus' engine and Mimir's engine are consistent in this regard.)

pkg/streamingpromql/operators/aggregation.go
charleskorn: Resolved in 60cfe30

jhesketh (approved):

PR #8666, opened 2024-07-09
query-frontend: fix case where error responses are double encoded

56quarters:
#### What this PR does

As part of #8595, the HTTP status code is now set correctly for errors. However, this introduced cases where the logic in PrometheusCodec wrapped a JSON APIError response in a duplicate APIError response. This fixes the logic to avoid wrapping when not required: only when there is no content type set, such as with gRPC errors.

Related #8595
Related #8139

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
LGTM

PR #8673, opened 2024-07-10
Fix chunks merging determinism when there are 2 samples with same timestamp but different value

pracucci:
#### What this PR does

We found that with out of order samples enabled a tenant is able to ingest duplicate in-order and out-of-order timeseries data. When queried, data returned form the in-order and out-of-order blocks on the ingester is returned and merged in a non-deterministic way and can appear as if the results are changing.

In this PR I'm proposing to change `batchStream.merge()` to make it more deterministic.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/querier/batch/stream.go
```diff
@@ -174,7 +212,22 @@ func (bs *batchStream) merge(batch *chunk.Batch, size int) {
...
+				// if bs.hPool is not nil, we put there the discarded histogram.Histogram object from batch, so it can be reused.
+				if lt == chunkenc.ValHistogram && bs.hPool != nil {
+					_, h := bs.curr().AtHistogram()
+					bs.hPool.Put((*histogram.Histogram)(h))
+				}
+				// if bs.fhPool is not nil, we put there the discarded histogram.FloatHistogram object from batch, so it can be reused.
+				if lt == chunkenc.ValFloatHistogram && bs.fhPool != nil {
+					_, fh := bs.curr().AtFloatHistogram()
+					bs.fhPool.Put((*histogram.FloatHistogram)(fh))
+				}
```
pracucci, line 227: @krajorama Do you know why this wasn't done before? Is it unsafe doing it?

pkg/querier/batch/stream.go
```diff
@@ -159,11 +159,49 @@ func (bs *batchStream) merge(batch *chunk.Batch, size int) {
...
+			case lt == rt:
+				if lt == chunkenc.ValFloat {
+					_, lValue := bs.curr().At()
+					_, rValue := batch.At()
+					pickLeftSide = lValue > rValue
+				} else if lt == chunkenc.ValHistogram {
+					_, lValue := bs.curr().AtHistogram()
+					_, rValue := batch.AtHistogram()
+
+					pickLeftSide = (*histogram.Histogram)(lValue).Sum > (*histogram.Histogram)(rValue).Sum
```
krajorama, line 178: Please .Count instead of Sum.  Observations can be negative, so Sum is not monotonic. However if Count is higher, you could argue that the histogram has more observations ~ more valueable.

krajorama, line 178: Also I'm wondering how strict we'd want to be. In theory you could have a situation where Count is the same, but the buckets are different, so you end up flapping anyway. Meaning that you'd have to compare every field to be 100% deterministic.

pkg/querier/batch/stream.go
krajorama: Not sure what you mean, we never had a situation before where we picked the right side and the other side wasn't a float which is _not_ pooled.

pkg/querier/batch/stream.go
krajorama: Comparing every field would introduce massive overhead here I feel.

pkg/querier/batch/stream.go
```diff
@@ -159,11 +159,49 @@ func (bs *batchStream) merge(batch *chunk.Batch, size int) {
...
-			if (rt == chunkenc.ValHistogram || rt == chunkenc.ValFloatHistogram) && lt == chunkenc.ValFloat {
-				// Prefer histograms than floats. Take left side if both have histograms.
-				populate(batch, rt)
-			} else {
+			// The two samples have the same timestamp. They may be the same exact sample or different one
+			// (e.g. we ingested an in-order sample and then OOO sample with same timestamp but different value).
+			// We want query results to be deterministic, so we select one of the two side deterministically.
+			var pickLeftSide bool
+
+			switch {
```
dimitarvdimitrov, line 167: i'd break all of your changes into a function. It's an edge case and I think it's distracting from the main body of the function


something like 

```go
func resolveTimestampConflict(lt, rt chunkenc.ValueType, leftBatch, rightBatch *chunk.Batch) (pickedVT chunkenc.ValueType, pickedB, discardedB *chunk.Batch)
```

pkg/querier/batch/stream.go
dimitarvdimitrov: my 2c is that we shouldn't be 100% deterministic. The issue we saw is probably caused by a relabelling rule which drops the `pod` label on series. In reality, having the same count and sum is very unlikely in those scenarios.

I suspect the client can also send duplicate data, but a misconfigured Prometheus is way more common than a custom remote-write client with a bug.

pkg/querier/batch/stream.go
krajorama: It's not an edge case, it's the most common case as usually two ingesters/store-gateways will have the same data.

dimitarvdimitrov: ah good point. Still think a function would make it more obvious

krajorama:
The issue comes up when implementing native histograms counter reset hints. See https://github.com/prometheus/prometheus/issues/15346 . In that case we also need to compare values of samples at the same timestamp to know if the counter reset became invalid. It can be costly, though we don't have numbers on this - the next priority for native histograms will be to work on read performance testing and optimization. If it is a problem, the mitigation - I think - is to make sure chunks are aligned from sources like store gateway/ingester, because we can and do compare whole chunks more easily and efficiently.

dimitarvdimitrov:
@pracucci can i take over this PR?

pracucci:
> @pracucci can i take over this PR?

Yes definitely @dimitarvdimitrov . It's not on my priority list.

pkg/querier/batch/merge_test.go
dimitarvdimitrov, line 1: I made some changes in this test. The test was added in https://github.com/grafana/mimir/pull/9909 by @fionaliao 

From what I could tell "overlapping chunks" doesn't contain any logic, the test case is there just so we can track which way the decision goes. Today on every one of these points we end up comparing the same histogram. At the same time whether we choose the left or the right determines where we consider there's a reset. This sounds sus, but if we're ok with the approach, then I also think that the changes I made should also be ok.

For the rest of the test cases, I think the new changes make sense.

@fionaliao when you have a minute, can you take a look and confirm the changes in this file? 🙏

dimitarvdimitrov:
this is now ready for review

pkg/querier/batch/stream_test.go
```diff
@@ -180,3 +181,358 @@ func requireBatchEqual(t *testing.T, b, o chunk.Batch) {
...
+		require.Equal(t, 1, s.len())
+		require.Equal(t, 2, s.batches[0].Length)
+
+		// Merge empty batch with non-empty stream
+		s.merge(&emptyBatch, chunk.BatchSize, 0)
+		require.Equal(t, 1, s.len())
+		require.Equal(t, 2, s.batches[0].Length)
+	})
+
+	t.Run("single sample batches with conflicts", func(t *testing.T) {
```
pracucci: [nit] This looks to be the same test case as `same timestamp different float values` in `TestBatchStream_MergeWithTimestampConflicts`. I would remove this.

pkg/querier/batch/stream_test.go
```diff
@@ -180,3 +181,358 @@ func requireBatchEqual(t *testing.T, b, o chunk.Batch) {
...
+	for i := 0; i < size; i++ {
+		batch.Timestamps[i] = from + int64(i)
+		h := test.GenerateTestFloatHistogram(int(from) + i)
+		batch.PointerValues[i] = unsafe.Pointer(h)
+	}
+	batch.Length = size
+	return batch
+}
+
+func TestBatchStream_EdgeCases(t *testing.T) {
```
pracucci: [nit] Have you considered merging this test with `TestBatchStream_MergeWithTimestampConflicts`? Looks the same type of tests.

pkg/querier/batch/stream_test.go
```diff
@@ -180,3 +181,358 @@ func requireBatchEqual(t *testing.T, b, o chunk.Batch) {
...
+		batch2 := mkGenericFloatBatch(100, 1)
+		batch2.Values[0] = 10.0
+		s.merge(&batch2, chunk.BatchSize, 0)
+
+		require.Equal(t, 1, s.len())
+		require.Equal(t, 1, s.batches[0].Length)
+		require.Equal(t, 10.0, s.batches[0].Values[0]) // Higher value should win
+	})
+
+	t.Run("extreme float values", func(t *testing.T) {
```
pracucci: [nit] This test is about `decideTimestampConflict()` and not `batchStream.merge()`. I would move this test into a separate function.

pracucci (commented):
LGTM

pkg/querier/batch/stream_test.go
dimitarvdimitrov: i integrated this into TestDecideTimestampConflict_DeterministicBehavior

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [BUGFIX] Query-frontend: Allow limit parameter to be 0 in label queries to explicitly request unlimited results. #12054
 * [BUGFIX] Distributor: Fix a possible panic in the OTLP push path while handling a gRPC status error. #12072
 * [BUGFIX] Tracing: Skip tracing configuration when no tracing environment variables were provided. #12074
+* [BUGFIX] Querier: Samples with the same timestamp are now merged deterministically. Previously this might lead to flapping query results when an out-of-order sample has been ingested which conflicts with a previously ingested in-order sample's value. #8673
```
```suggestion
* [BUGFIX] Querier: Samples with the same timestamp are merged deterministically. Previously, this could lead to flapping query results when an out-of-order sample is ingested that conflicts with a previously ingested in-order sample's value. #8673
```

tacole02 (approved):

pkg/querier/batch/merge_test.go
```diff
@@ -180,9 +180,9 @@ func TestMergeHistogramCheckHints(t *testing.T) {
 						}
 					}(),
 					expectedSamples: []checkHintTestSample{
-						{t: 0, v: 0, hint: histogram.UnknownCounterReset}, // 1 sample from c1
-						{t: 1, v: 8, hint: histogram.UnknownCounterReset}, // 1 sample from c2
-						{t: 2, v: 9, hint: histogram.NotCounterReset},     // 2 samples from c2, previous iterator was also c2, so keep the NCR hint
+						{t: 0, v: 7, hint: histogram.UnknownCounterReset}, // 2 samples from c2 (higher count wins: 75 > 12)
```
fionaliao:
```suggestion
						{t: 0, v: 7, hint: histogram.UnknownCounterReset}, // 2 samples from c2 (higher count wins: 7 > 2)
```

pkg/querier/batch/merge_test.go
```diff
@@ -224,8 +224,8 @@ func TestMergeHistogramCheckHints(t *testing.T) {
 						// batch, and the c1 sample at ts 7 and 8 is added to the batch stream before this c2 sample, so
```
fionaliao:
```suggestion
						// batch, and the c1 sample at ts 8 is added to the batch stream before this c2 sample, so
```

pkg/querier/batch/merge_test.go
```diff
@@ -224,8 +224,8 @@ func TestMergeHistogramCheckHints(t *testing.T) {
...
 						// the prevIteratorID is set to c1 rather than c2 when we append this sample.
 						{t: 6, v: 9, hint: histogram.UnknownCounterReset},
-						{t: 7, v: 4, hint: histogram.UnknownCounterReset},  // c1
-						{t: 8, v: 5, hint: histogram.NotCounterReset},      // c1
+						{t: 7, v: 10, hint: histogram.NotCounterReset},     // c1 (higher count wins)
```
fionaliao:
```suggestion
						{t: 7, v: 10, hint: histogram.NotCounterReset},     // c2 (higher count wins)
```

pkg/querier/batch/merge_test.go
fionaliao: > From what I could tell "overlapping chunks" doesn't contain any logic, the test case is there just so we can track which way the decision goes. 

Yes exactly, it's just sort of documentation (and to make anyone who dares to change the merging code's life harder 😉) 

> Today on every one of these points we end up comparing the same histogram. At the same time whether we choose the left or the right determines where we consider there's a reset. This sounds sus, but if we're ok with the approach, then I also think that the changes I made should also be ok.

The updates look fine to me, though which part is sus? It's not really the left or right that makes a difference to the reset hint, it's the underlying chunk that the sample comes from that does. We can only know for sure there's no counter reset if consecutive samples are from the same chunk, otherwise we have to mark it as unknown counter reset.

fionaliao: Oh, I think I see your point - if it's the same sample value, we could just retain the not counter reset hint rather than setting the hint to unknown even if we change chunks. That would require doing an equality check on the histograms rather than just overwriting the histogram pointer (and reset header) though.

dimitarvdimitrov: yess, you got it. How likely is it that we get a histogram for the exact same timestamp in the query engine and that the sum & count are the same? We'd still replace them the same way we replace them today, we'll just have a `for` loop over the buckets.

dimitarvdimitrov: i checked out the profiles for this. I looked at all queriers in all grafana cells over the last 3 minutes (more than that and `go tool pprof` started refusing to cooperate). 

`batchStream.merge` is 0.84% of all querier CPU. In there very little is spent actually resolving conflicts between samples. Most of the power goes into draining the samples from one of the iterators

```
github.com/grafana/mimir/pkg/querier/batch.(*batchStream).merge
/__w/backend-enterprise/backend-enterprise/vendor/github.com/grafana/mimir/pkg/querier/batch/stream.go

  Total:     213.57s    637.19s (flat, cum)  0.84%
    186       12.54s     12.61s           	for lt, rt := bs.hasNext(), batch.HasNext(); lt != chunkenc.ValNone && rt != chunkenc.ValNone; lt, rt = bs.hasNext(), batch.HasNext() { 
    187        3.27s      3.28s           		t1, t2 := bs.curr().AtTime(), batch.AtTime() 
    188        290ms      290ms           		if t1 < t2 { 
    189        530ms      2.32s           			populate(bs.curr(), lt, -1) 
    190        740ms      750ms           			bs.next() 
    191        300ms      300ms           		} else if t1 > t2 { 
    192        320ms      640ms           			populate(batch, rt, iteratorID) 
    193         30ms       30ms           			batch.Next() 
    194            .          .           		} else { 
    195        520ms      520ms           			if (rt == chunkenc.ValHistogram || rt == chunkenc.ValFloatHistogram) && lt == chunkenc.ValFloat { 
    196            .          .           				// Prefer histograms than floats. Take left side if both have histograms. 
    197            .          .           				populate(batch, rt, iteratorID) 
    198            .          .           			} else { 
    199        480ms      4.83s           				populate(bs.curr(), lt, -1) 
    200            .          .           				// if bs.hPool is not nil, we put there the discarded histogram.Histogram object from batch, so it can be reused. 
    201            .          .           				if rt == chunkenc.ValHistogram && bs.hPool != nil { 
    202            .          .           					_, h := batch.AtHistogram() 
    203            .       20ms           					bs.hPool.Put((*histogram.Histogram)(h)) 
    204            .          .           				} 
    205            .          .           				// if bs.fhPool is not nil, we put there the discarded histogram.FloatHistogram object from batch, so it can be reused. 
    206        250ms      250ms           				if rt == chunkenc.ValFloatHistogram && bs.fhPool != nil { 
    207            .          .           					_, fh := batch.AtFloatHistogram() 
    208            .          .           					bs.fhPool.Put((*histogram.FloatHistogram)(fh)) 
    209            .          .           				} 
    210            .          .           			} 
    211        3.11s      3.20s           			bs.next() 
    212        950ms      950ms           			batch.Next() 
    213            .          .           		} 
    214            .          .           	} 
    215            .          .            
    216        3.57s      3.60s           	for t := bs.hasNext(); t != chunkenc.ValNone; t = bs.hasNext() { 
    217         50ms      470ms           		populate(bs.curr(), t, -1) 
    218        340ms      340ms           		bs.next() 
    219            .          .           	} 
    220            .          .            
    221       64.95s     65.35s           	for t := batch.HasNext(); t != chunkenc.ValNone; t = batch.HasNext() { 
    222       24.37s    292.16s           		populate(batch, t, iteratorID) 
    223       20.91s     20.99s           		batch.Next() 
    224            .          .           	} 
    225            .          .            
    226            .          .           	// The Index is the place at which new sample 
    227            .          .           	// has to be appended, hence it tells the length. 
    228        7.51s      7.58s           	b.Length = b.Index 
    229            .          .            
    230            .          .           	// Store the last iterator id. 
    231        8.18s      8.21s           	bs.prevIteratorID = prevIteratorID 
    232            .          .            
    233       24.20s    172.17s           	bs.batches = append(origBatches, bs.batchesBuf[:resultLen]...) 
    234       35.06s     35.21s           	bs.reset() 
    235        1.10s      1.12s           } 
```

Just read it as a normal pprof file. i renamed it to trick github into accepting it as an attachment
[querier.pprof.zip](https://github.com/user-attachments/files/21257763/querier.pprof.zip)

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * [BUGFIX] Distributor: Check `max_inflight_push_requests_bytes` before decompressing incoming requests. #11967
 * [BUGFIX] Query-frontend: Allow limit parameter to be 0 in label queries to explicitly request unlimited results. #12054
 * [BUGFIX] Distributor: Fix a possible panic in the OTLP push path while handling a gRPC status error. #12072
+* [BUGFIX] Tracing: Skip tracing configuration when no tracing environment variables were provided. #12074
```
```suggestion
```

PR #8674, opened 2024-07-10
blockbuilder: fix logging and metrics

narqo:
This one is a follow-up to #8652

The changes here are the following:

- fix the KV pairs in the "done consuming partition" logs
- make sure we update the partition lag metric, when there is no lag
- rename internal variables to be more consistent

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -319,6 +319,8 @@ func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			continue
 		}
 
+		b.metrics.consumerLag.WithLabelValues(lag.Topic, fmt.Sprintf("%d", lag.Partition)).Set(float64(lag.Lag))
+
 		if lag.Lag <= 0 {
 			if err := lag.Err; err != nil {
 				level.Error(b.logger).Log("msg", "failed to get partition lag", "err", err, "part", part)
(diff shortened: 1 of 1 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -464,19 +462,25 @@ func (b *BlockBuilder) consumePartition(
 			return
 		}
 		b.metrics.processPartitionDuration.WithLabelValues(fmt.Sprintf("%d", part)).Observe(dur.Seconds())
-		level.Info(b.logger).Log("msg", "done consuming partition", "part", part, "dur", dur, "start_lag", startingLag,
-			"cycle_end", cycleEnd, "last_block_end", "num_blocks", numBlocks, time.UnixMilli(lastBlockMax), "curr_block_end", time.UnixMilli(retPl.LastBlockEnd),
```
narqo, line 468: Note: here we messed up the KV pairs in the log line a bit.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -528,28 +532,27 @@ func (b *BlockBuilder) consumePartition(
...
 				// this is the commit point.
 				if commitRec == nil {
-					// When the commit record does not change from the previous commit, we get nil here.
 					commitRec = &kgo.Record{
 						Timestamp:   time.UnixMilli(pl.CommitRecTs),
 						Topic:       lastCommit.Topic,
 						Partition:   lastCommit.Partition,
 						LeaderEpoch: lastCommit.LeaderEpoch,
-						Offset:      lastCommit.At - 1,
+						Offset:      lastCommit.At - 1, // TODO(v): lastCommit.At can we zero if there wasn't any commit yet
```
narqo, line 555: On this one

> `lastCommit.At` can we zero if there wasn't any commit yet

I'm not sure if there are other cases, where this can happen. We will need to investigate this separately.

codesome, line 555:
```suggestion
						Offset:      lastCommit.At - 1, // TODO(v): lastCommit.At can be zero if there wasn't any commit yet
```

codesome (approved):

PR #8676, opened 2024-07-11
Add a lookback config when block builder finds no kafka commit

codesome:
Temporary (or not) fix for the OOMs before we add an option to start consuming at the end (non-trivial because partitions can get rebalanced anytime). We might still have use of this config even when we have an option to start consuming at the end.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -372,15 +372,21 @@ type partitionInfo struct {
 }
 
 func (b *BlockBuilder) nextConsumeCycle(ctx context.Context, pl partitionInfo, cycleEnd time.Time) error {
-	// TODO(codesome): when we deploy it first, we will not have any kafka commit and we will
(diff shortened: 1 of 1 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -372,15 +372,21 @@ type partitionInfo struct {
 }
 
 func (b *BlockBuilder) nextConsumeCycle(ctx context.Context, pl partitionInfo, cycleEnd time.Time) error {
-	// TODO(codesome): when we deploy it first, we will not have any kafka commit and we will
-	// be lagging. Add an option to block builder to start consuming from the end for the first rollout.
-	// 		TODO Optionally also explore a way to consume in parts without the presence of commit
-	//      so that we don't go into a crash loop (because of low resources) in worst case.
```
codesome, line 378: Looks like we saw it coming before

narqo (approved):
I think it's worth trying to get some learnings. We can refactor it later 🔥

PR #8679, opened 2024-07-11
Tracing bridge: force compile error when Otel interface changes

bboreham:
#### What this PR does

Quoting from https://pkg.go.dev/go.opentelemetry.io/otel/trace#hdr-API_Implementations:
> requires them to choose a default behavior for unimplemented interface methods. There are three behavior choices they can make:
> * Compilation failure
> * Panic
> * Default to another implementation

Previous implementation made the third choice, silently ignoing any unimplemented methods.
Seems to me that a compile error is better, as it will let Mimir adapt where required.

(I noticed this while debugging the bridge, as the noop implementation was highly visible. It didn't cause any specific error or misbehaviour)

#### Checklist

- NA Tests updated.
- NA Documentation added.


narqo (approved):
🔥

PR #8680, opened 2024-07-11
Generalize passing options to queries via HTTP headers

krajorama:
#### What this PR does

Handle the read consistency header and engine force fallback header the same. In preparation to add a new header to be able to debug certain queries.

Related to #8599 

#### Checklist

- [x] Tests updated.
- N/A Documentation added.
- N/A `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.


pkg/querier/api/header_options.go
```diff
@@ -0,0 +1,166 @@
...
+
+			for header, validate := range cfg.Headers {
+				value := r.Header.Get(header)
+				if value == "" {
+					continue
+				}
+
+				if err := validate(value); err != nil {
+					// Send a Prometheus API-style JSON error response.
+					w.Header().Set("Content-Type", "application/json")
```
56quarters, line 46: We have to have at least 3 different versions of functions that already write errors as JSON like this.

56quarters (commented):
LGTM overall, just a few non-blocking comments.

pkg/querier/api/consistency.go
```diff
@@ -26,81 +22,21 @@ const (
...
-		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if c := r.Header.Get(ReadConsistencyHeader); IsValidReadConsistency(c) {
-				r = r.WithContext(ContextWithReadConsistency(r.Context(), c))
-			}
-			next.ServeHTTP(w, r)
-		})
-	})
-}
-
-const consistencyLevelGrpcMdKey = "__consistency_level__"
```
dimitarvdimitrov, line 63: if we change the metadata key, then old/new ingester/querier pairs won't be able to pass the consistency header correctly because the MD key is different.

Can you serialize each header into its own metadata key? Maybe change the implementation of `HeaderOptionsConfig` to something like this?


```go
type HeaderOptionsConfig struct {
	Headers []HeaderOptionConfig
}

type HeaderOptionConfig struct {
  HTTPHeaderName string
  GRPCMdKey      string
  Validate       func(string) error
}
```

dimitarvdimitrov (changes requested):

pkg/streamingpromql/compat/fallback_header.go
```diff
@@ -4,45 +4,25 @@ package compat
...
-
-				return
-			}
-
-			r = r.WithContext(withForceFallbackEnabled(r.Context()))
-		}
-
-		handler.ServeHTTP(w, r)
-	})
+func ForceFallbackHeaderNameValidator(v string) error {
```
dimitarvdimitrov, line 14: nit: 

```suggestion
func ValidateForceFallbackHeaderName(v string) error {
```

to match `ValidateReadConsistency`

pkg/querier/api/consistency_test.go
```diff
@@ -1,72 +0,0 @@
...
-			for n := 0; n < b.N; n++ {
-				_, _ = ReadConsistencyServerUnaryInterceptor(ctx, nil, nil, func(context.Context, any) (any, error) {
-					return nil, nil
-				})
-			}
-		})
-	}
-}
-
-func BenchmarkReadConsistencyServerStreamInterceptor(b *testing.B) {
```
dimitarvdimitrov, line 45: is it difficult to get these benchmarks into the new implementation? I guess we can write them again if we need them, but from afar they look like they can be copied without much overhead

dimitarvdimitrov (commented):
and a couple more questions/suggestions

pkg/streamingpromql/compat/fallback_header_test.go
```diff
@@ -43,7 +45,11 @@ func TestEngineFallbackInjector(t *testing.T) {
 
 	for name, testCase := range testCases {
 		t.Run(name, func(t *testing.T) {
-			injector := EngineFallbackInjector{}
+			injector := api.HeaderOptionsMiddleware(api.HeaderOptionsConfig{
+				Headers: map[string]func(string) error{
+					ForceFallbackHeaderName: ForceFallbackHeaderNameValidator,
+				},
+			})
```
charleskorn, line 52: Could we use or somehow share this config here and the config that's configured in `DefaultHeaderOptionsMiddlewareConfig`? Then we don't have to worry about one getting out of sync with the other

pkg/querier/api/header_options.go
```diff
@@ -0,0 +1,166 @@
...
+func HeaderOptionsFromContext(ctx context.Context) (HeaderOptionsContext, bool) {
+	hoc, ok := ctx.Value(headerOptionsContextKey).(HeaderOptionsContext)
+	return hoc, ok
+}
+
+func ContextWithHeaderOptions(ctx context.Context, hoc HeaderOptionsContext) context.Context {
+	prev, ok := HeaderOptionsFromContext(ctx)
+	if ok {
+		// Merge/overwrite with existing headers.
+		for k, v := range hoc.headers {
```
charleskorn, line 95: Could you please add a test for this merging behaviour?

pkg/querier/api/header_options.go
krajorama: The one in pkg/api/error [gives me something](https://github.com/grafana/mimir/blob/90bf98d91df3a9404fe0eca37d5e8c2bec0dcb1f/pkg/api/error/error.go#L85) I don't know how to use in the handler. Which one could I reuse?

krajorama:
Closing for now as it turned out to be too big a change for limited gain.

PR #8681, opened 2024-07-11
tests: work round assert.Equals not working on Labels any more

bboreham:
#### What this PR does

Under `-tags dedupelabels` there are hidden fields which don't compare equal.

~In some cases we can call `labels.Equal`, and in others~ can use `cmp.Equal` via a helper function in Prometheus.

#### Checklist

- [x] Tests updated.
- NA Documentation added.

### DIFF
--- a/pkg/distributor/distributor_test.go
+++ b/pkg/distributor/distributor_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/relabel"
 	"github.com/prometheus/prometheus/scrape"
+	promtestutil "github.com/prometheus/prometheus/util/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/twmb/franz-go/pkg/kfake"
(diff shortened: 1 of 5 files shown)


pkg/distributor/distributor_test.go
```diff
@@ -1291,7 +1292,7 @@ func TestDistributor_Push_LabelRemoval(t *testing.T) {
 			timeseries := ingesters[i].series()
 			assert.Equal(t, 1, len(timeseries))
 			for _, v := range timeseries {
-				assert.Equal(t, tc.expectedSeries, mimirpb.FromLabelAdaptersToLabels(v.Labels))
+				assert.True(t, labels.Equal(tc.expectedSeries, mimirpb.FromLabelAdaptersToLabels(v.Labels)))
```
colega, line 1295: This will be a hell to debug, especially given we're in a loop. `Expected: false, got: true` isn't really meaningful.

Can we create a `test.AssertEqualLabels(t *testing.T, a, b labels.Labels)` in `util/test` that would print a meaningful message if they aren't equal? Similar to what `promtestutil.RequireEqual` prints.

bboreham: sure, I can just do them all as `promtestutil.RequireEqual`

colega (approved):

pkg/distributor/distributor_test.go
colega: Thank you.

PR #8684, opened 2024-07-11
blockbuilder: add new option that sets the offset to start consuming from

narqo:
This is ~a potentially simpler version~ a different solution to same issue from #8676, which rely on [kgo.ConsumeResetOffset](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#ConsumeResetOffset) option. 

cc @codesome 

**TODO** This is still a draft, 

- [ ] I'm yet to try adding a unit test to this one
- [ ] we must figure out how to deal with lag, which is currently calculated from the partitions start, regardless


narqo:
Closing in favour of #8989

PR #6608, opened 2023-11-09
Distributor: Includes the Retry-After header in recoverable error responses

ying-jeanne:
#### What this PR does

This PR addresses [GitHub issue #5233](https://github.com/grafana/mimir/issues/5233) to implement the specified PIR. It introduces the final algorithm to set the retry-after header based on retry-attempts and retry.base [base*2^(retryAttempt-1), base*2^retryAttempt]. The algorithm underwent thorough load testing in a dev environment, involving 1000 agents and network policy rules that disabled requests to Mimir for 5 minutes before reenabling.

**Testing Approach:**

The load test encompassed several algorithms, each evaluated under simulated conditions, as the strategies documented in commit bb0f1662f945bfa05013e5ae2589886a759674c1.

**Decision Making:**

The ultimate algorithm selection is informed by test results, and also ensuring alignment with the existing usage in dskit's [backoff.go](https://github.com/grafana/dskit/blob/e59db7bb4403816f610c1451c104cee8eec5c3d1/backoff/backoff.go#L90).

**Test Result:**
Some load tests result, test with 1000 agents v0.37.3, pile up metrics data during 5 minutes and enabled network.

with retry-after header disabled. The request received by gateway
Screenshot 2023-11-09 at 22 41 24
with retry-after header enabled. The request received by gateway
Screenshot 2023-11-09 at 23 34 32
The observation indicates a significantly reduced peak, and the request per second (r/s) remains relatively stable thereafter.
#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated
- [x] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #8694, opened 2024-07-12
distributor: promote Retry-After header to stable, simplify config

dimitarvdimitrov:
#### What this PR does

Preparing for the 2.14 release. We've been running this in production since it was introduced and there haven't been any problems. The results have been positive in reducing the thundering herd retry behaviour.

Before promoting to stable we also wanted to simplify the configuration. After the discussion in this PR we stopped at replacing `-distributor.retry-after-header.max-backoff-exponent` and `-distributor.retry-after-header.base-seconds` with `.min-backoff` and `.max-backoff` for easier configuration.

Now the Retry-After value is calculated as `min_backoff * 2^(retry_attempt - 1)` with a random jitter of 50% in either direction and capped to `max_backoff`

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.


pracucci (approved):

ying-jeanne (approved):

tacole02 (approved):
Looks good 😄

pstibrany:
I was hoping we would fix the options for this feature before promoting it to stable.

pstibrany:
> I was hoping we would fix the options for this feature before promoting it to stable.

This was pointed out by all three reviewers in the original PR: https://github.com/grafana/mimir/pull/6608#discussion_r1392042223

dimitarvdimitrov:
> I was hoping we would fix the options for this feature before promoting it to stable.

Now is probably a very good time to do this, thanks for taking a look @pstibrany. 

### Simplifying options

From a look now it looks like the preference at the time was to have a simple duration option which caps the maximum duration. Something like `max_retry_backoff: 90s`. Here is one example usage that Peter [wrote up](https://github.com/grafana/mimir/pull/6608#discussion_r1389665631)

> * user sets max to 160s
> * we compute max attempts to 6, and use that value internally
> * it means that after 5 retries, interval that we use for generating next retry-after will be: [3_2^5 ... 3_2^6], ie. [96, 192].
> * when we compute value and it's higher than 160s, we set it to 160s
> [...]


This still makes sense to me. Is there anything that's changed? Tagging @pstibrany @ying-jeanne and @jhalterman since you folks were involved in this back in November.

### Setting upper bound without reducing randomness

From the same comment above:

> [...] What could go wrong is if user sets value like 200s. Then we would compute max attempts to be 7, and our last interval would be [3_2^6, 3_2^7], which is [192, 384], but since we limit at 200s, it would be [192s, 200s]. That wouldn't be so great, but it's solveable.


What if we extend the interval downwards? For example:

* user sets `max_retry_backoff: 200s`
* we compute max attempts to 6, and use that value internally; **6 attempts is the max **
* after 5 retries, interval that we use for generating next would be [3_2^5 ... 3_2^6], ie. [96, 192]. The size of this interval is 96. We use this interval.
* after 6 retries, interval that we use for generating next would be [3_2^6 ... 3_2^7], ie. [193, 384]. But `max_retry_backoff` is larger than the lower bound, so we create a new interval. The upper bound is `max_retry_backoff` and the lower bound is `max_retry_backoff - size(previous_interval)`. This gives the interval [200-96, 200] == [104, 200]

Thoughts?

pstibrany:
> What if we extend the interval downwards? For example:
> 
> * user sets `max_retry_backoff: 200s`
> * we compute max attempts to 6, and use that value internally; **6 attempts is the max **
> * after 5 retries, interval that we use for generating next would be [3_2^5 ... 3_2^6], ie. [96, 192]. The size of this interval is 96. We use this interval.
> * after 6 retries, interval that we use for generating next would be [3_2^6 ... 3_2^7], ie. [193, 384]. But `max_retry_backoff` is larger than the lower bound, so we create a new interval. The upper bound is `max_retry_backoff` and the lower bound is `max_retry_backoff - size(previous_interval)`. This gives the interval [200-96, 200] == [104, 200]
> 
> Thoughts?

Sounds reasonable to me. In the end, this doesn't need to be rocket science, we're generating random number from an interval in the end. As long as it's increasing between retries and doesn't go over some max, we're fine.

ying-jeanne:
Hi, thanks for the comment. I think it's challenging to quantify the effect of retry. I prefer the current approach because it ensures better randomness(smoother request rate) regardless of the user settings. With our current setup, the minimum base is at least 2, and the exponent is at least 1, so the minimum retry intervals would always be [2, 4], [4, 8], and so on. higher max component, means smoother request rate.

However, if we allow users to set arbitrary random numbers, the retry behaviour could become unpredictable. For example, if a user sets the max retry time to 33, the last retry window could be [32, 33]. This variability could result in all requests reaching the maximum retry 32|33 when the network is down for an extended period. In fact, I believe setting it to 32 in this case can smooth the retry rate more effectively. Therefore, a higher max retry time doesn't necessarily improve system behaviour compare to a lower max retry, which  doesn't seem intuitive. 

Unless there is evidence suggesting that the current setting is inadequate, I would prefer to keep it as is.

dimitarvdimitrov:
> the retry behaviour could become unpredictable. For example, if a user sets the max retry time to 33, the last retry window could be [32, 33]. This variability could result in all requests reaching the maximum retry 32|33 when the network is down for an extended period. In fact, I believe setting it to 32 in this case can smooth the request rate more effectively. Therefore, a higher max retry time doesn't necessarily improve system behaviour compare to a lower max retry, which doesn't seem intuitive.

Did you see the [example above](https://github.com/grafana/mimir/pull/8694#issuecomment-2242798601) with `max_retry_backoff: 200`? if the max backoff is 33, then the interval will be [17, 33] instead of [32, 33] (here I'm assuming base 1)

ying-jeanne:
> Setting upper bound without reducing randomness
Hi @dimitarvdimitrov , thanks for pointing that out. I understand your solution now, and it is actually nice and make sense. However, I believe the current calculation method is sufficient and straightforward, as it directly uses [base * 2^(retry-1), base * 2^retry].

Explaining the new solution clearly would require a deeper understanding and additional effort. Unless there is a strong indication that the current approach is not working, I would prefer to stick with it for now. 

That said, if the consensus is that the new approach provides significant benefits, I am open to discussing it further and making adjustments as needed.

pstibrany:
> That said, if the consensus is that the new approach provides significant benefits, I am open to discussing it further and making adjustments as needed.

The consensus in the original PR was that configuration is unnecessarily complicated and we want to simplify that eventually. There was an expectation for a follow-up PR to fix that, but that did not happen.

dimitarvdimitrov:
copying a convo with @ying-jeanne from slack:

@ying-jeanne:

> During the design, I think the useful part of base is when we increase the base we can get lower peak when outage just happened because the retry have a larger interval range from the beginning.  and it is actually the min.
>
> if we really want to change something, change the max exponent backoff, I won’t touch the base, I think it is better to keep control of the minimum retry. The thing is that I don’t think it is easier to understand the logic of round up intervals than the exponential calculation. It is easier to set and not understand, but as soon as you need to understand it, It is more efforts than a math equation.
>
> **so the setting would be base + max retry absolute value**


What's still unclear to me is under what circumstances would an operator want to change the base? Why doesn't the default of 3 work well enough in all cases?

ying-jeanne:
Hi, I believe there’s a balance to be struck between making configurations easy and ensuring their logics could be easily understood. I might be biased because I prefer reading mathematical equations over long text descriptions. If we need to remove something, I support Dimitar's proposal on Setting upper bound without reducing randomness. Since the base is the minimum retry interval, I would keep it. This default shouldn't seem insufficient as long as users aren’t encountering problems and stick with the default settings. Also answered in private what if user can't chose the base:

It ultimately depends on how much headroom users have on their systems. If we don’t allow them to adjust the base, the initial retry wave could have a high ratio. The higher we set the base, the lower the first wave of retries will be. This  can be helpful when want to lower the first retry wave.

dimitarvdimitrov:
I'm ok to take a middle-ground by having the configuration parameters be base + max delay. @pstibrany what do you think?

_in the interest of reducing back-and-forth, I'll continue with details about the two parameters_

The current base setting is called `-distributor.retry-after-header.base-seconds`, but it's a factor, not seconds. WDYT about renaming it to `-distributor.retry-after-header.base-factor`?

I propose to name the second parameter `-distributor.retry-after-header.max-backoff-delay`, which takes duration strings (`30s`, `1h`)

jhalterman:
The concern about randomness can be resolved by separating the jitter setting from the delay settings, ex:
- min-backoff: 1s
- max-backoff: 30s
- jitter-factor: .1

Without jitter, this would result in retry-after delays of 1,2,4,8,16,30. With jitter the delays would be [.9,1.1],[1.8,2.2],[3.6,4.4],[7.2,8.8],[14.4,17.6],[27,30]. This is simple to compute for mimir, and simple for users to think about.

ying-jeanne:
Hi, I like your ideas, I think it can all meet requirements, the thing is current approach works and tested, I think we shouldn't do major change to it, it doesn't worth our efforts. In order to find a middle ground, based on current algorithm, the best we can do is to make it less flexible, keep enabled + max_retry (so algorithm won't change, it is classic exponential backoff, make base a fix value), or enabled + max_delay using "Setting upper bound without reducing randomness" mentioned here https://github.com/grafana/mimir/pull/8694#issuecomment-2242798601.

dimitarvdimitrov:
This is easier to understand than having to do math with exponents and it still allows to increase the initial backoff.

With the current implementation the jitter factor is 50% - it takes the difference with the last power of two (for example upper_bound=32, lower_bound=16 - delay is [16,32]). In the interest of having fewer config options I think we can just hardcode the jitter as 0.5.

If there are no objections, I'll try to find some time for this this week.

dimitarvdimitrov:
sorry for the delay. I made the changes we talked about. 



> With the current implementation the jitter factor is 50%

What I was missing there was that the current (on main) behaviour uses 50% jitter, but only in the negative direction (i.e. 16 -> [8, 16]). Jonathan's suggested implementation makes it in either direction (i.e. 16 -> [8, 24]). I've implemented with hardcoded jitter factor of 0.5 in either direction. I'm open to revise this.

pkg/distributor/push.go
```diff
@@ -200,22 +201,27 @@ func handler(
...
-	delaySeconds := minSeconds + rand.Int63n(maxSeconds-minSeconds)
+	delaySeconds := float64(int64(1) << (retryAttempt - 1))
+	delaySeconds = min(maxBackoff.Seconds(), delaySeconds)
+	delaySeconds = max(minBackoff.Seconds(), delaySeconds)
+	if jitterAmount := int64(delaySeconds * jitterFactor); jitterAmount > 0 {
+		delaySeconds += float64(rand.Int63n(jitterAmount))
+	}
+	// Jitter might have pushed the delaySeconds over maxBackoff or minBackoff, so we need to clamp it again.
+	delaySeconds = min(maxBackoff.Seconds(), delaySeconds)
+	delaySeconds = max(minBackoff.Seconds(), delaySeconds)
```
jhalterman, line 222: I think this can be simplified a bit, to:

```go
	delaySeconds := math.Pow(minBackoff.Seconds(), float64(retryAttempt))
	if jitterAmount := int64(delaySeconds * jitterFactor); jitterAmount > 0 {
		delaySeconds += float64(rand.Int63n(jitterAmount))
	}
	delaySeconds = min(delaySeconds, maxBackoff.Seconds())
```

dimitarvdimitrov: doesn't this use minBackoff as the base? So if minBackoff: 6s, the backoffs will be 6s, 36s, 216s, 1296s,... (with jitter - [3,9], [18,54], [108,324], [648, 1944]...)

aside: I also realized that the intervals were a bit tight and not really with 50% jitter, it was more like 25% in the negative direction, not 50% in either direction. I fixed that in https://github.com/grafana/mimir/pull/8694/commits/c38f5342541c8c75e026d73dc5fb474e1c1e60aa

jhalterman: Yes, minBackoff would be the base, and should probably be something like 1s.

dimitarvdimitrov: sounds good; I updated the formula and the docs

jhalterman: Cool - with that update I think we can remove a few of the min and max checks, like above?

dimitarvdimitrov: I could remove the first `max(minBackoff, delay)` but I think we need the rest of them. Otherwise, the unit tests fail

jhalterman (approved):

pstibrany (approved):
Thank you!

CHANGELOG.md
```diff
@@ -22,7 +22,8 @@
...
 * [CHANGE] Ingester: circuit breakers do not open in case of per-instance limit errors anymore. Opening can be triggered only in case of push and pull requests exceeding the configured duration. #8854
 * [CHANGE] Query-frontend: Return `413 Request Entity Too Large` if a response shard for an `/active_series` request is too large. #8861
-* [CHANGE] Ingester: increase the default inactivity timeout of active series (`-ingester.active-series-metrics-idle-timeout`) from `10m` to `20m`. #8975
+* [CHANGE] Distributor: Promote replying with `Retry-After` header on retryable errors to stable and set `-distributor.retry-after-header.enabled=true` by default. #8694
+* [CHANGE] Distributor: Replace `-distributor.retry-after-header.max-backoff-exponent` and `-distributor.retry-after-header.base-seconds` with `-distributor.retry-after-header.min-backoff` and `-distributor.retry-after-header.max-backoff` for easier configuration. #8694* [CHANGE] Ingester: increase the default inactivity timeout of active series (`-ingester.active-series-metrics-idle-timeout`) from `10m` to `20m`. #8975
```
dimitarvdimitrov, line 26:
```suggestion
* [CHANGE] Distributor: Replace `-distributor.retry-after-header.max-backoff-exponent` and `-distributor.retry-after-header.base-seconds` with `-distributor.retry-after-header.min-backoff` and `-distributor.retry-after-header.max-backoff` for easier configuration. #8694
* [CHANGE] Ingester: increase the default inactivity timeout of active series (`-ingester.active-series-metrics-idle-timeout`) from `10m` to `20m`. #8975
```

dimitarvdimitrov:
appreciate all the rounds of reviews @pstibrany @jhalterman @ying-jeanne @tacole02 🙏

Issue #7847, opened 2024-04-08
Mismatch between ingester limiter and owned series when new ingesters are PENDING with no tokens


dimitarvdimitrov:
The issue is that when new ingesters come up, they [join the ring as PENDING with no tokens](https://github.com/grafana/dskit/blob/73feada6c0d7d02d419415a63fd5a31ae2a40830/ring/lifecycler.go#L638-L641), and only later generate new tokens and update their entry. There should be no delay between these two states, since the `JoinAfter` for Mimir ingesters is the default of `0`. 

However, if there happens to be a delay between ingester A joining the ring and generating new tokens, a run of the owned series service on ingester B would:
- see a lowered local limit, because there is an additional ingester in the ring
- see no change to the owned token ranges, because there have been no new tokens added to the ring.
This results in ingester B caching a new lowered local series limit, but not reducing the owned series count. Depending on how close the owned series count is to the limit, this can result in dropped samples until ingester A adds new tokens to the ring, and ingester B calculated new token ranges.

I was able to reproduce this locally by adding a new ingester with `JoinAfter` set to `60s`:
```
// Initial run of owned series service
ts=2024-03-29T05:35:10.547664799Z caller=user_tsdb.go:589 level=info component="owned series" msg="owned series: recomputed owned series for user" user=anonymous reason="new user" ownedSeriesBefore=20703 ownedSeriesNew=20703 shardSizeBefore=10 shardSizeNew=10 localLimitBefore=50000 localLimitNew=50000 duration=3.793875ms attempts=1 success=true

// New ingester added with no tokens -- local limit updated, but not owned series count
ts=2024-03-29T05:35:55.543544709Z caller=user_tsdb.go:589 level=info component="owned series" msg="owned series: recomputed owned series for user" user=anonymous reason="local series limit changed" ownedSeriesBefore=22822 ownedSeriesNew=22822 shardSizeBefore=10 shardSizeNew=10 localLimitBefore=50000 localLimitNew=25000 duration=3.914875ms attempts=1 success=true

// New ingester adds tokens -- owned series count updated
ts=2024-03-29T05:36:55.54486925Z caller=user_tsdb.go:589 level=info component="owned series" msg="owned series: recomputed owned series for user" user=anonymous reason="ring changed" ownedSeriesBefore=24829 ownedSeriesNew=12645 shardSizeBefore=10 shardSizeNew=10 localLimitBefore=25000 localLimitNew=25000 duration=6.787083ms attempts=1 success=true
```

[...]

PR #8695, opened 2024-07-12
ingester: promote owned series stable and enable by default

dimitarvdimitrov:
#### What this PR does

Preparing for the 2.14 release. We've been running this in production in GL and have since fixed the rough edges of this feature.

related to https://github.com/grafana/mimir/issues/7847

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * [CHANGE] Store-gateway: enabled `-blocks-storage.bucket-store.max-concurrent-queue-timeout` by default with a timeout of 5 seconds. #8496
 * [CHANGE] Store-gateway: enabled `-blocks-storage.bucket-store.index-header.lazy-loading-concurrency-queue-timeout` by default with a timeout of 5 seconds . #8667
 * [CHANGE] Distributor: Incoming OTLP requests were previously size-limited by using limit from `-distributor.max-recv-msg-size` option. We have added option `-distributor.max-otlp-request-size` for limiting OTLP requests, with default value of 100 MiB. #8574
+* [CHANGE] Ingester: enable owned series to stable and enable by default. Active series are now updated along with owned series. They decrease when series change ownership between ingesters. This helps provide a more accurate total of active series when ingesters are added. #8695
```
gotjosh, line 13:
```suggestion
* [CHANGE] Ingester: Promote `-ingester.track-ingester-owned-series` to stable and enabled by default. Active series are now updated along with owned series. They decrease when series change ownership between ingesters. This helps provide a more accurate total of active series when new ingesters are added. #8695
```

gotjosh (approved):
LGTM

dimitarvdimitrov:
feedback from @pr00se 

> i don't think we can do this until we work around [this limitation](https://github.com/grafana/dskit/blob/main/ring/token_range.go#L72-L73)

https://github.com/grafana/mimir/blob/42ccb78851f02cfe70e5681332fc2486b206a1f3/vendor/github.com/grafana/dskit/ring/token_range.go#L72-L77

dimitarvdimitrov:
from a slack conversation with @pr00se 

> can we enable it by default only when zone awareness is enabled - anything tricky there?


> the only other caveat is the replication factor == number of zones









> i'm not actually sure what the failure mode is if it is misconfigured. i think it might just throw an error in the logs and then not recalculate owned series, so if you're limiting based on them you just have no limits?
> and it should probably have some documentation somewhere too


because of these we agreed to keep the feature as experimental.

PR #8703, opened 2024-07-12
Fast fail write requests in distributor if Kafka client max buffered records is reached

pracucci:
#### What this PR does

franz-go Kafka client `Produce()` call blocks if the max number of buffered records is reached. The number of buffered records is the number of records buffered to be sent + the ones whose produce is on-flight but they haven't been acknowledged yet. If the `Produce()` blocks, `Writer.WriteSync()` will block too and not honor the context cancelation, because `WriteSync()` has been implemented assuming `Produce()` will never block, but that's not true.

In this PR I propose to use `TryProduce()` instead of `Produce()`. The only difference between the two is that `TryProduce()` fails immediately if the max buffered records / bytes limit is reached. Assuming a reasonably high limit, I think it's better to fast fail write requests in the distributor instead of keeping adding to the buffer and waiting for it.

Details:
- The franz-go kafka client defaults are 10K records and unlimited bytes. I think in Mimir case it's easier to reason in terms of bytes, so I've set an hardcoded unlimited records limits and introduce a config option for the max buffered bytes (default to 1GB).
- I've added a couple of metrics to easily track the rate of produce failures (we don't have such metrics exposed by franz-go directly)

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/writer.go
```diff
@@ -185,16 +195,25 @@ func (w *Writer) produceSync(ctx context.Context, client *kgo.Client, records []
 		res       kgo.ProduceResults
 	)
 
+	w.writeRequestsTotal.Add(float64(len(records)))
+
 	for _, record := range records {
 		// We use a new context to avoid that other Produce() may be cancelled when this call's context is
 		// canceled. It's important to note that cancelling the context passed to Produce() doesn't actually
 		// prevent the data to be sent over the wire (because it's never removed from the buffer) but in some
 		// cases may cause all requests to fail with context cancelled.
```
dimitarvdimitrov, line 204:
```suggestion
		// We use a new context to avoid that other TryProduce() may be cancelled when this call's context is
		// canceled. It's important to note that cancelling the context passed to TryProduce() doesn't actually
		// prevent the data to be sent over the wire (because it's never removed from the buffer) but in some
		// cases may cause all requests to fail with context cancelled.
```

dimitarvdimitrov (approved):

PR #8705, opened 2024-07-12
feat(config): multiline strings env variable are flatten during expansion

### LINKED ISSUE #8704: Google Cloud Service Accounts secrets can't be injected using Env variables
#### Describe the bug

The [official mimir documentation](https://grafana.com/docs/mimir/latest/configure/configure-object-storage-backend/#gcs) demonstrates how to connect `mimir` to an external storage based on GCS in Google Cloud. 

image

A comment states this should be done via an Env variable because this value has to secured. 

However, if we provide the content described in the documentation as an env variables, the system is not able to parse the configuration *because* the secret is a multiline JSON file and the [expansion system](https://github.com/grafana/mimir/blob/8571f6cc9876e2322d60ac03673d93f66646bd0b/cmd/mimir/main.go#L313-L329) is not yaml-aware. 

#### To Reproduce

Our system runs in k8s, so I provide `manifests` here:

```yaml
# mimir.yaml
target: compactor,distributor,ingester,querier,query-frontend,store-gateway

common:
  storage:
    backend: ${COMMON_STORAGE_BACKEND}
    gcs:
      bucket_name: ${COMMON_STORAGE_GCS_BUCKET_NAME}
      service_account: >-
        ${COMMON_STORAGE_GCS_SERVICE_ACCOUNT}
```

```yaml
# deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels: {}
  name: mimir
spec:
  replicas: 1
  revisionHistoryLimit: 3
  selector:
    matchLabels: {}
  strategy:
    type: Recreate
  template:
    metadata:
      labels: {}
    spec:
      containers:
        - args:
            - -config.expand-env=true
            - -config.file=/opt/mimir/mimir.yaml
          env:
            - name: COMMON_STORAGE_BACKEND
              value: gcs
            - name: COMMON_STORAGE_GCS_BUCKET_NAME
              value: a-bucket-just-for-this-issue
            - name: COMMON_STORAGE_GCS_SERVICE_ACCOUNT
              valueFrom:
                secretKeyRef:
                  key: key.json
                  name: google-cloud-service-account
          image: grafana/mimir:2.12.0
          imagePullPolicy: IfNotPresent
          livenessProbe:
            httpGet:
              path: /api/v1/status/buildinfo
              port: http-mimir
              scheme: HTTP
          name: mimir
          ports:
            - containerPort: 8080
              name: http-mimir
              protocol: TCP
            - containerPort: 9096
              name: grpc-mimir
              protocol: TCP
          readinessProbe:
            httpGet:
              path: /ready
              port: http-mimir
              scheme: HTTP
          volumeMounts:
            - mountPath: /data
              name: mimir-tmp
            - mountPath: /opt/mimir/mimir.yaml
              name: config
              subPath: mimir.yaml
      volumes:
        - emptyDir: {}
          name: mimir-tmp
        - configMap:
            name: mimir
          name: config
```

```json
{
  "type": "service_account",
  "project_id": "my-project",
  "private_key_id": "1234abc",
  "private_key": "-----BEGIN PRIVATE KEY-----\n\n-----END PRIVATE KEY-----\n",
  "client_email": "test@my-project.iam.gserviceaccount.com",
  "client_id": "5678",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/test%40my-project.iam.gserviceaccount.com"
}
```

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: mimir-configuration-with-gcp-service-account

labels:
  - pairs:
      app.kubernetes.io/part-of: mimir
      app.kubernetes.io/for: configuration-with-gcp-service-account
    includeSelectors: true

resources:
  - deployment.yaml
  - namespace.yaml

configMapGenerator:
  - name: mimir
    files:
      - mimir.yaml

secretGenerator:
  - name: google-cloud-service-account
    files:
      - key.json
```

Zip with all resources: [issue.zip](https://github.com/user-attachments/files/16194948/issue.zip)

After the deployment, we can see this from the mimir output: 

```
error loading config from /opt/gradle/mimir/mimir.yaml: Error parsing config file: yaml: line 3: did not find expected key
```

#### Expected behavior

I expect the env variable to be injected in a way I can leverage the service account for authentication, and I don't produce an invalid `yaml`. 

This can be down by doing an "yaml aware" injection or by stripping new lines from the secret, the implementation doesn't matter as soon as I can achieve what the doc recommends 😇

I've checked the mimir source code and confirmed the `yaml` is not correctly indented after the expansion done by [this function](https://github.com/grafana/mimir/blob/8571f6cc9876e2322d60ac03673d93f66646bd0b/cmd/mimir/main.go#L313-L329)

#### Environment

- Infrastructure: K8s, but should be the same for every conjunction of `mimir` + `expandVars` + multiline env vars
- Deployment tool: `kubectl`

davinkevin:
#### What this PR does

To support GoogleCloud Service account keys injected via env variables, the system has to manage multilines strings. Because YAML parsing is complicated, this PR assumes the multilines env variables can be flatten (removing `\n`) before injection. This is applied ONLY to multilines variables, so it won't break compatibility with already existing expanded variables.

Fixes #8704

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/cmd/mimir/main.go
+++ b/cmd/mimir/main.go
@@ -319,6 +319,11 @@ func expandEnvironmentVariables(config []byte) []byte {
 		if v == "" && len(keyAndDefault) == 2 {
 			v = keyAndDefault[1] // Set value to the default.
 		}
+
+		if strings.Contains(v, "\n") {
+			return strings.Replace(v, "\n", "", -1)
+		}
+
 		return v
 	}))
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 * [BUGFIX] Alertmanager: Fixes a number of bugs in silences which could cause an existing silence to be deleted/expired when updating the silence failed. This could happen when the replacing silence was invalid or exceeded limits. #8525
 * [BUGFIX] Alertmanager: Fix help message for utf-8-strict-mode. #8572
 * [BUGFIX] Query-frontend: Ensure that internal errors result in an HTTP 500 response code instead of 422. #8595 #8666
+* [BUGFIX] Configuration: Multi line envs variables are flatten during injection to be compatible with YAML syntax
 
 ### Mixin
 
--- a/cmd/mimir/main_test.go
+++ b/cmd/mimir/main_test.go
@@ -374,32 +374,84 @@ func TestExpandEnvironmentVariables(t *testing.T) {
 	var tests = []struct {
 		in  string
 		out string
+		env string
 	}{
 		// Environment variables can be specified as ${env} or $env.
-		{"x$y", "xy"},
-		{"x${y}", "xy"},
+		{"x$y", "xy", "y"},
+		{"x${y}", "xy", "y"},
 
 		// Environment variables are case-sensitive. Neither are replaced.
-		{"x$Y", "x"},
-		{"x${Y}", "x"},
+		{"x$Y", "x", "y"},
+		{"x${Y}", "x", "y"},
 
 		// Defaults can only be specified when using braces.
-		{"x${Z:D}", "xD"},
-		{"x${Z:A B C D}", "xA B C D"}, // Spaces are allowed in the default.
-		{"x${Z:}", "x"},
+		{"x${Z:D}", "xD", "y"},
+		{"x${Z:A B C D}", "xA B C D", "y"}, // Spaces are allowed in the default.
+		{"x${Z:}", "x", "y"},
 
 		// Defaults don't work unless braces are used.
-		{"x$y:D", "xy:D"},
+		{"x$y:D", "xy:D", "y"},
+
+		// multiline case are managed, useful for Google Cloud Service Accounts
+		{"x$y", "x{\t\t\t\"foo\": \"bar\"\t\t}", `{
+			"foo": "bar"
+		}`},
 	}
 
 	for _, test := range tests {
 		test := test
 		t.Run(test.in, func(t *testing.T) {
-			_ = os.Setenv("y", "y")
+			_ = os.Setenv("y", test.env)
 			output := expandEnvironmentVariables([]byte(test.in))
 			assert.Equal(t, test.out, string(output), "Input: %s", test.in)
 		})
 	}
+}
+
+func TestWithGoogleCloudServiceAccountEnvVariable(t *testing.T) {
+	var configuration = `
+	common:
+	  storage:
+		gcs:
+		  service_account: >-
+		    ${COMMON_STORAGE_GCS_SERVICE_ACCOUNT}
+	
+	blocks_storage:
+	  storage_prefix: monitoringmetricsv1blocks
+	  tsdb:
+		flush_blocks_on_shutdown: true
+	`
+	var serviceAccountEnvValue = `{
+	  "type": "service_account",
+	  "project_id": "my-project",
+	  "private_key_id": "1234abc",
+	  "private_key": "-----BEGIN PRIVATE KEY-----\n\n-----END PRIVATE KEY-----\n",
+	  "client_email": "test@my-project.iam.gserviceaccount.com",
+	  "client_id": "5678",
+	  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+	  "token_uri": "https://oauth2.googleapis.com/token",
+	  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+	  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/test%40my-project.iam.gserviceaccount.com"
+	}`
+
+	var expectedResult = `
+	common:
+	  storage:
+		gcs:
+		  service_account: >-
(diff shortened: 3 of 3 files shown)


cmd/mimir/main.go
```diff
@@ -319,6 +319,11 @@ func expandEnvironmentVariables(config []byte) []byte {
 		if v == "" && len(keyAndDefault) == 2 {
 			v = keyAndDefault[1] // Set value to the default.
 		}
+
+		if strings.Contains(v, "\n") {
```
56quarters, line 323: nit: Could this just be `return strings.Replace(v, "\n", "", -1)`, without the conditional?

56quarters (approved):
Thanks!

cmd/mimir/main.go
davinkevin: I wanted to do that only to apply this transformation to files with `\n`, so we reduce the risk of regression until a potentially better solution is found. 

Because it's just a nit, I'm opening the discussion, but if you want me to apply this modification, it's not a problem.

colega:
Can this break some scenarios where newlines are expected? Maybe someone was already fixing this by putting a properly indented string into their environment variables.

What if we implement two indenting functions like [indent and nindent in Helm](https://helm.sh/docs/chart_template_guide/function_list/)?

It shouldn't be hard to match with a regexp whether the env var ends with `|\s*n?indent\s+[\d]+`, and in that case build a pipe function that would properly indent the newlines.

davinkevin:
> Can this break some scenarios where newlines are expected? Maybe someone was already fixing this by putting a properly indented string into their environment variables.

The code provided does the modification *ONLY* to files with newlines in it. If nothing is done (like currently on `main`), the configuration file is broken, and `mimir` won't start. The `TLDR;` is this is better than the previous state. 

> What if we implement two indenting functions like [indent and nindent in Helm](https://helm.sh/docs/chart_template_guide/function_list/)?

My solution is a quick fix to support the case the documentation presents. A function like those you mentioned is definitely better because they will be "yaml" aware and include content with the proper indentation. 

> It shouldn't be hard to match with a regexp whether the env var ends with `|\s*n?indent\s+[\d]+`, and in that case build a pipe function that would properly indent the newlines.

I thought about this, but how do we know the number of space characters we need before the value? It depends on the value location in the YAML tree. I'm open to suggestions; I don't know if I would be able to implement it, but I am curious to see it 😇

colega:
> The code provided does the modification ONLY to files with newlines in it. 

I understand that.

> If nothing is done (like currently on main), the configuration file is broken, and mimir won't start.

I disagree. It's only broken if you threw your env var without pre-processing it, but maybe someone already did a workaround for this?

Say I have a key:
```
$ cat key.json
{
    "foo": "bar"
}
```
I can put it into an env var like this:
```
$ export KEY=$(sed -e '1!s/^/  /' key.json)
$ echo $KEY
{
      "foo": "bar"
  }
```
When I substitute that in a YAML, it generates a correct yaml:
```
$ cat key.json
{
    "foo": "bar"
}

$ envsubst < config.yaml |  yq .
key: |
  {
      "foo": "bar"
  }
```
And the contents are my original key:
```
$ envsubst < config.yaml |  yq . -ojson
{
  "key": "{\n    \"foo\": \"bar\"\n}\n"
}
```

> I thought about this, but how do we know the number of space characters we need before the value? It depends on the value location in the YAML tree. I'm open to suggestions; I don't know if I would be able to implement it, but I am curious to see it

Yes, but you know where you put the env var in the yaml file, right? So you know how much spaces it would need _there_.

This is how the entire Helm is built:
https://github.com/grafana/mimir/blob/d0c2a79809bdddef8cd5cfa98a20085a51556fa8/operations/helm/charts/mimir-distributed/templates/mimir-config.yaml#L17-L19

For example, in the test you added, it would have to be:
```yaml
common:
  storage:
    gcs:
      service_account: |
        ${COMMON_STORAGE_GCS_SERVICE_ACCOUNT | nindent 8 }
```

davinkevin:
> I disagree. It's only broken if you threw your env var without pre-processing it, but maybe someone already did a workaround for this?

Yes, but in my use-case, I would like to consume an env variable created by a 3rd party user, and because **Google Cloud** provides those `JSON` indented like this, I expect 100% of our secrets to be like this 🤷 .

> Yes, but you know where you put the env var in the yaml file, right? So you know how much spaces it would need there.

I'm not sure to understand. You mean "as a `mimir` user" or as a "developer" trying to offer something to user? 

My point is from a `mimir` developer perspective. With the [current implementation](https://github.com/grafana/mimir/blob/8571f6cc9876e2322d60ac03673d93f66646bd0b/cmd/mimir/main.go#L313-L329) used to substitute values, we don't know where the substitution happens, so we can't implement a logic to add extra spaces for lines 2 to n of the env variable. It would be possible with a more complicated implementation. 

> For example, in the test you added, it would have to be…

Yes, but it's another level of complexity compared to what I've provided. And this makes it the responsibility of the one in charge of the template… it's ok but could generate more support case to the `mimir` project. 

Whatever the solution is (the PR I did or introduction of sprig templating functions), it's ok or me 👌.

56quarters:
I can't find any places in configuration besides GCS credentials where accept a value that could reasonably have newlines in it so I'm going to merge this as-is since it fixes the issue and is simple.

PR #8706, opened 2024-07-12
Improve TestWriter_WriteSync

pracucci:
#### What this PR does

I was investigating alternatives to https://github.com/grafana/mimir/pull/8703 and so I re-tested again what happens when the context passed to `Produce()` gets canceled. When testing it, I realised that the `should interrupt the WriteSync() on context cancelled but other concurrent requests should not fail` test case doesn't (always) fail if cancel the context passed to `Produce()`, due to a timing issue (it depends which between the 2nd and 3rd request gets buffered first).

This PR fixes it.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
makes sense, LGTM

PR #8708, opened 2024-07-12
Add cortex_ingest_storage_writer_buffered_produce_bytes metric

pracucci:
#### What this PR does

This PR is a follow up of https://github.com/grafana/mimir/pull/8703. In this PR I propose to add the `cortex_ingest_storage_writer_buffered_produce_bytes` tracked as summary at fixed short intervals using the same technique used in the query-scheduler. Why? Because this way we can get a good approximation of the peak buffered produce bytes over 1m period, which is what I want to track (we mostly care about the peak for this metric, which is the p100).

In addition, I've added `cortex_ingest_storage_writer_buffered_produce_bytes_limit`, so we I can easily alert once the buffer gets close to the limit.

In a follow up PR I will also update dashboards and alerts.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):

pstibrany (approved):
lgtm. I think we can miss p100, if we are sampling every 250ms, but I'm not sure if that's a problem.

pracucci:
> lgtm. I think we can miss p100, if we are sampling every 250ms, but I'm not sure if that's a problem.

p100 is just the 100th percentile among values samples every 250ms. The 250ms interval is the granularity we get.

PR #8711, opened 2024-07-13
Concurrent processing of partitions in block builder

codesome:
This PR adds support to consume partitions concurrently in a single block builder.

The idea is that we resume record fetching concurrently with max parallelism and then a single record fetcher distributes the records to appropriate channels where it will be consumed in parallel.

I still need to fix a bunch of rough edges, like properly stopping consumption of records without missing/leaking any records when we resume it later, and making sure issue in one partition consumption does not impact/block fetching of records indefinitely.

@narqo this is ready for a peek for general design while I still fix the edge cases.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


codesome:
Not one of our goals right now.

PR #8724, opened 2024-07-15
query-tee: add support for comparing native histograms in results

charleskorn:
#### What this PR does

This PR adds support for comparing native histograms in query results to query-tee.

Previously, any native histograms in query results were ignored.

I've also taken this opportunity to clarify some of the error messages generated when results do not match, and to replace the use of `errors.Wrap` with `fmt.Errorf` in `response_comparator.go`.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/querytee/response_comparator.go
+++ b/tools/querytee/response_comparator.go
@@ -12,6 +12,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/common/model"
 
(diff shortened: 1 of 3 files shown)


tools/querytee/response_comparator.go
```diff
@@ -231,26 +288,64 @@ func compareSamplePair(expected, actual model.SamplePair, opts SampleComparisonO
...
-		return math.Float64bits(f) == math.Float64bits(s)
+		return math.Float64bits(first) == math.Float64bits(second)
+	}
+	if opts.UseRelativeError && second != 0 {
+		return math.Abs(first-second)/math.Abs(second) <= opts.Tolerance
+	}
+	return math.Abs(first-second) <= opts.Tolerance
+}
+
+func compareSampleHistogramPair(expected, actual model.SampleHistogramPair, opts SampleComparisonOptions) error {
```
krajorama, line 312: nit: call inputs first and second as compareSampleValue. Function doesn't care and in line with the other function.

tools/querytee/response_comparator.go
```diff
@@ -231,26 +288,64 @@ func compareSamplePair(expected, actual model.SamplePair, opts SampleComparisonO
...
+	}
+
+	if !slices.EqualFunc(expected.Histogram.Buckets, actual.Histogram.Buckets, compareSampleHistogramBuckets(opts)) {
+		return fmt.Errorf("expected buckets %s for timestamp %v but got %s", expected.Histogram.Buckets, expected.Timestamp, actual.Histogram.Buckets)
+	}
+
+	return nil
+}
+
+func compareSampleHistogramBuckets(opts SampleComparisonOptions) func(expected, actual *model.HistogramBucket) bool {
```
krajorama, line 336: nit: same as above, rename arguments to first, second

krajorama (approved):
Approved with tiny nit, thank you for the extensive test coverage! According to VS code all branches are covered.

tools/querytee/response_comparator.go
charleskorn: Addressed in a21ea79.

tools/querytee/response_comparator.go
charleskorn: This function does care about which input is which, as they're used to produce the error message when comparison fails (eg. "expected timestamp X but got Y").

The names here mirror those of `compareSamplePair`, which uses these names for the same reason.

PR #8728, opened 2024-07-15
Use github.com/okzk/sdnotify fork to unblock upgrades

aknuds1:
#### What this PR does
The latest revision of github.com/okzk/sdnotify fails to build for us, since it depends on CGo. The revision changes a function we don't use, i.e. `Reloading`, but it's all the same a hassle that automatic updating of the dependency fails. I've submitted a [PR](https://github.com/okzk/sdnotify/pull/4) to sdnotify to drop the CGo dependency, in the meantime I propose we use my fork.

Also fixing the changelog entry to refer to the original PR, instead of the corresponding issue.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
 * [ENHANCEMENT] Query-frontend: added `remote_read` to `op` supported label values for the `cortex_query_frontend_queries_total` metric. #8412
 * [ENHANCEMENT] Query-frontend: log the overall length and start, end time offset from current time for remote read requests. The start and end times are calculated as the miminum and maximum times of the individual queries in the remote read request. #8404
 * [ENHANCEMENT] Storage Provider: Added option `-<prefix>.s3.dualstack-enabled` that allows disabling S3 client from resolving AWS S3 endpoint into dual-stack IPv4/IPv6 endpoint. Defaults to true. #8405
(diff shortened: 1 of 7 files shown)


pstibrany:
Can't we use old version until it's fixed in the upstream?

aknuds1:
@pstibrany yes, we can do that, I was more being proactive in case my PR doesn't get merged upstream.

dimitarvdimitrov (approved):
SGTM, thanks for fixing this Arve.

dimitarvdimitrov:
and I didn't actually see the existing comments

> Can't we use old version until it's fixed in the upstream?


happy to go this route too.

aknuds1:
@dimitarvdimitrov do you suggest waiting a while to see if upstream accepts my fix? I'm just worried about being unable to update sdnotify.

pstibrany:
> @dimitarvdimitrov do you suggest waiting a while to see if upstream accepts my fix? I'm just worried about being unable to update sdnotify.

We can always move to a fork when we need to, right? I'd prefer to avoid it, if we can.

dimitarvdimitrov:
yeah maybe it's better if we change fewer things. We have a very limited use of sdnotify, so it should be ok to keep an older version.

aknuds1:
OK, I guess I can close this PR in the meantime.

aknuds1:
Closing until we find the sdnotify fork is needed.

PR #8737, opened 2024-07-16
Fetch records concurrently with benchmarks

codesome:
A quick change to see if we make fetching records and processing them concurrent, do we see any gains. I wrote a quick benchmark, and the numbers vary greatly with different number of records/samples/series. With the benchmark as is, I see ~16% gains.

Although https://github.com/grafana/mimir/pull/8711 is underway, it will be nice to try this out in the meantime.

```
benchmark                    old ns/op      new ns/op      delta
BenchmarkBlockBuilder-10     3157905625     2640359666     -16.39%
```

### DIFF
--- a/pkg/blockbuilder/blockbuilder_test.go
+++ b/pkg/blockbuilder/blockbuilder_test.go
@@ -5,7 +5,6 @@ package blockbuilder
 import (
 	"context"
 	"errors"
-	"fmt"
 	"os"
 	"path"
 	"testing"
@@ -474,7 +473,6 @@ func TestBlockBuilder_StartupWithExistingCommit(t *testing.T) {
 	)
 }
 
-// Testing block builder starting up with an existing kafka commit.
 func BenchmarkBlockBuilder(b *testing.B) {
 	ctx, cancel := context.WithCancelCause(context.Background())
 	b.Cleanup(func() { cancel(errors.New("test done")) })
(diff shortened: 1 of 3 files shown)


pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -473,9 +474,54 @@ func TestBlockBuilder_StartupWithExistingCommit(t *testing.T) {
...
+		testBuilder := newTestTSDBBuilder(cfg, overrides)
+		testBuilder.compactFunc = func() {
+			compactCalled <- struct{}{}
+		}
+		return testBuilder
+	}
+
+	// It was difficult to reliably get the time taken to consume the records and produce a block
+	// by using b.ResetTimer. So for now, printing the required time to analyse.
+	// TODO(codesome): figure out how we can get the required numbers without prints.
```
codesome, line 513: To be noted about the benchmark. The 8% gains come from the printed time below.

narqo: > figure out how we can get the required numbers without prints

You can use `b.ResetTimer` and [`b.ReportMetric`](https://pkg.go.dev/testing#B.ReportMetric) for that:

```go
bb := New()
bb.staring()

b.ResetTimer()

err := bb.NextConsumeCycle(ctx, cycleEnd)
if err != nil {
    b.Fatal(err)
}
<-compactCalled

b.ReportMetric(float64(b.Elapsed().Nanoseconds()), "ns/op")
```

codesome: I tried the ResetTimer but the benchmark seemed to be running many iterations like forever. So I had removed it to make the benchmark timer go long. TIL about ReportMetric! I will try that

codesome:
I somehow managed to push the code without the change and undo the changes locally as well. Bummer! Re-doing it.

codesome:
I just fixed a race and I see more gains. 16%. Updated PR description.

narqo:
_We've discussed that in the DM, but I'm putting it here, just so it didn't slip, and I didn't forget while I'm off._

The idea looks good to me. But, in its current implementation, I'm concerned this re-introduces the issue with Kafka rebalance, that we were addressing in #8591. That is, a consumer that is a part of a group, should not block the group's rebalance for extended period of time. Otherwise, after around one minute of blockage (or so), the broker kicks the consumer out of the group, assuming it is malfunctioning.

codesome:
Updated the PR to call `AllowRebalance` more often. Making this PR work same as before without only fetching and processing made concurrent. I will test it out and see how it goes.

codesome:
I tested this, and it ended up taking a little longer to consume and the memory consumption was much higher (3x). Memory usage might be because of a high channel size, leading to a lot of records buffering up. But the time increase is unexpected.

After rolling back this change in dev, the memory consumption stayed higher with consumption time still staying higher. So I am not sure if this test was conclusive.

Need to test it again in a controlled setting where the load does not change much between tests.

codesome:
Code has changed a lot and not one of our priority at the moment. Hence closing.

PR #8740, opened 2024-07-16
ingest: encapsulate record producction logic into `KafkaProducer` type

ortuman:
#### What this PR does

This PR extracts the logic for producing Kafka records into a public exported type (`ingest.KafkaProducer`), so that it can be re-used in other projects. 

Fixes n/a

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (commented):
From a design perspective, I'm not a big fan of taking client in input to ProduceSync(). A client wrapper would have been nicer. I guess this conflict with kafkaWriterClient, but looks solvable.

Also calling promise for each record looks an antipattern to me, because works differently than franz-go. I think we need to re-think this design.

pracucci:
Moved to draft cause I attempt a slightly different approach in another PR, then we can compare.

PR #8747, opened 2024-07-17
MQE: Refactor range functions to allow functions other than rate

jhesketh:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/functions/common.go
```diff
@@ -5,8 +5,10 @@ package functions
 import (
 	"github.com/grafana/mimir/pkg/streamingpromql/pooling"
 	"github.com/grafana/mimir/pkg/streamingpromql/types"
+	"github.com/prometheus/prometheus/promql"
 )
 
+// SeriesMetadataFunction is a function to operate on the metadata of a single series.
```
charleskorn, line 11: This isn't quite right - `SeriesMetadataFunction` operates over many series, not a single series.

pkg/streamingpromql/functions/common.go
```diff
@@ -17,6 +19,7 @@ func DropSeriesName(seriesMetadata []types.SeriesMetadata, _ *pooling.LimitingPo
 	return seriesMetadata, nil
 }
 
+// InstantVectorFunction is a function to operate on an instant vector series.
```
charleskorn, line 22: This might be a bit clearer:

```suggestion
// InstantVectorFunction is a function that takes in a instant vector and produces an instant vector.
```

pkg/streamingpromql/functions/common.go
```diff
@@ -43,3 +46,7 @@ func FloatTransformationDropHistogramsFunc(transform func(f float64) float64) In
 func Passthrough(seriesData types.InstantVectorSeriesData, _ *pooling.LimitingPool) (types.InstantVectorSeriesData, error) {
 	return seriesData, nil
 }
+
+// RangeVectorStepFunction is a function to operate on a range vector step.
+// floatBuffer and histogramBuffer will contain the values for the step range.
+type RangeVectorStepFunction func(step types.RangeVectorStepData, rangeSeconds float64, floatBuffer *types.FPointRingBuffer, histogramBuffer *types.HPointRingBuffer) (*promql.FPoint, *promql.HPoint, error)
```
charleskorn, line 52: I'd prefer not to return pointers for `FPoint` and `HPoint` here - perhaps we could return `bool`s to indicate their presence instead of checking the points for `nil`?

pkg/streamingpromql/functions/common.go
```diff
@@ -43,3 +46,7 @@ func FloatTransformationDropHistogramsFunc(transform func(f float64) float64) In
 func Passthrough(seriesData types.InstantVectorSeriesData, _ *pooling.LimitingPool) (types.InstantVectorSeriesData, error) {
 	return seriesData, nil
 }
+
+// RangeVectorStepFunction is a function to operate on a range vector step.
+// floatBuffer and histogramBuffer will contain the values for the step range.
```
charleskorn, line 51: This line (`floatBuffer and histogramBuffer ...`) is a bit misleading: it might make it sound like implementations of this method don't need to consider `step.RangeEnd`.

pkg/streamingpromql/functions.go
```diff
@@ -43,17 +43,17 @@ func SingleInputVectorFunctionOperatorFactory(name string, metadataFunc function
 	}
 }
 
-// TransformationFunctionOperatorFactory creates an InstantVectorFunctionOperatorFactory for functions
+// SingleInputVectorTransformationFunctionOperatorFactory creates an InstantVectorFunctionOperatorFactory for functions
```
charleskorn, line 46: This name is quite wordy - how about `InstantVectorTransformationFunctionOperatorFactory`?

Any function that doesn't take in a single argument doesn't qualify as a simple transformation function in my mind, so I think we can drop that part of the name.

pkg/streamingpromql/functions.go
```diff
@@ -43,17 +43,17 @@ func SingleInputVectorFunctionOperatorFactory(name string, metadataFunc function
...
 // Parameters:
 //   - name: The name of the function.
 //   - seriesDataFunc: The function to handle series data
-func TransformationFunctionOperatorFactory(name string, seriesDataFunc functions.InstantVectorFunction) InstantVectorFunctionOperatorFactory {
+func SingleInputVectorTransformationFunctionOperatorFactory(name string, seriesDataFunc functions.InstantVectorFunction) InstantVectorFunctionOperatorFactory {
 	return SingleInputVectorFunctionOperatorFactory(name, functions.DropSeriesName, seriesDataFunc)
 }
 
-// LabelManipulationFunctionOperatorFactory creates an InstantVectorFunctionOperator for functions
+// SingleInputVectorLabelManipulationFunctionOperatorFactory creates an InstantVectorFunctionOperator for functions
```
charleskorn, line 56: Same as above - I think we can drop the `SingleInput` and "just" (😅) call it `InstantVectorLabelManipulationFunctionOperatorFactory`

charleskorn (commented):
I think this will need to be reworked slightly after https://github.com/grafana/mimir/pull/8660 is merged, but otherwise the general design LGTM.

pkg/streamingpromql/functions/common.go
charleskorn, line 52: One more thought: what if we split this in two, one for computing the output over float samples and one for the same over native histograms? Or have a simple case for most functions perhaps?

With the exception of `count_over_time` (and maybe `last_over_time`?), I don't believe there are any functions over range vectors that support both floats and histograms at once for a single output point, so this should simplify the implementation of most of these functions somewhat.

colega: (Also `present_over_time`?)

charleskorn: One more idea: rather than returning a `FPoint` or `HPoint`, what if this returned `(float64, bool)` or `*FloatHistogram`? Then the implementation of the function doesn't need to think about the output sample timestamp.

jhesketh: I think they may need to consider what happens when both a histogram and float exists a single point. For example `rate` will return nil points where both exist. I don't think we should split these up at this point, I think it is more logical for the operating function to consider how to handle when both are present rather than move that logic up into the definition of the function.

jhesketh: What is the opposition to returning a pointer here?

I don't like that we have to have a value for the float64 that we only check if a bool tells us to.

The RangeVectorStepFunction needs the `step` so it already understands the timestamp.

We only create the FPoint or HPoint if it is needed, so returning nil should be minimal overhead.

charleskorn: This is how I'm thinking about it: there three categories of functions over range vectors:

* functions that only work with floats and ignore native histograms, even if they occur in a single step where floats are also present: `min_over_time`, `max_over_time`, `quantile_over_time`, `stddev_over_time`, `stdvar_over_time`, `mad_over_time`, `changes`, `deriv`, `holt_winters`, `idelta`, `irate`, `predict_linear`

* functions that handle a single step seeing both floats and native histograms: `count_over_time`, `resets`, `last_over_time`, `present_over_time`

* functions that will not emit a sample if the step has both floats and native histograms, and otherwise produce a sample matching the input type (eg. float input produces float output and histogram input produces histogram output): `delta`, `increase`, `rate`, `avg_over_time`, `sum_over_time`

I'm wondering if we should have specific cases for each of these three that handle the common functionality shared by each group, so that the function-specific implementations can be very simple.

charleskorn: >What is the opposition to returning a pointer here?

It adds significantly to garbage collection: we'd be creating a new pointer for every single point emitted by a function over a range vector, only to immediately drop it on the floor once we store the values in the `[]FPoint` / `[]HPoint` slice we've already allocated for the `InstantVectorSeriesData`.

>The RangeVectorStepFunction needs the `step` so it already understands the timestamp.

Agree - but if we just return the `float64` or `*FloatHistogram`, then the function-specific implementation can be even simpler, as most of the implementations then become "call `floatBuffer.UnsafePoints(step.RangeEnd)`, compute result over slices, return result".

charleskorn: If you're curious about the impact of the pointers, I'd suggest comparing the benchmarks for `rate()` before and after the changes in this PR - I would expect a significant impact.

jhesketh: > I'm wondering if we should have specific cases for each of these three that handle the common functionality shared by each group, so that the function-specific implementations can be very simple.

I guess my concern here is how do we determine the cases? We'd still need `RangeVectorStepFunc` in `FunctionOverRangeVector` or otherwise have two separate funcs like `RangeVectorFloatStepFunc` and `RangeVectorHistogramStepFunc`, perhaps with a enum for each of the cases you describe.

However we'd likely still want the function to get the actual points as otherwise we're handling too many different states in `FunctionOverRangeVector` (imo).

> * functions that only work with floats and ignore native histograms, even if they occur in a single step where floats are also present: `min_over_time`, `max_over_time`, `quantile_over_time`, `stddev_over_time`, `stdvar_over_time`, `mad_over_time`, `changes`, `deriv`, `holt_winters`, `idelta`, `irate`, `predict_linear`

Wouldn't these just call `floatBuffer.UnsafePoints(step.RangeEnd)` and never bother with the histograms returning nil?

> * functions that handle a single step seeing both floats and native histograms: `count_over_time`, `resets`, `last_over_time`, `present_over_time`

These need both, so any re-arrangements are moot for this case.

> * functions that will not emit a sample if the step has both floats and native histograms, and otherwise produce a sample matching the input type (eg. float input produces float output and histogram input produces histogram output): `delta`, `increase`, `rate`, `avg_over_time`, `sum_over_time`

Unless `FunctionOverRangeVector` checks for the presence of both floats+histograms, the functions would still need both buffers to check themselves. Having FunctionOverRangeVector check doesn't make as much sense to me.

I think either I don't fully understand what you're envisioning here or I disagree with it. Would it be okay if we leave this for now and revisit it in a followup PR if necessary?

jhesketh: Okay, makes sense, thanks.

pkg/streamingpromql/functions.go
```diff
@@ -65,56 +65,78 @@ func TransformationFunctionOperatorFactory(name string, seriesDataFunc functions
...
+//
+// Parameters:
+//   - name: The name of the function.
+//   - rangeStepFunc: The function to handle a range vector step
 
-	return &operators.FunctionOverRangeVector{
-		Inner: inner,
-		Pool:  pool,
-	}, nil
+func SingleRangeVectorTransformationFunctionOperatorFactory(name string, rangeStepFunc functions.RangeVectorStepFunction) InstantVectorFunctionOperatorFactory {
```
charleskorn, line 108: [nit] For consistency with `InstantVectorTransformationFunctionOperatorFactory`, should we call this `RangeVectorTransformationFunctionOperatorFactory`?

pkg/streamingpromql/functions/common.go
```diff
@@ -43,3 +46,21 @@ func FloatTransformationDropHistogramsFunc(transform func(f float64) float64) In
...
+//   - rangeSeconds: the duration of the range in seconds.
+//   - floatBuffer: a ring buffer containing float values for the step range.
+//   - histogramBuffer: a ring buffer containing histogram values for the step range.
+//
+// Returns:
+//   - hasFloat bool: a boolean indicating if a float value is present.
+//   - f float64: the float value.
+//   - h *histogram.FloatHistogram: nil if no histogram is present.
+//   - err error.
+type RangeVectorStepFunction func(step types.RangeVectorStepData, rangeSeconds float64, floatBuffer *types.FPointRingBuffer, histogramBuffer *types.HPointRingBuffer) (hasFloat bool, f float64, h *histogram.FloatHistogram, err error)
```
charleskorn, line 66: [nit] The convention for these OK / present bools is for them to follow the parameter they relate to:
```suggestion
type RangeVectorStepFunction func(step types.RangeVectorStepData, rangeSeconds float64, floatBuffer *types.FPointRingBuffer, histogramBuffer *types.HPointRingBuffer) (f float64, hasFloat bool, h *histogram.FloatHistogram, err error)
```

pkg/streamingpromql/operators/function_over_range_vector.go
```diff
@@ -73,193 +75,35 @@ func (m *FunctionOverRangeVector) NextSeries(ctx context.Context) (types.Instant
 			return types.InstantVectorSeriesData{}, err
 		}
 
-		err = m.computeNextStep(&data, step)
+		hasFloat, f, h, err = m.RangeVectorStepFunc(step, m.rangeSeconds, m.floatBuffer, m.histogramBuffer)
```
charleskorn, line 78: [nit] Couldn't we declare and initialise these here?

```suggestion
		hasFloat, f, h, err := m.RangeVectorStepFunc(step, m.rangeSeconds, m.floatBuffer, m.histogramBuffer)
```

pkg/streamingpromql/functions/common.go
```diff
@@ -43,3 +46,21 @@ func FloatTransformationDropHistogramsFunc(transform func(f float64) float64) In
...
 }
+
+// RangeVectorStepFunction is a function that operates on a range vector step.
+//
+// floatBuffer and histogramBuffer will contain the values for the step range.
+// They may also contain values outside of the step, so step.RangeEnd needs to be considered.
+//
+// Parameters:
+//   - step: the range vector step data to be processed.
+//   - rangeSeconds: the duration of the range in seconds.
```
charleskorn, line 57: [nit] Not something we have to do in this PR, but I wonder if it makes sense to include `rangeSeconds` in `RangeVectorStepData` and drop the parameter from this function (or just expect the few functions that need this to compute it themselves from `RangeVectorStepData.RangeEnd - RangeVectorStepData.RangeStart` perhaps?)

pkg/streamingpromql/functions/rate.go
```diff
@@ -0,0 +1,176 @@
...
+
+import (
+	"github.com/prometheus/prometheus/model/histogram"
+	"github.com/prometheus/prometheus/promql"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func Rate(step types.RangeVectorStepData, rangeSeconds float64, floatBuffer *types.FPointRingBuffer, histogramBuffer *types.HPointRingBuffer) (bool, float64, *histogram.FloatHistogram, error) {
+	var err error
```
charleskorn, line 16: [nit] Does this need to be here? Looks like it could be declared inline on line 66.

pkg/streamingpromql/functions/rate.go
```diff
@@ -0,0 +1,176 @@
...
+	// Floats
+	fHead, fTail := floatBuffer.UnsafePoints(step.RangeEnd)
+	fCount := len(fHead) + len(fTail)
+
+	// Histograms
```
charleskorn, line 21: [nit] The `// Floats` and `// Histograms` comments seem superfluous to me.

pkg/streamingpromql/functions/rate.go
```diff
@@ -0,0 +1,176 @@
...
+	hCount := len(hHead) + len(hTail)
+
+	if fCount > 0 && hCount > 0 {
+		// We need either at least two Histograms and no Floats, or at least two
+		// Floats and no Histograms to calculate a rate. Otherwise, drop this
+		// Vector element.
+		return false, 0, nil, nil
+	}
+
+	if fCount >= 2 {
```
charleskorn, line 32: [nit] This `Rate(...)` function is getting quite long - perhaps the bodies of this `if` and the following `if` could be split out to their own functions?

pkg/streamingpromql/functions/common.go
charleskorn: I think this will be easier to discuss once we implement some more of these functions - let's pause this discussion until we get to that.

charleskorn (approved):
LGTM modulo nits and linting failure

pkg/streamingpromql/operators/function_over_range_vector.go
jhesketh: I was concerned about re-assigning it each loop iteration instead of re-using the existing variables. However, I think they'll only be on the stack not in the heap so it should be okay. I'll change it to your suggestion, but is my assumption here correct?

charleskorn: Yep, you're correct - these should only be on the stack (although the `FloatHistogram` instance that `h` points to will be on the heap, but that's expected and unavoidable). So there's no harm in using `:=` here as far as I know.

PR #8750, opened 2024-07-17
Refactoring: extract ProduceSync() into a new KafkaProducer component

pracucci:
#### What this PR does

This PR is an alternative to #8740. In this PR I propose to create a `kgo.Client` wrapper named `KafkaProducer` which provides:

1. Higher level `ProduceSync()` with support for our "max buffered bytes" limit
2. The following metrics:
   - <prefix>_buffered_produce_bytes
   - <prefix>_buffered_produce_bytes_limit
   - <prefix>_produce_requests_total
   - <prefix>_produce_failures_total

There should be no functional difference in this PR, expect:
- The "max buffered bytes" limit is per Kafka client. This is how it was before #8738. In practice we run with only 1 client so there's no practical difference (the reason why I haven't removed multiple clients support is because I have no idea how to write `TestDistributor_Push_ShouldSupportIngestStorage` otherwise).
- The metrics tracked by KafkaProducer have `client_id` label because we could have multiple KafkaProducer(s) for a single Writer

Closes #8740

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


ortuman (approved):
lgtm, nice.

tacole02 (approved):

PR #8755, opened 2024-07-17
Generalise and make some functions of astmapper public

zenador:
#### What this PR does

Generalises astmapper's shardSummer, and make some functions public to allow reuse outside of query sharding

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

None as these are minor, non-user-facing changes.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/sharding.go
+++ b/pkg/frontend/querymiddleware/astmapper/sharding.go
@@ -19,7 +19,7 @@ import (
 
 // NewSharding creates a new query sharding mapper.
 func NewSharding(ctx context.Context, shards int, logger log.Logger, stats *MapperStats) (ASTMapper, error) {
-	shardSummer, err := newShardSummer(ctx, shards, vectorSquasher, logger, stats, nil)
+	shardSummer, err := newShardSummer(ctx, shards, vectorSquasher, logger, stats)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -18,39 +18,67 @@ import (
...
 	stats        *MapperStats
 
+	shardLabeller ShardLabeller
+
 	canShardAllVectorSelectorsCache map[string]bool
 }
 
-// newShardSummer instantiates an ASTMapper which will fan out sum queries by shard
-func newShardSummer(ctx context.Context, shards int, squasher squasher, logger log.Logger, stats *MapperStats) (ASTMapper, error) {
+func NewShardSummer(ctx context.Context, shards int, squasher Squasher, logger log.Logger, stats *MapperStats, shardLabeller ShardLabeller) (*shardSummer, error) { //nolint:revive // We do not need shardSummer to be exported.
```
charleskorn, line 76: Is this called from anywhere except line 55 above? It seems weird that this method is exported but the type it returns is not. If the only place that calls this method is line 55 above, then can we make this method unexported too?

zenador: In Mimir it's only line 55, but we're exporting it so that GEM can re-use this for the cluster sharding.

charleskorn: Right, makes sense, thanks for explaining.

charleskorn: I think in this case, it makes sense to export `shardSummer` then, no?

charleskorn (approved):

PR #8766, opened 2024-07-18
Use offset instead of timestamp for commit meta

codesome:
The kafka record timestamps are not required to be monotonic. The logic of block builder requires the number being used be monotonic w.r.t. the order of records. The offset of record is monotonic. So use it instead.

cc @narqo

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -891,8 +891,7 @@ func (b *BlockBuilder) consumePartition(
 		LastBlockEnd:   commitBlockEnd,
 	}
 
-	err = commitRecord(ctx, b.logger, client, b.cfg.Kafka.ConsumerGroup, pl.Commit)
-	return pl, err
+	return pl, commitRecord(ctx, b.logger, client, b.cfg.Kafka.ConsumerGroup, pl.Commit)
 }
 
 func (b *BlockBuilder) seekPartition(client *kgo.Client, part int32, rec kgo.EpochOffset) {
--- a/pkg/blockbuilder/blockbuilder_test.go
+++ b/pkg/blockbuilder/blockbuilder_test.go
@@ -797,26 +797,15 @@ func TestBlockBuilderNonMonotonicRecordTimestamps(t *testing.T) {
 
 	bb, err := New(cfg, log.NewNopLogger(), prometheus.NewPedanticRegistry(), overrides)
 	require.NoError(t, err)
-	compactCalled := make(chan struct{}, 2)
-	bb.tsdbBuilder = func() builder {
-		testBuilder := newTestTSDBBuilder(cfg, overrides)
-		testBuilder.compactFunc = func() {
-			compactCalled <- struct{}{}
-		}
-		return testBuilder
-	}
 
 	require.NoError(t, bb.starting(ctx))
 	t.Cleanup(func() {
 		require.NoError(t, bb.stopping(nil))
 	})
 
-	run := func(name string, end time.Time, compactions int, expSamples []mimirpb.Sample) {
+	run := func(name string, end time.Time, expSamples []mimirpb.Sample) {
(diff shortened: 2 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -471,7 +471,7 @@ func (b *BlockBuilder) consumePartition(
 		level.Info(b.logger).Log("msg", "done consuming partition", "part", part, "dur", dur,
 			"start_lag", startingLag, "cycle_end", cycleEnd,
 			"last_block_end", time.UnixMilli(lastBlockEnd), "curr_block_end", time.UnixMilli(retPl.LastBlockEnd),
-			"last_seen_till", time.UnixMilli(seenTillTs), "curr_seen_till", time.UnixMilli(retPl.SeenTillTs),
+			"last_seen_till", seenTillOffset, "curr_seen_till", time.UnixMilli(retPl.SeenTillOffset),
```
narqo, line 474:
```suggestion
			"last_seen_till", seenTillOffset, "curr_seen_till", retPl.SeenTillOffset,
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -481,7 +481,7 @@ func (b *BlockBuilder) consumePartition(
 	level.Info(b.logger).Log(
 		"msg", "consuming partition", "part", part, "lag", lag,
 		"cycle_end", cycleEnd, "last_block_end", time.UnixMilli(lastBlockEnd), "curr_block_end", blockEndAt,
-		"last_seen_till", time.UnixMilli(seenTillTs))
+		"last_seen_till", time.UnixMilli(seenTillOffset))
```
narqo, line 484:
```suggestion
		"last_seen_till", seenTillOffset)
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -612,9 +612,9 @@ func (b *BlockBuilder) consumePartition(
 	// We should take the max of "seen till" timestamp. If the partition was lagging
 	// due to some record not being processed because of a future sample, we might be
 	// coming back to the same consume cycle again.
-	commitSeenTillTs := seenTillTs
-	if commitSeenTillTs < lastRec.Timestamp.UnixMilli() {
-		commitSeenTillTs = lastRec.Timestamp.UnixMilli()
+	commitSeenTillOffset := seenTillOffset
+	if commitSeenTillOffset < lastRec.Offset {
+		commitSeenTillOffset = lastRec.Offset
```
narqo, line 617: Could we also update the comment for this chunk. It refers to the timestamp.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -712,14 +712,14 @@ const (
...
-	return fmt.Sprintf("%d,%d,%d,%d", kafkaCommitMetaV1, commitRecTs, lastRecTs, blockEnd)
+func marshallCommitMeta(commitRecTs, lastRecOffset, blockEnd int64) string {
+	return fmt.Sprintf("%d,%d,%d,%d", kafkaCommitMetaV1, commitRecTs, lastRecOffset, blockEnd)
 }
 
 // commitRecTs: timestamp of the record which was committed (and not the commit time).
 // lastRecTs: timestamp of the last record processed (which will be >= commitRecTs).
 // blockEnd: timestamp of the block end in this cycle.
-func unmarshallCommitMeta(meta string) (commitRecTs, lastRecTs, blockEnd int64, err error) {
+func unmarshallCommitMeta(meta string) (commitRecTs, lastRecOffset, blockEnd int64, err error) {
```
narqo, line 722: Similar to the above, the code comments here are also a bit off now.

narqo (approved):
There are a couple of small typos that we should fix before merging. But overall it's great to see that the change set here isn't that drastic here 👍👍👍

PR #8688, opened 2024-07-11
Add partitionOffsetClient.FetchLastProducedOffsets()

pracucci:
#### What this PR does

In this PR I'm adding `partitionOffsetClient.FetchLastProducedOffsets()`. It's a simple function to fetch the last produce offset of all partitions. It will be used in the next PR.

Note: I'm updating franz-go kadm with my fork, until https://github.com/twmb/franz-go/pull/775 is merged (I think it's non controversial).

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8778, opened 2024-07-20
Add TopicOffsetsReader

pracucci:
#### What this PR does

This PR is a follow up of https://github.com/grafana/mimir/pull/8688. In this PR I'm introducing `TopicOffsetsReader`, which is like `partitionOffsetReader` but reads the last produce offset of all partitions in a topic instead of a single partition. `TopicOffsetsReader` is not used yet, but will be used in the next PR.

Notes to revewiers:
- I introduced `genericOffsetReader` which is a generic-based version of the offset reader, then used by the already existing `partitionOffsetReader` and the new `TopicOffsetsReader`.
- `TopicOffsetsReader`  is exported because it will be used outside of the `pkg/storage/ingest/` package (on the contrary, `partitionOffsetReader` is only used internally)

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
LGTM

PR #8787, opened 2024-07-22
fix(helm): Ensure global volumes are mounted in memcached exporter

bradleypettit:
#### What this PR does

Recently global volume support was added to the helm chart, however it wasn't added to the exporter container in memcached statefulsets. This PR adds the missing reference in the exporter container template ensuring that global volumes are mounted as expected.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/Chart.yaml
+++ b/operations/helm/charts/mimir-distributed/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 5.4.1
+version: 5.4.0
 appVersion: 2.13.0
 description: "Grafana Mimir"
 home: https://grafana.com/docs/helm-charts/mimir-distributed/latest/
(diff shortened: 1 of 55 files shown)


operations/helm/charts/mimir-distributed/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 5.4.0
+version: 5.4.1
```
dimitarvdimitrov, line 2: normally we don't do patch releases for these fixes. There are [weekly versions released](https://artifacthub.io/packages/helm/grafana/mimir-distributed/5.4.0-weekly.298) every Monday (usually on time) which contain the changes on main. Will these weekly releases work for you?

dimitarvdimitrov (commented):
Thanks for the fix! I left a comment about the patch version bump

operations/helm/charts/mimir-distributed/Chart.yaml
bradleypettit: > Normally contributors need not bump the version.

My mistake, totally missed that. Weekly is definitely fine, I'll revert that change shortly.

dimitarvdimitrov (approved):
LGTM, thank you!

PR #8803, opened 2024-07-23
Update values.yaml

felipefrx:
It is advisable to enable persistence in the ingester to avoid losing metrics.


dimitarvdimitrov (approved):
good call, we should probably remove the option to run without persistence because of temporary data gaps that can occur if the ingester runs without PVs

PR #8804, opened 2024-07-23
BlockBuilder: Make the LookbackOnNoCommit actually effective

codesome:
Even with LookbackOnNoCommit the first cycle would still consume records from far in the past. With this change, we will now discard records that are before LookbackOnNoCommit, hence make this config actually do what it was meant to do.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -240,6 +240,7 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 	for {
 		select {
 		case cycleEnd := <-time.After(waitTime):
+			level.Info(b.logger).Log("msg", "triggering next consume from running", "cycle_end", cycleEnd, "cycle_time", nextCycleTime)
 			err := b.NextConsumeCycle(ctx, cycleEnd.Add(-time.Second))
 			if err != nil {
 				b.metrics.consumeCycleFailures.Inc()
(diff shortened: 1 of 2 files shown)


narqo:
So we actually want to discard the records before the `LookbackOnNoCommit`. That makes sense, and is what I wanted to do in #8684. My version bumped into a problem with how the existing code relies on the lag, and I wanted to expolore if we can get rid of lag completely, first.

---

Unrealtaed to above: there seem to be a deadlock in the tests ([CI job](https://github.com/grafana/mimir/actions/runs/10065748579/job/27825575158)):

```
panic: test timed out after 30m0s
running tests:
	TestBlockBuilder (30m0s)
	TestBlockBuilder/starting_fresh_with_existing_data_but_no_kafka_commit (30m0s)
```

codesome:
I thought I had fixed that test in https://github.com/grafana/mimir/pull/8823 since it was happening locally as well, but looks like it is not fixed :/. There were 2 issues in that test - one was when getting the group lag, we only tried once, and hard failed without retry. Second was mismatch in number of compactions expected depending on wall clock, which made the test hand at particular times and not at other times.
The issue is unrelated to this PR, so we can fix it separately (I have some ideas that I will try out).

codesome:
> But meanwhile, I'm fine with moving forward with this version also.

Sounds good. I see this PR as a temporary fix to help us test the correctness and stability sooner. I tried the start at end in https://github.com/grafana/mimir/pull/8823 as well but I faced the same issue with how we use lag.

narqo (approved):
Works for me 🔥

PR #8808, opened 2024-07-24
Fetch last produced offset in the query-frontend

pracucci:
#### What this PR does

In this PR I'm adding the fetching of the last produced offsets to the query-frontend. This is done only when ingest storage is enabled **and** the request has strong read consistency enabled. The offsets and then propagated through an HTTP header and context metadata down to ingesters. When strong read consistency is enabled, the ingesters will wait until the offset specified in the request is consumed; if no offset is specified in the request, they will fallback to the previous behaviour which is waiting until the "last produced offset" (fetched at the time when the request hit the ingester) is consumed.

I've done few manual tests, and offsets look to be propagated correctly.

Out of the scope of this PR (but will be done in subsequent ones):

- Update dashboards / alerts (in particular, alert if offsets are not passed to ingesters)
- Add / improve an integration test
-  Do not fetch last produced offsets if a query doesn't touch ingesters at all

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/frontend/querymiddleware/codec.go
```diff
@@ -673,8 +673,8 @@ func (c prometheusCodec) EncodeLabelsQueryRequest(ctx context.Context, req Label
 		return nil, fmt.Errorf("unknown query result response format '%s'", c.preferredQueryResultResponseFormat)
 	}
 
-	if consistency, ok := api.ReadConsistencyFromContext(ctx); ok {
-		r.Header.Add(api.ReadConsistencyHeader, consistency)
+	if level, ok := api.ReadConsistencyFromContext(ctx); ok {
```
pracucci, line 676: Note to reviewers: we should propagate HTTP headers in `EncodeLabelsQueryRequest()` too, **but** this function is currently dead code (not used) and `LabelsQueryRequest` doesn't have headers yet, so I haven't added support in this PR but it's something I'm tracking as part of this work.

pkg/storage/ingest/reader.go
```diff
@@ -881,9 +863,6 @@ func newReaderMetrics(partitionID int32, reg prometheus.Registerer) readerMetric
 			Buckets:                         prometheus.DefBuckets,
 		}),
 		lastConsumedOffset: lastConsumedOffset,
-		kprom: kprom.NewMetrics("cortex_ingest_storage_reader",
-			kprom.Registerer(prometheus.WrapRegistererWith(prometheus.Labels{"partition": strconv.Itoa(int(partitionID))}, reg)),
```
pracucci, line 885: Note to reviewers: I removed the "partition" label here. I don't think it was worth it.

pkg/storage/ingest/reader_client.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "creating kafka client")
+	}
+
+	return client, nil
+}
+
+func NewKafkaReaderClientMetrics(component string, reg prometheus.Registerer) *kprom.Metrics {
+	return kprom.NewMetrics("cortex_ingest_storage_reader",
+		kprom.Registerer(prometheus.WrapRegistererWith(prometheus.Labels{"component": component}, reg)),
```
pracucci, line 41: Note to reviewers: I've added the "component" label here, in order to not have conflicting metrics in single binary mode (on the other side, we never tested ingest storage in single binary so other stuff may not work at all).

development/mimir-ingest-storage/config/mimir.yaml
```diff
@@ -16,7 +16,6 @@ ingest_storage:
     topic:   mimir-ingest
 
 ingester:
-  return_only_grpc_errors:     true
```
dimitarvdimitrov, line 19: oops, thanks for this

pkg/util/validation/limits.go
```diff
@@ -1151,6 +1153,43 @@ func MaxDurationPerTenant(tenantIDs []string, f func(string) time.Duration) time
 	return result
 }
 
+// PreferredStringPerTenant returns the preferred string value among the actual settings for the
+// given tenants. The order of preference is specified in the input preferences slice. A preference
+// with lower index means higher importance. If no setting matches the any of the specified preferences,
+// then the value from the first tenant is returned.
+func PreferredStringPerTenant(tenantIDs []string, f func(string) string, preferences []string) string {
```
dimitarvdimitrov, line 1160: the godoc doesn't mention the function parameter. Can you include that too?

dimitarvdimitrov: and also - i don't quite get the point of this function. Can't we just say "if any tenant has strong read consistency, then the query uses that, otherwise use weak consistency"?

pkg/mimir/modules.go
```diff
@@ -702,6 +704,26 @@ func (t *Mimir) initQueryFrontendCodec() (services.Service, error) {
 	return nil, nil
 }
 
+// initQueryFrontendTopicOffsetsReader instantiates the topic offsets reader used by the query-frontend
+// when the ingest storage is enabled.
+func (t *Mimir) initQueryFrontendTopicOffsetsReader() (services.Service, error) {
+	if !t.Cfg.IngestStorage.Enabled {
```
dimitarvdimitrov, line 710: this means that we should also change the migration procedure, right? At the very least query-frontends should get ingest storage enabled on them along with queriers. And maybe ideally enabled before queriers, so that we don't fall back to relying on ingesters fetching the offsets

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -253,17 +260,30 @@ func newQueryTripperware(
...
-			cardinality = newCardinalityQueryCacheRoundTripper(c, cacheKeyGenerator, limits, cardinality, log, registerer)
-			labels = newLabelsQueryCacheRoundTripper(c, cacheKeyGenerator, limits, labels, log, registerer)
-		}
-
 		if cfg.ShardActiveSeriesQueries {
 			activeSeries = newShardActiveSeriesMiddleware(activeSeries, cfg.UseActiveSeriesDecoder, limits, log)
 			activeNativeHistogramMetrics = newShardActiveNativeHistogramMetricsMiddleware(activeNativeHistogramMetrics, limits, log)
 		}
 
+		// Enforce read consistency after caching.
```
dimitarvdimitrov, line 268: this made me realize that strong read consistency can still use cached responses, which might not always be expected. Should we skip caching when strong read consistency is requested? This feels outside of the scope of this PR though, so it's mostly a discussion point

dimitarvdimitrov (commented):
some initial comments, I'm about 1/3 of the way

pkg/util/validation/limits.go
pracucci: > Can't we just say "if any tenant has strong read consistency, then the query uses that, otherwise use weak consistency"?

Yes, but that's what the function does. The function is just a generalization.

pracucci: > the godoc doesn't mention the function parameter. Can you include that too?
Done in https://github.com/grafana/mimir/pull/8808/commits/b7eebcda87d416801555f8dabe05ce0c60543b34

pkg/mimir/modules.go
pracucci: > this means that we should also change the migration procedure, right? At the very least query-frontends should get ingest storage enabled on them along with queriers.

Yes, absolutely. Will do in a separate PR.

> And maybe ideally enabled before queriers, so that we don't fall back to relying on ingesters fetching the offsets

This is not required. It doesn't matter the rollout order. If offsets are not propagated by query-frontend to ingesters, it will keep working as of today.

pkg/storage/ingest/reader_client.go
dimitarvdimitrov: we used to have the partition label here, but I don't think it ever made much sense because there's only one partition per consumer. Agree? 

This might break the block builder, but provided it's still not merged maybe we shouldn't worry about it

pkg/frontend/querymiddleware/roundtrip.go
pracucci: In my opinion cache comes first. It's like work in Mimir classic architecture too, and I would just keep parity. If you want to bypass, then you should pass `Cache-Control: no-store`. But I'm open to discuss it.

pkg/util/validation/limits.go
dimitarvdimitrov: it looks overcomplicated. Is there another place which needs this kind of logic? Wouldn't something like this be simpler

```go
level = querierapi.ReadConsistencyEventual
for _, t := range tenantIDs {
  if r.limits.IngestStorageReadConsistency(t) == querierapi.ReadConsistencyStrong {
    consistency = querierapi.ReadConsistencyStrong
    break
  }
}
```

pkg/frontend/querymiddleware/read_consistency_test.go
```diff
@@ -0,0 +1,136 @@
...
+			var downstreamReq *http.Request
+			downstream := RoundTripFunc(func(req *http.Request) (*http.Response, error) {
+				downstreamReq = req
+				return nil, nil
+			})
+
+			ctx := context.Background()
+			logger := log.NewNopLogger()
+
+			_, clusterAddr := testkafka.CreateCluster(t, numPartitions, topic)
```
dimitarvdimitrov, line 69: a fully-fledged kafka looks like an overkill for this test. `TopicOffsetsReader` is already extensively tested in the `ingest` package. WDYT about mocking the `NewTopicOffsetsReader`?

dimitarvdimitrov (commented):
a few more comments. I've gone through the whole PR now. Biggest one is probably about propagating partial offsets to each ingester instead of the whole encoded header

pkg/frontend/querymiddleware/roundtrip.go
dimitarvdimitrov: I guess it comes down to what we promise with strong read consistency. If we want to give strong read consistency on all samples, then we should disable cache for strong read consistency queries.

Otherwise, if we keep caching, we'd potentially not guarantee strong read consistency on out of order samples. It's hard to explain that was we promise is "the latest data except if the query was ran before and there's older data points which arrived in the last 10 minutes".

We don't want to promote this feature, but I still feel like a less surprising behaviour is better. Even if that means worse query performance (which also happens to create a disincentive to use strong consistency when you don't need it).

pkg/querier/api/DESIGN.md
```diff
@@ -0,0 +1,81 @@
...
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Varint + base64                    | 23399 ns/op | 48193 B/op  | 13 allocs/op |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Varint + snappy + base64           | 18063 ns/op | 38720 B/op  | 12 allocs/op |
+| Binary-encoded list of <offset id> only, encoded with: Varint + snappy + base64                           | 13120 ns/op | 24384 B/op  | 10 allocs/op |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Int32 / int64 + base64             | 18974 ns/op | 48192 B/op  | 13 allocs/op |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Int32 / int64 + snappy + base64    | 17451 ns/op | 48193 B/op  | 13 allocs/op |
+| Binary-encoded list of <offset id> only, encoded with: Int + snappy + base64                              | 9615 ns/op  | 22336 B/op  | 10 allocs/op |
+
+### Considerations
+
+- The Mimir use case is encode once (in the query-frontend) and decode / lookup many (in each ingester)
```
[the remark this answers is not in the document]
pracucci: I see it as an optimization not a design thing. It's something we can consider for the future, but I wouldn't block on this.

Morever, I want to stress that this PR is just the first step. I don't want to wait even 1ns in the ingesters as final goal. My idea is that medium terms these offsets won't be propagated at all, and all the waiting will happen in the query-frontend. However, since we haven't designed the 2nd step of the work yet, I can't say that it will happen for sure, but that's my goal.

pkg/util/math/digits_test.go
```diff
@@ -0,0 +1,63 @@
...
+	assert.Equal(t, 7, Digits[int64](-123456))
+	assert.Equal(t, len(strconv.Itoa(math.MaxInt)), Digits[int64](math.MaxInt))
+	assert.Equal(t, len(strconv.Itoa(math.MinInt+1)), Digits[int64](math.MinInt+1))
+}
+
+func TestEstimatedDigitsInt32(t *testing.T) {
+	// Small numbers.
+	for i := int32(-10000); i < 10000; i++ {
+		assert.GreaterOrEqual(t, EstimatedDigitsInt32(i), Digits(i))
+		assert.LessOrEqual(t, EstimatedDigitsInt32(i), Digits(i)+1)
```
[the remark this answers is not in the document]
pracucci: No, it's a between `Digits(i)` and `Digits(i) + 1`. I may use a `assert.True(t, ... comparison here ...)` but the advantage of using `GreaterOrEqual()` and `LessOrEqual()` is that we can see the actual value on failure.

pkg/frontend/querymiddleware/read_consistency_test.go
pracucci: I personally prefer testing it more end-to-end rather than mocking `TopicOffsetsReader`. Using `kfake` and the real `TopicOffsetsReader` increases my confidence.

pkg/ingester/ingester_ingest_storage_test.go
```diff
@@ -239,8 +239,35 @@ func TestIngester_QueryStream_IngestStorageReadConsistency(t *testing.T) {
...
+
+			// To keep the test simple, use a trick passing a negative offset so the query will return immediately.
+			// In this test we just want to check that the passed offset is honored.
+			readConsistencyOffsets: map[int32]int64{0: -2},
+			expectedQueriedSeries:  0,
+		},
+	}
+
+	for testName, testData := range tests {
+		t.Run(testName, func(t *testing.T) {
```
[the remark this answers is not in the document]
pracucci: Done in https://github.com/grafana/mimir/pull/8808/commits/e946e498f9e658930c5e757af264a8a9c1b21b20

pkg/querier/api/DESIGN.md
```diff
@@ -0,0 +1,81 @@
...
+
+| Encoding format                                                                                           | Bytes length |
+| --------------------------------------------------------------------------------------------------------- | ------------ |
+| Comma-separated string: "<partition id>:<offset id>"                                                      | 23889        |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Varint + base64                    | 15916        |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Varint + snappy + base64           | 7980         |
+| Binary-encoded list of <offset id> only, encoded with: Varint + snappy + base64                           | 5364         |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Int32 / int64 + base64             | 16000        |
+| Binary-encoded list of <partition id> <offset id> pairs, encoded with: Int32 / int64 + snappy + base64    | 10688        |
+| Binary-encoded list of <offset id> only, encoded with: Int + snappy + base64                              | 5356         |
```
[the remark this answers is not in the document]
pracucci: You're right, I didn't consider it. Fixed in https://github.com/grafana/mimir/pull/8808/commits/e946e498f9e658930c5e757af264a8a9c1b21b20

pkg/querier/api/consistency_test.go
```diff
@@ -70,3 +203,185 @@ type serverStreamMock struct {
...
+		"multiple partitions where partition IDs are in reverse order": {
+			encoded:              "v1=10:100,1:10",
+			expectedPartitions:   map[int32]int64{1: 10, 10: 100},
+			unexpectedPartitions: []int32{0, 100},
+		},
+	}
+
+	for testName, testData := range tests {
+		t.Run(testName, func(t *testing.T) {
+			serialized := EncodedOffsets(testData.encoded)
```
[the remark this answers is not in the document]
pracucci: Done in https://github.com/grafana/mimir/pull/8808/commits/e946e498f9e658930c5e757af264a8a9c1b21b20

pkg/frontend/querymiddleware/roundtrip.go
pracucci: I see your point and I don't disagree. But in the ingest storage the strong read consistency has been introduced to have a way to run queries on par with classic architecture, and that's just what I'm doing here.

pkg/storage/ingest/reader_client.go
pracucci: > we used to have the partition label here, but I don't think it ever made much sense because there's only one partition per consumer. Agree?

Agree. I commented about it here: https://github.com/grafana/mimir/pull/8808/files#r1690885330

pkg/util/validation/limits.go
pracucci: Ok. Simplified as you suggested here: https://github.com/grafana/mimir/pull/8808/commits/bb45335c6bf64289e24e13f5f2ad4344727dcb78

pkg/util/math/digits_test.go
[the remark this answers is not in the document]
dimitarvdimitrov: ok, now I think I get it. So being equal to either `Digits(i)` or `Digits(i) + 1` is ok.

pkg/storage/ingest/reader.go
```diff
@@ -602,50 +602,36 @@ func (r *PartitionReader) WaitReadConsistencyUntilOffset(ctx context.Context, of
...
-		ctx, cancel = context.WithTimeoutCause(ctx, r.kafkaCfg.WaitStrongReadConsistencyTimeout, errWaitStrongReadConsistencyTimeoutExceeded)
-		defer cancel()
-	}
-
-	defer func() {
-		// Do not track failure or latency if the request was canceled (because the tracking would be incorrect).
-		if errors.Is(returnErr, context.Canceled) {
-			return
+func (r *PartitionReader) waitReadConsistency(ctx context.Context, getOffset func(context.Context) (int64, error)) error {
+	_, err := r.metrics.strongConsistencyInstrumentation.Observe(func() (any, error) {
```
dimitarvdimitrov, line 606: nit on the signature: `struct{}` instead of `any` would make it more obvious that the return value is ignored

pkg/frontend/querymiddleware/roundtrip_test.go
```diff
@@ -794,6 +800,167 @@ func TestTripperware_RemoteRead(t *testing.T) {
...
+		testData := testData
+
+		t.Run(testName, func(t *testing.T) {
+			t.Parallel()
+
+			for _, consistencyLevel := range []string{querierapi.ReadConsistencyEventual, querierapi.ReadConsistencyStrong} {
+				consistencyLevel := consistencyLevel
+
+				t.Run(fmt.Sprintf("consistency level: %s", consistencyLevel), func(t *testing.T) {
+					// Create a roundtripper that captures the downstream HTTP request.
```
dimitarvdimitrov, line 924:
```suggestion
					t.Parallel()
					// Create a roundtripper that captures the downstream HTTP request.
```

maybe in parallel? I see the tripperware is shared between tests, but it should be concurrency-safe, right?

dimitarvdimitrov (approved):
LGTM, only two nitpicks, but I don't need to see this again

pkg/storage/ingest/reader.go
pracucci: Ah good point!

pkg/frontend/querymiddleware/roundtrip_test.go
pracucci: Yes, that was the plan, but I missed it.

pkg/storage/ingest/reader.go
pracucci: Done in https://github.com/grafana/mimir/pull/8808/commits/94d767c523f97d70cffbef4f5f3a3e574b6674e9

pkg/frontend/querymiddleware/roundtrip_test.go
pracucci: Done in https://github.com/grafana/mimir/pull/8808/commits/94d767c523f97d70cffbef4f5f3a3e574b6674e9

pkg/frontend/querymiddleware/codec_test.go
```diff
@@ -740,6 +741,7 @@ func TestPrometheusCodec_EncodeMetricsQueryRequest_ShouldPropagateHeadersInAllow
 	require.NoError(t, err)
 	require.Equal(t, []string{"true"}, req.Header.Values(compat.ForceFallbackHeaderName))
 	require.Equal(t, []string{"label"}, req.Header.Values(chunkinfologger.ChunkInfoLoggingHeader))
+	require.Equal(t, []string{string(api.EncodeOffsets(map[int32]int64{0: 1, 1: 2}))}, req.Header.Values(api.ReadConsistencyOffsetsHeader))
```
charleskorn, line 744: This assertion can sometimes flake due to the unpredictable iteration order of maps, for example:

```
--- FAIL: TestPrometheusCodec_EncodeMetricsQueryRequest_ShouldPropagateHeadersInAllowList (0.00s)
    codec_test.go:744:
        	Error Trace:	/mimir/pkg/frontend/querymiddleware/codec_test.go:744
        	Error:      	Not equal:
        	            	expected: []string{"v1=1:2,0:1"}
        	            	actual  : []string{"v1=0:1,1:2"}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 ([]string) (len=1) {
        	            	- (string) (len=10) "v1=1:2,0:1"
        	            	+ (string) (len=10) "v1=0:1,1:2"
        	            	 }
        	Test:       	TestPrometheusCodec_EncodeMetricsQueryRequest_ShouldPropagateHeadersInAllowList
```

pracucci: Thanks for raising it. I will fix it, thanks.

PR #8821, opened 2024-07-25
Fix setting of receivedAt timestamp by HA tracker.

pstibrany:
#### What this PR does

Fix small bugs in HA tracker:

* HA tracker updates KV store with up-to-date values every update interval. During such update, HA tracker used current time as "received at" value for all updated cluster, but that's incorrect.
* Furthermore, when HA tracker received update about new elected replica, it did not update `electedLastSeenTimestamp` correctly, keeping the old value.

Both of these could lead to storing more recent timestamp for "received at" field than compared to actually received requests, which could delay failover.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [na] Documentation added.
- [na] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.


pkg/distributor/ha_tracker_test.go
```diff
@@ -135,7 +136,7 @@ func TestHATrackerConfig_Validate(t *testing.T) {
 }
 
 // Test that values are set in the HATracker after WatchPrefix has found it in the KVStore.
-func TestWatchPrefixAssignment(t *testing.T) {
+func TestHATrackerWatchPrefixAssignment(t *testing.T) {
```
pstibrany, line 139: All HA tracker tests now have `TestHATracker` prefix, so that we can run them easily.

seizethedave (approved):
LGTM!

PR #8825, opened 2024-07-26
feat: support string extraobjects

davhdavh:
#### What this PR does

This PR fixes it so that both of these way of defining `extraObjects` work:

```yaml
extraObjects:
  - |
      apiVersion: v1
      kind: ConfigMap
      ...
extraObjects:
  - apiVersion: v1
    kind: ConfigMap
    ...
```

The former syntax allows to use pre-defined templates easier, see problem in https://stackoverflow.com/questions/75338320/helm-tpl-template-yaml-object-without-using-a-string


dimitarvdimitrov (commented):
Thanks for fixing this. Can you also expand the example in the values.yaml file just above `extraObjects` to include a string object too?

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -4462,4 +4462,15 @@ continuous_test:
...
+# alternatively, you can use strings, which lets you use the mimir defines:
+# extraObjects:
+# - |
+#   kind: ConfigMap
+#   apiVersion: v1
+#   metadata:
+#     name: extra-cm-{{ .Release.Name }}
+#   data: |
+#     extra.yml: "{{ include some-mimir-define }}"
+
```
dimitarvdimitrov, line 4474:
```suggestion
```

dimitarvdimitrov (approved):
Awesome! Thank you

dimitarvdimitrov:
I pushed a commit to add a changelog entry, hope you don't mind :)

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -4464,4 +4464,14 @@ continuous_test:
...
+# extraObjects:
+# - |
+#   kind: ConfigMap
+#   apiVersion: v1
+#   metadata:
+#     name: extra-cm-{{ .Release.Name }}
+#   data: |
+#     extra.yml: "{{ include some-mimir-define }}"
 extraObjects: []
+
```
dimitarvdimitrov, line 4477:
```suggestion
```

PR #8826, opened 2024-07-26
fix: remove double check for existence of config

davhdavh:
#### What this PR does

There is absolutely no reason for the `No component section mapping for xxx not found in values; submit a bug report if you are a user, edit mimir.componentSectionFromName if you are a contributor` error when the next line checks if it can find the config section. Just fall-back to the name given.

### DIFF
--- a/operations/helm/charts/mimir-distributed/templates/_helpers.tpl
+++ b/operations/helm/charts/mimir-distributed/templates/_helpers.tpl
@@ -455,7 +455,7 @@ Examples:
   "tokengen" "tokengenJob"
 -}}
 {{- $componentSection := index $componentsMap .component -}}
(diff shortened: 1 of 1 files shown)


operations/helm/charts/mimir-distributed/templates/_helpers.tpl
```diff
@@ -455,7 +455,7 @@ Examples:
   "tokengen" "tokengenJob"
 -}}
 {{- $componentSection := index $componentsMap .component -}}
-{{- if not $componentSection -}}{{- printf "No component section mapping for %s not found in values; submit a bug report if you are a user, edit mimir.componentSectionFromName if you are a contributor" .component | fail -}}{{- end -}}
+{{- if not $componentSection -}}{{- $componentSection := .component -}}{{- end -}}
```
dimitarvdimitrov, line 458: this might leave subtle bugs that we won't find out until someone complains. For example because some specialized value isn't set in our other tests and is missed during code review.

 It's clunky to edit this code, but it prevents bugs as it is

Issue #6008, opened 2023-09-12
All errors should be safe to be wrapped and retained


colega:
#### Is your feature request related to a problem? Please describe.

Some of our errors retain references to unsafe-ly built things like labels:

https://github.com/grafana/mimir/blob/e928a4b7b7ca243f92760e5ce6014d1e5cdff86f/pkg/ingester/errors.go#L77-L83

@pracucci [said](https://github.com/grafana/mimir/pull/5585/files#r1283453383):

> some errors carry the series labels but these series labels are only safe to read during the execution of push() because they're unmarshalled from the write request into a pool. So before returning, we effectively "make a copy" to ensure the error will be safe after the push() has returned.
> 
> As a retrospective, we could have better handled the returned error, maybe implementing a specific interface for unsafe error messages, because it's currently very obscure. But that's for another day :)

And I think that day has come and it's time to do it. Errors are something meant to escape the execution of a function, and there's no way to ensure they won't escape the caller. It's just a matter of time for this to cause corrupted error messages or even something worse.

#### Describe the solution you'd like

Any error should be safe to be stored for any time, even if that means a more expensive error creation/formatting.

#### Describe alternatives you've considered

Just ignore the problem.

duricanikolic:
We have made the first steps towards creating safe errors: this [PR](https://github.com/grafana/mimir/pull/5584) introduces the following interface, belonging to the `ingester` package:
```
type safeToWrap interface {
  safeToWrap()
}
```
as well as a type implementing that interface:
```
type safeToWrapError string

func (s safeToWrapError) safeToWrap() {}

func (s safeToWrapError) Error() string {
	return string(s)
}
```

Most of the errors returned by ingester are created by calling `safeToWrapError(msg)`.

duricanikolic:
This comment describes the current status of this issue. So far error handling has been improved in the following components: ingester and distributor.

## Main Achievements ##
- Consistency.
- Ability to send additional details about the error via a gRPC communication.
- Use gRPC status codes as `status_code` label in the request duration metrics and in Mimir's dashboards.

### Additional error details ###
Additional details about the gRPC errors returned by ingester and distributor are modelled by the following protobuf messages belonging to the `mimirpb` package:
```
message ErrorDetails {
  ErrorCause Cause = 1;
}

enum ErrorCause {
  UNKNOWN_CAUSE = 0;
  REPLICAS_DID_NOT_MATCH = 1;
  TOO_MANY_CLUSTERS = 2;
[...]

PR #8828, opened 2024-07-26
Removing legacy coded related to -ingester.return-only-grpc-errors

duricanikolic:
#### What this PR does
In #8699 we got rid of the `-ingester.return-only-grpc-codes` CLI option. This PR removes the legacy code that was present only because of the removed CLI flag.

Part of https://github.com/grafana/mimir/issues/6008 (migration plan)

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -260,9 +260,8 @@ func otlpHandler(
 				errorMsg string
 			)
 			if st, ok := grpcutil.ErrorToStatus(err); ok {
-				// TODO: This code is needed for backwards compatibility,
-				// and can be removed once -ingester.return-only-grpc-errors
-				// is removed.
+				// This code is needed for a correct handling of errors returned by the supplier function.
+				// These errors are created by using the httpgrpc package.
 				httpCode = httpRetryableToOTLPRetryable(int(st.Code()))
 				grpcCode = st.Code()
 				errorMsg = st.Message()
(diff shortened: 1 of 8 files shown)


pkg/distributor/otel.go
```diff
@@ -260,9 +260,8 @@ func otlpHandler(
 				errorMsg string
 			)
 			if st, ok := grpcutil.ErrorToStatus(err); ok {
-				// TODO: This code is needed for backwards compatibility,
-				// and can be removed once -ingester.return-only-grpc-errors
-				// is removed.
+				// This code is needed for a correct handling of errors returned by the supplier function.
+				// These errors are created by using the httpgrpc package.
```
dimitarvdimitrov, line 264: can we just return a `newValidationError` from the supplier and delete this piece of code too?

duricanikolic: Hi @dimitarvdimitrov 
Errors can also come from distributors, and they are never built by the `httprgrpc` package.

dimitarvdimitrov: I'm not sure I follow. The comment now says `This code is needed for a correct handling of errors returned by the supplier function.`. Can we change the supplier function to return non-httpgrpc errors so that we can remove the `if` branch? My understanding is that then there will be no piece of code within the distributor which returns an `httprgrpc` error

duricanikolic: We can decide to change the current implementation and to remove usage of the `httpgrpc` package from the `distributor`package, but that might not be an easy task, since the full parsing of requests both in otlp and http handler have to be changed. I am not even sure that we want to do it, because the otlp handler is supposed to return `httpgrpc` errors (to be verified).
However, that is not the goal of this PR, which removes the legacy code remained after `-ingester.return-only-grpc-errors` was removed.

duricanikolic: Hi @dimitarvdimitrov, 
WDYT about my previous comment?

dimitarvdimitrov: ah, ok, then I misunderstood what the comment above said. By `supplier` you actually mean the `supplier` and the `praser` together, which vastly increases the scope. 

Then I'm happy with the changes you've made

dimitarvdimitrov (approved):

PR #8829, opened 2024-07-26
dashboards: remove "All" option for namespace variable

dimitarvdimitrov:
#### What this PR does

Opening the dashboard and defaulting to "All" makes loading slow.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pracucci (approved):
Nice! Should we do the same for cluster variable?

dimitarvdimitrov:
I wasn't sure - the cluster label shouldn't affect the scope of the queries if you've already selected a single namespace. Also if we allow to select multiple namespaces, they can be in different clusters, so an "All" clusters might make the selection easier.

PR #8837, opened 2024-07-29
enhancement(helm): Add support for running continuous-test with GEM

bradleypettit:
#### What this PR does
When using the enterprise metrics image, continuous test pods will fail as they don't have any license file available. This PR adds the license when enterprise is enabled, as well as loading the mimir config file as a command line arg.

Additionally, it add supports for setting an endpoint for continuous-test in the values file, where before this was automatically derived from the configuration. This is necessary in certain environments where you might want to test functionality from an external endpoint (like an ingress), in order to test the same write/read path that other users would be using.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


aldernero (approved):
This looks good, thanks!

bradleypettit:
Thanks @aldernero - I've fixed the conflicts so should be good to merge.

PR #8847, opened 2024-07-29
Purge ConcurrencyController on membership changes

gotjosh:
#### What this PR does

This introduces a way to remove users from concurrencyController on membership changes in a safe way.

We need to wait until the manager is fully stopped to know that rule evaluation has completed, even if the manager for that user has been deleted already.

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/ruler/manager_test.go
+++ b/pkg/ruler/manager_test.go
@@ -489,10 +489,10 @@ func (cm *controllerMock) Allow(_ context.Context, _ *promRules.Group, _ promRul
 func (cm *controllerMock) Done(_ context.Context) {
 }
 
-func (cm *controllerMock) RemoveTenant(UserID string) {
+func (cm *controllerMock) MarkTenantForRemoval(userID string, _ chan struct{}) {
 	cm.mtx.Lock()
 	defer cm.mtx.Unlock()
-	cm.removed = append(cm.removed, UserID)
+	cm.removed = append(cm.removed, userID)
 }
 
 func (cm *controllerMock) Removed() []string {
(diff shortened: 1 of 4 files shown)


pkg/ruler/manager_test.go
```diff
@@ -67,6 +69,7 @@ func TestDefaultMultiTenantManager_SyncFullRuleGroups(t *testing.T) {
 
 		// Ensure the ruler manager has been stopped for all users.
 		assertManagerMockStopped(t, initialUser1Manager)
+		// assertUserRemovedFromConcurrencyController(t, concurrencyController, user1)
```
pracucci, line 72: Pay attention this is commented.

pkg/ruler/rule_concurrency.go
```diff
@@ -102,6 +103,13 @@ type MultiTenantConcurrencyController struct {
 	tenantConcurrency    map[string]*DynamicSemaphore
 }
 
+// RemoveTenant removes a tenant from the concurrency controller.
+func (c *MultiTenantConcurrencyController) RemoveTenant(userID string) {
+	c.tenantConcurrencyMtx.Lock()
+	defer c.tenantConcurrencyMtx.Unlock()
+	delete(c.tenantConcurrency, userID)
```
pracucci, line 110: If there's a race condition and we remove a tenant when the semaphore is != 0 then we'll introduce an inconsistency and possibly a panic once `Done()` will be called. I would remove a tenant only if `acquired` is `0` and return a `bool` from `RemoveTenant()` to signal whether the tenant was removed or not (if the tenant doesn't exist it should return `true`), so that the caller can log it as an error if tenant was not removed.

gotjosh:
We decided to go in a different direction in https://github.com/grafana/mimir/pull/8880

PR #8850, opened 2024-07-30
Mimir query engine: fix issue where `rate()` over native histograms could panic or return incorrect results

charleskorn:
#### What this PR does

This PR fixes an issue where multiple queries running `rate()` over native histograms simultaneously could panic or return incorrect results.

This could happen when multiple queries reused the same `FloatHistogram` instance due to a bug in `HPointRingBuffer`. `HPointRingBuffer` could retain a reference to `FloatHistogram`s while also returning a slice containing those `FloatHistogram`s to the pool, so a second query could modify those `FloatHistogram`s concurrently with the original query.

Note to reviewers: I've tried to build up this PR incrementally. I'd suggest reviewing each commit individually.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -161,7 +161,6 @@ func (b *FPointRingBuffer) Reset() {
 
 // Close releases any resources associated with this buffer.
 func (b *FPointRingBuffer) Close() {
-	b.Reset()
```
aknuds1, line 164: Why is the `Reset` call unnecessary? Which practical consequences does removing it have?

aknuds1 (commented):
Just wondering about the practical consequences of removing the `Reset` calls.

pkg/streamingpromql/types/hpoint_ring_buffer.go
```diff
@@ -172,6 +173,11 @@ func (b *HPointRingBuffer) NextPoint() (*promql.HPoint, error) {
 		copy(newSlice, b.points[b.firstIndex:])
 		copy(newSlice[pointsAtEnd:], b.points[:b.firstIndex])
 
+		// We must clear b.points before returning it to the pool, as the current query could continue using the
+		// FloatHistogram instances it contains a reference to, but a later user of b.points may otherwise reuse
+		// those instances instead of creating new FloatHistograms.
+		clear(b.points)
```
jhesketh, line 179: Just to help my understanding here.. is this because a future query may end up with the released slice pool?

charleskorn: Yep, exactly.

pkg/streamingpromql/types/fpoint_ring_buffer.go
charleskorn: Calling `Reset` is unnecessary because a `FPointRingBuffer` isn't expected to be used after `Close` is called.

aknuds1 (approved):
LGTM

jhesketh (approved):

PR #8855, opened 2024-07-30
blockbuilder: get rid of groupConsumer

narqo:
As we discussed in DM, this PR updates the #8823 getting rid of a separate `groupClient`. Having multiple kafka clients per instance adds unnecessary pressure to the broker, as the (group) client has to poll for the metadata and do the rebalance. 

Also, relying on lag of the group becomes trickier, because `kadm.Client.Lag` only returns the lag for the live group-members. That means, the consuming cycle, where we calculate the lag, has to take into account the group client's liveness, and readiness. Otherwise, the calls to lag calculation may return unexpected results.


codesome (approved):
Great work!

PR #8856, opened 2024-07-30
Chore: replace store_query_gate_ismyturn span with a log

colega:
There's no need for this verbose span that is almost always just a few microseconds log. We can just log on the parent span how much time we spent waiting.

I was also tempted to avoid logging at all if it was <1ms wait, but finally preferred to be consistent.

### DIFF
--- a/pkg/storegateway/bucket.go
+++ b/pkg/storegateway/bucket.go
@@ -746,7 +746,7 @@ func (s *BucketStore) limitConcurrentQueries(ctx context.Context, stats *safeQue
 	waited := time.Since(waitStart)
 
 	stats.update(func(stats *queryStats) { stats.streamingSeriesConcurrencyLimitWaitDuration = waited })
-	level.Info(spanlogger.FromContext(ctx, s.logger)).Log("msg", "waited for turn", "duration", waited)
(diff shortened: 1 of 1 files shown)


pkg/storegateway/bucket.go
```diff
@@ -741,14 +741,13 @@ func (s *BucketStore) recordRequestAmbientTime(stats *safeQueryStats, requestSta
...
 	waitStart := time.Now()
-	err = s.queryGate.Start(spanCtx)
-	stats.update(func(stats *queryStats) {
-		stats.streamingSeriesConcurrencyLimitWaitDuration = time.Since(waitStart)
-	})
+	err = s.queryGate.Start(ctx)
+	waited := time.Since(waitStart)
+
+	stats.update(func(stats *queryStats) { stats.streamingSeriesConcurrencyLimitWaitDuration = waited })
+	level.Info(spanlogger.FromContext(ctx, s.logger)).Log("msg", "waited for turn", "duration", waited)
```
dimitarvdimitrov, line 749: i guess this makes it a bit less cryptic

```suggestion
	level.Info(spanlogger.FromContext(ctx, s.logger)).Log("msg", "waited for turn on query concurrency gate", "duration", waited)
```

dimitarvdimitrov (approved):
thanks

PR #8857, opened 2024-07-30
docs: define ALLOY_VERSION

krajorama:
Otherwise defaults to current mimir or helm chart version, see https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution


narqo (approved):
Thank you 👍

PR #8846, opened 2024-07-29
End to end test rule concurrency

gotjosh:
#### What this PR does

Adds an end to ends test to ensure rule concurrency is working correctly. A follow up to https://github.com/grafana/mimir/pull/8146

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [x] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8858, opened 2024-07-30
Validate `-ruler.independent-rule-evaluation-concurrency-min-duration-percentage`

gotjosh:
#### What this PR does

To ensure that its value is 0 or greater.

NB: I double-checked that the code doesn't panic with 0, and I'll add a test in https://github.com/grafana/mimir/pull/8846.

Do let me know if you think a unit test is also a good idea because, in my head, it seemed superfluous.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/ruler/ruler.go
```diff
@@ -137,8 +138,8 @@ type Config struct {
 	RingCheckPeriod             time.Duration `yaml:"-"`
 	rulerSyncQueuePollFrequency time.Duration `yaml:"-"`
 
-	MaxIndependentRuleEvaluationConcurrency                    int64   `yaml:"max_independent_rule_evaluation_concurrency" category:"experimental"`
-	IndependentRuleEvaluationConcurrencyMinDurationPercentange float64 `yaml:"threshold_independent_rule_evaluation_concurrency" category:"experimental"`
+	MaxIndependentRuleEvaluationConcurrency                   int64   `yaml:"max_independent_rule_evaluation_concurrency" category:"experimental"`
```
gotjosh, line 141: typo fix on the variable name and the YAML name was completely incorrect.

pracucci (approved):

PR #8859, opened 2024-07-30
Alertmanager: Use default Mimir globals in Grafana configuration

santihernandezc:
This PR updates the `computeConfig()` method to use the Alertmanager's default global configuration even if the current config is empty. This allows the Alertmanager to use the default SMTP configuration in the email sender.


alexander-akhmetov (approved):
LGTM!

PR #8861, opened 2024-07-31
Query-frontend: Return `413` if active series response is too large

flxbk:
#### What this PR does

Response shards for responses to the `/active_series` endpoint need to be accumulated in the querier to deduplicate series from ingesters, and there's a limit for the response shard size to protect queriers from OOMs. If this limit is exceeded, the query-frontend currently responds with a `400 (Bad Request)` status code. This PR makes it respond with `413 (Request Entity Too Large)` instead.

Clients can use this status code to detect the condition and resubmit the request with an increased shard size.

n/a

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/shard_by_series_base.go
+++ b/pkg/frontend/querymiddleware/shard_by_series_base.go
@@ -72,7 +72,7 @@ func (s *shardBySeriesBase) shardBySeriesSelector(ctx context.Context, spanLog *
 	resp, err := doShardedRequests(ctx, reqs, s.upstream)
 	if err != nil {
 		if errors.Is(err, errShardCountTooLow) {
-			return nil, apierror.New(apierror.TypeBadData, fmt.Errorf("%w: try increasing the requested shard count", err).Error())
+			return nil, apierror.New(apierror.TypeTooLargeEntry, fmt.Errorf("%w: try increasing the requested shard count", err).Error())
 		}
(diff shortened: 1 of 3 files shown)


pkg/frontend/querymiddleware/shard_by_series_base.go
```diff
@@ -72,7 +72,7 @@ func (s *shardBySeriesBase) shardBySeriesSelector(ctx context.Context, spanLog *
 	resp, err := doShardedRequests(ctx, reqs, s.upstream)
 	if err != nil {
 		if errors.Is(err, errShardCountTooLow) {
-			return nil, apierror.New(apierror.TypeBadData, fmt.Errorf("%w: try increasing the requested shard count", err).Error())
```
dimitarvdimitrov, line 75: Doesn't this change the status code for all errors, regardless what caused them?

flxbk: What do you mean by "all errors"? This only applies to errors for which `errors.Is(err, errShardCountTooLow)` returns true. Am I missing something? 🤔

dimitarvdimitrov: 🙈 you're right. That's why I shouldn't do code reviews from my phone while waiting for a coffee

LGTM

dimitarvdimitrov (approved):

PR #8865, opened 2024-07-31
Prevent returning HTTP status 499 instead of 5xx

### LINKED ISSUE #8815: Mimir returns HTTP status 499 instead of 5xx
We've been receiving several `499` HTTP responses from Mimir when 5xx response code seems more appropriate. This prevents many HTTP clients from automatically retrying the request if enabled.

```
{"status":"error","errorType":"canceled","error":"expanding series: grpc: the client connection is closing"}
```
```
{"status":"error","errorType":"canceled","error":"attempted to read series at index 0 from store-gateway chunks stream, but the stream has failed: grpc: the client connection is closing"}
```

duricanikolic:
#### What this PR does

This PR fixes the bug reported in this [issue](https://github.com/grafana/mimir/issues/8815). Namely, before this PR in some cases Mimir used to return HTTP status code `499`, which is internally used to track the `context canceled` errors, but that should never be reported back to the client.

Function `globalerror.WrapGRPCErrorWithContextError()` is used for wrapping gRPC errors with their corresponding context errors. It assumes that all gRPC errors with the code `codes.Canceled` and `codes.DeadlineExceeded` could actually be wrapped with the `context.Canceled` and `context.DeadlineExceeded` errors. Although this is a valid assumptions, there are some cases where it is not true. One of those cases is the `grpc: the client connection is closing` error. This error was erroneously treated as `context canceled` error. This PR fixed the implementation of `globalerror.WrapGRPCErrorWithContextError()` accordingly.

Errors wrapped by `globalerror.WrapGRPCErrorWithContextError()` are used inside `querier.fetchSeriesFromStores()` to determine whether an erroneous attempt to fetch series from blocks should be retried or not. In case of `context.canceled` errors, the request should not be retried because it was actually canceled by the caller. But in the case of `grpc: the client connection is closing` error, the request should be required because the connection closing was not triggered by the caller. Before this PR the requests running into `grpc: the client connection is closing` errors were not retried, and their status code `499` used to be sent back to the caller.

Fixes #8815 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/globalerror/grpc_test.go
```diff
@@ -284,3 +300,77 @@ func checkErrorWithStatusDetails(t *testing.T, details []any, expected *mimirpb.
...
+	require.Error(t, err)
+	require.NotErrorIs(t, err, context.Canceled)
+
+	errWithCtx := WrapGRPCErrorWithContextError(err)
+	stat, ok := grpcstatus.FromError(errWithCtx)
+	require.True(t, ok)
+	require.Equal(t, codes.Canceled, stat.Code())
+	require.Equal(t, grpcClientConnectionIsClosingErr, stat.Message())
+	require.NotErrorIs(t, errWithCtx, context.Canceled)
+}
```
colega, line 368: I think the most important thing here is the last line:
```go
require.NotErrorIs(t, errWithCtx, context.Canceled)
```

But the method name doesn't reflect that. IMO, it makes harder to understand the test, because the test wasn't checking what "the error isn't context.Canceled".

OTOH, this test also does the wrapping itself, which is the other important part of the test.

In other words: I'd move the contents of this method to the test itself.

duricanikolic: Hi @colega, 
I have refactored the test. I kept the `checkGRPCConnectionIsClosing()` function, which actually checks that the passed error is a `grpc: the client connection is closing` error.
OTOH, I put the wrapping and the checks that both original and wrapped errors are not a `context cancled` error to the test itself.

pkg/util/globalerror/grpc_test.go
```diff
@@ -77,16 +83,24 @@ func TestWrapContextError(t *testing.T) {
 				expectedGrpcCode:   codes.DeadlineExceeded,
 				expectedContextErr: context.DeadlineExceeded,
 			},
+			"grpc.ErrClientConnClosing": {
+				origErr:            status.Error(codes.Canceled, grpcClientConnectionIsClosingErr),
```
jhalterman, line 87: Could use `grpc.ErrClientConnClosing` here

pkg/util/globalerror/grpc.go
```diff
@@ -24,10 +28,12 @@ func WrapGRPCErrorWithContextError(err error) error {
 	if stat, ok := grpcutil.ErrorToStatus(err); ok {
 		switch stat.Code() {
 		case codes.Canceled:
-			return &ErrorWithStatus{
-				UnderlyingErr: err,
-				Status:        stat,
-				ctxErr:        context.Canceled,
+			if stat.Message() != grpcClientConnectionIsClosingErr {
```
jhalterman, line 31: It's a bummer that we depend on the error message not changing. Perhaps we could compare the err to `grpc.ErrClientConnClosing`? Or convert that to a status, then compare the error message?

duricanikolic: Hi @jhalterman 
If we use `grpc.ErrClientConnClosing` lint will complain because that error is deprecated. And as far as I know `grpc.ErrClientConnClosing` is the only exported occurrence of the error message we are looking for.

pkg/util/globalerror/grpc_test.go
duricanikolic: If we use `grpc.ErrClientConnClosing` lint will complain because that error is deprecated.
We can use that error, but we need to ignore lint

pkg/util/globalerror/grpc.go
```diff
@@ -7,12 +7,23 @@ import (
...
 )
 
+var grpcClientConnectionIsClosingErr string
+
+func init() {
+	// Ignore deprecation warning for now
+	//nolint:staticcheck
+	if stat, ok := grpcutil.ErrorToStatus(grpc.ErrClientConnClosing); ok && stat.Code() == codes.Canceled {
+		grpcClientConnectionIsClosingErr = stat.Message()
+	}
```
colega, line 24: What if the condition isn't met? The string would be empty. Should we panic then to make sure that the code doesn't compile?

pkg/util/globalerror/grpc_test.go
```diff
@@ -284,3 +298,80 @@ func checkErrorWithStatusDetails(t *testing.T, details []any, expected *mimirpb.
 		require.Equal(t, expected, errDetails)
 	}
 }
+
+func TestGRPCClientClosingConnectionError(t *testing.T) {
```
colega, line 302:
```suggestion
func TestGRPCClientClosingConnectionError_IsNotContextCanceled(t *testing.T) {
```

pkg/util/globalerror/grpc.go
duricanikolic: Hi @colega,
My idea in this case was to consider all errors with `codes.Canceled` as `context canceled`.

duricanikolic: Or better:
- if `grpc.ErrClientConnClosing` gets removed (it will happen at a poi t), this code will not compile anymore, which is good
- if `grpc.ErrClientConnaclosing` gets modified  and it is not a gRPC error and its status is not `codes.Canceled` (not likely to happen, why should a deprecated error change?),  we’ll not filter out it anymore
- if `grpc.ErrClientConnaclosing` gets modified but it is still a gRPC error with status `codes.Canceled` (not likely to happen, why should a deprecated error change?),  we’ll not filter out the new message

colega: But the variable would become a lie.

duricanikolic: I personally prefer to have an empty string, to change a name of the variable, or to hardcode `grpc: the client connection is closing` than to panic. 
But I am flexible, I'll do what you prefer. So what do we want to do? Panic?

colega: I liked the hardcoded string as it was clear and tested, this is extra logic, extra things that can fail. But I don't seem to be aligned with @jhalterman on this, so I'm going to approve it as is.

colega (approved):

pkg/util/globalerror/grpc.go
duricanikolic: @colega, @jhalterman 
I also prefer the hardcoded version.
Anyway, I am going to merge this version, but if we agree to use the hardcoded string, I’ll prepare another PR.
Please let me know you are OK with it @jhalterman

PR #8867, opened 2024-07-31
ingest: make PartitionReader wait for in-flight fetches when it's stopping

narqo:
#### What this PR does

These PR updates the ingest storage's `PartitionReader` to shut down gracefully. That is, when its service enters the stopping phase, `PartitionReader` shouldn't try to interrupt the processing of the in-flight fetches batch. Also, it shouldn't commit if the processing didn't succeed and was cancelled due to stopping.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -172,13 +172,17 @@ func (r *PartitionReader) stopDependencies() error {
 
 func (r *PartitionReader) run(ctx context.Context) error {
 	for ctx.Err() == nil {
-		r.processNextFetches(ctx, r.metrics.receiveDelayWhenRunning)
+		err := r.processNextFetches(ctx, r.metrics.receiveDelayWhenRunning)
+		if err != nil && !errors.Is(err, context.Canceled) {
+			// Fail the whole service in case of a non-recoverable error.
+			return err
+		}
 	}
 
 	return nil
 }
 
-func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver prometheus.Observer) {
+func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver prometheus.Observer) error {
 	fetches := r.pollFetches(ctx)
 	r.recordFetchesMetrics(fetches, delayObserver)
 	r.logFetchErrors(fetches)
@@ -186,11 +190,11 @@ func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver 
 
 	err := r.consumeFetches(ctx, fetches)
 	if err != nil {
-		level.Error(r.logger).Log("msg", "error while consuming fetches", "err", err, "num_records", fetches.NumRecords())
-	} else {
-		r.enqueueCommit(fetches)
-		r.notifyLastConsumedOffset(fetches)
+		return fmt.Errorf("consume %d records: %w", fetches.NumRecords(), err)
 	}
+	r.enqueueCommit(fetches)
+	r.notifyLastConsumedOffset(fetches)
+	return nil
 }
 
 // processNextFetchesUntilTargetOrMaxLagHonored process records from Kafka until at least the maxLag is honored.
@@ -301,9 +305,11 @@ func (r *PartitionReader) processNextFetchesUntilLagHonored(ctx context.Context,
 				break
 			}
 
-			r.processNextFetches(ctx, r.metrics.receiveDelayWhenStarting)
+			err := r.processNextFetches(ctx, r.metrics.receiveDelayWhenStarting)
+			if err != nil {
+				return 0, err
+			}
 		}
-
 		if boff.Err() != nil {
 			return 0, boff.ErrCause()
 		}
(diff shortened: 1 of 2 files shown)


pkg/storage/ingest/reader.go
```diff
@@ -184,12 +184,13 @@ func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver
...
 
-	// TODO consumeFetches() may get interrupted in the middle because of ctx canceled due to PartitionReader stopped.
-	// 		We should improve it, but we shouldn't just pass a context.Background() because if consumption is stuck
-	// 		then PartitionReader will never stop.
-	r.consumeFetches(ctx, fetches)
-	r.enqueueCommit(fetches)
-	r.notifyLastConsumedOffset(fetches)
+	err := r.consumeFetches(ctx, fetches)
+	if err != nil {
+		level.Error(r.logger).Log("msg", "error while consuming fetches", "err", err, "num_records", fetches.NumRecords())
```
pracucci, line 189: I'm not sure we log as error when the context is canceled (which is the only error that could be returned). I would rather propagate the error back to the caller, up until `run()` and then in `run()` we check whether it's context canceled or a different error. If it's context canceled, we interrupt `run()` returning `nil`, while if it's an other we return the `err` (this will make the entire ingester to fail, which is what we want in case of a non recoverable). I know we currently can only return `context.Canceled` but I think will make code more future proof.

pracucci (commented):
Overall LGTM! I left a minor comment. Can we also test this without getting crazy? In a unit test you can mock the `Pusher` to control the consumption.

pkg/storage/ingest/reader.go
```diff
@@ -407,13 +408,18 @@ func (r *PartitionReader) consumeFetches(ctx context.Context, fetches kgo.Fetche
 		MaxBackoff: 2 * time.Second,
 		MaxRetries: 0, // retry forever
 	})
-
 	for boff.Ongoing() {
+		// If the PartitionReader is stopping and the ctx was cancelled, we don't want to interrupt the in-flight
+		// processing midway. Instead, we let it finish, assuming it'll succeed.
+		// If the processing fails while stopping, we log the error and let the backoff stop and bail out.
+		// There is an edge-case when the processing gets stuck and doesn't let the stopping process. In such a case,
+		// we expect the infrastructure (e.g. k8s) to eventually kill the process.
```
dimitarvdimitrov, line 416: the log message is a bit deceiving now - it says "will retry," but it won't retry if the context was cancelled. Can we make the message conditional ("not retrying" and "will retry")?

narqo: I update the log line below, removing the "will retry" part. I think, from the presence of the `num_retries` field, it should be relatively clear, what is happening in the logs.

narqo:
🆙 

I've updated the changes, following Marco's suggestions. PTAL

pkg/storage/ingest/reader.go
dimitarvdimitrov: if we're in a retry loop and then shut down the ingester, you can't know whether the last batch was successfully ingested or the backoff was interrupted. I'm still more inclined to make it dead obvious. Another way is to add a line which says "successfully ingested data from kafka."

pkg/storage/ingest/reader_test.go
```diff
@@ -142,6 +142,52 @@ func TestPartitionReader_ConsumerError(t *testing.T) {
 	assert.Equal(t, [][]byte{[]byte("1"), []byte("2")}, records)
 }
 
+func TestPartitionReader_ConsumerStopping(t *testing.T) {
+	const (
```
dimitarvdimitrov, line 146: can you make this a `t.Parallel()`? It takes 1.5s and most of it is just sleeping

pkg/storage/ingest/reader.go
narqo: Ah, I see it now. Unfortunately, we cannot tell if the loop will retry, while we are inside the loop. I.e. the context, the `boff` holds to can be cancelled between the calls to `boff.Wait()` and `boff.Ongoing()`. So the loop can bail, even though the log told that it will retry. This will make it even more confusing, IMHO.

Apart from logging every successful attempt, we can explicitly log the `context canceled`, when exit with an error (like the PR did it initially). But Marco had a good point, that it's, perhaps, redundant (see https://github.com/grafana/mimir/pull/8867#discussion_r1698707575).

dimitarvdimitrov: ok, this is more complicated than I anticipated :D let's keep it as-is.

dimitarvdimitrov (approved):

pkg/storage/ingest/reader.go
```diff
@@ -407,23 +414,30 @@ func (r *PartitionReader) consumeFetches(ctx context.Context, fetches kgo.Fetche
...
 		consumeStart := time.Now()
-		err := r.consumer.consume(ctx, records)
+		err := r.consumer.consume(consumeCtx, records)
 		r.metrics.consumeLatency.Observe(time.Since(consumeStart).Seconds())
 		if err == nil {
-			break
+			return nil
 		}
 		level.Error(r.logger).Log(
-			"msg", "encountered error while ingesting data from Kafka; will retry",
```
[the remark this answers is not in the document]
narqo: I've updated it to `should retry`. To my eye this reads as an _ok'ish_ middle ground between yours and @dimitarvdimitrov (see [above][1]) comments about this log line.

[1]: https://github.com/grafana/mimir/pull/8867#discussion_r1703795088

PR #8871, opened 2024-08-01
Introduce leaf nodes and "unknown" query component

chencs:
#### What this PR does
This PR introduces the concept of "leaf nodes" to the experimental `MultiQueuingAlgorithmTreeQueue`, and enforces a couple of rules:
* Leaf nodes may only exist at a tree's maximum depth (`len(queuingAlgorithms)` at tree creation)
* Non-leaf nodes may not contain a local queue
* Leaf nodes may not have children (this is already enforced by tree's max-depth)

This lets us efficiently skip memory allocation for non-leaf nodes, saving memory and CPU cycles.

#### Changes to structure/behavior
One structural change is required to enable this: previously, if `query-frontend.additional-query-queue-dimensions-enabled` was set and scheduler requests `AdditionalQueueDimensions` ([code](https://github.com/grafana/mimir/blob/035a8c44bfeb8fd376aa01f534b50064a8ceb235/pkg/scheduler/queue/queue.go#L54)) was _empty_, said requests would be enqueued to the parent node's local queue, e.g.:
```
root
└─ tenant-1 --> local queue: [queries with no additional queue dimensions]
   ├─ ingester
   ├─ store-gateway
   └─ ingester-and-store-gateway
```

With this change, such queries will instead be nested under a new node in the tree:
```
root
└─ tenant-1
   ├─ unknown
   │  └─ queries with no additional queue dimensions
   ├─ ingester
   ├─ store-gateway
   └─ ingester-and-store-gateway
```

The overall behavior is roughly the same, since tenant nodes round-robin between dequeuing from their children and their local queue. There is one fairly non-impactful edge case which will be removed: previously, a query with "unknown" queue dimensions may have jumped the queue in the following example scenario:
1. the tenant node is created by enqueuing queries _with_ queue dimensions (e.g., `ingester-and-store-gateway`)
2. queries with unknown queue dimensions are queued to the tenant's local queue
3. the first dequeue for the tenant will dequeue from the tenant's local queue, skipping all the `ingester-and-store-gateway` queries which were enqueued _first_.

With this change, because "unknown" queue dimensions become their own nodes, they will be queued and dequeued from in the expected FIFO order.

There is a possible unexpected behavior change -- if anyone is using `unknown` as a queue dimension, scheduler requests with empty `AdditionalQueueDimensions` and requests with `unknown` `AdditionalQueueDimensions` will be mixed.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
+++ b/pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
@@ -140,15 +140,16 @@ func newNode(name string, height int, da QueuingAlgorithm) (*Node, error) {
 		name:             name,
 		localQueue:       nil,
 		queuePosition:    localQueueIndex,
-		queueOrder:       make([]string, 0),
-		queueMap:         make(map[string]*Node, 1),
 		height:           height,
 		queuingAlgorithm: da,
 	}
-	// if the node is a leaf node, it gets memory allocated towards a local queue and cannot have child nodes
+	// if the node is a leaf node, it gets memory allocated towards a local queue and cannot have child queues
 	if height == 0 {
 		n.localQueue = list.New()
-		n.queueMap = nil
+	} else {
+		// any other kind of node is allowed to have children, but not a local queue
+		n.queueMap = make(map[string]*Node, 1)
+		n.queueOrder = make([]string, 0)
 	}
 	return n, nil
 }
--- a/pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
+++ b/pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
@@ -43,7 +43,7 @@ func NewTree(queuingAlgorithms ...QueuingAlgorithm) (*MultiQueuingAlgorithmTreeQ
 	if len(queuingAlgorithms) == 0 {
 		return nil, fmt.Errorf("cannot create a tree without defined QueuingAlgorithm")
 	}
-	root, err := newNode("root", 0, queuingAlgorithms[0])
+	root, err := newNode("root", 0, queuingAlgorithms[0], len(queuingAlgorithms) == 1)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
```diff
@@ -85,6 +85,9 @@ func (t *MultiQueuingAlgorithmTreeQueue) Dequeue() (QueuePath, any) {
 // path is relative to the root node; providing a QueuePath beginning with "root"
 // will create a child node of the root node which is also named "root."
 func (t *MultiQueuingAlgorithmTreeQueue) EnqueueBackByPath(path QueuePath, v any) error {
+	if len(path) != len(t.algosByDepth)-1 {
+		return fmt.Errorf("can't enqueue object to non-child node")
```
francoposa, line 89: > non-child node
non-leaf node?

pkg/scheduler/queue/tenant_queues.go
```diff
@@ -140,11 +147,20 @@ func (qb *queueBroker) enqueueRequestFront(request *tenantRequest, tenantMaxQuer
...
+		dimensions := QueuePath{}
 		if schedulerRequest, ok := request.req.(*SchedulerRequest); ok {
+			// TODO (casie): Clean this up when deprecating old TreeQueue
+			if _, ok := qb.tree.(*MultiQueuingAlgorithmTreeQueue); ok && len(schedulerRequest.AdditionalQueueDimensions) == 0 {
+				// if using new MultiQueueingAlgorithmTreeQueue and AdditionalQueueDimensions is empty,
+				// enqueue request to a node "unknown" to keep all requests enqueued on leaf nodes.
+				dimensions = append(dimensions, unknownQueueDimension)
+			} else {
+				dimensions = schedulerRequest.AdditionalQueueDimensions
+			}
```
francoposa, line 159: let's leverage `schedulerRequest.ExpectedQueryComponentName()` for this if we can? It's supposed to help hide the details of grokking this stuff from `AdditionalQueueDimensions` which will probably go away/move/be renamed.

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
```diff
@@ -126,37 +129,56 @@ type Node struct {
 	childrenChecked  int
 }
 
-func newNode(name string, depth int, da QueuingAlgorithm) (*Node, error) {
+func newNode(name string, depth int, da QueuingAlgorithm, isLeaf bool) (*Node, error) {
```
francoposa, line 132: could we pass in height instead of depth, and have `height == 0` be the check for whether it is a leaf?

francoposa (approved):
looks good. we can come back and remove all "local queue selection" stuff from the queuing algo(s), either in a single PR or as part of adding the new algos

PR #8874, opened 2024-08-01
Jsonnet: change min/max ingester replicas setting to per-zone when ingest storage is enabled

pracucci:
#### What this PR does

In this PR I'm changing the min/max ingester replicas setting to be per-zone when ingest storage is enabled. It makes it easier to reason, given in the ingest storage zones can be added and removed.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,9 @@
 
 ### Jsonnet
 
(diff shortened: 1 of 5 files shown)


narqo (approved):
🔥

PR #8878, opened 2024-08-01
Dashboards: add 'Read path' selector to 'Mimir / Queries' dashboard

pracucci:
#### What this PR does

The Queries dashboard doesn't offer a consistent view over queries. As supported by the OSS jsonnet, at Grafana Labs we currently have 2 read path stacks:
- one for user queries
- one for Mimir rule evaluations

The panels in the Queries dashboard show a mix of the two. For components shared between the 2 read stacks (ingesters and store-gateways) we show metrics related to all queries. For components specific to a stack, we don't query `ruler-*` components, so we miss rule evaluations from Queries dashboard. This is confusing to me.

The easy solution would be to always add `ruler-*` components to the query, but after a quick internal feedback loop looks like there's people liking the ability to see metrics for a specific read path (I see value as well).

In this PR I'm proposing to add a "Read path" selector to the Queries dashboard. It has 3 options:

- All
- Main
- Remote ruler

Note: I'm looking for better option names, any suggestion?

Here you can see a video of the selector in action:

https://github.com/user-attachments/assets/6a397d13-b0d7-4fb3-9161-6179634a409b

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir-mixin/config.libsonnet
+++ b/operations/mimir-mixin/config.libsonnet
@@ -88,6 +88,10 @@
       alertmanager: ['alertmanager', 'cortex', 'mimir', 'mimir-backend.*'],
       overrides_exporter: ['overrides-exporter', 'mimir-backend.*'],
 
+      // The following arejob matchers used to select all components in the read path.
+      standard_read_path: std.uniq(std.sort(self.query_frontend + self.query_scheduler + self.querier)),
+      ruler_read_path: std.uniq(std.sort(self.ruler_query_frontend + self.ruler_query_scheduler + self.ruler_querier)),
+
       // The following are job matchers used to select all components in a given "path".
       write: ['distributor.*', 'ingester.*', 'mimir-write.*'],
       read: ['query-frontend.*', 'querier.*', 'ruler-query-frontend.*', 'ruler-querier.*', 'mimir-read.*'],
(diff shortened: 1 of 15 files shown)


pracucci:
If we agree on the approach, then I will spend a week or so fixing the linter 😠

operations/mimir-mixin/config.libsonnet
```diff
@@ -88,6 +88,10 @@
       alertmanager: ['alertmanager', 'cortex', 'mimir', 'mimir-backend.*'],
       overrides_exporter: ['overrides-exporter', 'mimir-backend.*'],
 
+      // The following arejob matchers used to select all components in the read path.
```
charleskorn, line 91:
```suggestion
      // The following are job matchers used to select all components in the read path.
```

charleskorn (approved):
LGTM modulo typo

charleskorn:
> Note: I'm looking for better option names, any suggestion?

Perhaps "user" rather than "standard"? It's still not perfect, but might be a bit clearer

operations/mimir-mixin/config.libsonnet
pracucci: Done, thanks

pracucci:
> > Note: I'm looking for better option names, any suggestion?
> 
> Perhaps "user" rather than "standard"? It's still not perfect, but might be a bit clearer

Discussed offline. Charles offered few better options, I picked "main".

PR #8884, opened 2024-08-02
Make indepedent rule concurrency metrics per tenant

gotjosh:
#### What this PR does

I think that makes these per tenant is worth the trouble as they will allow us to easily spot which tenants have a need more concurrency over there without having to any corelation between the rule evalution metrics.

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/ruler/rule_concurrency.go
gotjosh, line 1: Do I know need to keep track of when a tenant gets shared away to unregister these metrics? I think the volume is OK enough (4 metrics per tenant) that I might not have a need to unregister them but I could be wrong - please let me know if there are any easy ways to do this that does not involve keeping track of registries...

pkg/ruler/rule_concurrency.go
```diff
@@ -135,8 +139,8 @@ type TenantConcurrencyController struct {
 // Done releases a slot from the concurrency controller.
 func (c *TenantConcurrencyController) Done(_ context.Context) {
 	c.globalConcurrency.Release(1)
-	c.metrics.CurrentConcurrency.Dec()
-	c.metrics.ReleaseTotal.Inc()
+	c.metrics.SlotsInUser.WithLabelValues(c.tenantID).Dec()
```
pracucci, line 142: [nit] We may store the per-tenant metric directly in the `TenantConcurrencyController` struct, so that we don't have to call `WithLabelValues()` each time. There's a small perf overhead calling `WithLabelValues()` which we could simply avoid.

pracucci (approved):
Now we re-introduced the purge problem :) A tenant metrics are never removed. Don't get crazy, I would not fix it for now, just something to be aware of.

pkg/ruler/rule_concurrency.go
gotjosh: Done in [1e6a32a](https://github.com/grafana/mimir/pull/8884/commits/1e6a32ad76d9a3776402d6bd744c96dd736d643b)

gotjosh:
> Now we re-introduced the purge problem :) A tenant metrics are never removed. Don't get crazy, I would not fix it for now, just something to be aware of.

https://github.com/grafana/mimir/pull/8884/files/5ae12cd85f6816c08ef6b59da08b65e3e4d4745d#diff-3e8efecf11fa9185935eea3d3c9c7993f99422353a391190657877e902875a37

Thanks for answering - let me ponder about what we can do here.

PR #8886, opened 2024-08-02
Optimize 'last produced offsets' fetching by caching existing partitions

pracucci:
#### What this PR does

This PR is a follow up of https://github.com/grafana/mimir/pull/8808.

In `main` we use `kgo.Client.ListEndOffsets()` to fetch the end offset of each partition in the topic, which is very convenient but under the hood it does 2 things sequentially:

1. Refresh topic metadata to get the updated list of partitions
2. Fetch the end offset of each partition (a partition is queried from the partition leader). Multiple brokers are queried concurrently.

In Mimir case, we don't need to a perfectly fresh view over existing partitions (and we also pre-provision partitions so we don't even expect partitions to change), we just need the guarantee that the end offsets we fetch is updated (no stale). If some partitions are skipped, it's not a big deal, because strong consistency would be enforced anyway in the ingester.

So, in this PR I propose to cache the step (1) "Refresh topic metadata to get the updated list of partitions" for a short period of time (15s). To do it, we have to build the request ourselves, but it's not complicated and it's the same approach we already took in `partitionOffsetClient.fetchPartitionOffset()`.

> 2. Fetch the end offset of each partition (a partition is queried from the partition leader). Multiple brokers are queried concurrently.

I know what you're thinking. If the Kafka-compatible backend you use doesn't have leaders (any broker is a leader for any partition) then there's no need to shard such request! We can request all partitions to 1 random broker. And you're right! It's what I will try to do in a next follow up PR (idea is to add a config option to specify whether the in-use Kafka-compatible backend has leader brokers or not). The changes in this PR will make it easier to do.

#### Test

I've tested the change in a Mimir dev cluster and here you can see the impact (change rolled out at 10:40):
- Left: average latency
- Right: p99 latency

Screenshot 2024-08-05 at 17 12 34

The results match the expectation. I was expecting to cut about 100ms for the average latency, which is close to what we observe.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


twmb:
Not sure the place to put this, so commenting here; there is a bit of caching within kgo itself --

* kgo.Request splits ListOffsets and issues it to the proper brokers
* To discover the proper brokers, it issues Metadata (pt 1 of the PR description)
* kgo actually tries to see if the metadata for the topics is already cached, https://github.com/twmb/franz-go/blob/6b61d17383b1e0c766143118efa39f1f320a3b2b/pkg/kgo/client.go#L2602
* If topics are not cached, a metadata request is issued https://github.com/twmb/franz-go/blob/6b61d17383b1e0c766143118efa39f1f320a3b2b/pkg/kgo/client.go#L2431-L2436
* The cache itself uses MetadataMinAge for expiry: https://github.com/twmb/franz-go/blob/master/pkg/kgo/client.go#L2415-L2416

The problem this PR is working around is that kadm.ListEndOffsets issues two requests: one Metadata request, and then a follow up ListOffsets request. The caching described above works for ListOffsets. However, the Metadata request is _not_ cached.

I think when I implemented the requesting in kgo, I opted user-issued Metadata requests _out_ of caching figuring that if a person is deliberately issuing a metadata request, we shouldn't return stale information. Perhaps that's not the best assumption though, and over the years, I've introduced more metadata caching for other requests.

There could be a few changes within kgo to improve the situation for (1) here. First, as follow up to your commit [`993544c3`](https://github.com/twmb/franz-go/commit/993544c3), any manually issued metadata request should also call `storeCachedMappedMetadata`. This would at least prepopulate the metadata cache.

Second, three options:
* Always use the mapped metadata cache for user issued `Metadata` requests
* Introduce a new API, `RequestCachedMetadata(req *kmsg.MetadataRequest, expiry time.Duration) (*kmsg.MetadataResponse, error)`. If the metadata exists and is cached for less than `expiry`, return it. If not, issue the metadata request (and force the response through the cache)
* Introduce a new API, `UseCache(context.Context) context.Context` that can be used to query any internal cache when manually issuing a metadata request

My vote is the first or second bullet point. I _think_ making these changes would avoid the need for this PR, but I'm not positive (not looking too closely).

pracucci:
Thanks @twmb for always be so kind to look at our work and give suggestions! I will reply to you tomorrow.

pracucci:
Thanks a lot @twmb for jumping on this! I'm also not sure where is the best place to have this conversation. Let me quickly reply here, but then we can probably move the conversation to a franz-go issue.

> To discover the proper brokers, it issues Metadata (pt 1 of the PR description)

I'm not sure you're talking about the same thing, so let me just clarify my take. The pt1 of the PR description refers to the Metadata request issued to discover the partitions:
https://github.com/twmb/franz-go/blob/6b61d17383b1e0c766143118efa39f1f320a3b2b/pkg/kadm/metadata.go#L432-L436

Due to how we run Mimir, we don't expect partitions to change frequently (actually they may never change). Reason is that we expect partitions to be pre-provisioned (e.g. create at least 1K partitions in the topic, even if we use just few, so that when we autoscale the number of partitions used by Mimir, we can assume partitions are already created in Kafka).

Given we don't expect the number of partitions to change frequently, we can cache the list of existing partitions for some time.

gotjosh (approved):

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -238,3 +249,136 @@ func (p *partitionOffsetClient) FetchTopicLastProducedOffsets(ctx context.Contex
...
+	return list, nil
+}
+
+// listTopicPartitionIDs returns a list of all partition IDs in the topic.
+func (p *partitionOffsetClient) listTopicPartitionIDs(ctx context.Context) ([]int32, error) {
+	topics, err := p.admin.ListTopics(ctx, p.topic)
+	if err != nil {
+		return nil, err
+	}
+	if err := topics.Error(); err != nil {
```
gotjosh, line 326: [nit] this threw me off for a minute.

```suggestion
        // Calling .Error() returns only the first error. However, we're requesting only one topic, so there should only be 1 error anyway. 
	if err := topics.Error(); err != nil {
```

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -238,3 +249,136 @@ func (p *partitionOffsetClient) FetchTopicLastProducedOffsets(ctx context.Contex
...
+
+	// Fetch partition IDs.
+	partitionIDs, err := p.listTopicPartitionIDs(ctx)
+	if err != nil {
+		return partitionIDs, err
+	}
+
+	// Cache it.
+	p.topicPartitionIDsMx.Lock()
+	p.topicPartitionIDs = slices.Clone(partitionIDs)
```
gotjosh, line 379: [nit] Do we need a clone here? The return type from `listTopicPartitionIDs` are a slice of `[]int32` anyways.

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -238,3 +249,136 @@ func (p *partitionOffsetClient) FetchTopicLastProducedOffsets(ctx context.Contex
...
+	// Execute the request.
+	shards := p.client.RequestSharded(ctx, req)
+
+	for _, shard := range shards {
+		if shard.Err != nil {
+			return nil, shard.Err
+		}
+
+		res := shard.Resp.(*kmsg.ListOffsetsResponse)
+		if len(res.Topics) != 1 {
```
gotjosh, line 295: Do we also need to check `for != 0` if the response contains 0 topics it might panic below.

gotjosh (approved):
LGTM! Great job.

My comments are optional and I have one question.

pkg/storage/ingest/partition_offset_client.go
pracucci: Not sure adding this comment is much helpful. It returns only 1 error, but we care about the 1st error anyway. This is how all `Error()` functions work in franz-go (they return the 1st error) and this pattern is used everywhere in `pkg/storage/ingest/`.

pkg/storage/ingest/partition_offset_client.go
pracucci: I'm cloning it, so that if the `listTopicPartitionIDsWithCache()` caller modifies the returned `[]int32` then the cached version is not modified too.

pkg/storage/ingest/partition_offset_client.go
pracucci: I can't understand this comment. If the number of `res.Topics` is different than 1, we return error. So, once we reach the lines below there's the guarantee that `res.Topics[0]` exists. I can't see how it could panic 🤔

pracucci:
I'm going to merge this PR to move forward. You can follow up on the `panic` comment. If turns out we have a bug, then I will open a follow up PR to fix it.

pkg/storage/ingest/partition_offset_client.go
gotjosh: Ignore me - I read this entirely wrong 🤦

PR #8890, opened 2024-08-02
blockbuilder: pass partition assignment via flag

narqo:
These changes are the follow-up to #8823

Instead of configuring the number of instances and kafka partitions, as long as asking for a `POD_NAME` env, here we add two new flags

- `-block-builder.instance-id`, which defaults to the hostname;
- `-block-builder.kafka.partition-assignment`, which must be a mapping from the instance-id to a static list of partitions, assigned to the instance.

Example of usage (with docker-compose):

```yaml
mimir-block-builder-0:
  command:
  - /mimir
  - -block-builder.instance-id=instance-0
  - -block-builder.kafka.partition-assignment={"0":[0,1,2]}

mimir-block-builder-1:
  hostname: mimir-block-builder-1
  command:
  - /mimir
  - -block-builder.kafka.partition-assignment={"1":[3,4,5]}
```

When using in Kubernetes, we will generate the full assignment via jsonnet, and pass the assignment to the block builder's stateful set.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -4,6 +4,7 @@ package blockbuilder
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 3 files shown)


narqo:
We've tried it in one of our test instances and it works as expected. I will merge it for now. We will fix anything new in later PRs.

PR #8905, opened 2024-08-05
Compactor: fix a race condition between different compactor replicas causing "corrupted" bucket index

### LINKED ISSUE #8687: compactor: comcurrent cleanups for the same user may result in invalid bucket index
#### Describe the bug

When two compactors are cleaning up the same user the bucket index may contain blocks which have already been deleted. As a result there is an increase in consistency check errors because the store-gateways cannot sync all the blocks from the bucket index, while the queriers keep requesting them.

#### To Reproduce

General flow of cleaning up a user (over-simplified and not all steps are listed):

1. fetch bucket index
1. start updating bucket index:
    1. list blocks in bucket
    1. fetch metas for new blocks (not in the index until now)
    1. list deletion markers
    1. fetch metas for new deletion markers (not in the index until now)
1. delete blocks marked for deletion; this also modifies the bucket index in memory
1. clean up partial blocks
1. upload updated bucket index to the bucket

In this case compactor-18 and compactor-36 were cleaning up the same tenant. As 

* compactor-18 was at step 3 while compactor-36 went through steps 1, 2a, 2b. 
* compactor-36 started with step 2c about 40 seconds after -18 had finished the cleanup (so after step 5)

<details><summary>Logs</summary>
<p>

```
2024-07-05T07:27:51.043194181Z	compactor-18 started blocks cleanup and maintenance 
2024-07-05T07:27:51.321589764Z	compactor-18 fetched existing bucket index 
2024-07-05T07:27:51.323114709Z	compactor-18 marked blocks for deletion 
2024-07-05T07:27:56.663910075Z	compactor-18 listed all blocks in storage 
2024-07-05T07:29:15.128260749Z	compactor-18 fetched blocks metas for newly discovered blocks 
2024-07-05T07:29:18.066056362Z	compactor-18 listed deletion markers 
2024-07-05T07:29:59.045060356Z	compactor-18 updated deletion markers for recently marked blocks 
2024-07-05T07:29:59.599287921Z	compactor-36 started blocks cleanup and maintenance 
2024-07-05T07:29:59.833082702Z	compactor-36 fetched existing bucket index 
2024-07-05T07:29:59.834543052Z	compactor-36 marked blocks for deletion 
2024-07-05T07:30:05.206666849Z	compactor-36 listed all blocks in storage 
2024-07-05T07:30:12.863034434Z	compactor-18 deleted block marked for deletion 01J1ZN4WFB4F58YNECD5KK2SDZ
2024-07-05T07:30:13.835748541Z	compactor-18 cleaned up partial blocks 
2024-07-05T07:30:18.003318897Z	compactor-18 completed blocks cleanup and maintenance 
2024-07-05T07:31:54.247794639Z	compactor-36 fetched blocks metas for newly discovered blocks 
2024-07-05T07:31:57.143349628Z	compactor-36 listed deletion markers 
2024-07-05T07:32:44.656111872Z	compactor-36 updated deletion markers for recently marked blocks 
2024-07-05T07:32:47.67501721Z	compactor-36 cleaned up partial blocks 
2024-07-05T07:32:50.994079855Z	compactor-36 completed blocks cleanup and maintenance 
```

</p>
</details> 

@pstibrany writes:

so compactor-36 has seen that block `01J1ZN4WFB4F58YNECD5KK2SDZ` (directory) exists ([done in this method](https://github.com/grafana/mimir/blob/94028daf35acc5a90489b97ad6fe6acac98c6c2f/pkg/storage/tsdb/bucketindex/updater.go#L76)). Notice that this method logs `fetched blocks metas for newly discovered blocks` just before it stops. In the meantime (after `listed all blocks in storage`, but before `fetched blocks metas for newly discovered blocks`) block `01J1ZN4WFB4F58YNECD5KK2SDZ` was deleted.

Since block was already in the bucket index, compactor-36 then tried to see if deletion mark exists ([this method](https://github.com/grafana/mimir/blob/94028daf35acc5a90489b97ad6fe6acac98c6c2f/pkg/storage/tsdb/bucketindex/updater.go#L177-L218)). But block was deleted in the meantime, so deletion mark was not found.

Since block was no longer marked for deletion, it was simply left in the index that compactor-38 uploaded, overriding correct index from compactor-18

#### Expected behavior

No consistency check errors

pracucci:
#### What this PR does

This PR should fix #8687. The issue describes the problem, and the simple idea of this solution is explained in the (intentionally) long comment in `pkg/storage/tsdb/bucketindex/updater.go`. The test I've added fails on `main` and passes in this PR.

Fixes #8687

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
LGTM. I still haven't figured out how to have a more general test for block cleaning concurrency. Thanks for the fix

pracucci:
> I still haven't figured out how to have a more general test for block cleaning concurrency. Thanks for the fix

Me neither. I tried reproducing the issue adding random sleeps but I couldn't reproduce it.

seizethedave:
> >I still haven't figured out how to have a more general test for block cleaning concurrency. Thanks for the fix
> 
> Me neither. I tried reproducing the issue adding random sleeps in the objstore.Bucket functions but I couldn't reproduce it.

I've had thoughts of turning substantial parts of the block cleaner into an airflow-like DAG process that advances through the steps it needs to perform with step retries where possible. Easier to imagine enumerating all possible concurrent executions if it were like that. (Though, it wouldn't address different executions inside of bucketindex methods.)

seizethedave (approved):

pracucci:
> > > I still haven't figured out how to have a more general test for block cleaning concurrency. Thanks for the fix
> > 
> > 
> > Me neither. I tried reproducing the issue adding random sleeps in the objstore.Bucket functions but I couldn't reproduce it.
> 
> I've had thoughts of turning substantial parts of the block cleaner into an airflow-like DAG process that advances through the steps it needs to perform with step retries where possible. Easier to imagine enumerating all possible concurrent executions if it were like that. (Though, it wouldn't address different executions inside of bucketindex methods.)

Not sure rewriting everything is ever a good idea. The few times we tried we failed spectacularly. I always strongly recommend small iterative changes, and try to do the smallest change to fix an issue when possible (not always possible, but sometimes it is).

pracucci:
@dimitarvdimitrov I thought about another possible race condition: the deletion happens between listing deletion markers and fetching their content for **new** markers not existing in the index yet. Due to how deletion delay works it shouldn't happen in the real work (we always expect to delete blocks for which the deletion markers were already in the index), but since the fix is even more trivial I've just fixed it (and added a test case for it too).

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -1266,104 +1267,180 @@ func TestBlocksCleaner_RaceCondition_CleanerUpdatesBucketIndexWhileAnotherCleane
...
+								<-cleaner2ListMarkersStarted
+
+								// Then signal the deletion happened with some delay. The delay is used to make
+								// sure this deletion completes before the 2nd cleaner is unblocked.
+								once.Do(func() {
+									go func() {
+										time.Sleep(100 * time.Millisecond)
+										close(cleaner2GetDeletionMarkUnblocked)
+									}()
+								})
```
dimitarvdimitrov, line 1351: this looks more appropriate for a `postDeleteHook`. It will be less flaky

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -1266,104 +1267,180 @@ func TestBlocksCleaner_RaceCondition_CleanerUpdatesBucketIndexWhileAnotherCleane
...
+			mockBucketClients: func(bucketClient objstore.Bucket) (cleaner1BucketClient, cleaner2BucketClient objstore.Bucket) {
+				cleaner2ListMarkersStarted := make(chan struct{})
+				cleaner2GetDeletionMarkUnblocked := make(chan struct{})
+
+				// Mock the 1st cleaner to delete deletion marks when the 2nd cleaner is between the listing
+				// of deletion marks and fetching their content (for new markers).
+				cleaner1BucketClient = &hookBucket{
+					Bucket: bucketClient,
+					preDeleteHook: func() func(context.Context, string) {
+						once := sync.Once{}
```
dimitarvdimitrov, line 1336: Isn't this a new `Once` instance every time the `preDeleteHook` is called? I don't understand how it's different from just calling the func below upon every invocation of `preDeleteHook`. Having said that, I still think a `postDeleteHook` is better.

pkg/compactor/blocks_cleaner_test.go
pracucci: Agree! I don't know why I haven't thought about it before. Done in https://github.com/grafana/mimir/pull/8905/commits/55b097c390f393f939793a330d2d5fd843cfd0b3

pkg/compactor/blocks_cleaner_test.go
pracucci: The `Once` is global because a channel can only be closed once, so we protect from it. However, in this specific test we only expect 1 block to be deleted, so it shouldn't matter much.

pkg/compactor/blocks_cleaner_test.go
pracucci: I've re-run the test on top of `main` and both test cases fail.

pkg/compactor/blocks_cleaner_test.go
dimitarvdimitrov: I had missed that it's a `func() func(...)`, so the closure with the once is reused for every call of `Delete()` on the client. SGTM

dimitarvdimitrov (approved):
Thanks!

Issue #8875, opened 2024-08-01
Flaky TestBucketStore_PersistsLazyLoadedBlocks


pracucci:
I've seen `TestBucketStore_PersistsLazyLoadedBlocks` being flaky in [this CI run](https://github.com/grafana/mimir/actions/runs/10194927456/job/28202561611?pr=8873):

```
--- FAIL: TestBucketStore_PersistsLazyLoadedBlocks (1.08s)
    logger.go:25: 2024-08-01 08:43:57.524147263 +0000 UTC m=+365.990533634 level debug component block.MetaFetcher msg fetching meta data concurrency 20
    logger.go:25: 2024-08-01 08:43:57.531194746 +0000 UTC m=+365.997581117 level info component block.MetaFetcher msg successfully synchronized block metadata duration 7.097585ms duration_ms 7 cached 6 returned 6 partial 0
    logger.go:25: 2024-08-01 08:43:57.531317253 +0000 UTC m=+365.997703624 level debug msg loading new block id 01J46HRCD9XS8BERCYGHW3PFQR
    logger.go:25: 2024-08-01 08:43:57.531389988 +0000 UTC m=+365.997776369 level debug msg the index-header doesn't exist on disk; recreating path /tmp/TestBucketStore_PersistsLazyLoadedBlocks3180052183/001/01J46HRCD9XS8BERCYGHW3PFQR/index-header
    logger.go:25: 2024-08-01 08:43:57.532117506 +0000 UTC m=+365.998503877 level debug msg loading new block id 01J46HRCHC4AZG34SFTDEWVW22
    logger.go:25: 2024-08-01 08:43:57.532199388 +0000 UTC m=+365.998585789 level debug msg the index-header doesn't exist on disk; recreating path /tmp/TestBucketStore_PersistsLazyLoadedBlocks3180052183/001/01J46HRCHC4AZG34SFTDEWVW22/index-header
    logger.go:25: 2024-08-01 08:43:57.532711376 +0000 UTC m=+365.999097747 level debug msg loading new block id 01J46HRCMP0TNC6AJKP0SEXMYG
    logger.go:25: 2024-08-01 08:43:57.532789451 +0000 UTC m=+365.999175822 level debug msg the index-header doesn't exist on disk; recreating path /tmp/TestBucketStore_PersistsLazyLoadedBlocks3180052183/001/01J46HRCMP0TNC6AJKP0SEXMYG/index-header
[...]

PR #8907, opened 2024-08-05
store-gateway: fix flaky TestBucketStore_PersistsLazyLoadedBlocks

dimitarvdimitrov:
This test used to use `time.Sleep()` which is flaky. I changed it to use assert.Eventually(), which is less flaky (but still has the potential :()

I'm debating whether we should use fsnotify, but that adds an extra dependency (it's currently only an indirect dep), and we'll be leaking the name of the lazy-loaded.json outside the `indexheader` package. If reviewers think it's a good idea, I can make the change.

Fixes https://github.com/grafana/mimir/issues/8875

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket_e2e_test.go
+++ b/pkg/storegateway/bucket_e2e_test.go
@@ -795,6 +795,11 @@ func TestBucketStore_PersistsLazyLoadedBlocks(t *testing.T) {
 	cfg.bucketStoreConfig.IndexHeader.EagerLoadingStartupEnabled = true
 	cfg.bucketStoreConfig.IndexHeader.LazyLoadingIdleTimeout = persistInterval * 3
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


narqo (approved):
Works for me 🔥

---

As for using `fsnotify` in the test,

> I'm debating whether we should use fsnotify.

That feels an overkill for the `RestoreLoadedBlocks` function, which is only 40 LOC (with helpers inlined).

Maybe, if that becomes annoying, we can mock the `os.ReadFile` in the tests, as a package-level global (e.g. [random example from github](https://github.com/open-telemetry/opentelemetry-go/blob/127d068751eb1c82262b65db3c3ff6c757402f40/exporters/otlp/otlplog/otlploghttp/config.go#L427)).

dimitarvdimitrov:
agree on fsnotify. Thanks for the review. Let's see how flaky this remains

PR #8270, opened 2024-06-04
Mimir query engine: report estimated peak memory consumption as a metric and in traces

charleskorn:
#### What this PR does

This PR builds upon https://github.com/grafana/mimir/pull/8230 to add reporting of the per-query estimated peak memory consumption. 

The estimate is reported as a metric and in a `query stats` event added to the query's trace span.

Note that for consistency with the metrics added in https://github.com/grafana/mimir/pull/7898,  I've used the prefix `cortex_streaming_promql_engine...` for the metric name. I'll change the name to reflect the new name for the engine in a separate PR.

#### Which issue(s) this PR fixes or relates to

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [n/a] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8909, opened 2024-08-06
Mimir query engine: split `LimitingPool`

charleskorn:
#### What this PR does

This PR is a refactoring of `LimitingPool`, which was introduced in https://github.com/grafana/mimir/pull/8270.

`LimitingPool` previously served two purposes:
* tracking the current estimated memory consumption of a query and enforcing the max memory consumption limit
* pooling slices of objects

However, there's no need to couple these two things together, and in fact, doing so makes future changes (eg. incorporating in-memory chunk bytes into the memory limit) more difficult.

This PR splits `LimitingPool` into two parts: `MemoryConsumptionTracker` and a number of global `LimitingBucketedPool`s.

Unfortunately this introduced an import cycle between the `types` and `pooling` packages, so I've opted to merge them together into the `types` package.

Performance is unchanged within the margin of error of our benchmarks.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/types/limiting_pool_test.go
```diff
@@ -0,0 +1,208 @@
...
+	p.Put(s1, tracker)
+	require.Equal(t, 8*FPointSize, tracker.CurrentEstimatedMemoryConsumptionBytes)
+	require.Equal(t, 9*FPointSize, tracker.PeakEstimatedMemoryConsumptionBytes)
+	assertRejectedQueryCount(t, reg, 0)
+
+	// Try to get a slice where the requested size would push us over the limit.
+	_, err = p.Get(4, tracker)
+	expectedError := fmt.Sprintf("the query exceeded the maximum allowed estimated amount of memory consumed by a single query (limit: %d bytes) (err-mimir-max-estimated-memory-consumption-per-query)", limit)
+	require.ErrorContains(t, err, expectedError)
+	require.Equal(t, 8*FPointSize, tracker.CurrentEstimatedMemoryConsumptionBytes)
```
jhesketh, line 111: should we reach into the pool and check the capacity of the slices here? Even though we reject the Get due to the limit, we still end up performing a Get and Put against the pool so it's possible the slice has grown.

pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -0,0 +1,142 @@
...
+	FPointSize           = uint64(unsafe.Sizeof(promql.FPoint{}))
+	HPointSize           = uint64(FPointSize * nativeHistogramSampleSizeFactor)
+	VectorSampleSize     = uint64(unsafe.Sizeof(promql.Sample{})) // This assumes each sample is a float sample, not a histogram.
+	Float64Size          = uint64(unsafe.Sizeof(float64(0)))
+	BoolSize             = uint64(unsafe.Sizeof(false))
+	HistogramPointerSize = uint64(unsafe.Sizeof((*histogram.FloatHistogram)(nil)))
+)
+
+var (
+	FPointSlicePool = NewLimitingBucketedPool(
```
jhesketh, line 32: Why maintain these globals?

We could still have a Pool separate from the MemoryConsumptionTracker that we pass around as needed.

jhesketh (commented):
Mostly concerned about the use of globals, otherwise lgtm.

We could perhaps keep some kind of LimitingPool that allows us to reach into the MemoryConsumptionTracker for chunks etc in the future.

pkg/streamingpromql/types/limiting_pool.go
charleskorn: What advantage do we gain from passing around a separate `Pool`?

pkg/streamingpromql/types/limiting_pool_test.go
charleskorn: Not sure I'm following sorry - are you suggesting we should check that the slice we took from the pool has been returned before we returned the error?

jhesketh: Yes, although I'm not entirely sure there's a trivial way to do that.
Additionally it's interesting to confirm that the capacity grew even though we didn't use it.

pkg/streamingpromql/types/limiting_pool.go
jhesketh: Not having globals :-p.

Globals have a few anti-patterns to them that we generally like to avoid.

pkg/streamingpromql/types/limiting_pool_test.go
charleskorn: >Yes, although I'm not entirely sure there's a trivial way to do that.

One way to do it would be to introduce an interface for `BucketedPool` so we can mock it out in the test and confirm that the slice is returned, but that will introduce a performance overhead for the production code which I'd prefer to avoid.

The consequences of not returning the slice to the pool are relatively minor, especially given we don't expect to hit the limit too often. 

>Additionally it's interesting to confirm that the capacity grew even though we didn't use it.

Which capacity are you referring to here?

pkg/streamingpromql/types/limiting_pool.go
charleskorn: In general, I agree with you. In this particular case, we need to have a global somewhere for the pool, otherwise we can't share the pool between queries, and using a global variable like this is fairly common in Go for this purpose.

jhesketh: If it's a common pattern that's fine by me :+1:

jhesketh (approved):

pkg/streamingpromql/types/limiting_pool_test.go
charleskorn: `Get` creates a new slice (or takes it from the pool) with a particular capacity, and `Put` puts that slice back in the pool. If something else later needs a slice of the same capacity of the discarded slice, it'll be given that slice, unless GC collects it first.

PR #8912, opened 2024-08-06
[r302] Compactor: fix a race condition between different compactor replicas causing "corrupted" bucket index (#8905)

pracucci:
Cherry pick https://github.com/grafana/mimir/pull/8905 into r302.


narqo (approved):
🔥

PR #8920, opened 2024-08-06
querytee: update compareMatrix to handle difference in series with recent samples

narqo:
#### What this PR does

We've noticed a corner case that is currently failing in `querytee`. The comparison should not fail, when 

1. "expected" and "actual" series has a different number of samples AND
2. the difference is within the `SkipRecentSamples` duration

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/querytee/response_comparator_test.go
+++ b/tools/querytee/response_comparator_test.go
@@ -98,6 +98,18 @@ func TestCompareMatrix(t *testing.T) {
 						]`),
 		},
 		{
+			name: "first series match but later series has difference in sample value",
+			expected: json.RawMessage(`[
+							{"metric":{"foo":"bar"},"values":[[1,"1"],[2,"2"]]},
+							{"metric":{"oops":"bar"},"values":[[1,"1"],[2,"2"]]}
+						]`),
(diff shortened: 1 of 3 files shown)


tools/querytee/response_comparator_test.go
```diff
@@ -1736,6 +1735,19 @@ func TestCompareSamplesResponse(t *testing.T) {
 			skipRecentSamples: time.Hour,
 			err:               errors.New("expected 1 metrics but got 2"),
 		},
+		{
+			name: "should not fail when there is a different number of samples in a series in a matrix, where some float samples are more recent than the configured skip recent samples interval",
```
pracucci, line 1739: Can we add a similar test with native histograms, please?

pracucci (approved):
I think the change is more complicated than it could have been (I think it was enough to change the lines like `expectedFloatSamplesCount := len(expectedMetric.Values)` into `expectedFloatSamplesCount := countSkippingRecentSamples(expectedMetric.Values)`), but I can't see any issue with the new code neither, so LGTM.

PR #8921, opened 2024-08-07
Mimir query engine: improve benchmark setup time by ~10% by reusing slices for histograms

charleskorn:
#### What this PR does

This PR reduces the time taken to load data into a test ingester during the MQE benchmarks.

It does this by reusing the same slices for spans and deltas for all samples, rather than allocating a new one for each sample.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


jhesketh (approved):
if we ever want more interesting data in the benchmarks (rather than identical histograms) this might need to change. However, it's good for now :+1:

PR #8923, opened 2024-08-07
MQE: Use kahan in sum aggregation

jhesketh:
Implements the improvements from
https://github.com/prometheus/prometheus/pull/14074 https://github.com/prometheus/prometheus/pull/14362

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/operators/aggregation.go
```diff
@@ -96,6 +97,7 @@ type group struct {
 
 	// Sum, presence, and histograms for each step.
 	floatSums           []float64
+	floatMeans          []float64 // Mean, or "compensating value" for Kahan summation.
```
charleskorn, line 100: Would `floatCompensatingValues` or something like that be a clearer name?

pkg/streamingpromql/testdata/upstream/aggregators.test
```diff
@@ -602,8 +602,8 @@ load 10s
 	data{test="nan",group="2",point="b"} NaN
 
 # Unsupported by streaming engine.
```
charleskorn, line 604: We can remove this comment now 🙂

seizethedave, line 604: is it now supported?

pkg/streamingpromql/operators/aggregation.go
jhesketh: Perhaps. We may use this for mean's in the future, but will rename for now (can always go back)

charleskorn (approved):

PR #8924, opened 2024-08-07
Mimir query engine: add native histogram custom bucket-related annotations

charleskorn:
#### What this PR does

This PR adds support for the following warnings in `rate` and `sum` in MQE:

* histogram samples with a mixture of exponential and custom buckets at the same timestamp
* histogram samples with incompatible custom buckets at the same timestamp

This uncovered two issues in Prometheus' engine: 

* https://github.com/prometheus/prometheus/pull/14609
* https://github.com/prometheus/prometheus/pull/14611

I've opted to skip the affected tests that confirm our test cases match Prometheus' engine to unblock this PR.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/functions/rate.go
```diff
@@ -95,6 +100,10 @@ func histogramRate(histogramBuffer *types.HPointRingBuffer, step types.RangeVect
 				emitAnnotation(annotations.NewNativeHistogramNotCounterWarning)
 			}
 
+			if p.H.UsesCustomBuckets() != usingCustomBuckets {
```
jhesketh, line 103: Why not do this check first thing in this for? It could potentially avoid doing an unnecessary CopyToSchema

pkg/streamingpromql/testdata/ours/native_histograms.test
```diff
@@ -145,3 +144,32 @@ load 1m
...
+# Test mixing exponential and custom buckets.
+load 6m
+  metric{series="exponential"}         {{sum:4 count:3 buckets:[1 2 1]}}  _                                                                 {{sum:4 count:3 buckets:[1 2 1]}}
+  metric{series="other-exponential"}   {{sum:3 count:2 buckets:[1 1 1]}}  _                                                                 {{sum:3 count:2 buckets:[1 1 1]}}
+  metric{series="custom"}              _                                  {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}}     {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}}
+  metric{series="other-custom"}        _                                  {{schema:-53 sum:15 count:2 custom_values:[5 10] buckets:[0 2]}}  {{schema:-53 sum:15 count:2 custom_values:[5 10] buckets:[0 2]}}
+
+# T=0: only exponential
+# T=6: only custom
+# T=12: mixed, should be ignored and emit an warning
```
jhesketh, line 159: would be good to see some valid points after an ignored point due to the way point count is decremented.

pkg/streamingpromql/operators/aggregation.go
```diff
@@ -276,14 +288,20 @@ func (a *Aggregation) accumulateUntilGroupComplete(ctx context.Context, g *group
...
+
+		a.currentSeriesIndex++
 	}
 	return nil
 }
 
+// Sentinel value used to indicate a sample has seen an invalid combination of histograms and should be ignored.
+//
+// Invalid combinations include exponential and custom buckets, and histograms with incompatible custom buckets.
+var invalidCombinationOfHistograms = &histogram.FloatHistogram{}
```
jhesketh, line 303: Why use a global here instead of on the struct? I don't think the memory overhead would be significant

pkg/streamingpromql/operators/aggregation.go
```diff
@@ -411,10 +429,20 @@ func (a *Aggregation) accumulateSeriesIntoGroup(s types.InstantVectorSeriesData,
...
 
-		if seriesGroup.histogramSums[idx] != nil {
+		if seriesGroup.histogramSums[idx] == invalidCombinationOfHistograms {
+			// We've already seen an invalid combination of histograms at this timestamp. Ignore this point.
+			continue
+		} else if seriesGroup.histogramSums[idx] != nil {
 			seriesGroup.histogramSums[idx], err = seriesGroup.histogramSums[idx].Add(p.H)
 			if err != nil {
-				return err
+				// We've seen an invalid combination of histograms. Make sure we don't emit a sample at this timestamp.
```
jhesketh, line 438: (nit)
```suggestion
				// Unable to add histograms together (likely due to an invalid combination). Make sure we don't emit a sample at this timestamp.
```

pkg/streamingpromql/operators/aggregation.go
```diff
@@ -411,10 +429,20 @@ func (a *Aggregation) accumulateSeriesIntoGroup(s types.InstantVectorSeriesData,
...
+				seriesGroup.histogramSums[idx] = invalidCombinationOfHistograms
+				seriesGroup.histogramPointCount--
+
+				if err := functions.NativeHistogramErrorToAnnotation(err, a.emitAnnotationFunc); err != nil {
+					// We couldn't convert the error to an annotation. Give up.
```
jhesketh, line 443: (nit)
```suggestion
					// Unknown error, or we couldn't convert the error to an annotation. Give up.
```

pkg/streamingpromql/operators/aggregation.go
charleskorn: I would make this a const if that was valid in Go.

It doesn't make sense to me to add it to the struct, as there's nothing about it that's specific to this aggregation instance.

pkg/streamingpromql/functions/rate.go
charleskorn: Changed in 1558d29

pkg/streamingpromql/operators/aggregation.go
charleskorn: Changed in 20a3fc0

pkg/streamingpromql/testdata/ours/native_histograms.test
charleskorn: Done: 20a3fc0

jhesketh (approved):
LGTM!

PR #8935, opened 2024-08-08
ruler: map invalid org-id errors to status code `400`

ortuman:
#### What this PR does

Recently, we were paged in production because the ruler maps invalid org-id errors to status code `500`.

The PR fixes this behaviour by returning `400` code instead. 

Fixes n/a


pkg/ruler/api.go
```diff
@@ -160,7 +160,7 @@ func (a *API) PrometheusRules(w http.ResponseWriter, req *http.Request) {
 	userID, err := tenant.TenantID(ctx)
 	if err != nil || userID == "" {
 		level.Error(logger).Log("msg", "error extracting org id from context", "err", err)
-		respondServerError(logger, w, "no valid org id found")
```
gotjosh, line 163: Can we add a test for all the routes? 

It looks like the same will happen on all routes where errors failing to extract the `org_id` in `parseRequest` will cause a server error, looks like we'll have to use `errors.Is` to identify the kind of error it is.

gotjosh (commented):
Great catch - please take a look at my comments.

pkg/ruler/api.go
ortuman: as suggested, i reworked the PR to consider all routes, and added the new `TestAPIRoutesCorrectlyHandleInvalidOrgID` unit test.

ready for review again.

gotjosh (approved):
LGTM

PR #8937, opened 2024-08-08
api: clean up outdated chunk from earlier block-builder prototyping

narqo:
This one cleans up an outdated piece of code from the earlier prototype. Block-builder doesn't create its own hash-ring (for now). We may revisit that in the future.


narqo:
There is nothing to review here, so I'll merge it to move on.

PR #8944, opened 2024-08-08
Code cleanup for `pusherConsumer`

gotjosh:
#### What this PR does

This small refactor cleans up a bit the code within pusherConsumer to address some dependency inversion of some of the components.

Some of the things I've done are:

- Move functions around to comply with Go code's standards e.g. Functions that are part of the public interface go after the struct definition
- Avoid passing individual metrics around and instead have a single metrics struct that is passed around to the different structs
- Simplify the `consumerFactory` definition  to push the construction of the `pusherCloser` to itself.

There are other things I want to do such as: Remove the "multitenant pusher" abstraction but I think this is a enough for a first PR.

NB: This does not the change the behaviour of the code at all.

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -46,9 +46,6 @@ type pusherConsumer struct {
 	metrics                  *pusherConsumerMetrics
 	logger                   log.Logger
 
-	shards    int
-	batchSize int
-
 	pusher PusherCloser
 }
 
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -28,16 +28,6 @@ type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
-type pusherConsumerPrototype struct {
-	processingTimeSeconds prometheus.Observer
-	clientErrRequests     prometheus.Counter
-	serverErrRequests     prometheus.Counter
-	totalRequests         prometheus.Counter
-
-	fallbackClientErrSampler *util_log.Sampler // Fallback log message sampler client errors that are not sampled yet.
-	logger                   log.Logger
-}
-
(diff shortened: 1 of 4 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -267,11 +287,6 @@ func (c multiTenantPusher) Close() []error {
 	return errs
 }
 
-type shardedPush struct {
```
gotjosh, line 270: This was unused.

pkg/storage/ingest/reader.go
```diff
@@ -339,9 +337,9 @@ func (r *PartitionReader) processNextFetchesUntilLagHonored(ctx context.Context,
 	} else {
 		fetcher = r
 	}
-	//defer func() {
+	// defer func() {
 	//	r.setPollingStartOffset(r.consumedOffsetWatcher.LastConsumedOffset())
-	//}()
+	// }()
```
dimitarvdimitrov, line 342:
```suggestion
```

this was long left over

pkg/storage/ingest/pusher.go
```diff
@@ -267,11 +287,6 @@ func (c multiTenantPusher) Close() []error {
...
-	mimirpb.PreallocTimeseries
-	context.Context
-}
-
```
dimitarvdimitrov, line 274: that's unused now, but there was a TODO referring to it. I'm fine removing it now, but FYI in case that comment is then confusing

https://github.com/grafana/mimir/blob/37c2780a1197d1769d07f2bbe8644ce41a18bda5/pkg/storage/ingest/pusher.go#L368

dimitarvdimitrov (approved):
LGTM, thanks

pracucci (approved):

PR #8946, opened 2024-08-09
doc-generator: include default values for config options decorated with `nocli`, fix formatting of slices, and export Markdown generation package

charleskorn:
#### What this PR does

This PR fixes two issues in `tools/doc-generator`:

* config options that have `nocli` set do not have their default values included in the generated text
* config options that take slices do not include details of the slice elements in the generated text

It also exports the Markdown generation code so that it can be used in other repositories (eg. our enterprise product).

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/doc-generator/write/writer.go
+++ b/tools/doc-generator/write/writer.go
@@ -3,7 +3,7 @@
 // Provenance-includes-license: Apache-2.0
 // Provenance-includes-copyright: The Cortex Authors.
 
-package main
+package write
 
 import (
 	"fmt"
@@ -18,8 +18,14 @@ import (
 	"github.com/grafana/mimir/tools/doc-generator/parse"
 )
 
+const (
+	maxLineWidth = 80
+	tabWidth     = 2
+)
+
 type specWriter struct {
-	out strings.Builder
+	out                strings.Builder
+	modifyDescriptions func(string) string
 }
 
 func (w *specWriter) writeConfigBlock(b *parse.ConfigBlock, indent int) {
@@ -43,7 +49,7 @@ func (w *specWriter) writeConfigEntry(e *parse.ConfigEntry, indent int) {
 		// so here we've just to write down the reference without re-iterating on it.
 		if e.Root {
 			// Description
-			w.writeComment(e.BlockDesc, indent, 0)
+			w.writeComment(w.modifyDescriptions(e.BlockDesc), indent, 0)
 			if e.Block.FlagsPrefix != "" {
 				w.writeComment(fmt.Sprintf("The CLI flags prefix for this block configuration is: %s", e.Block.FlagsPrefix), indent, 0)
 			}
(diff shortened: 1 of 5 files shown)


tools/doc-generator/write/writer.go
```diff
@@ -82,6 +88,18 @@ func (w *specWriter) writeConfigEntry(e *parse.ConfigEntry, indent int) {
 			w.out.WriteString(pad(indent) + "[" + e.Name + ": <" + e.FieldType + "> | default = " + fieldDefault + "]\n")
 		}
 	}
+
+	if e.Kind == parse.KindSlice {
```
jhesketh, line 92:
```suggestion
	} else if e.Kind == parse.KindSlice {
```

jhesketh (commented):
I know there is no existing tests, but it would be nice to see some. Especially if we're going to start exporting this and using it elsewhere.

charleskorn:
>I know there is no existing tests, but it would be nice to see some. Especially if we're going to start exporting this and using it elsewhere.

I agree it's not ideal, but we have been using the output of the doc generation process (https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/configuration-parameters/index.md) as a test of sorts.

At present, we've copied the same code into our enterprise product as well, and the two have diverged, so we currently have two slightly different, equally untested versions of the same thing in two places.

Given we rarely change this code, using the produced docs as a test is sufficient for me, but open to dissenting opinions 🙂

tools/doc-generator/write/writer.go
charleskorn: Addressed in 6d7d1fe

jhesketh:
> > I know there is no existing tests, but it would be nice to see some. Especially if we're going to start exporting this and using it elsewhere.
> 
> I agree it's not ideal, but we have been using the output of the doc generation process (https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/configuration-parameters/index.md) as a test of sorts.
> 
> At present, we've copied the same code into our enterprise product as well, and the two have diverged, so we currently have two slightly different, equally untested versions of the same thing in two places.
> 
> Given we rarely change this code, using the produced docs as a test is sufficient for me, but open to dissenting opinions 🙂

I feel like it wouldn't be too difficult to add a test for a few slice formats with default values. That would at least get us started on some coverage and it can grow over time. Otherwise we'll keep ignoring this for the reasons you put forward (which I don't entirely disagree with, but feel like overcoming them isn't too hard if done incrementally).

Happy not to block on this if it's more effort than I think it is :-).

charleskorn:
I've added a basic test in c267491.

jhesketh (approved):

jhesketh:
> I've added a basic test in [c267491](https://github.com/grafana/mimir/commit/c2674919c4d33d164ec5e88474e07aea6fa7c7c8).

Thanks!

PR #8947, opened 2024-08-09
query-tee: log all samples for matrix series when a mismatch occurs

charleskorn:
#### What this PR does

This PR adds more information to the error messages produced by query-tee when comparing series from a matrix. 

Specifically, it will now include all samples for a series when a mismatch occurs.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/querytee/response_comparator.go
+++ b/tools/querytee/response_comparator.go
@@ -179,7 +179,7 @@ func compareMatrix(expectedRaw, actualRaw json.RawMessage, opts SampleComparison
 
 		err := compareMatrixSamples(expectedMetric, actualMetric, opts)
 		if err != nil {
-			return err
+			return fmt.Errorf("%w\nFull expected result:\n%v\n\nFull actual result:\n%v", err, expectedMetric, actualMetric)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


tools/querytee/response_comparator.go
```diff
@@ -179,7 +179,7 @@ func compareMatrix(expectedRaw, actualRaw json.RawMessage, opts SampleComparison
 
 		err := compareMatrixSamples(expectedMetric, actualMetric, opts)
 		if err != nil {
-			return err
+			return fmt.Errorf("%w\nFull expected result:\n%v\n\nFull actual result:\n%v", err, expectedMetric, actualMetric)
```
jhesketh, line 182: (nit) why "full"? Just Expected result would be sufficient imo.

jhesketh (commented):
Would be nice to see a test with the full matrix response

tools/querytee/response_comparator.go
charleskorn: I was trying to disambiguate it from the rest of the error message, which will refer to a single point.

Does 9875b9d look better to you?

charleskorn:
> Would be nice to see a test with the full matrix response

Done: d94d1a5

PR #8948, opened 2024-08-09
Dashboards: add Kafka end-to-end latency outliers panel in the "Mimir / Writes" dashboard

pracucci:
#### What this PR does

Yesterday I spent few hours investigating high end-to-end latency in a dev cell and it took me a while to figure out that the high reported latency was actually caused by 1 outlier ingester. In this PR I propose to add a panel to show outliers. It's not perfect, but it gives an hint. I also tried to compute outliers using a more complex approach (using stddev and z-score but I couldn't get better results, so I'm proposing the simple approach).

Screenshot 2024-08-09 at 12 50 58

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
SGTM

PR #8953, opened 2024-08-09
kafka replay speed: use concurrent fetchers in tests

dimitarvdimitrov:
#### What this PR does

1.  It uses concurrent fetchers in tests. This adds coverage to the concurrent fetchers. Because of that I also found two bugs
    * we no longer ignore the first record with offset `0`
    * when we reach the end of a partition we back off and try fetching again - currently with exponential backoff of 10ms, up to 1s

    The latter actually unblocks using this code for regular fetching too (see 3.)

2. Improve maxBytes estimations - now we make sure to not overflow int32 (it happened during testing) and generally keep the logic closer together in `fetchWant`
3. Use concurrent fetchers for regular consuming too. This required changes to handle the case where we're fetching after the last available offset, but the changes were minor. 
   * This also unblocks undoing some changes compared to `main`: the changes in `processNextFetches` and in `processNextFetchesUntilTargetOrMaxLagHonored`
   * When testing I realized that the metrics in `newPusherConsumer` are initialized each time causing panics, so I included this fix too.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -494,8 +494,10 @@ func (r *PartitionReader) consumeFetches(ctx context.Context, fetches kgo.Fetche
 		consumeCtx := context.WithoutCancel(ctx)
 		err := consumer.consume(consumeCtx, records)
 		if err == nil {
-			_ = consumer.Close(consumeCtx)
-			break
+			err = consumer.Close(consumeCtx)
+			if err == nil {
+				break
+			}
 		}
 		level.Error(r.logger).Log(
 			"msg", "encountered error while ingesting data from Kafka; should retry",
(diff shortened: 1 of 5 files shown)


pkg/storage/ingest/reader.go
```diff
@@ -842,41 +864,43 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
...
 		Partitions: []kmsg.FetchRequestTopicPartition{{
-			Partition:          r.partitionID,
-			FetchOffset:        w.startOffset,
-			LastFetchedEpoch:   -1,
+			Partition:   r.partitionID,
+			FetchOffset: w.startOffset,
+			// CurrentLeaderEpoch: -1 means we don't know. Kafka brokers are ok with that.
+			// It does mean that we might end up fetching from an out-of-sync replica.
+			// If we provide this the broker would check if we have up-to-date data.
 			CurrentLeaderEpoch: -1,
```
dimitarvdimitrov, line 882: this is probably the shoddiest of this implementation. Do we want to deal with it?

pkg/storage/ingest/reader.go
```diff
@@ -246,14 +259,14 @@ func (r *PartitionReader) processFetches(ctx context.Context, fetches kgo.Fetche
 // processNextFetchesUntilTargetOrMaxLagHonored process records from Kafka until at least the maxLag is honored.
 // This function does a best-effort to get lag below targetLag, but it's not guaranteed that it will be
 // reached once this function successfully returns (only maxLag is guaranteed).
-func (r *PartitionReader) processNextFetchesUntilTargetOrMaxLagHonored(ctx context.Context, startOffset int64, targetLag, maxLag time.Duration) error {
+func (r *PartitionReader) processNextFetchesUntilTargetOrMaxLagHonored(ctx context.Context, targetLag, maxLag time.Duration) error {
```
dimitarvdimitrov, line 262: this is a diff to the base branch, but it restores the version from `main`

pkg/storage/ingest/reader.go
dimitarvdimitrov: update: I found some implementation notes in franz-go 

https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kgo/topics_and_partitions.go#L530-L537

and the list of all possible errors. I think we should handle all of them as best as we can. Currently we handle only the most common one.

https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kmsg/generated.go#L4705-L4745

looks like a lot of code we'll have to write and test :(

dimitarvdimitrov:
closing in favour of https://github.com/grafana/mimir/pull/8967 and https://github.com/grafana/mimir/pull/8968

PR #8956, opened 2024-08-10
Add support for ingester downscaling

jhalterman:
#### What this PR does

This PR adds support for downscaling ingesters via a new prepare-instance-ring-downscale endpoint. This endpoint sets or clears the read-only status of an ingester in the ring, and will be called by the rollout-operator.

Part of https://github.com/grafana/mimir-squad/issues/1797

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/distributor/query.go
```diff
@@ -137,11 +137,10 @@ func (d *Distributor) getIngesterReplicationSetsForQuery(ctx context.Context) ([
 	// Lookup ingesters ring because ingest storage is disabled.
 	shardSize := d.limits.IngestionTenantShardSize(userID)
 	r := d.ingestersRing
-
-	// If tenant uses shuffle sharding, we should only query ingesters which are part of the tenant's subring.
-	if lookbackPeriod := d.cfg.ShuffleShardingLookbackPeriod; shardSize > 0 && lookbackPeriod > 0 {
-		r = r.ShuffleShardWithLookback(userID, shardSize, lookbackPeriod, time.Now())
+	if shardSize == 0 {
```
pstibrany, line 140: This change is not necessary. We could keep querying entire ring, if shuffle-sharding is not enabled -- entire ring contains all read-only and read-write ingesters, which is fine.

Also setting `shardSize` to `r.InstancesCount()` is not necessary -- `ShuffleShardWithLookback` handles `shardSize == 0` already.

pkg/ingester/downscale.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"net/http"
+
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/services"
+
+	"github.com/grafana/mimir/pkg/util"
+)
+
+// PrepareInstanceRingDownscaleHandler prepares the ingester ring entry for downscaling. It can mark ingester as READONLY or back to ACTIVE.
```
pstibrany, line 12: This entire comment is referring to `READONLY` ring state, but we moved away from that design, and now only mark ingester as read-only, without changing its state.

pkg/ingester/downscale.go
```diff
@@ -0,0 +1,65 @@
...
+//
+// Following methods are supported:
+//
+//   - GET
+//     Returns timestamp when instance was switched to READONLY state, or 0, if partition is not in READONLY state.
```
pstibrany, line 17: `PrepareInstanceRingDownscaleHandler` doesn't deal with partitions. (See also comment about `DELETE` method below)

pkg/ingester/downscale.go
```diff
@@ -0,0 +1,65 @@
...
+//
+//   - POST
+//     Switches the instance to READONLY state (if not yet), and returns the timestamp when the switch to
+//     READONLY state happened.
+//
+//   - DELETE
+//     Sets partition back from READONLY to ACTIVE state.
+func (i *Ingester) PrepareInstanceRingDownscaleHandler(w http.ResponseWriter, r *http.Request) {
```
pstibrany, line 25: Shall we add a unit test for this handler?

pkg/util/atomicfs/fsync.go
```diff
@@ -68,22 +68,7 @@ func (a *File) Close() error {
...
-	// > needed.
-	dir, err := os.Open(filepath.Dir(a.finalPath))
-	if err != nil {
-		return err
-	}
-
-	merr.Add(dir.Sync())
-	merr.Add(dir.Close())
-	return merr.Err()
+	return syncDir(filepath.Dir(a.finalPath))
```
pstibrany, line 71: Changes in `atomicfs` and `shutdownmarker` packages come from previous design, when we planned to store "downscale marker" on the disk. However we have moved on from that design, and now changes in these two packages are unrelated to this PR.

I suggest we remove changes to `atomicfs` and `shutdownmarker` packages from this PR.

pstibrany: Done in https://github.com/grafana/mimir/pull/8956/commits/0d64c249f3461cce83c8f3a730fdbdb0abe1ef8e.

pkg/ingester/downscale.go
pstibrany: Fixed in https://github.com/grafana/mimir/pull/8956/commits/96afdc1c3d2f62572d1e7f296b0f96552980a9df.

pkg/ingester/downscale.go
pstibrany: Updated in https://github.com/grafana/mimir/pull/8956/commits/96afdc1c3d2f62572d1e7f296b0f96552980a9df.

pkg/ingester/downscale.go
pstibrany: I started in https://github.com/grafana/mimir/pull/8956/commits/9bb33466be1ee28abebd65fcd91f4c3cfb1fd328, but it needs more work.

jhalterman: Added some more test cases

pstibrany (approved):
Thank you! I've pushed some comments and test changes, but nothing substantial, so I'm approving this.

PR #8962, opened 2024-08-12
Mimir query engine: add support for `offset` in instant and range vector selectors

charleskorn:
#### What this PR does

This PR adds support for the `offset` modifier in instant and range vector selectors in Mimir's query engine.

For example, `metric{foo="bar"} offset 2m` and `rate(metric{foo="bar"}[2m] offset 1h)` are now supported.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/engine_test.go
```diff
@@ -594,6 +603,94 @@ func TestRangeVectorSelectors(t *testing.T) {
...
+					},
+					{
+						Metric: labels.FromStrings("__name__", "some_metric", "env", "2"),
+						Floats: []promql.FPoint{
+							{T: timestamp.FromTime(baseT.Add(time.Minute)), F: 2},
+							{T: timestamp.FromTime(baseT.Add(2 * time.Minute)), F: 4},
+						},
+					},
+				},
+			},
```
jhesketh, line 692: how about a test for when offset takes us to before `ts`

pkg/streamingpromql/testdata/ours/selectors.test
```diff
@@ -19,3 +19,36 @@ eval range from 0 to 4m step 1m some_metric
...
+# Output T=0m to T=3m should be looking for sample at one minute in the future
+# T=5m should be looking for sample at T=4m, and then fallback to lookback to previous point within 5m lookback window
+eval range from 0 to 4m step 1m some_metric offset -1m
+  some_metric{env="prod", cluster="eu"} 1 2 3 4 4
+  some_metric{env="prod", cluster="us"} 2 4 6 8 8
+  some_metric{env="test", cluster="eu"} 3 6 9 12 12
+  some_metric{env="test", cluster="us"} 4 8 12 16 16
+
+# Range query with @ modifier.
+eval range from 0 to 10m step 1m some_metric @ 2m
```
jhesketh, line 43: what about a testing
`eval range from 0 to 10m step 1m some_metric @ 20m`
and
`eval range from 0 to 10m step 1m some_metric @ -5m`

jhesketh (commented):
lgtm, just some suggestions for extra tests

pkg/streamingpromql/engine_test.go
charleskorn: Doesn't https://github.com/grafana/mimir/pull/8962/files/795c3a8328dc8fa3ac2c0282342ac8c953c7c3d0#diff-df3817badf68588818b73b443096f0edd778d0019a2bc135e6a7f6eeb218946cR606-R608 cover this? Not sure I'm following the suggestion.

pkg/streamingpromql/testdata/ours/selectors.test
charleskorn: Added in 329094c.

tacole02 (approved):

pkg/streamingpromql/engine_test.go
jhesketh: I mean where the offset takes us to before `baseT` (ie < T = 0). Obviously not a realistic scenario, but we should check that the engines handle the edge case the same. Perhaps also when T is at max and offset looks forward too.

charleskorn: Right, makes sense. Added in d820aed.

jhesketh (approved):

PR #8968, opened 2024-08-12
kafka replay speed: use concurrent fetchers for regular fetching too

dimitarvdimitrov:
#### What this PR does

It is based on #8967

1.  It uses concurrent fetchers in tests. This adds coverage to the concurrent fetchers. Because of that I also found two bugs
    * we no longer ignore the first record with offset `0`
    * when we reach the end of a partition we back off and try fetching again - currently with exponential backoff of 10ms, up to 1s

    The latter actually unblocks using this code for regular fetching too (see 2.)

3. Use concurrent fetchers for regular consuming too. This required changes to handle the case where we're fetching after the last available offset, but the changes were minor. 
   * This also unblocks undoing some changes compared to `main`: the changes in `processNextFetches` and in `processNextFetchesUntilTargetOrMaxLagHonored`
   * When testing I realized that the metrics in `newPusherConsumer` are initialized each time causing panics, so I included this fix too.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -891,27 +891,28 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
 	}
 }
 
-func (r *concurrentFetchers) fetchSingle(ctx context.Context, w fetchWant) fetchResult {
+func (r *concurrentFetchers) fetchSingle(ctx context.Context, fw fetchWant) fetchResult {
 	req := kmsg.NewFetchRequest()
 	req.MinBytes = 1
 	req.Version = 13
 	req.MaxWaitMillis = 10000
-	req.MaxBytes = w.MaxBytes()
-	req.Topics = []kmsg.FetchRequestTopic{{
-		Topic:   r.topicName,
-		TopicID: r.topicID,
-		Partitions: []kmsg.FetchRequestTopicPartition{{
-			Partition:   r.partitionID,
-			FetchOffset: w.startOffset,
-			// CurrentLeaderEpoch: -1 means we don't know. Kafka brokers are ok with that.
-			// It does mean that we might end up fetching from an out-of-sync replica.
-			// If we provide this the broker would check if we have up-to-date data.
-			CurrentLeaderEpoch: -1,
-			LastFetchedEpoch:   -1,
-			LogStartOffset:     -1, // this is broker-follower only field. Set it to -1 to not use it.
-			PartitionMaxBytes:  req.MaxBytes,
-		}},
-	}}
+	req.MaxBytes = fw.MaxBytes()
+
+	reqTopic := kmsg.NewFetchRequestTopic()
+	reqTopic.Topic = r.topicName
+	reqTopic.TopicID = r.topicID
+
+	// Using NewFetchRequestTopicPartition gives us a partition with the default values.
+	// One of them is CurrentLeaderEpoch: -1, which means we don't know the leader epoch and Kafka brokers are ok with that.
+	// It does mean that we might end up fetching from an out-of-sync replica.
+	// If we provide this the broker would check if we have up-to-date data.
+	reqPartition := kmsg.NewFetchRequestTopicPartition()
+	reqPartition.Partition = r.partitionID
+	reqPartition.FetchOffset = fw.startOffset
+	reqPartition.PartitionMaxBytes = req.MaxBytes
+
+	reqTopic.Partitions = append(reqTopic.Partitions, reqPartition)
+	req.Topics = append(req.Topics, reqTopic)
 
 	resp, err := req.RequestWith(ctx, r.client)
 	if err != nil {
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -908,7 +908,7 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
 	}
 }
 
-func (r *concurrentFetchers) fetchSingle(ctx context.Context, w fetchWant, _ log.Logger) (_ kgo.FetchPartition, fetchedBytes int) {
+func (r *concurrentFetchers) fetchSingle(ctx context.Context, w fetchWant) fetchResult {
 	req := kmsg.NewFetchRequest()
 	req.MinBytes = 1
 	req.Version = 13
@@ -930,18 +930,18 @@ func (r *concurrentFetchers) fetchSingle(ctx context.Context, w fetchWant, _ log
 	resp, err := req.RequestWith(ctx, r.client)
 	if err != nil {
 		if errors.Is(err, context.Canceled) {
-			return kgo.FetchPartition{}, 0
+			return fetchResult{kgo.FetchPartition{}, 0}
 		}
-		return kgo.FetchPartition{
+		return fetchResult{kgo.FetchPartition{
 			Err: fmt.Errorf("fetching from kafka: %w", err),
-		}, 0
+		}, 0}
 	}
 	rawPartitionResp := resp.Topics[0].Partitions[0]
-	fetchedBytes = len(rawPartitionResp.RecordBatches)          // TODO dimitarvdimitrov make this conditional on the kafka backend - for WS we use uncompressed bytes (sumRecordLengths), for kafka we use the size of the response (rawPartitionResp.RecordBatches)
-	r.metrics.fetchesCompressedBytes.Add(float64(fetchedBytes)) // This doesn't include overhead in the response, but that should be small.
+	// TODO dimitarvdimitrov make this conditional on the kafka backend - for WS we use uncompressed bytes (sumRecordLengths), for kafka we use the size of the response (rawPartitionResp.RecordBatches)
+	r.metrics.fetchesCompressedBytes.Add(float64(len(rawPartitionResp.RecordBatches))) // This doesn't include overhead in the response, but that should be small.
 	partition := processRespPartition(&rawPartitionResp, r.topicName)
 	partition.EachRecord(r.tracer.OnFetchRecordBuffered) // TODO dimitarvdimitrov we might end up buffering the same record multiple times - what happens then?
-	return partition, sumRecordLengths(partition.Records)
+	return fetchResult{partition, sumRecordLengths(partition.Records)}
 }
 
 func sumRecordLengths(records []*kgo.Record) (sum int) {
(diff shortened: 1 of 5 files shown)


pkg/storage/ingest/reader.go
```diff
@@ -721,27 +719,8 @@ func (r *PartitionReader) waitReadConsistency(ctx context.Context, withOffset bo
...
-					if record.Offset > r.consumedOffsetWatcher.LastConsumedOffset() {
-						afterConsumed = i
-						break
-					}
-				}
-				f[fIdx].Topics[tIdx].Partitions[pIdx].Records = partition.Records[afterConsumed:]
-			}
-		}
-	}
-	return f
```
dimitarvdimitrov, line 744: this is no longer necessary since we're using the same fetcher implementation for startup fetching and regular fetching. Previously we needed a way to synchronise the two - make sure the latter picked up where the former left off.

pkg/storage/ingest/reader.go
```diff
@@ -900,30 +880,33 @@ func (r *concurrentFetchers) fetchSingle(ctx context.Context, w fetchWant, _ log
...
-			FetchOffset:        w.startOffset,
-			LastFetchedEpoch:   -1,
+			Partition:   r.partitionID,
+			FetchOffset: w.startOffset,
+			// CurrentLeaderEpoch: -1 means we don't know. Kafka brokers are ok with that.
+			// It does mean that we might end up fetching from an out-of-sync replica.
+			// If we provide this the broker would check if we have up-to-date data.
 			CurrentLeaderEpoch: -1,
-			LogStartOffset:     -1,
+			LastFetchedEpoch:   -1,
```
dimitarvdimitrov, line 889: this is probably the shoddiest of this implementation. Do we want to deal with this problem? I couldn't find any docs on how to handle this, but I found some implementation notes in franz-go and the list of all possible errors

https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kgo/topics_and_partitions.go#L530-L537

I think we should handle all of them as best as we can. Currently we handle only the most common one.

https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kmsg/generated.go#L4705-L4745

looks like a lot of code we'll have to write and test :( I'll work on this next

pkg/storage/ingest/reader.go
```diff
@@ -73,7 +73,10 @@ type PartitionReader struct {
 	partitionID   int32
 	consumerGroup string
 
-	client *kgo.Client
+	client  *kgo.Client
+	fetcher interface {
+		pollFetches(context.Context) kgo.Fetches
+	}
```
gotjosh, line 79: [nit] is this a pattern we commonly use? Would rather see the interface as a separate type.

pkg/storage/ingest/reader.go
```diff
@@ -118,8 +121,9 @@ func (noopPusherCloser) Close() []error {
 }
 
 func NewPartitionReaderForPusher(kafkaCfg KafkaConfig, partitionID int32, instanceID string, pusher Pusher, logger log.Logger, reg prometheus.Registerer) (*PartitionReader, error) {
+	metrics := newPusherConsumerMetrics(reg)
```
gotjosh, line 124: What's the reasoning for this change? I'd rather let the constructor of `newPusherConsumer` create the dependency it needs. I don't see `metrics` being used anywhere else.

pkg/storage/ingest/reader.go
gotjosh: After reading more of the code I found this is a bit simpler to understand:

```
diff --git a/pkg/storage/ingest/reader.go b/pkg/storage/ingest/reader.go
index d8f826118..0c63ac245 100644
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -65,6 +65,8 @@ type recordConsumer interface {
 	consume(context.Context, []record) error
 }
 
+type pollingFetcherFunc func(context.Context) kgo.Fetches
+
 type PartitionReader struct {
 	services.Service
 	dependencies *services.Manager
@@ -73,10 +75,8 @@ type PartitionReader struct {
 	partitionID   int32
 	consumerGroup string
 
-	client  *kgo.Client
-	fetcher interface {
-		pollFetches(context.Context) kgo.Fetches
-	}
+	client      *kgo.Client
+	pollingFunc pollingFetcherFunc
 
 	newConsumer consumerFactory
 	metrics     readerMetrics
@@ -143,7 +143,6 @@ func newPartitionReader(kafkaCfg KafkaConfig, partitionID int32, instanceID stri
 		logger:                log.With(logger, "partition", partitionID),
 		reg:                   reg,
 	}
-
 	r.Service = services.NewBasicService(r.start, r.run, r.stop)
 	return r, nil
 }
@@ -187,12 +186,19 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 	}
 
 	if r.kafkaCfg.ReplayConcurrency > 1 {
-		r.fetcher, err = newConcurrentFetchers(ctx, r.client, r.logger, r.kafkaCfg.Topic, r.partitionID, startOffset, r.kafkaCfg.ReplayConcurrency, r.kafkaCfg.RecordsPerFetch, &r.metrics)
+		fetcher, err := newConcurrentFetchers(ctx, r.client, r.logger, r.kafkaCfg.Topic, r.partitionID, startOffset, r.kafkaCfg.ReplayConcurrency, r.kafkaCfg.RecordsPerFetch, &r.metrics)
 		if err != nil {
 			return errors.Wrap(err, "creating concurrent fetchers")
 		}
+
+		r.pollingFunc = fetcher.pollFetches
 	} else {
-		r.fetcher = r
+		r.pollingFunc = func(ctx context.Context) kgo.Fetches {
+			defer func(start time.Time) {
+				r.metrics.fetchWaitDuration.Observe(time.Since(start).Seconds())
+			}(time.Now())
+			return r.client.PollFetches(ctx)
+		}
 	}
 
 	// Enforce the max consumer lag (if enabled).
@@ -242,7 +248,7 @@ func (r *PartitionReader) run(ctx context.Context) error {
 }
 
 func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver prometheus.Observer) error {
-	fetches := r.fetcher.pollFetches(ctx)
+	fetches := r.pollingFunc(ctx)
 	r.recordFetchesMetrics(fetches, delayObserver)
 	r.logFetchErrors(fetches)
 	fetches = filterOutErrFetches(fetches)
@@ -716,13 +722,6 @@ func (r *PartitionReader) waitReadConsistency(ctx context.Context, withOffset bo
 	return err
 }
 
-func (r *PartitionReader) pollFetches(ctx context.Context) (result kgo.Fetches) {
-	defer func(start time.Time) {
-		r.metrics.fetchWaitDuration.Observe(time.Since(start).Seconds())
-	}(time.Now())
-	return r.client.PollFetches(ctx)
-}
-
 // fetchWant represents a range of offsets to fetch.
 // Based on a given number of records, it tries to estimate how many bytes we need to fetch, given there's no support for fetching offsets directly.
 // fetchWant also contains the channel on which to send the fetched records for the offset range.

```

WDYT?

pkg/storage/ingest/reader.go
```diff
@@ -842,15 +821,16 @@ type concurrentFetchers struct {
...
-		tracer:          kotel.NewTracer(kotel.TracerPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))),
-		orderedFetches:  make(chan kgo.FetchPartition),
+		client:             client,
+		logger:             logger,
+		concurrency:        concurrency,
+		topicName:          topic,
+		partitionID:        partition,
+		metrics:            metrics,
+		recordsPerFetch:    recordsPerFetch,
+		lastReturnedRecord: -1, // we still haven't returned the 0 offset.
```
gotjosh, line 831: [nit] `const noRecords = -1` ?

gotjosh:
Dropping the first half of my review, but I have yet to finish all the changes in `reader.go`

pkg/storage/ingest/reader.go
```diff
@@ -218,8 +231,7 @@ func (r *PartitionReader) stopDependencies() error {
 
 func (r *PartitionReader) run(ctx context.Context) error {
 	for ctx.Err() == nil {
-		fetches := r.pollFetches(ctx)
-		err := r.processFetches(ctx, fetches, r.metrics.receiveDelayWhenRunning)
+		err := r.processNextFetches(ctx, r.metrics.receiveDelayWhenRunning)
```
gotjosh, line 234: Now that we have a separate pattern for calling pollFetches, you no longer need to pass the metrics to `processNextFetchets`.

So you can just do:

```golang
r.metrics.receiveDelayWhenRunning.Observe(now.Sub(record.Timestamp).Seconds())
```

instead of:

```golang
delayObserver.Observe(now.Sub(record.Timestamp).Seconds())
```

dimitarvdimitrov: ah, right. good point 👍

pkg/storage/ingest/reader.go
gotjosh: Right! Great catch.

pkg/storage/ingest/reader.go
gotjosh: Ok, I think I get now - this looks like something worth exploring, but we can do it in the next PR. I don't think we need to implement them but just think about what kind of semantics to we want for them and then do it incrementally.

As for the implementation of this request definition itself. Have you tried:

```golang
	req := kmsg.NewFetchRequest()
	req.MinBytes = 1
	req.Version = 13
	req.MaxWaitMillis = 10000
	req.MaxBytes = w.MaxBytes()

	topic := kmsg.NewFetchRequestTopic()
	topic.Topic = r.topicName
	topic.TopicID = r.topicID

	// Using NewFetchRequestTopicPartition gives us a partition with the default values.
	// One of them is CurrentLeaderEpoch: -1, which means we don't know the leader epoch and Kafka brokers are ok with that.
	// It does mean that we might end up fetching from an out-of-sync replica.
	// If we provide this the broker would check if we have up-to-date data.
	partition := kmsg.NewFetchRequestTopicPartition()
	partition.Partition = r.partitionID
	partition.FetchOffset = w.startOffset
	partition.PartitionMaxBytes = req.MaxBytes

	topic.Partitions = append(topic.Partitions, partition)
	req.Topics = append(req.Topics, topic)
```

Upon a closer look:

```golang
CurrentLeaderEpoch: -1,
LastFetchedEpoch:   -1,
LogStartOffset:     -1,
```

The definition of this is already handled in (and potentially more in the future) by:

```golang
func (v *FetchRequestTopicPartition) Default() {
	v.CurrentLeaderEpoch = -1
	v.LastFetchedEpoch = -1
	v.LogStartOffset = -1
}
```

So I would rather use that.

pkg/storage/ingest/reader.go
dimitarvdimitrov: actually, I'm not sure I get it - we still have two flows for calling `processNextFetches` - one passing `receiveDelayWhenRunning` and the other `receiveDelayWhenStarting`.

pkg/storage/ingest/reader.go
```diff
@@ -974,64 +953,66 @@ func (r *concurrentFetchers) getEndOffset(ctx context.Context) (int64, error) {
 }
 
 func (r *concurrentFetchers) runFetchers(ctx context.Context, startOffset int64) {
-	defer level.Info(r.logger).Log("msg", "done running fetchers")
-	wg := sync.WaitGroup{}
-	wg.Add(r.concurrency)
-	defer wg.Wait()
+	fetchersWg := &sync.WaitGroup{}
```
gotjosh, line 956: [nit] Have you considered `dskit_concurrency.ForEachJob` here?

pkg/storage/ingest/reader_test.go
```diff
@@ -871,6 +871,122 @@ func TestPartitionReader_ConsumeAtStartup(t *testing.T) {
...
+		// We expect the reader to catch up, and then switch to Running state.
+		test.Poll(t, 5*time.Second, services.Running, func() interface{} {
+			return reader.State()
+		})
+
+		// We expect the reader to have switched to running because target consumer lag has been honored.
+		assert.Contains(t, logs.String(), "partition reader consumed partition and current lag is lower than configured target consumer lag")
+
+		// We expect the reader to have consumed the partition from start.
+		test.Poll(t, time.Second, []string{"record-1", "record-2"}, func() interface{} {
```
gotjosh, line 954: [nit] (applies for the rest of the use-case) have you considered `require.Eventually` to control how often we poll?

gotjosh (approved):
Overall, LGTM. 

I have might more comments later on once I understand this a bit better and see a final state with `main` but for now, I'd like to keep us moving.

pkg/storage/ingest/reader.go
dimitarvdimitrov: i tried it, but it's not working great - i still need a wait group because ForEachJob blocks and needs to be run in a separate goroutine at which point it's the same as running our own goroutines

pkg/storage/ingest/reader_test.go
dimitarvdimitrov: the rest of the tests run with `test.Poll`, so I opted for consistency within the file at least

pkg/storage/ingest/reader.go
dimitarvdimitrov: I removed the anonymous interface in 9f27ea065a18a6122e84c0cd5f881b1c93b894ed. Looks good?

pkg/storage/ingest/reader.go
dimitarvdimitrov: > As for the implementation of this request definition itself. Have you tried:

Agree, I totally forgot about these utilities. Changed to use them now

pkg/storage/ingest/reader.go
gotjosh: Good point, I missed this. I'm not in love with this pattern and I've got an idea on how to simplify it but let's keep the train moving we can do that later.

pkg/storage/ingest/reader_test.go
gotjosh: Makes sense, thanks!

PR #8969, opened 2024-08-12
Docs: clarify namespace groupName escaping in the rule config API path segments

pracucci:
#### What this PR does

I propose to clarify that `{namespace}` and `{groupName}` path segments in the ruler config API should be URL-escaped.

N/A

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.

### DIFF
--- a/docs/sources/mimir/references/http-api/index.md
+++ b/docs/sources/mimir/references/http-api/index.md
@@ -913,8 +913,8 @@ GET <prometheus-http-prefix>/config/v1/rules/{namespace}
 ```
 
 Returns the rule groups defined for a given namespace.
-The `{namespace}` path segment should be escaped using percent-encoding as defined by
-[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) (for example, `/` should escaped to `%2F`).
+Escape the `{namespace}` path segment using percent-encoding, as defined by
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986). For example, escape `/` to `%2F`.
 
 This endpoint can be disabled via the `-ruler.enable-api` CLI flag (or its respective YAML config option).
 
@@ -946,8 +946,8 @@ GET <prometheus-http-prefix>/config/v1/rules/{namespace}/{groupName}
 ```
 
 Returns the rule group matching the request namespace and group name.
-The `{namespace}` and `{groupName}` path segments should be escaped using percent-encoding as defined by
-[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) (for example, `/` should escaped to `%2F`).
+Escape the `{namespace}` and `{groupName}` path segments using percent-encoding, as defined by
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986). For example, escape `/` to `%2F`.
 
 This endpoint can be disabled via the `-ruler.enable-api` CLI flag (or its respective YAML config option).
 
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

pracucci:
@tacole02 In this PR I'm adding quite a lot of repetition between different documented API endpoints. WDYT? Is it too much repetition? If yes, do we have any recommendation from docs squad on how to avoid it?

docs/sources/mimir/references/http-api/index.md
```diff
@@ -913,6 +913,8 @@ GET <prometheus-http-prefix>/config/v1/rules/{namespace}
 ```
 
 Returns the rule groups defined for a given namespace.
+The `{namespace}` path segment should be escaped using percent-encoding as defined by
```
tacole02, line 916:
```suggestion
Escape the `{namespace}` path segment using percent-encoding, as defined by
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -913,6 +913,8 @@ GET <prometheus-http-prefix>/config/v1/rules/{namespace}
...
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) (for example, `/` should escaped to `%2F`).
```
tacole02, line 917:
```suggestion
[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986). For example, escape `/` to `%2F`.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -944,6 +946,8 @@ GET <prometheus-http-prefix>/config/v1/rules/{namespace}/{groupName}
 ```
 
 Returns the rule group matching the request namespace and group name.
+The `{namespace}` and `{groupName}` path segments should be escaped using percent-encoding as defined by
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) (for example, `/` should escaped to `%2F`).
```
tacole02, line 950:
```suggestion
Escape the `{namespace}` and `{groupName}` path segments using percent-encoding, as defined by
[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986). For example, escape `/` to `%2F`.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -962,6 +966,8 @@ POST /<prometheus-http-prefix>/config/v1/rules/{namespace}
 Creates or updates a rule group.
 This endpoint expects a request with `Content-Type: application/yaml` header and the rules group **YAML** definition in the request body, and returns `202` on success.
 The request body must contain the definition of one and only one rule group.
+The `{namespace}` path segment should be escaped using percent-encoding as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)
+(for example, `/` should escaped to `%2F`).
```
tacole02, line 970:
```suggestion
Escape the `{namespace}` path segment using percent-encoding, as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
For example, escape `/` to `%2F`.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -995,6 +1001,8 @@ DELETE /<prometheus-http-prefix>/config/v1/rules/{namespace}/{groupName}
 ```
 
 Deletes a rule group by namespace and group name. This endpoints returns `202` on success.
+The `{namespace}` and `{groupName}` path segments should be escaped using percent-encoding as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)
+(for example, `/` should escaped to `%2F`).
```
tacole02, line 1005:
```suggestion
Escape the `{namespace}` and `{groupName}` path segments using percent-encoding, as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
For example, escape `/` to `%2F`.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -1011,6 +1019,8 @@ DELETE /<prometheus-http-prefix>/config/v1/rules/{namespace}
 ```
 
 Deletes all the rule groups in a namespace (including the namespace itself). This endpoint returns `202` on success.
+The `{namespace}` path segment should be escaped using percent-encoding as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)
+(for example, `/` should escaped to `%2F`).
```
tacole02, line 1023:
```suggestion
Escape the `{namespace}` path segment using percent-encoding, as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
For example, escape `/` to `%2F`.
```

tacole02 (approved):
Great job! I made a few suggestions to use active voice.

tacole02:
@pracucci I think this repetition is okay, because people aren't necessary reading the docs for each endpoint sequentially. Rather, people are likely using the navigation to jump straight to the endpoint in question.

PR #8972, opened 2024-08-12
change the pod anti affinity when setting zone aware replication

KyriosGN0:
#### What this PR does
this PR changes the second element of the zone aware replication pod anti affinity map from using the NotIn operator to In operator, the reason is that this causes it to sometimes pod the same ingester pod on the same node in the same zone 

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -47,7 +47,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [BUGFIX] Add missing container security context to run `continuous-test` under the restricted security policy. #8653
 * [BUGFIX] Add `global.extraVolumeMounts` to the exporter container on memcached statefulsets #8787
 * [BUGFIX] Fix helm releases failing when `querier.kedaAutoscaling.predictiveScalingEnabled=true`. #8731
-* [BUGFIX] change the pod anti affinity when setting zone aware replication. #8955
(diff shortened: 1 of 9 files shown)


dimitarvdimitrov:
The idea of anti affinity is that two pods of different zones are never scheduled on the same node. Your use case looks like a change better suited for [topology spread constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/). 

https://github.com/grafana/mimir/blob/76c0281226d62a025497b8c01f7771f3144388ae/operations/helm/charts/mimir-distributed/values.yaml#L934-L946

Have you tried configuring that?

dimitarvdimitrov:
can you explain why TSC don't work for you? I was thinking about somehting like 

```yaml
 topologySpreadConstraints: 
   maxSkew: 1 
   topologyKey: kubernetes.io/hostname 
   whenUnsatisfiable: DoNotSchedule
```

This can schedule two ingesters on the same node if and only if all nodes in your cluster already have an ingester.

Another option is to use `ingester.zoneAwareReplication.zones[].extraAffinity` in the values.yaml to add extra `requiredDuringSchedulingIgnoredDuringExecution` affinity.

KyriosGN0:
using `extraAffinity` Worked, however its worth mentioning that i used to overrride the default not add, since if i added only a `requiredDuringSchedulingIgnoredDuringExecution` rules, it will add it outside the pod anti affinity and thus make it an invalid sts manifest

PR #8973, opened 2024-08-12
ingest: improve TestPartitionReader_ConsumerStopping

narqo:
#### What this PR does

This one improves the `TestPartitionReader_ConsumerStopping` test, introduced in #8867. The changes fix occasional deadlock, we see in the CI ([example](https://github.com/grafana/mimir/actions/runs/10349790880/job/28644937482#step:8:77)). As a side effect, the changes speed up the test by eliminating intermediate polling for atomic values.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/reader_test.go
```diff
@@ -155,40 +155,35 @@ func TestPartitionReader_ConsumerStopping(t *testing.T) {
 
 	_, clusterAddr := testkafka.CreateCluster(t, partitionID+1, topicName)
 
-	invocations := atomic.NewInt64(0)
-	blockingConsumer := newTestConsumer(0)
-	consumer := consumerFunc(func(ctx context.Context, records []record) error {
-		invocations.Inc()
-		return blockingConsumer.consume(ctx, records)
+	// consumerErrs will store the last error returned by the consumer; it's initial value doesn't matter, but it must be a non-nil.
```
aknuds1, line 158: [Nit]
```suggestion
	// consumerErrs will store the last error returned by the consumer; its initial value doesn't matter, but it must be non-nil.
```

aknuds1 (approved):
LGTM, modulo a nit. You might want someone more familiar with the code to review though.

pkg/storage/ingest/reader_test.go
```diff
@@ -155,40 +155,35 @@ func TestPartitionReader_ConsumerStopping(t *testing.T) {
...
-		return blockingConsumer.consume(ctx, records)
+	// consumerErrs will store the last error returned by the consumer; its initial value doesn't matter, but it must be non-nil.
+	consumerErrs := atomic.NewError(errors.New("dummy error"))
+	consumeCalls := make(chan chan error)
+	consumer := consumerFunc(func(context.Context, []record) (err error) {
+		defer consumerErrs.Store(err)
+
+		call := make(chan error, 1)
+		consumeCalls <- call
+		return <-call
```
dimitarvdimitrov, line 166: can we just share a `chan error` instead of a `chan chan error`? From what I can see `consumeCalls` is only read from and written to once

dimitarvdimitrov (approved):
SGTM, but the test no longer asserts that the records were all consumed. I'd bring back the assertion on `[][]byte{[]byte{"1"}, []byte{"2"}`

pkg/storage/ingest/reader_test.go
```diff
@@ -155,40 +155,47 @@ func TestPartitionReader_ConsumerStopping(t *testing.T) {
...
+		return <-call.resp
 	})
 	reader := createReader(t, clusterAddr, topicName, partitionID, consumer)
 	require.NoError(t, services.StartAndAwaitRunning(ctx, reader))
 
 	// Write to Kafka.
 	writeClient := newKafkaProduceClient(t, clusterAddr)
-
 	produceRecord(ctx, t, writeClient, topicName, partitionID, []byte("1"))
-	produceRecord(ctx, t, writeClient, topicName, partitionID, []byte("2"))
```
dimitarvdimitrov, line 171: couldn't we avoid the deadlock by just deleting this line?

dimitarvdimitrov, line 171: so my understanding of the deadlock is that record 2 gets produced after record 1 is consumed. So then the reader waits for records 2 to get consumed, which never happens.

Aren't you solving this by just removing record 2? Why are the rest of the changes necessary?

narqo: Good point about records 2. The idea for the changes was to get rid of any polling for changes in atomic values, and explicitly sync the state of the goroutines instead.

Looking at it now, I don't think this test verifies what we initially wanted. I.e. the idea was to test the behaviour around [this line](https://github.com/grafana/mimir/blob/bcde1f9235c5060580b3c269c3f3393c979a58be/pkg/storage/ingest/reader.go#L418-L426). For that we must synchronize around several key points:

1. the data is already fetched from kafka
2. the service is about to stop
3. the records (and the context) are passed to the consumer

I've updated the test to try make it closer to what we expected to test.

dimitarvdimitrov (approved):

PR #8977, opened 2024-08-12
kafka replay speed: fix sharding writer tests

dimitarvdimitrov:
#### What this PR does

This fixes up the tests in the ingest storage package. I needed to make two changes (in two commits) for this

1. include the tenant ID in each unflushed shard. This allows to propagate some valid context when we flush each shard on Close(). Missing tenant ID was failing some tests
2. modify `TestShardingPusher` and `newShardingPusher` so that we can configure the buffer we use for each shard. This allows to trigger blocking in tests and allow draining the errors from a shard

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher_test.go
+++ b/pkg/storage/ingest/pusher_test.go
@@ -464,7 +464,7 @@ func TestShardingPusher(t *testing.T) {
 			upstreamPushErrs: []error{nil},
 			expectedCloseErr: nil,
 		},
-		"push to multiple shards and fill exactly capacity": {
+		"push to multiple shards and fill exact capacity": {
 			shardCount: 2,
 			batchSize:  2,
 			requests: []*mimirpb.WriteRequest{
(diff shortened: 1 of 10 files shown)


pkg/storage/ingest/pusher_test.go
```diff
@@ -425,11 +426,14 @@ type mockPusher struct {
 }
 
 func (m *mockPusher) PushToStorage(ctx context.Context, request *mimirpb.WriteRequest) error {
-	args := m.Called(ctx, request)
+	c := &mimirpb.WriteRequest{}
+	c.Timeseries = slices.Clone(request.Timeseries)
+	args := m.Called(ctx, c)
```
dimitarvdimitrov, line 431:
```suggestion
	args := m.Called(ctx, request)
```

dimitarvdimitrov (commented):
LGTM

PR #8984, opened 2024-08-12
Make `additional-query-queue-dimensions-enabled` into no-op flags

chencs:
#### What this PR does
This PR makes `query-frontend.additional-query-queue-dimensions-enabled` and `query-scheduler.additional-query-queue-dimensions-enabled` into no-op flags, in preparation for fully removing them. In future changes, Mimir will always include "query components" as a queue dimension, so the option is no longer needed.

Users can (and should) delete any `additional-query-queue-dimensions-enabled` flags from their configurations to avoid running into issues when the flags are removed.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/frontend/v2/frontend_test.go
```diff
@@ -142,7 +142,7 @@ func TestFrontendBasicWorkflow(t *testing.T) {
 	})
 
 	req := &httpgrpc.HTTPRequest{
-		Url: "/api/v1/query_range?start=946684800&end=946771200&step=60",
+		Url: "/api/v1/query_range?start=946684800&end=946771200&step=60&query=up{}",
```
chencs, line 145: @francoposa not entirely sure why this worked before without a query, but I'm pretty sure they _should_ all have one.

francoposa: pretty sure it's because flag was off for that test.
The helper method parses the entire query using promql parser, which won't accept a missing a `query_range` request with no `query` param.

chencs: Yeah, I didn't think it was off before my changes because of this line: https://github.com/grafana/mimir/blob/4bce24dc690d9939e968491ad8f5c1328a1263e3/pkg/frontend/v2/frontend_test.go#L71

But I see now that the line immediately following disregards any config already set in `cfg` and sets everything to defaults, which is a little confusing.

francoposa: yeah I guess I missed or didn't understand what that line did

francoposa (approved):
LGTM (assuming the NO-OP marker is in fact what we're supposed to do? I'd leave that to docs team).

I <3 deleting code.

tacole02 (approved):
Instead of `NO-OP`, we could use `Deprecated` or `Non-operational` to describe CLI flags that are retired and soon to be removed.

PR #8988, opened 2024-08-13
Optimization: restrict fetching of last produced offsets only to in-use partitions

pracucci:
#### What this PR does

In this PR I'm proposing an optimization to restrict the fetch of "last produced offsets" in query-frontend only to in-use partitions. In-use partitions are any partitions in the partitions ring, regardless their state (PENDING, ACTIVE, INACTIVE).

Before this change, the query-frontend fetched the last produced offsets of any partition created in Kafka. Since we pre-provision partitions (i.e. we create 1K partitions regardless how many we use, in order to simplify autoscaling), before this change we were always fetching the offsets of 1K partitions. After this change, we only fetch the offsets for the actual number of in-use partitions.

**How it works**:

- Added partitions ring watcher dependency to query-frontend, in order to discover partitions.
- Query-frontend already joined the memberlist cluster, even in the classic architecture. When Mimir is deployed with the classic architecture, the partitions ring watcher and offsets reader are not started (they have no mean in classic architecture), so I don't expect any significant difference in classic architecture.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/querier/api/consistency.go
```diff
@@ -223,11 +222,6 @@ func EncodeOffsets(offsets map[int32]int64) EncodedOffsets {
 	// re-allocations because we under-counted digits.
 	size := versionLen
 	for partitionID, offset := range offsets {
-		// Skip empty partitions.
```
pracucci, line 226: Note to reviewers: this was a trick used to only restrict the propagated offsets to the in-use partitions. We don't need this anymore, because now we fetch only the offsets for in-use partitions.

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -317,8 +308,8 @@ func (p *partitionOffsetClient) fetchTopicPartitionsEndOffsets(ctx context.Conte
 	return list, nil
 }
 
-// listTopicPartitionIDs returns a list of all partition IDs in the topic.
-func (p *partitionOffsetClient) listTopicPartitionIDs(ctx context.Context) ([]int32, error) {
+// ListTopicPartitionIDs returns a list of all partition IDs in the topic.
+func (p *partitionOffsetClient) ListTopicPartitionIDs(ctx context.Context) ([]int32, error) {
```
pracucci, line 312: Note to reviewers: this function is not used anymore, but I would prefer to keep it for now, if there are no objections (otherwise I will remove it). It's a generic function that we may find useful in the future.

pkg/mimir/modules.go
```diff
@@ -719,7 +719,16 @@ func (t *Mimir) initQueryFrontendTopicOffsetsReader() (services.Service, error)
...
 
-	t.QueryFrontendTopicOffsetsReader = ingest.NewTopicOffsetsReader(kafkaClient, t.Cfg.IngestStorage.KafkaConfig.Topic, t.Cfg.IngestStorage.KafkaConfig.LastProducedOffsetPollInterval, t.Registerer, util_log.Logger)
+	// The Kafka partitions may have been pre-provisioned, so the existing partitions in Kafka may be way more
+	// than the actual number of partitions we use. To improve performance, we only look up the actual partitions
+	// we're currently using in Mimir. We include all partition states because ACTIVE and INACTIVE partitions
+	// must be queried, and PENDING partitions may switch to ACTIVE between when the query-frontend fetch the offsets
+	// and the querier builds the replicaset of partitions to query.
+	getPartitionIDs := func() []int32 {
+		return t.IngesterPartitionRingWatcher.PartitionRing().PartitionIDs()
+	}
```
narqo, line 729: To double-check I got it: since `memberlist` itself — implementation underneath the ring — is eventually consistent. Could there be any problems when the state any partition didn't yet fully propagate through the frontend's peers?

gotjosh (approved):
LGTM

I also compared the change to the code introduces in https://github.com/grafana/mimir/pull/8886 and I can see the relevant things have been removed now that we don't need the cache.

pkg/mimir/modules.go
```diff
@@ -719,7 +719,16 @@ func (t *Mimir) initQueryFrontendTopicOffsetsReader() (services.Service, error)
 		return nil, err
 	}
 
-	t.QueryFrontendTopicOffsetsReader = ingest.NewTopicOffsetsReader(kafkaClient, t.Cfg.IngestStorage.KafkaConfig.Topic, t.Cfg.IngestStorage.KafkaConfig.LastProducedOffsetPollInterval, t.Registerer, util_log.Logger)
+	// The Kafka partitions may have been pre-provisioned, so the existing partitions in Kafka may be way more
+	// than the actual number of partitions we use. To improve performance, we only look up the actual partitions
```
narqo, line 723: This comment reads a bit mouthful. Can we rephrase it a bit:

```suggestion
	// The Kafka partitions may have been pre-provisioned. There are may be much more existing partitions in Kafka
	// than the actual number we use. To improve performance, we only look up the actual partitions
```

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -202,22 +196,24 @@ func (p *partitionOffsetClient) fetchPartitionOffset(ctx context.Context, partit
...
 // The Kafka client used under the hood may retry a failed request until the retry timeout is hit.
-//
-// New partitions may not be immediately discovered by this function because the list of existing partitions is cached
-// for a short period of time under the hood. However, the returned offsets for the known partitions are guaranteed
-// to be always updated.
-func (p *partitionOffsetClient) FetchTopicLastProducedOffsets(ctx context.Context) (_ map[int32]int64, returnErr error) {
+func (p *PartitionOffsetClient) FetchPartitionsLastProducedOffsets(ctx context.Context, partitionIDs []int32) (_ map[int32]int64, returnErr error) {
+	// Skip lookup and don't track any metric if no partition was requested.
+	if len(partitionIDs) == 0 {
+		return nil, nil
```
narqo, line 207: (nit) How about we return an empty map in this case?

```suggestion
		return make(map[int32]int64), nil
```

This is a redundant allocation, but it will preserve the invariant, that the resulting map can be accessed by key (i.e. no nil check is required if err is `nil`). The existing code already handles this. But this can make it more future-proof.

integration/e2emimir/services.go
```diff
@@ -178,14 +178,17 @@ func getBinaryNameForBackwardsCompatibility() string {
 	return "mimir"
 }
 
-func NewQueryFrontend(name string, flags map[string]string, options ...Option) *MimirService {
+func NewQueryFrontend(name string, consulAddress string, flags map[string]string, options ...Option) *MimirService {
```
dimitarvdimitrov, line 181: maybe mention that you only need a valid consul address if you're running ingest storage

narqo (approved):
I left a couple of comments, but those aren't blocking. Overall changes work for me 🔥

pkg/storage/ingest/partition_offset_client.go
```diff
@@ -20,18 +19,13 @@ import (
 	"github.com/twmb/franz-go/pkg/kmsg"
 )
 
-// partitionOffsetClient is a client used to read partition offsets.
-type partitionOffsetClient struct {
+// PartitionOffsetClient is a client used to read partition offsets.
+type PartitionOffsetClient struct {
```
dimitarvdimitrov, line 23: why was this and its constructor exported?

pkg/mimir/modules.go
pracucci: Yes, absolutely. The partitions discovery done by query-frontend is just a best effort. The ultimate discovery is done by querier (which BTW could suffer the same issue too, and there's not much we can do about it today). If a strongly consistency request hits an ingester and the offset is missing in the request, then the ingester will fetch the last produced offset again for their partition in order to enforce it. This is an edge case that we don't expect to happen in the real world (excluding the few seconds after we add a new partition) and there's an alert firing if it happens for few minutes consecutively.

dimitarvdimitrov (approved):
LGTM, but I don't understand the reason to export `NewPartitionOffsetClient` which is a good part of the changes in this PR

pracucci:
>  I don't understand the reason to export NewPartitionOffsetClient which is a good part of the changes in this PR

It's handy to use in tests, like in `pkg/frontend/querymiddleware/roundtrip_test.go`.

pkg/storage/ingest/partition_offset_client.go
dimitarvdimitrov: you can index into a `nil` map: https://go.dev/play/p/5d7pZvb8i7g

dimitarvdimitrov:
> > I don't understand the reason to export NewPartitionOffsetClient which is a good part of the changes in this PR
> 
> It's handy to use in tests, like in `pkg/frontend/querymiddleware/roundtrip_test.go`.

I think it's only used as a way to get the list of partitions, right? You can expose a new constructor for `TopicOffsetsReader` which uses the `NewPartitionOffsetClient` it already creates to look up the offsets once. This way you keep the surface area of the ingest package a bit smaller. Not feeling strongly about this though. Feel free to merge as-is

integration/e2emimir/services.go
pracucci: Sure, done in https://github.com/grafana/mimir/pull/8988/commits/5e3dbc02394f3d6288762faac3bf1da4fd839e9d

pracucci:
> > > I don't understand the reason to export NewPartitionOffsetClient which is a good part of the changes in this PR
> > 
> > 
> > It's handy to use in tests, like in `pkg/frontend/querymiddleware/roundtrip_test.go`.
> 
> I think it's only used as a way to get the list of partitions, right? You can expose a new constructor for `TopicOffsetsReader` which uses the `NewPartitionOffsetClient` it already creates to look up the offsets once. This way you keep the surface area of the ingest package a bit smaller. Not feeling strongly about this though. Feel free to merge as-is

Very good feedback Dimitar. I should have done it in https://github.com/grafana/mimir/pull/8988/commits/5111b8c1ec1a4c268339cbfa5c4f2843701c48d9.

PR #8989, opened 2024-08-13
 blockbuilder: consume from lookback when partition doesn't have commit

narqo:
~_This one seats atop the #8982 for now_~

These changes update block-builder making it to start consuming from a point, after the `-block-builder.lookback-on-no-commit` time, when there is no commit for a partition. That is, on the first deploy, instead of starting from the partition's earliest offset, the block-builder consume from the nearest offset after the `lookback-on-no-commit` from its start.

Note: in the future, I want to extend the `-block-builder.lookback-on-no-commit` to handle the value of `-1`, as a way to start from the partition's start. This may be useful when a user doesn't want to lose any historical data.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -45,6 +45,7 @@ type BlockBuilder struct {
 
 	parts []int32
 	// fallbackOffset is the low watermark from where a partition which doesn't have a commit will be consumed.
+	// It can be either a timestamp or the kafkaStartOffset.
 	// TODO(v): to support for setting it to kafkaStartOffset we need to rework how a lagging cycle is chopped into time frames.
 	fallbackOffset int64
 
@@ -176,6 +177,9 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
 	return nil
 }
 
+// kafkaOffsetStart is a special offset value that means the beginning of the partition.
+const kafkaOffsetStart = int64(-2)
+
 func (b *BlockBuilder) findOffsetsToStartAt(ctx context.Context, metrics *kprom.Metrics) (map[int32]kgo.Offset, int64, error) {
 	// We use an ephemeral client to fetch the offset and then create a new client with this offset.
 	// The reason for this is that changing the offset of an existing client requires to have used this client for fetching at least once.
@@ -188,14 +192,25 @@ func (b *BlockBuilder) findOffsetsToStartAt(ctx context.Context, metrics *kprom.
 
 	admClient := kadm.NewClient(cl)
 
-	fallbackOffset := int64(b.cfg.LookbackOnNoCommit)
+	var fallbackOffset int64
 	defaultKOffset := kgo.NewOffset()
-	if fallbackOffset < 0 {
-		defaultKOffset = defaultKOffset.At(fallbackOffset)
-	} else {
+
+	// When LookbackOnNoCommit is positive, its value is used to calculate the timestamp after which the consumption starts,
+	// otherwise it is a special offset for the "start" of the partition.
+	if b.cfg.LookbackOnNoCommit > 0 {
 		ts := time.Now().Add(-b.cfg.LookbackOnNoCommit)
 		fallbackOffset = ts.UnixMilli()
 		defaultKOffset = defaultKOffset.AfterMilli(fallbackOffset)
+	} else {
+		fallbackOffset = int64(b.cfg.LookbackOnNoCommit)
+		switch fallbackOffset {
+		case kafkaOffsetStart:
+			defaultKOffset = defaultKOffset.AtStart()
+		default:
+			// We don't support consuming from the "end" of the partition because starting a cycle from the end always results to a zero lag.
+			// This may be done later.
+			return nil, -1, fmt.Errorf("unexpected fallback offset value %v", b.cfg.LookbackOnNoCommit)
+		}
 	}
 
 	fetchOffsets := func(ctx context.Context) (offsets map[int32]kgo.Offset, err error) {
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -176,11 +176,6 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
 	return nil
 }
 
-const (
-	// kafkaOffsetStart is a special offset value that means the beginning of the partition.
-	kafkaOffsetStart = int64(-2)
-)
-
 func (b *BlockBuilder) findOffsetsToStartAt(ctx context.Context, metrics *kprom.Metrics) (map[int32]kgo.Offset, int64, error) {
 	// We use an ephemeral client to fetch the offset and then create a new client with this offset.
 	// The reason for this is that changing the offset of an existing client requires to have used this client for fetching at least once.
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -169,47 +176,49 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
...
 	cl, err := kgo.NewClient(commonKafkaClientOptions(b.cfg.Kafka, b.logger, metrics)...)
 	if err != nil {
-		return nil, fmt.Errorf("unable to create bootstrap client: %w", err)
+		return nil, -1, fmt.Errorf("unable to create bootstrap client: %w", err)
 	}
 	defer cl.Close()
 
 	admClient := kadm.NewClient(cl)
 
+	fallbackOffset := int64(b.cfg.LookbackOnNoCommit)
```
codesome, line 191: Assigning it like this is a little confusing because it is nanoseconds here while ts.UnixMilli() assigns milliseconds below. But I believe the nanoseconds is never used. How about we directly use `b.cfg.LookbackOnNoCommit < 0` in the `if` condition below and set `fallbackOffset` appropriately?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -169,47 +176,49 @@ func (b *BlockBuilder) starting(ctx context.Context) (err error) {
...
+	defaultKOffset := kgo.NewOffset()
+	if fallbackOffset < 0 {
+		defaultKOffset = defaultKOffset.At(fallbackOffset)
```
codesome, line 194: Are we using `LookbackOnNoCommit` as a hack to set -1 and -2 to consume from beginning or end? Needs to be documented here in code and for the config.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -441,17 +456,35 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, topic, group strin
...
-	if len(offsets) == 0 {
-		offsets = make(kadm.OffsetResponses)
-		for t, pt := range startOffsets.Offsets() {
-			resp := make(map[int32]kadm.OffsetResponse)
-			for _, o := range pt {
-				resp[o.Partition] = kadm.OffsetResponse{
-					Offset: o,
-				}
+	listOffsetsAfterMilli := sync.OnceValues(func() (kadm.ListedOffsets, error) {
+		return admClient.ListOffsetsAfterMilli(ctx, fallbackOffset, topic)
```
codesome, line 460: How does this work when fallbackOffset is negative?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -441,17 +456,35 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, topic, group strin
...
-	// For now, if the group doesn't have any commits, we use startOffsets as the starting point.
-	if len(offsets) == 0 {
-		offsets = make(kadm.OffsetResponses)
-		for t, pt := range startOffsets.Offsets() {
-			resp := make(map[int32]kadm.OffsetResponse)
-			for _, o := range pt {
-				resp[o.Partition] = kadm.OffsetResponse{
-					Offset: o,
-				}
+	listOffsetsAfterMilli := sync.OnceValues(func() (kadm.ListedOffsets, error) {
```
codesome, line 459: TIL about `sync.OnceValues` 👏

pkg/blockbuilder/blockbuilder.go
```diff
@@ -441,17 +456,35 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, topic, group strin
...
+				continue
+			}
+			fallbackOffsets, err := listOffsetsAfterMilli()
+			if err != nil {
+				return nil, err
+			}
+			o, ok := fallbackOffsets.Lookup(topic, part)
+			if !ok {
+				// This should not happen.
+				return nil, fmt.Errorf("partition %d not found in fallback offsets for topic %s", part, topic)
```
codesome, line 480: Can this error occur if the partition has no data?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -40,9 +41,13 @@ type BlockBuilder struct {
 	register    prometheus.Registerer
 	limits      *validation.Overrides
 	kafkaClient *kgo.Client
-	parts       []int32
 	bucket      objstore.Bucket
 
+	parts []int32
+	// fallbackOffset is the low watermark from where a partition which doesn't have a commit will be consumed.
+	// TODO(v): to support for setting it to kafkaStartOffset we need to rework how a lagging cycle is chopped into time frames.
+	fallbackOffset int64
```
codesome, line 49: Maybe appropriate to name it like this because we are doing stuff like `time.UnixMilli(b.fallbackOffset)`
```suggestion
	fallbackTimeOffset int64
```

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -718,19 +723,10 @@ func TestBlockBuilder_LongCatchupWithNoRecordsProcessed(t *testing.T) {
...
-	// second compaction happens.
-	select {
-	case <-compactCalled:
-	case <-time.After(15 * time.Second):
-		t.Fatal("second compaction did not happen within expected time")
-	}
+	time.Sleep(5 * time.Second)
+
+	// Nothing is consumed due to zero lag.
+	require.Zero(t, compactCalled.Load(), "expected skipping all records before lookback period")
```
codesome, line 729: Do these test change correctly test the changes? I will have a proper look again.

pkg/blockbuilder/blockbuilder.go
narqo: Yes, moving forward, I want to allow `b.cfg.LookbackOnNoCommit` to accept a special string value "start", that converts into `-1` (this also how kgo works with the offsets internally)

pkg/blockbuilder/blockbuilder_test.go
narqo: I think it does. Before the changes we tried to consume from the lag up until the lookback and called compact on an empty tsdb. Now, there is no "first" (empty) cycle, because we start consuming straight at the lookback record.

pkg/blockbuilder/blockbuilder.go
narqo: Below there is a check

```
// If fallback is not a timestamp, it's the kafkaOffsetStart value.
if fallbackOffset < 0 {
    ...
```

I will make this chunk more explicit

pkg/blockbuilder/blockbuilder.go
narqo: Here, we're inside the loop that iterates over topic-partitions from the resolved `startOffsets`. I.e. the partition here is known to have data.

narqo: I've added a test for this and other cases. PHAL

pkg/blockbuilder/blockbuilder.go
narqo: I'm not sure about this one: the comment for the `fallbackOffset` mentions, that it is expected to be either a special `kafkaStartOffset` or a timestamp (in the future). Thus, the code should expect it to be a non-time value. I've expanded the comment to be more explicit about that.

codesome (approved):
I just added a TODO in the code. If that is actually done in tests and I failed to notice is, then we can remove the TODO. In the meantime, I will merge this PR. LGTM 👍 

```
	// TODO: add a test to test the case where the consumption on startup happens
	// after the LookbackOnNoCommit with records before the after the consumption
	// start point.
```

PR #8991, opened 2024-08-13
 blockbuilder: consume partition up to its lag

narqo:
~_This one seats atop the https://github.com/grafana/mimir/pull/8989_~

This is part 3 of the internal's refactoring. Here we improve how `consumePartition` tracks the partition's lag and make the partition's lag, the high watermark of the cycles. I.e. a full cycle now is expected to consume at most `partitionInfo.Lag` records. This brings a couple of benefits:

1. Previously, we didn't take into account that the number of seen records isn't the same as number of processed records. In the case when the consumer catches up, it may incorrectly report lag `-1` in to logs (this didn't hurt but was confusing).
2. This simplifies the consumer loop, because we no longer need to care about occasional empty fetches. The consumer will retry fetching until the "initial_lag" records got consumed.
3. This should unblock further refactoring to consume multiple partitions in parallel.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -408,7 +408,7 @@ func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			level.Warn(b.logger).Log("msg", "error unmarshalling commit metadata", "err", err, "part", part, "offset", offset.At, "metadata", offset.Metadata)
 		}
 
-		pl := &partitionInfo{
+		pinf := &partitionInfo{
 			Partition:      part,
 			Lag:            lag.Lag,
 			Commit:         offset,
@@ -416,7 +416,7 @@ func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			SeenTillOffset: seenTillOffset,
 			LastBlockEnd:   lastBlockEnd,
 		}
-		if err := b.nextConsumeCycle(ctx, b.kafkaClient, pl, cycleEnd); err != nil {
+		if err := b.nextConsumeCycle(ctx, b.kafkaClient, pinf, cycleEnd); err != nil {
 			level.Error(b.logger).Log("msg", "failed to consume partition", "err", err, "part", part)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -867,54 +848,49 @@ func (b *BlockBuilder) consumePartition(
 			Offset: firstRec.Offset,
 		}
 		b.seekPartition(client, part, rec)
-		return pl, nil
+		return nil
 	}
 
 	// All samples in all records were processed. We can commit the last record's offset.
 	if commitRec == nil {
 		commitRec = lastRec
```
codesome, line 856: Just noticed that if all samples faced tsdb soft error, `lastRec` will be nil. We should avoid panics and probably skip commits if lastRec and commitRec both are nil.


But this may cause BB to get stuck in replaying the same data, in which case committing the last record may make sense even if TSDB rejected it. But avoiding panics is better than trying again.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -867,54 +848,49 @@ func (b *BlockBuilder) consumePartition(
...
+		Topic:       commitRec.Topic,
+		Partition:   commitRec.Partition,
+		At:          commitRecOffset,
+		LeaderEpoch: commitRec.LeaderEpoch,
+		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), pl.SeenTillOffset, pl.LastBlockEnd),
 	}
+	pl.CommitRecTs = commitRec.Timestamp.UnixMilli()
 
-	return pl, commitOffset(ctx, b.logger, client, b.cfg.Kafka.ConsumerGroup, pl.Commit)
+	return commitOffset(ctx, b.logger, client, b.cfg.Kafka.ConsumerGroup, pl.Commit)
```
codesome, line 893: We should ideally update `pl` only when `commitOffset()` does not return an error. How about we call `commitOffset()` before making changes to `pl`?

codesome (commented):
LGTM with a couple of comments (one unrelated to this PR but its a small fix)

pkg/blockbuilder/blockbuilder.go
narqo: Isn't that what we handle below https://github.com/grafana/mimir/blob/9280ea7fadb2a30201b395f33aae59b68021a9e3/pkg/blockbuilder/blockbuilder.go#L809-L818 ?

narqo:
I'll merge it, and we can address the https://github.com/grafana/mimir/pull/8991#discussion_r1728110768 — if still needed — as part of backporting the branch into `main`.

PR #8992, opened 2024-08-13
Fix mergeIterator.Seek()

codesome:
#### What this PR does

mergeIterator.Seek() was skipping samples in some cases. This was happening because in the Seek() method we were discarding the current batch from mergeIterator if the batch's first sample was after the seek time.

Fixes #<issue number>

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/batch/merge_test.go
+++ b/pkg/querier/batch/merge_test.go
@@ -64,8 +64,8 @@ func TestMergeHarder(t *testing.T) {
 	}
 }
 
-// TestBugInMergeIterator tests a bug while calling Seek() on mergeIterator.
-func TestBugInMergeIteratorSeek(t *testing.T) {
+// TestMergeIteratorSeek tests a bug while calling Seek() on mergeIterator.
+func TestMergeIteratorSeek(t *testing.T) {
 	// Samples for 3 chunks.
 	chunkSamples := [][]int64{
 		{10, 20, 30, 40},
--- a/pkg/querier/batch/merge_test.go
+++ b/pkg/querier/batch/merge_test.go
@@ -6,13 +6,13 @@
 package batch
 
 import (
-	"github.com/prometheus/prometheus/tsdb/chunkenc"
-	"github.com/stretchr/testify/require"
 	"testing"
 	"time"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/tsdb/chunkenc"
+	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/mimir/pkg/storage/chunk"
 )
--- a/pkg/querier/batch/merge_test.go
+++ b/pkg/querier/batch/merge_test.go
@@ -65,13 +65,12 @@ func TestMergeHarder(t *testing.T) {
 }
 
 // TestBugInMergeIterator tests a bug while calling Seek() on mergeIterator.
-// The samples used for testing is from a live Mimir cluster where the bug was first observed.
 func TestBugInMergeIteratorSeek(t *testing.T) {
 	// Samples for 3 chunks.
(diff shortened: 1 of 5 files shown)


charleskorn (commented):
Thanks for tracking this down and fixing it!

Looks like `chunkIterator.Seek` has a similar check (and therefore I assume the same bug) - should we change that as well?

I'm also wondering if we should clarify the description of `iterator.Seek` in `batch.go` to be clear that it'll seek to the batch with a sample at or after the desired timestamp, not just a batch at or after the desired timestamp:

```diff
type iterator interface {
-	// Seek to the batch at (or after) time t.
+	// Seek to the batch with sample at (or after) time t.
	Seek(t int64, size int) chunkenc.ValueType
```

pkg/querier/batch/merge_test.go
```diff
@@ -61,3 +63,47 @@ func TestMergeHarder(t *testing.T) {
 		})
 	}
 }
+
+// TestBugInMergeIterator tests a bug while calling Seek() on mergeIterator.
+func TestBugInMergeIteratorSeek(t *testing.T) {
```
dimitarvdimitrov, line 68: nit on naming: aren't all tests testing for a bug? 
```suggestion
// TestMergeIterator tests a bug while calling Seek() on mergeIterator.
func TestMergeIteratorSeek(t *testing.T) {
```

pkg/querier/batch/merge_test.go
```diff
@@ -61,3 +63,47 @@ func TestMergeHarder(t *testing.T) {
...
+			require.NoError(t, err)
+			require.NoError(t, chk.UnmarshalFromBuf(ch.Bytes()))
+			return chk.NewIterator(reuse)
+		}))
+	}
+
+	c3It := NewGenericChunkMergeIterator(nil, labels.EmptyLabels(), genericChunks)
+
+	c3It.Seek(15)
+	// These Next() calls are necessary to reproduce the bug.
```
dimitarvdimitrov, line 96: are they necessary so that we cross the chunk boundary or?

dimitarvdimitrov (commented):
thanks!

pkg/querier/batch/merge_test.go
codesome: Yes I think so. These three Next() cross from chunk 1 to 2. I will probably simply the test if it still reproduces the earlier bug

codesome: Looks like there is something specific with these three Next() that reproduces it with this input. I don't think I want to dig further into details of why at the moment :)

codesome:
> Looks like chunkIterator.Seek has a similar check (and therefore I assume the same bug) - should we change that as well?

I updated it similar to mergeIterator, but I am not sure if we can reproduce the bug in that one with same input since chunkIterator is just for a single chunk. I think we can leave it as it is with the potential fix. Not sure if we want to invest time in trying to break chunkIterator.Seek.

charleskorn (approved):

charleskorn:
>I think we can leave it as it is with the potential fix. Not sure if we want to invest time in trying to break chunkIterator.Seek.

I think it'd be good to add a test to ensure we don't reintroduce this bug in the future, but this can come in a follow-up PR.

dimitarvdimitrov (approved):
LGTM. Plus one on the test for chunkIterator (but better to merge this now before the weekend)

Maybe we can create some general test for all iterators which ensures that they comply with the interface. I think I found a bug in memoizingIterator in that it skips the first sample if you first call Next() on it. But we never use that, so it's still not causing problems. I can have a go at that next week.

codesome:
> but this can come in a follow-up PR

+1

charleskorn:
> > but this can come in a follow-up PR
> 
> +1

I've added a test in https://github.com/grafana/mimir/pull/9040

PR #8993, opened 2024-08-14
Mimir query engine: emit annotations for binary operations between incompatible native histograms

charleskorn:
#### What this PR does

This PR adds the same annotations as https://github.com/grafana/mimir/pull/8924, this time for binary operations such as addition and subtraction.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/engine_test.go
```diff
@@ -1444,6 +1444,21 @@ func TestAnnotations(t *testing.T) {
 			},
 		},
 
+		"binary operation between native histograms with exponential and custom buckets": {
+			data: nativeHistogramsWithCustomBucketsData,
+			expr: `metric{series="exponential-buckets"} + ignoring(series) metric{series="custom-buckets-1"}`,
+			expectedWarningAnnotations: []string{
+				`PromQL warning: vector contains a mix of histograms with exponential and custom buckets schemas for metric name "" (1:1)`,
```
charleskorn, line 1451: Note to reviewers: these messages look weird given we have no metric name in the case of binary operations, but it's what Prometheus does as well, so I've kept it the same for the sake of consistency.

jhesketh: We should open an upstream issue to fix it there and then eventually in MQE

pkg/streamingpromql/operators/binary_operation.go
```diff
@@ -812,7 +822,13 @@ func (b *BinaryOperation) computeResult(left types.InstantVectorSeriesData, righ
 			// Timestamps match at this step
 			resultFloat, resultHist, ok, err := b.opFunc(lF, rF, lH, rH)
 			if err != nil {
-				return types.InstantVectorSeriesData{}, err
+				err = functions.NativeHistogramErrorToAnnotation(err, b.emitAnnotation)
+				if err == nil {
+					// Error was converted to an annotation, continue without emitting a sample here.
+					ok = false
```
jhesketh, line 828: (nit) I think `ok` will always be false when there is an error. No harm in being explicit though.

jhesketh (approved):

pkg/streamingpromql/operators/binary_operation.go
charleskorn: Yep, that's exactly my thinking.

pkg/streamingpromql/engine_test.go
charleskorn: Yep, done: https://github.com/prometheus/prometheus/issues/14669

PR #8995, opened 2024-08-14
Mimir query engine: add support for scalar constants and `pi()`

charleskorn:
#### What this PR does

This PR adds support for scalar constants and the `pi()` function in MQE.

(none)

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


tacole02 (approved):
Thank you!

pkg/streamingpromql/operators/scalar_constant.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	samples = samples[:numSteps]
+
+	for step := 0; step < numSteps; step++ {
+		samples[step].T = s.Start + int64(step)*s.Interval
+		samples[step].F = s.Value
+	}
+
+	return types.ScalarData{Samples: samples}, nil
```
jhesketh, line 59: I wonder (and this probably applies for other operators too) if we run the risk of forgetting to put something back into the pool. Would it be better for each operator to use its `Close` to return items to the pool once it is done with them? We could have GetValues return a closer that the caller needs to handle if the timing is difficult.

pkg/streamingpromql/query.go
```diff
@@ -492,6 +574,8 @@ func (q *Query) Close() {
 		types.PutMatrix(v)
 	case promql.Vector:
 		types.VectorPool.Put(v, q.memoryConsumptionTracker)
+	case promql.Scalar:
+		// Nothing to do, we already returned the slice in populateScalarFromScalarOperator.
```
jhesketh, line 593: do we need to return the slice in the case of a Matrix?

pkg/streamingpromql/query.go
```diff
@@ -473,6 +535,26 @@ func (q *Query) populateMatrixFromRangeVectorOperator(ctx context.Context, o typ
...
+	return promql.Matrix{
+		{
+			Metric: labels.EmptyLabels(),
+			Floats: d.Samples,
+		},
+	}
+}
+
+func (q *Query) populateScalarFromScalarOperator(d types.ScalarData) promql.Scalar {
+	defer types.FPointSlicePool.Put(d.Samples, q.memoryConsumptionTracker)
```
jhesketh, line 563: this seems like an odd place to put this. I mean it makes sense, we're done with the samples, but it feels easy for us to miss something and not return something to a pool somewhere.

jhesketh (commented):
lgtm, just a question around handling of pools. These can likely be moved to another discussion or PR given this is in line with what we're currently doing. (But still a question on the matrix).

pkg/streamingpromql/operators/scalar_constant.go
charleskorn: >if we run the risk of forgetting to put something back into the pool

It's not ideal if we forget to put something back into the pool, but the impact is that we'll just allocate more slices more often than we'd otherwise need to do.

>Would it be better for each operator to use its `Close` to return items to the pool once it is done with them? 

The downside of this is that an operator doesn't know how its slices will be used, and each operator would need to then track all slices it returns, which would keep them alive and increase peak memory consumption.

For example, let's say operator A returns a slice of `FPoint`s to operator B.

If operator B is an operation like `-metric`, it could mutate that slice in place and pass it to the operator consuming it, without getting another slice. If both operators implemented the "return to pool on `Close`" pattern, then both A and B would return the same slice to the pool, so it would be used twice (pools don't do any kind of deduplication). 

Furthermore, imagine operator A returns many series - it would then need to hold a reference to the slice for each series so that it can return them to the pool when `Close` is called.

Lastly, delaying returning slices to the pool until `Close` delays returning those slices until the query is complete. This means those slices can't be reused by the same query, but a single query tends to need many slices of the same size, one for each series, so this would be a great candidate for pooling.

pkg/streamingpromql/query.go
charleskorn: Yep - this is already handled a few lines above in this method.

pkg/streamingpromql/query.go
charleskorn: I'm not sure I follow sorry - we've consistently followed the pattern of "return the slice when we don't need the slice any more". This method is the place where we stop needing the slice, and after this point, we won't have a reference to the slice any more, so this is our last opportunity to return the slice to the pool.

jhesketh: I was thinking as a defer on line 405 so that it's next to where we get the values.

pkg/streamingpromql/operators/scalar_constant.go
jhesketh: Thanks for the detailed examples. I'm aware there would be extra overhead, but I'm wondering if we can add some more safeguards in place. Unfortunately I don't have any concrete ideas how this might look. Perhaps some kind of test that can check everything has been put back at the end, or a way of asserting that if you receive a value from this (or similar methods) the caller is responsible for putting xyz back. Anyway, this is probably a discussion outside of this PR, but it's something to keep in mind.

jhesketh (approved):
lgtm :-)

pkg/streamingpromql/operators/scalar_constant.go
charleskorn: >Perhaps some kind of test that can check everything has been put back at the end, 

I've been thinking about this as well - perhaps we could add a test to ensure that the estimated current memory consumption is 0 after a query is complete. If it was negative, then we've returned something to the pool multiple times, and if it was positive, then we've missed returning something to the pool.

I've also wondered if we should just panic if the estimated current memory consumption goes negative, as that would indicate something being returned multiple times and would almost certainly lead to incorrect query results.

But yep, I think this is something for another PR.

>or a way of asserting that if you receive a value from this (or similar methods) the caller is responsible for putting xyz back

The convention we've been following is that if you receive a slice from an operator (samples or metadata), you can mutate it as you please, and you're responsible for returning the slice to the pool when you're done (or passing it to something else that has the same responsibility).

pkg/streamingpromql/query.go
charleskorn: Are you talking about near line 388, where we call `GetValues()`? (line 405 is populating the warnings on the query results)

If so, it's not safe to return the slice at that point (unless it's in the `if` branch that calls `populateScalarFromScalarOperator`), as the slice will be used as-is for the matrix result.

jhesketh: I think our line numbers might be getting confused. Ignoring that though, you're right, it would need to be in that if branch, so this seems like a good enough place :+1:

pkg/streamingpromql/operators/scalar_constant.go
jhesketh: > But yep, I think this is something for another PR.

Sounds good :-).

> The convention we've been following is that if you receive a slice from an operator (samples or metadata), you can mutate it as you please, and you're responsible for returning the slice to the pool when you're done (or passing it to something else that has the same responsibility).

Yep, but I meant a way of enforcing or checking that convention.

PR #8975, opened 2024-08-12
ingester: set active series default to 20m

dimitarvdimitrov:
#### What this PR does

In GL we bill based on active series in the last 20 minutes. This is also the value we want bill GEM for. It would be easier to reason about active series if they have the same default value everywhere. Because of this I'm changing the default value from 10m to 20m.

I'll leave this open until the end of the week to gather more feedback.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #8998, opened 2024-08-14
[gem-release-2.13] ingester: set active series default to 20m 

dimitarvdimitrov:
backports #8975 (1501b90e1cd93e4c39c21fba9ded7987089f2512)

### Backport process

The idea is that we don't want to backport this PR to the Mimir 2.13 release because it's not worth the hassle and it's a strange change to include. Plus we're not doing 2.13.1 now and just putting this change here might surprise someone in the future if 2.13.1 is necessary. So I branched off from `release-2.13` into `gem-release-2.13` and am now backporting this change there. I will use `gem-release-2.13` in GEM instead of `release-2.13`


pstibrany:
Could we override default value in GEM 2.13 by using approach similar to https://github.com/grafana/mimir/blob/1fd9d2f95c61511f0fcb81d082426721973c7b5b/pkg/mimir/mimir.go#L546-L572 ? (in GEM code)

Then we would remove it in future GEM releases. WDYT?

dimitarvdimitrov:
we could. Can only do this in 2.13 actually, so we don't have to deal with it in future versions.

PR #9006, opened 2024-08-14
 blockbuilder: handle non-terminal errors when processing record samples

narqo:
#### What this PR does

~_This is the part 2 or N, which is currently seats atop the #8978. I will move it out of the draft after the latter gets merged._~

This PR updates the block-builder's tsdb wrapper to properly handle non-terminal (soft) errors, raised while a record's samples are being appended. While block-builder is still an early experiment, we skip testing every possible case for a soft error. We can add more tests as we get more comfortable with the design.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/tsdb.go
+++ b/pkg/blockbuilder/tsdb.go
@@ -101,44 +101,6 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record, lastBlockMax
 		}
 	}()
 
-	mapAppendError := func(err error) error {
-		if err == nil {
-			return nil
-		}
-
-		// Check if error is one of the "soft errors" we can proceed on without terminating.
-		// Same as https://github.com/grafana/mimir/blob/1eb4b8e1e3293df100d7fc4df0c94712c31a0930/pkg/ingester/ingester.go#L1283-L1284
-		switch {
-		case errors.Is(err, storage.ErrOutOfBounds):
-			return nil
-		case errors.Is(err, storage.ErrOutOfOrderSample):
-			return nil
-		case errors.Is(err, storage.ErrTooOldSample):
-			return nil
-		case errors.Is(err, globalerror.SampleTooFarInFuture):
-			return nil
-		case errors.Is(err, storage.ErrDuplicateSampleForTimestamp):
-			return nil
-		case errors.Is(err, globalerror.MaxSeriesPerUser):
-			return nil
-		case errors.Is(err, globalerror.MaxSeriesPerMetric):
-			return nil
-
-		// Map TSDB native histogram validation errors to soft errors.
-		case errors.Is(err, histogram.ErrHistogramCountMismatch):
-			return nil
-		case errors.Is(err, histogram.ErrHistogramCountNotBigEnough):
-			return nil
-		case errors.Is(err, histogram.ErrHistogramNegativeBucketCount):
-			return nil
-		case errors.Is(err, histogram.ErrHistogramSpanNegativeOffset):
-			return nil
-		case errors.Is(err, histogram.ErrHistogramSpansBucketsMismatch):
-			return nil
-		}
-		return err
-	}
-
 	var (
 		labelsBuilder       labels.ScratchBuilder
 		nonCopiedLabels     labels.Labels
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/tsdb.go
```diff
@@ -100,6 +101,44 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record, lastBlockMax
 		}
 	}()
 
+	mapAppendError := func(err error) error {
```
dimitarvdimitrov, line 104: maybe move this to a function of its own so that Process can be read a bit more quickly?

dimitarvdimitrov, line 104: also - shouldn't you use this mapping when committing the appender too? I think it might return some similar errors as during  Append

codesome: I guess any error in Commit() is a hard error. Ingester also treats as a hard error. If there is any out of order sample or any soft error, Commit() ignores it (from top of my head).

codesome (commented):
How about we add a metric for soft and hard errors? They can be useful.

pkg/blockbuilder/tsdb_test.go
```diff
@@ -170,28 +165,48 @@ func TestTSDBBuilder(t *testing.T) {
...
+				rec := createRequest(oooUserID, floatSample(lastEnd+20, 1), nil, false)
+				allProcessed, err := builder.Process(context.Background(), rec, lastEnd, currEnd, false)
+				require.NoError(t, err)
+				require.True(t, allProcessed)
+
+				// A. "ErrOutOfOrderSample" (soft error)
+				rec = createRequest(oooUserID, floatSample(lastEnd-20, 1), nil, false)
+				allProcessed, err = builder.Process(context.Background(), rec, lastEnd, currEnd, false)
+				require.NoError(t, err)
+				require.True(t, allProcessed)
```
codesome, line 209: To actually verify that it took the first sample and not the second, we should query the `test-ooo-tenant`'s block after it is created and expect only `lastEnd+20`.

codesome (commented):
LGTM, after the above and below comment. Thanks!

narqo:
> How about we add a metric for soft and hard errors?

I'd rather revisited metrics after #9068 (and its follow-up). In the `blockbuilder` branch we already have base metrics about compaction timings, etc. We will refactor those, pushing them down to the `TSDBBuilder, and will add the missing ones, like soft-errors.

pkg/blockbuilder/tsdb.go
dimitarvdimitrov: TIL that the [headAppender.Commit](https://github.com/grafana/mimir/blob/fefa35e43bcc36c58f8baa3b1de0171b5f590b28/vendor/github.com/prometheus/prometheus/tsdb/head_append.go#L797) just never returns any errors

pkg/blockbuilder/tsdb.go
```diff
@@ -183,6 +192,45 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record, lastBlockMax
...
 }
 
+// checkTSDBAppendError checks if err is a non-terminal error, that should not block processing other series in the batch.
+func checkTSDBAppendError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Check if error is one of the "soft errors" we can proceed on without terminating.
+	// Same as https://github.com/grafana/mimir/blob/1eb4b8e1e3293df100d7fc4df0c94712c31a0930/pkg/ingester/ingester.go#L1283-L1284
```
dimitarvdimitrov, line 202: do you want to keep those two functions in sync? did you consider adding a test or moving them to a shared package? Not super likely we change them, but I can see how _if_ we fix one, we're likely to fix the other.

dimitarvdimitrov (approved):
LGTM, maybe a minor comment about keeping this in sync with the ingester

pkg/blockbuilder/tsdb.go
narqo: Right now almost all of `tsdb.go` is a stripped version of the code from the ingester's tsdb wrapper. We should definitely find a way to share pieces of ingesters, ingest-store and block-builder after it all gets into shape.

PR #9012, opened 2024-08-15
add HA dedupe functionality to docker dev environment

aldernero:
#### What this PR does

This PR updates the microservers docker compose development environment so that it can more easily be used to test the functionality and changes to the HA tracker. Specifically it
- adds an option `enable_ha_tracker` to toggle the functionality
- when the above is enabled, 2 additional prometheii are started that scrape the `memcache-exporter` and `load-generator` and add the relevant cluster and replica labels required by the HA tracker.
- the two targets scraped for the HA tracker are now removed from the base prometheus scrape targets
- the relevant flags to enable HA tracking were added to the mimir config

I chose to use 2 smaller prometheii instead of having 2 prometheii that scrape all the targets to save on memory requirements.

#### Checklist

- [N/A] Tests updated.
- [N/A] Documentation added.


development/mimir-microservices-mode/docker-compose.jsonnet
```diff
@@ -62,12 +64,14 @@ std.manifestYamlDoc({
       name: 'distributor-1',
       target: 'distributor',
       httpPort: 8000,
+      extraArguments: '-distributor.ha-tracker.consul.hostname=consul:8500',
```
56quarters, line 67: This seems like it should be conditional based on the `enable_ha_tracker` config as well as the other HA config bits in `mimir.yaml`.

development/mimir-microservices-mode/docker-compose.jsonnet
```diff
@@ -33,6 +33,7 @@ std.manifestYamlDoc({
 
     // If true, a query-tee instance with a single backend is started.
     enable_query_tee: false,
+    enable_ha_tracker: true,
```
56quarters, line 36: Can you include some docs here (some of the info from the description of this PR would be good)?

development/mimir-microservices-mode/docker-compose.jsonnet
56quarters: Or alternatively, the jsonnet config should be removed so HA is always used.

56quarters (approved):

development/mimir-microservices-mode/docker-compose.jsonnet
aldernero: I went with option 2, always have HA used and group the 2 new prometheus with existing `enable_prometheus` toggle

PR #9013, opened 2024-08-15
query scheduler: querier-worker queue priority algo implementation

francoposa:
#### What this PR does

This PR introduces the implementation of our querier-worker queue prioritization algorithm for the scheduler's multi-algorithm queue tree and benchmarks it against other combinations of queuing algorithms.

The implementation and benchmark scenarios are intended to address scenarios where one query component (ingesters or store-gateways) is experiencing high latency in query processing due to high load or other issues.

In our current state, the queue selection algorithms have no mechanism to prevent the slow query component from dominating the utilization of the available querier-worker capacity, leaving queries for the other query component sitting in the queue even though those queries will not utilize the slow query component and therefore could be being serviced without issue or significant delay.

#### What this PR does not do
This PR does not actually configure the new queueing algorithm to be utilized in the production `MultiAlgorithmTreeQueue`.
That will be done in a later PR after further benchmarking and observation in live Mimir clusters.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


francoposa:
sample output from `TestMultiDimensionalQueueAlgorithmSlowConsumerEffects`

```
tree: non-flipped round-robin tree, 1 tenant, 05pct slow queries: seconds in queue: [ingester: mean: 0.4663 stddev: 0.06 store-gateway: mean: 0.1768 stddev: 0.13]
tree: querier-worker priority tree, 1 tenant, 05pct slow queries: seconds in queue: [ingester: mean: 0.1291 stddev: 0.07 store-gateway: mean: 0.2487 stddev: 0.15]
tree: non-flipped round-robin tree, 1 tenant, 10pct slow queries: seconds in queue: [ingester: mean: 0.8560 stddev: 0.14 store-gateway: mean: 0.3831 stddev: 0.25]
tree: querier-worker priority tree, 1 tenant, 10pct slow queries: seconds in queue: [ingester: mean: 0.1239 stddev: 0.07 store-gateway: mean: 0.4673 stddev: 0.27]
tree: non-flipped round-robin tree, 1 tenant, 25pct slow queries: seconds in queue: [ingester: mean: 1.9395 stddev: 0.49 store-gateway: mean: 1.0058 stddev: 0.61]
tree: querier-worker priority tree, 1 tenant, 25pct slow queries: seconds in queue: [ingester: mean: 0.1033 stddev: 0.06 store-gateway: mean: 1.1097 stddev: 0.63]
tree: non-flipped round-robin tree, 1 tenant, 50pct slow queries: seconds in queue: [ingester: mean: 3.0945 stddev: 1.32 store-gateway: mean: 2.0037 stddev: 1.18]
tree: querier-worker priority tree, 1 tenant, 50pct slow queries: seconds in queue: [ingester: mean: 0.0716 stddev: 0.04 store-gateway: mean: 2.0371 stddev: 1.18]
tree: non-flipped round-robin tree, 1 tenant, 75pct slow queries: seconds in queue: [ingester: mean: 2.0081 stddev: 1.19 store-gateway: mean: 3.1358 stddev: 1.84]
tree: querier-worker priority tree, 1 tenant, 75pct slow queries: seconds in queue: [ingester: mean: 0.0397 stddev: 0.02 store-gateway: mean: 3.0144 stddev: 1.75]
tree: non-flipped round-robin tree, 1 tenant, 90pct slow queries: seconds in queue: [ingester: mean: 0.6751 stddev: 0.41 store-gateway: mean: 3.7965 stddev: 2.22]
tree: querier-worker priority tree, 1 tenant, 90pct slow queries: seconds in queue: [ingester: mean: 0.0147 stddev: 0.01 store-gateway: mean: 3.7246 stddev: 2.17]
tree: non-flipped round-robin tree, 1 tenant, 95pct slow queries: seconds in queue: [ingester: mean: 0.4905 stddev: 0.30 store-gateway: mean: 3.8926 stddev: 2.27]
tree: querier-worker priority tree, 1 tenant, 95pct slow queries: seconds in queue: [ingester: mean: 0.0062 stddev: 0.00 store-gateway: mean: 3.9796 stddev: 2.32]
```

pkg/scheduler/queue/query_component_utilization.go
```diff
@@ -97,65 +72,6 @@ func NewQueryComponentUtilization(
 	}, nil
 }
 
-// ExceedsThresholdForComponentName checks whether a query component has exceeded the capacity utilization threshold.
```
francoposa, line 100: this was all for when we were going to use this for the load-balancing algorithm

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -0,0 +1,117 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package queue
+
+type QuerierWorkerQueuePriorityAlgo struct {
+	currentQuerierWorker  int
+	currentNodeOrderIndex int
+	nodeOrder             []string
+	nodeCounts            map[string]int
```
chencs, line 9: What's the purpose of this? Like what do we read this for?

francoposa: this is for the equivalent of the `tenantNodes` in TQA - namely for when the algorithm is not at the root of the tree and therefore than can be multiple of each node type, we track that we don't need to add a node to the rotation if an instance of that node type already exists somewhere, and we shouldn't delete it from the rotation until the final instance is removed - I plan to add more docstring on this, but there is an full unit test for the behavior in this PR.

pkg/scheduler/queue/queue_test.go
```diff
@@ -93,156 +104,6 @@ func makeSchedulerRequest(tenantID string, additionalQueueDimensions []string) *
...
-// If queue items belonging to the slow category are in the same queue ("normal-channel") in front of the normal queue
-// items, the normal queue items must wait for all slow queue items to be cleared before they can be serviced.
-// In this way, the degraded performance of the slow query component equally degrades the performance of the
-// queries which *could* be serviced quickly, but are waiting behind the slow queries in the queue.
-//
-// When using multiple queue dimensions, the queues are split by which "component" the query will utilize -- in this
-// test, those components are called "normal-channel" and "slow-channel" for clarity. The queue broker then
-// round-robins between the multiple queues, which has the effect of alternately dequeuing from the slow queries
-// and normal queries rather than blocking normal queries behind slow queries.
-func TestMultiDimensionalQueueFairnessSlowConsumerEffects(t *testing.T) {
```
francoposa, line 112: benchmark only relevant for comparing old versions of the tree and we have a better benchmark test now, introduced in `pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go`

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
chencs: Do we ever expect this algo to not be at the root of the tree? i.e., is this more than we realistically need to keep track of?

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority_test.go
```diff
@@ -0,0 +1,345 @@
...
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{ingesterQueueDimension}, "obj-4"))
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{storeGatewayQueueDimension}, "obj-5"))
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{ingesterAndStoreGatewayQueueDimension}, "obj-6"))
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{ingesterQueueDimension}, "obj-7"))
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{storeGatewayQueueDimension}, "obj-8"))
+	require.NoError(t, tree.EnqueueBackByPath(QueuePath{ingesterAndStoreGatewayQueueDimension}, "obj-9"))
+
+	// node order was set by initial enqueue order;
+	// order will remain until a node is deleted for being empty or a new node is added by an enqueue
+	expectedInitialNodeOrder := []string{ingesterQueueDimension, storeGatewayQueueDimension, ingesterAndStoreGatewayQueueDimension}
```
chencs, line 102: Do you need this check? You can also essentially check order via dequeuing objects (you expect the first thing dequeued by querier-worker 0 to be from `ingesterQueueDimension`, etc), and it involves less peeping of internals.

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
```diff
@@ -0,0 +1,468 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+	"golang.org/x/sync/errgroup"
+)
+
+const querierForgetDelay = 0
+const maxOutStandingPerTenant = 1000000
+
+// newBenchmarkRequestQueue mirrors the constructor for RequestQueue, but with the ability to pass in a Tree
```
chencs, line 29: Could probably just call `q := NewRequestQueue()` in here with all the same args, then set `q.tree = tree`, right?

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
francoposa: I was under the impression that we would eventually make it so `use-multi-algorithm-tree` would use this algo regardless, and `prioritize-query-components` would put it at the first level.
Whether or not there is a config option placing this at a non-root level of the tree moving forward, I am not comfortable really having one of these algorithms in the codebase that breaks if it's not put at the correct level.

The breakage scenario is really bad too - prematurely deleting a node type from the rotation means it would not be selected and all queries sitting in that queue would be ignored & unreachable indefinitely until another enqueue of that node type re-adds it to the broken rotation tracking.

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority_test.go
francoposa: So technically yes, *anything* could be happening to the node order behind the scenes and as long as the workerID->deueueNode relation holds from outside observation, then it's working.

However because the dequeue order depends on which workerID is dequeuing, I want both checks - that the order has not been changed when it should not be changed, and that the queue selection by workerID is correct.

I generally take the opinion that these lowest-level unit tests are also supposed to serve as unambiguous documentation of intent for how something is intended to work, even if that behavior is not necessarily visible when observing the system from a wider angle. It's for the next engineer who works on it to understand the intended inner machinations a bit better.

This is the lowest-level most contained direct unit test of this class, so I believe this is the right place to peek internals (when it matters). I would not want to assert on this in `queue_test.go` or `multi_queuing_algorithm_tree_queue_test.go` or the benchmarks, etc.

francoposa:
**New Benchmark Results with Multi-Tenant Scenarios**

<details>

<summary>

**Results by Query Component**

</summary>

```
tree: non-flipped round-robin tree, 1 tenant, 05pct slow queries: seconds in queue: [ingester: mean: 0.4751 stddev: 0.07 store-gateway: mean: 0.1816 stddev: 0.13]
tree: new flipped round-robin tree, 1 tenant, 05pct slow queries: seconds in queue: [ingester: mean: 0.4748 stddev: 0.07 store-gateway: mean: 0.1814 stddev: 0.13]
tree: querier-worker priority tree, 1 tenant, 05pct slow queries: seconds in queue: [ingester: mean: 0.1318 stddev: 0.08 store-gateway: mean: 0.2142 stddev: 0.13]
tree: non-flipped round-robin tree, 1 tenant, 10pct slow queries: seconds in queue: [ingester: mean: 0.7331 stddev: 0.11 store-gateway: mean: 0.3096 stddev: 0.20]
tree: new flipped round-robin tree, 1 tenant, 10pct slow queries: seconds in queue: [ingester: mean: 0.8821 stddev: 0.14 store-gateway: mean: 0.3945 stddev: 0.25]
tree: querier-worker priority tree, 1 tenant, 10pct slow queries: seconds in queue: [ingester: mean: 0.1265 stddev: 0.07 store-gateway: mean: 0.3650 stddev: 0.21]
tree: non-flipped round-robin tree, 1 tenant, 25pct slow queries: seconds in queue: [ingester: mean: 2.0674 stddev: 0.55 store-gateway: mean: 1.0932 stddev: 0.66]
tree: new flipped round-robin tree, 1 tenant, 25pct slow queries: seconds in queue: [ingester: mean: 2.0442 stddev: 0.53 store-gateway: mean: 1.0725 stddev: 0.65]
tree: querier-worker priority tree, 1 tenant, 25pct slow queries: seconds in queue: [ingester: mean: 0.1001 stddev: 0.06 store-gateway: mean: 1.1315 stddev: 0.65]
tree: non-flipped round-robin tree, 1 tenant, 50pct slow queries: seconds in queue: [ingester: mean: 3.1152 stddev: 1.36 store-gateway: mean: 2.0439 stddev: 1.21]
tree: new flipped round-robin tree, 1 tenant, 50pct slow queries: seconds in queue: [ingester: mean: 3.1476 stddev: 1.40 store-gateway: mean: 2.0742 stddev: 1.23]
tree: querier-worker priority tree, 1 tenant, 50pct slow queries: seconds in queue: [ingester: mean: 0.0648 stddev: 0.04 store-gateway: mean: 2.1826 stddev: 1.26]
tree: non-flipped round-robin tree, 1 tenant, 75pct slow queries: seconds in queue: [ingester: mean: 1.9731 stddev: 1.16 store-gateway: mean: 3.1599 stddev: 1.85]
tree: new flipped round-robin tree, 1 tenant, 75pct slow queries: seconds in queue: [ingester: mean: 2.0608 stddev: 1.22 store-gateway: mean: 3.1114 stddev: 1.82]
tree: querier-worker priority tree, 1 tenant, 75pct slow queries: seconds in queue: [ingester: mean: 0.0366 stddev: 0.02 store-gateway: mean: 3.0599 stddev: 1.78]
tree: non-flipped round-robin tree, 1 tenant, 90pct slow queries: seconds in queue: [ingester: mean: 0.7781 stddev: 0.48 store-gateway: mean: 3.7424 stddev: 2.19]
tree: new flipped round-robin tree, 1 tenant, 90pct slow queries: seconds in queue: [ingester: mean: 0.8182 stddev: 0.49 store-gateway: mean: 3.7271 stddev: 2.18]
tree: querier-worker priority tree, 1 tenant, 90pct slow queries: seconds in queue: [ingester: mean: 0.0142 stddev: 0.01 store-gateway: mean: 3.7285 stddev: 2.18]
tree: non-flipped round-robin tree, 1 tenant, 95pct slow queries: seconds in queue: [ingester: mean: 0.4828 stddev: 0.30 store-gateway: mean: 3.8950 stddev: 2.27]
tree: new flipped round-robin tree, 1 tenant, 95pct slow queries: seconds in queue: [ingester: mean: 0.3220 stddev: 0.21 store-gateway: mean: 3.9738 stddev: 2.32]
tree: querier-worker priority tree, 1 tenant, 95pct slow queries: seconds in queue: [ingester: mean: 0.0071 stddev: 0.00 store-gateway: mean: 3.9417 stddev: 2.30]
tree: non-flipped round-robin tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue: [ingester: mean: 2.1504 stddev: 1.22 store-gateway: mean: 2.0608 stddev: 1.22]
tree: new flipped round-robin tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue: [ingester: mean: 3.1455 stddev: 1.40 store-gateway: mean: 2.0840 stddev: 1.23]
tree: querier-worker priority tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue: [ingester: mean: 0.0707 stddev: 0.04 store-gateway: mean: 2.0341 stddev: 1.18]
tree: non-flipped round-robin tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue: [ingester: mean: 2.1583 stddev: 1.18 store-gateway: mean: 2.0540 stddev: 1.21]
tree: new flipped round-robin tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue: [ingester: mean: 3.1210 stddev: 1.37 store-gateway: mean: 2.0637 stddev: 1.22]
tree: querier-worker priority tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue: [ingester: mean: 0.0694 stddev: 0.04 store-gateway: mean: 2.0687 stddev: 1.20]
tree: non-flipped round-robin tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue: [ingester: mean: 2.5163 stddev: 1.04 store-gateway: mean: 2.0650 stddev: 1.22]
tree: new flipped round-robin tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue: [ingester: mean: 3.1690 stddev: 1.44 store-gateway: mean: 2.1476 stddev: 1.27]
tree: querier-worker priority tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue: [ingester: mean: 0.0682 stddev: 0.04 store-gateway: mean: 2.1087 stddev: 1.22]
tree: non-flipped round-robin tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue: [ingester: mean: 3.1026 stddev: 1.33 store-gateway: mean: 2.0115 stddev: 1.19]
tree: new flipped round-robin tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue: [ingester: mean: 3.0782 stddev: 1.29 store-gateway: mean: 1.9658 stddev: 1.16]
tree: querier-worker priority tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue: [ingester: mean: 0.0679 stddev: 0.04 store-gateway: mean: 2.1317 stddev: 1.23]
```

</details>

<details>

<summary>

**Results for Ingester-Only Queries by Tenant ID**

</summary>

```
tree: non-flipped round-robin tree, 1 tenant, 05pct slow queries: seconds in queue:[tenant-0: mean: 0.4751 stddev: 0.07]
tree: new flipped round-robin tree, 1 tenant, 05pct slow queries: seconds in queue:[tenant-0: mean: 0.4748 stddev: 0.07]
tree: querier-worker priority tree, 1 tenant, 05pct slow queries: seconds in queue:[tenant-0: mean: 0.1318 stddev: 0.08]
tree: non-flipped round-robin tree, 1 tenant, 10pct slow queries: seconds in queue:[tenant-0: mean: 0.7331 stddev: 0.11]
tree: new flipped round-robin tree, 1 tenant, 10pct slow queries: seconds in queue:[tenant-0: mean: 0.8821 stddev: 0.14]
tree: querier-worker priority tree, 1 tenant, 10pct slow queries: seconds in queue:[tenant-0: mean: 0.1265 stddev: 0.07]
tree: non-flipped round-robin tree, 1 tenant, 25pct slow queries: seconds in queue:[tenant-0: mean: 2.0674 stddev: 0.55]
tree: new flipped round-robin tree, 1 tenant, 25pct slow queries: seconds in queue:[tenant-0: mean: 2.0442 stddev: 0.53]
tree: querier-worker priority tree, 1 tenant, 25pct slow queries: seconds in queue:[tenant-0: mean: 0.1001 stddev: 0.06]
tree: non-flipped round-robin tree, 1 tenant, 50pct slow queries: seconds in queue:[tenant-0: mean: 3.1152 stddev: 1.36]
tree: new flipped round-robin tree, 1 tenant, 50pct slow queries: seconds in queue:[tenant-0: mean: 3.1476 stddev: 1.40]
tree: querier-worker priority tree, 1 tenant, 50pct slow queries: seconds in queue:[tenant-0: mean: 0.0648 stddev: 0.04]
tree: non-flipped round-robin tree, 1 tenant, 75pct slow queries: seconds in queue:[tenant-0: mean: 1.9731 stddev: 1.16]
tree: new flipped round-robin tree, 1 tenant, 75pct slow queries: seconds in queue:[tenant-0: mean: 2.0608 stddev: 1.22]
tree: querier-worker priority tree, 1 tenant, 75pct slow queries: seconds in queue:[tenant-0: mean: 0.0366 stddev: 0.02]
tree: non-flipped round-robin tree, 1 tenant, 90pct slow queries: seconds in queue:[tenant-0: mean: 0.7781 stddev: 0.48]
tree: new flipped round-robin tree, 1 tenant, 90pct slow queries: seconds in queue:[tenant-0: mean: 0.8182 stddev: 0.49]
tree: querier-worker priority tree, 1 tenant, 90pct slow queries: seconds in queue:[tenant-0: mean: 0.0142 stddev: 0.01]
tree: non-flipped round-robin tree, 1 tenant, 95pct slow queries: seconds in queue:[tenant-0: mean: 0.4828 stddev: 0.30]
tree: new flipped round-robin tree, 1 tenant, 95pct slow queries: seconds in queue:[tenant-0: mean: 0.3220 stddev: 0.21]
tree: querier-worker priority tree, 1 tenant, 95pct slow queries: seconds in queue:[tenant-0: mean: 0.0071 stddev: 0.00]
tree: non-flipped round-robin tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue:[tenant-0: mean: 2.2437 stddev: 1.17 tenant-1: mean: 0.2929 stddev: 0.18]
tree: new flipped round-robin tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue:[tenant-0: mean: 3.2674 stddev: 1.30 tenant-1: mean: 0.4010 stddev: 0.34]
tree: querier-worker priority tree, 2 tenants, first with 5pct slow queries, second with 95pct slow queries: seconds in queue:[tenant-0: mean: 0.0747 stddev: 0.04 tenant-1: mean: 0.0099 stddev: 0.00]
tree: non-flipped round-robin tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue:[tenant-0: mean: 2.3074 stddev: 1.12 tenant-1: mean: 0.5588 stddev: 0.32]
tree: new flipped round-robin tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue:[tenant-0: mean: 3.3485 stddev: 1.22 tenant-1: mean: 1.1224 stddev: 0.91]
tree: querier-worker priority tree, 2 tenants, first with 10pct slow queries, second with 90pct slow queries: seconds in queue:[tenant-0: mean: 0.0750 stddev: 0.04 tenant-1: mean: 0.0120 stddev: 0.01]
tree: non-flipped round-robin tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue:[tenant-0: mean: 2.7256 stddev: 0.98 tenant-1: mean: 1.9386 stddev: 0.98]
tree: new flipped round-robin tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue:[tenant-0: mean: 3.4509 stddev: 1.30 tenant-1: mean: 2.1399 stddev: 1.45]
tree: querier-worker priority tree, 2 tenants, first with 25pct slow queries, second with 75pct slow queries: seconds in queue:[tenant-0: mean: 0.0787 stddev: 0.04 tenant-1: mean: 0.0297 stddev: 0.02]
tree: non-flipped round-robin tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue:[tenant-0: mean: 3.1190 stddev: 1.32 tenant-1: mean: 3.0862 stddev: 1.34]
tree: new flipped round-robin tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue:[tenant-0: mean: 3.0698 stddev: 1.33 tenant-1: mean: 3.0866 stddev: 1.25]
tree: querier-worker priority tree, 2 tenants, first with 50pct slow queries, second with 50pct slow queries: seconds in queue:[tenant-0: mean: 0.0644 stddev: 0.04 tenant-1: mean: 0.0712 stddev: 0.04]

```

</details>

chencs:
> new algo flipped tree still looks great, but vanilla round-robin flipped tree looks worse than the un-flipped version.

I believe this is expected because we are round-robining across fewer subtrees (maximum 4 query components instead of potentially thousands of tenants), so we're _guaranteeing_ that every fourth query or so, a querier is getting stuck waiting on the slow store-gateway, regardless of tenant it's from.

This means that `query-scheduler.prioritize-query-components` should _not_ be used without this querier-worker queue priority algorithm. I'll temporarily change it to a no-op in order to prevent operators from accidentally degrading performance until we have this in.

jhesketh (approved):

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
```diff
@@ -0,0 +1,533 @@
...
+// significant delay while the queue consumer processes it and before the consumer can dequeue the next item.
+// This emulates a situation where one of the query components - the ingesters or store-gateways - is under load.
+//
+// In a single-queue or single-queue-per-tenant scenario, the slow query component will cause the queue to back up.
+// When queue items belonging to the slow category are in the same queue in front of the normal queue items,
+// the normal queue items must wait for all slow queue items to be cleared before they can be serviced.
+// In this way, the degraded performance of the slow query component equally degrades the performance of the
+// queries which *could* be serviced quickly, but are waiting behind the slow queries in the queue.
+//
+// In this benchmark, queries for the store-gateway query component are artificially slowed down.,
```
chencs, line 237:
```suggestion
// In this benchmark, queries for the store-gateway query component are artificially slowed down.
```

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -0,0 +1,187 @@
...
+//
+//  4. (b) Otherwise, if none of those tenants are sharded to this querier, the tree traversal algorithm will return
+//     back up to the parent level and ask the QuerierWorkerQueuePriorityAlgo to select its next node.
+//     We continue to step 5.
+//
+//  5. The QuerierWorkerQueuePriorityAlgo will select the next node in the nodeOrder, "store-gateway".
+//     We return to step 3 and continue through steps 3, 4b, and 5, until we reach step 4a and exit.
+//
+// This process of continuing to search for requests to dequeue helps prevent querier-worker capacity from sitting idle
+// when there are no requests to dequeue for the query component node that the querier-worker was originally mapped to.
```
chencs, line 74:
```suggestion
// QuerierWorkerQueuePriorityAlgo implements QueuingAlgorithm, mapping worker IDs (assigned incrementally) to a queue
// node to prioritize. Querier-workers' priority is determined by currentQuerierWorker % len(nodeOrder), resulting in
// fair assignment of querier-workers across query component subtrees. Thus, when one query component is experiencing
// high latency (e.g., store-gateways), about 25% of querier-workers continue prioritizing queries for unaffected
// components.
//
// There are four possible query components: "ingester", "store-gateway", "ingester-and-store-gateway", and "unknown".
// When all four possible query component queue nodes exist, one out of four worker IDs starts dequeuing from any given
// queue node. For example, a querier-worker with ID 4 and node order ["ingester", "store-gateway",
// "ingester-and-store-gateway", "unknown"] first attempts to dequeue a query from the "ingester" node. However, it is
// not required that the number of querier-workers be divisible by four; this algorithm assumes that
// MultiQueuingAlgorithmTreeQueue always deletes empty leaf nodes and nodes with no children after a dequeue operation,
// and if a new request is enqueued for a previously-deleted paths, nodes are re-created. Thus:
//  1. The modulo operation may modulo the worker ID by 1, 2, 3, or 4 depending on the number of node types
//     currently present in the node order, which can change which node a worker ID is prioritized for.
//  2. The node order changes as queues are deleted and re-created, so the worker ID-to-node mapping changes as the
//     enqueue order places query component nodes in different positions in the order.
//
// These assumptions guarantee that even when the number of querier-workers is not evenly divisible by the number of
// query component nodes, through the randomized changes in node order over time, they are more likely to be evenly
// distributed across query component nodes than if length and order of those nodes were fixed.
//
// For a given worker, queue nodes are prioritized -- not assigned. Ceteris paribus, the same worker ID starts at a
// consistent queue node, but moves on to other nodes if it cannot dequeue a request from the subtree of its first
// prioritized queue node. This maximally utilizes querier-worker capacity when there are no process-able requests for
// a specific querier-worker and query component.
//
// A querier-worker can process queries for nodes it has not prioritized when this queue algorithm is applied at the
// highest layer of the tree and the tenant-querier-shuffle-shard queue algorithm applied at the second layer of the
// tree. If shuffle-sharding is enabled, a querier-worker that prioritizes ingester-only queries may not find
// ingester-only queries for any tenant it is assigned to, and move on to the next query component subtree. E.g.:
//
//  1. The QuerierWorkerQueuePriorityAlgo has node order:
//     ["ingester", "store-gateway", "ingester-and-store-gateway", "unknown"].
//
//  2. A querier-worker with workerID 0 requests to dequeue; it prioritizes the "ingester" queue node.
//
//  3. The dequeue operation recursively dequeues from "ingester" node. Each child node is a tenant-specific
//     queue of ingester-only queries. The tenantQuerierAssignments QueuingAlgorithm checks if any of its child nodes
//     (tenant queues) is assigned to this querier, and finds none.
//
//  4. We walk back up to the QuerierWorkerQueuePriorityAlgo level, not having dequeued anything. The
//     QuerierWorkerQueuePriorityAlgo increments currentNodeOrderIndex and selects the next node in nodeOrder (in
//     this example, "store-gateway"), and checks for dequeue-able queries again, from step 3, etc. until a
//     dequeue-able child is found, or every query component node has been checked for dequeue-able queries.
//
// TestMultiDimensionalQueueAlgorithmSlowConsumerEffects illustrates that this algorithm significantly outperforms
// a simple round-robin algorithm. Because round-robining selects a given query-component node ~25% of the time, when
// one query component is experiencing high latency, eventually, almost every querier-worker will be stuck processing
// queries for that query component.
```

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority_test.go
```diff
@@ -0,0 +1,345 @@
...
+		├── b
+		│   ├── ingester
+		│	│	├── obj-b-i-1
+	    │   │   ├── obj-b-i-2
+		│   ├── store-gateway
+		│	│	├── obj-b-sg-1
+	    │   │   ├── obj-b-sg-2
+		│   └── ingester-and-store-gateway
+		│	 	├── obj-b-isg-1
+	    │       ├── obj-b-isg-2
```
chencs, line 227: Not sure what Github is doing here; formatting looks weird in browser, but looks fine when I pull down the branch.

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
```diff
@@ -0,0 +1,533 @@
...
+// We assume that in such a slowdown scenario that queries hitting both query components (ingester and store-gateway)
+// will be constrained by the latency of the slowest query component, so we treat them as store-gateway queries.
+// The goal of the queueing algorithms is to enable ingester-only queries to continue to be serviced quickly
+// when the store-gateway queries are slow; we are not concerned with de-loading the store-gateways
+// or otherwise improving the performance for the store-gateway queries.
+// Benchmark results should be interpreted primarily in terms of the performance of the ingester-only queries.
+func TestMultiDimensionalQueueAlgorithmSlowConsumerEffects(t *testing.T) {
```
chencs, line 244: IIUC there's no assertion in this test to ensure that we perform better/worse with specific test configurations; is there a way we can add this, but not have it run with the general unit tests? Running the tests in CI for the `queue` package went from taking ~30s to ~172s with this change.

francoposa: I wanna check with some other people on how they feel about this - I got some pushback on asserting on the benchmark outputs on the previous version of this for fear of them being flaky in CI, but I was able to configure it where it was stable.

I think there's also a fair question of like, when this is all said and done after the cleanup we won't be using vanilla round robin anywhere. Should we still keep something like that in a code base just to show it's worse?

w.r.t the total test runtime, we have a few levers:
* remove one of the two slow tree configurations
* lower the iterations
* lower the slow consumer delay
* crank up worker concurrency
* nuclear: skip in CI altogether somehow until we come up with a larger strategy for regular benchmarking

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
```diff
@@ -0,0 +1,533 @@
...
+			name string
+			tree Tree
+		}{
+			// keeping these names the same length keeps logged results aligned
+			{
+				"non-flipped round-robin tree",
+				nonFlippedRoundRobinTree,
+			},
+			{
+				"new flipped round-robin tree",
```
chencs, line 429:
```suggestion
				"flipped round-robin tree",
```

francoposa: :laughing: I hear you.. but you know how much time I have spent trying to make the names both somewhat accurate but also the same length? The output is a bear to read if we don't line everything up and I preferred not to just pad with extra characters.

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
francoposa: so I lopped off the 5% / 95%  cases since they generally show the same effects as the 10% / 90% effects and these cases take the longest
I also cranked up querier and worker parallellism - total runtime on my local is down to 14s.

As far as asserting on differences, it looks like I could probably assert on the new algo being 10x better in all cases _except_ the 10% case. In general with lower % slow queries, the gains are less pronounced so it would need to be a sliding scale. Gonna ask some more mimir people how they generally feel about this sort of approach.

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
francoposa: after some effort I think I got better descriptions while managing to keep the same length:

```golang
{
	"tenant-querier -> query component round-robin tree",
	nonFlippedRoundRobinTree,
},
{
	"tenant-querier -> query component round-robin tree",
	flippedRoundRobinTree,
},
{
	"worker-queue prioritization -> tenant-querier tree",
	querierWorkerPrioritizationTree,
},
```

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
francoposa: this was helpful - did a merge on our different wording and ordering here, and managed to cut the same number of lines (20) as your suggestions here.

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority_test.go
francoposa: Might be your UI setting for tab width appearance - which is at https://github.com/settings/appearance

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue_benchmark_test.go
francoposa: done

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority_test.go
chencs: You're right; mine was defaulted to 8

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -4,74 +4,54 @@ package queue
...
-// This reservation ensures that the querier-worker connections can continue to process queries
-// which do not need to utilize the query component experiencing high latency.
-//
-// Performance:
-// This significantly outperforms the previous round-robin algorithm which simply rotated through the node order
-// (see TestMultiDimensionalQueueAlgorithmSlowConsumerEffects benchmark outputs for comparison).
+// This significantly outperforms the previous round-robin approach which simply rotated through the node order.
 // Although a vanilla round-robin algorithm will select a given query-component node 1 / 4 of the time,
-// in situations of high latency on a query component that one slow query component will still grow asymptotically
+// in situations of high latency on a query component, the one slow query component will still grow asymptotically
```
chencs, line 12:
```suggestion
// in situations of high latency on a query component, the queue size of one slow query component will still grow asymptotically
```

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -4,74 +4,54 @@ package queue
...
 // to dominate the utilization the querier-worker connections, as measured by inflight query processing time.
 //
-// Implementation Details & Assumptions:
-// The MultiAlgorithmTreeQueue which utilizes this and other QueuingAlgorithm implementations always deletes
-// nodes for paths through the tree which lead to an empty leaf node after a dequeue operation.
-// Nodes for paths through the tree are then re-created when a new request is enqueued which requires that path.
-// This means that a tree will not always have all 4 node types for the 4 possible query component assignments.
+// There are 4 possible query components: "ingester", "store-gateway", "ingester-and-store-gateway", and "unknown".
+// With all 4 queue nodes active, approximately 1 / 4 of the querier-workers are prioritized to each queue node.
```
chencs, line 16: I think the concept of a node being "active" is a little confusing; they either exist or don't, but there's never a component node that exists but is "inactive."
```suggestion
// When all 4 queue nodes exist, approximately 1 / 4 of the querier-workers are prioritized to each queue node.
```

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -4,74 +4,54 @@ package queue
...
+//  3. The dequeue operation attempts to dequeue first from the child nodes of the "ingester" node,
+//     where each child node is a tenant-specific queue of ingester-only queries. The tenantQuerierAssignments
+//     QueuingAlgorithm checks if any of its tenant queue nodes is sharded to this querier, and finds none.
 //
-// This process of continuing to search for requests to dequeue helps prevent querier-worker capacity from sitting idle
-// when there are no requests to dequeue for the query component node that the querier-worker was originally mapped to.
+//  4. The dequeue operation walks back up to the QuerierWorkerQueuePriorityAlgo level, not having dequeued anything.
+//     The QuerierWorkerQueuePriorityAlgo moves on and selects the next query-component node in the nodeOrder,
+//     and recurs again to search that next subtree for tenant queue nodes sharded to this querier, from step 3, etc.,
+//     until a dequeue-able tenant queue node is found, or every query component node subtree has been exhausted.
```
chencs, line 54: Is it possible for a querier to find _no_ appropriate tenant node to dequeue from in the entire tree? This seems like it would be occasion for a reshuffle.

chencs: (I'm not saying we should reshuffle in this case; just that this case:
> or every query component node subtree has been exhausted.

seems impossible to reach

francoposa: as far as the "could a querier not find a tenant node", this is not something we are changing, the querier could always possibly not find a tenant node but the only way I can think that it would actually occur would be some absurd shuffle shard config.

I think it would have to have num queriers > num tenants, and every tenant would have to have very few max queriers (like 1) so that some queriers would actually get shuffled to zero tenants.
I don't know if our shuffle shard logic guards against this at all but we dont use querier shuffle shard that much or that restrictively.

In that case a reshuffle wouldn't solve anyway.

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
francoposa: done

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
francoposa: I updated to this:

```
// in situations of high latency on a query component, the utilization of the querier-worker connections
// as measured by inflight query processing time will grow asymptotically to be dominated by the slow query component.
```

I am trying to express that the round robin still allows the slow query component to grow to use all connections:
With the slow component only being 8x slower than the fast component - taking 8 time slices to process instead of 1, by time slice 10, all consumers are processing the slow queries, and we will spend the rest of the time essentially bouncing between 0 and 1 consumers processing fast queries.

|  | round-robin selection state | consumer 1 | consumer 2 | consumer 3 | consumer 4 |
| :---- | :---- | :---- | :---- | :---- | :---- |
| time 1 | fast | dequeue fast (advance to slow) |  |  |  |
| time 2 | slow | process fast | dequeue slow (advance to fast) |  |  |
| time 3 | fast | wait | process slow | dequeue fast (advance to slow) |  |
| time 4 | slow | wait | process slow | process fast | dequeue slow (advance to fast) |
| time 5 | fast | dequeue fast (advance to slow) | process slow | wait | process slow |
| time 6 | slow | process fast | process slow | dequeue slow (advance to fast) | process slow |
| time 7 | fast | dequeue fast (advance to slow) | process slow | process slow | process slow |
| time 8 | slow | process fast | process slow | process slow | process slow |
| time 9 | slow | dequeue slow (advance to fast) | process slow | process slow | process slow |
| time 10 | fast | process slow | process slow | process slow | process slow |
| time 11 | fast | process slow | dequeue fast (advance to slow) | process slow | process slow |
| time 12 | slow | process slow | process fast | process slow | process slow |
| time 13 | fast | process slow | wait | process slow | dequeue slow (advance to fast) |
| time 14 | fast | process slow | dequeue fast (advance to slow) | process slow | process slow |
| time 15 | slow | process slow | process fast | dequeue slow (advance to fast) | process slow |
| time 16 | fast | process slow | dequeue fast (advance to slow) | process slow | process slow |
|  |  | process slow | process fast | process slow | process slow |
|  |  |  |  | process slow | process slow |
|  |  |  |  | process slow | process slow |
|  |  |  |  | process slow | process slow |
|  |  |  |  | process slow | process slow |
|  |  |  |  | process slow |  |
|  |  |  |  | process slow |  |
|  |  |  |  |  |  |
|  |  |  |  |  |  |

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
Logiraptor: Long term I do think we should limit the number of ways to configure it to what we use in production. Once that is the case, I think it would be simpler and likely faster to remove `nodeCounts` and just use `slices.Contains(nodeOrder, name)` instead. The map is expected to only have 4 elements - searching a slice of that size is faster than a map lookup.

pkg/scheduler/queue/queue.go
```diff
@@ -655,38 +659,46 @@ func FirstTenant() TenantIndex {
 	return TenantIndex{last: -1}
 }
 
-// waitingQuerierConn is a "request" indicating that the querier is ready to receive the next query request.
-// It embeds the unbuffered `recvChan` to receive the requestForQuerier "response" from the RequestQueue.
+// QuerierWorkerDequeueRequest is a request from a querier-worker which is ready to receive the next query.
+// It embeds the unbuffered `recvChan` to receive the querierWorkerDequeueResponse "response" from the RequestQueue.
 // The request/response terminology is avoided in naming to disambiguate with the actual query requests.
```
Logiraptor, line 664: This comment may be out of date? Since the types are renamed to include request / response

pkg/scheduler/queue/tree_queue_algo_querier_worker_queue_priority.go
```diff
@@ -0,0 +1,167 @@
...
+		childName := dequeuedFrom.Name()
+
+		// decrement the global nodeCounts
+		qa.nodeCounts[childName]--
+
+		// only delete from global nodeOrder if the global nodeCount is now zero
+		// meaning there are no nodes with this name remaining in the tree
+		if qa.nodeCounts[childName] == 0 {
+			for idx, name := range qa.nodeOrder {
+				if name == childName {
```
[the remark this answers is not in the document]
Logiraptor: Idk what happened with that diff, but it is supposed to replace the whole loop, not just those two lines 🤷

Logiraptor (approved):

chencs (approved):

PR #9016, opened 2024-08-15
Create flag to prioritize dequeuing from query components over tenant fairness

chencs:
#### What this PR does
This PR introduces a flag, `query-scheduler.prioritize-query-components`, which allows operators to configure how preferentially the query scheduler dequeues from query components (also known as queue dimensions) vs. from tenants.

In the current state, the query scheduler constructs a tree queue which looks something like this:
```
root
├─ tenant-1
│  ├─ unknown
│  ├─ ingester
│  ├─ store-gateway
│  └─ ingester-and-store-gateway
...
└─ tenant-n
   ├─ unknown
   ├─ ingester
   ├─ store-gateway
   └─ ingester-and-store-gateway
```
and when a query is dequeued from this tree, it walks the tree down to, e.g., tenant-1 and checks every subtree for a dequeuable item, only checking the next tenant if it finds nothing to dequeue. This means that it is possible for slow requests from one tenant to eventually back up the entire request queue if a query component is experiencing degradation. 

**Example:** store-gateways are having a problem, or are working on expensive queries. Ingesters are fine, so we'd like to keep executing a steady stream of ingester-only queries. With this tree, though, once we dequeue to a tenant, each query component subtree will be checked until a dequeue-able query is found, regardless of whether it is an ingester-only query, which means we can't guarantee that steady stream of queries on healthy components.

With these changes and `query-scheduler.prioritize-query-components` enabled, the tree would instead be constructed as:
```
root
├─ unknown
│  ├─ tenant-1
│  ├─ ...
│  └─ tenant-n
├─ ingester
│  ├─ tenant-1
│  ├─ ...
│  └─ tenant-n
...
└─ store-gateway
   ├─ tenant-1
   ├─ ...
   └─ tenant-n
```
Thus, in the scenario of the same example above, once we get to the ingester-only subtree, we will dequeue a query for any tenant in the fairness order which _has an ingester-only query queued_. 

It will still be possible to get into the scenario where all queriers are busy with degraded-component queries, since the `root` node is still only round-robining between components. @francoposa is working on a change to improve the component subtree selection.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/scheduler/queue/queue_test.go
```diff
@@ -844,9 +855,20 @@ func TestRequestQueue_tryDispatchRequestToQuerier_ShouldReEnqueueAfterFailedSend
 				require.NoError(t, queueBroker.enqueueRequestBack(&tr, tenantMaxQueriers))
 				require.False(t, tq.getNode(QueuePath{"tenant-1"}).IsEmpty())
 			} else if itq, ok := queueBroker.tree.(*MultiQueuingAlgorithmTreeQueue); ok {
-				require.Nil(t, itq.GetNode(QueuePath{"tenant-1"}))
+				var path QueuePath
+				if queueDim == nil {
+					queueDim = []string{unknownQueueDimension}
+				}
+				switch queueBroker.prioritizeQueryComponents {
+				case true:
```
francoposa, line 863: let's do an if-else on the boolean, I can't find examples of using a switch on a boolean anywhere else

pkg/scheduler/queue/tenant_queues_test.go
```diff
@@ -254,17 +252,38 @@ func TestQueuesRespectMaxTenantQueueSizeWithSubQueues(t *testing.T) {
...
 				var checkPath QueuePath
 				if v == nil {
-					checkPath = append(QueuePath{"tenant-1"}, unknownQueueDimension)
-				} else {
+					v = []string{unknownQueueDimension}
+				}
+				switch qb.prioritizeQueryComponents {
+				case true:
+					checkPath = append(append(checkPath, v...), "tenant-1")
+				default:
```
francoposa, line 283: nits:
1. should we just replace the `0: nil,` from `additionalQueueDimensions` variable above with the unknown so we don't need this case? we only use unknown now anyway
2. switch on bool, let's do if-else

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 * [ENHANCEMENT] Expose a new `s3.session-token` configuration option to enable using temporary security credentials. #8952
 * [ENHANCEMENT] Make `-query-frontend.additional-query-queue-dimensions-enabled` and `-query-scheduler.additional-query-queue-dimensions-enabled` non-operational flags in preparation for removal. #8984
 * [ENHANCEMENT] Add a new ingester endpoint to prepare instances to downscale. #8956
+* [ENHANCEMENT] Query-scheduler: Add `query-scheduler.prioritize-query-components` to allow dequeuing from specific query components. #9016
```
francoposa, line 70: I might change this a bit to be more specific like the flag description - on its own doesn't really allow anyone or anything to choose to dequeue for specific query components

francoposa (approved):
LGTM save for some minor nits & I will leave more of the flag doc wording for docs team to approve.

I kept looking for where the flag was actually used in the production code in this PR, but it was already in `makeQueuePath` :)

pkg/scheduler/queue/tenant_queues_test.go
chencs: 1. We will still _receive_ requests that have nil `SchedulerRequest.AdditionalQueueDimensions`, so I think I prefer to keep it as nil to cover that case.
2. Fair enough!

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1722,6 +1722,14 @@ The `query_scheduler` block configures the query-scheduler.
 # CLI flag: -query-scheduler.use-multi-algorithm-query-queue
 [use_multi_algorithm_query_queue: <boolean> | default = false]
 
+# (experimental) When enabled, the query scheduler will primarily prioritize
+# dequeuing fairly from queue components, and secondarily prioritize dequeuing
+# fairly across tenants. When disabled, tenant fairness is primarily
+# prioritized. `query-scheduler.use-multi-algorithm-query-queue` must be enabled
+# in order to use this flag.
```
tacole02, line 1729:
```suggestion
# (experimental) When enabled, the query scheduler primarily prioritizes
# dequeuing fairly from queue components and secondarily prioritizes dequeuing
# fairly across tenants. When disabled, the query scheduler primarily
# prioritizes tenant fairness. You must enable the `query-scheduler.use-multi-algorithm-query-queue` setting to use this flag.
```

tacole02: We avoid using "will" in docs and instead always use the present tense.

tacole02 (approved):
Thanks for updating this doc!

PR #9017, opened 2024-08-16
Mimir query engine: add support for `vector()` and `scalar()`

charleskorn:
#### What this PR does

This PR builds on https://github.com/grafana/mimir/pull/8995 to add support for the `vector()` and `scalar()` functions in MQE.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/operators/instant_vector_to_scalar.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	for t := i.Start; t <= i.End; t += i.Interval {
+		output = append(output, promql.FPoint{
+			T: t,
+			F: math.NaN(),
+		})
+	}
+
+	for seriesIdx := 0; seriesIdx < seriesCount; seriesIdx++ {
```
jhesketh, line 74: Why do this instead of looping until we see EOS?

jhesketh (commented):
lgtm, just one thing I didn't quite follow

pkg/streamingpromql/operators/instant_vector_to_scalar.go
charleskorn: I did it this way so we could detect a situation where the inner operator said it would give X series but only returned Y series.

Granted, that should never happen, but if it did, that'd be weird and we'd want to fail the query rather than silently continue.

jhesketh (approved):

PR #9018, opened 2024-08-16
Mimir query engine: add tool to find disabled but supported test cases

charleskorn:
#### What this PR does

This PR adds a tool to find upstream test cases that are currently disabled but are now supported by MQE, and enables those test cases that are supported.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


jhesketh (commented):
Are we perhaps at a point where we can just enable all the tests but not fail on ones that emit `NotSupportedError`?

Granted it's possible that we might miss something if we're inadvertently returning NotSupportedError but if we only check that on upstream test cases hopefully we're not marking anything as not supported incorrectly and it would be detected by our test cases if that were to happen.

I feel like the risk here is low enough, and we add a TODO to remove the disabled test once we reach parity. Or once we're very close to parity we check for a specific disabled reason.

charleskorn:
> Are we perhaps at a point where we can just enable all the tests but not fail on ones that emit `NotSupportedError`?

Unfortunately `promqltest` doesn't support filtering for expected errors like these, and I'm not sure Prometheus would be keen on adding support for this given it's not relevant for Prometheus, and won't be relevant for us once we reach feature parity.

I also prefer the explicitness of enabling and disabling tests to ensure we don't accidentally start rejecting queries as unsupported.

It also allows us to disable tests that are supported but are expected to return incorrect results for features under development.

jhesketh (approved):
Fair enough. This is already done, no point messing around with other options (eg exporting `test` from prometheus so we can write our own `runTest`).

PR #9019, opened 2024-08-16
Mimir query engine: correctly handle functions that produce multiple series with the same labels

charleskorn:
#### What this PR does

This PR fixes MQE's handling of queries that produce results with multiple series with the same labels.

This is currently only possible in MQE in places where we drop the metric name from the input series: in all other cases it's not possible to return multiple series with the same labels. (Functions like `label_replace` and unary negation will need similar checks when we implement those features.)

See https://github.com/grafana/mimir/pull/9019#issuecomment-2295413980 for more explanation of the logic in this PR.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/functions/common_test.go
```diff
@@ -30,6 +30,16 @@ func TestDropSeriesName(t *testing.T) {
 	require.Equal(t, expected, modifiedMetadata)
 }
 
+func TestDropSeriesName_DuplicateSeries(t *testing.T) {
```
jhesketh, line 33: not `SadPath`? :-(

pkg/streamingpromql/types/data.go
```diff
@@ -102,3 +105,36 @@ type ScalarData struct {
...
+	case 2:
+		if metadata[0].Labels.Hash() == metadata[1].Labels.Hash() {
+			return true, metadata[0].Labels
+		}
+
+		return false, labels.EmptyLabels()
+
+	default:
+		// Note that there's a risk here that we incorrectly flag two series as duplicates when in reality they're different
+		// due to hash collisions.
```
jhesketh, line 122: (nit) this applies to case 2 above as well. Perhaps move this there?

pkg/streamingpromql/functions/common.go
```diff
@@ -22,6 +24,11 @@ func DropSeriesName(seriesMetadata []types.SeriesMetadata, _ *limiting.MemoryCon
 		seriesMetadata[i].Labels = seriesMetadata[i].Labels.DropMetricName()
 	}
 
+	if hasDuplicate, example := types.HasDuplicateSeries(seriesMetadata); hasDuplicate {
```
jhesketh, line 27: Should this check not be done in query.go when we get all of the series names?
It would help catch other operators incorrectly handling series labels such as aggregations. Although that's more likely a bug and unnecessary overhead. Or perhaps we do it in query.go in `convertFunctionCallToOperator` so that we apply it to all functions that can manipulate labels. Otherwise if we manipulate labels differently in a future function we may forget to apply this check.

jhesketh (commented):
lgtm, just one question inline

charleskorn:
Thinking about this more, I realised the solution isn't quite as simple as this, so I'm going to put this PR back to a draft until I can implement this properly:

It's not enough to check for duplicate series, what we need to do is check for conflicting samples for the same output series, and merge the results from input series that map to the same output series.

eg. imagine we're running the query `ceil({__name__=~"some_metric|some_other_metric"})`.

If the input data is:

```
some_metric        1 2 _ _
some_other_metric  _ _ 3 4
```

then we should not produce an error and should merge the result to a single series:

```
{} 1 2 3 4
```

However, the implementation here will cause the query to return an error. 

And, without this PR, we'll currently return this result, which is also invalid:

```
{} 1 2 _ _
{} _ _ 3 4
```

The only time we should return an error is if the input series map to the same output series and have a conflicting sample, for example:

```
some_metric       1 2 _ _
some_other_metric 1 _ _ _
```

pkg/streamingpromql/functions/common.go
charleskorn: >Should this check not be done in query.go when we get all of the series names? It would help catch other operators incorrectly handling series labels such as aggregations. Although that's more likely a bug and unnecessary overhead.

This wouldn't be a thorough-enough check - duplicate series are not permitted in any intermediate stage, not just the end result. And yep, I think it would be unnecessary overhead and a bug.

> Or perhaps we do it in query.go in `convertFunctionCallToOperator` so that we apply it to all functions that can manipulate labels. Otherwise if we manipulate labels differently in a future function we may forget to apply this check.

I'm thinking something more along these lines makes sense, especially given [this realisation](https://github.com/grafana/mimir/pull/9019#issuecomment-2295413980) over the weekend. 

Given we need similar checking and merging logic in many places[^1], I'm wondering if this should be done in an operator that wraps operators that manipulate labels. This wrapping operator would be responsible for:
- merging the results from different input series that map to the same output series
- rejecting queries where results from different input series conflict at a single timestamp

We can then apply this operator whenever the expression manipulates labels.

The other option would be to add this logic to `FunctionOverInstantVector` and `FunctionOverRangeVector`, and possibly other places in the future, but I think that will make them more complex to reason about and unnecessarily duplicate some logic that can live in one place.

What do you think?

[^1]: all the functions over instant vectors and range vectors, `label_replace`, `label_join`, unary negation, possibly others?

jhesketh: I think a wrapper operator makes the most sense. I think the only reason not to would be if we have an opportunity in a downstream operator to detect such a conflict. However, we can likely do both and if we have an opportunity for such an optimisation we can percolate the error back up.

pkg/streamingpromql/functions.go
```diff
@@ -37,7 +37,8 @@ type ScalarFunctionOperatorFactory func(
 //   - name: The name of the function
 //   - metadataFunc: The function for handling metadata
 //   - seriesDataFunc: The function to handle series data
-func SingleInputVectorFunctionOperatorFactory(name string, metadataFunc functions.SeriesMetadataFunction, seriesDataFunc functions.InstantVectorFunction) InstantVectorFunctionOperatorFactory {
+//   - needsSeriesDeduplication: Set to true if metadataFunc may produce multiple series with the same labels and therefore deduplication is required
```
charleskorn, line 40: Note to reviewers: we've previously talked about extracting a type like `FunctionOverRangeVector` for functions over instant vectors, and I think the addition of this argument is another reason to do this, but I don't want to do this in this PR in the interests of keeping it to a manageable size.

pkg/streamingpromql/types/data.go
charleskorn: Done in 0cc5310

pkg/streamingpromql/operators/series_merging_test.go
```diff
@@ -0,0 +1,763 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func TestMergeSeries(t *testing.T) {
+	testCases := map[string]struct {
+		input               []types.InstantVectorSeriesData
+		sourceSeriesIndices []int
```
jhesketh, line 19: (nit) would be good to document what these variables represent (or at least sourceSeriesIndices in the context of the test case)

pkg/streamingpromql/operators/deduplicate_and_merge.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/prometheus/prometheus/promql/parser/posrange"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+var errVectorContainsMetricsWithSameLabels = errors.New("vector cannot contain metrics with the same labelset")
+
+type DeduplicateAndMerge struct {
```
jhesketh, line 18: (nit) should this be `DeduplicateAndMergeOperator` since it implements the operator interface?

jhesketh (approved):
Nice work :tada:, it's a great improvement on what we had. LGTM sans nits :-)

pkg/streamingpromql/operators/series_merging_test.go
charleskorn: `sourceSeriesIndices` has the same meaning as the parameter to `MergeSeries`, and this is already documented for `MergeSeries` - is there something special you think we should mention in the context of the tests?

pkg/streamingpromql/operators/deduplicate_and_merge.go
charleskorn: None of the other operators have an `Operator` suffix - eg. `Aggregation`, `FunctionOverInstantVector` etc.

Adding an `Operator` suffix would introduce a stutter when the type is written with the package name, eg. `operators.DeduplicateAndMerge`.

charleskorn:
Given the approval, I'm going to merge this, but happy to address any post-merge feedback in a follow-up PR.

pkg/streamingpromql/operators/deduplicate_and_merge.go
jhesketh: I confused myself with BinaryOperation :-p

pkg/streamingpromql/operators/series_merging_test.go
jhesketh: That's fair enough, although I don't think it would hurt to have it in the tests for clarity on how to write tests, but I'm not worried about it.

PR #9020, opened 2024-08-16
Refactor `pusherConsumer.Consume`

gotjosh:
#### What this PR does

There's a couple of things going on here:

- Make `Consume` part of the public interface, it's a private struct so having a public method is OK in this package.
- Make the unmarshalling and pushing to storage process a bit easier to understand and document it.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -27,15 +27,6 @@ type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
-type parsedRecord struct {
-	*mimirpb.WriteRequest
-	// ctx holds the tracing baggage for this record/request.
-	ctx      context.Context
-	tenantID string
-	err      error
-	index    int
-}
-
 type PusherCloser interface {
 	Pusher
 	Close() []error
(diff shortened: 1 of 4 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -103,15 +103,57 @@ func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsu
...
+					WriteRequest: &mimirpb.WriteRequest{},
+					index:        index,
+				}
+
+				// We don't free the WriteRequest slices because they are being freed by a level below.
+				err := parsed.WriteRequest.Unmarshal(r.content)
+				if err != nil {
+					parsed.err = fmt.Errorf("parsing ingest consumer write request: %w", err)
+				}
+				recordsChannel <- parsed
```
dimitarvdimitrov, line 137: shouldn't this also look at the `<-unmarshalCtx.Done():`? The context may be cancelled between checking the select on line 121 and us unmarshalling the request and trying to send it on line 137

pkg/storage/ingest/pusher.go
```diff
@@ -103,15 +103,57 @@ func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsu
...
+	for r := range recordsChannel {
+		if r.err != nil {
+			level.Error(c.logger).Log("msg", "failed to parse write request; skipping", "err", r.err)
+			continue
+		}
+
+		// If we get an error at any point, we need to stop processing the records. They will be retried at some point.
+		err := c.pushToStorage(r.ctx, r.tenantID, r.WriteRequest)
+		if err != nil {
+			cancel() // Stop the unmarshalling goroutine.
```
dimitarvdimitrov, line 151: there's also a `defer cancel()`, do we need both?

pkg/storage/ingest/pusher.go
```diff
@@ -30,10 +29,11 @@ type Pusher interface {
 
 type parsedRecord struct {
```
dimitarvdimitrov, line 30: `parsedRecord` is only only used in `pusherConsumer.Consume`. You can move it into that method if you want to reduce its scope. Not sure if you like that kind of type definitions

dimitarvdimitrov (commented):
i think there's a locking problem with sending to parsedRecords, but otherwise looks good.

pkg/storage/ingest/pusher.go
```diff
@@ -103,15 +103,57 @@ func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsu
...
+
+				// We don't free the WriteRequest slices because they are being freed by a level below.
+				err := parsed.WriteRequest.Unmarshal(r.content)
+				if err != nil {
+					parsed.err = fmt.Errorf("parsing ingest consumer write request: %w", err)
+				}
+				recordsChannel <- parsed
+			}
+		}
+	}(ctx, records)
```
dimitarvdimitrov, line 140: we pass `recordsChannel` by closure and `ctx` and `records` by parameter. Should we do it just one way?

pkg/storage/ingest/pusher.go
gotjosh: 1. Good catch!

pkg/storage/ingest/pusher.go
gotjosh: I've removed it in favour of manually calling `cancel` I think it has something to do with how defer works and our go verify leak where I get inconsistent results without calling cancel as we're done (and also the linter complains about it)

pkg/storage/ingest/pusher.go
gotjosh: I moved it.

gotjosh: Done in https://github.com/grafana/mimir/pull/9020/commits/9c310be3c4388cb3838034dfcf76547087a73359

pkg/storage/ingest/pusher.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9020/commits/9c310be3c4388cb3838034dfcf76547087a73359

pkg/storage/ingest/pusher.go
```diff
@@ -103,15 +94,75 @@ func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsu
...
+				parsed.err = fmt.Errorf("parsing ingest consumer write request: %w", err)
+			}
 
-	recC := make(chan parsedRecord)
+			// Now that we're done, check again before we send it to the channel.
+			select {
+			case <-unmarshalCtx.Done():
+				return
+			default:
+				ch <- parsed
```
dimitarvdimitrov, line 145: this still allows a race. You need to have ch <- parsed as a select case

gotjosh: ohhh, you're right - thanks!

dimitarvdimitrov (approved):

PR #9022, opened 2024-08-16
kafka replay speed: make error handling more explicit

dimitarvdimitrov:
### What this PR does

This is a small cleanup to make error handling a bit more explicit. The basic idea is that we take different actions based on the response:

1. Back off and wait a bit
2. Adjust the range of records we want to fetch
3. (TODO) refresh broker metadata

In each action we look at the error we received and decide.

#### Errors to handle

This is what is listed in the [franz-go package](https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kmsg/generated.go#L4705-L4745) for fetch responses. Next to each is how I think we should handle each. franz-go has more explanations for what each error means _I couldn't find official kafka docs which list this_.

 
<details> <summary>All errors we have to handle </summary>

```
// OffsetOutOfRange - already handled - modify fetchWant and backoff if we're requesting after end; or abort fetchWant if we're fetching before the LSO - no log
// TopicAuthorizationFailed - log error and backoff
// UnknownTopicOrPartition - log error and backoff
// UnsupportedCompressionType - this shouldn't happen - only happens when the request version was under 10, but we always use 13 - log error and backoff - we can't afford to lose records
// UnsupportedVersion - in this case our client is too old - log error and backoff - can't afford to lose records
// KafkaStorageError - it's a server error, log warning and backoff
// UnknownTopicID - log error and backoff
// OffsetMovedToTieredStorage - this should be only intra-broker error and we shouldn't get it. log error and backoff
// NotLeaderForPartition - log warning and backoff; our metadata is likely outdated, so we can retry
// ReplicaNotAvailable - happens when the replica hasn't had the time to catch up with the topic - log warning and backoff
// UnknownLeaderEpoch - our metadata is out of date, we should refresh and try again with a leader who is up to date - log error and backoff // TODO this current doesn't happen
// FencedLeaderEpoch - our metadata is out of date, we should refresh and try again with a leader who is up to date - log error and backoff // TODO this current doesn't happen
```

</details>

### Alternative

 I wasn't sure if that's the best approach or whether it's not something like this which shows all the actions we take for a given error in one place. Open to feedback, but also don't want to bikeshed :smile: I spent way too long trying out different ways and neither looked perfect.

```go
func handleError(f fetchResult) {
  switch f.Err {
  case OffsetOutOfRange:
    w = adjustWant()
    if consumingAfterEnd() {
      waitBackoff()
    }
  case TopicAuthorizationFailed:
    waitBackoff()
  case UnsupportedCompressionType:
    waitBackoff()
  case ... // list all the possible errors, so we're very explicit what happens on each
  default:
     panic("") // or a very loud error
  }
}
```

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -980,11 +980,12 @@ func (r *concurrentFetchers) runFetcher(ctx context.Context, fetchersWg *sync.Wa
 
 	for w := range wants {
 		for attempt := 0; errBackoff.Ongoing() && w.endOffset > w.startOffset; attempt++ {
-			f := r.fetchSingle(ctx, w, log.With(logger, "attempt", attempt))
-
-			w = adjustedFetchWantFromErr(f, w)
+			attemptLogger := log.With(logger, "attempt", attempt)
+			f := r.fetchSingle(ctx, w, attemptLogger)
+			if f.Err != nil {
+				w = handleKafkaFetchErr(f, w, errBackoff, newRecordsProducedBackoff, attemptLogger)
+			}
 			if len(f.Records) == 0 {
-				backoffFromKafkaFetchErr(f, w, errBackoff, newRecordsProducedBackoff)
 				// TODO proper handling of NotLeaderForPartition ReplicaNotAvailable UnknownLeaderEpoch FencedLeaderEpoch; we need to be able to select the right broker to handle those
 				continue
 			}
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov:
> All errors we have to handle

franz-go will already retry the requests with retriable errors. Some of these errors are considered not retriable by Kafka ([according to franz-go](https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kerr/kerr.go#L76-L192)), but we retry them anyway:

- OffsetOutOfRange: we adjust our request if necessary. I think it's ok to retry (we shouldn't skip offset ranges, so we can't just abort)
- TopicAuthorizationFailed
- UnsupportedCompressionType
- UnsupportedVersion
- OffsetMovedToTieredStorage

For the first I think we should still retry. For the latter 4 we can also try to fail the service - i.e. exit code 1 and some error message. **I _think_ that's actually a better idea because it makes a potential problem easier to spot. I'm working on a patch to do that and will push it here.** From what I can tell franz-go just returns these errors to the client application, so there's no special handling (what even can we do?)

pkg/storage/ingest/reader.go
```diff
@@ -1076,22 +1070,84 @@ func (r *concurrentFetchers) runFetchers(ctx context.Context, startOffset int64)
 	}
 }
 
-func logCompletedFetch(logger log.Logger, f fetchResult, fetchStartTime time.Time, attempt int, w fetchWant) {
+// handleKafkaFetchErr handles all the errors listed in the franz-go documentation as possible errors when fetching records.
+// For most of them we just apply a backoff. They are listed here so we can be explicit in what we're handling and how.
+// handleKafkaFetchErr returns an adjusted fetchWant in case the error indicated we were consuming not yet produced records or records already deleted due to retention.
```
gotjosh, line 1075:
```suggestion
// it may also return an adjusted fetchWant in case the error indicated, we were consuming not yet produced records or records already deleted due to retention.
```

pkg/storage/ingest/reader.go
```diff
@@ -1076,22 +1070,84 @@ func (r *concurrentFetchers) runFetchers(ctx context.Context, startOffset int64)
...
+func handleKafkaFetchErr(fr fetchResult, fw fetchWant, shortBackoff, longBackoff *backoff.Backoff, logger log.Logger) fetchWant {
+	err := fr.Err
+	switch {
+	case err == nil:
+	case errors.Is(err, kerr.OffsetOutOfRange):
```
gotjosh, line 1080: I'm not a fan of all this logic here - but I don't have a better suggestion so I'm good with it

gotjosh (approved):
LGTM

pkg/storage/ingest/reader.go
dimitarvdimitrov: you mean the one for `OffsetOutOfRange` specifically or the whole switch statement?

gotjosh: OffsetOutOfRange specifically.

dimitarvdimitrov:
converting to draft because I realized that we should handle some errors such as ReplicaNotAvailable or NotLeaderForPartition by refreshing metadata. While I'm working on this, I'll keep the PR in a draft

dimitarvdimitrov:
chatted with @gotjosh and agreed to add a test for handleKafkaFetchErr and merge as-is. The PR is long enough and old enough. The other unhandled errors ReplicaNotAvailable or NotLeaderForPartition will be handled in another PR.

I pushed `TestHandleKafkaFetchErr_OffsetOutOfRange` and will merge on green.

PR #9026, opened 2024-08-16
Jsonnet: allow to easily override Kafka client ID on a per-component basis

pracucci:
#### What this PR does

In this PR I propose a jsonnet refactoring to allow to easily override the Kafka client ID on a per-component basis. The idea is to configure the client ID as key-value pairs, which then gets serialised in a comma-separated string (e.g. `{keyA: 'valueA', keyB: 'valueB'}` gets serialised to `keyA=valueA,keyB=valueB`).

I've introduce config fields to set the Kafka client ID for specific Mimir components (distributor, ingester, query-frontend, ruler). By default, they inherit the default producer or consumer settings, but client ID values can be overriden in the per-component config field. Similarly to CLI args, a `null` value in the override removes that setting from the client ID.

The config fields are suffixed with `_settings`. They work similarly to `_args` ones, but since they're not CLI flags I preferred to name them differently.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
As discussed privately, I'm not sure what use the KV-settings bring for an arbitrary Kafka-compatible broker (e.g. Kafka, Confluent, MSK, Redpanda, etc) 'cause, generally, the client-id is just a string. There are indeed Kafka-compatible brokers, that add special meaning to the client-id, for which the changes can simplify the configuration.

Yet, I don't feel strong about the above. The changes look fine for me 🔥

PR #9028, opened 2024-08-16
Cache non-transient error responses from the query-frontend

### LINKED ISSUE #2676: Query-frontend should cache errors resulting from data

Suppose you have a dashboard containing a large join query which results in a many-to-many error.
This dashboard is unduly expensive to run - the entire data range is re-fetched on every dashboard update but each time the same error is returned by Mimir.

Query-frontend should cache data errors, similarly to how it caches empty results today.
Note that join errors can come and go - if there is only 1 row returned then you won't get a many-to-many error - so we need to cache each query result in time extents, just as with sample results or empty results.

Not all query errors result from data; for example exceeding the chunk limit is something that will clear if the administrator raises the limit.  Perhaps we can get round needing to know the difference by a relatively short cache expiry on errors.

Nothing.

### LINKED ISSUE #7340: Inhibit executing a query asap for 5 minutes if the same query was rejected due to busting limits

If a query loads too much data (chunks/samples/series) and we rejected, nothing stops the user from issuing the same query again right away. This is not a good strategy as we can be reasonably sure that we cannot execute the query a second time around and generate useless load.

Reject repeated queries as soon as possible for 5 minutes if the query failed due to limits and has the same query and same/similar time range.

N/A

#### Additional context

N/A

56quarters:
#### What this PR does

Create a new query-frontend middleware that caches errors returned by queries if they are non-transient and will fail again if executed again. This allows us to save work when running a query that hits, e.g., a limit error: running the query again will not help and is a waste of work.

Fixes #2676

Fixes #7340

#### Checklist

- [X] Tests updated.
- [X] Documentation added.


56quarters:
Some example of errors that would be cached pulled from the logs of an internal cluster over the last seven days. Each of the errors falls into one of two categories: bad queries or limits. These are the intended type of errors that would be cached by this feature.

Examples of `errorType = "execution"` that result in HTTP 422:

* `many-to-many matching not allowed: matching labels must be unique on one side`. Bad query
* `multiple matches for labels: many-to-one matching must be explicit (group_left/group_right)`. Bad query
* `multiple matches for labels: grouping labels must ensure unique matches`. Bad query
* `expanding series: invalid destination label name in label_join(): `. Bad query
* `expanding series: invalid source label name in label_join(): (.*)`. Bad query
* `Can't query aggregated metric X without aggregation because...`. Bad query
* `query processing would load too many samples into memory in query execution`. Limit
* `expanding series: the query exceeded the maximum number of chunks (limit: N chunks) (err-mimir-max-chunks-per-query)`. Limit
* `expanding series: the query time range exceeds the limit (query length: X, limit: Y) (err-mimir-max-query-length)`. Limit

Examples of `errorType = "bad_data"` that result in HTTP 400"

* `invalid parameter \"query\": 1:135: parse error: ..."`. Bad query
* `invalid parameter \"query\": invalid expression type \"range vector\" for range query, must be Scalar or instant Vector`. Bad query
* `the total query time range exceeds the limit (query length: X, limit: Y) (err-mimir-max-total-query-length)`. Limit
* `exceeded maximum resolution of 11,000 points per timeseries. Try decreasing the query resolution`. Limit

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -339,6 +340,15 @@ func newQueryMiddlewares(
 		newStepAlignMiddleware(limits, log, registerer),
 	)
 
+	if cfg.CacheResults && cfg.CacheErrors {
+		// TODO: Use a real TTL
+		queryRangeMiddleware = append(
```
dimitarvdimitrov, line 345: an additional middleware increases the cache calls - now we look up one more key from the cache sequentially. I'm mostly thinking about increasing cache lookup RPCs and increasing latency because we do this cache lookup before we proceed with the request. Did you consider including this logic in the current split&cache middleware?

pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -0,0 +1,182 @@
...
+func (e *errorCachingHandler) isCacheable(apiErr *apierror.APIError, req MetricsQueryRequest, tenantIDs []string) (bool, string) {
+	if apiErr.Type != apierror.TypeBadData && apiErr.Type != apierror.TypeExec {
+		return false, reasonNotCacheableError
+	}
+
+	maxCacheFreshness := validation.MaxDurationPerTenant(tenantIDs, e.limits.MaxCacheFreshness)
+	maxCacheTime := int64(model.Now().Add(-maxCacheFreshness))
+	cacheUnalignedRequests := validation.AllTrueBooleansPerTenant(tenantIDs, e.limits.ResultsCacheForUnalignedQueryEnabled)
+
+	return isRequestCachable(req, maxCacheTime, cacheUnalignedRequests, e.logger)
```
dimitarvdimitrov, line 181: why do we need to respect step alignment and max cache times? The way I was thinking about this is that we want to prevent the exact same query from running again because it will, almost certainly, fail again.

pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		level.Warn(e.logger).Log("msg", "unable to marshal cached error", "err", err)
+		return
+	}
+
+	e.cache.SetAsync(hashedKey, bytes, e.ttl)
+}
+
+func (e *errorCachingHandler) cacheKey(tenantID string, r MetricsQueryRequest) string {
+	return fmt.Sprintf("EC:%s:%s:%d:%d:%d", tenantID, r.GetQuery(), r.GetStart(), r.GetEnd(), r.GetStep())
```
dimitarvdimitrov, line 169: the design doc ([internal link](https://docs.google.com/document/d/1SO74280ZziBrzfzca-ZUhzkh93bI-ihEmg01TsNwIhA/edit?tab=t.0#heading=h.5sybau7waq2q)) talked about grouping the queries into time buckets. Are you thinking of doing something similar too?

pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (e *errorCachingHandler) isCacheable(apiErr *apierror.APIError, req MetricsQueryRequest, tenantIDs []string) (bool, string) {
+	if apiErr.Type != apierror.TypeBadData && apiErr.Type != apierror.TypeExec {
```
dimitarvdimitrov, line 173: I saw we also have HTTP 413 responses. I couldn't find what they're used for though. Should we cache them too? https://github.com/grafana/mimir/blob/e78426575885594438de073d8221f544b92add62/pkg/frontend/querymiddleware/codec.go#L724-L725

pkg/frontend/querymiddleware/roundtrip.go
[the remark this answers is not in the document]
56quarters: I'll see if I can adapt it to label queries.

pkg/frontend/querymiddleware/model.proto
```diff
@@ -43,6 +43,12 @@ message SampleStream {
   repeated cortexpb.FloatHistogramPair histograms = 3 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "histograms"];
 }
 
+message CachedError {
+  string key = 1 [(gogoproto.jsontag) = "key"];
+  string type = 2 [(gogoproto.jsontag) = "errorType"];
+  string message = 3 [(gogoproto.jsontag) = "error"];
```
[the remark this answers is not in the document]
56quarters: No reason, I can change this.

pkg/frontend/querymiddleware/roundtrip.go
56quarters: I considered it. I wanted to avoid changes to the split&cache middleware because it's quite complicated and very focused on caching results. Adding caching of errors to it would add more complexity. 

I also wanted to keep the error caching _before_ queries are split so that we have the best chance of avoiding work when any part of a query is going to fail. We would have do a lot more work if each individual split or sharded query can fail.

I'm not particularly worried about the latency added here for each request because it's only a single cache lookup which is a few milliseconds on hit or miss.

pkg/frontend/querymiddleware/error_caching.go
56quarters: Using a bucket is less precise than start/end/step. I'm not sure if that's an appropriate thing to do since it could make a difference with respect to hitting limits.

pkg/frontend/querymiddleware/error_caching.go
56quarters: I think that's reasonable. I'll take a look and try to find some examples.

pkg/frontend/querymiddleware/error_caching.go
dimitarvdimitrov: exact matching looked more intuitive to me too. I think I prefer it, just wanted to check with you

pkg/frontend/querymiddleware/roundtrip.go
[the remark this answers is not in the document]
56quarters: Currently, label queries don't use middlewares: only range queries, instant queries, and remote reads. In order to get something in place and evaluated quicker I'd rather limit this middleware to only those types of queries. I can open an issue to apply the same treatment to label queries if this turns out to be useful.

pkg/frontend/querymiddleware/error_caching.go
56quarters: I don't have a particularly good reason for this. It felt correct to use the same criteria for caching errors as we use for caching requests but I can see why that wouldn't be desirable (we keep getting hit by expensive failing queries and don't have any recourse). Do you think we should ignore step alignment here? And do you think we should respect the max cache freshness setting here?

pkg/frontend/querymiddleware/roundtrip.go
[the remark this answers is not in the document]
dimitarvdimitrov: I remember @francoposa had some plans to change the label queries too. Franco, is this still planned or did you drop it?

But yeah, sounds good to just focus on those 3 for now.

dimitarvdimitrov: ok, I guess if we measure these latencies it will be ok and we'd know if it's too much. Maybe we can stack them in the cache lookup panel in the reads dashboard?
Screenshot 2024-08-20 at 17 20 21

pkg/frontend/querymiddleware/error_caching.go
dimitarvdimitrov: I think it's ok to ignore them.

For query results, they make sense because the cached value may become outdated as new samples (possibly out of order) come in. From the errors you listed I don't see any of them becoming less likely as new data comes in.

pkg/frontend/querymiddleware/roundtrip.go
56quarters: Testing in a dev cell, this middleware ended up taking about 500us on cache hit or miss. I'll follow up this PR with some changes to the reads/queries dashboards if this turns out to be useful.

tacole02 (approved):
Docs look good!

pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -0,0 +1,182 @@
...
+		ttl := validation.MinDurationPerTenant(tenantIDs, e.limits.ResultsCacheTTLForErrors)
+		e.storeErrorToCache(key, hashedKey, ttl, apiErr, spanLog)
+	}
+
+	return res, err
+}
+
+func (e *errorCachingHandler) loadErrorFromCache(ctx context.Context, key, hashedKey string, spanLog *spanlogger.SpanLogger) *apierror.APIError {
+	res := e.cache.GetMulti(ctx, []string{hashedKey})
+	if cached, ok := res[hashedKey]; ok {
```
dimitarvdimitrov, line 132: nit to invert this: `if !ok { return nil }`

pkg/frontend/querymiddleware/error_caching_test.go
```diff
@@ -0,0 +1,221 @@
...
+		require.Equal(t, 0, c.CountFetchCalls())
+		require.Equal(t, 0, c.CountStoreCalls())
+	})
+
+	t.Run("cache hit", func(t *testing.T) {
+		c := cache.NewInstrumentedMockCache()
+
+		innerRes := newEmptyPrometheusResponse()
+		inner := &mockHandler{}
+		inner.On("Do", mock.Anything, mock.Anything).Return(innerRes, nil)
```
dimitarvdimitrov, line 81: inner should never be called, right? So we can also remove this setup call. Otherwise a wrong behaviour of the cache can pass the test

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -263,13 +264,12 @@ func isRequestCachable(req MetricsQueryRequest, maxCacheTime int64, cacheUnalign
 	return true, ""
 }
 
-// isResponseCachable says whether the response should be cached or not.
-func isResponseCachable(r Response, logger log.Logger) bool {
-	headerValues := getHeaderValuesWithName(r, cacheControlHeader)
-	for _, v := range headerValues {
-		if v == noStoreValue {
-			level.Debug(logger).Log("msg", fmt.Sprintf("%s header in response is equal to %s, not caching the response", cacheControlHeader, noStoreValue))
```
dimitarvdimitrov, line 271: why the change to remove `logger`?

dimitarvdimitrov (approved):
Thanks for you patience 🙏 I'm really looking forward to this running in prod

LGTM, I left a few small comments, but I probably don't need to take a look at this again

pkg/frontend/querymiddleware/results_cache.go
56quarters: Just didn't seem like a useful log line on its own since we never run at debug level

dimitarvdimitrov: it always gets added to the span though

PR #9039, opened 2024-08-19
Add Grafana data source with results caching disabled in microservices Docker Compose stack

charleskorn:
#### What this PR does

This PR adds a second Grafana data source with results caching disabled to the microservices Docker Compose stack.

This is useful when debugging incorrect query results.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


development/mimir-microservices-mode/config/datasources.yaml
```diff
@@ -0,0 +1,36 @@
...
+  access: proxy
+  uid: mimir-no-caching
+  orgID: 1
+  url: http://nginx:8080/prometheus
+  jsonData:
+    prometheusType: Mimir
+    exemplarTraceIdDestinations:
+      - name: traceID
+        datasourceUid: jaeger
+    httpHeaderName1: Cache-Control
```
jhesketh, line 27: Does this not need a value?

jhesketh (commented):
change lgtm, but I'm a bit confused about how the header value is set?

development/mimir-microservices-mode/config/datasources.yaml
charleskorn: It is set in `secureJsonData` below.

jhesketh: ah thanks. Don't know how I missed that

jhesketh (approved):

PR #9040, opened 2024-08-19
Add unit test for `chunkIterator.Seek()` when the current batch can satisfy the sought timestamp

charleskorn:
#### What this PR does

This PR is a follow up to https://github.com/grafana/mimir/pull/8992.

In that PR, we noticed that `chunkIterator` had similar logic to `mergeIterator`, but decided to merge the change without adding a test for `chunkIterator`. In this PR, I'm adding a test for that case.

This revealed that while `chunkIterator` would have behaved correctly without the change in #8992, it would unnecessarily skip the optimisation to satisfy the `Seek` from the current batch if the sought timestamp was before the first sample of the current batch.

Note that `Seek` is expected to only move forwards, and I've clarified the docstring for the interface to reflect this.

I haven't added a changelog entry given this PR only adds a test.

Relates to https://github.com/grafana/mimir/pull/8992.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/querier/batch/batch.go
+++ b/pkg/querier/batch/batch.go
@@ -38,7 +38,10 @@ func (c GenericChunk) Iterator(reuse chunk.Iterator) chunk.Iterator {
 
 // iterator iterates over batches.
 type iterator interface {
-	// Seek advances the iterator forward to the sample at or after the given timestamp.
+	// Seek advances the iterator forward to batch containing the sample at or after the given timestamp.
+	// If the current batch contains a sample at or after the given timestamp, then Seek retains the current batch.
+	//
+	// The batch's Index is advanced to point to the sample at or after the given timestamp.
 	Seek(t int64, size int) chunkenc.ValueType
 
 	// Next moves to the next batch.
--- a/pkg/querier/batch/batch.go
+++ b/pkg/querier/batch/batch.go
@@ -38,7 +38,7 @@ func (c GenericChunk) Iterator(reuse chunk.Iterator) chunk.Iterator {
 
 // iterator iterates over batches.
 type iterator interface {
-	// Seek to the batch with sample at (or after) time t.
+	// Seek advances the iterator forward to the sample at or after the given timestamp.
 	Seek(t int64, size int) chunkenc.ValueType
 
 	// Next moves to the next batch.
--- a/pkg/querier/batch/chunk.go
+++ b/pkg/querier/batch/chunk.go
@@ -32,8 +32,6 @@ func (i *chunkIterator) reset(chunk GenericChunk) {
 	i.batch.Index = 0
 }
 
-// Seek advances the iterator forward to the value at or after
-// the given timestamp.
 func (i *chunkIterator) Seek(t int64, size int) chunkenc.ValueType {
 	// We assume seeks only care about a specific window; if this chunk doesn't
 	// contain samples in that window, we can shortcut.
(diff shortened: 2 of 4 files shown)


pkg/querier/batch/batch.go
```diff
@@ -38,7 +38,7 @@ func (c GenericChunk) Iterator(reuse chunk.Iterator) chunk.Iterator {
 
 // iterator iterates over batches.
 type iterator interface {
-	// Seek to the batch with sample at (or after) time t.
+	// Seek advances the iterator forward to the sample at or after the given timestamp.
```
aknuds1, line 41: Doesn't `Seek` advance to a batch, considering the iterator operates at the batch granularity? E.g., `Next` moves to the next batch, not sample. Wondering if the previous wording isn't more descriptive in this respect.

pkg/querier/batch/chunk.go
```diff
@@ -32,8 +32,6 @@ func (i *chunkIterator) reset(chunk GenericChunk) {
 	i.batch.Index = 0
 }
 
-// Seek advances the iterator forward to the value at or after
-// the given timestamp.
```
aknuds1, line 36: What's the reason for entirely dropping the doc comment of this method?

pkg/querier/batch/chunk_test.go
```diff
@@ -222,3 +222,40 @@ func (i *mockIterator) Batch(_ int, valueType chunkenc.ValueType, _ *zeropool.Po
...
+	for _, ts := range chunkTimestamps {
+		app.Append(ts, float64(ts))
+	}
+
+	genericChunk := NewGenericChunk(chunkTimestamps[0], chunkTimestamps[len(chunkTimestamps)-1], func(reuse chunk.Iterator) chunk.Iterator {
+		chk, err := chunk.NewForEncoding(chunk.PrometheusXorChunk)
+		require.NoError(t, err)
+		require.NoError(t, chk.UnmarshalFromBuf(ch.Bytes()))
+
+		// We should never need to reset this iterator, as the Seek call below should be able to be satisfied by the initial batch.
```
aknuds1, line 241: [Nit]
```suggestion
		// We should never need to reset this iterator, as the Seek call below should be satisfiable by the initial batch.
```

aknuds1 (approved):
LGTM, but please see comments.

dimitarvdimitrov (approved):
other than Arve's comments, LGTM

pkg/querier/batch/chunk_test.go
charleskorn: Resolved in 0ee758f

pkg/querier/batch/batch.go
charleskorn: `Seek` does both: it advances to the batch with a sample at or after the given timestamp, and advances within that batch to the sample at or after the given timestamp by updating `batch.Index`. 

I've tried to make this clearer in 0ee758f - how's this now?

pkg/querier/batch/chunk.go
charleskorn: None of the other methods on `chunkIterator` have doc comments, and this comment was potentially confusing considering the slightly different comment on the `iterator` interface. So I opted to remove this comment and make the comment on `iterator` clearer.

charleskorn:
I'm taking the silence since my last changes as approval, but let me know if there's anything further you'd like me to change and I'll do that in a follow up PR

Issue #9009, opened 2024-08-15
Goroutine leak in TestStoreGateway_YYY


narqo:
We've noticed a failure in the CI, that reported a goroutine leak in a couple of tests for `storegateway/indexheader.(*LazyBinaryReader)` ([CI logs](https://github.com/grafana/mimir/actions/runs/10396338913/job/28790161079))

@dimitarvdimitrov that seems related to the recent changes from #8380, right?

```
--- FAIL: TestStoreGateway_BlocksSyncWithDefaultSharding_RingTopologyChangedAfterScaleUp (16.89s)
    leak.go:14: found unexpected goroutines:
        [Goroutine 2040876 in state select, with github.com/grafana/mimir/pkg/storegateway/indexheader.(*LazyBinaryReader).controlLoop on top of the stack:
        github.com/grafana/mimir/pkg/storegateway/indexheader.(*LazyBinaryReader).controlLoop(0xc000e00ea0)
        	/__w/mimir/mimir/pkg/storegateway/indexheader/lazy_binary_reader.go:341 +0x17f
        created by github.com/grafana/mimir/pkg/storegateway/indexheader.NewLazyBinaryReader in goroutine 2040872
        	/__w/mimir/mimir/pkg/storegateway/indexheader/lazy_binary_reader.go:167 +0xc4b
        ]
--- FAIL: TestStoreGateway_ShouldSupportLoadRingTokensFromFile (1.42s)
    leak.go:14: found unexpected goroutines:
        [Goroutine 2040876 in state select, with github.com/grafana/mimir/pkg/storegateway/indexheader.(*LazyBinaryReader).controlLoop on top of the stack:
        github.com/grafana/mimir/pkg/storegateway/indexheader.(*LazyBinaryReader).controlLoop(0xc000e00ea0)
        	/__w/mimir/mimir/pkg/storegateway/indexheader/lazy_binary_reader.go:341 +0x17f
        created by github.com/grafana/mimir/pkg/storegateway/indexheader.NewLazyBinaryReader in goroutine 2040872
        	/__w/mimir/mimir/pkg/storegateway/indexheader/lazy_binary_reader.go:167 +0xc4b
--- FAIL: TestStoreGateway_Series_QueryShardingShouldGuaranteeSeriesShardingConsistencyOverTheTime (0.91s)
    leak.go:14: found unexpected goroutines:
[...]

PR #9048, opened 2024-08-19
store-gateway: gracefully shut down BucketStores

dimitarvdimitrov:
### What this PR does

This PR fixes a bug which resulted in a flaky test. 
#### BucketStores graceful shutdown

BucketStores would get shut down immediately when we receive SIGTERM instead of waiting for ongoing bucket syncs to finish first. This sometimes results in leaving index header readers open before shutting down.

This made some tests flaky. It could possibly mean that `BucketStores` stop `BucketStore` instances before waiting for ongoing queries to finish, but I am not certain of this and couldn't find proof in 2 weeks worth of store-gateway and querier logs. Still this is a bug and we should fix it.

By using `context.Background()` we rely on the shutdown here to take effect instead of detecting the parent context being cancelled

https://github.com/grafana/mimir/blob/fcef902e727c603411614d41bdb5d95223ac539f/pkg/storegateway/gateway.go#L326-L328

This is also the pattern I used for starting and stopping the index reader pool here

https://github.com/grafana/mimir/blob/b3a8b39b5fcf7b327239160229bbb1c13a97240f/pkg/storegateway/bucket.go#L265-L266

### Flaky test

Previously the test from https://github.com/grafana/mimir/issues/9009 failed 3 out of 70 runs (once every ~23 runs). With this fix I ran it 100 times and it didn't fail.

### Which issue(s) this PR fixes or relates to

Fixes https://github.com/grafana/mimir/issues/9009

### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket.go
+++ b/pkg/storegateway/bucket.go
@@ -330,6 +330,9 @@ func (s *BucketStore) syncBlocks(ctx context.Context) error {
 		}
 		select {
 		case <-ctx.Done():
+			close(blockc)
+			wg.Wait()
+			return nil
 		case blockc <- meta:
 		}
 	}
--- a/pkg/storegateway/bucket.go
+++ b/pkg/storegateway/bucket.go
@@ -330,9 +330,6 @@ func (s *BucketStore) syncBlocks(ctx context.Context) error {
 		}
 		select {
 		case <-ctx.Done():
-			close(blockc)
-			wg.Wait()
-			return nil
 		case blockc <- meta:
 		}
 	}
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov:
I realized I had another fix locally that's not committed here. I'll integrate and that and and then mark as ready for review

pkg/storegateway/bucket.go
```diff
@@ -330,6 +330,9 @@ func (s *BucketStore) syncBlocks(ctx context.Context) error {
 		}
 		select {
 		case <-ctx.Done():
+			close(blockc)
+			wg.Wait()
+			return nil
```
narqo, line 335: _Just to double-check, to get the idea of the problem here_ 

You mentioned,

> previously `BucketStore` would not wait for `syncBlocks` to finish before returning

Was this about this chunk? The new version looks like it does the same as what the code right below the `range metas` loop does. So the change here is to bail faster. But the previous code did wait, didn't it?

narqo (approved):
I have one comment/question to clarify. But changes overall look good to me 🔥

pkg/storegateway/bucket.go
dimitarvdimitrov: actually yeah, you're right. We don't need this change
```suggestion
```

PR #9049, opened 2024-08-19
indexheader: abort writing index-header when data transfer from bucket is slow

### LINKED ISSUE #8364: store-gateway: abot object store requests when data transfer is slow
### Background

The store-gateway fetches data from object store to either 

1. fetch index bytes and store them in memory ([code](https://github.com/grafana/mimir/blob/2a0da30ed761a80c84bfd832633fb2e90b5db639/pkg/storegateway/bucket.go#L1927-L1931))
2. fetch chunk bytes and store them in memory ([code](https://github.com/grafana/mimir/blob/2a0da30ed761a80c84bfd832633fb2e90b5db639/pkg/storegateway/bucket.go#L1958))
3. fetch index bytes and store them on disk to construct an index header ([code](https://github.com/grafana/mimir/blob/8b8d822f39a4bb95189ef72cad9d2596f7225575/pkg/storegateway/indexheader/binary_reader.go#L73-L106))

### Problem

The last item usually contains the most data and can sometimes take a very long time due to problems on the object store's side. For example creating this index header took more than 2h30m

```
ts=2024-06-13T08:01:16.282008923Z caller=bucket.go:407 level=error user=*** msg="loading block failed" elapsed=2h34m23.851638646s id=01J07YKACB3901WQW9E1G27ERH err="create index header reader: write index header: copy posting offsets: context canceled"
```

### Proposal

Add a configurable option which puts a lower bound on the data transfer rate from the object store. This means that we will not be waiting for 2h30m to download a 100-200MB file.

This can be enforced in two ways

1. an `io.Reader` implementation which tracks the download rate since the start of the request (or in the last X seconds) and aborts it if the transfer rate fall below a threshold; the reader has a starting grace period during which it doesn't abort the connection
2. set a timeout on `GetRange` requests based on the size of the data we want to read. `GetRange` accepts the length of the range, so computing the timeout is trivial

This threshold should take a low value (256KB/s for example), but still be configurable to not potentially block users with very slow object stores.

narqo:
#### What this PR does

_This one is still a draft, as I wanted to get some feedback on the approach. I'm not too fancy about how this ended up, but it seems the simplest way to solve what we outlined in #8364. Note, I'll be looking at which tests to add here, after we agree on the direction._

Here we update the internals of `indexheader` with the support for estimating how much time the download of an index-header's chunks should take, base on the configured `min-download-throughput`.

cc @dimitarvdimitrov @pracucci 

Fixes #8364

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/indexheader/bucket_reader.go
+++ b/pkg/storegateway/indexheader/bucket_reader.go
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package indexheader
+
+import (
+	"context"
+	"errors"
+	"io"
+	"math"
+	"time"
+
+	"github.com/thanos-io/objstore"
+)
+
+var errLowThroughput = errors.New("throughput is lower than expected minimum")
+
+type bucketRangeReader interface {
+	// GetRange returns a new range reader for the given object name and range.
+	GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error)
+	// Attributes returns information about the specified object.
+	Attributes(ctx context.Context, name string) (objstore.ObjectAttributes, error)
+}
+
+type minThroughputBucketRangeReader struct {
+	bucketRangeReader
+	// Minimum expected throughput in bytes/s.
+	throughput int
+}
+
+func bucketRangeReaderWithEnforcedThroughput(bkt bucketRangeReader, throughput int) bucketRangeReader {
+	return &minThroughputBucketRangeReader{
+		bucketRangeReader: bkt,
+		throughput:        throughput,
+	}
+}
+
+func (m *minThroughputBucketRangeReader) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	ctx, cancel := context.WithTimeoutCause(ctx, m.expectedTime(length), errLowThroughput)
+	defer cancel()
+	return m.bucketRangeReader.GetRange(ctx, name, off, length)
+}
+
+func (m *minThroughputBucketRangeReader) expectedTime(n int64) time.Duration {
+	if m.throughput == 0 {
+		return time.Duration(math.MaxInt)
+	}
+	return time.Duration(n) * time.Second / time.Duration(m.throughput)
+}
--- a/pkg/storegateway/indexheader/header.go
+++ b/pkg/storegateway/indexheader/header.go
@@ -70,12 +70,12 @@ type Config struct {
 	// Controls whether index-header lazy loading is enabled.
 	LazyLoadingEnabled     bool          `yaml:"lazy_loading_enabled" category:"advanced"`
 	LazyLoadingIdleTimeout time.Duration `yaml:"lazy_loading_idle_timeout" category:"advanced"`
-
 	// Maximum index-headers loaded into store-gateway concurrently
 	LazyLoadingConcurrency             int           `yaml:"lazy_loading_concurrency" category:"advanced"`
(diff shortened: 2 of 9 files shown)


dimitarvdimitrov:
sorry for the delay, i'll try to have a look next week

pkg/storegateway/indexheader/bucket_reader.go
```diff
@@ -0,0 +1,48 @@
...
+func bucketRangeReaderWithEnforcedThroughput(bkt bucketRangeReader, throughput int) bucketRangeReader {
+	return &minThroughputBucketRangeReader{
+		bucketRangeReader: bkt,
+		throughput:        throughput,
+	}
+}
+
+func (m *minThroughputBucketRangeReader) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	ctx, cancel := context.WithTimeoutCause(ctx, m.expectedTime(length), errLowThroughput)
+	defer cancel()
```
dimitarvdimitrov, line 39: don't we need the context to survive beyond the scope of this method? The io.ReadCloser I think uses the `ctx` we build

pkg/storegateway/indexheader/reader_pool.go
```diff
@@ -85,12 +89,33 @@ func (p *ReaderPool) NewBinaryReader(ctx context.Context, logger log.Logger, bkt
...
+	}
+
 	readerFactory = func() (Reader, error) {
-		return NewStreamBinaryReader(ctx, logger, bkt, dir, id, postingOffsetsInMemSampling, p.metrics.streamReader, cfg)
+		return NewStreamBinaryReader(ctx, logger, brr, dir, id, postingOffsetsInMemSampling, p.metrics.streamReader, cfg)
 	}
 
 	if p.lazyReaderEnabled {
-		reader, err = NewLazyBinaryReader(ctx, readerFactory, logger, bkt, dir, id, p.metrics.lazyReader, p.onLazyReaderClosed, p.lazyLoadingGate)
+		binPath := filepath.Join(dir, id.String(), block.IndexHeaderFilename)
```
dimitarvdimitrov, line 102: why was this code moved from NewLazyBinaryReader?

pkg/storegateway/indexheader/header.go
```diff
@@ -89,6 +89,7 @@ func (cfg *Config) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.DurationVar(&cfg.LazyLoadingConcurrencyQueueTimeout, prefix+"lazy-loading-concurrency-queue-timeout", 5*time.Second, "Timeout for the queue of index header loads. If the queue is full and the timeout is reached, the load will return an error. 0 means no timeout and the load will wait indefinitely.")
 	f.BoolVar(&cfg.EagerLoadingStartupEnabled, prefix+"eager-loading-startup-enabled", true, "If enabled, store-gateway will periodically persist block IDs of lazy loaded index-headers and load them eagerly during startup. Ignored if index-header lazy loading is disabled.")
 	f.DurationVar(&cfg.EagerLoadingPersistInterval, prefix+"eager-loading-persist-interval", time.Minute, "Interval at which the store-gateway persists block IDs of lazy loaded index-headers. Ignored if index-header eager loading is disabled.")
+	f.IntVar(&cfg.MinDownloadThroughput, prefix+"min-download-throughput", 256*1024, "The floor of the throughput, in bytes/sec, with which store-gateway expected to download index-header.")
```
dimitarvdimitrov, line 92: i'd mention what happens when the throughput fails to be met

dimitarvdimitrov:
overall the approach looks ok, I think it will do the job of aborting early. I left a couple of comments. The one about the context is the most critical I think

PR #9062, opened 2024-08-21
Add encode time to query stats

chencs:
#### What this PR does
Adds a field `encode_time_seconds` to query stats log messages, which records the time it takes to encode a response from the frontend. 

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/querier/stats/stats.go
```diff
@@ -187,6 +187,22 @@ func (s *Stats) LoadQueueTime() time.Duration {
...
+func (s *Stats) LoadEncodeTime() time.Duration {
+	if s == nil {
+		return 0
+	}
+
+	return time.Duration(atomic.LoadInt64((*int64)(&s.EncodeTime)))
+}
+
 // Merge the provided Stats into this one.
 func (s *Stats) Merge(other *Stats) {
```
dimitarvdimitrov, line 207: i think Merge also needs changes

pkg/querier/stats/stats.proto
```diff
@@ -34,4 +34,6 @@ message Stats {
   uint64 estimated_series_count = 8;
   // The sum of durations that the query spent in the queue, before it was handled by querier.
   google.protobuf.Duration queue_time = 9 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
+  // The time spent encoding the query's results for returning from the frontend.
+  google.protobuf.Duration encode_time = 10 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
```
dimitarvdimitrov, line 38: maybe mention that it doesn't include the time queriers spent marshalling the result?

dimitarvdimitrov (commented):
Does this include the time it takes queriers to marshal the results? IIRC they use protobuf to talk to the frontend, but I may be wrong. For the purposes of measuring JSON overhead in the query-frontend I think it's better to only measure at the query-frontend.

Would be nice to also get a changelog entry so we document at least somewhere what this new field means.

chencs:
> Does this include the time it takes queriers to marshal the results? IIRC they use protobuf to talk to the frontend, but I may be wrong. 

Yeah, the queriers use protobuf to talk to the frontend, and this doesn't include serialization time for that. I think that serialization time (among other things) would be included in the total `query_wall_time_seconds`, _and_ I think encoding time is _not_ included in `query_wall_time_seconds`. 

> For the purposes of measuring JSON overhead in the query-frontend I think it's better to only measure at the query-frontend.

Are you suggesting that we record the time somewhere other than the current changes propose? Sorry if I'm misunderstanding this; I'm pretty sure the current changes are only recording at the query-frontend.

dimitarvdimitrov (approved):
> > For the purposes of measuring JSON overhead in the query-frontend I think it's better to only measure at the query-frontend.
> 
> Are you suggesting that we record the time somewhere other than the current changes propose? Sorry if I'm misunderstanding this; I'm pretty sure the current changes are only recording at the query-frontend.

oh no, I was just making sure that the changes in this PR don't include the time it takes to marshal the responses in queriers. I think that's the approach we want.

LGTM

PR #9068, opened 2024-08-21
blockbuilder: backport top-level service

narqo:
#### What this PR does

This is part of #8635; refer to it for more details.

Here's the continuation of what we started in #8978, backporting chunks of the new experimental `blockbuilder` component into main. This PR focusing on top-level `BlockBuilder` service, and the general mechanics of starting the service up.

To make the review simpler here, we intentionally skip how the `BlockBuilder` service consumes data from Kafka, how it interacts with the `TSDBBuilder`, and the related testing. This will be the part of the follow-up PRs (refer to the TODO comments).

cc @codesome 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/kafkautil.go
```diff
@@ -0,0 +1,205 @@
...
+	"sync"
+
+	"github.com/twmb/franz-go/pkg/kadm"
+	"github.com/twmb/franz-go/pkg/kerr"
+)
+
+// getGroupLag is the inlined version of `kadm.Client.Lag` but updated to work with when the group
+// doesn't have live participants.
+// Similar to `kadm.CalculateGroupLagWithStartOffsets`, it takes into account that the group may not have any commits.
+func getGroupLag(ctx context.Context, admClient *kadm.Client, topic, group string, fallbackOffset int64) (kadm.GroupLag, error) {
```
narqo, line 18: We briefly talked about this `getGroupLag` chunk @dimitarvdimitrov, asking if we should aim at keeping this utils in the forked version of `kgo`.

Thinking a bit more, I don't think I'm fancy about that idea. That is, in my mind, these are simple helper functions, that do "maths" against four sets of offsets (partition's start, end, commit, fallback). It's not something that we're "fixing" in kgo.

Instead, moving forward, I'd rather simplified this code, with the idea that our consumers never consume more than one topic/partition. I.e. we don't even need the complexity of `kadm.GroupLag` struct here.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+
+	for _, part := range b.parts {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		if part < 0 {
+			return nil
+		}
+
+		// TODO(v): calculating lag for each individual partition requests data for the whole group every time. This is redundant.
```
dimitarvdimitrov, line 367: i agree we do a lot of work multiple times and throw away most of it. But wouldn't lag change as we move along the partitions? For example the lag now might be lower than the lag after we process one partition.

Also - this should happen fairly infrequently, so doing the work multiple times and throwing away the result may be ok.

dimitarvdimitrov (commented):
A lot of work and some of it isn't that trivial. I didn't find any critical issues.

As for splitting this up into smaller PRs: after having reviewed it I would have approached it by splitting into 3 PRs:

1. just config and services setup (I still don't see where `blockbuilder.New()` is used btw)
2. everything that happens in blockBuilder.starting
3. everything that happens in blockBuilder.running

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -0,0 +1,138 @@
...
+		cfg.LookbackOnNoCommit = -1
+
+		bb, err := New(cfg, test.NewTestingLogger(t), prometheus.NewPedanticRegistry(), overrides)
+		require.NoError(t, err)
+
+		require.Error(t, services.StartAndAwaitRunning(ctx, bb))
+	})
+}
+
+func TestBlockBuilder_NextConsumeCycle(t *testing.T) {
```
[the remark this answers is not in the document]
narqo: We have 1K more lines of tests in the `blockbuilder` branch :D ([ref](https://github.com/grafana/mimir/blob/e1827ef467b2148b9c6ccb0afdc94ad28bf999cc/pkg/blockbuilder/blockbuilder_test.go)). This PR doesn't include the "meat" of the `NextConsumeCycle` so testing it here would be very redundant.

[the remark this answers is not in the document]
dimitarvdimitrov: ok, sounds good

pkg/blockbuilder/blockbuilder.go
narqo: > wouldn't lag change as we move along the partitions? For example the lag now might be lower than the lag after we process one partition

Yes, but also note that block-builder's cycles consume data from the past hour, up until beginning of this hour. I.e. if cycle starts at time `T`, block-builder needs to consume records produced in interval `[T-1h, T)`. For us, records produced after `T` (i.e. during the cycle) belong to the next cycle.

There is an edge case, when someone produces Kafka records with time set in the past. We cannot gracefully handle that in the current iteration, so those will be consumed in the next cycle anyway.

pkg/blockbuilder/kafkautil.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+
+	return calculateGroupLag(offsets, startOffsets, endOffsets), nil
+}
+
+var errListMissing = errors.New("missing from list offsets")
+
+// Inlined from https://github.com/grafana/mimir/blob/04df05d32320cd19281591a57c19cceba63ceabf/vendor/github.com/twmb/franz-go/pkg/kadm/groups.go#L1711
+func calculateGroupLag(commit kadm.OffsetResponses, startOffsets, endOffsets kadm.ListedOffsets) kadm.GroupLag {
```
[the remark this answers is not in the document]
narqo: Oh, you seem right. I've refactored the `getGroupLag` a couple of times in the past months, and this current version is indeed can just call `kadm.CalculateGroupLagWithStartOffsets` now 👍

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+
+	// Empty any previous artifacts.
+	if err := os.RemoveAll(b.cfg.BlocksStorageConfig.TSDB.Dir); err != nil {
+		return fmt.Errorf("removing tsdb dir: %w", err)
+	}
+	if err := os.MkdirAll(b.cfg.BlocksStorageConfig.TSDB.Dir, os.ModePerm); err != nil {
+		return fmt.Errorf("creating tsdb dir: %w", err)
+	}
+
+	// TODO: add a test to test the case where the consumption on startup happens
```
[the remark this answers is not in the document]
narqo: This will be a part of future PRs, where we will actually consume data from Kafka.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+	if cycleEnd.After(time.Now()) {
+		cycleEnd = cycleEnd.Add(-interval)
+	}
+	return cycleEnd
+}
+
+// NextConsumeCycle manages consumption of currently assigned partitions.
+// The cycleEnd argument indicates the timestamp (relative to Kafka records) up until which to consume from partitions
+// in this cycle. That is, Kafka records produced after the cycleEnd mark will be consumed in the next cycle.
+func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time) error {
```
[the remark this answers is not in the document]
narqo: We use it extensively in the tests (which aren't part of this PR, [example][1]).

[1]: https://github.com/grafana/mimir/blob/e1827ef467b2148b9c6ccb0afdc94ad28bf999cc/pkg/blockbuilder/blockbuilder_test.go#L231

narqo:
@dimitarvdimitrov 

> As for splitting this up into smaller PRs: after having reviewed it I would have approached it by splitting into 3 PRs...

I can open a new PR with your proposal, split this one by three large commits, or we continue iterating on this one, as I'm addressing the feedback. What would be you preference?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+
+	return nil
+}
+
+type KafkaConfig struct {
+	Address             string          `yaml:"address"`
+	Topic               string          `yaml:"topic"`
+	ClientID            string          `yaml:"client_id"`
+	DialTimeout         time.Duration   `yaml:"dial_timeout"`
+	ConsumerGroup       string          `yaml:"consumer_group"`
```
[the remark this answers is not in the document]
narqo: I added a TODO to address this one later. My only concern was that Marco actively refactored `storage/ingest` for his work to use `KafkaReader` from query-frontend. I didn't want to expose more internals of `storage/ingest` while the package is still in the jelly form.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+
+	for {
+		select {
+		case <-time.After(waitTime):
+			cycleEnd := nextCycleTime
+			level.Info(b.logger).Log("msg", "triggering next consume from running", "cycle_end", cycleEnd, "cycle_time", nextCycleTime)
+			err := b.NextConsumeCycle(ctx, cycleEnd)
+			if err != nil {
+				b.metrics.consumeCycleFailures.Inc()
+				level.Error(b.logger).Log("msg", "consume cycle failed", "cycle_end", cycleEnd, "err", err)
```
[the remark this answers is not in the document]
narqo: TBF, I'm not sure what is the desired behaviour when an error reaches the `NextConsumeCycle`. Should we crash the whole app? 

Note that the pieces of the business logic, that 
1. consume from Kafka,
2. build blocks,
3. upload blocks,
4. update consumer's offset and commit the offset

are parts of the future PRs.

dimitarvdimitrov:
> I can open a new PR with your proposal, split this one by three large commits, or we continue iterating on this one, as I'm addressing the feedback. What would be you preference?

i'm fine to continue with this one 👍 this was more for future PRs

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+	limits      *validation.Overrides
+	kafkaClient *kgo.Client
+	bucket      objstore.Bucket
+
+	instanceID int
+	parts      []int32
+	// fallbackOffset is the low watermark from where a partition which doesn't have a commit will be consumed.
+	// It can be either a timestamp or the kafkaStartOffset (the latter isn't supported still).
+	// TODO(v): to support setting it to kafkaStartOffset we need to rework how a lagging cycle is chopped into time frames.
+	fallbackOffset int64
```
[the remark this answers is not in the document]
dimitarvdimitrov: my concern was having this field with double meaning - a timestamp or maybe -2 which is also a valid timestamp, but really it means "start" 🥴 I know kafka does this too and I don't like it there either. If we can't figure it out, it's not a big deal, IIRC we do a similar thing in the ingest storage reader

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
dimitarvdimitrov: ok, but then it can be unexported, right? The tests are in the same package

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
dimitarvdimitrov: what we do with regular consumption in the ingester is retry indefinitely. It's also dangerous because the app can block forever, but it's harder to lose data. There'd be some errors which you can ignore and others on which you always retry. I think you already have the error classification in another PR that I've reviewed

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: aha, that makes sense. So caching the lag for each consume cycle can only potentially delay ingesting some data until the next cycle. But that's probably not a big deal. Yeah I think what you have is easier to reason about.

pkg/blockbuilder/kafkautil.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if fallbackOffset > 0 {
+			return admClient.ListOffsetsAfterMilli(ctx, fallbackOffset, topic)
+		}
+		// This should not happen because fallbackOffset already went through the validation by this point.
+		return nil, fmt.Errorf("cannot resolve fallback offset for value %v", fallbackOffset)
+	})
+	// If the group-partition in offsets doesn't have a commit, fall back depending on where fallbackOffset points at.
+	for topic, pt := range startOffsets.Offsets() {
+		for part := range pt {
```
pracucci, line 49: [nit] I see using `part` name to refer to partition. "Part" means a different thing in english. Please be a bit more verbose and rename everywhere "part" into "partition" or "partitionID" to make it clear, please.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,509 @@
...
+	services.Service
+
+	cfg         Config
+	logger      log.Logger
+	register    prometheus.Registerer
+	limits      *validation.Overrides
+	kafkaClient *kgo.Client
+	bucket      objstore.Bucket
+
+	parts []int32
```
pracucci, line 43: [nit] Let's make it clear. "Parts" means a different thing in english.

```suggestion
	partitionIDs []int32
```

pracucci: Should we even rename it to `ownedPartitionIDs` to be even more clear?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,509 @@
...
+func (b *BlockBuilder) NextConsumeCycle(ctx context.Context, cycleEnd time.Time) error {
+	defer func(t time.Time) {
+		b.metrics.consumeCycleDuration.Observe(time.Since(t).Seconds())
+	}(time.Now())
+
+	for _, part := range b.parts {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		if part < 0 {
```
pracucci, line 328: In which case could it happen? And... if happens, why do we stop iterating partitions and `return nil`?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	// TODO(codesome): validate the consumption interval. Must be <=2h and can divide 2h into an integer.
+	if cfg.ConsumeInterval < 0 {
+		return fmt.Errorf("-consume-interval must be non-negative")
+	}
+
+	return nil
+}
+
+type KafkaConfig struct {
```
pracucci, line 434: [nit] I suggest to move all config structs to a `config.go` file.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,542 @@
...
+		level.Error(logger).Log("msg", "failed to get hostname", "err", err)
+		os.Exit(1)
+	}
+
+	cfg.Kafka.RegisterFlagsWithPrefix("block-builder.", f)
+
+	f.StringVar(&cfg.InstanceID, "block-builder.instance-id", hostname, "Instance id.")
+	f.DurationVar(&cfg.ConsumeInterval, "block-builder.consume-interval", time.Hour, "Interval between block consumption cycles.")
+	f.DurationVar(&cfg.ConsumeIntervalBuffer, "block-builder.consume-interval-buffer", 15*time.Minute, "Extra buffer between subsequent block consumption cycles to avoid small blocks.")
+	f.DurationVar(&cfg.LookbackOnNoCommit, "block-builder.lookback-on-no-commit", 12*time.Hour, "How much of the historical records to look back when there is no kafka commit for a partition.")
```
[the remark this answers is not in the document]
pracucci: Discussed in a call. My suggestion is to not allow special (negative) values at all. If you want to replay from partition start you can just set this value to a very long time (higher than Kafka retention). If you want to replay from the end of the partition, you should be able to do it setting this value to 0.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+) (*BlockBuilder, error) {
+	b := &BlockBuilder{
+		cfg:      cfg,
+		logger:   logger,
+		register: reg,
+		limits:   limits,
+		metrics:  newBlockBuilderMetrics(reg),
+	}
+
+	instanceID, err := parseIDFromInstance(b.cfg.InstanceID)
```
pracucci, line 66: Can we avoid the parsing at all? Looks an unnecessary extra complication to me.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+	defer func() {
+		if err != nil {
+			if b.kafkaClient != nil {
+				b.kafkaClient.Close()
+			}
+		}
+	}()
+
+	// Empty any previous artifacts.
+	if err := os.RemoveAll(b.cfg.BlocksStorageConfig.TSDB.Dir); err != nil {
```
pracucci, line 112: I think we should use dedicated folder (specified in the block builder Config) instead of the TSDB one. TSDB one is the one used by ingesters and I don't want to run into any conflict when running, for example, the monolithic mode during development / local testing.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+
+	metrics := ingest.NewKafkaReaderClientMetrics("block-builder", b.register)
+	opts = append(
+		commonKafkaClientOptions(b.cfg.Kafka, b.logger, metrics),
+		opts...,
+	)
+	b.kafkaClient, err = kgo.NewClient(opts...)
+	if err != nil {
+		return fmt.Errorf("creating kafka client: %w", err)
+	}
```
pracucci, line 152: Can we reuse `ingest.NewKafkaReaderClient()`? If you pick my suggestion to reuse `ingest.KafkaConfig` in block-builder then it should be easy to do.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+		return fmt.Errorf("removing tsdb dir: %w", err)
+	}
+	if err := os.MkdirAll(b.cfg.BlocksStorageConfig.TSDB.Dir, os.ModePerm); err != nil {
+		return fmt.Errorf("creating tsdb dir: %w", err)
+	}
+
+	// TODO: add a test to test the case where the consumption on startup happens
+	// after the LookbackOnNoCommit with records before the after the consumption
+	// start point.
+	startAtOffsets, fallbackOffset, err := b.findOffsetsToStartAt(ctx)
```
pracucci, line 122: Why is finding the `startAtOffsets` important at all when starting the block-builder? Don't you control the offsets to consume anyway when the consumption will happen? I guess I'm missing something.

pkg/blockbuilder/blockbuilder_metrics.go
```diff
@@ -0,0 +1,67 @@
...
+		Help: "Total number of errors while fetching by the consumer.",
+	})
+
+	m.assignedPartitions = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+		Name: "cortex_blockbuilder_assigned_partitions",
+		Help: "The number of partitions currently assigned to this instance.",
+	}, []string{"topic"})
+
+	m.consumerLag = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+		Name: "cortex_blockbuilder_consumer_lag",
```
pracucci, line 62: I would add the "records" unit because in ingest storage we measure lag as "seconds", so it's better to be clear here.

```suggestion
		Name: "cortex_blockbuilder_consumer_lag_records",
```

pkg/blockbuilder/blockbuilder_metrics.go
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package blockbuilder
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+type blockBuilderMetrics struct {
```
pracucci, line 10: I would remove all metrics not tracked for now. It's quite difficult to review metrics that are not currently used. You can introduce them in the actual PRs where these metrics are tracked.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+	f.StringVar(&cfg.ClientID, prefix+"kafka.client-id", "", "The Kafka client ID.")
+	f.DurationVar(&cfg.DialTimeout, prefix+"kafka.dial-timeout", 5*time.Second, "The maximum time allowed to open a connection to a Kafka broker.")
+	f.StringVar(&cfg.ConsumerGroup, prefix+"kafka.consumer-group", "block-builder", "The consumer group used to keep track of the consumed offsets for each partition.")
+}
+
+func (cfg *KafkaConfig) Validate() error {
+	return nil
+}
+
+type partitionAssignmentVar map[int][]int32
```
pracucci, line 458: I would prefer a more explicit config, where the key is the full instance ID (e.g. the full pod name) not just part of it. If you follow this suggestion, then I think you can just use `validation.LimitsMap` adding support for `[]int32` as value.

```suggestion
type partitionAssignmentVar map[string][]int32
```

pkg/blockbuilder/blockbuilder.go
narqo: The hairy part is how both `ingest` and `blockbuilder` create a bootstrapping (ephemeral) admin client to construct the final Kafka client's options. And this ephemeral client needs some config too (see [1][1], [2][2]).

As mentioned above (https://github.com/grafana/mimir/pull/9068#discussion_r1732255080) I didn't want to refactor the internals of `ingest` package until you finish with the refactoring for using Kafka client from query-frontend.

[1]: https://github.com/grafana/mimir/blob/ba81e660c004f177d7c23408681d222d240790d0/pkg/storage/ingest/reader.go#L499-L506
[2]: https://github.com/grafana/mimir/blob/00f696fa4e1425e1f7c442090910ce41f2cf2da8/pkg/storage/ingest/reader.go#L428-L435

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
narqo: After another discussion, the `fallbackOffsets` is no longer overloaded. Now it's the actual `timestamp`, that the `getGroupLag` uses to figure look up the first record used as the starting point for lag calculation.

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
narqo: I think the best we can do here is to fail the application, similar to how ingest storage does it ([code](https://github.com/grafana/mimir/blob/fdb1225907b237a4ffd4cbd6950385e87d71f2cf/pkg/storage/ingest/reader.go#L173-L181)). Everything that can handled gracefully (by retrying or logging) is handled inside `NextConsumeCycle`

pkg/blockbuilder/blockbuilder.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+	// TODO(codesome): validate the consumption interval. Must be <=2h and can divide 2h into an integer.
+	if cfg.ConsumeInterval < 0 {
+		return fmt.Errorf("-consume-interval must be non-negative")
+	}
+
+	return nil
+}
+
+type KafkaConfig struct {
```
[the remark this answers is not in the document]
narqo: > [..] so you can just add `BlockBuilderConsumerGroup` to `KafkaConfig`

How about we keep it as _for now_. After finishing with this iteration, we revisit the ingest storage's `KafkaConfig` and extract general parts into a shared package with config and client, similar to how it's for bucket client?

My concern is that we're already in an internal discussion about having more "components", that will utilize this same Kafka instance (referring to the internal discussions about _scheduling the assigned block-building jobs_ and also the  _limits_). I don't think it's sustainable to keep adding everything that touches Kafka into ingest storage's config.

pkg/blockbuilder/config.go
```diff
@@ -0,0 +1,116 @@
...
+
+	ConsumeInterval       time.Duration `yaml:"consume_interval"`
+	ConsumeIntervalBuffer time.Duration `yaml:"consume_interval_buffer"`
+	LookbackOnNoCommit    time.Duration `yaml:"lookback_on_no_commit" category:"advanced"`
+
+	Kafka KafkaConfig `yaml:"kafka"`
+
+	// BlocksStorageConfig is defined outside the block-builder config and is injected dynamically.
+	BlocksStorageConfig tsdb.BlocksStorageConfig `yaml:"-"`
+}
```
dimitarvdimitrov, line 31: Didn't we agree to use the same configuration from ingest storage - `ingest_strage.kafka`? Why are there yaml tags on the `kafka` field?

narqo: I'm doing it in the next PR #9147

pkg/blockbuilder/kafkautil.go
```diff
@@ -0,0 +1,91 @@
...
+	endOffsets, err := admClient.ListEndOffsets(ctx, topic)
+	if err != nil {
+		return nil, err
+	}
+
+	resolveFallbackOffsets := sync.OnceValues(func() (kadm.ListedOffsets, error) {
+		if fallbackOffset > 0 {
+			return admClient.ListOffsetsAfterMilli(ctx, fallbackOffset, topic)
+		}
+		switch fallbackOffset {
```
dimitarvdimitrov, line 54: didn't we remove the special meanings of the fallbackOffset, so now it's only fallbackTimestampMillis?

pkg/blockbuilder/kafkautil_test.go
```diff
@@ -0,0 +1,136 @@
...
+	})
+
+	t.Run("fallbackOffset=end", func(t *testing.T) {
+		groupLag, err := getGroupLag(ctx, admClient, testTopic, testGroup, kafkaOffsetEnd)
+		require.NoError(t, err)
+
+		require.EqualValues(t, 0, getTopicPartitionLag(t, groupLag, testTopic, 0), "partition 0 must have no lag")
+		require.EqualValues(t, 0, getTopicPartitionLag(t, groupLag, testTopic, 1), "partition 1 must fall back to the end and get no lag")
+		require.EqualValues(t, 0, getTopicPartitionLag(t, groupLag, testTopic, 2), "partition 2 has no data and must have no lag")
+	})
```
dimitarvdimitrov, line 92: do we still support those?

dimitarvdimitrov (approved):
Nice work 🚀 LGTM % the two remaining fallbackOffset mentions

PR #9082, opened 2024-08-23
mimir-build-image: Explicitly set GOTOOLCHAIN=local

aknuds1:
#### What this PR does
In mimir-build-image/Dockerfile, explicitly override environment variable `GOTOOLCHAIN` to be `local`, to make sure that the Docker image's bundled Go toolchain is used, rather than the one specified via the `toolchain` directive in go.mod. The same is already done in [official Go images](https://github.com/docker-library/golang/issues/472), but being explicit helps also to document.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/mimir-build-image/Dockerfile
+++ b/mimir-build-image/Dockerfile
@@ -9,6 +9,10 @@ FROM golang:1.22.5-bookworm
 ARG goproxyValue
 ENV GOPROXY=${goproxyValue}
 ENV SKOPEO_DEPS="libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config"
+# Override toolchain directive in go.mod, to ensure the image's Go version is used.
+# Be aware that the official Go Dockerfiles already do this, but let's be explicit.
+# https://github.com/docker-library/golang/issues/472
+RUN go env -w GOTOOLCHAIN=local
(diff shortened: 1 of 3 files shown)


pstibrany (approved):
Thank you very much @aknuds1! This cleans up confusion about which toolchain is used in our builds.

mimir-build-image/Dockerfile
```diff
@@ -9,6 +9,10 @@ FROM golang:1.22.5-bookworm
 ARG goproxyValue
 ENV GOPROXY=${goproxyValue}
 ENV SKOPEO_DEPS="libgpgme-dev libassuan-dev libbtrfs-dev libdevmapper-dev pkg-config"
+# Override toolchain directive in go.mod, to ensure the image's Go version is used.
+# Be aware that the official Go Dockerfiles already do this, but let's be explicit.
+# https://github.com/docker-library/golang/issues/472
+RUN go env -w GOTOOLCHAIN=local
```
narqo, line 15: (This is super nit), but this Dockerfile sets `GOPROXY` and `GOCACHE` as explicit environment variables. Is there any reason to set `GOTOOLCHAIN` via `go env`?

```suggestion
ENV GOTOOLCHAIN=local
```

aknuds1: Good point @narqo, as [the official Go image](https://github.com/docker-library/golang/blob/539882fb23e90d31854a51a773accf8731cf0c9d/1.22/bookworm/Dockerfile#L123) also uses `ENV`. I thought they were using `go env -w`.

Issue #9065, opened 2024-08-21
New config option in mixin to use native histograms only in dashboards


krajorama:
Currently Mimir dashboards execute queries on some latency histograms using both native and classic histograms. This puts extra load on the backend and can be slower than just running one of the queries.

Add a mixin configuration parameter to be able to restrict which kind of query is run at build time.
MVP: allow generating dashboards with either both queries or only native histograms queries.

This would not be necessary if  https://github.com/grafana/grafana/issues/79848 was implemented or PromQL would optimize away impossible queries (x < -Inf).

krajorama:
No push for this over a year, closing.

PR #9086, opened 2024-08-23
Add query pruning middleware

zenador:
#### What this PR does

(Hackathon project) Query pruning optimises PromQL queries by rewriting them to avoid evaluating expressions that can't produce a result, which is particularly useful in [dashboards with toggles between native and classic histograms](https://github.com/grafana/jsonnet-libs/pull/1164/files#diff-040bce91f6d46ff86f067fe7e11934183b2ea095ff6d4c62c5e798f24b6d9269R86).

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/pruning.go
+++ b/pkg/frontend/querymiddleware/astmapper/pruning.go
@@ -13,9 +13,9 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 )
 
-func NewQueryPruner(ctx context.Context, logger log.Logger) (ASTMapper, error) {
+func NewQueryPruner(ctx context.Context, logger log.Logger) ASTMapper {
 	pruner := newQueryPruner(ctx, logger)
-	return NewASTExprMapper(pruner), nil
+	return NewASTExprMapper(pruner)
 }
 
 type queryPruner struct {
--- a/pkg/frontend/querymiddleware/astmapper/pruning.go
+++ b/pkg/frontend/querymiddleware/astmapper/pruning.go
@@ -34,6 +34,12 @@ func (pruner *queryPruner) MapExpr(expr parser.Expr) (mapped parser.Expr, finish
 	}
 
 	switch e := expr.(type) {
+	case *parser.ParenExpr:
+		mapped, finished, err = pruner.MapExpr(e.Expr)
+		if err != nil {
+			return e, false, err
+		}
+		return &parser.ParenExpr{Expr: mapped, PosRange: e.PosRange}, finished, nil
 	case *parser.BinaryExpr:
 		return pruner.pruneBinOp(e)
 	default:
--- a/pkg/frontend/querymiddleware/astmapper/pruning.go
+++ b/pkg/frontend/querymiddleware/astmapper/pruning.go
@@ -49,6 +49,8 @@ func (pruner *queryPruner) pruneBinOp(expr *parser.BinaryExpr) (mapped parser.Ex
 		return pruner.handleMultiplyOp(expr), false, nil
 	case parser.GTR, parser.LSS:
 		return pruner.handleCompOp(expr), false, nil
+	case parser.LOR:
+		return pruner.handleOrOp(expr), false, nil
 	default:
 		return expr, false, nil
 	}
@@ -149,5 +151,31 @@ func (pruner *queryPruner) isInfinite(expr parser.Expr) (bool, bool) {
 		return false, false
 	default:
 		return false, false
+	}
+}
+
+func (pruner *queryPruner) handleOrOp(expr *parser.BinaryExpr) parser.Expr {
+	switch {
+	case pruner.isEmpty(expr.LHS):
+		return expr.RHS
+	case pruner.isEmpty(expr.RHS):
+		return expr.LHS
+	}
+	return expr
+}
+
+func (pruner *queryPruner) isEmpty(expr parser.Expr) bool {
+	mapped, _, err := pruner.MapExpr(expr)
+	if err == nil {
+		expr = mapped
+	}
+	switch e := expr.(type) {
+	case *parser.ParenExpr:
+		return pruner.isEmpty(e.Expr)
+	default:
+		if e.String() == `vector(0) < -Inf` {
+			return true
+		}
+		return false
 	}
 }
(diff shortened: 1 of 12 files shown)


tacole02 (approved):
LGTM 😄

krajorama (commented):
awesome work, I only have  a couple of minor comments

pkg/frontend/querymiddleware/prune_test.go
```diff
@@ -0,0 +1,105 @@
...
+	seriesName := `test_float`
+	for i := 0; i < numSeries; i++ {
+		nss := promql.NewStorageSeries(promql.Series{
+			Metric:     labels.FromStrings("__name__", seriesName, "series", fmt.Sprint(i)),
+			Floats:     floats,
+			Histograms: histograms,
+		})
+		storageSeries = append(storageSeries, nss)
+	}
+	queryable := storageSeriesQueryable(storageSeries)
```
charleskorn, line 42: `promqltest.LoadedStorage` might help simplify this setup

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -65,6 +65,7 @@ type Config struct {
 	MaxRetries               int           `yaml:"max_retries" category:"advanced"`
 	NotRunningTimeout        time.Duration `yaml:"not_running_timeout" category:"advanced"`
 	ShardedQueries           bool          `yaml:"parallelize_shardable_queries"`
+	PrunedQueries            bool          `yaml:"prune_queries"`
```
charleskorn, line 68: Should we mark this as experimental?

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -92,6 +93,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
 	f.DurationVar(&cfg.SplitQueriesByInterval, "query-frontend.split-queries-by-interval", 24*time.Hour, "Split range queries by an interval and execute in parallel. You should use a multiple of 24 hours to optimize querying blocks. 0 to disable it.")
 	f.BoolVar(&cfg.CacheResults, "query-frontend.cache-results", false, "Cache query results.")
 	f.BoolVar(&cfg.ShardedQueries, "query-frontend.parallelize-shardable-queries", false, "True to enable query sharding.")
+	f.BoolVar(&cfg.PrunedQueries, "query-frontend.prune-queries", false, "True to enable query pruning.")
```
charleskorn, line 96: [nit] I'm not sure "pruning" makes this feature clear enough - what about something like this?

```suggestion
	f.BoolVar(&cfg.PrunedQueries, "query-frontend.prune-queries", false, "True to enable pruning dead code (eg. expressions that cannot produce any results) and simplifying expressions (eg. expressions that can be evaluated immediately) in queries.")
```

pkg/frontend/querymiddleware/prune.go
```diff
@@ -0,0 +1,72 @@
...
+	log := spanlogger.FromContext(ctx, p.logger)
+
+	prunedQuery, err := p.pruneQuery(ctx, r.GetQuery())
+	if err != nil {
+		level.Warn(log).Log("msg", "failed to prune the input query, falling back to the original query", "query", r.GetQuery(), "err", err)
+
+		return p.next.Do(ctx, r)
+	}
+
+	level.Debug(log).Log("msg", "query has been purged", "original", r.GetQuery(), "rewritten", prunedQuery)
```
charleskorn, line 46:
```suggestion
	level.Debug(log).Log("msg", "query has been rewritten by pruning", "original", r.GetQuery(), "rewritten", prunedQuery)
```

charleskorn, line 46: [nit] It'd be nice to skip this log message if the expression is not modified (or emit a different message if the expression is not modified)

pkg/frontend/querymiddleware/astmapper/pruning.go
```diff
@@ -0,0 +1,181 @@
...
+	case parser.GTR, parser.LSS:
+		return pruner.handleCompOp(expr), false, nil
+	case parser.LOR:
+		return pruner.handleOrOp(expr), false, nil
+	default:
+		return expr, false, nil
+	}
+}
+
+func calcInf(sign bool, coeff string) (*parser.NumberLiteral, bool) {
```
charleskorn, line 59: [nit] `isPositive` might be a clearer name than `sign`

pkg/frontend/querymiddleware/astmapper/pruning.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+}
+
+func (pruner *queryPruner) pruneBinOp(expr *parser.BinaryExpr) (mapped parser.Expr, finished bool, err error) {
+	switch expr.Op {
+	case parser.MUL:
+		return pruner.handleMultiplyOp(expr), false, nil
+	case parser.GTR, parser.LSS:
+		return pruner.handleCompOp(expr), false, nil
+	case parser.LOR:
```
charleskorn, line 52: Might be something for another PR: we can do something similar for `and` and `unless`:

* if either side of an `and` is empty, then the whole op will return no results
* if the left side of an `unless` is empty, then the whole op will return no results
* if the right side of an `unless` is empty, then we can just use the left side and ignore the right side

charleskorn (commented):
Thanks for putting this together!

krajorama:
Solves https://github.com/grafana/mimir/issues/9065

charleskorn (approved):
LGTM 🚀

pkg/frontend/querymiddleware/prune_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{`(((-1 * -Inf) < avg(rate(foo[3m]))) unless avg(rate(foo[2m]))) or avg(rate(%s[1m]))`, true},
+		{`((((-1 * -Inf) < avg(rate(foo[4m]))) unless avg(rate(foo[3m]))) and avg(rate(foo[2m]))) or avg(rate(%s[1m]))`, true},
+	}
+	for _, template := range templates {
+		t.Run(template.query, func(t *testing.T) {
+			query := fmt.Sprintf(template.query, seriesName)
+			req := &PrometheusRangeQueryRequest{
+				path:      "/query_range",
+				start:     0,
+				end:       int64(endTime * 1000 * 60),
```
charleskorn, line 65: [nit] This might be clearer as something like `(100 * time.Minute).Milliseconds()`

Issue #1617, opened 2022-04-04
Upgrade mixtool


pracucci:
The mixtool version used in `mimir-build-image` is quite old and we should update it. I've pushed an updated build image with the latest mixtool commit, but dashboard linter shows many failures. We should investigate and fix it:
https://github.com/grafana/mimir/compare/update-mixtool

_It's not high priority._

pracucci:
Done in https://github.com/grafana/mimir/pull/8032

PR #9117, opened 2024-08-27
makefile: Run `mixtool` in container

julienduchesne:
Also updates mixtool to latest version which closes https://github.com/grafana/mimir/issues/1617 
Mixtool was actually updated two months ago, so the dashboards are already all compliant!

**Makefile changes:**
I was looking at the issue above and I wanted to test it using the build image. I noticed that it wasn't actually running in the container. This makes the `check-mixin-*` steps run in the build image for local uses and should make it easier to test in the future.


pracucci (approved):
I don't see why we shouldn't run mixtool in container, so LGTM.

PR #9121, opened 2024-08-28
MQE: Panic when memory isn't managed as we expect

jhesketh:
#### What this PR does

* Panics when we try to Put more objects back into the pool then expected. This likely means we have queries interfering with each other, so panic'ing is appropriate.
* During tests, panic when we don't put all the objects back into their pools.
* Fix a bug where we would unnecessarily get extra slices in binary operations.

Related to https://github.com/grafana/mimir-squad/issues/2280

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/query.go
```diff
@@ -595,6 +595,12 @@ func (q *Query) Close() {
 	default:
 		panic(fmt.Sprintf("unknown result value type %T", q.result.Value))
 	}
+
+	if q.engine.pedantic {
+		if q.memoryConsumptionTracker.CurrentEstimatedMemoryConsumptionBytes > 0 {
```
charleskorn, line 600: We should only do this if the query succeeded (`q.result.Err == nil`) - if something goes wrong during the query, not returning everything to the pool is OK.

pkg/streamingpromql/functions/common_test.go
```diff
@@ -43,6 +44,8 @@ func TestFloatTransformationFunc(t *testing.T) {
 			{H: &histogram.FloatHistogram{Count: 1, Sum: 2}},
 		},
 	}
+	// Increase the memory tracker for 2 FPoints, and 1 HPoint
+	memoryConsumptionTracker.IncreaseMemoryConsumption(types.FPointSize*2 + types.HPointSize*1)
```
charleskorn, line 48: Why is this necessary? Can't we just assert that `CurrentEstimatedMemoryConsumptionBytes` is 0 below?

(same comment applies to the other similar changes in this file)

pkg/streamingpromql/limiting/memory_consumption.go
```diff
@@ -51,5 +51,8 @@ func (l *MemoryConsumptionTracker) IncreaseMemoryConsumption(b uint64) error {
 
 // DecreaseMemoryConsumption decreases the current memory consumption by b bytes.
 func (l *MemoryConsumptionTracker) DecreaseMemoryConsumption(b uint64) {
+	if b > l.CurrentEstimatedMemoryConsumptionBytes {
+		panic("Memory consumption tracking is negative. This indicates something has been returned to a pool more than once.")
```
charleskorn, line 55: [nit] I think something like this would be clearer:

```suggestion
		panic("Estimated memory consumption of this query is negative. This indicates something has been returned to a pool more than once, which is a bug.")
```

pkg/streamingpromql/limiting/memory_consumption_test.go
```diff
@@ -44,6 +44,15 @@ func TestMemoryConsumptionTracker_Unlimited(t *testing.T) {
...
 	assertRejectedQueriesCount(t, reg, 0)
+
+	// Test reducing memory consumption to a negative panics
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("We did not panic on negative memory consumption")
+		}
+	}()
+
+	tracker.DecreaseMemoryConsumption(150)
```
charleskorn, line 55: This could be simplified a bit with `require.Panics()`. 

(same comment applies below)

pkg/streamingpromql/operators/binary_operation_test.go
```diff
@@ -1130,7 +1135,10 @@ func TestBinaryOperationSeriesBuffer(t *testing.T) {
 	}
 
 	seriesUsed := []bool{true, false, true, true, true}
-	buffer := newBinaryOperationSeriesBuffer(inner, seriesUsed, limiting.NewMemoryConsumptionTracker(0, nil))
+	memoryConsumptionTracker := limiting.NewMemoryConsumptionTracker(0, nil)
+	// We have 6 FPoints from the inner series
+	memoryConsumptionTracker.IncreaseMemoryConsumption(types.FPointSize * 6)
```
charleskorn, line 1140: Why is this needed?

pkg/streamingpromql/query.go
jhesketh: I think this is guaranteed by `if q.result == nil` above. But also, pedantic is only enabled during testing and I think it's reasonable to be pedantic about errors during tests too.

pkg/streamingpromql/functions/common_test.go
jhesketh: Because the functions we're testing actually call `DecreaseMemoryConsumption` at some point, which requires the current estimated memory to be larger than what is returned. I figured may as well put in real values if that path is being executed.

pkg/streamingpromql/operators/binary_operation_test.go
jhesketh: See other comment. Will resolve this one and move conversation there.

pkg/streamingpromql/query.go
charleskorn: `q.result` will be non-nil if the query was evaluated, even if the query failed.

If the query succeeds, `q.result.Err` will be nil and `q.result.Value` will be set.

If the query fails, `q.result.Err` will be non-nil.

>But also, pedantic is only enabled during testing and I think it's reasonable to be pedantic about errors during tests too.

I don't think we should do this: this will force us to ensure that things are returned to the pool even when errors occur, which may make the logic in places more complex than it needs to be.

pkg/streamingpromql/functions/common_test.go
charleskorn: Sorry, I'm missing something: where is `DecreaseMemoryConsumption` called in this test? `floatTransformationFunc` reuses the slice it's passed.

(I can see this is necessary for `TestFloatTransformationDropHistogramsFunc` though)

pkg/streamingpromql/functions/common_test.go
```diff
@@ -81,7 +88,8 @@ func TestFloatTransformationDropHistogramsFunc(t *testing.T) {
 		Histograms: nil, // Histograms should be dropped
 	}
 
-	modifiedSeriesData, err := transformFunc(seriesData, limiting.NewMemoryConsumptionTracker(0, nil))
+	modifiedSeriesData, err := transformFunc(seriesData, memoryConsumptionTracker)
 	require.NoError(t, err)
 	require.Equal(t, expected, modifiedSeriesData)
+	require.Equal(t, types.FPointSize*2, memoryConsumptionTracker.CurrentEstimatedMemoryConsumptionBytes)
```
charleskorn, line 94: Would be good to add a comment here explaining that we expect the histogram slice to have been returned to the pool.

pkg/streamingpromql/functions/common_test.go
jhesketh: oh, sorry, I see your point. I'm doing this for consistency with `TestFloatTransformationDropHistogramsFunc`

charleskorn (approved):
LGTM, and nice catch on the issue in the binary operator 🚀

PR #9125, opened 2024-08-28
Use grafana URL to test receivers

titolins:
Fixes https://github.com/grafana/alerting-squad/issues/872

As described, alerts should use the Grafana URL instead of the remote AM's. For regular alert notifications this was already done, this PR fixes for tests

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -192,7 +192,7 @@ type Replicator interface {
 }
 
 // New creates a new Alertmanager.
-func New(cfg *Config, reg *prometheus.Registry, tmplExternalURL *url.URL) (*Alertmanager, error) {
+func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
 	if cfg.TenantDataDir == "" {
 		return nil, fmt.Errorf("directory for tenant-specific AlertManager is not configured")
 	}
@@ -210,8 +210,6 @@ func New(cfg *Config, reg *prometheus.Registry, tmplExternalURL *url.URL) (*Aler
 			Name: "alertmanager_notification_rate_limited_total",
 			Help: "Number of rate-limited notifications per integration.",
 		}, []string{"integration"}), // "integration" is consistent with other alertmanager metrics.
-
-		tmplExternalURL: tmplExternalURL,
 	}
 
 	am.registry = reg
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -92,6 +92,7 @@ type Config struct {
 	Retention                         time.Duration
 	MaxConcurrentGetRequestsPerTenant int
 	ExternalURL                       *url.URL
+	TmplExternalURL                   *url.URL
 	Limits                            Limits
 	Features                          featurecontrol.Flagger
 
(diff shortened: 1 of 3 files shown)


pkg/alertmanager/alertmanager.go
```diff
@@ -92,6 +92,7 @@ type Config struct {
 	Retention                         time.Duration
 	MaxConcurrentGetRequestsPerTenant int
 	ExternalURL                       *url.URL
+	TmplExternalURL                   *url.URL
```
santihernandezc, line 95: This new field is not "config" in the same sense as the other fields in the struct, i.e. global config taken from a YAML file and applied to all Alertmanagers.

I think the `Alertmanager` struct would be a better fit for this field (you could add `tmplExternalURL` [here](https://github.com/grafana/mimir/blob/369932771fb4eacef7209b4fd5046965af999513/pkg/alertmanager/alertmanager.go#L130-L131)).

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +901,7 @@ func (am *MultitenantAlertmanager) newAlertmanager(userID string, amConfig *defi
 		Retention:                         am.cfg.Retention,
 		MaxConcurrentGetRequestsPerTenant: am.cfg.MaxConcurrentGetRequestsPerTenant,
 		ExternalURL:                       am.cfg.ExternalURL.URL,
+		TmplExternalURL:                   tmplExternalURL,
```
santihernandezc, line 904: The external URL can change when a new config is applied. For example, consider the following scenario:
1. A user has no Grafana configuration, they are using Mimir's external URL
2. They upload some Grafana configuration, they should be using Grafana's external URL

We should assign the correct external URL on every config sync (check [here](https://github.com/grafana/mimir/blob/369932771fb4eacef7209b4fd5046965af999513/pkg/alertmanager/alertmanager.go#L518-L520), we're updating the templates in the Alertmanager struct, we could use the same lock to update the external URL).

santihernandezc (changes requested):
Nice one! A few comments.

pkg/alertmanager/alertmanager.go
titolins: Got it! Changed it now 👍

pkg/alertmanager/multitenant.go
titolins: Ah yeah, ofc.. I missed the `else` and presumed a new instance would be built every time. Tks, fixed now 👍

pkg/alertmanager/alertmanager.go
```diff
@@ -191,7 +192,7 @@ type Replicator interface {
 }
 
 // New creates a new Alertmanager.
-func New(cfg *Config, reg *prometheus.Registry) (*Alertmanager, error) {
+func New(cfg *Config, reg *prometheus.Registry, tmplExternalURL *url.URL) (*Alertmanager, error) {
```
santihernandezc, line 195: If we're already passing the URL in `ApplyConfig()`, we can skip adding a new parameter here. On the initial sync ([right after the Alertmanager is created](https://github.com/grafana/mimir/blob/f52911d917c8c52e0da6a59348a64dd7f7622072/pkg/alertmanager/multitenant.go#L907-L929)), the AM will get the correct URL to be used when testing receivers.

I'd remove this new parameter.

santihernandezc (approved):
I just tested it, it works as expected, nice job! 🎉 
Please address my comment, once that's fixed you can merge the PR. I don't need to re-review it.

PR #9127, opened 2024-08-28
Convert `Mimir / Config` dashboard to grafonnet

julienduchesne:
POC on a single dashboard to show what a conversion to grafonnet would look like

Benefits:
- Maintained library (new features, easier to upgrade, etc)
- Auto-complete and go-to definition with the language server

image

**I've tested the dashboard. Looks the same and seems to work the same as the current one, but this is a pretty simple case**

### DIFF
--- a/operations/mimir-mixin/dashboards/grafonnet.libsonnet
+++ b/operations/mimir-mixin/dashboards/grafonnet.libsonnet
@@ -1,0 +1,90 @@
+local g = import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet';
+local sortAscending = 1;
+
+
+function(config) g {
+  // Overrides of grafonnet functions
+  dashboard+:: {
+    new(title)::
+      g.dashboard.new('%(prefix)s%(title)s' % { prefix: config.dashboard_prefix, title: title })
+      + g.dashboard.withVariables(
+        g.dashboard.variable.datasource.new('datasource', 'prometheus')
+        + g.dashboard.variable.datasource.generalOptions.withLabel('Data source')
+        + g.dashboard.variable.datasource.generalOptions.withCurrent(config.dashboard_datasource, config.dashboard_datasource)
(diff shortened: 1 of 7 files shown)


operations/mimir-mixin/dashboards/grafonnet.libsonnet
```diff
@@ -0,0 +1,90 @@
...
+        variable('cluster', config.dashboard_variables.cluster_query, config.per_cluster_label),
+
+        variable('namespace', config.dashboard_variables.namespace_query, config.per_namespace_label)
+        + g.dashboard.variable.query.selectionOptions.withIncludeAll(false),
+      ] else [
+        variable('cluster', config.dashboard_variables.cluster_query, config.per_cluster_label)
+        + g.dashboard.variable.query.selectionOptions.withIncludeAll(true, '.*'),
+
+        variable('namespace', config.dashboard_variables.namespace_query, config.per_namespace_label),
+      ];
```
Duologic, line 68: nit: I generally put if/then/else at the beginning of the line so that it's easier to read.

```suggestion
      local selectors =
        if config.singleBinary
        then
          [
            variable('job', config.dashboard_variables.job_query, config.per_job_label),
          ]
        else if multi
        then
          [
            variable('cluster', config.dashboard_variables.cluster_query, config.per_cluster_label),

            variable('namespace', config.dashboard_variables.namespace_query, config.per_namespace_label)
            + g.dashboard.variable.query.selectionOptions.withIncludeAll(false),
          ]
        else
          [
            variable('cluster', config.dashboard_variables.cluster_query, config.per_cluster_label)
            + g.dashboard.variable.query.selectionOptions.withIncludeAll(true, '.*'),

            variable('namespace', config.dashboard_variables.namespace_query, config.per_namespace_label),
          ];
```

aldernero (approved):
LGTM

julienduchesne:
I think making use of grafonnet in `grafana-builder` would probably be a better approach than redoing everything. Will look into that

PR #9131, opened 2024-08-29
OTLP: Convert start timestamps to Prom created timestamps

aknuds1:
#### What this PR does

In the OTLP endpoint, support converting OTel data point start timestamps to Mimir zero samples 1 millisecond before the sample itself, iff the following are true:

* The OTel start timestamp is greater than zero and equal to the data point timestamp

Includes https://github.com/grafana/mimir-prometheus/pull/684.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


jesusvazquez:
This PR includes the latest changes from Mimir Prometheus so I'm waiting on https://github.com/grafana/mimir/pull/9148 before moving forward.

docs/sources/mimir/configure/about-versioning.md
```diff
@@ -81,6 +81,8 @@ The following features are currently experimental:
     - `-distributor.max-request-pool-buffer-size`
   - Enable direct translation from OTLP write requests to Mimir equivalents
     - `-distributor.direct-otlp-translation-enabled`
+  - Enable conversion of OTel start timestamps to Prometheus zero samples to mark series start
+    - `-distributor.otel-created-timestamp-zero-ingestion-enabled`
```
ArthurSens, line 85: In OTel it is called `StartTimeUnixNano`, should we call this flag `-distributor.otel-starttimeunixnano-zero-ingestion-enabled`?

aknuds1: As a point of reference, the OTel SDK calls it `StartTimestamp`. Do you gain anything from including the unixnano suffix? Technically, Prometheus created timestamps are in milliseconds, but there's no point in including that kind of detail when referring to them, is there?

I guess it might make sense to refer to "start timestamps" in this flag rather than "created timestamps", since you're converting the former. My original thinking, OTOH, was to refer to support for created timestamps via zero samples (since zero samples are a way to represent created timestamps) in the OTLP endpoint.

ArthurSens: > As a point of reference, the OTel SDK calls it StartTimestamp. Do you gain anything from including the unixnano suffix?

No, you have a good point to not add unixnano!

> I guess it might make sense to refer to "start timestamps" in this flag rather than "created timestamps", since you're converting the former. My original thinking, OTOH, was to refer to support for created timestamps via zero samples (since zero samples are a way to represent created timestamps) in the OTLP endpoint.

I don't have a strong preference, I was just thinking that people coming from OTel might not understand this flag. Quite unfortunate that OpenMetrics and OTel are using different names here :/

aknuds1: Yeah, I guess you've got a point. Maybe e.g. `-otel-start-time-zero-ingestion-enabled`?

go.mod
```diff
@@ -280,6 +280,10 @@ require (
 // Using a fork of Prometheus with Mimir-specific changes.
 replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v0.0.0-20240830150301-6b342fac9c48
 
+// client_golang v1.20.0 has some bugs https://github.com/prometheus/client_golang/issues/1605, https://github.com/prometheus/client_golang/issues/1607
+// Stick to v1.19.1 until they are fixed.
+replace github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.19.1
```
ArthurSens, line 285: Taking a look at them right now!

jesusvazquez (approved):
LGTM!

krajorama (approved):
LGTM

PR #9133, opened 2024-08-29
Refactor the `PusherConsumer` interactions

gotjosh:
#### What this PR does

- Introduce a new `BatchingQueue` to reveal the intentions of a queue per shard.
- Removed the need to call `Close` as we pushed data into TSDB from the main loop. This was really confusing given we were using close as a semantic to "no more items are coming" and ensure any incomplete batches were done.
- Renamed and moved the `noopPusherCloser` which is more of an alternative way to push data without any sort of concurrency.
- Inlined a few methods to get rid of certain level of indirection that made the code harder to understand.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -24,21 +24,28 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+const shardForSeriesBuffer = 2000 // TODO dimitarvdimitrov 2000 is arbitrary; the idea is that we don't block the goroutine calling PushToStorage while we're flushing. A linked list with a sync.Cond or something different would also work
+
 type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
 type PusherCloser interface {
-	Pusher
+	PushToStorage(context.Context, *mimirpb.WriteRequest) error
+	// Calls to close are safe and will not be called concurrenctly.
 	Close() []error
 }
 
+// pusherConsumer receivers records from Kafka and pushes them to the storage.
+// Each time a batch of records is received from Kafka, we instantiate a new pusherConsumer, this is to ensure we can retry if necessary and know whether we have completed that batch or not.
 type pusherConsumer struct {
 	fallbackClientErrSampler *util_log.Sampler
 	metrics                  *pusherConsumerMetrics
 	logger                   log.Logger
 
-	pusher PusherCloser
+	kafkaConfig KafkaConfig
+
+	pusher Pusher
 }
 
 type pusherConsumerMetrics struct {
@@ -49,6 +56,7 @@ type pusherConsumerMetrics struct {
 	totalRequests         prometheus.Counter
 }
 
+// newPusherConsumerMetrics creates a new pusherConsumerMetrics instance.
 func newPusherConsumerMetrics(reg prometheus.Registerer) *pusherConsumerMetrics {
 	errRequestsCounter := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "cortex_ingest_storage_reader_records_failed_total",
@@ -78,20 +86,14 @@ func newPusherConsumerMetrics(reg prometheus.Registerer) *pusherConsumerMetrics 
 	}
 }
 
+// newPusherConsumer creates a new pusherConsumer instance.
 func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsumerMetrics, logger log.Logger) *pusherConsumer {
-	var p PusherCloser
-	if kafkaCfg.ReplayShards == 0 {
-		p = newNoopPusherCloser(metrics, pusher)
-	} else {
-		p = newMultiTenantPusher(metrics, pusher, kafkaCfg.ReplayShards, kafkaCfg.BatchSize)
-	}
-
 	return &pusherConsumer{
+		pusher:                   pusher,
+		kafkaConfig:              kafkaCfg,
 		metrics:                  metrics,
 		logger:                   logger,
 		fallbackClientErrSampler: util_log.NewSampler(kafkaCfg.FallbackClientErrorSampleRate),
-
-		pusher: p,
 	}
 }
 
@@ -147,6 +149,7 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) error {
 		}
 	}(ctx, records, recordsChannel)
 
+	writer := c.newStorageWriter()
 	for r := range recordsChannel {
 		if r.err != nil {
 			level.Error(spanlogger.FromContext(ctx, c.logger)).Log("msg", "failed to parse write request; skipping", "err", r.err)
@@ -154,7 +157,7 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) error {
 		}
 
 		// If we get an error at any point, we need to stop processing the records. They will be retried at some point.
-		err := c.pushToStorage(r.ctx, r.tenantID, r.WriteRequest)
+		err := c.pushToStorage(r.ctx, r.tenantID, r.WriteRequest, writer)
 		if err != nil {
 			cancel(cancellation.NewErrorf("error while pushing to storage")) // Stop the unmarshalling goroutine.
 			return fmt.Errorf("consuming record at index %d for tenant %s: %w", r.index, r.tenantID, err)
@@ -162,12 +165,11 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) error {
 	}
 
 	cancel(cancellation.NewErrorf("done unmarshalling records"))
-	return nil
-}
 
-func (c pusherConsumer) Close(ctx context.Context) error {
+	// We need to tell the storage writer that we're done and no more records are coming.
+	// err := c.close(ctx, writer)
 	spanLog := spanlogger.FromContext(ctx, log.NewNopLogger())
-	errs := c.pusher.Close()
+	errs := writer.Close()
 	for eIdx := 0; eIdx < len(errs); eIdx++ {
 		err := errs[eIdx]
 		isServerErr := c.handlePushErr(ctx, "TODO", err, spanLog)
@@ -177,10 +179,19 @@ func (c pusherConsumer) Close(ctx context.Context) error {
 			eIdx--
 		}
 	}
+
 	return multierror.New(errs...).Err()
 }
 
-func (c pusherConsumer) pushToStorage(ctx context.Context, tenantID string, req *mimirpb.WriteRequest) error {
+func (c pusherConsumer) newStorageWriter() PusherCloser {
+	if c.kafkaConfig.ReplayShards == 0 {
+		return newSequentialStoragePusher(c.metrics, c.pusher)
+	}
+
+	return newParallelStoragePusher(c.metrics, c.pusher, c.kafkaConfig.ReplayShards, c.kafkaConfig.BatchSize, c.logger)
+}
+
+func (c pusherConsumer) pushToStorage(ctx context.Context, tenantID string, req *mimirpb.WriteRequest, writer PusherCloser) error {
 	spanLog, ctx := spanlogger.NewWithLogger(ctx, c.logger, "pusherConsumer.pushToStorage")
 	defer spanLog.Finish()
 
@@ -188,7 +199,7 @@ func (c pusherConsumer) pushToStorage(ctx context.Context, tenantID string, req 
 
 	// Note that the implementation of the Pusher expects the tenantID to be in the context.
 	ctx = user.InjectOrgID(ctx, tenantID)
-	err := c.pusher.PushToStorage(ctx, req)
+	err := writer.PushToStorage(ctx, req)
 
 	// TODO dimitarvdimitrov processing time is flawed because it's only counting enqueuing time, not processing time.
 	c.metrics.processingTimeSeconds.Observe(time.Since(processingStart).Seconds())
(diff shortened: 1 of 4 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -24,21 +24,28 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+const shardForSeriesBuffer = 2000 // TODO dimitarvdimitrov 2000 is arbitrary; the idea is that we don't block the goroutine calling PushToStorage while we're flushing. A linked list with a sync.Cond or something different would also work
+
 type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
 type PusherCloser interface {
```
gotjosh, line 33: I also wanted to refactor this but turned out to be too big of a change and decided against it.

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +253,239 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
 
-	pusher.shards = shards
-	return pusher
+	p.start()
+
+	return p
 }
 
-func (p *shardingPusher) PushToStorage(ctx context.Context, request *mimirpb.WriteRequest) error {
+func (p *parallelStorageShards) ShardWriteRequest(ctx context.Context, request *mimirpb.WriteRequest) error {
```
gotjosh, line 373: I introduced different names here as I found it really strange that we had levels of equal method name that are not part of interface e.g. `shardingPusher` was not fulfilled the `PusherCloser` interface but had similar method names.

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +253,239 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
-		// drain errors to avoid blocking the shard loop
-	drainErrs:
-		for {
-			select {
-			case err := <-p.errs:
-				errs.Add(err)
-			default:
-				break drainErrs
-			}
-		}
```
gotjosh, line 368: In the end, I realised the value of these two selects - the trade-off here is that I have to `p.errs = make(chan error, p.numShards)` re-create the error channels once we're done as we use the closing of an channels as an indication that no more data was mean to go through. I'm open to the idea of bring this back if you feel strongly about it as I'm not sure how much it'll cost in terms of resources us to re-create the channels again.

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +253,260 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
+		p.shards[shard].AddToBatch(ctx, ts)
 
-		if len(s.Timeseries) < p.batchSize {
-			continue
-		}
-		p.unfilledShards[shard] = flushableWriteRequest{
-			WriteRequest: &mimirpb.WriteRequest{Timeseries: mimirpb.PreallocTimeseriesSliceFromPool()},
+		// After each addition, check for any errors in the error channel
+		select {
+		case err := <-p.errs:
```
dimitarvdimitrov, line 395: This way turned out to be super expensive. I had it like this initially, but replay speed falttened-out because we end up checking the channel about 1M times per second. So I ended up checking this channel once per flush.

Maybe `AddToBatch` can return an error and retain something like the previous behaviour

dimitarvdimitrov (commented):
nice one! I think there's a place where we can have a deadlock and a place where checking the channel  every time is too expensive. Otherwise, this is definitely an improvement. 

You mentioned some race condition, but I couldn't spot any. Isn't the test failure caused by different series sharding on different CPU architectures?

pkg/storage/ingest/pusher.go
gotjosh: Oh, that's clever - I like it!

pkg/storage/ingest/pusher.go
```diff
@@ -107,6 +109,8 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) error {
 		index    int
 	}
 
+	writer := c.newStorageWriter()
```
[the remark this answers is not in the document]
gotjosh: Done in https://github.com/grafana/mimir/pull/9133/commits/6f7409771850b93e0419065363ab476c94d9b8c2

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +253,260 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
+	close(q.ch)
+}
+
+// Channel returns the channel where the batches are pushed.
+func (q *BatchingQueue) Channel() <-chan FlushableWriteRequest {
+	return q.ch
+}
+
+func (q *BatchingQueue) push(fwr FlushableWriteRequest) {
+	q.ch <- fwr
```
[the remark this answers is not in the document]
gotjosh: Please take a look at https://github.com/grafana/mimir/pull/9133/commits/ff3c48c0b5e9ff4991b0a85272858b14ee87cd98 if the approach is OK, I'm going to do a quick scan / cleanup for comments and then follow through with this PR.

pkg/storage/ingest/pusher.go
gotjosh: Please take a look at https://github.com/grafana/mimir/pull/9133/commits/ff3c48c0b5e9ff4991b0a85272858b14ee87cd98 if the approach is OK, I'm going to do a quick scan / cleanup for comments and then follow through with this PR.

pkg/storage/ingest/pusher.go
```diff
@@ -24,21 +24,28 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+const shardForSeriesBuffer = 2000 // TODO dimitarvdimitrov 2000 is arbitrary; the idea is that we don't block the goroutine calling PushToStorage while we're flushing. A linked list with a sync.Cond or something different would also work
```
dimitarvdimitrov, line 27: I thought a bit about this - i think it shouldn't be larger than 1 or 2. 

We want to batch and flush in parallel. There's no point in batching faster than we can flush, right? If we flush 1 batch/sec, then batching 2 batches/sec doesn't make us faster. So if we're still flushing batch `N`, batch `N+1` is enqueued for flushing and we've just finished buffering batch `N+2` we're going too fast - in the time to flush `N` we've created `N+2` and `N+1`.

So I think a buffer of 1 or 2 is enough. What do you think?

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +252,281 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
+	ch    chan FlushableWriteRequest
+	errCh chan error
+	done  chan struct{}
+
+	currentBatch FlushableWriteRequest
+	batchSize    int
+}
+
+// NewBatchingQueue creates a new BatchingQueue instance.
+func NewBatchingQueue(capacity int, batchSize int) *BatchingQueue {
```
dimitarvdimitrov, line 456: do these need to be exported? I wouldn't be opposed ot moving them to a different package though. Something like `pkg/ingest/push`? If not, we can unexport them. Same for `FlushableWriteRequest`

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +252,281 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
+}
+
+// Done signals the queue that there is no more data coming for both the channel and the error channel.
+// It is necessary to ensure we don't close the channel before all the data is flushed.
+func (q *BatchingQueue) Done() {
+	close(q.done)
+}
+
+func (q *BatchingQueue) push(fwr FlushableWriteRequest) error {
+	for {
```
dimitarvdimitrov, line 524: the `for` loop looks redundant - the select will always return

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +252,281 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
 		}
 	}
-	for _, shard := range p.shards {
-		close(shard)
+
+	close(q.ch)
+	<-q.done
+
+	select {
+	case err := <-q.errCh:
```
dimitarvdimitrov, line 498: depending on the capacity there may be more errors in this channel. So by closing early we might ignore some of them (some may be internal and may need to be retried for example).

So maybe we can either: 1) have errCh always with capacity 0 or 1; 2) loop over errCh to make sure we've drained the errors

I think I prefer 1) with capacity 0 because it matches the current implementation that we don't continue pushing if we've encountered an internal error. wdyt?

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +252,281 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
+func newSequentialStoragePusher(metrics *pusherConsumerMetrics, pusher Pusher) sequentialStoragePusher {
+	return sequentialStoragePusher{
+		metrics: metrics,
+		pusher:  pusher,
+	}
+}
+
+// PushToStorage implements the PusherCloser interface.
+func (ssp sequentialStoragePusher) PushToStorage(ctx context.Context, wr *mimirpb.WriteRequest) error {
+	// TODO: What about time??
```
dimitarvdimitrov, line 272: what about it? :smile: did you mean to address this now or before we merge to `main`?

dimitarvdimitrov (commented):
i'll take another look of the whole PR in about 30m, going into a meeting now

pkg/storage/ingest/pusher.go
dimitarvdimitrov: forgot to mention that we don't have to finish this discussion in this PR. Just saw the line changed and decided to bring it up

pkg/storage/ingest/pusher.go
```diff
@@ -241,163 +252,281 @@ func (c pusherConsumer) shouldLogClientError(ctx context.Context, err error) (bo
...
-func (p *shardingPusher) close() error {
-	for shard, wr := range p.unfilledShards {
-		if len(wr.Timeseries) > 0 {
-			p.shards[shard] <- wr
+// Close closes the BatchingQueue, it'll push the current branch to the channel if it's not empty.
+// and then close the channel.
+func (q *BatchingQueue) Close() error {
+	if len(q.currentBatch.Timeseries) > 0 {
+		if err := q.push(q.currentBatch); err != nil {
+			return err
```
dimitarvdimitrov, line 490: so push doesn't actually ensure that we've pushed the current batch. If there was a buffered error, push will return that error and we'd end up discarding currentBatch.

Maybe instead of removing the `for` loop in `push()` as I suggested in another comment, you can kep the loop and just keep looping until the batch is sent and `continue` if we get an error.

pkg/storage/ingest/pusher_test.go
```diff
@@ -624,10 +626,129 @@ func TestShardingPusher(t *testing.T) {
...
+			defer queue.Done()
+			for b := range queue.Channel() {
+				batch = b
+				queue.ErrorChannel() <- nil
+			}
+		}()
+
+		// Close the queue, and the items should be flushed.
+		require.NoError(t, queue.Close())
+		wg.Wait()
```
dimitarvdimitrov, line 701: here wouldn't `queue.Close()` only return after that goroutine above has also returned? Why do we need the `WaitGroup`?

pkg/storage/ingest/pusher_test.go
```diff
@@ -624,10 +626,129 @@ func TestShardingPusher(t *testing.T) {
...
-			closeErr := shardingP.close()
+			closeErr := shardingP.Stop()
 			assert.ErrorIs(t, closeErr, tc.expectedCloseErr)
 			pusher.AssertNumberOfCalls(t, "PushToStorage", len(tc.expectedUpstreamPushes))
 			pusher.AssertExpectations(t)
 		})
 	}
 }
+
+func TestBatchingQueue(t *testing.T) {
```
dimitarvdimitrov, line 637: now that we also added error handling there I think it would make sense to add more tests for that. For example, what happens if the error channel fills up after the queue is already at capacity: any future calls to AddToBatch should return an error until we drain all errors, then AddToBatch should continue working. Also what happens if we push multiple errors after the last AddToBatch: they should all come out when we call Close. What if we push more errors than write requests? 

claude.ai is pretty good at writing these btw 😅

dimitarvdimitrov (commented):
ok, i've reviewed the whole PR. I think there's another place where we might lose data

pkg/storage/ingest/pusher.go
gotjosh: I haven't thought too much about this yet; I'll put this discussion for the next PR if you don't mind.

pkg/storage/ingest/pusher.go
gotjosh: I don't have any telemetry here so I hope you don't mind if the comment stays, I'm trying to avoid adding any until we're set this is the direction we want to take.

pkg/storage/ingest/pusher.go
gotjosh: I'm not sure I follow here, I thought what we wanted here is to return an error as soon as possible - so if it's true that we do have an error, why do we want to push the current batch if the whole write request is aimed to be retried?

pkg/storage/ingest/pusher.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9133/commits/7e56c2f2af28a31ac6f4e1dc832f4d89e287b583

I made them private, I don't think they warrant a different package / file just yet.

pkg/storage/ingest/pusher_test.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9133/commits/944a449cb9ce72c8d91efb182c683ed18316860c

pkg/storage/ingest/pusher.go
gotjosh: Discussed this offline, we went a different direction by collecting all errors on either push or close and keeping the capacity as is.

PR #9138, opened 2024-08-30
Fix issue where annotations would not be returned for sharded queries

charleskorn:
#### What this PR does

This PR fixes an issue where some annotations would not be returned for sharded queries.

This would occur if the annotation was generated during the evaluation of one of the shards. 

For example, the query `sum(rate(foo[5m]))` should return a `metric might not be a counter, name does not end in _total/_sum/_count/_bucket: "foo"` info annotation. When this query is sharded, the annotation is generated during the evaluation of one of the `sum(rate(foo{__query_shard__="x_of_y"}[5m]))` shards on queriers, but the query-frontend would not combine annotations from individual shards into the final result.

This PR changes the behaviour of query-frontends to accumulate and return annotations from each sharded query.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/querysharding.go
+++ b/pkg/frontend/querymiddleware/querysharding.go
@@ -8,6 +8,8 @@ package querymiddleware
 import (
 	"context"
 	"fmt"
+	"slices"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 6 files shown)


pkg/frontend/querymiddleware/querysharding.go
```diff
@@ -160,17 +164,25 @@ func (s *querySharding) Do(ctx context.Context, r MetricsQueryRequest) (Response
 	if err != nil {
 		return nil, mapEngineError(err)
 	}
+	// Note that the positions based on the original query may be wrong as the rewritten
+	// query which is actually used is different, but the user does not see the rewritten
+	// query, so we pass in an empty string as the query so the positions will be hidden.
```
jhesketh, line 169: Seems like a bug we should track in an issue. Ie: Reconstruct annotation positions after sharding.

jhesketh (approved):

pkg/frontend/querymiddleware/querysharding.go
charleskorn: I've created https://github.com/grafana/mimir/issues/9141 to track this

PR #9142, opened 2024-08-30
Mimir query engine: don't include `cpu: ...` line before every benchmark run in `tools/benchmark-query-engine`

charleskorn:
#### What this PR does

Go 1.23 added printing the CPU name to benchmarks on macOS, which clutters the output from `tools/benchmark-query-engine`.

This PR fixes the behaviour of `tools/benchmark-query-engine` to print the CPU name just once per run.

Given this is just an internal tooling change I haven't added a changelog entry.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


jhesketh (approved):
Change looks good to me, but the PR/commit message is very long. (As a nit I personally like the 50/70-75 formatting that the kernel and other projects use)

PR #9143, opened 2024-08-30
query-tee: optionally consider equivalent errors the same when comparing responses

charleskorn:
#### What this PR does

This PR extends query-tee's response comparison to accept equivalent but non-identical error messages as the same.

For example, Prometheus' engine and MQE return slightly different error messages in some circumstances, and Prometheus' engine is non-deterministic when selecting example series in some error messages. Both of these situations can cause query-tee to consider a response as different, but we usually don't care as long as the errors are equivalent.

This new behaviour is enabled by default but can be disabled with `-proxy.require-exact-error-match=true`.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -39,6 +39,7 @@ type ProxyConfig struct {
 	UseRelativeError                    bool
 	PassThroughNonRegisteredRoutes      bool
 	SkipRecentSamples                   time.Duration
+	RequireExactErrorMatch              bool
 	BackendSkipTLSVerify                bool
 	AddMissingTimeParamToInstantQueries bool
 	SecondaryBackendsRequestProportion  float64
@@ -64,6 +65,7 @@ func (cfg *ProxyConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&cfg.ValueComparisonTolerance, "proxy.value-comparison-tolerance", 0.000001, "The tolerance to apply when comparing floating point values in the responses. 0 to disable tolerance and require exact match (not recommended).")
 	f.BoolVar(&cfg.UseRelativeError, "proxy.compare-use-relative-error", false, "Use relative error tolerance when comparing floating point values.")
 	f.DurationVar(&cfg.SkipRecentSamples, "proxy.compare-skip-recent-samples", 2*time.Minute, "The window from now to skip comparing samples. 0 to disable.")
+	f.BoolVar(&cfg.RequireExactErrorMatch, "proxy.require-exact-error-match", false, "If true, errors will be considered the same only if they are exactly the same. If false, errors will be considered the same if they are considered equivalent.")
 	f.BoolVar(&cfg.PassThroughNonRegisteredRoutes, "proxy.passthrough-non-registered-routes", false, "Passthrough requests for non-registered routes to preferred backend.")
 	f.BoolVar(&cfg.AddMissingTimeParamToInstantQueries, "proxy.add-missing-time-parameter-to-instant-queries", true, "Add a 'time' parameter to proxied instant query requests if they do not have one.")
 	f.Float64Var(&cfg.SecondaryBackendsRequestProportion, "proxy.secondary-backends-request-proportion", 1.0, "Proportion of requests to send to secondary backends. Must be between 0 and 1 (inclusive), and if not 1, then -backend.preferred must be set.")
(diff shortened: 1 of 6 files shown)


tools/querytee/response_comparator.go
```diff
@@ -116,6 +118,60 @@ func (s *SamplesComparator) Compare(expectedResponse, actualResponse []byte) (Co
 	return ComparisonSuccess, nil
 }
 
+var errorEquivalenceClasses = [][]*regexp.Regexp{
+	{
+		// Invalid expression type for range query: MQE and Prometheus' engine return different error messages.
+		// Prometheus' engine:
+		regexp.MustCompile(`invalid parameter "query": invalid expression type "range vector" for range query, must be Scalar or instant Vector`),
+		// MQE:
+		regexp.MustCompile(`invalid parameter "query": query expression produces a range vector, but expression for range queries must produce an instant vector or scalar`),
```
jhesketh, line 127: Should MQE be changed to be the same for this one? (and later promql can be fixed to be more descriptive)

tools/querytee/response_comparator.go
```diff
@@ -116,6 +118,60 @@ func (s *SamplesComparator) Compare(expectedResponse, actualResponse []byte) (Co
...
+	},
+	{
+		// Binary operation conflict on right (one-to-one) / many (one-to-many/many-to-one) side: MQE and Prometheus' engine return different error messages, and there's no guarantee they'll pick the same series as examples.
+		// Even comparing Prometheus' engine to another instance of Prometheus' engine can produce different results: the series selected as examples are not deterministic.
+		// Prometheus' engine:
+		regexp.MustCompile(`found duplicate series for the match group \{.*\} on the (left|right) hand-side of the operation: \[.*\];many-to-many matching not allowed: matching labels must be unique on one side`),
+		// MQE:
+		regexp.MustCompile(`found duplicate series for the match group \{.*\} on the (left|right) side of the operation at timestamp \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z: \{.*\} and \{.*\}`),
```
jhesketh, line 135: Again should the engines be closer to the same but with regex to fix the side etc.

tools/querytee/response_comparator.go
```diff
@@ -116,6 +118,60 @@ func (s *SamplesComparator) Compare(expectedResponse, actualResponse []byte) (Co
...
+	},
+	{
+		// Same as above, but for left (one-to-one) / one (one-to-many/many-to-one) side.
+		// Prometheus' engine:
+		regexp.MustCompile(`multiple matches for labels: many-to-one matching must be explicit \(group_left/group_right\)`),
+		// MQE:
+		regexp.MustCompile(`found duplicate series for the match group \{.*\} on the (left|right) side of the operation at timestamp \d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z: \{.*\} and \{.*\}`),
```
jhesketh, line 142: Ditto

tools/querytee/response_comparator.go
charleskorn: We've intentionally decided to provide clearer error messages where we can in MQE, even if that means diverging from Prometheus' engine.

tools/querytee/response_comparator.go
charleskorn: Same as above: we intentionally decided to provide clearer error messages where we can in MQE.

tools/querytee/proxy.go
```diff
@@ -64,6 +65,7 @@ func (cfg *ProxyConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&cfg.ValueComparisonTolerance, "proxy.value-comparison-tolerance", 0.000001, "The tolerance to apply when comparing floating point values in the responses. 0 to disable tolerance and require exact match (not recommended).")
 	f.BoolVar(&cfg.UseRelativeError, "proxy.compare-use-relative-error", false, "Use relative error tolerance when comparing floating point values.")
 	f.DurationVar(&cfg.SkipRecentSamples, "proxy.compare-skip-recent-samples", 2*time.Minute, "The window from now to skip comparing samples. 0 to disable.")
+	f.BoolVar(&cfg.RequireExactErrorMatch, "proxy.require-exact-error-match", false, "If true, errors will be considered the same only if they are exactly the same. If false, errors will be considered the same if they are considered equivalent.")
```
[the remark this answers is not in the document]
charleskorn: I've renamed the flag to `compare-exact-error-matching`: `compare-...` to mirror the other comparison-related flags and I've retained the `exact` as that tells you exactly (heh) what it will do.

tools/querytee/response_comparator.go
jhesketh: Yep, but I'm wondering if we should be the same to ease testing, and open PRs upstream to fix up wording. Nonetheless, not a blocker here :-)

jhesketh (commented):
lgtm, but the new flag should probably be added to `docs/sources/mimir/manage/tools/query-tee.md` too.

charleskorn:
> lgtm, but the new flag should probably be added to `docs/sources/mimir/manage/tools/query-tee.md` too.

Done in 56f9674.

jhesketh (approved):

PR #7421, opened 2024-02-19
Helm / Gateway: Allow to configure whether or not NGINX binds IPv6

wilfriedroset:
#### What this PR does

This PR allow to configure whether or not the gateway's NGINX binds IPv6.
The primary reason to do that is to avoid to duplicate the whole configuration to remove the line `listen [::]:{{ include "mimir.serverHttpListenPort" . }};`
I've purposedely not done it for `nginx` as it is marked as deprecated.

Without this change the deployment of the gateway will fail with the following message if IPv6 is disable at the system level in the whole k8s cluster:
```
/docker-entrypoint.sh: No files found in /docker-entrypoint.d/, skipping configuration
2024/02/19 12:23:08 [emerg] 1#1: socket() :8080 failed (97: Address family not supported by protocol)
nginx: [emerg] socket() :8080 failed (97: Address family not supported by protocol)
```

#### Which issue(s) this PR fixes or relates to

Fixes #7210 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

Issue #4560, opened 2023-03-22
[EPIC] Grafana Mimir Helm chart 6.0 tracker


krajorama:
#### Is your feature request related to a problem? Please describe.

Keep track of things we might want to introduce with 6.0 of mimir-distributed Helm chart.

> **Note** non breaking items can be done before that

#### Describe the solution you'd like

Example, not agreed yet:
- bump minimum required kubernetes version
- nginx/gateway configuration only via [unified gateway configuration](https://grafana.com/docs/helm-charts/mimir-distributed/latest/migration-guides/migrate-to-unified-proxy-deployment/)
- topologySpreadConstraints is an array (don't allow single object anymore)
- get rid of [single zone](https://grafana.com/docs/helm-charts/mimir-distributed/latest/migration-guides/migrate-from-single-zone-with-helm/) ingesters/store-gateways (? this would not allow RF=1)
- changes related to alignment between loki/tempo/mimir/rollout-operator charts.
  - naming things (maybe even align on bitnami naming? https://github.com/bitnami/charts/tree/main/bitnami/grafana-mimir) 
  - globals (!)
  - reusable library - templates for basic deployments and such
- opinionated metamonitoring setup: https://github.com/grafana/mimir/issues/3315
  - switch to Grafana Agent Flow: already planned: #5860 
- bring in auto scaling from GL jsonnet - once stable #4687 
- memberlist cluster label migration (!) https://github.com/grafana/mimir/issues/2865
- change store-gateway's podManagementPolicy to `Parallel`. This will match jsonnet and the ingester statefulSet in helm - see https://github.com/grafana/mimir/pull/5757
- change compactor's podManagementPolicy to `Parallel`. See #5902 

#### Describe alternatives you've considered

n/a

krajorama:
Timing wise, I'd prefer to not have breaking changes too often. The last was end of '22 with zone aware replication, so end of '23 would give us and users plenty of time to prepare

osg-grafana:
Removing `type/docs` from Epic, because the issues themselves have the label where needed.

PR #9144, opened 2024-08-30
Give option to disable IPv6 in nginx config in helm chart

Raboo:
#### What this PR does

The helm chart value section `nginx.nginxConfig` has no way of turning off IPv6 listener unless modifying the whole `nginx.nginxConfig.file` multi line value.

This PR would make it an value that can easily be turned true or false.

I haven't created an issue for this.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -29,7 +29,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## main / unreleased
 
-* [ENHANCEMENT] Add support for turning off IPv6 listener in nginx.
+* [ENHANCEMENT] Add support for turning off IPv6 listener in nginx. #9144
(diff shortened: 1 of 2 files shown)


narqo:
Wasn't that what #7421 has fixed already?

Raboo:
Oh crap, I missed that `.nginx` is a deprecated section...
Well if you want merge, otherwise close the PR.
I will switch to `.gateway.nginx`.

Well actually, if you don't specify `.gateway.enabledNonEnterprise: true`, the helm chart defaults into using the nginx configured by `.nginx`.

If you truly want to deprecate `.nginx`, then it should not be used it by default when not specified.

dimitarvdimitrov:
> If you truly want to deprecate `.nginx`, then it should not be enabled by default when not actively chosen to.

unfortunately, that'd be a breaking change and the upgrade will cause some downtime. So we want to release a major version which does the switch (#4560)

Closing this for now

PR #9147, opened 2024-08-30
 blockbuilder: simplify KafkaClient management

narqo:
#### What this PR does

_This one seats atop #9068 for now_

Following the discussion in #9068 this PR updates the `blockbuilder` package, to reuse the Kafka configuration and the client from the `storage/ingest` package.

Note that here I'm getting rid of `findOffsetsToStartAt` helper. It's turned out that after the several rounds of fixes and refactorings, we no longer need to explicitly set the client's starting position. Instead, here I'm going back to what I'd tried in grafana/mimir#8684. That is, previously I gave up on the idea of [`kgo.ConsumeResetOffset`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#ConsumeResetOffset) because fixing lag calculation seemed too complicated. Now as I've already fixed lag calculation with the introduction of `fallbackOffset`, and looking at it with a fresh eye, we are in a good position to simplify a bunch of internals here.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/config.go
```diff
@@ -20,14 +21,14 @@ type Config struct {
...
+	ConsumerGroup         string        `yaml:"consumer_group"`
 	ConsumeInterval       time.Duration `yaml:"consume_interval"`
 	ConsumeIntervalBuffer time.Duration `yaml:"consume_interval_buffer"`
 	LookbackOnNoCommit    time.Duration `yaml:"lookback_on_no_commit" category:"advanced"`
 
-	Kafka KafkaConfig `yaml:"kafka"`
-
-	// BlocksStorageConfig is defined outside the block-builder config and is injected dynamically.
-	BlocksStorageConfig tsdb.BlocksStorageConfig `yaml:"-"`
+	// Config parameters defined outside the block-builder config and are injected dynamically.
```
dimitarvdimitrov, line 29: do you also want to remove the `cfg.Kafka.RegisterFlagsWithPrefix` call below?

dimitarvdimitrov (approved):
nice, LGTM. only one comment on registering the config

PR #9149, opened 2024-08-30
Add proposal for reducing the cost of multitenancy

wilfriedroset:
#### What this PR does

This PR adds a new proposal that aims to reducing the cost of multitenancy

#### Additional context

'm currently working on a usecase where I would like to store the metrics across several thousand of small tenants. The number of tenants to store in a given cluster is 400k and each tenant is expected to have from 50 to 200 actives series at a scrape interval of 15sec, this would make 80M actives series with a replication factor=3 (aka RF).

Mimir is known to efficiently scale for both active series and sample/sec, 80M is expected to be within Mimir working range. However the number of tenants come with an overhead which impact the performance/resources consumption which makes this scenario hard to run in production.

I've run the following proof-of-concept:
* Mimir 2.13.0 is deployed on k8s with [small.yaml](https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/small.yaml) recommendation
  * see `values.yaml` after.
  * The resources limits of the ingester have been removed. 
```yaml
global:
  podAnnotations:
    profiles.grafana.com/cpu.port_name: http-metrics
    profiles.grafana.com/cpu.scrape: "true"
    profiles.grafana.com/goroutine.port_name: "http-metrics"
    profiles.grafana.com/goroutine.scrape: "true"
    profiles.grafana.com/memory.port_name: "http-metrics"
    profiles.grafana.com/memory.scrape: "true"

alertmanager:
  persistentVolume:
    enabled: true
  replicas: 1
  resources:
    limits:
      memory: 1.4Gi
    requests:
      cpu: 1
      memory: 1Gi
  statefulSet:
    enabled: true

compactor:
  persistentVolume:
    size: 20Gi
  resources:
    limits:
      memory: 2.1Gi
    requests:
      cpu: 1
      memory: 1.5Gi

distributor:
  replicas: 2
  resources:
    limits:
      memory: 5.7Gi
    requests:
      cpu: 2
      memory: 4Gi

ingester:
  persistentVolume:
    size: 50Gi
  replicas: 3
  resources:
    # limits:
    #   memory: 12Gi
    requests:
      cpu: 3.5
      memory: 8Gi
  topologySpreadConstraints: {}
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchExpressions:
              - key: target # support for enterprise.legacyLabels
                operator: In
                values:
                  - ingester
          topologyKey: "kubernetes.io/hostname"

        - labelSelector:
            matchExpressions:
              - key: app.kubernetes.io/component
                operator: In
                values:
                  - ingester
          topologyKey: "kubernetes.io/hostname"

  zoneAwareReplication:
    enabled: false
    topologyKey: "kubernetes.io/hostname"

admin-cache:
  enabled: true
  replicas: 2

chunks-cache:
  enabled: true
  replicas: 2

index-cache:
  enabled: true
  replicas: 3

metadata-cache:
  enabled: true

results-cache:
  enabled: true
  replicas: 2

minio:
  enabled: true

overrides_exporter:
  replicas: 0
  resources:
    limits:
      memory: 128Mi
    requests:
      cpu: 100m
      memory: 128Mi

querier:
  replicas: 1
  resources:
    limits:
      memory: 5.6Gi
    requests:
      cpu: 2
      memory: 4Gi

query_frontend:
  replicas: 1
  resources:
    limits:
      memory: 2.8Gi
    requests:
      cpu: 2
      memory: 2Gi

ruler:
  replicas: 0
  resources:
    limits:
      memory: 2.8Gi
    requests:
      cpu: 1
      memory: 2Gi

store_gateway:
  persistentVolume:
    size: 10Gi
  replicas: 3
  resources:
    limits:
      memory: 2.1Gi
    requests:
      cpu: 1
      memory: 1.5Gi
  topologySpreadConstraints: {}
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchExpressions:
              - key: target # support for enterprise.legacyLabels
                operator: In
                values:
                  - store-gateway
          topologyKey: "kubernetes.io/hostname"

        - labelSelector:
            matchExpressions:
              - key: app.kubernetes.io/component
                operator: In
                values:
                  - store-gateway
          topologyKey: "kubernetes.io/hostname"
  zoneAwareReplication:
    enabled: false
    topologyKey: "kubernetes.io/hostname"

nginx:
  replicas: 1
  resources:
    # limits:
    #   memory: 731Mi
    requests:
      cpu: 1
      memory: 512Mi

# Grafana Enterprise Metrics feature related
admin_api:
  replicas: 1
  resources:
    limits:
      memory: 128Mi
    requests:
      cpu: 100m
      memory: 64Mi

gateway:
  replicas: 1
  resources:
    limits:
      memory: 731Mi
    requests:
      cpu: 1
      memory: 512Mi

metaMonitoring:
  dashboards:
    enabled: true
  serviceMonitor:
    enabled: true
    interval: 5s
  grafanaAgent:
    enabled: false
    installOperator: false
```
* Every 5 min one sample per tenant is sent to Mimir, the number tenant is gradually increased from 1 to 5k
  * see the go sample after. (note: the sample is far from being perfect 😅)
```go
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
	"net/http"
	"sync"
	"time"

	"github.com/golang/snappy"
	"github.com/prometheus/prometheus/prompb"
)

func generateSample() ([]byte, error) {
	// Create the WriteRequest object with metric data populated.
	wr := prompb.WriteRequest{
		Timeseries: []prompb.TimeSeries{
			{
				Labels: []prompb.Label{
					{
						Name:  "__name__",
						Value: "foo",
					},
				},
				Samples: []prompb.Sample{
					{
						Timestamp: time.Now().UnixNano() / int64(time.Millisecond),
						Value:     100.0,
					},
				},
			},
		},
	}
	// Marshal the data into a byte slice using the protobuf library.
	data, err := wr.Marshal()
	if err != nil {
		return nil, err
	}
	encoded := snappy.Encode(nil, data)
	return encoded, nil
}

func remoteWrite(endpoint string, encodedSample []byte, tenantID int) error {
	body := bytes.NewReader(encodedSample)
	req, err := http.NewRequest("POST", endpoint, body)
	if err != nil {
		return err
	}

	req.Header.Set("Content-Type", "application/x-protobuf")
	req.Header.Set("Content-Encoding", "snappy")
	req.Header.Set("X-Prometheus-Remote-Write-Version", "0.1.0")
	req.Header.Set("X-Scope-OrgID", fmt.Sprintf("tenant_%d", tenantID))

	resp, err := http.DefaultClient.Do(req)
	defer resp.Body.Close()
	if err != nil {
		return err
	}
	return nil
}

type Annotation struct {
	Text string   `json:"text"`
	Time int64    `json:"time"`
	Tags []string `json:"tags"`
}

func addAnnotation(endpoint, username, password string, tenantsCount int) error {
	annotation := Annotation{
		Text: fmt.Sprintf("running with %d", tenantsCount),
		Tags: []string{"tenant"},
		Time: time.Now().UTC().UnixMilli(),
	}
	marsheldAnnotation, err := json.Marshal(annotation)
	if err != nil {
		return err
	}
	req, err := http.NewRequest("POST", endpoint, bytes.NewBuffer(marsheldAnnotation))
	if err != nil {
		return err
	}

	req.Header.Set("Content-Type", "application/json")
	req.SetBasicAuth(username, password)

	resp, err := http.DefaultClient.Do(req)
	defer resp.Body.Close()
	if err != nil {
		return err
	}
	return nil
}

func main() {
	concurrentGoroutines := make(chan struct{}, 10)

	tenantsCount := []int{1, 5, 10, 20, 50, 75, 100, 150, 200, 250, 300, 500, 600, 700, 800, 900, 1000, 2000, 3000, 4000, 5000}
	for _, tenantCount := range tenantsCount {
		fmt.Printf("== Starting batch with %d tenants\n", tenantCount)

		err := addAnnotation("http://localhost:3000/api/annotations", "admin", "prom-operator", tenantCount)
		if err != nil {
			fmt.Printf("failed to create an annotation: %v\n", err)
		}

		encodedSample, err := generateSample()
		if err != nil {
			panic(err) // we can't do anything without a sample
		}

		var wg sync.WaitGroup

		tenantID := 0
		for tenantID < tenantCount {
			wg.Add(1)
			go func(t int) {
				defer wg.Done()
				// Throttle the number of request to send and avoid slaming the distributor
				concurrentGoroutines <- struct{}{}
				defer func() { <-concurrentGoroutines }()

				fmt.Printf("(batch with %d tenants) Remote write for tenant %d\n", tenantCount, t)
				err = remoteWrite("http://localhost:8080/api/v1/push", encodedSample, t)
				if err != nil {
					fmt.Printf("(batch with %d tenants) failed to push sample for tenantID: %d\n", tenantCount, t)
				}
			}(tenantID)
			tenantID = tenantID + 1
		}
		fmt.Printf("Batch with %d tenants complete, waiting for goroutine\n", tenantCount)
		wg.Wait()
		time.Sleep(5 * time.Minute)
	}
}
```

Here is a screenshot of the resources consumption evolution during the benchmark that can be read as follow:
* The blue annotations shows each time the number of tenants is increased
* The first panel "CPU" query is
```
sum(
    node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate{cluster="", namespace="mimir-test"}
  * on(namespace,pod)
    group_left(workload, workload_type) namespace_workload_pod:kube_pod_owner:relabel{cluster="", namespace="mimir-test", workload="mimir-test-ingester", workload_type="statefulset"}
) by (pod)
```
* The second panel "Memory" query is
```
sum(
    container_memory_working_set_bytes{cluster="", namespace="mimir-test", container!="", image!=""}
  * on(namespace,pod)
    group_left(workload, workload_type) namespace_workload_pod:kube_pod_owner:relabel{cluster="", namespace="mimir-test", workload="mimir-test-ingester", workload_type="statefulset"}
) by (pod)
```
* The last panel "In-memory series per ingester"query is
```
sum by (pod) (cortex_ingester_memory_series_created_total{cluster=~"mimir-test", job=~"(mimir-test)/((ingester.*|cortex|mimir|mimir-write.*))"})
- sum by (pod) (cortex_ingester_memory_series_removed_total{cluster=~"mimir-test", job=~"(mimir-test)/((ingester.*|cortex|mimir|mimir-write.*))"})
```

Capture d’écran 2024-08-20 à 16 34 24

Here are a couple of flamgraph during the benchmark:
* **memory:inuse_space**: https://flamegraph.com/share/ce24e81a-5f04-11ef-ab1b-b6898751f631
* **memory:inuse_objects**: https://flamegraph.com/share/08a70331-5f05-11ef-ab1b-b6898751f631
* **memory:alloc_space**: https://flamegraph.com/share/21f2b848-5f05-11ef-aba3-6a3d1814cbe4
* **memory:alloc_objects**: https://flamegraph.com/share/3eb0f66b-5f05-11ef-aba3-6a3d1814cbe4

This proof-of-concept demonstrate that each tenant cost ~6MB of ram and it is linear throughout all the proof-of-concept, I assume the samples stored in memory are neglectable.

For the usecase described at the beginning I would need `6MB * 400k * RF = 7,2TB` of memory only to create the tenants in the ingester. The replication factor, the shuffle sharding and the replica count could also impact the overhead. The worst case scenario is expected to be `6MB * tenants_count * ingesters_count`.

I've not benchmarked the compactor and the store-gateway but they might be impacted the overhead.

Similarly the number files created on disk and the objects in the bucket rapidly increase.

On top of the above resources, we should account for the regular capacity planning.

According to the[ capacity planning ](https://grafana.com/docs/mimir/latest/manage/run-production-environment/planning-capacity/#ingester)we should planned for 80M actives series * RF * 2.5GB / 300k active series = 2TB

My understanding is that Mimir creates a Prometheus' TSDB in memory for each tenant and store the sample accordingly.

One way to workaround this overhead is to store multiple tenant in a single TSDB (e.g: use a label to identify the owner of a given metrics).

However, this might not be as easy as it looks in particular for the rate-limiting and usage statistics, especially if this is done outside of Mimir.

I reckon it would be a good addition to Mimir to facilitate the multi-tenancy (not to be mistaken with actives series or ingestion rate) at such scale. Hence the attached proposal.

At OVHcloud, we are willing to pick up the work if the maintainers see the proposal worthy 🫡 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/proposals/reduce-multitenancy-cost.md
+++ b/docs/proposals/reduce-multitenancy-cost.md
@@ -1,0 +1,221 @@
+---
+title: "XXXX: Tenancy multiplexing"
+description: "Tenancy multiplexing"
+draft: true
+---
+
+# XXXX: Tenancy multiplexing
+
+**Authors:** @wilfriedroset @vaxvms @bubu11e
+
+**Date:** 09/2024
+
+**Sponsor(s):** @username of maintainer(s) willing to shepherd this MID
+
+**Type:** Feature
+
+**Status:** Draft
+
+**Related issues/PRs:**
+
+---
+
+## Background
+
+The `compactor` is a component that improves query performance by compacting multiple blocks for a given tenant in larger optimized ones.
+The `ingester` is a component that is responsible for the persistency of the data of each tenant in a TSDB.
+The `store-gateway` is a component that is responsible for providing access to the object storage where the blocks are stored.
+
+## Problem statement
+
+For each new tenant hosted on a given ingester, a new TSDB is created/opened. With this new TSDB to manage, Mimir consumes more resources (e.g: CPU, RAM, files on disk, objects in the bucket). With the load balancing, replication factor and shuffle sharding the overhead resources consumed by Mimir can be multiplied by the number of replicas of each component. Example: - For each new tenant, the `ingester` consumes ~6MB of ram - Mimir is currently running with 3 ingesters and a replication factor of 3 - The per-tenant RAM overhead is 18MB (3 ingesters _ 1 tenant _ 6MB/tenant)
+
+Given an increasing number of tenants and the horizontal scaling of Mimir, this overhead becomes bigger.
+The problem affects all components that are sensible to the tenant count and their associated TSDB.
+
+## Goals
+
+- From an end-user's point of view, the feature set offered by Mimir works the same as today and there is no breaking changes.
+- Reduce the impact on performance of a high number of Mimir tenants, and therefore enable Mimir to support a large number of tenants.
+
+## Non-Goals
+
+- Provide a migration path from the regular multitenancy to the tenancy multiplexing.
+
+## Proposals
+
+### Proposal 0: Do nothing
+
+In this proposal, it is the operator's responsibility to implement a Mimir gateway to expose tenancy multiplexing using a unique Mimir tenant as the backend.
+This gateway would modify and proxy requests to the Mimir cluster.
+
+The challenge doing so is to re-implement all the tenant limits and isolation built into Mimir.
+Moreover, given the label convention used to enable tenancy multiplexing, the proxy should be implemented in such a way that this private label is not returned in the response to a given Prometheus query.
+
+Note: the gateway of this proposal must not be confused with the GEM gateway.
+
+#### Write path
+
+All the limits enforced by the distributor should be straightforward to implement in another frontend, as the distributor is a stateless process. If the gateway is coded in Go, the code could even be partially reused.
+
+The limits enforced [by the ingester](https://github.com/grafana/mimir/blob/main/pkg/ingester/ingester.go#L102-L109) could be more complex.
+
+#### per_user_series_limit
+
+Gateway could keep in memory a representation of the series already owned by the tenant. This representation should: - be fast to access (as we need to access it for every serie of every write request) - have a small footprint (those information will be exchanged between members of the gateway cluster) - be mergeable (to aggregate information from various members of the gateway cluster) - serializable (for exchange purposes)
+
+A naive approach would be to store in memory all the metadata associated with all the series of every tenants, but this would waste hundreds of gigabytes of memory and would be hard to exchange between frontend cluster members.
+
+A better approach would be to use [HyperLogLog++ estimators](https://research.google/pubs/hyperloglog-in-practice-algorithmic-engineering-of-a-state-of-the-art-cardinality-estimation-algorithm/). Such an estimator is able to keep a representation of the cardinality of a data set using a limited footprint (a few kilobytes per estimator), and is serializable and mergeable with a limited error margin (depends of the precision of the estimator).
+
+Sharing those estimators between members of the gateway would ensure us all members share an up-to-date and sufficiently accurate representation of each tenant's cardinality. Communication between the horizontally scaled gateway could be done via memberlist and gRPC or a pub/sub service like Kafka.
+However, the amount of data to be exchanged between members could be rather large, as we need one estimator per meta-tenant. A way to mitigate this issue could be to propagate only recently updated estimators.
+For example exchange occurs every 15 seconds, for all estimators updated within the last minute.
+
+#### per_metric_series_limit
+
+We could implement this limit in the same way as for `per_user_series_limit`, but this would lead to significant memory consumption as the cardinality of the series is proportional to the cardinality of the tenants.
+
+#### sample-out-of-order
+
+This is not tenant-related, it could be forwarded from the backend.
+
+#### sample-too-old
+
+This is not tenant-related, it could be forwarded from the backend.
+
+#### sample-too-far-in-future
+
+This is not tenant-related, it could be forwarded from the backend.
+
+#### new-value-for-timestamp
+
+This is not tenant-related, it could be forwarded from the backend.
+
+#### sample-out-of-bounds
+
+This is not tenant-related, it could be forwarded from the backend.
+
+#### invalid-native-histogram
+
+This is not tenant-related, it could be forwarded from the backend.
+
+After checking the limits, the gateway adds a dedicated label to each metric containing the name of the meta tenant.
+
+### Read path
+
+The gateway should modify queries on the fly and add the correct label/value to all nodes of the query.
+
+**Pros:**
+
+- Nothing to be done on Mimir side and all the logic is kept apart
+- Operators are free to build any custom made solution
+
+**Cons:**
+
+- The gateway adds an extra hop and therefore latency
+- The limits validation are computed several times
+- All distinct operators faced with the same situation have to solve it themselves
+
+### Proposal 1: Implement tenancy multiplexing through label
+
+In this proposal, we add the possibility for Mimir to host multiples tenants in a single TSDB instead of having one TSDB per tenant. The distinction is done via a label.
+The tenancy multiplexing can be switched on via a configuration file that goes along with the regular multitenancy one.
+
+Example:
+
+```yaml
+# When set to true, incoming HTTP requests must specify tenant ID in HTTP
+# X-Scope-OrgId header. When set to false, tenant ID from -auth.no-auth-tenant
+# is used instead.
+# CLI flag: -auth.multitenancy-enabled
+[multitenancy_enabled: <boolean> | default = true]
+
+# The mode of multitenancy to use. Either "standard" or "multiplexing".
+# When using the "multiplexing" multitenancy several tenants share the same TSDB.
+# CLI flag: -auth.multitenancy-mode
+[multitenancy_mode: <string> | default = "standard"]
+```
+
+All components responsible for manipulating the blocks and/or the TSDB must by adjusted to take into account the tenancy multiplexing.
+
+#### Write path
+
+On the write path we should create a middleware that is responsible for taking the tenant ID and adding the value as a label on each series.
+This middleware is only used when "multiplexing" multitenancy mode is selected.
+
+Option: in order to avoid having a single large TSDB we could distribute the tenants over a fixed numbers of TSDBs. Doing so would reduce the number of tenants per TSDB.
+
+Neither the distributor nor the compactor needs to be modified as their current behavior is not impacted by the multiplexing tenancy.
+
+#### Read path
+
+Ingester and store-gateway use the tenant ID as a matcher when querying storage instead of opening the tenant's TSDB
+
+Ruler still work as-is (see <https://grafana.com/docs/mimir/latest/references/architecture/components/ruler/>): - In internal mode it runs its own distributor. Given the following proposal, we must ensure that we apply the same relabeling as in the rest of the write path - In remote mode, only the read path is changed. Indeed, filtering is performed by the ingester and store-gateway using the tenant ID.
+
+Alertmanager multitenancy still working as-is
+
+    - Ruler call the alertmanager with the tenant ID
+
+The tenants federation still works as-is.
+
+#### Limits
+
+Depending on the component in charge of enforcing the limits, we might have to adapt them:
+
+- distributor: no change expected
+- ingester: max_series_per_user and max_series_per_metrics have to be computed differently
+- querier/query-frontend/query-scheduler: no change expected
+- compactor: retention is per TSDB. It can be enforced in the read path by clamping the "start" of the query. Depending on the compute cost vs the storage cost, this might be more cost-efficient to keep the block for longer instead of processing them to discard the data that have exceeded their retention.
+- ruler: no change expected
+- alertmanager: no change expected
+
+#### Metrics
+
+Depending on the component in charge of metrics exposition, we might have to adapt them:
+
+- ingester: TSDB metrics aren't per user anymore.
+- compactor: metrics are per TSDB.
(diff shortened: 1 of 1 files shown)


56quarters:
Hello,

Thank you for this proposal. We've discussed doing something similar to this, proposal 1, internally a few times but never actually moved forward due to the large amount of work and complexity this would entail. Depending on your specific use case, you may be able to get most of the value of this change with a minimal amount of work. Some thoughts are below about why I don't think enforcing limits in proposal 0 or proposal 1 in its entirety is worth the amount of effort it entails.

To start, both proposal 0 as written and proposal 1 are a significant amount of work to make happen and potentially make Mimir more complicated to understand and operate. The idea of tenants being a trust boundary _and_ how quality of service is enforced is threaded throughout all of Mimir. Enforcing most limits at something other than the tenant level is going to be prohibitively difficult. I think it's possible to get most of what (I'm assuming) you are looking for out of a simpler version of proposal 0 that works similar to the [GEM LBAC feature](https://grafana.com/docs/enterprise-metrics/latest/tenant-management/lbac/).

At Grafana, we make "usage" metrics to our cloud users. These metrics are similar to the usecase you describe: thousands of different users/tenants but only a few hundred series for each one. This information originates from metrics emitted by ingesters and distributors with a `user` label. We have dedicated Prometheus instances in each GEM/Mimir cluster that scrape these particular metrics, which are then written  to a dedicated "usage" GEM/Mimir cluster. The interesting thing to note here is that while we ensure each series has a `user` label and a few other bits of metadata: all series are written to a single tenant in the GEM/Mimir usage cluster.

When users access these metrics, queries are sent to Billing Gateways which use per-user authentication supplied to inject an LBAC matcher into the query. These queries are all sent to the same single tenant mentioned above but the LBAC matchers limit access to only series associated with a single user. The LBAC feature handles transparently dealing with the `user` label like you mention the gateway would do.

We used to use separate tenants per user for billing information but found the overhead per tenant prohibitive like you mention. Because this data all lives in a single tenant in GEM/Mimir, we don't really enforce that many limits on it: we control the number of series since we operate the Prometheus instances writing data. We can't really control the way this data is queried but this isn't a problem in practice due to the nature of the data: an extremely small number of series.

In summary:
* I think the amount of work entailed as written is enough that there's not much chance of this change being made to Mimir
* Based on what I assume the problem you are solving is, I think scoping down this proposal would allow you to solve this problem with _only_ a gateway to transparently add a per-customer label on metrics stored under a single tenant and query by that label on the read path. 

Thank you

wilfriedroset:
Thank you @56quarters for your feedbacks I really appreciate the time you spent on reading the whole piece and answering.

In our usecase each tenant will be used for a different usecase and probably by different end-users. This is the main reason why we want to have a rate limiting.

> I think it's possible to get most of what (I'm assuming) you are looking for out of a simpler version of proposal 0 that works similar to the [GEM LBAC feature](https://grafana.com/docs/enterprise-metrics/latest/tenant-management/lbac/).

We are also evaluating the LBAC solution but we have yet to find a way to address both the ruler and alertmanager.
Manipulating the remote write payload to adding a label on the fly is fairly straighforward. On the read path we can manipulate the query before it reaches the query-frontend to add the correct label in all places.

However, for the rules and the alerts not only do we need to manipulate the path but the yaml content itself and manipulate it.

> When users access these metrics, queries are sent to Billing Gateways which use per-user authentication supplied to inject an LBAC matcher into the query. These queries are all sent to the same single tenant mentioned above but the LBAC matchers limit access to only series associated with a single user. The LBAC feature handles transparently dealing with the user label like you mention the gateway would do.

I'm no expert in Grafana Cloud but are users allowed to define the rules and alerts and those metrics or are they only read-only?

56quarters:
> I'm no expert in Grafana Cloud but are users allowed to define the rules and alerts and those metrics or are they only read-only?

In this case the recording rules that operate on the billing data are controlled by Grafana. The users are consumers of the metrics generated by them.

wilfriedroset:
This is the key difference between our usecases. On my end I would like end-users to be able to use the whole feature set of Mimir which mean rulers and alertmanager :) 

Doing it outside of Mimir seems harder due to this need.

56quarters:
> This is the key difference between our usecases. On my end I would like end-users to be able to use the whole feature set of Mimir which mean rulers and alertmanager :)

Ah, I see. I'm assuming you are already using shuffle sharding to reduce the number of ingesters each tenant is spread over? In this case, my preference would be for proposal 2. 

My reasoning is:

* You need the full set of features that tenants offer but the overhead is too high. Proposal 2 is to reduce the overhead.
* The work needed to get there is unknown but likely not as much as the other proposals.
* Even if we don't reduce overhead by an order of magnitude, Mimir will still be in a better place. Compared to adding tenant functionality by label which is all-or-nothing.

In any case, I think it makes sense to merge this proposal.

docs/proposals/reduce-multitenancy-cost.md
```diff
@@ -0,0 +1,221 @@
...
+- Unless modifications, all tenants within the same TSDB share some of the limits. This might be a blocker especially for the retention.
+- There is no easy migration back and forth between regular tenancy and tenancy multiplexing.
+- It is all or nothing as there is no way to define the mapping between the tenant ID and the TSDB.
+- Compactor blocks upload (<https://grafana.com/docs/mimir/latest/configure/configure-tsdb-block-upload/>) will not work properly as block are uploaded to LTS storage as-is. There is no way to add the metatenancy label without rewriting the block.
+- Compactor tenant deletion will not work as it relies on mapping the tenant to a dedicated TSDB.
+- Migrating a metatenant from one TSDB/cluster to another is not straightforward nor easy.
+
+### Proposal 2: Reduce the overhead by at least an order of magnitude
+
+Disclaimer: We list this proposal for the sake of completeness but our knowledge is limited and we don't know how realistic it is.
```
bboreham, line 201: It would be great to have a "tuning" page in the docs, with sections aimed at different scenarios.  
Some ideas for your scenario:
 * Reduce `-blocks-storage.tsdb.head-chunks-write-buffer-size-bytes`, default 4MB.  E.g. try 1MB, or 128KB.
 * Reduce `-blocks-storage.tsdb.stripe-size`, default 16384. E.g. try 256 or even 64.

wilfriedroset:
Thank you @bboreham  & @56quarters  for your feedbacks. This is a good place to start and I would be happy to contribute a tuning page.
Here is the list of the tuning we have discovered so far:
* grpc_compression (gzip or snappy) comes with a price if you are limited by your cpu power consider disabling it or switching to the less cpu heavy namely snappy
* wal_compression_enabled same as above
* block_ranges_period could be reduce (e.g 1h) when mimir have a significant number of sample per time series within a block. we are doing this for our 1Hz cluster (1 sample per second per time series)
* configure the memcache to use ext_path to increase the cache size (similar to loki: https://grafana.com/blog/2023/08/23/how-we-scaled-grafana-cloud-logs-memcached-cluster-to-50tb-and-improved-reliability/)

With that we could also add @bboreham  recommendations.

WDYT?

56quarters (approved):

PR #9152, opened 2024-08-30
kafka replay speed: handle metadata refreshes and OffsetOutOfRange errors

dimitarvdimitrov:
#### What this PR does

1. Deploys 3 kafka brokers with 2x replication so that you can simulate electing a new kafka broker for a partition locally
2. Handle leader brokers in concurrent fetchers
   1. Selecting the leader broker to send a request to. Currently we don't support the preferred replicas that the kafka API provides.
   2. Handle errors around leader election and broker selection - this is basically triggering a metadata refresh whenever we see some specific errors. franz-go already refreshes metadata when it notices changes in brokers, but it's hard to tell whether this will work in all cases for us. Refreshing metadata has few (no) downsides.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -339,8 +339,6 @@ func (p *shardingPusher) PushToStorage(ctx context.Context, request *mimirpb.Wri
 		s.Context = ctx // retain the last context in case we have to flush it when closing shardingPusher
 
 		if len(s.Timeseries) < p.batchSize {
-			// We modified a value so far. Write it back.
-			p.unfilledShards[shard] = s
 			continue
 		}
 		p.unfilledShards[shard] = flushableWriteRequest{
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -339,6 +339,8 @@ func (p *shardingPusher) PushToStorage(ctx context.Context, request *mimirpb.Wri
 		s.Context = ctx // retain the last context in case we have to flush it when closing shardingPusher
 
 		if len(s.Timeseries) < p.batchSize {
+			// We modified a value so far. Write it back.
+			p.unfilledShards[shard] = s
 			continue
 		}
 		p.unfilledShards[shard] = flushableWriteRequest{
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -193,7 +193,7 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 	}
 
 	if r.kafkaCfg.ReplayConcurrency > 1 {
-		r.fetcher, err = newConcurrentFetchers(ctx, r.client, r.logger, r.kafkaCfg.Topic, r.partitionID, startOffset, r.kafkaCfg.ReplayConcurrency, r.kafkaCfg.RecordsPerFetch, &r.metrics)
+		r.fetcher, err = newConcurrentFetchers(ctx, r.client, r.logger, r.kafkaCfg.Topic, r.partitionID, startOffset, r.kafkaCfg.ReplayConcurrency, r.kafkaCfg.RecordsPerFetch, r.kafkaCfg.MinBytesMaxWaitDuration, &r.metrics)
 		if err != nil {
 			return errors.Wrap(err, "creating concurrent fetchers")
 		}
@@ -821,15 +821,16 @@ type concurrentFetchers struct {
 	metrics     *readerMetrics
 	tracer      *kotel.Tracer
 
-	concurrency     int
-	recordsPerFetch int
+	concurrency      int
+	recordsPerFetch  int
+	minBytesWaitTime time.Duration
 
 	orderedFetches     chan kgo.FetchPartition
 	lastReturnedRecord int64
 }
 
 // newConcurrentFetchers creates a new concurrentFetchers. startOffset can be kafkaOffsetStart, kafkaOffsetEnd or a specific offset.
-func newConcurrentFetchers(ctx context.Context, client *kgo.Client, logger log.Logger, topic string, partition int32, startOffset int64, concurrency int, recordsPerFetch int, metrics *readerMetrics) (*concurrentFetchers, error) {
+func newConcurrentFetchers(ctx context.Context, client *kgo.Client, logger log.Logger, topic string, partition int32, startOffset int64, concurrency int, recordsPerFetch int, minBytesWaitTime time.Duration, metrics *readerMetrics) (*concurrentFetchers, error) {
 	const noReturnedRecords = -1 // we still haven't returned the 0 offset.
 	f := &concurrentFetchers{
 		client:             client,
@@ -839,6 +840,7 @@ func newConcurrentFetchers(ctx context.Context, client *kgo.Client, logger log.L
 		partitionID:        partition,
 		metrics:            metrics,
 		recordsPerFetch:    recordsPerFetch,
+		minBytesWaitTime:   minBytesWaitTime,
 		lastReturnedRecord: noReturnedRecords,
 		tracer:             kotel.NewTracer(kotel.TracerPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))),
 		orderedFetches:     make(chan kgo.FetchPartition),
@@ -902,10 +904,6 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
 	}
 }
 
-// fetchMinBytesWaitTime is the time the Kafka broker can wait for MinBytes to be filled.
-// This is usually used when there aren't enough records available to fulfil MinBytes, so the broker waits for more records to be produced.
-var fetchMinBytesWaitTime = 10 * time.Second
-
 // fetchSingle sends a fetch request to the leader Kafka broker for a partition for the fetchWant and parses the responses.
 // fetchSingle returns a fetchResult which may or may not fulfil the entire fetchWant.
 // If ctx is cancelled, fetchSingle will return an empty fetchResult without an error.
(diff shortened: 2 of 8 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -339,6 +339,8 @@ func (p *shardingPusher) PushToStorage(ctx context.Context, request *mimirpb.Wri
 		s.Context = ctx // retain the last context in case we have to flush it when closing shardingPusher
 
 		if len(s.Timeseries) < p.batchSize {
+			// We modified a value so far. Write it back.
+			p.unfilledShards[shard] = s
```
dimitarvdimitrov, line 343: this is unrelated. I found the bug 2 weeks ago for some reason, so now I'm commiting it.

gotjosh: I don' think we need this since the introduction of https://github.com/grafana/mimir/pull/9133 - I fixed it there.

pkg/storage/ingest/reader.go
```diff
@@ -807,6 +807,10 @@ type fetchResult struct {
 	fetchedBytes int
 }
 
+func emptyFetchResult(err error) fetchResult {
```
gotjosh, line 810: nit: I feel like this prefix suits this better.
```suggestion
func newEmptyFetchResult(err error) fetchResult {
```

pkg/storage/ingest/reader.go
```diff
@@ -853,6 +857,8 @@ func newConcurrentFetchers(ctx context.Context, client *kgo.Client, logger log.L
 	topics, err := kadm.NewClient(client).ListTopics(ctx, topic)
 	if err != nil {
 		return nil, fmt.Errorf("failed to find topic ID: %w", err)
+	} else if !topics.Has(topic) {
```
gotjosh, line 860: nit: Why the nested else?

We could simply add another `if`

pkg/storage/ingest/reader.go
```diff
@@ -894,45 +900,57 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
 	}
 }
 
+var errUnknownPartitionLeader = fmt.Errorf("unknown partition leader")
```
gotjosh, line 903: nit: should we place this sat the top next to the other errors?

pkg/storage/ingest/reader.go
```diff
@@ -894,45 +900,57 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
...
+
+// fetchMinBytesWaitTime is the time the Kafka broker can wait for MinBytes to be filled.
+// This is usually used when there aren't enough records available to fulfil MinBytes, so the broker waits for more records to be produced.
+var fetchMinBytesWaitTime = 10 * time.Second
```
gotjosh, line 907: nit: I tend to discourage the use of global variables for methods that have a pointer receiver. You could simplify this and add this as an attribute to `concurrentFetchers` then rename this  to `defaultFetchMinBytesWaitTime` which is set on the constructor.

Also, non-nit can you please specify the unit in the name, it was non-obvious that even though this in seconds, the actual client requires millis.

gotjosh: Very similar to what you have done with `noReturnedRecords` which btw, I'd suggest you consider moving that to the top too.

pkg/storage/ingest/reader.go
```diff
@@ -894,45 +900,57 @@ func (r *concurrentFetchers) pollFetches(ctx context.Context) (result kgo.Fetche
...
+
+// fetchSingle sends a fetch request to the leader Kafka broker for a partition for the fetchWant and parses the responses.
```
gotjosh, line 909:
```suggestion
// fetchSingle attempts to find out the leader leader Kafka broker for a partition and then sends a fetch request to the leader of the fetchWant request and parses the responses
```

pkg/storage/ingest/reader.go
```diff
@@ -1122,20 +1144,50 @@ func handleKafkaFetchErr(fr fetchResult, fw fetchWant, shortBackoff, longBackoff
...
+	// Typically franz-go will update its own metadata when it detects a change in brokers. But it's hard to verify this.
+	// So we force a metadata refresh here to be sure.
+	// It's ok to call this from multiple fetchers concurrently. franz-go will only be sending one metadata request at a time (whether automatic, periodic, or forced).
+	//
+	// Metadata refresh is asynchronous. So even after forcing the refresh we might have outdated metadata.
+	// Hopefully the backoff that will follow is enough to get the latest metadata.
+	// If not, the fetcher will end up here again on the next attempt.
+	client.ForceMetadataRefresh()
+}
+
```
gotjosh, line 1190: nit: I don't think we need this level of interaction - if you care more about the comment than anything else, you can just inline the metadata refresh in a closure within `handleKafkaFetchErr`.

e.g.
```golang
func handleKafkaFetchErr(fr fetchResult, fw fetchWant, shortBackoff, longBackoff waiter, client *kgo.Client, logger log.Logger) fetchWant {
	refreshMetadata := client.ForceMetadataRefresh
```

and then add your comment on top, I do understand that this makes it simpler to test but for now, It feels like we should have a `ReaderClient` or something interface around here, which is kind of a different problem.

gotjosh (approved):
LGTM, almost all of my comments are nits.

pkg/storage/ingest/pusher.go
dimitarvdimitrov:
```suggestion
```

pkg/storage/ingest/reader.go
dimitarvdimitrov: good spot, moved

pkg/storage/ingest/reader.go
dimitarvdimitrov: > Also, non-nit can you please specify the unit in the name, it was non-obvious that even though this in seconds, the actual client requires millis.

it's a `time.Duration`, isn't the type enough? It's effectively in nanoseconds, but it's a proper type

dimitarvdimitrov: the reason i didn't want to do it in the config is because it requires a lot of prod code changes just to ease tests. Take a look at this - a404ec94168f8ee11858037ba42e6b8fb838e3fe - if you like it, we can keep it

pkg/storage/ingest/reader.go
dimitarvdimitrov: i do care about the comment and can inline this function, but what's wrong with the interface? we have the two `waiter`s too? for now we need only this if we later have a need for `ReaderClient` we can replace `metadataRefresher`, but we don't now (do we?)

pkg/storage/ingest/reader.go
gotjosh: I was thinking much simpler than the above:

```git
diff --git a/pkg/storage/ingest/reader.go b/pkg/storage/ingest/reader.go
index 854dd0aaf..3bfbc3261 100644
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -45,6 +45,8 @@ const (
 
 	// kafkaOffsetEnd is a special offset value that means the end of the partition.
 	kafkaOffsetEnd = int64(-1)
+
+	defaultMinBytesWaitTimeSeconds = 10 * time.Second
 )
 
 var (
@@ -840,7 +842,7 @@ func newConcurrentFetchers(ctx context.Context, client *kgo.Client, logger log.L
 		partitionID:        partition,
 		metrics:            metrics,
 		recordsPerFetch:    recordsPerFetch,
-		minBytesWaitTime:   minBytesWaitTime,
+		minBytesWaitTime:   defaultMinBytesWaitTimeSeconds,
 		lastReturnedRecord: noReturnedRecords,
 		tracer:             kotel.NewTracer(kotel.TracerPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}))),
 		orderedFetches:     make(chan kgo.FetchPartition),

```

pkg/storage/ingest/reader.go
gotjosh: I was more concerned with the indirection than interface to be honest. I'm generally not a fan of single method interfaces but can appreciate it's very subjective, this LGTM.

gotjosh (approved):
LGTM, I don't need to look at this again.

pkg/storage/ingest/reader.go
dimitarvdimitrov: oh, that works too. I prefer that. Done in a6b2e87c59aff7881a2a24ced5e2ee98407739c7

pkg/storage/ingest/reader.go
```diff
@@ -829,8 +829,12 @@ type concurrentFetchers struct {
 	lastReturnedRecord int64
 }
 
+// defaultMinBytesWaitTime is the time the Kafka broker can wait for MinBytes to be filled.
+// This is usually used when there aren't enough records available to fulfil MinBytes, so the broker waits for more records to be produced.
+var defaultMinBytesWaitTime = 10 * time.Second
```
gotjosh, line 834: I think we need to move this next to the top of the fiile.

pkg/storage/ingest/reader_test.go
```diff
@@ -30,6 +30,11 @@ import (
 	"github.com/grafana/mimir/pkg/util/testkafka"
 )
 
+func init() {
+	// Reduce the time the fake kafka would wait for new records. Sometimes this blocks startup.
+	defaultMinBytesWaitTime = time.Second
```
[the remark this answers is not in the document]
dimitarvdimitrov: i think this should do it https://github.com/grafana/mimir/pull/9202

pkg/storage/ingest/reader.go
dimitarvdimitrov: in #9202 it's now a constant and at the top

PR #9167, opened 2024-09-02
Update usage for alertmanager.utf8-strict-mode-enabled

grobinson-grafana:
#### What this PR does

This commit updates the usage for UTF-8 strict mode to link to the Grafana Mimir docs, and removes redundant text that is mentioned in those docs. This was not done earlier as the Mimir docs did not exist at the time this flag was created.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/alertmanager/multitenant.go
```diff
@@ -129,7 +129,7 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet, logger
...
+	f.BoolVar(&cfg.UTF8StrictMode, "alertmanager.utf8-strict-mode-enabled", false, "Enable UTF-8 strict mode. Allows UTF-8 characters in the matchers for routes and inhibition rules, in silences, and in the labels for alerts. It is recommended that all tenants run the `migrate-utf8` command in mimirtool before enabling this mode. Otherwise, some tenant configurations might fail to load. For more information, refer to https://grafana.com/docs/mimir/<MIMIR_VERSION>/references/architecture/components/alertmanager/#enable-utf-8. Enabling and then disabling UTF-8 strict mode can break existing Alertmanager configurations if tenants added UTF-8 characters to their Alertmanager configuration while it was enabled.")
```
pstibrany, line 132: Could we use real version to make link work correctly without any manual work? Or perhaps refer to docs in some other way?

grobinson-grafana: Hmm. We could use `latest`, but Mimir `2.14` will have changes that are not available in Mimir `2.13` or `2.12`. We could also make the link relative, for example, `/references/architecture/components/alertmanager/#enable-utf-8` but I don't know if that makes it more difficult.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -2482,16 +2482,8 @@ alertmanager_client:
...
-# parser for labels and matchers, and this input is incompatible`. To find
-# tenant configurations that are valid but contain ambiguous matchers, search
-# for log lines containing `Matchers input has disagreement`. Each log line
-# includes the invalid input, a suggestion on how to fix the input (excluding
-# ambiguous matchers, as these require manual correction), and the ID of the
-# affected tenant. You must run Mimir with debug-level logging enabled.
-# Otherwise, these lines aren't logged. For more information, refer to
-# https://prometheus.io/docs/alerting/latest/configuration/#label-matchers.
+# might fail to load. For more information, refer to
+# https://grafana.com/docs/mimir/<MIMIR_VERSION>/references/architecture/components/alertmanager/#enable-utf-8.
```
tacole02, line 2486: I think we should include link text here instead of just the URL.

tacole02 (approved):
One minor comment, otherwise LGTM!

docs/sources/mimir/configure/configuration-parameters/index.md
grobinson-grafana: Hi, do you mean like a Markdown URL? `[Enable UTF-8](https://grafana.com/docs/mimir/<MIMIR_VERSION>/references/architecture/components/alertmanager/#enable-utf-8)`?

tacole02: Exactly! You could even make it more specific like:
```
refer to [Enable UTF-8](https://grafana.com/docs/mimir/<MIMIR_VERSION>/references/architecture/components/alertmanager/#enable-utf-8) in the Grafana Mimir Alertmanager documentation.
```

56quarters (approved):

PR #8175, opened 2024-05-23
mixin dashbaords: add support for ingest storage replication

dimitarvdimitrov:
#### What this PR does

Some panels in the dashboards rely on `cortex_distributor_replication_factor`. This metric is not exported if running in ingest storage (https://github.com/grafana/mimir/pull/7809).

This PR updates all queries in the mixin that were relaying on `cortex_distributor_replication_factor`. If the replication factor metric isn't present now the queries fall back to the maximum of each partition for the metric in question (such as `cortex_ingester_active_series`).

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9170, opened 2024-09-02
 mixin: improve ingest storage queries

narqo:
#### What this PR does

This is the follow-up to #8175

Here I'm updating the dashboards' queries for the "classic storage" instances, to exclude the ingesters running under the "ingest storage" mode. This is to prevent overcounting during the migration from classic to ingest storage.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


colega (approved):

dimitarvdimitrov (approved):
can you add this to the list of PRs related to ingest storage in the changelog? IIRC there should be one entry with a long list of PRs

PR #9173, opened 2024-09-02
Add a feature flag to enable logging when parsing label matchers

grobinson-grafana:
#### What this PR does

To validate UTF-8 strict mode is working as intended, Grafana Mimir logs parsing label matchers to validate that the correct parser is used and also identify any label matchers that are incompatible or ambigious. However, this also produces a huge amount of log volume, and doesn't need to be enabled all the time. This puts it behind an experimental feature flag which will be removed when we also remove the feature flag for UTF-8 strict mode.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/mimir/modules.go
```diff
@@ -970,7 +970,12 @@ func (t *Mimir) initAlertManager() (serv services.Service, err error) {
 	}
 	features, err := featurecontrol.NewFlags(util_log.Logger, mode)
 	util_log.CheckFatal("initializing Alertmanager feature flags", err)
-	compat.InitFromFlags(util_log.Logger, features)
+
+	compatLogger := log.NewNopLogger()
+	if t.Cfg.Alertmanager.LogParsingLabelMatchers {
+		compatLogger = util_log.Logger
+	}
+	compat.InitFromFlags(compatLogger, features)
```
grobinson-grafana, line 978: This is where the logger is initialized for the label matchers parser code. You can find it [here](https://github.com/grafana/mimir/blob/mimir-2.13.0/vendor/github.com/prometheus/alertmanager/matchers/compat/parse.go#L59).

santihernandezc (approved):
LGTM! 👍

PR #9174, opened 2024-09-02
Add feature flag to enable logging for UTF-8 migration

grobinson-grafana:
#### What this PR does

This commit moves the logging for UTF-8 migration behind a feature flag, so it can be disabled once UTF-8 strict mode is enabled. The motivation for this is that this logging produces a huge amount of log volume, and doesn't need to be enabled all the time.

This pull request is different from #9173 as this feature flag uses a different code path that is specific to the UTF-8 migration, where as #9173 controls general logging of label matchers in tenant configurations.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/alertmanager/multitenant.go
```diff
@@ -806,12 +808,14 @@ type amConfig struct {
 // setConfig applies the given configuration to the alertmanager for `userID`,
 // creating an alertmanager if it doesn't already exist.
 func (am *MultitenantAlertmanager) setConfig(cfg amConfig) error {
-	// Instead of using "config" as the origin, as in Prometheus Alertmanager, we use "tenant".
-	// The reason for this that the config.Load function uses the origin "config",
-	// which is correct, but Mimir uses config.Load to validate both API requests and tenant
-	// configurations. This means metrics from API requests are confused with metrics from
-	// tenant configurations. To avoid this confusion, we use a different origin.
-	validateMatchersInConfigDesc(am.logger, "tenant", cfg.AlertConfigDesc)
+	if am.cfg.UTF8Migration {
```
grobinson-grafana, line 811: The reason I made this a separate feature flag instead of `!am.cfg.UTF8StrictMode` is because I believe some users might want to keep running this logging for a little while after enabling UTF-8 strict mode just to be 100% sure all their tenant configurations are OK. If I use `!am.cfg.UTF8StrictMode` then this use case is no longer possible.

santihernandezc (approved):
LGTM 👍 

Nit: the name of the feature toggle doesn't give a great cue about what it does, maybe we could make its function more explicit in the name itself (e.g. `utf8-migration-logging-enabled`). Just a suggestion, this should not block you from merging the PR.

grobinson-grafana:
> LGTM 👍
> 
> Nit: the name of the feature toggle doesn't give a great cue about what it does, maybe we could make its function more explicit in the name itself (e.g. `utf8-migration-logging-enabled`). Just a suggestion, this should not block you from merging the PR.

I changed the name to the suggestion! I think it's much better!

pstibrany (approved):
🙈

PR #9175, opened 2024-09-02
globalerror: improve wrapping errors with their context error counterparts

duricanikolic:
#### What this PR does

In #8865 we partially fixed a problem of erroneously wrapping gRPC errors containing `codes.Canceled` status code with the `context.Canceled` error, and their consecutive mapping to HTTP error with status code `499`. 

This PR is a generalization of #8865. The heuristic we use is the following:
- In case of gRPC errors with status code `codes.Canceled`, errors will be wrapped with the `context.Canceled` error, only if the underlying context actually contains an error. 
- In case of gRPC errors with status code `codes.DeadlineExceeded` will be wrapped with the `context.DeadlineExceeded` error only if the underlying context actually contains an error. 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/globalerror/grpc.go
```diff
@@ -13,27 +13,24 @@ import (
...
 // to a standard golang context error, and if it is, wraps the former with the latter.
 // If the given error isn't a gRPC error, or it doesn't correspond to a standard golang
 // context error, the original error is returned.
-func WrapGRPCErrorWithContextError(err error) error {
+func WrapGRPCErrorWithContextError(ctx context.Context, err error) error {
 	if err == nil {
 		return nil
 	}
+	if ctx.Err() == nil {
+		return err
```
colega, line 25: Why `err` and not `ctx.Err()`?

duricanikolic: Because at this point there is an error, `err`, it is not nil, but there is no error in the given context `ctx`, so we shouldn't wrap `err` with any context-related errors, but return `err` itself.

colega: Oh right, I misread the condition.

pkg/util/globalerror/grpc_test.go
```diff
@@ -29,70 +29,124 @@ import (
...
+				expectedGrpcCode:    codes.Canceled,
+				expectedContextErr:  context.Canceled,
+			},
+			"ErrorWithStatus with Canceled status and without canceled context": {
+				ctxCanceled:         false,
+				ctxDeadlineExceeded: false,
+				origErr:             WrapErrorWithGRPCStatus(errors.New("cancel error"), codes.Canceled, nil),
+				expectedGrpcCode:    codes.Canceled,
+				expectedContextErr:  nil,
 			},
```
colega, line 107: I'm afraid of the fact that if someone passes a wrong context to `WrapGRPCErrorWithContextError` it will never map the `context.Canceled` anymore, but I guess that's exactly what we want here.

PR #9189, opened 2024-09-03
Move childrenChecked check out of queuing algorithms

chencs:
#### What this PR does
Previously, queuing algorithms were responsible for managing how to define `checkedAllNodes`, a condition to signal the tree to move on to a different subtree (e.g., after checking every child of the current node for a dequeue-able object and finding none).

This change moves the responsibility for managing `checkedAllNodes` out of the queuing algorithm and instead codifying the condition for moving on to the next subtree -- it happens when we have attempted a dequeue on every child of the current node and found no dequeue-able value.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


Logiraptor (approved):

Logiraptor (commented):
Thank you!

francoposa (approved):

PR #9190, opened 2024-09-03
Use dequeue args and setup within tree instead of making pre-dequeue updates the responsibility of the tree caller

chencs:
#### What this PR does
Instead of relying on the tree caller to use things like `updateQueuingAlgorithmState` before a dequeue operation, define a `DequeueArgs` which is passed to the tree on dequeue. Each queuing algorithm has a `setup` function which is called before dequeue which takes what it needs from `DequeueArgs` to update its own state. If `DequeueArgs` is nil, the dequeue will still be attempted without updating any algorithm state.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/scheduler/queue/queue.go
```diff
@@ -676,7 +668,7 @@ func NewQuerierWorkerDequeueRequest(querierWorkerConn *QuerierWorkerConn, lastTe
 }
 
 // querierWorkerDequeueResponse is the response for a QuerierWorkerDequeueRequest,
-// to be written to the dequeue requests 's receiver channel.
+// to be written to the dequeue requests' receiver channel.
```
francoposa, line 671: keep as singular possessive

pkg/scheduler/queue/multi_queuing_algorithm_tree_queue.go
```diff
@@ -61,17 +67,23 @@ func (t *MultiQueuingAlgorithmTreeQueue) IsEmpty() bool {
 	return t.rootNode.IsEmpty()
 }
 
-// Dequeue removes and returns an item from the front of the next appropriate Node in the MultiQueuingAlgorithmTreeQueue, as
-// well as the path to the Node which that item was dequeued from.
+// Dequeue removes and returns an item from the front of the next appropriate Node in the
+// MultiQueuingAlgorithmTreeQueue, as well as the path to the Node which that item was dequeued from. If DequeueArgs
```
francoposa, line 71: nit: attempt to line break at sentence or clause endings as we do for our docs style guide

francoposa (approved):

PR #9198, opened 2024-09-04
development: enable metric suffixes by default on OTLP by default

krajorama:
I was confused why the docker compose profiles yielded different metrics:
```
 ./compose-up.sh --profile otel-collector-otlp-push
```
 had no unit suffix like milliseconds.

But
```
./compose-up.sh --profile otel-collector-remote-write
```
did. It was because we don't add the suffixes by default.


gouthamve (approved):
LGTM

aknuds1 (approved):

PR #9199, opened 2024-09-04
blockbuilder: consume and build blocks

narqo:
#### What this PR does

This PR continues the work in #9068 and #9147 and updates the `BlockBuilder` glueing together the kafka- and tsdb-parts. 

Note that this PR 
1. doesn't yet port all the tests we have in the `blockbuilder` branch. I'll add those as the follow-up. This one is already large enough to review.
2. doesn't expose `BlockBuilder` as a target in the Mimir's main. This will be the last part.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"time"
 
 	"github.com/go-kit/log"
@@ -20,6 +21,8 @@ import (
 
 	"github.com/grafana/mimir/pkg/storage/bucket"
 	"github.com/grafana/mimir/pkg/storage/ingest"
+	mimir_tsdb "github.com/grafana/mimir/pkg/storage/tsdb"
+	"github.com/grafana/mimir/pkg/storage/tsdb/block"
 	"github.com/grafana/mimir/pkg/util/validation"
 )
 
@@ -197,17 +200,15 @@ func (b *BlockBuilder) nextConsumeCycle(ctx context.Context, cycleEnd time.Time)
 			if err := lag.Err; err != nil {
 				level.Error(b.logger).Log("msg", "failed to get partition lag", "err", err, "partition", partition, "cycle_end", cycleEnd)
 			} else {
-				level.Info(b.logger).Log(
-					"msg", "nothing to consume in partition",
-					"partition", partition,
-					"offset", lag.Commit.At,
-					"end_offset", lag.End.Offset,
-					"lag", lag.Lag,
-				)
+				level.Info(b.logger).Log("msg", "nothing to consume in partition", "partition", partition, "offset", lag.Commit.At, "end_offset", lag.End.Offset, "lag", lag.Lag)
 			}
+			continue
 		}
 
-		// TODO(v): backport the code for consuming partition up to its lag.Lag
+		state := partitionStateFromLag(b.logger, lag, b.fallbackOffsetMillis)
+		if err := b.consumePartition(ctx, state, cycleEnd); err != nil {
+			level.Error(b.logger).Log("msg", "failed to consume partition", "err", err, "partition", partition)
+		}
 	}
 	return nil
 }
@@ -238,4 +239,309 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32) 
 	}
 
 	return kadm.GroupMemberLag{}, lastErr
+}
+
+type partitionState struct {
+	Partition          int32
+	Lag                int64
+	Commit             kadm.Offset
+	CommitRecTimestamp time.Time
+	SeenTillOffset     int64
+	LastBlockEnd       int64
+}
+
+func partitionStateFromLag(logger log.Logger, lag kadm.GroupMemberLag, fallbackMillis int64) *partitionState {
+	commitRecTs, seenTillOffset, lastBlockEnd, err := unmarshallCommitMeta(lag.Commit.Metadata)
+	if err != nil {
+		// If there is an error in unmarshalling the metadata, treat it as if
+		// we have no commit. There is no reason to stop the cycle for this.
+		level.Warn(logger).Log("msg", "error unmarshalling commit metadata", "err", err, "partition", lag.Partition, "offset", lag.Commit.At, "metadata", lag.Commit.Metadata)
+	}
+
+	if commitRecTs == 0 {
+		// If there was no commit metadata, we use the fallback to replay a set amount of
+		// records because it is non-trivial to peek at the first record in a partition to determine
+		// the range of replay required. Without knowing the range, we might end up trying to consume
+		// a lot of records in a single partition consumption call and end up in an OOM loop.
+		// TODO(codesome): add a test for this.
+		commitRecTs = fallbackMillis
+	}
+
+	return &partitionState{
+		Partition:          lag.Partition,
+		Lag:                lag.Lag,
+		Commit:             lag.Commit,
+		CommitRecTimestamp: time.UnixMilli(commitRecTs),
+		SeenTillOffset:     seenTillOffset,
+		LastBlockEnd:       lastBlockEnd,
+	}
+}
+
+func (b *BlockBuilder) consumePartition(ctx context.Context, state *partitionState, cycleEnd time.Time) error {
+	lagging := cycleEnd.Sub(state.CommitRecTimestamp) > 3*b.cfg.ConsumeInterval/2
+	if !lagging {
+		// Either we did not find a commit offset or we are not lagging behind by
+		// more than 1.5 times the consume interval.
+		// When there is no kafka commit, we play safe and assume seenTillOffset and
+		// lastBlockEnd was 0 to not discard any samples unnecessarily.
+		err := b.consumePartitionCycle(ctx, state, cycleEnd)
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", state.Commit.Partition, err)
+		}
+
+		return nil
+	}
+
+	// We are lagging behind. We need to consume the partition in sections.
+	// We iterate through all the ConsumeInterval intervals, starting from the first one after the last commit until the cycleEnd.
+	//cycleEndStartAt := commitRecTime.Truncate(b.cfg.ConsumeInterval).Add(b.cfg.ConsumeInterval + b.cfg.ConsumeIntervalBuffer)
+	cycleEndStartAt, _ := nextCycleEnd(
+		func() time.Time {
+			return state.CommitRecTimestamp
+		},
+		b.cfg.ConsumeInterval,
+		b.cfg.ConsumeIntervalBuffer,
+	)
+	level.Info(b.logger).Log("msg", "partition is lagging behind", "partition", state.Partition, "lag", state.Lag, "cycle_end_start", cycleEndStartAt, "cycle_end", cycleEnd)
+	for ce := cycleEndStartAt; cycleEnd.Sub(ce) >= 0; ce = ce.Add(b.cfg.ConsumeInterval) {
+		// Instead of looking for the commit metadata for each iteration, we use the data returned by consumePartition in the next iteration.
+		err := b.consumePartitionCycle(ctx, state, ce)
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", state.Partition, err)
+		}
+
+		// If adding the ConsumeInterval takes it beyond the cycleEnd, we set it to the cycleEnd to not
+		// exit the loop without consuming until cycleEnd.
+		if ce.Compare(cycleEnd) != 0 && ce.Add(b.cfg.ConsumeInterval).After(cycleEnd) {
+			ce = cycleEnd
+		}
+	}
+
+	return nil
+}
+
+// consumePartition consumes records from the given partition until the cycleEnd timestamp.
+// It modifies state, tracking the consumption progress. If the partition is lagging behind, the caller needs to take care
+// of calling consumePartition in sections.
+func (b *BlockBuilder) consumePartitionCycle(ctx context.Context, state *partitionState, cycleEnd time.Time) (retErr error) {
+	blockEndAt := cycleEnd.Truncate(b.cfg.ConsumeInterval)
+	blockEnd := blockEndAt.UnixMilli()
+
+	var (
+		numBlocks     int
+		compactionDur time.Duration
+	)
+	defer func(t time.Time, startState partitionState) {
+		dur := time.Since(t)
+
+		// Don't propagate any errors derived from the canceled context.
+		if errors.Is(retErr, context.Canceled) {
+			retErr = nil
+		}
+
+		if retErr != nil {
+			level.Error(b.logger).Log("msg", "partition consumption failed", "partition", state.Partition, "dur", dur, "start_lag", startState.Lag, "end_lag", state.Lag, "err", retErr)
+			return
+		}
+
+		b.metrics.processPartitionDuration.WithLabelValues(fmt.Sprintf("%d", state.Partition)).Observe(dur.Seconds())
+		level.Info(b.logger).Log("msg", "done consuming", "partition", state.Partition, "dur", dur,
+			"start_lag", startState.Lag, "end_lag", state.Lag, "cycle_end", cycleEnd,
+			"last_block_end", time.UnixMilli(startState.LastBlockEnd), "curr_block_end", time.UnixMilli(state.LastBlockEnd),
+			"last_seen_till", startState.SeenTillOffset, "curr_seen_till", state.SeenTillOffset,
+			"num_blocks", numBlocks, "compact_and_upload_dur", compactionDur)
+	}(time.Now(), *state)
+
+	builder := NewTSDBBuilder(b.logger, b.cfg.DataDir, b.cfg.BlocksStorage, b.limits)
+	defer builder.Close()
+
+	level.Info(b.logger).Log("msg", "start consuming", "partition", state.Partition, "offset", state.Commit.At, "lag", state.Lag, "cycle_end", cycleEnd)
+
+	// TopicPartition to resume consuming on this iteration.
+	tp := map[string][]int32{b.cfg.Kafka.Topic: {state.Partition}}
+	b.kafkaClient.ResumeFetchPartitions(tp)
+	defer b.kafkaClient.PauseFetchPartitions(tp)
+
+	// We always rewind the partition's offset to the commit offset. This is so the cycle started exactly at the commit point,
+	// and not at what was (potentially over-) consumed previously.
+	b.seekPartition(state.Partition, state.Commit.At)
+
+	var (
+		firstRec  *kgo.Record
+		lastRec   *kgo.Record
+		commitRec *kgo.Record
+	)
+
+consumerLoop:
+	for remaining := state.Lag; remaining > 0; {
+		if err := context.Cause(ctx); err != nil {
+			return err
+		}
+		fetches := b.kafkaClient.PollFetches(ctx)
+		if fetches.IsClientClosed() {
+			level.Warn(b.logger).Log("msg", "client closed when fetching records")
+			return nil
+		}
+		fetches.EachError(func(_ string, _ int32, err error) {
+			if !errors.Is(err, context.Canceled) {
+				level.Error(b.logger).Log("msg", "failed to fetch records", "partition", state.Partition, "err", err)
+				b.metrics.fetchErrors.Inc()
+			}
+		})
+
+		numRecs := fetches.NumRecords()
+		remaining -= int64(numRecs)
+		b.metrics.fetchRecordsTotal.Add(float64(numRecs))
+
+		recIter := fetches.RecordIter()
+		for !recIter.Done() {
+			rec := recIter.Next()
+
+			if firstRec == nil {
+				firstRec = rec
+			}
+
+			level.Debug(b.logger).Log("msg", "consumed record", "offset", rec.Offset, "rec_ts", rec.Timestamp, "last_bmax", state.LastBlockEnd, "bmax", blockEnd)
+
+			// Stop consuming after we reached the cycleEnd marker.
+			// NOTE: the timestamp of the record is when the record was produced relative to distributor's time.
+			if rec.Timestamp.After(cycleEnd) {
+				break consumerLoop
+			}
+
+			recProcessedBefore := rec.Offset <= state.SeenTillOffset
+			allSamplesProcessed, err := builder.Process(ctx, rec, state.LastBlockEnd, blockEnd, recProcessedBefore)
+			if err != nil {
+				// All "non-terminal" errors are handled by the TSDBBuilder.
+				return fmt.Errorf("process record in partition %d at offset %d: %w", rec.Partition, rec.Offset, err)
+			}
+			if !allSamplesProcessed {
+				if lastRec == nil {
+					// The first record was not fully processed, meaning the record before this is the commit point.
+					// We have to re-commit this one to update the seenTillOffset; this is so the next cycle handled partially processed record properly.
+					commitRec = &kgo.Record{
+						Topic:       state.Commit.Topic,
+						Partition:   state.Commit.Partition,
+						Offset:      state.Commit.At - 1, // Previous commit's offset-1 means we expect the next cycle to start from the commit's offset.
+						Timestamp:   state.CommitRecTimestamp,
+						LeaderEpoch: state.Commit.LeaderEpoch,
+					}
+				} else if commitRec == nil {
+					// The commit offset should be the last record that was fully processed and not the first record that was not fully processed.
+					commitRec = lastRec
+				}
+			}
+			lastRec = rec
+		}
+	}
+
+	compactStart := time.Now()
+	var err error
+	numBlocks, err = builder.CompactAndUpload(ctx, b.uploadBlocks)
+	if err != nil {
+		return err
+	}
+	compactionDur = time.Since(compactStart)
+	b.metrics.compactAndUploadDuration.WithLabelValues(fmt.Sprintf("%d", state.Partition)).Observe(compactionDur.Seconds())
+
+	// Nothing was consumed from Kafka at all.
+	if firstRec == nil {
+		level.Info(b.logger).Log("msg", "no records were consumed", "partition", state.Partition)
+		return nil
+	}
+
+	// No records were processed for this cycle.
+	if lastRec == nil {
+		level.Info(b.logger).Log("msg", "nothing to commit due to first record fetched is from next cycle", "partition", state.Partition, "first_rec_offset", firstRec.Offset)
+		return nil
+	}
+
+	// All samples in all records were processed. We can commit the last record's offset.
+	if commitRec == nil {
+		commitRec = lastRec
+	}
+
+	commitRecOffset := commitRec.Offset + 1 // offset+1 means everything up to (including) the offset was processed
+
+	// We should take the max of "seen till" offset. If the partition was lagging
+	// due to some record not being processed because of a future sample, we might be
+	// coming back to the same consume cycle again.
+	seenTillOffset := max(lastRec.Offset, state.SeenTillOffset)
+	// Take the max of block max times because of the same reasons above.
+	lastBlockEnd := max(blockEnd, state.LastBlockEnd)
+
+	commit := kadm.Offset{
+		Topic:       commitRec.Topic,
+		Partition:   commitRec.Partition,
+		At:          commitRecOffset,
+		LeaderEpoch: commitRec.LeaderEpoch,
+		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), seenTillOffset, lastBlockEnd),
+	}
+	if err := b.commitOffset(ctx, b.cfg.Kafka.ConsumerGroup, commit); err != nil {
+		return err
+	}
+
+	// The new lag is the distance between fully processed offsets. If the partition's cycle was chopped into sections due to lagging,
+	// the cycle will continue consuming the next portion of the lag.
+	state.Lag -= commitRecOffset - firstRec.Offset
+
+	state.SeenTillOffset = seenTillOffset
+	state.LastBlockEnd = lastBlockEnd
+	state.Commit = commit
+
+	return nil
+}
+
+func (b *BlockBuilder) seekPartition(partition int32, offset int64) {
+	offsets := map[string]map[int32]kgo.EpochOffset{
+		b.cfg.Kafka.Topic: {
+			partition: {
+				Epoch:  -1, // Epoch doesn't play any role outside the group consumption.
+				Offset: offset,
+			},
+		},
+	}
+	b.kafkaClient.SetOffsets(offsets)
+}
+
+func (b *BlockBuilder) commitOffset(ctx context.Context, group string, offset kadm.Offset) error {
+	offsets := make(kadm.Offsets)
+	offsets.Add(offset)
+	if err := kadm.NewClient(b.kafkaClient).CommitAllOffsets(ctx, group, offsets); err != nil {
+		return fmt.Errorf("commit with partition %d, offset %d: %w", offset.Partition, offset.At, err)
+	}
+
+	level.Debug(b.logger).Log("msg", "successfully committed to Kafka", "partition", offset.Partition, "offset", offset.At)
+
+	return nil
+}
+
+func (b *BlockBuilder) uploadBlocks(ctx context.Context, tenantID, dbDir string, blockIDs []string) error {
+	buc := bucket.NewUserBucketClient(tenantID, b.bucket, b.limits)
+	for _, bid := range blockIDs {
+		blockDir := path.Join(dbDir, bid)
+		meta, err := block.ReadMetaFromDir(blockDir)
+		if err != nil {
+			return err
+		}
+
+		if meta.Stats.NumSamples == 0 {
+			// No need to upload empty block.
+			return nil
+		}
+
+		meta.Thanos.Source = block.BlockBuilderSource
+		meta.Thanos.SegmentFiles = block.GetSegmentFiles(blockDir)
+
+		if meta.Compaction.FromOutOfOrder() && b.limits.OutOfOrderBlocksExternalLabelEnabled(tenantID) {
+			// At this point the OOO data was already ingested and compacted, so there's no point in checking for the OOO feature flag
+			meta.Thanos.Labels[mimir_tsdb.OutOfOrderExternalLabel] = mimir_tsdb.OutOfOrderExternalLabelValue
+		}
+
+		// Upload block with custom metadata.
+		if err := block.Upload(ctx, b.logger, buc, blockDir, meta); err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/pkg/blockbuilder/blockbuilder_metrics.go
+++ b/pkg/blockbuilder/blockbuilder_metrics.go
@@ -8,8 +8,12 @@ import (
 )
 
 type blockBuilderMetrics struct {
-	consumeCycleDuration prometheus.Histogram
-	consumerLagRecords   *prometheus.GaugeVec
+	consumeCycleDuration     prometheus.Histogram
+	processPartitionDuration *prometheus.HistogramVec
+	compactAndUploadDuration *prometheus.HistogramVec
+	fetchRecordsTotal        prometheus.Counter
+	fetchErrors              prometheus.Counter
+	consumerLagRecords       *prometheus.GaugeVec
 }
 
 func newBlockBuilderMetrics(reg prometheus.Registerer) blockBuilderMetrics {
@@ -20,6 +24,27 @@ func newBlockBuilderMetrics(reg prometheus.Registerer) blockBuilderMetrics {
 		Help: "Time spent consuming a full cycle.",
 
 		NativeHistogramBucketFactor: 1.1,
+	})
+
+	m.processPartitionDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Name:                        "cortex_blockbuilder_process_partition_duration_seconds",
+		Help:                        "Time spent processing one partition.",
+		NativeHistogramBucketFactor: 1.1,
+	}, []string{"partition"})
+
+	m.compactAndUploadDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Name:                        "cortex_blockbuilder_compact_and_upload_duration_seconds",
+		Help:                        "Time spent compacting and uploading one partition.",
+		NativeHistogramBucketFactor: 1.1,
+	}, []string{"partition"})
+
+	m.fetchRecordsTotal = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "cortex_blockbuilder_fetch_records_total",
+		Help: "Total number of records received by the consumer.",
+	})
+	m.fetchErrors = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "cortex_blockbuilder_fetch_errors_total",
+		Help: "Total number of errors while fetching by the consumer.",
 	})
 
 	m.consumerLagRecords = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
--- a/pkg/blockbuilder/blockbuilder_test.go
+++ b/pkg/blockbuilder/blockbuilder_test.go
@@ -7,15 +7,23 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/go-kit/log"
 	"github.com/grafana/dskit/services"
 	"github.com/prometheus/client_golang/prometheus"
 	promtest "github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/tsdb"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/twmb/franz-go/pkg/kadm"
+	"github.com/twmb/franz-go/pkg/kgo"
 
+	"github.com/grafana/mimir/pkg/mimirpb"
 	"github.com/grafana/mimir/pkg/util/test"
 	"github.com/grafana/mimir/pkg/util/testkafka"
 )
@@ -25,8 +33,6 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
 
 	ctx, cancel := context.WithCancelCause(context.Background())
 	t.Cleanup(func() { cancel(errors.New("test done")) })
-
-	const numPartitions = 2
 
 	_, kafkaAddr := testkafka.CreateClusterWithoutCustomConsumerGroupsSupport(t, numPartitions, testTopic)
 	cfg, overrides := blockBuilderConfig(t, kafkaAddr)
@@ -49,13 +55,11 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
 	require.Empty(t, list, "expected data_dir to be empty")
 }
 
-func TestBlockBuilder_NextConsumeCycle(t *testing.T) {
+func TestBlockBuilder_consumerLagRecords(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancelCause(context.Background())
 	t.Cleanup(func() { cancel(errors.New("test done")) })
-
-	const numPartitions = 2
 
 	_, kafkaAddr := testkafka.CreateClusterWithoutCustomConsumerGroupsSupport(t, numPartitions, testTopic)
 	kafkaClient := mustKafkaClient(t, kafkaAddr)
@@ -63,9 +67,6 @@ func TestBlockBuilder_NextConsumeCycle(t *testing.T) {
 	produceRecords(ctx, t, kafkaClient, time.Now().Add(-time.Hour), "1", testTopic, 0, []byte(`test value`))
 
 	cfg, overrides := blockBuilderConfig(t, kafkaAddr)
-	cfg.PartitionAssignment = map[string][]int32{
-		"block-builder-0": {0, 1}, // instance 0 -> partitions 0, 1
-	}
 
 	reg := prometheus.NewPedanticRegistry()
 	bb, err := New(cfg, test.NewTestingLogger(t), reg, overrides)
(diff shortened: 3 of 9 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		b.metrics.processPartitionDuration.WithLabelValues(fmt.Sprintf("%d", state.Partition)).Observe(dur.Seconds())
+		level.Info(b.logger).Log("msg", "done consuming", "partition", state.Partition, "dur", dur,
+			"start_lag", startState.Lag, "end_lag", state.Lag, "cycle_end", cycleEnd,
+			"last_block_end", time.UnixMilli(startState.LastBlockEnd), "curr_block_end", time.UnixMilli(state.LastBlockEnd),
+			"last_seen_till", startState.SeenTillOffset, "curr_seen_till", state.SeenTillOffset,
+			"num_blocks", numBlocks, "compact_and_upload_dur", compactionDur)
+	}(time.Now(), *state)
+
+	builder := NewTSDBBuilder(b.logger, b.cfg.DataDir, b.cfg.BlocksStorage, b.limits)
+	defer builder.Close()
```
dimitarvdimitrov, line 356: can you handle this error? there is `runutil.CloseWithErrCapture` which in `defer` uses

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+}
+
+func (b *BlockBuilder) consumePartition(ctx context.Context, state *partitionState, cycleEnd time.Time) error {
+	lagging := cycleEnd.Sub(state.CommitRecTimestamp) > 3*b.cfg.ConsumeInterval/2
+	if !lagging {
+		// Either we did not find a commit offset or we are not lagging behind by
+		// more than 1.5 times the consume interval.
+		// When there is no kafka commit, we play safe and assume seenTillOffset and
+		// lastBlockEnd was 0 to not discard any samples unnecessarily.
+		err := b.consumePartitionCycle(ctx, state, cycleEnd)
```
dimitarvdimitrov, line 287: if there's no commit don't we want to still chop up the period into multiple cycles? I mean do we want to compact the last 10 hours in 1 sections or in 10 sections?

Which also leads to another question: what's the difference between consuming only the last section and consuming multiple sections? i.e. why do we need this `if`, can't we just use the loop below?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", state.Commit.Partition, err)
+		}
+
+		return nil
+	}
+
+	// We are lagging behind. We need to consume the partition in sections.
```
dimitarvdimitrov, line 295: nitpick on the comment - you use "cycle" more often then "section", let's stick to one term - "cycle"?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	// We iterate through all the ConsumeInterval intervals, starting from the first one after the last commit until the cycleEnd.
+	//cycleEndStartAt := commitRecTime.Truncate(b.cfg.ConsumeInterval).Add(b.cfg.ConsumeInterval + b.cfg.ConsumeIntervalBuffer)
+	cycleEndStartAt, _ := nextCycleEnd(
```
dimitarvdimitrov, line 298: is this the end of the next cycle? doesn't `cycleEnd` fit better? I don't understand what it's starting. Maybe `firstCycleEnd`?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	for ce := cycleEndStartAt; cycleEnd.Sub(ce) >= 0; ce = ce.Add(b.cfg.ConsumeInterval) {
+		// Instead of looking for the commit metadata for each iteration, we use the data returned by consumePartition in the next iteration.
+		err := b.consumePartitionCycle(ctx, state, ce)
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", state.Partition, err)
+		}
+
+		// If adding the ConsumeInterval takes it beyond the cycleEnd, we set it to the cycleEnd to not
+		// exit the loop without consuming until cycleEnd.
+		if ce.Compare(cycleEnd) != 0 && ce.Add(b.cfg.ConsumeInterval).After(cycleEnd) {
```
dimitarvdimitrov, line 315: `ce.Compare(cycleEnd) != 0` -> `!ce.Equal(cycleEnd)`

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	)
+	level.Info(b.logger).Log("msg", "partition is lagging behind", "partition", state.Partition, "lag", state.Lag, "cycle_end_start", cycleEndStartAt, "cycle_end", cycleEnd)
+	for ce := cycleEndStartAt; cycleEnd.Sub(ce) >= 0; ce = ce.Add(b.cfg.ConsumeInterval) {
+		// Instead of looking for the commit metadata for each iteration, we use the data returned by consumePartition in the next iteration.
+		err := b.consumePartitionCycle(ctx, state, ce)
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", state.Partition, err)
+		}
+
+		// If adding the ConsumeInterval takes it beyond the cycleEnd, we set it to the cycleEnd to not
```
dimitarvdimitrov, line 313: i read this three times and I'm still confused. If we `ce = cycleEnd`, then wouldn't the loop action increment it with `b.cfg.ConsumeInterval` and the loop condition fail?

Another way of asking it as: aren't you advancing `ce` by setting it to `cycleEnd`? How would that keep us in the loop?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+
+		recIter := fetches.RecordIter()
+		for !recIter.Done() {
+			rec := recIter.Next()
+
+			if firstRec == nil {
+				firstRec = rec
+			}
+
+			level.Debug(b.logger).Log("msg", "consumed record", "offset", rec.Offset, "rec_ts", rec.Timestamp, "last_bmax", state.LastBlockEnd, "bmax", blockEnd)
```
dimitarvdimitrov, line 404: bmax isn't very informative if you haven't read the code. Can you rename it to `block_max_time` or something similar? Same with `last_bmax`

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
 
 	return kadm.GroupMemberLag{}, lastErr
 }
+
+type partitionState struct {
+	Partition          int32
+	Lag                int64
+	Commit             kadm.Offset
+	CommitRecTimestamp time.Time
+	SeenTillOffset     int64
```
dimitarvdimitrov, line 249: nits on naming: till is the short version of until, why not use Until? Also maybe something to make it more obvious if this offset was consumed or not - `latestConsumedOffset` or `oldestSeenOffset`, `firstUnconsummedOffset`?

dimitarvdimitrov (commented):
some initial comments. Most are minor. I still haven't gone through most of `consumePartitionCycle`

seizethedave (commented):
Rando nits. Out of time, I'll try to look again later!

pkg/blockbuilder/blockbuilder.go
narqo: Here a "section" means a part of a cycle (but we used to follow the kgo's vocabulary, and assumed a "part" means "partition"). That is, a "cycle" is an interval from `[T, cycleEnd)`. Sections of the cycle are chunks `[T, T+interval), [T+interval, T+2*interval), ... [T+S*interval, cycleEnd)` (assuming the duration between `cycleEnd` and `T` is N hours).

pkg/storage/tsdb/block/meta.go
```diff
@@ -29,6 +29,7 @@ const (
 	CompactorSource       SourceType = "compactor"
 	CompactorRepairSource SourceType = "compactor.repair"
 	BucketRepairSource    SourceType = "bucket.repair"
+	BlockBuilderSource    SourceType = "block-builder"
```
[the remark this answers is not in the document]
narqo: I read `compactor.repair` as `<service>.<action>`, that's so `block-builder`. Although `receive` doesn't make much sense to me. Is there any semantic for those sources?

narqo:
Looking through the first round of comments, I thought maybe it will help if I try to outline the general mechanics of processing in a pseudocode.

The `nextConsumeCycle` retrieves the lag and other details of a partition's state, and keep it in the `partitionState`. Then the `consumePartitionCycle` fetches records from the partitions and tracks its progress by updating this state:


```go
func nextConsumeCycle(cycleEnd) {
  ...
    state = &partitionState{
        Lag: lag,
        OldestSeenOffset: offset,
        ...
    }

    for s in cycle.Sections() {
        consumePartitionCycle(state, s.CycleEnd)
    }
}

func consumePartitionCycle(state, end) {
    // seek to the offset, because the previous section could overconsume
    kafka.Seek(state.CommitOffset)

loop:
    for lag := state.Lag; lag > 0; lag-- {
        recs := kafka.Consume()
        for rec := range recs {
            // we may've overconsumed: ie. lag is positive, but the rec is already after the end;
            // break out of fetching, leaving some Lag to be processed on the next cycle
            if rec.Time > end {
                break loop
            }

            // NOTE: not all samples in rec may be from this cycle
            // 1. some could be from the future cycle
            // 2. some could be from the past, i.e. we've already consumed this offset on the prev. cycle
            // that's why we keep the progress in the state
            tsdb.process(rec)
        }
    }

    tsdb.compact()
    uploadBlocks()

    // commit the first _fully consumed_ offset to kafka
    // update the state's Lag, CommitOffset, OldestSeenOffset, etc
}
```

pkg/blockbuilder/blockbuilder.go
narqo: The problem is that the `cycleEnd` is relative to the record's timestamp — not the timestamp of the commit (i.e. when distributor pushed the original record to kafka). If we haven't committed the metadata, that contains `CommitRecTimestamp`, we cannot tell the timestamp of the first record, w/o consuming it. Does it make sense?

narqo: I recall that, initially, we tried chopping the cycle from inside the consumption loop:

```
for recs := range kafka.Consume() {
  for rec := range recs {
    if rec.Ts > cycleEnd {
      return
    }
    if cycleEnd - rec.Ts > interval {
      tsdb.compact()
      tsdb.upload()
      kafka.commit()

      cycleEnd = rec.Ts + interval
    }
    ...
  }
}
```

But that kept showing edge cases, and we've refactored it the current form (with keeping the `rec.Ts` in the commit's metadata to make it simpler). I'm not a fan of this one, but it works. Maybe we can improve it now 🙃

pkg/blockbuilder/blockbuilder.go
narqo: TBF I cannot recall why it's like that either. I've refactored this part now, and, at least the current subset of tests, pass 🙃

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
 
 	return kadm.GroupMemberLag{}, lastErr
 }
+
+type partitionState struct {
```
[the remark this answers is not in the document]
narqo: Good call, I've added the comments for all fields.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		Partition:          lag.Partition,
+		Lag:                lag.Lag,
+		Commit:             lag.Commit,
+		CommitRecTimestamp: time.UnixMilli(commitRecTs),
+		SeenTillOffset:     seenTillOffset,
+		LastBlockEnd:       lastBlockEnd,
+	}
+}
+
+func (b *BlockBuilder) consumePartition(ctx context.Context, state *partitionState, cycleEnd time.Time) error {
```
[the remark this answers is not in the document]
narqo: Refer to the https://github.com/grafana/mimir/pull/9199#issuecomment-2333952030 for how `state` is used in `consumePartitionCycle`. Please, share if we can refactor it to something better.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +239,298 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+			return nil
+		}
+
+		meta.Thanos.Source = block.BlockBuilderSource
+		meta.Thanos.SegmentFiles = block.GetSegmentFiles(blockDir)
+
+		if meta.Compaction.FromOutOfOrder() && b.limits.OutOfOrderBlocksExternalLabelEnabled(tenantID) {
+			// At this point the OOO data was already ingested and compacted, so there's no point in checking for the OOO feature flag
+			meta.Thanos.Labels[mimir_tsdb.OutOfOrderExternalLabel] = mimir_tsdb.OutOfOrderExternalLabelValue
+		}
```
dimitarvdimitrov, line 528: Normally we don't expect empty blocks, right?. I'd log this at info level with the ulid of the block so we can trace why there were no samples in this block (maybe all the samples were invalid? unlikely, but possible).

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -49,25 +55,144 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
...
+	kafkaClient.AddConsumeTopics(testTopic)
+
+	cfg, overrides := blockBuilderConfig(t, kafkaAddr)
+
+	// Producing some records
+	cycleEndStartup := cycleEndAtStartup(time.Now(), cfg.ConsumeInterval, cfg.ConsumeIntervalBuffer)
+
+	var producedSamples []mimirpb.Sample
+	kafkaRecTime := cycleEndStartup.Truncate(cfg.ConsumeInterval).Add(-7 * time.Hour).Add(29 * time.Minute)
+	for kafkaRecTime.Before(cycleEndStartup) {
```
dimitarvdimitrov, line 109: this loop looks similar to the variable `produceSamples` in another test. Consider making that a helper function and reusing in the tests. Would make future tests a bit briefer and easier to write

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -49,25 +55,144 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
...
+	lastRec := commitRec
+	blockEnd := commitRec.Timestamp.Truncate(cfg.ConsumeInterval).Add(cfg.ConsumeInterval)
+	offset := kadm.Offset{
+		Topic:       commitRec.Topic,
+		Partition:   commitRec.Partition,
+		At:          commitRec.Offset + 1,
+		LeaderEpoch: -1, // not a group consumer
+		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), lastRec.Timestamp.UnixMilli(), blockEnd.UnixMilli()),
+	}
+	commitOffset(ctx, t, kafkaClient, testGroup, offset)
```
dimitarvdimitrov, line 142: did you consider just running the block builder multiple times? That's a more block-box approach and also tests that the same block builder can pick up from where it left off. Here we copy the logic for commiting and that can break later on if the BB start commiting in a different way.

i think it's mostly a matter of preference, but wanted to hear your thoughts on it

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -49,25 +55,144 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
...
+	bb, err := New(cfg, test.NewTestingLogger(t), prometheus.NewPedanticRegistry(), overrides)
+	require.NoError(t, err)
+
+	require.NoError(t, services.StartAndAwaitRunning(ctx, bb))
+	t.Cleanup(func() {
+		require.NoError(t, services.StopAndAwaitTerminated(ctx, bb))
+	})
+
+	// Wait for all cycles.
+	time.Sleep(time.Second * 3)
```
dimitarvdimitrov, line 153: cue flaky tests. Can you assert on a side effect like metrics or that there are N blocks in the bucket?

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -49,25 +55,144 @@ func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
...
+
+	// Because there is a commit, on startup, block-builder must consume samples only after the commit.
+	expSamples := producedSamples[1+(len(producedSamples)/2):]
+
+	bucketDir := path.Join(cfg.BlocksStorage.Bucket.Filesystem.Directory, "1")
+	db, err := tsdb.Open(bucketDir, log.NewNopLogger(), nil, nil, nil)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, db.Close()) })
+	compareQuery(t,
```
dimitarvdimitrov, line 162: nice that it has some correctness tests too

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -76,16 +201,178 @@ func TestBlockBuilder_NextConsumeCycle(t *testing.T) {
...
+			cortex_blockbuilder_consumer_lag_records{partition="0",topic="test"} 0
+			cortex_blockbuilder_consumer_lag_records{partition="1",topic="test"} 0
+			# HELP cortex_blockbuilder_fetch_records_total Total number of records received by the consumer.
+			# TYPE cortex_blockbuilder_fetch_records_total counter
+			cortex_blockbuilder_fetch_records_total 0
+		`), "cortex_blockbuilder_consumer_lag_records", "cortex_blockbuilder_fetch_records_total"))
+	}, 3*time.Second, 100*time.Millisecond, "expected skipping all records before lookback period")
+}
+
+func TestBlockBuilder_WithMultiplyTenants(t *testing.T) {
```
dimitarvdimitrov, line 220: multiply -> multiple

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -76,16 +201,178 @@ func TestBlockBuilder_NextConsumeCycle(t *testing.T) {
...
+		compareQuery(t,
+			db,
+			producedSamples[tenant],
+			nil,
+			labels.MustNewMatcher(labels.MatchRegexp, "foo", ".*"),
+		)
+	}
+}
+
+func TestBlockBuilder_WithNonMonotonicRecordTimestamps(t *testing.T) {
```
dimitarvdimitrov, line 277: nice test

pkg/blockbuilder/config_test.go
```diff
@@ -83,7 +85,7 @@ func blockBuilderConfig(t *testing.T, addr string) (Config, *validation.Override
 
 	cfg.InstanceID = "block-builder-0"
 	cfg.PartitionAssignment = map[string][]int32{
-		"block-builder-0": {0}, // instance 0 -> partition 0
+		"block-builder-0": {0, 1}, // instance 0 -> partition 0
```
dimitarvdimitrov, line 88: the comment is now outdated :smile: I think you can just remove it

dimitarvdimitrov (commented):
I'm done with the review. I'm leaving a lot of comments, but most of them are minor and aimed mostly at readability and structure. The code looks solid otherwise, I couldn't find any bugs.

There shouldn't be any major ones, but feel free to address these here or in follow-up PRs

pkg/storage/tsdb/block/meta.go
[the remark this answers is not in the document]
dimitarvdimitrov: oh, ok. Yeah, maybe. I think it's treated opaquely. `block-builder` is ok

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +239,298 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		// we have no commit. There is no reason to stop the cycle for this.
+		level.Warn(logger).Log("msg", "error unmarshalling commit metadata", "err", err, "partition", lag.Partition, "offset", lag.Commit.At, "metadata", lag.Commit.Metadata)
+	}
+
+	if commitRecTs == 0 {
+		// If there was no commit metadata, we use the fallback to replay a set amount of
+		// records because it is non-trivial to peek at the first record in a partition to determine
+		// the range of replay required. Without knowing the range, we might end up trying to consume
+		// a lot of records in a single partition consumption call and end up in an OOM loop.
+		// TODO(codesome): add a test for this.
```
dimitarvdimitrov, line 271: doesn't `TestBlockBuilder_StartWithLookbackOnNoCommit` now cover this?

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
dimitarvdimitrov: my preference is to return a new value for `state`. It makes it more obvious to the caller of `consumePartitionCycle` that you now have new state. So you'd have


```go
func consumePartitionCycle(ctx context.Context, state partitionState, cycleEnd time.Time) (newState partitionState, retErr error) {}
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +239,298 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		CommitRecTimestamp: time.UnixMilli(commitRecTs),
+		OldestSeenOffset:   oldestSeenOffset,
+		LastBlockEnd:       time.UnixMilli(lastBlockEndTs),
+	}
+}
+
+func (b *BlockBuilder) consumePartition(ctx context.Context, state *partitionState, cycleEnd time.Time) error {
+	sectionCycleEnd := cycleEnd
+	if sectionCycleEnd.Sub(state.CommitRecTimestamp) > 3*b.cfg.ConsumeInterval/2 {
+		// We are lagging behind by more thatn 1.5*interval or there is no commit. We need to consume the partition in sections.
```
dimitarvdimitrov, line 288:
```suggestion
// We are lagging behind by more than 1.5*interval or there is no commit. We need to consume the partition in sections.
```

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: i like how you did it now. I had basically something like that in mind.

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: thanks for explaining 💡

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: leaving suggestions from goland is a mess. i meant this 
```suggestion
		// We are lagging behind by more than 1.5*interval or there is no commit. We need to consume the partition in sections.
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	defer b.kafkaClient.PauseFetchPartitions(tp)
+
+	// We always rewind the partition's offset to the commit offset. This is so the cycle started exactly at the commit point,
+	// and not at what was (potentially over-) consumed previously.
+	b.seekPartition(state.Partition, state.Commit.At)
+
+	var (
+		firstRec  *kgo.Record
+		lastRec   *kgo.Record
+		commitRec *kgo.Record
```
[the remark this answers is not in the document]
narqo: In order to commit the state, we need two pieces:

1. offset of the record (`kgo.Record.Offset`)
2. timestamp of the record (`kgo.Record.Timestamp.UnixMilli()`)

I don't think there is a single struct in `kgo` that has both of those, other than the whole `Record`.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+			recProcessedBefore := rec.Offset <= state.SeenTillOffset
+			allSamplesProcessed, err := builder.Process(ctx, rec, state.LastBlockEnd, blockEnd, recProcessedBefore)
+			if err != nil {
+				// All "non-terminal" errors are handled by the TSDBBuilder.
+				return fmt.Errorf("process record in partition %d at offset %d: %w", rec.Partition, rec.Offset, err)
+			}
+			if !allSamplesProcessed {
+				if lastRec == nil {
+					// The first record was not fully processed, meaning the record before this is the commit point.
+					// We have to re-commit this one to update the seenTillOffset; this is so the next cycle handled partially processed record properly.
```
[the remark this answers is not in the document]
narqo: I've rephrased this comment. The point here is that we re-commit the same offset to update its metadata, with the latest value of `oldestSeenOffset`. Otherwise, committing what has been already committed on the previous cycle is redundant.

pkg/blockbuilder/blockbuilder_metrics.go
```diff
@@ -22,6 +26,27 @@ func newBlockBuilderMetrics(reg prometheus.Registerer) blockBuilderMetrics {
...
+		Help:                        "Time spent compacting and uploading one partition.",
+		NativeHistogramBucketFactor: 1.1,
+	}, []string{"partition"})
+
+	m.fetchRecordsTotal = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "cortex_blockbuilder_fetch_records_total",
+		Help: "Total number of records received by the consumer.",
+	})
+	m.fetchErrors = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "cortex_blockbuilder_fetch_errors_total",
```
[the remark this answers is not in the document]
narqo: Wouldn't it be confusing to run mimir in monolith mode, if both `ingester` and `blockbuilder` packages used the same metrics?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +240,308 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	lastBlockEnd := max(blockEnd, state.LastBlockEnd)
+
+	commit := kadm.Offset{
+		Topic:       commitRec.Topic,
+		Partition:   commitRec.Partition,
+		At:          commitRecOffset,
+		LeaderEpoch: commitRec.LeaderEpoch,
+		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), seenTillOffset, lastBlockEnd),
+	}
+	if err := b.commitOffset(ctx, b.cfg.Kafka.ConsumerGroup, commit); err != nil {
```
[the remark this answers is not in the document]
narqo: That's actually how it was initially. But in one of the reviews we agreed to not update the state before we make sure the commit succeeded (ref https://github.com/grafana/mimir/pull/8991#discussion_r1728113980) 😅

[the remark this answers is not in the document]
dimitarvdimitrov: you wouldn't have this problem if you weren't using pointers 👀 👀 but your call, this isn't a major comment

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
dimitarvdimitrov: ok, then what about the naming?

pkg/blockbuilder/blockbuilder_metrics.go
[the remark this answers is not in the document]
dimitarvdimitrov: idk, i don't think so. You'd have a single metric to look at for errors. Otherwise you always need to check two metrics.

It's also an option to unify the metrics when we end up refactoring the block builder to be able to consume faster with the replay speed work

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +241,289 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	state.Commit = commit
+
+	return nil
+}
+
+func (b *BlockBuilder) seekPartition(partition int32, offset int64) {
+	offsets := map[string]map[int32]kgo.EpochOffset{
+		b.cfg.Kafka.Topic: {
+			partition: {
+				Epoch:  -1, // Epoch doesn't play any role outside the group consumption.
```
dimitarvdimitrov, line 481: doesn't epoch help detect when leaders get reelected and might help detect that the client is out of date?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +241,289 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+				Offset: offset,
+			},
+		},
+	}
+	b.kafkaClient.SetOffsets(offsets)
```
dimitarvdimitrov, line 486: the godoc on SetOffsets says that you must've used the same client to consume from that partition to be able to adequately set the offset. From what I could tell you're not guaranteed this, right? 

I believe that's the reason we used two clients in the ingest storage ingester

narqo: > the godoc on SetOffsets says that you must've used the same client to consume from that partition to be able to adequately set the offset.

I think you refer to [this part][1] 👇 right?

> it is invalid to set topics that were not yet returned from a PollFetches: this function sets only partitions that were previously consumed, any extra partitions are skipped.

I don't think this caveat is meaningful to us. Here we only need to move the client's internal cursor to the offset of the last commit, in case the client overconsumed on the previous cycle-section.

On the very first call to `consumePartitionCycle` after the rest, this is a noop, because the client didn't consume anything, so it will consume from the commit (or, if there is no commit, it will reset to the fallback, due to `kgo.ConsumeResetOffset` option).

Effectively, what we are doing here is as if we re-created a brand-new Kafka client, every time we enter `consumePartitionCycle`.

[1]: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Client.SetOffsets

dimitarvdimitrov: > It will start consuming from last commit

what guarantees this? Since we're consuming out of a consumer group (i.e. not relying on the kafka client to manage the commits), we manage the start offsets. Then doesn't that mean that we'll always offset to `fallbackOffsetMillis` the first time we start consuming a partition?

narqo: > doesn't that mean that we'll always offset to `fallbackOffsetMillis` the first time we start consuming a partition?

That's a good point to check. In our unit-tests, I saw that it behaved as expected (i.e. the [`TestBlockBuilder_StartWithExistingCommit`](https://github.com/grafana/mimir/blob/cd35789928f6d4763526cea84ec235046197be43/pkg/blockbuilder/blockbuilder_test.go#L95) test consumes from the last commit). But since the past refactoring in the backporting PRs, I also haven't tested it against the real Kafka with existing data. I should do that.

pkg/blockbuilder/blockbuilder.go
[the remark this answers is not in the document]
narqo: >  I feel like some name such as `firstPartiallyProcessedRecord` would be more obvious.

I've spent too much time debating about that with myself already. The `commitRec` points at the record whose offset we want to commit.

- if we fully processed all records in this cycle, this will be the last consumed record
- otherwise, it will be the **last** fully processed record before the first partially processed one 🤯
- or, if the very first record was partially processed, it points to the previous commit (updating the commit's metadata).

I'm fine renaming it, but this name doesn't make sense to me.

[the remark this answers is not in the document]
dimitarvdimitrov: let's keep it as-is

pkg/blockbuilder/blockbuilder.go
narqo: I've reworked this part, see https://github.com/grafana/mimir/pull/9199/commits/d3786fdf03894049e43505fe4ddafd3d37ac2630.

This was indeed a bug that we introduced during the refactorings between #9068 and #9147. `kfake` didn't show it, because it behaves differently in when partitions and offsets are assigned (I haven't delved into the roots of it). The new version works in real Kafka, though.

pkg/blockbuilder/blockbuilder_metrics.go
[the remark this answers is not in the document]
narqo: Yeah, let's come back to the metrics naming separately.

pkg/blockbuilder/blockbuilder.go
narqo: I cannot figure out which piece of code you refer to here. Should this log be done during the compaction in tsdb?

narqo:
I think, I've addressed everything here. Should be good for another review round

pkg/blockbuilder/blockbuilder.go
```diff
@@ -437,63 +429,46 @@ consumerLoop:
...
-		At:          commitRecOffset,
+		At:          commitRec.Offset + 1, // offset+1 means everything up to (including) the offset was processed
 		LeaderEpoch: commitRec.LeaderEpoch,
-		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), oldestSeenOffset, lastBlockEnd.UnixMilli()),
+		Metadata:    marshallCommitMeta(commitRec.Timestamp.UnixMilli(), lastSeenOffset, lastBlockEnd.UnixMilli()),
 	}
-	if err := b.commitOffset(ctx, partitionLogger, b.cfg.Kafka.ConsumerGroup, commit); err != nil {
-		return err
+	newState := partitionState{
+		Lag:                state.Lag - (commit.At - firstRec.Offset), // the new lag is the distance between fully processed offsets
```
dimitarvdimitrov, line 451: i just realized that offsets aren't guaranteed to be continuous if we use transactions. There may be aborted offsets for which there are no records. Then `Lag` could become negative after doing `state.Lag - (commit.At - firstRec.Offset)`. Isn't this a problem?

I'm not sure what's the best way to deal with this. 1) leave a comment? 2) store the end offset of the partition instead of the Lag and have lag be computed like this.
```go
func (s partitionState) Lag() int64 {
  return s.EndOffset - s.Commit
}
```

In this case Lag shouldn't go below 0.

dimitarvdimitrov (commented):
now I'm not sure how we reset offsets when we start out of range 🤔

pracucci (commented):
I'm half way through my review. Submitting it, I will continue it asap.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+			}
+			if !allSamplesProcessed {
+				if lastRec == nil {
+					// The first record was not fully processed, meaning the record before this is the commit point.
+					// We hand-craft the commitRec from the data in the state to re-commit it. On commit the commit's meta is updated
+					// with the new value of LastSeenOffset. This is so the next cycle handled partially processed record properly.
+					commitRec = &kgo.Record{
+						Topic:       state.Commit.Topic,
+						Partition:   state.Commit.Partition,
+						Offset:      state.Commit.At - 1, // Previous commit's offset-1 means we expect the next cycle to start from the commit's offset.
```
pracucci, line 399: Could you explain me why - 1? The comment on this line didn't help me.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+
+	// Nothing was consumed from Kafka at all.
+	if firstRec == nil {
+		level.Info(logger).Log("msg", "no records were consumed")
+		return state, nil
+	}
+
+	// No records were processed for this cycle.
+	if lastRec == nil {
+		level.Info(logger).Log("msg", "nothing to commit due to first record fetched is from next cycle", "first_rec_offset", firstRec.Offset)
```
pracucci, line 420: [nit] I would clarify it. I would also log the firstRec timestamp and cycle end.

```suggestion
		level.Info(logger).Log("msg", "nothing to commit due to first record fetched has a timestamp greater than this cycle end", "first_rec_offset", firstRec.Offset)
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	if err := kadm.NewClient(b.kafkaClient).CommitAllOffsets(ctx, group, offsets); err != nil {
+		return fmt.Errorf("commit with partition %d, offset %d: %w", state.Commit.Partition, state.Commit.At, err)
+	}
+
+	level.Debug(logger).Log("msg", "successfully committed to Kafka", "offset", state.Commit.At)
+
+	return nil
+}
+
+func (b *BlockBuilder) uploadBlocks(ctx context.Context, tenantID, dbDir string, blockIDs []string) error {
```
pracucci, line 476: I would wrap it into another function `uploadBlocksWithRetries()`. I don't want to re-run the whole cycle just because of a transient networking issue.

pkg/blockbuilder/blockbuilder.go
pracucci: Do we use transactions?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	level.Info(logger).Log("msg", "start consuming", "offset", state.Commit.At)
+
+	var (
+		firstRec  *kgo.Record
+		lastRec   *kgo.Record
+		commitRec *kgo.Record
+	)
+
+consumerLoop:
+	for remaining := state.Lag; remaining > 0; {
```
pracucci, line 356: Generally speaking, the `remaining > 0` is always a red flag to me, because it's hard to reason about possible edge cases where `remaining > 0` but we actually consumed the partition till the end (i.e. if `state.Lag` was not computed correctly). I'm thinking in particular to edge cases related to inactive partitions, where no new records are produced, and so this `for` loop may potentially never end.

To prevent for such starvation in case of issues we're not considering right now, what if we pass to `PollFetches()` a context with a long timeout (e.g. few minutes) and we break the consumption (like if `remaining` has reached 0) if such timeout expires?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+
+	return newState, nil
+}
+
+func (b *BlockBuilder) commitState(ctx context.Context, logger log.Logger, group string, state partitionState) error {
+	offsets := make(kadm.Offsets)
+	offsets.Add(state.Commit)
+	if err := kadm.NewClient(b.kafkaClient).CommitAllOffsets(ctx, group, offsets); err != nil {
+		return fmt.Errorf("commit with partition %d, offset %d: %w", state.Commit.Partition, state.Commit.At, err)
+	}
```
pracucci, line 469: We should have retries here. I don't want to discard the whole compaction cycle just because the commit failed due to a transient error. You could wrap this function into a `commitStateWithRetries()` (it's an approach we already use in multiple places in Mimir).

pracucci (commented):
Nice work! I've found the code comments very helpful. Keep commenting the logic explaining the "why": it's very useful! I also find the code quite easy to follow, good job!

I left some nits, I hope you don't mind taking a look. Tomorrow, with fresh eyes, I want to take again a look at the commit state (I also left a question to better understand it). Keep pushing! 🚀

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: we don't

pracucci: Then I think a simple mention in a comment is enough. We don't need to handle an edge for a feature we don't use.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		b.blockBuilderMetrics.processPartitionDuration.WithLabelValues(fmt.Sprintf("%d", partition)).Observe(dur.Seconds())
+		level.Info(logger).Log("msg", "done consuming", "duration", dur, "curr_lag", state.Lag,
+			"last_block_end", startState.LastBlockEnd, "curr_block_end", blockEnd,
+			"last_seen_offset", startState.LastSeenOffset, "curr_seen_offset", state.LastSeenOffset,
+			"num_blocks", numBlocks)
+	}(time.Now(), state)
+
+	// We always rewind the partition's offset to the commit offset by reassigning the partition to the client (this triggers partition assignment).
+	// This is so the cycle started exactly at the commit offset, and not at what was (potentially over-) consumed previously.
+	b.kafkaClient.AddConsumePartitions(map[string]map[int32]kgo.Offset{
```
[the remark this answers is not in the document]
narqo: _I actually had to pause and look this up, 'cause I'm already getting confused myself with how it works_ 😵‍💫 

When the `getGroupLag` resolves the commit it falls back to the `fallbackOffsetMillis` when a partition doesn't have a commit. I.e. the `offset` in `state.Commit.At` is always the expected one now.

pkg/blockbuilder/blockbuilder.go
narqo: > what if we pass to `PollFetches()` a context with a long timeout (e.g. few minutes) and we break the consumption (like if remaining has reached 0) if such timeout expires?

This was exactly what I did in the earlier versions. I bumped into the problem that `PollFetches` can randomly return an empty set of fetches if the broker decides so. If we get rid of the `remaining` guard, we cannot tell if we should bail out from the polling loop (because we consumed the whole backlog), or we must retry the polling.

pkg/blockbuilder/blockbuilder.go
narqo: `commitRec` is the record we "commit" to Kafka (similar to [`kgo.Client.CommitRecords`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Client.CommitRecords)). But in Kafka, one doesn't commit a record — one commits the offset _after_ the record (i.e. the offset at which next consumption will start).

That was the backstory of this `offset-1`. We create a "fake" record that point at the offset **before** the last commit offset, so the code below committed the offset _after the before_ the last commit offset.

In pseudo-code:

```
lastCommitOffset = 0
recOffset  = lastCommitOffset - 1

commitOffset = recOffset + 1
assert(commitOffset == lastCommitOffset)
kafka.Commit(commitOffset)
```

pkg/blockbuilder/blockbuilder.go
narqo: > [..] store the end offset of the partition instead of the Lag and have lag be computed like this...

I don't really follow the problem. Lag, as returned by [`kadm.CalculateGroupLagWithStartOffsets`][1] is the difference between the partition's end offset and commit offset, no?

I don't remember the details about the transaction's implementation, but I assumed that the offsets are guaranteed to be monotonic. With that it shouldn't matter if there are gaps between the offset of the previous last commit and the new last commit offset. What do I miss?

[1]: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#CalculateGroupLagWithStartOffsets

pkg/blockbuilder/blockbuilder.go
```diff
@@ -83,22 +89,19 @@ func (b *BlockBuilder) starting(context.Context) (err error) {
...
 	if err != nil {
 		return fmt.Errorf("creating kafka reader: %w", err)
 	}
 
-	// Immediately pause fetching all assigned partitions. We control the order and the pace of partition fetching within the cycles.
-	b.kafkaClient.PauseFetchPartitions(map[string][]int32{b.cfg.Kafka.Topic: b.assignedPartitionIDs})
+	// Immediately unassign (remove) all partitions from the client. We control the order and the pace of fetching within the cycles.
+	b.kafkaClient.RemoveConsumePartitions(map[string][]int32{
+		b.cfg.Kafka.Topic: b.assignedPartitionIDs,
+	})
```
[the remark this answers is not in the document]
narqo: Good call. Probably, I wanted it to trigger a metadata update for some reason, but there were so many iterations here, so I don't remember already if it solved anything. I removed this chunk — everything works without it 👍

narqo:
I've addressed all the latest comments. PTAL

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +237,306 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+func (b *BlockBuilder) consumePartitionSection(ctx context.Context, logger log.Logger, builder *TSDBBuilder, partition int32, state partitionState, sectionEnd time.Time) (_ partitionState, retErr error) {
+	// Oppose to the section's range (and cycle's range), that include the ConsumeIntervalBuffer, the block's range doesn't.
+	// Thus, truncate the timestamp with ConsumptionInterval here to round the block's range.
+	blockEnd := sectionEnd.Truncate(b.cfg.ConsumeInterval)
+
+	var numBlocks int
+	defer func(t time.Time, startState partitionState) {
+		dur := time.Since(t)
+
+		// Don't propagate any errors derived from the cancelled context.
```
pracucci, line 332: I'm not convinced it's correct to not propagate the `context.Canceled` error, for two reasons:

1. Design: we're used to propagate it to the caller, so they can take an informed decision
2. The caller doesn't stop the loop with the current code

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +238,267 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+	}(time.Now(), state)
+
+	// We always rewind the partition's offset to the commit offset by reassigning the partition to the client (this triggers partition assignment).
+	// This is so the cycle started exactly at the commit offset, and not at what was (potentially over-) consumed previously.
+	b.kafkaClient.AddConsumePartitions(map[string]map[int32]kgo.Offset{
+		b.cfg.Kafka.Topic: {
+			partition: kgo.NewOffset().At(state.Commit.At),
+		},
+	})
+	defer b.kafkaClient.RemoveConsumePartitions(map[string][]int32{b.cfg.Kafka.Topic: {partition}})
```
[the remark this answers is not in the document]
pracucci: My take is that we don't have to think about removing partitions consumption (and don't have to deal with any potential bug or edge case in the Kafka client) if we just create a new ephemeral Kafka client inside `consumePartitionSection()`. I do understand you need another Kafka client to get the lag, and I'm perfectly fine using a shared one, but my comment is just about `consumePartitionSection()`: I simply find it safer to create an ephemeral client that removes the need to call `RemoveConsumePartitions()`.

pkg/blockbuilder/blockbuilder.go
pracucci: Ok great. You should condense this explanation in a code comment at the line where we do `state.Commit.At - 1` 😉

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +237,306 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		meta.Thanos.SegmentFiles = block.GetSegmentFiles(blockDir)
+
+		if meta.Compaction.FromOutOfOrder() && b.limits.OutOfOrderBlocksExternalLabelEnabled(tenantID) {
+			// At this point the OOO data was already ingested and compacted, so there's no point in checking for the OOO feature flag
+			meta.Thanos.Labels[mimir_tsdb.OutOfOrderExternalLabel] = mimir_tsdb.OutOfOrderExternalLabelValue
+		}
+
+		boff := backoff.New(ctx, backoff.Config{
+			MinBackoff: 100 * time.Millisecond,
+			MaxBackoff: time.Second,
```
pracucci, line 526: I would significantly increase the backoff here. If there's a temporary networking issue it's unlikely will be resolved in 10s (10 x 1s max backoff). We could have way longer backoffs instead, explaining the reason with a comment.

pkg/blockbuilder/blockbuilder.go
pracucci: > I bumped into the problem that PollFetches can randomly return an empty set of fetches if the broker decides so.

Interesting. Add a code comment to explain it, please.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +237,306 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
 
 	return kadm.GroupMemberLag{}, lastErr
 }
+
+type partitionState struct {
+	// Lag is the upperbound number of records the cycle will need to consume.
+	Lag int64
+	// Commit is the current offset commit for the partition.
```
pracucci, line 244: [nit] I suggest to clarify it.

```suggestion
	// Commit is the offset of the next record we'll start consuming.
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +237,306 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+		if err != nil {
+			return fmt.Errorf("consume partition %d: %w", partition, err)
+		}
+		sectionEnd = sectionEnd.Add(b.cfg.ConsumeInterval)
+	}
+
+	return nil
+}
+
+func (b *BlockBuilder) consumePartitionSection(ctx context.Context, logger log.Logger, builder *TSDBBuilder, partition int32, state partitionState, sectionEnd time.Time) (_ partitionState, retErr error) {
```
pracucci, line 323: I went with Vladimir through different cases to better understand how the offset work and LGTM. At a first glance I found it non trivial to understand, so I'm leaving my notes here.

Basic info:
- Next consumption cycle should process `partitionState.Commit` offset as 1st record.


Partition has 1 record, and the record has been partially processed:

- AddConsumePartitions() called with `state.Commit.At`
  - The first record will have offset `state.Commit.At`
- firstRec  = first record (not fully processed)
  - It has offset `state.Commit.At`
- commitRec = Offset set to "state.Commit.At - 1"
- lastRec = firstRec
- We commit `commit.At` set to `commitRec.Offset + 1`, which is `state.Commit.At - 1 + 1`, which is `state.Commit.At`, correct


Partition has 2 records, the 1st record is fully processed, but the 2nd record is partially processed:

- AddConsumePartitions() called with `state.Commit.At`
  - The first record will have offset `state.Commit.At`
- firstRec  = first record (fully processed)
  - It has offset `state.Commit.At`
- The 2nd record is partially processed
  - It has offset `state.Commit.At + 1`
- commitRec = firstRecord
  - It has offset `state.Commit.At`
- We commit `commit.At` set to `commitRec.Offset + 1`, which is `state.Commit.At + 1`, which is the 2nd record, which is the earlier record with samples not fully processed and so the next time we'll have to restart from there, correct


Partition has 2 records, we fully process all records:

- AddConsumePartitions() called with `state.Commit.At`
  - The first record will have offset `state.Commit.At`
- firstRec  = first record (fully processed)
  - It has offset `state.Commit.At`
- The 2nd record (fully processed)
  - It has offset `state.Commit.At + 1`
- commitRec = second record
  - It has offset `state.Commit.At + 1`
- We commit `commit.At` set to `commitRec.Offset + 1`, which is `state.Commit.At + 1 + 1`, which is the offset that will have the 3rd record, which is correct because records up until the 2nd one (included) have been fully processed, correct


Partition has 1 record, it's first record ever in the partition so its offset is 0 and the record is partially processed:

- AddConsumePartitions() called with `state.Commit.At = 0`
  - The first record will have offset `state.Commit.At = 0`
- firstRec  = first record (partially processed)
  - It has offset `state.Commit.At = 0`
- commitRec = offset set to `state.Commit.At - 1`, which is `-1`
- We commit `commit.At` set to `commitRec.Offset + 1`, which is `-1 +1`, which is `0`, correct

pracucci (approved):
Good job! I left few last minor comments, but no need for me to re-review it.

pkg/blockbuilder/blockbuilder.go
narqo: After brainstorming with @dimitarvdimitrov I've added a comment, that explains the theoretical edge-case and potential way to rework it later https://github.com/grafana/mimir/pull/9199/commits/ac269454239f58e801bf0b06da93d5b8d4bdffa6

pkg/blockbuilder/blockbuilder.go
```diff
@@ -239,3 +237,319 @@ func (b *BlockBuilder) getLagForPartition(ctx context.Context, partition int32)
...
+
+		dur := time.Since(t)
+
+		if retErr != nil {
+			level.Error(logger).Log("msg", "partition consumption failed", "err", retErr, "duration", dur, "curr_lag", state.Lag)
+			return
+		}
+
+		b.blockBuilderMetrics.processPartitionDuration.WithLabelValues(fmt.Sprintf("%d", partition)).Observe(dur.Seconds())
+		level.Info(logger).Log("msg", "done consuming", "duration", dur, "curr_lag", state.Lag,
```
dimitarvdimitrov, line 343: now we don't mutate `state`, so `curr_lag`, `curr_seen_offset` will the same as when this function started

pkg/blockbuilder/kafkautil.go
```diff
@@ -60,6 +60,11 @@ func getGroupLag(ctx context.Context, admClient *kadm.Client, topic, group strin
 			if !ok {
 				return nil, fmt.Errorf("partition %d not found in fallback offsets for topic %s", partition, topic)
 			}
+			if o.Offset < startOffset.At {
+				// Skip the resolved fallback offset if it's before the partition's start offset (i.e. before the earliest offset of the partition).
+				// This should not happen in Kafka, but can happen in Kafka-compatible systems, e.g. Warpstream.
+				continue
```
dimitarvdimitrov, line 66: so here we continue which effectively skips this partition. Is the assumption that the fallbackOffset call will fix itself eventually? What if we've set fallbackOffset == 12h and warpstream retention == 12h: doesn't this make this race condition more likely to occur? 

Why not just use the start offset in this case?

dimitarvdimitrov (approved):
LGTM, nice work! 🚀 kudos for getting this done!

pkg/blockbuilder/kafkautil.go
narqo: We pass all three sets of offsets to `kadm.CalculateGroupLagWithStartOffsets`. It uses the start offset if the partition doesn't have a commit, i.e. when the partition isn't in the `offsets`. So we should be good here.

PR #9203, opened 2024-09-04
kafka replay speed: use franz-go fork

dimitarvdimitrov:
#### Using a fork

Starts using a fork of franz-go until we can get a version of https://github.com/twmb/franz-go/pull/803 merged upstream. The maintainer was ok with the approach, so I think using a fork won't diverge us in the long run. 

As a result this PR also deletes all the copied code we have from franz-go for partition parsing.

#### Metrics

Using a fork allows us to just use the metrics that kprom exposes instead of tracking them ourselves. 

* `cortex_ingest_storage_reader_fetches_compressed_bytes_total` which was a copy of kprom's `cortex_ingest_storage_reader_fetch_compressed_bytes_total`
* `cortex_ingest_storage_reader_fetched_bytes_total` which was a copy of kprom's `cortex_ingest_storage_reader_fetch_bytes_total`

#### Clean up fetching

* Properly close spans when we double-fetch a record. Also emit a span event to show that we've ignored the record.

* We also assert on the response we get instead of indexing into an array and hoping it contains what we asked for. I don't think this check should ever fail in normal operations. It's there as a stopgap to consuming wrong data in case there's a bug somewhere.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


seizethedave (approved):

gotjosh (approved):
LGTM

PR #9210, opened 2024-09-05
Remove TreeQueue Tree implementation

chencs:
#### What this PR does
This PR removes the legacy `TreeQueue` implementation of `Tree`, and the experimental `use-multi-algorithm-query-queue` flag which allowed operators to switch between `TreeQueue` and the new `MultiQueuingAlgorithmTreeQueue`. It leaves in place the `Tree` interface.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/scheduler/queue/tenant_queues.go
```diff
@@ -203,23 +171,12 @@ func (qb *queueBroker) dequeueRequestForQuerier(
...
-	} else if itq, ok := qb.tree.(*MultiQueuingAlgorithmTreeQueue); ok {
-		queueNodeAfterDequeue := itq.GetNode(queuePath)
-		if queueNodeAfterDequeue == nil && len(qb.tenantQuerierAssignments.tenantNodes[string(tenantID)]) == 0 {
-			// queue node was deleted due to being empty after dequeue
-			qb.tenantQuerierAssignments.removeTenant(tenantID)
-		}
+	queueNodeAfterDequeue := qb.tree.GetNode(queuePath)
+	if queueNodeAfterDequeue == nil && len(qb.tenantQuerierAssignments.tenantNodes[string(tenantID)]) == 0 {
+		// queue node was deleted due to being empty after dequeue
+		qb.tenantQuerierAssignments.removeTenant(tenantID)
```
francoposa, line 177: Related to making `GetNode` part of the public tree interface:
not really worth tackling in this PR, but as we have moved more of the node state management back into the tree, I find myself more and more wanting to have the tree just delete the empty and report to the queueing algorithm that it did so - then each algorithm can update its internal state but not actually do the deleting.
Deleting the empty node is generally assumed to happen by all our implementations and interface dosctrings, but still has to be done by each algorithm implementation separately.

This is related to `GetNode` because this above is the only non-test usage of it, which could be eliminated if we just had the tree report that the node was deleted instead.
`GetNode` is a great thing to have for tests, but I don't really know if it belongs in the public interface in the long term, since we want all interactions from consumers of the interface to start at the root of the tree to work as expected.

chencs: I think this'll need a bit more discussion, yeah. I agree with some of your points, but no part of the tenantQuerierAssignments' `QueuingAlgorithm` implementation currently touches `tenantsByID`, and having it do so would (I think) greatly complicate the task of separating the many responsibilities `tenantQuerierAssignments` currently holds.

pkg/scheduler/queue/tenant_queues.go
```diff
@@ -99,23 +90,12 @@ func (qb *queueBroker) enqueueRequestBack(request *tenantRequest, tenantMaxQueri
...
-			itemCount += tenantNode.ItemCount()
-		}
-		if itemCount+1 > qb.maxTenantQueueSize {
-			return ErrTooManyRequests
-		}
+	itemCount := 0
+	for _, tenantNode := range qb.tenantQuerierAssignments.tenantNodes[string(request.tenantID)] {
+		itemCount += tenantNode.ItemCount()
+	}
+	if itemCount+1 > qb.maxTenantQueueSize {
```
francoposa, line 97: `if itemCount >= qb.maxTenantQueuSize` ?

chencs: The (+1) is the request we want to enqueue -- if enqueuing it would exceed the max tenant queue size, we should not enqueue it.

pkg/scheduler/queue/tenant_queues.go
francoposa: :+1: I sort of conflated two concepts:
1. Not making deleting the node the queue algorithm's responsibility
2. Reporting via `tree.Dequeue` return values to the caller that the queue path is now empty/gone so that the caller can do things `qb.tenantQuerierAssignments.removeTenant(tenantID)` without using the returned `queuePath` to reach down into the tree via `GetNode`, check for existence, then doing `removeTenant`

francoposa (approved):

pkg/scheduler/queue/tenant_queues.go
francoposa: It's just a simpler way of expressing the same thing (I think?) 
`if itemCount == qb.maxTenantQueueSize` means "is the queue is already full up, we cannot add anymore" - I just added `>=` for safety.

"Current item count plus one is greater than the limit" vs. "Current item count is already at the limit"

chencs: Oh I see what you mean; I guess it's up for debate which is easier to understand, I hadn't actually changed any logic here (the change is an indent since I got rid of a type assertion).

Issue #73, opened 2021-08-09
Fix goroutine leaks in unit tests


pracucci:
We have some unit tests leaking goroutines. Typically it's not a big deal, considering they're short lived but in some circumstances they could significantly increase the memory allocated by tests.

To give an example, in https://github.com/grafana/mimir/pull/69 we've fixed a goroutines leak in consul in-memory KV store which was accounting for > 1GB when running `pkg/storegateway`.

As a chore activity, would be great to expand the usage of `goleak` in unit tests, to eventually spot more goroutine leaks (see #69).

PR #9211, opened 2024-09-05
Alertmanager: Fix goroutine leak when stored config fails to apply and there is no existing tenant alertmanager

alexweav:
#### What this PR does

When a tenant has bad config in the store and no prior running alertmanager, we start a new one and try to apply the bad config to it. When the apply fails, we forget to shut down the alertmanager we just started.

This means we leak an alertmanager every time we sync that user's config.

There is a check for config correctness earlier on when syncing configs, but it is only a partial check - for example, it does not check templates - so it is easily bypassed (nor should we trust it). This PR fixes the leak directly.

Contrib https://github.com/grafana/mimir/issues/73

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/alertmanager/distributor_test.go
+++ b/pkg/alertmanager/distributor_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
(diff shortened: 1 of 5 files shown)


pkg/alertmanager/distributor_test.go
```diff
@@ -330,6 +336,7 @@ func prepare(t *testing.T, numAM, numHappyAM, replicationFactor int, responseBod
 
 	return d, ams, func() {
 		require.NoError(t, services.StopAndAwaitTerminated(context.Background(), d))
+		require.NoError(t, services.StopAndAwaitTerminated(context.Background(), amRing))
```
alexweav, line 339: Drive-by fix: enabling the leak detector for `TestMultitenantAlertmanager_loadAndSyncConfigs` would fail due to unrelated goroutines leaked by this test. , This test only happens to fall earlier in the test ordering.

santihernandezc (approved):
LGTM 👍

56quarters (approved):
LGTM, thanks!

PR #9212, opened 2024-09-06
MQE: Handle min/max NaN's same as promql

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/aggregations/min_max.go
```diff
@@ -34,14 +34,14 @@ func NewMinMaxAggregationGroup(max bool) *MinMaxAggregationGroup {
...
-	if !g.floatPresent[idx] || g.floatPresent[idx] && f > g.floatValues[idx] {
+	if !g.floatPresent[idx] || f > g.floatValues[idx] || math.IsNaN(g.floatValues[idx]) {
 		g.floatValues[idx] = f
 		g.floatPresent[idx] = true
 	}
 }
 
 func (g *MinMaxAggregationGroup) minAccumulatePoint(idx int64, f float64) {
-	if !g.floatPresent[idx] || g.floatPresent[idx] && f < g.floatValues[idx] {
+	if !g.floatPresent[idx] || f < g.floatValues[idx] || math.IsNaN(g.floatValues[idx]) {
```
charleskorn, line 44: [nit] Might be good to explain the expected behaviour here (eg. we return a NaN only if there are no other values to return)

charleskorn (approved):
LGTM, thanks for fixing this 🚀

PR #9218, opened 2024-09-06
Make continuous-test writes hash more uniformly.

seizethedave:
#### What this PR does
This adds more fluctuation to the byte contents of the the labels emitted by continuous-test, because if the only difference in the labels is a monotonically increasing integer, the resulting tokens are poorly distributed.

Histogram of 10K hashes before:
image

and after:
image

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,7 @@
 * [ENHANCEMENT] Include human-friendly timestamps in diffs logged when a test fails. #8630
 * [ENHANCEMENT] Add histograms to measure latency of read and write requests. #8583
 * [ENHANCEMENT] Log successful test runs in addition to failed test runs. #8817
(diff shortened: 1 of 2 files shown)


leizor (approved):
Neat!

jhalterman (approved):
Awesome

charleskorn (commented):
Would it be possible to add a second label that influences the hash, rather than modifying the `series_id` label like this?

continuous-test is primarily a testing tool, and being able to easily check for the presence of each series in a result is useful when debugging failures. If every value is multiplied by a somewhat random value, it makes it harder to quickly verify each series is present.

seizethedave:
@charleskorn awesome, that makes sense. I'll do that.

PR #9224, opened 2024-09-09
Fix issue where both recently compacted blocks and their source blocks can be skipped during querying

charleskorn:
#### What this PR does

This PR fixes an issue where both recently compacted blocks and their source blocks can be skipped during querying if store-gateways are restarting around the time of the query.

Prior to this PR:

- The default value of `-blocks-storage.bucket-store.sync-interval` is 15 minutes. Any blocks that have been uploaded within `3× -blocks-storage.bucket-store.sync-interval` are queried with best effort (ie. no retries if the first store-gateway queried doesn't have the block). So any block uploaded within the last 45 minutes was only queried with best effort by default. ([source](https://github.com/grafana/mimir/blob/d16f21ce30356d53f6882e71cc05f043266ff01a/pkg/querier/blocks_store_queryable.go#L254-L256))
- The default value of `-blocks-storage.bucket-store.ignore-deletion-marks-delay` is 60 minutes. Any blocks that were marked as deleted more than `-blocks-storage.bucket-store.ignore-deletion-marks-delay / 2` ago are queried with best effort. So any block deleted 30 minutes ago or more was only queried with best effort by default. ([source](https://github.com/grafana/mimir/blob/d16f21ce30356d53f6882e71cc05f043266ff01a/pkg/querier/blocks_store_queryable.go#L257-L260))

This meant that if blocks A, B and C are compacted into block X, and block X is uploaded at 01:00 and blocks A, B and C are marked as deleted at 01:00, then there's a 15 minute window when all blocks containing the same data are queried with best effort only:

- Old blocks A, B and C must be queried until 01:30, then are best effort until 02:00, then not queried from 02:00
- New block X is queried with best effort until 01:45, then must be queried
- Between 01:30 and 01:45, all blocks are queried with best effort

This means that between 01:30 and 01:45, we may skip querying all blocks containing the data from A, B and C, which may lead to incorrect query results. This could happen if the first store-gateways selected to query for blocks A, B, C and X are restarting, for example, and other store-gateways in the same zone (which do not own those blocks) are selected instead.

This PR changes this behaviour to avoid this problem. 

Queriers will now always query blocks marked for deletion until `-blocks-storage.bucket-store.ignore-deletion-marks-while-querying-delay` has elapsed since the block has been marked for deletion, with no window where they are queried with best effort. This CLI flag defaults to 50 minutes: 
* this is more than `3× -blocks-storage.bucket-store.sync-interval`, so queriers will start querying newly compacted blocks before stopping querying the source blocks
* this is less than `-blocks-storage.bucket-store.ignore-deletion-marks-delay`, so queriers will stop querying deleted blocks before store-gateways stop serving them

This is guaranteed to be safe: compactors mark old blocks for deletion after they upload the new block, so with this change we are always guaranteed to either:

* query all the old blocks, and optionally the new block (01:00 to 01:45 in our example above)
* query all the old blocks and the new block (01:45 to 01:50)
* query only the new block (01:50 onwards)

Visually, the changes here look like this (using the same example from before): 

PR-9224

In addition to the changes above, I've also added some validation to the various CLI flags used here to ensure they are configured in a way that ensures queries will return correct results.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -106,8 +106,7 @@ const (
 
 	DefaultMaxTSDBOpeningConcurrencyOnStartup = 10
 
-	seriesSelectionStrategyFlag = "blocks-storage.bucket-store.series-selection-strategy"
-	bucketIndexFlagPrefix       = "blocks-storage.bucket-store.bucket-index."
+	bucketIndexFlagPrefix = "blocks-storage.bucket-store.bucket-index."
 )
 
 // Validation errors
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -409,7 +409,7 @@ type BucketStoreConfig struct {
 	ChunksCache                            ChunksCacheConfig   `yaml:"chunks_cache"`
 	MetadataCache                          MetadataCacheConfig `yaml:"metadata_cache"`
 	IgnoreDeletionMarksInStoreGatewayDelay time.Duration       `yaml:"ignore_deletion_mark_delay" category:"advanced"`
-	IgnoreDeletionMarksWhileQueryingDelay  time.Duration       `yaml:"ignore_deletion_mark_while_querying_delay" category:"advanced"`
+	IgnoreDeletionMarksWhileQueryingDelay  time.Duration       `yaml:"ignore_deletion_mark_while_querying_delay" category:"experimental"`
 	BucketIndex                            BucketIndexConfig   `yaml:"bucket_index"`
 	IgnoreBlocksWithin                     time.Duration       `yaml:"ignore_blocks_within" category:"advanced"`
 
--- a/pkg/storage/tsdb/config.go
+++ b/pkg/storage/tsdb/config.go
@@ -98,6 +98,14 @@ const (
 	// partitioner aggregates together two bucket GET object requests.
 	DefaultPartitionerMaxGapSize = uint64(512 * 1024)
 
+	// NewBlockDiscoveryDelayMultiplier is the factor used (multiplied with BucketStoreConfig.SyncInterval) to determine
+	// when querying a newly uploaded block is required.
+	// For example, if BucketStoreConfig.SyncInterval is 15 minutes, then a querier will require that it can query all
+	// blocks uploaded at least 45 minutes ago, and not fail queries when a store-gateway does not respond to queries for
+	// newer blocks (those uploaded in the last 45 minutes).
+	// This gives store-gateways time to discover and load newly created blocks.
+	NewBlockDiscoveryDelayMultiplier = 3
+
 	headChunksEndTimeVarianceHelp = "How much variance (as percentage between 0 and 1) should be applied to the chunk end time, to spread chunks writing across time. Doesn't apply to the last chunk of the chunk range. 0 means no variance."
 	headStripeSizeHelp            = "The number of shards of series to use in TSDB (must be a power of 2). Reducing this will decrease memory footprint, but can negatively impact performance."
 	headChunksWriteQueueSizeHelp  = "The size of the write queue used by the head chunks mapper. Lower values reduce memory utilisation at the cost of potentially higher ingest latency. Value of 0 switches chunks mapper to implementation without a queue."
@@ -107,6 +115,10 @@ const (
 	DefaultMaxTSDBOpeningConcurrencyOnStartup = 10
 
 	bucketIndexFlagPrefix = "blocks-storage.bucket-store.bucket-index."
+
+	syncIntervalFlag                           = "blocks-storage.bucket-store.sync-interval"
+	ignoreDeletionMarksInStoreGatewayDelayFlag = "blocks-storage.bucket-store.ignore-deletion-marks-delay"
+	ignoreDeletionMarksWhileQueryingDelayFlag  = "blocks-storage.bucket-store.ignore-deletion-marks-while-querying-delay"
 )
 
 // Validation errors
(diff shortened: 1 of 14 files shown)


pkg/storage/tsdb/config.go
```diff
@@ -98,6 +98,14 @@ const (
 	// partitioner aggregates together two bucket GET object requests.
 	DefaultPartitionerMaxGapSize = uint64(512 * 1024)
 
+	// NewBlockDiscoveryDelayMultiplier is the factor used (multiplied with BucketStoreConfig.SyncInterval) to determine
+	// when querying a newly uploaded block is required.
+	// For example, if BucketStoreConfig.SyncInterval is 15 minutes, then a querier will require that it can query all
```
jhesketh, line 103: (nit)
```suggestion
	// For example, if BucketStoreConfig.SyncInterval is 15 minutes and NewBlockDiscoveryDelayMultiplier is 3, then a querier will require that it can query all
```

pkg/storage/tsdb/config.go
```diff
@@ -122,6 +133,8 @@ var (
 	errInvalidEarlyHeadCompactionMinSeriesReduction = errors.New("early compaction minimum series reduction percentage must be a value between 0 and 100 (included)")
 	errEarlyCompactionRequiresActiveSeries          = fmt.Errorf("early compaction requires -%s to be enabled", activeseries.EnabledFlag)
 	errEmptyBlockranges                             = errors.New("empty block ranges for TSDB")
+	errInvalidIgnoreDeletionMarksDelayConfig        = fmt.Errorf("value for -%s must be less than -%s", ignoreDeletionMarksWhileQueryingDelayFlag, ignoreDeletionMarksInStoreGatewayDelayFlag)
+	errIgnoreDeletionMarksDelayTooShort             = fmt.Errorf("value for -%s must be greater than %v× -%s to ensure that newly compacted blocks are queried before old blocks are ignored", ignoreDeletionMarksWhileQueryingDelayFlag, NewBlockDiscoveryDelayMultiplier, syncIntervalFlag)
```
jhesketh, line 137:
```suggestion
	errIgnoreDeletionMarksDelayTooShort             = fmt.Errorf("value for -%s must be greater than `%v × -%s` to ensure that newly compacted blocks are queried before old blocks are ignored", ignoreDeletionMarksWhileQueryingDelayFlag, NewBlockDiscoveryDelayMultiplier, syncIntervalFlag)
```

charleskorn:
> Rather than creating a new flag, would it work to have the grace period be: `storageCfg.BucketStore.IgnoreDeletionMarksDelay + (storageCfg.BucketStore.IgnoreDeletionMarksDelay - 3*storageCfg.BucketStore.SyncInterval) / 2`?

We would have to make it something like:

`3*storageCfg.BucketStore.SyncInterval + ((storageCfg.BucketStore.IgnoreDeletionMarksDelay - 3*storageCfg.BucketStore.SyncInterval) / 2)`

My personal preference is for the extra flag: to me at least, it makes the logic much clearer ("new blocks must be queried until 3× `-blocks-storage.bucket-store.sync-interval` after they're uploaded, old blocks won't be queried once `-blocks-storage.bucket-store.ignore-deletion-marks-while-querying-delay` has elapsed since they were marked as deleted"). It also gives us more flexibility to tweak things if we find the defaults I've chosen aren't quite right (eg. if it turns out we somehow need to allow more time due to clock skew).

pkg/storage/tsdb/config.go
charleskorn: I'm not sure this adds much - the constant is defined to be 3 a few lines later.

pkg/storage/tsdb/config.go
charleskorn: I'm not sure this is necessary - this description is never formatted as Markdown. (The description included on `docs/sources/mimir/configure/configuration-parameters/index.md` is in a code block, so backticks are shown literally.)

jhesketh:
> My personal preference is for the extra flag: to me at least, it makes the logic much clearer ("new blocks must be queried until 3× `-blocks-storage.bucket-store.sync-interval` after they're uploaded, old blocks won't be queried once `-blocks-storage.bucket-store.ignore-deletion-marks-while-querying-delay` has elapsed since they were marked as deleted"). It also gives us more flexibility to tweak things if we find the defaults I've chosen aren't quite right (eg. if it turns out we somehow need to allow more time due to clock skew).

Sure, I have an aversion to flags, and an arguement could be made that we can tweak the value as constants etc. but it otherwise lgtm sans nits.

jhesketh (approved):

56quarters (approved):
Thank you for fixing this.

Don't block on me but I wonder if we couldn't replace both of the "ignore deletion" settings with some multiple of the sync interval. I think adding another flag that is subtly different than an existing flag introduces a lot of cognitive overhead trying to understand why store-gateways do what they do.

Maybe we could mark the new flag as "experimental" and consolidate after testing this change for a while.

charleskorn:
> Thank you for fixing this.
> 
> Don't block on me but I wonder if we couldn't replace both of the "ignore deletion" settings with some multiple of the sync interval. I think adding another flag that is subtly different than an existing flag introduces a lot of cognitive overhead trying to understand why store-gateways do what they do.
> 
> Maybe we could mark the new flag as "experimental" and consolidate after testing this change for a while.

Makes sense to me - as I said above, I'd prefer to keep the flags as-is for now while we confirm this works as we hope, but happy to revisit the flags in the future to simplify things where we can.

And good point on marking the flag as experimental, I missed that and have changed it in 2dc7947.

PR #9225, opened 2024-09-09
MQE: Remove workaround copying spans

jhesketh:
With https://github.com/prometheus/prometheus/pull/14771 vendored in, we no longer need this workaround

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/operators/instant_vector_selector.go
charleskorn, line 1: Don't we need to remove the workaround starting on line 102/103 as well?

jhesketh: Yes, I meant to do that, thanks.

charleskorn (approved):

PR #9233, opened 2024-09-09
Add tool to upload a block to local development environment

charleskorn:
#### What this PR does

This PR adds a script that will upload a block to a local development environment for testing and debugging.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/development/tools/upload-block/upload-block.sh
+++ b/development/tools/upload-block/upload-block.sh
@@ -1,4 +1,5 @@
 #! /usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-only
 
 set -euo pipefail
 
--- a/development/tools/upload-block/upload-block.sh
+++ b/development/tools/upload-block/upload-block.sh
@@ -1,0 +1,45 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+TENANT="anonymous"
+
+function main() {
+  if [ "$#" -ne 1 ]; then
+    echo "Expected exactly one argument: the path to a block directory." >/dev/stderr
+    exit 1
+  fi
+
+  local BLOCK_DIR="$1"
+
+  if [ ! -d "$BLOCK_DIR" ]; then
+    echo "'$BLOCK_DIR' is not a directory." >/dev/stderr
+    exit 1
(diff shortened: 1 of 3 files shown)


flxbk (approved):

development/tools/upload-block/upload-block.sh
```diff
@@ -0,0 +1,45 @@
...
+
+  cat <<EOF | aws_with_creds s3 cp - "s3://mimir-tsdb/$TENANT/markers/$BLOCK_ULID-no-compact.json"
+  {
+    "id": "$BLOCK_ULID",
+    "version": 1,
+    "no_compact_time": $(date -u +%s),
+    "reason": "manual",
+    "details": "block uploaded for debugging purposes"
+  }
+EOF
```
pstibrany, line 33: we have `markblocks` tool that could be used, if it skipped the blocks-exists check.

For correctness, this marker file should also be in the block directory itself.

development/tools/upload-block/upload-block.sh
```diff
@@ -0,0 +1,45 @@
...
+  fi
+
+  local BLOCK_DIR="$1"
+
+  if [ ! -d "$BLOCK_DIR" ]; then
+    echo "'$BLOCK_DIR' is not a directory." >/dev/stderr
+    exit 1
+  fi
+
+  BLOCK_ULID="$(basename "$BLOCK_DIR")"
```
pstibrany, line 20: If `BLOCK_ULID` isn't valid ulid, uploaded block will be ignored. We could check for it by calling `ulidtime` (from mimir/tools/ulidtime) and warn if it fails.

pstibrany (approved):
Left some nit comments. Thank you.

development/tools/upload-block/upload-block.sh
charleskorn: Good point, I've switched to using `markblocks` in 28f3244.

development/tools/upload-block/upload-block.sh
charleskorn: I've added a check in ff0d022.

PR #9257, opened 2024-09-10
Add active series reloading annotation to Tenants dashboard

pr00se:
#### What this PR does

When active series are purged the counts are inaccurate until the idle timeout elapses. During this time, we stop exporting the active series counts, which can be jarring to encounter on the Tenants dashboard if you don't have the context for what is happening.

This PR adds an annotation to three panels on the Tenants dashboard that indicates when active series for the tenant are being reloaded, as indicated by the `cortex_ingester_active_series_loading` metric.

The panel IDs are currently hard-coded, which means it will break if we change the order in the dashboard 😢 but I couldn't think of a way to be more clever here, without rewriting a chunk of the jsonnet dashboard scaffolding.

SCR-20240910-ktia

Fixes n/a

#### Checklist
- [ ] Tests updated.
- [ ] Documentation added.


leizor (approved):
LGTM

PR #9260, opened 2024-09-11
Mimir query engine: fix issue where `sum` and `avg` results over native histograms can be corrupted

charleskorn:
#### What this PR does

This PR fixes an issue in `sum` and `avg` in MQE where results for native histograms could be corrupted.

This would occur when `sum` or `avg` retained a `FloatHistogram` instance, but the `HPoint` slice containing a reference to that `FloatHistogram` instance was returned to the pool and the next user of that slice modified the `FloatHistogram` slice in place before the query completed.

For example, when filling a `HPointRingBuffer` for a range vector selector, the `FloatHistogram` instances in the `HPoint` slice are reused and modified in place.

I haven't added a test for this as constructing a test for this case is quite difficult, and likely wouldn't help catch a reintroduction of this bug in the future as it would rely on the current behaviour of a variety of parts of MQE. Open to ideas on how to test this though.

(none)

#### Checklist

- [ ] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/aggregations/avg.go
```diff
@@ -166,42 +166,46 @@ func (g *AvgAggregationGroup) accumulateHistograms(data types.InstantVectorSerie
...
 
-		if g.histograms[idx] == invalidCombinationOfHistograms {
+		if g.histograms[outputIdx] == invalidCombinationOfHistograms {
 			// We've already seen an invalid combination of histograms at this timestamp. Ignore this point.
 			continue
 		}
 
-		if g.histograms[idx] == nil {
+		if g.histograms[outputIdx] == nil {
+			data.Histograms[inputIdx].H = nil // Ensure the FloatHistogram instance is not reused when the HPoint slice is reused, as we'll retain a reference to the FloatHistogram instance below.
```
charleskorn, line 179: Note to reviewers: this is the key line in this file, the rest of the diff is just renaming things

pkg/streamingpromql/aggregations/sum.go
```diff
@@ -89,34 +89,36 @@ func (g *SumAggregationGroup) accumulateHistograms(data types.InstantVectorSerie
...
 
-		if g.histogramSums[idx] == invalidCombinationOfHistograms {
+		if g.histogramSums[outputIdx] == invalidCombinationOfHistograms {
 			// We've already seen an invalid combination of histograms at this timestamp. Ignore this point.
 			continue
 		}
 
-		if g.histogramSums[idx] == nil {
+		if g.histogramSums[outputIdx] == nil {
+			data.Histograms[inputIdx].H = nil // Ensure the FloatHistogram instance is not reused when the HPoint slice is reused, as we'll retain a reference to the FloatHistogram instance below.
```
charleskorn, line 101: Note to reviewers: this is the key line in this file, the rest of the diff is just renaming things

jhesketh (changes requested):
Thanks for tracking this down, nice work!

I agree, testing in general that the buffer isn't reusing a histogram is difficult (since we want to do that in some cases), however, I think we can test that sum and avg take the histogram away from the buffer.

My suggestion is to test that after `AccumulateSeries` `data`'s HPoints all have nil Histograms.

We could potentially extend this test to function over range vector, or anything using a range vector/the buffer, and make an explicit list of things that we expect to reset the histograms on or not.

Maybe for aggregations we can have a flag just for tests along the lines of `mutatesHistograms` that if true we do this check for.

charleskorn:
Good idea, I was over thinking how to test this 🙂 

And the test revealed a case that the original fix didn't cover 🎉

pkg/streamingpromql/aggregations/sum.go
```diff
@@ -98,20 +98,27 @@ func (g *SumAggregationGroup) accumulateHistograms(data types.InstantVectorSerie
...
-			data.Histograms[inputIdx].H = nil // Ensure the FloatHistogram instance is not reused when the HPoint slice is reused, as we'll retain a reference to the FloatHistogram instance below.
-
 			if lastUncopiedHistogram == p.H {
 				// We've already used this histogram for a previous point due to lookback.
 				// Make a copy of it so we don't modify the other point.
 				g.histogramSums[outputIdx] = p.H.Copy()
 				g.histogramPointCount++
+
+				// Ensure the FloatHistogram instance is not reused when the HPoint slice is reused, as we're retaining a reference to it.
+				data.Histograms[inputIdx].H = nil
```
jhesketh, line 108: is this necessary? Aren't we creating a copy of `p.H` in this case so `data.Histograms[inputIdx].H` could actually be reused. (also applies to avg)

pkg/streamingpromql/aggregations/aggregations_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/prometheus/prometheus/promql"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func TestAggregationGroupNativeHistogramSafety(t *testing.T) {
+	// This test exists to ensure that AggregationGroup implementations correctly remove FloatHistogram instances they retain,
+	// so that retained FloatHistogram instances are not mutated when the HPoint slice they receive is later reused by something else.
```
jhesketh, line 18:
```suggestion
	// so that retained FloatHistogram instances are not mutated when the HPoint slice they received is later reused by something else.
```

pkg/streamingpromql/aggregations/aggregations_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	groups := map[string]AggregationGroup{
```
jhesketh, line 20: (nit)
```suggestion
	// These are aggregations that are expected to retain FloatHistogram instances.
	groups := map[string]AggregationGroup{
```

pkg/streamingpromql/aggregations/aggregations_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			// First series: all histograms should be nil-ed out after returning, as they're all retained for use.
+			histograms, err := types.HPointSlicePool.Get(3, memoryConsumptionTracker)
+			require.NoError(t, err)
+
+			h1 := &histogram.FloatHistogram{Sum: 1}
+			h2 := &histogram.FloatHistogram{Sum: 2}
+			histograms = append(histograms, promql.HPoint{T: 0, H: h1})
+			histograms = append(histograms, promql.HPoint{T: 1, H: h2})
+			histograms = append(histograms, promql.HPoint{T: 2, H: h2})
```
jhesketh, line 37: (nit)
```suggestion
			// T=2 is a lookback and points to the same histogram as T=1
			histograms = append(histograms, promql.HPoint{T: 2, H: h2})
```

pkg/streamingpromql/aggregations/aggregations_test.go
```diff
@@ -0,0 +1,70 @@
...
+			histograms, err = types.HPointSlicePool.Get(5, memoryConsumptionTracker)
+			require.NoError(t, err)
+			h3 := &histogram.FloatHistogram{Sum: 3}
+			h4 := &histogram.FloatHistogram{Sum: 4}
+			h5 := &histogram.FloatHistogram{Sum: 5}
+			h6 := &histogram.FloatHistogram{Sum: 6}
+			histograms = append(histograms, promql.HPoint{T: 0, H: h3})
+			histograms = append(histograms, promql.HPoint{T: 1, H: h4})
+			histograms = append(histograms, promql.HPoint{T: 2, H: h5})
+			histograms = append(histograms, promql.HPoint{T: 3, H: h5})
```
jhesketh, line 53: (nit)
```suggestion
			// T=3 is a lookback
			histograms = append(histograms, promql.HPoint{T: 3, H: h5})
```

pkg/streamingpromql/aggregations/sum.go
```diff
@@ -89,34 +89,43 @@ func (g *SumAggregationGroup) accumulateHistograms(data types.InstantVectorSerie
...
+				g.histogramSums[outputIdx] = p.H.Copy()
 				g.histogramPointCount++
+
+				// Ensure the FloatHistogram instance is not reused when the HPoint slice is reused, as we're retaining a reference to it.
+				data.Histograms[inputIdx].H = nil
+
 				continue
 			}
+
 			// This is the first time we have seen this histogram.
```
jhesketh, line 113: (nit) I think this clarification makes the below `removeReferencesToRetainedHistogram` clearer why we need to do it. Since we may have seen the histogram before (also applies to avg)
```suggestion
			// We have not previously used this histogram as the start of a point.
```

pkg/streamingpromql/aggregations/sum.go
charleskorn: `data.Histograms[inputIdx].H` refers to `lastUncopiedHistogram` in this case, so we must `nil` it out - otherwise something can mutate `lastUncopiedHistogram`.

pkg/streamingpromql/aggregations/aggregations_test.go
charleskorn: Addressed in b5ea123.

pkg/streamingpromql/aggregations/aggregations_test.go
charleskorn: Addressed in b5ea123.

pkg/streamingpromql/aggregations/aggregations_test.go
charleskorn: Addressed in b5ea123.

pkg/streamingpromql/aggregations/aggregations_test.go
charleskorn: Addressed in b5ea123.

pkg/streamingpromql/aggregations/sum.go
charleskorn: Addressed in f88c488.

pkg/streamingpromql/aggregations/common.go
```diff
@@ -32,3 +33,16 @@ var AggregationGroupFactories = map[parser.ItemType]AggregationGroupFactory{
 //
 // Invalid combinations include exponential and custom buckets, and histograms with incompatible custom buckets.
 var invalidCombinationOfHistograms = &histogram.FloatHistogram{}
+
+// removeReferencesToRetainedHistogram searches backwards through s, starting at lastIndex, removing any
```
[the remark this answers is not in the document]
charleskorn: Good point, moved in 96bbda1.

pkg/streamingpromql/aggregations/sum.go
charleskorn: In fact, we need to do this check regardless of `g.histogramSums[outputIdx]` being `nil` - let me fix that.

jhesketh: yes, but it's already nil'd out by this point right? as when we set `lastUncopiedHistogram` we call `RemoveReferencesToRetainedHistogram`

charleskorn: No - `RemoveReferencesToRetainedHistogram` only looks backwards from the current point, not forwards. If we modified the later points when we call `RemoveReferencesToRetainedHistogram`, we'd remove the histograms for later points before we'd had a chance to use them.

pkg/streamingpromql/aggregations/aggregations_test.go
```diff
@@ -28,41 +28,42 @@ func TestAggregationGroupNativeHistogramSafety(t *testing.T) {
...
-				{T: 0, H: h3},  // h3 not retained (added to h1)
-				{T: 1, H: h4},  // h4 not retained (added to h2)
-				{T: 2, H: nil}, // h5 is retained for T=3
-				{T: 3, H: nil}, // h5 is retained for this point
-				{T: 4, H: nil}, // h6 is retained for this point
+				{T: 0, H: h4},  // h4 not retained (added to h1)
+				{T: 1, H: h5},  // h5 not retained (added to h2)
+				{T: 2, H: nil}, // h6 is retained for T=3
+				{T: 3, H: nil}, // h6 is retained for this point
+				{T: 4, H: nil}, // h7 is retained for T=4
```
jhesketh, line 66: you refer to h7, but never create it  (you've reused h6)

jhesketh (approved):
one last nit, otherwise lgtm

pkg/streamingpromql/aggregations/aggregations_test.go
charleskorn: 🤦 Thanks, fixed in e7e4fb5.

PR #9266, opened 2024-09-11
Return the Flush() function in log so others can call it

zenador:
#### What this PR does

Return the Flush() function in log so others can call it

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -32,8 +32,9 @@ type RateLimitedLoggerCfg struct {
 	Registry      prometheus.Registerer
 }
 
-// InitLogger initialises the global gokit logger (util_log.Logger) and returns that logger.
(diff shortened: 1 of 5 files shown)


pkg/util/log/log.go
```diff
@@ -50,7 +51,7 @@ func InitLogger(logFormat string, logLevel dslog.Level, buffered bool, rateLimit
 
 	// Set global logger.
 	Logger = logger
-	return logger
+	return logger, Flush
```
dimitarvdimitrov, line 54: `Flush` is an exported package-level function. Can't the callers just use that?

zenador: When I first tried to use it there was a CI error complaining about the import https://github.com/grafana/backend-enterprise/actions/runs/10221465647/job/28298783016

However @narqo suggested an alternative that doesn't require this change, so I'm marking this as draft for now, and will close it if we can go ahead without it.

zenador:
No longer needed as mentioned above

PR #9267, opened 2024-09-11
add new gauge metric for alertmanager config size

titolins:
## https://github.com/grafana/alerting-squad/issues/907

### DIFF
--- a/pkg/alertmanager/api_grafana.go
+++ b/pkg/alertmanager/api_grafana.go
@@ -316,7 +316,6 @@ func (am *MultitenantAlertmanager) SetUserGrafanaConfig(w http.ResponseWriter, r
 		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errReadingGrafanaConfig, err.Error())})
 		return
 	}
-	am.multitenantMetrics.grafanaConfigSize.WithLabelValues(userID).Set(float64(len(payload)))
 
 	cfg := &UserGrafanaConfig{}
 	err = json.Unmarshal(payload, cfg)
--- a/pkg/alertmanager/api_grafana.go
+++ b/pkg/alertmanager/api_grafana.go
@@ -316,6 +316,7 @@ func (am *MultitenantAlertmanager) SetUserGrafanaConfig(w http.ResponseWriter, r
 		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errReadingGrafanaConfig, err.Error())})
 		return
 	}
+	am.configSize.WithLabelValues(userID).Set(float64(len(payload)))
 
 	cfg := &UserGrafanaConfig{}
 	err = json.Unmarshal(payload, cfg)
--- a/pkg/alertmanager/api_grafana.go
+++ b/pkg/alertmanager/api_grafana.go
@@ -316,7 +316,7 @@ func (am *MultitenantAlertmanager) SetUserGrafanaConfig(w http.ResponseWriter, r
 		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errReadingGrafanaConfig, err.Error())})
 		return
 	}
-	am.configSize.WithLabelValues(userID).Set(float64(len(payload)))
+	am.multitenantMetrics.grafanaConfigSize.WithLabelValues(userID).Set(float64(len(payload)))
 
 	cfg := &UserGrafanaConfig{}
 	err = json.Unmarshal(payload, cfg)
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -308,6 +308,7 @@ type MultitenantAlertmanager struct {
 	tenantsDiscovered prometheus.Gauge
 	syncTotal         *prometheus.CounterVec
 	syncFailures      *prometheus.CounterVec
+	configSize        *prometheus.GaugeVec
 }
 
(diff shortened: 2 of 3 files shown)


santihernandezc (changes requested):
Nice! A few comments:

pkg/alertmanager/multitenant.go
```diff
@@ -401,6 +402,10 @@ func createMultitenantAlertmanager(cfg *MultitenantAlertmanagerConfig, fallbackC
 			Name: "cortex_alertmanager_tenants_owned",
 			Help: "Current number of tenants owned by the Alertmanager instance.",
 		}),
+		configSize: promauto.With(registerer).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cortex_alertmanager_config_size_bytes",
+			Help: "Size of the last received alertmanager configuration.",
+		}, []string{"user"}),
```
santihernandezc, line 408: The name is a bit ambiguous, it's not clear that we're only using this for Grafana configurations. Maybe something like `cortex_alertmanager_grafana_config_size_bytes` is enough to avoid confusions.

The [`multitenantAlertmanagerMetrics` struct](https://github.com/grafana/mimir/blob/b810cbe07175145c9c63d046df58637916ac9a46/pkg/alertmanager/multitenant.go#L185-L188) might be a better place for defining this metric. We could add a new field (e.g. `grafanaConfigSize`) and initialize the gauge in [`newMultitenantAlertmanagerMetrics()`](https://github.com/grafana/mimir/blob/b810cbe07175145c9c63d046df58637916ac9a46/pkg/alertmanager/multitenant.go#L190-L206)

titolins: > The name is a bit ambiguous, it's not clear that we're only using this for Grafana configurations. Maybe something like cortex_alertmanager_grafana_config_size_bytes is enough to avoid confusions.

Yep, good point - changed now 👍 

> The [multitenantAlertmanagerMetrics struct](https://github.com/grafana/mimir/blob/b810cbe07175145c9c63d046df58637916ac9a46/pkg/alertmanager/multitenant.go#L185-L188) might be a better place for defining this metric. We could add a new field (e.g. grafanaConfigSize) and initialize the gauge in [newMultitenantAlertmanagerMetrics()](https://github.com/grafana/mimir/blob/b810cbe07175145c9c63d046df58637916ac9a46/pkg/alertmanager/multitenant.go#L190-L206)

Agreed. Also fixed now 👍

pkg/alertmanager/api_grafana.go
```diff
@@ -316,6 +316,7 @@ func (am *MultitenantAlertmanager) SetUserGrafanaConfig(w http.ResponseWriter, r
 		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errReadingGrafanaConfig, err.Error())})
 		return
 	}
+	am.multitenantMetrics.grafanaConfigSize.WithLabelValues(userID).Set(float64(len(payload)))
```
santihernandezc, line 319: A few problems with this approach:
- If the AM restarts, the Grafana configuration might be there, but we won't get metrics for it until the next POST request from Grafana to update the config
- The metric doesn't change when the configuration is deleted

What if we updated this metric during the config sync for each tenant? E.g. in [`computeConfig()`](https://github.com/grafana/mimir/blob/b810cbe07175145c9c63d046df58637916ac9a46/pkg/alertmanager/multitenant.go#L701), by setting it to the `len(cfgs.Grafana.RawConfig)`.

pkg/alertmanager/multitenant.go
```diff
@@ -183,13 +183,20 @@ func (cfg *MultitenantAlertmanagerConfig) CheckExternalURL(alertmanagerHTTPPrefi
...
 	lastReloadSuccessfulTimestamp *prometheus.GaugeVec
 }
 
 func newMultitenantAlertmanagerMetrics(reg prometheus.Registerer) *multitenantAlertmanagerMetrics {
 	m := &multitenantAlertmanagerMetrics{}
 
+	m.grafanaConfigSize = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "alertmanager_grafana_config_size_bytes",
+		Help:      "Size of the last received grafana alertmanager configuration.",
```
santihernandezc, line 197:
```suggestion
		Help:      "Size of the grafana alertmanager configuration.",
```

pkg/alertmanager/multitenant.go
```diff
@@ -721,10 +728,12 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 	// Grafana configuration.
 	case cfgs.Mimir.RawConfig == am.fallbackConfig:
 		level.Debug(am.logger).Log("msg", "mimir configuration is default, using grafana config with the default globals", "user", cfgs.Mimir.User)
+		am.multitenantMetrics.grafanaConfigSize.WithLabelValues(cfgs.Grafana.User).Set(float64(len(cfgs.Grafana.RawConfig)))
```
santihernandezc, line 731: We could do this once at the top of the method, before entering the `switch`. Even if the Grafana config is ignored, we'll get the size in bytes.

titolins: Sure, I wasn't sure if we actually wanted to record it in those cases.
I've moved it out of `computeConfigs` and into `syncConfigs` instead, mainly because of the user reference from the `cfgMap` but also considering there are other metrics being recorded there. Please let me know if you agree 🙌

PR #9268, opened 2024-09-11
Ruler: Make the sync queue polling intervals configurable.

stevesg:
In some environments, it may be desirable to trade off increased resource usage for faster rule group updates.

- For the outbound queue - polling more often results in more gRPC calls to other rulers, but nothing else. I think reducing this below 10s could be valuable.
- For the inbound queue - polling more often results in more rule sync operations, which can be quite expensive for tenants with many rule groups. Reducing this is likely not a wise idea.

In a follow up PR, I want to experiment with making the sync operations more specific (e.g. include rule group), this would potentially allow more frequent inbound queue polling.

Part of https://github.com/grafana/mimir/issues/9329

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -12658,7 +12658,7 @@
           "kind": "field",
           "name": "inbound_sync_queue_poll_interval",
           "required": false,
-          "desc": "Interval between sending queued rule sync requests to ruler replicas.",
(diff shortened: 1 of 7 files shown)


gotjosh (approved):
LGTM

gotjosh:
Remember to modify the about versioning documentation.

tacole02 (approved):
😄

PR #9204, opened 2024-09-04
kafka replay speed: compressed/uncompressed bytes for MaxBytes in fetch requests

dimitarvdimitrov:
Normally all kafka flavours support the same API, except when they don't. Warpstream treats the `MaxBytes` field of a Fetch request as a limit on the uncompressed bytes of all the records. Apache Kafka treats it as a limit on the compressed bytes.

This difference in behaviour is quite significant and leads to very suboptimal performance when using the wrong interpretation.

This PR adds a config flag to be able to choose between the Kafka flavours. Open to feedback on what to call the flag and what to call the options.

#### Note to reviewers
 
This PR is based on #9203, but will also conflict with #9153 when merged. The conflicts should be tiny, so I think this is ready for a review. I'll rebase when the two PRs above are merged.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9269, opened 2024-09-11
kafka replay speed: instrument concurrent fetching pipeline

dimitarvdimitrov:
#### Note to reviewers

This PR is based on #9204 to avoid small paper-cut conflicts. The changes here are mostly independent of the changes there.

#### What this PR does

* Replaces `cortex_ingest_storage_reader_fetched_discarded_records_total` with  `cortex_ingest_storage_reader_fetched_discarded_bytes_total` so that we can better estimate whether overfetching is a real problem (avoid the question "are 1000 overfetched records too much?")
* Adds tracing for the fetch pipeline - there's a single trace for a single continuous section of records. My idea was to allow debugging increased e2e latency based on this. I want to add some events in the `PusherConsumer` that might help with debugging sharding performance, but I'll do that after #9133 is merged to avoid conflicts
  * I tried including necessary information to be able to infer almost all of the internal state like bytes per record, backoff duration, error handling without cluttering the trace too much
  
Screenshot 2024-09-11 at 12 10 00

* Adds jaeger all-in-one to the ingest-storage docker-compose to see the produced traces

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/development/mimir-ingest-storage/config/datasource-mimir.yaml
+++ b/development/mimir-ingest-storage/config/datasource-mimir.yaml
@@ -10,3 +10,9 @@ datasources:
   jsonData:
     prometheusType: Mimir
     timeInterval: 5s
+- name: Jaeger
+  type: jaeger
+  access: proxy
+  uid: jaeger
+  orgID: 1
+  url: http://jaeger:16686/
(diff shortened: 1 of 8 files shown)


gotjosh (approved):
LGTM

PR #8201, opened 2024-05-28
Clamp GOMAXPROCS when higher than runtime.NumCPU

dimitarvdimitrov:
#### Background

We are trying to automatically set GOMAXPROCS based on the number of CPUs that an ingester pod requests in Kubernetes. We're going with 2x the requested cores. The reason for this is that the default values of GOMAXPROCS is NumCPU. When running on large nodes and only utilizing a small % of the underlying node results in high scheduling overhead.

#### Problem

Sometimes the setting of GOMAXPROCS might exceed the number of cores of the node. We also don't want to restrict the nodes on which pods run. In those cases setting GOMAXPROCS to a higher value than NumCPU has the opposite effect - it increases scheduling overhead instead of reducing it.

#### What this PR does

Clamps the value of GOMAXPROCS to runtime.NumCPU. The idea of this PR is to basically make automating the GOMAXPROCS setting in deployment tooling easier by having some support from the code.

#### Considerations

It's possible that the original NumCPU is not correctly detected or that the operator intended to run with higher GOMAXPROCS. I didn't want to add more configuration options, so instead we're logging a warning message. If there is a real use case for GOMAXPROCS > NumCPU, then we can add the configuration option later.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9273, opened 2024-09-11
jsonnet: set GOMAXPROCS on ingesters

dimitarvdimitrov:
#### What this PR does

We've been experimenting with GOMAXPROCS on ingesters at GL for the past few months. The test included all production cells and was comparing ingester-zone-a and zone-b (with GOMAXPROCS) to zone-c (without GOMAXPROCS)

 This is the summary of results:
 > * Push & Query outlier pods are significantly lower with GOMAXPROCS. This makes the cell more resilient to bad nodes and to inefficient NUMA node assignment; this was the goal of the experiment
 > * QueryStream performance under sudden load is worse
 > * CPU usage is generally slightly lower. Reduced available cores via GOMAXPROCS explains this
 > * heap might appear slightly increased. However, reduced CPU time (due to fewer cores, see point above) for garbage collection can explain why peak heap can be left to rise before GC kicks in. This is generally safe and should not lead to OOMs for example
 >
 > To me GOMAXPROCS is a tradeoff between lower stable-state resource usage and resilience to outlier infrastructure **and** being able to handle sudden usage peaks without increasing latency. With the help of query-scheduler scheduling improvements (prioritizing queries that wouldn't touch slowed-down components) and autoscaling (more frequent ingester vertical autoscaling, querier autoscaling) I think we can accept the increased QueryStream latency.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir/ingester.libsonnet
+++ b/operations/mimir/ingester.libsonnet
@@ -56,6 +56,20 @@
 
   ingester_env_map:: {
     JAEGER_REPORTER_MAX_QUEUE_SIZE: '1000',
+
+    local requests = $.util.parseCPU($.ingester_container.resources.requests.cpu),
+    local requestsWithHeadroom = std.ceil(
+      // double the requests, but with headroom of at least 3 and at most 6 CPU
+      // too little headroom can lead to throttling
+      // too much headroom can lead to setting the limit above the available CPU cores
+      requests + std.max(3, std.min(6, requests)),
+    ) + 1,
+    // If the ingester read path is limited,
+    // we don't want to set GOMAXPROCS to something higher than that because then the read path limit will never be hit.
+    local limitBasedGOMAXPROCS = std.parseInt($.ingester_args['ingester.read-path-cpu-utilization-limit']) + 1,
+    local GOMAXPROCS = if 'ingester.read-path-cpu-utilization-limit' in $.ingester_args then limitBasedGOMAXPROCS else requestsWithHeadroom,
+
+    GOMAXPROCS: std.toString(GOMAXPROCS),
   },
 
   ingester_node_affinity_matchers:: [],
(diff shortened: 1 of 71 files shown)


operations/mimir/ingester.libsonnet
```diff
@@ -56,6 +56,20 @@
 
   ingester_env_map:: {
     JAEGER_REPORTER_MAX_QUEUE_SIZE: '1000',
+
+    local requests = $.util.parseCPU($.ingester_container.resources.requests.cpu),
+    local requestsWithHeadroom = std.ceil(
+      // double the requests, but with headroom of at least 3 and at most 6 CPU
+      // too little headroom can lead to throttling
+      // too much headroom can lead to setting the limit above the available CPU cores
+      requests + std.max(3, std.min(6, requests)),
```
narqo, line 65: Wouldn't this _potentially_ set it to more than what is available for the host/container? E.g. if a VM has only 4 cores, and there is only one container running on the host (kinda line in AWS fargate), we set maxprocs to 7?

dimitarvdimitrov: very good point, Peter already thought about that 🙂

*  #8201

narqo: Oh, I didn't know that 👍  

Maybe it's worth it to link to clamping PR from the comment, to make sure this doesn't trigger anyone later

```suggestion
      // double the requests, but with headroom of at least 3 and at most 6 CPU
      // too little headroom can lead to throttling
      // too much headroom can lead to setting the limit above the available CPU cores
      // if the value exceeds the number of available CPU cores on the host, mimir caps the maxprocs on application start (ref grafana/mimir#8201)
      requests + std.max(3, std.min(6, requests)),
```

narqo (approved):
🔥 works for me, although I'd expand the "headroom" comment.

Also, CI seems to be unhappy.

PR #9277, opened 2024-09-12
Mimir query engine: binary operations between two scalars

charleskorn:
#### What this PR does

This PR adds support for binary operations between two scalars in MQE.

For example, expressions like `2 + 3` and `scalar(metric) * 2` are now supported.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/engine_test.go
```diff
@@ -88,6 +87,15 @@ func TestUnsupportedPromQLFeaturesWithFeatureToggles(t *testing.T) {
 
 		requireRangeQueryIsUnsupported(t, featureToggles, "metric{} + other_metric{}", "binary expressions")
 		requireInstantQueryIsUnsupported(t, featureToggles, "metric{} + other_metric{}", "binary expressions")
+
+		requireRangeQueryIsUnsupported(t, featureToggles, "metric{} + 1", "binary expressions")
```
jhesketh, line 91: (nit) etc..
```suggestion
		requireRangeQueryIsUnsupported(t, featureToggles, "metric{} + 1", "binary expressions with scalar and instant")
```

pkg/streamingpromql/operators/scalar_scalar_binary_operation.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// Binary operations between two scalars always produce a float value, as only arithmetic operators or comparison
+	// operators (with the bool keyword) are supported between two scalars.
+	//
+	// Furthermore, scalar values always have a value at each step.
+	//
+	// So we can just compute the result of each pairwise operation without examining the timestamps of each sample.
+	//
+	// We store the result in the slice from the left operator, and return the right operator's slice once we're done.
```
jhesketh, line 69: I think for safety we should have a check such as `len(leftValues) != len(rightValues)` and raise an error

jhesketh (changes requested):
It would be good to see a test with `scalar(series)` (or am I blind)

pkg/streamingpromql/operators/scalar_scalar_binary_operation.go
charleskorn: We rely on this fact in other places, and don't have checks there either (eg. scalar / vector binary operations). 

If the lengths are different, the code will panic, which seems like a reasonable thing to do for a bug where an invariant has been broken.

pkg/streamingpromql/engine_test.go
charleskorn: Adding this would complicate [the code](https://github.com/grafana/mimir/blob/6a54260008a00d5c848eb0870527a2d897a0d09f/pkg/streamingpromql/query.go#L182) that generates the error message, and given this message will now only be generated when the feature toggle is turned off, I'm not sure it's worth it.

charleskorn:
> It would be good to see a test with `scalar(series)` (or am I blind)

There were existing tests that exercised `scalar()`, but I've added one specifically that uses `scalar(metric) + 2` to test it with binary operations between two scalars.

pkg/streamingpromql/operators/scalar_scalar_binary_operation.go
jhesketh: It only panics if the right side is shorter than the left. I don't think the overhead here is very high, what do you think about it adding it to the other scalar operators too?

charleskorn: I'm not sure this makes sense as a precedent to set: we don't do similar validation-type checks in other situations (eg. checking that the data points from an instant vector operator are within the query time range, or align to the requested step), and if those situations were to occur, we'd have similarly undefined behaviour.

jhesketh: Hmm, okay. I still think it's reasonable, but I don't hold it strongly. LGTM :-)

jhesketh (approved):

PR #9280, opened 2024-09-12
Mimir query engine: extract query time range values to their own type

charleskorn:
#### What this PR does

This PR refactors some parts of MQE to pass around a `types.QueryTimeRange` struct rather than a bunch of `int64`s that represent the start, end and interval of a query.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


jhesketh (approved):
Nice! I've been thinking we should tidy this up for a while too

PR #9281, opened 2024-09-12
MQE: Add tests for various mixed data and edge cases

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


jhesketh:
FYI this test seems expensive (it runs a lot of permutations), but it only takes < 5s on my laptop. We can reduce some of the cases if it grows too much as we add more operations/functions etc.

pkg/streamingpromql/engine_test.go
```diff
@@ -1674,3 +1674,131 @@ func TestAnnotations(t *testing.T) {
...
+		interval time.Duration
+	}{
+		{loadStep: 1, interval: 1 * time.Minute},
+		{loadStep: 1, interval: 6 * time.Minute},
+		{loadStep: 1, interval: 5 * time.Minute},
+		{loadStep: 6, interval: 6 * time.Minute},
+		{loadStep: 6, interval: 5 * time.Minute},
+	}
+
+	// Total tests:
```
charleskorn, line 1772: Something else that would be good to test here is something with a range selector (eg. `rate(foo[1m])`), and combinations with aggregations etc. (eg. `sum(rate(foo[1m]))`) - this might have helped catch the issue fixed in https://github.com/grafana/mimir/pull/9260.

charleskorn (commented):
Nice idea, thanks for putting this together!

pkg/streamingpromql/query.go
```diff
@@ -546,6 +549,10 @@ func (q *Query) populateVectorFromInstantVectorOperator(ctx context.Context, o t
 }
 
 func (q *Query) populateMatrixFromInstantVectorOperator(ctx context.Context, o types.InstantVectorOperator, series []types.SeriesMetadata) (promql.Matrix, error) {
+	if len(series) == 0 {
```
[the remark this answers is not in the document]
jhesketh: I was initially doing a crude `requires.Equal` between the engines and the results would be different where one is nil and one is empty. It's not necessary but makes the results more like prometheus. I'll fix the inconsistency, but also happy to take it out if you prefer.

pkg/streamingpromql/testing.go
```diff
@@ -24,3 +30,196 @@ func NewTestEngineOpts() EngineOpts {
...
+	require.Equal(t, len(b1), len(b2), "bucket lengths match")
+	for i, b := range b1 {
+		require.InEpsilon(t, b, b2[i], 1e-10, "bucket values match")
+	}
+}
+
+// Copied from prometheus as it is not exported
+// spansMatch returns true if both spans represent the same bucket layout
+// after combining zero length spans with the next non-zero length span.
+func spansMatch(s1, s2 []histogram.Span) bool {
```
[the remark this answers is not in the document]
jhesketh: Good point, I think this can be much simpler.

pkg/streamingpromql/engine_test.go
jhesketh: Good idea. I had intended to do that later, but no need to hold off.

This actually uncovered an inconsistency in handling annotations that cause a bit of a headache :-/. But good for tracking down differences!

jhesketh:
This uncovered a difference in annotation handling. I looked at making MQE return the same results, but handling it was messy and I think it's easier to fix in Prometheus: https://github.com/prometheus/prometheus/pull/14910

jhesketh:
https://github.com/prometheus/prometheus/pull/14910 has merged, but it may be a while before prometheus is vendored back into mimir. Until then this PR temporarily doesn't compare annotations.

pkg/streamingpromql/types/metric_names.go
charleskorn, line 1: Why did this need to move?

pkg/streamingpromql/engine_test.go
```diff
@@ -1682,3 +1702,133 @@ func TestAnnotations(t *testing.T) {
...
+
+	for _, labels := range labelCombinations {
+		labelRegex := strings.Join(labels, "|")
+		// Aggregations
+		for _, aggFunc := range []string{"sum", "avg", "min", "max"} {
+			expressions = append(expressions, fmt.Sprintf(`%s(series{label=~"(%s)"})`, aggFunc, labelRegex))
+			expressions = append(expressions, fmt.Sprintf(`%s by (group) (series{label=~"(%s)"})`, aggFunc, labelRegex))
+			expressions = append(expressions, fmt.Sprintf(`%s without (group) (series{label=~"(%s)"})`, aggFunc, labelRegex))
+		}
+		expressions = append(expressions, fmt.Sprintf(`rate(series{label=~"(%s)"}[45s])`, labelRegex))
```
charleskorn, line 1793: Why 45s? Doesn't this mean `rate` will always see at most one input sample per output sample and so never produce any output samples?

pkg/streamingpromql/engine_test.go
```diff
@@ -1682,3 +1702,133 @@ func TestAnnotations(t *testing.T) {
...
+				defer q.Close()
+				expectedResults := q.Exec(context.Background())
+
+				q, err = mimirEngine.NewRangeQuery(context.Background(), storage, nil, expr, start, end, tr.interval)
+				require.NoError(t, err)
+				defer q.Close()
+				mimirResults := q.Exec(context.Background())
+
+				// We currently omit checking the annotations due to a difference between the engines.
+				// This can be re-enabled once https://github.com/prometheus/prometheus/pull/14910 is vendored.
```
charleskorn, line 1829: As mentioned offline - given https://github.com/prometheus/prometheus/pull/14910 has been merged, could we cherry-pick it into mimir-prometheus and then vendor that into Mimir so we don't need these workarounds without needing to do the pending Prometheus 3.0 merge?

pkg/streamingpromql/engine_test.go
```diff
@@ -1628,6 +1628,26 @@ func TestAnnotations(t *testing.T) {
 				`PromQL info: metric might not be a counter, name does not end in _total/_sum/_count/_bucket: "other_float_metric" (1:99)`,
 			},
 		},
+		"no rate annotation when sample ends with histogram": {
```
charleskorn, line 1631: Would a description like "rate() over metric without counter suffix with single float or histogram in range" be clearer? (the use of "sample" in the current description is a bit ambiguous - it's not clear if it's referring to the input or output)

pkg/streamingpromql/engine_test.go
```diff
@@ -1628,6 +1628,26 @@ func TestAnnotations(t *testing.T) {
...
+			data: `
+				series 3 1 {{schema:3 sum:12 count:7 buckets:[2 2 3]}}
+			`,
+			expr:                       "rate(series[45s])",
+			expectedWarningAnnotations: []string{},
+			expectedInfoAnnotations:    []string{},
+			// This can be removed once https://github.com/prometheus/prometheus/pull/14910 is vendored.
+			skipComparisonWithPrometheusReason: "Prometheus only considers the type of the last point in the vector selector rather than the output value",
+		},
+		"no rate annotation when only 1 point in range vector": {
```
charleskorn, line 1641: [nit] For consistency with the other names, perhaps "rate() over one point in range" would be a better name here

pkg/streamingpromql/engine_test.go
```diff
@@ -1628,6 +1628,26 @@ func TestAnnotations(t *testing.T) {
...
+			data: `
+				series 1
+			`,
+			expr:                       "rate(series[1m])",
+			expectedWarningAnnotations: []string{},
+			expectedInfoAnnotations:    []string{},
+			// This can be removed once https://github.com/prometheus/prometheus/pull/14910 is vendored.
+			skipComparisonWithPrometheusReason: "Prometheus only considers the type of the last point in the vector selector rather than the output value",
+		},
```
charleskorn, line 1650: [nit] It'd be good to move these cases next to the other cases above for `rate()`

pkg/streamingpromql/query.go
```diff
@@ -493,7 +493,10 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
 		return &promql.Result{Err: compat.NewNotSupportedError(fmt.Sprintf("unsupported result type %s", parser.DocumentedType(q.statement.Expr.Type())))}
 	}
 
-	q.result.Warnings = *q.annotations
+	// To make comparing to Prometheus' engine easier, only return Annotations if there are some, otherwise, return nil.
```
charleskorn, line 496: [nit] 

```suggestion
	// To make comparing to Prometheus' engine easier, only return the annotations if there are some, otherwise, return nil.
```

pkg/streamingpromql/query.go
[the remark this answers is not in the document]
charleskorn: I'd prefer to take it out - I'd prefer to keep this as simple as possible, and I think the special case you've added later (on line 581) will cover this scenario too.

pkg/streamingpromql/testing.go
```diff
@@ -24,3 +30,142 @@ func NewTestEngineOpts() EngineOpts {
 		Pedantic:       true,
 	}
 }
+
+// Why do we do this rather than require.Equal(t, expected, actual)?
+// It's possible that floating point values are slightly different due to imprecision, but require.Equal doesn't allow us to set an allowable difference.
+func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Result, checkAnnotations bool) {
```
charleskorn, line 36: What do you think about moving this to a test utilities package or something like that?

As it stands, this introduces a production dependency on `testify/require`, which feels a bit weird.

pkg/streamingpromql/types/metric_names.go
jhesketh: I had refactored it for somethign else, then undid the other thing, but decided to keep it here since it seemed to make more sense to me. We can move it back though if you'd prefer.

pkg/streamingpromql/engine_test.go
jhesketh: Yes. This is an edge case I want to test. It can also affect the lookback etc.

pkg/streamingpromql/engine_test.go
jhesketh: Lets deal with this separately as I think it's a bigger and more complicated task and we have a way forward here for now.

pkg/streamingpromql/types/metric_names.go
charleskorn: Let's move it back - it's only used by operators so I think it makes sense to keep it there until we have a case to use it elsewhere.

pkg/streamingpromql/engine_test.go
charleskorn: Might be good to add a comment explaining that, and would be good to add a case that does select multiple input samples as well.

charleskorn (approved):
🚀

PR #9287, opened 2024-09-12
kafka replay speed: add support for metadata & source

dimitarvdimitrov:
#### What this PR does

1. Parallelization based on: 
   - The `parallelStoragePusher` now supports separate pushers for different write request sources (e.g., API, RULE) per tenant.
   - This is necessary because the ingester keep track of metrics based on the source of the request.
   - The tradeoff is that the config option for batch size isn't what it looks like - we have `2N` batches per tenant, not the `N` that's configured.

2. Metadata Handling:
   - Added support for pushing metadata alongside time series data.
   - Metadata is distributed across shards in a round-robin fashion.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -290,7 +290,8 @@ type parallelStoragePusher struct {
 	metrics *pusherConsumerMetrics
 	logger  log.Logger
 
-	pushers        map[string]map[mimirpb.WriteRequest_SourceEnum]*parallelStorageShards
+	// pushers is map["$tenant|$source"]*parallelStorageShards
+	pushers        map[string]*parallelStorageShards
 	upstreamPusher Pusher
 	numShards      int
 	batchSize      int
@@ -300,7 +301,7 @@ type parallelStoragePusher struct {
 func newParallelStoragePusher(metrics *pusherConsumerMetrics, pusher Pusher, numShards int, batchSize int, logger log.Logger) *parallelStoragePusher {
 	return &parallelStoragePusher{
 		logger:         log.With(logger, "component", "parallel-storage-pusher"),
-		pushers:        make(map[string]map[mimirpb.WriteRequest_SourceEnum]*parallelStorageShards),
+		pushers:        make(map[string]*parallelStorageShards),
 		upstreamPusher: pusher,
 		numShards:      numShards,
 		batchSize:      batchSize,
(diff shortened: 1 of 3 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -471,24 +492,28 @@ func newBatchingQueue(capacity int, batchSize int) *batchingQueue {
 
 // AddToBatch adds a time series to the current batch. If the batch size is reached, the batch is pushed to the Channel().
 // If an error occurs while pushing the batch, it returns the error and ensures the batch is pushed.
-func (q *batchingQueue) AddToBatch(ctx context.Context, ts mimirpb.PreallocTimeseries) error {
+func (q *batchingQueue) AddToBatch(ctx context.Context, source mimirpb.WriteRequest_SourceEnum, ts mimirpb.PreallocTimeseries) error {
```
dimitarvdimitrov, line 495: so these methods keep growing. I'm wondering if we shouldn't extract `(context.Context, mimirpb.WriteRequest_SourceEnum)` into the constructor of `batchingQueue` and use them from there? It also conveys the assumption that we can only use one of them at a time, and not every one which is passed to AddToBatch and AddMetadataToBatch actually makes a difference

gotjosh (approved):
LGTM - we discussed optimising the metadata sending offline (whether we wanted to keep them separate or mix both timeseries + metadata at the same time) but ended settling for separate them.

Our thinking here is that expensive price to pay will be within `:= p.pusher.PushToStorage` and no the batching so mixing them up could have helped us get some speed.

pkg/storage/ingest/pusher.go
gotjosh: I think this is fine for now, 3 is still manageable - if we add the 4 one, maybe we should be looking at doing what you've said but for now, it's fine.

pkg/storage/ingest/pusher.go
```diff
@@ -285,7 +285,7 @@ type parallelStoragePusher struct {
 	metrics *pusherConsumerMetrics
 	logger  log.Logger
 
-	pushers        map[string]*parallelStorageShards
+	pushers        map[string]map[mimirpb.WriteRequest_SourceEnum]*parallelStorageShards
```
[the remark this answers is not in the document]
dimitarvdimitrov: did that in 83f04bbd85f5dbd835cd365d29fd32d9b342cf43

PR #9291, opened 2024-09-12
Splitting querier worker connection management out of tenantQuerierAssignments

chencs:
#### What this PR does
This PR moves logic that governs querier-worker connections to the request queue _out_ of `tenantQuerierAssignments` and into its own struct (`querierConnections`). The `queueBroker` is now responsible for coordinating `tenantQuerierAssignments` and `querierConnections`, making sure tenant-querier assignments are reshuffled when necessary on querier connection and removal/disconnection.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/scheduler/queue/querier_connections.go
+++ b/pkg/scheduler/queue/querier_connections.go
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
 package queue
 
 import (
--- a/pkg/scheduler/queue/querier_connections.go
+++ b/pkg/scheduler/queue/querier_connections.go
@@ -7,6 +7,8 @@ import (
 
 const unregisteredWorkerID = -1
 
+// querierConnections manages information about queriers connected to the request queue. The queueBroker
+// receives information about querier connections via RequestQueue's querierWorkerOperations channel.
 type querierConnections struct {
 	queriersByID map[QuerierID]*querierState
 
@@ -23,18 +25,19 @@ func newQuerierConnections(forgetDelay time.Duration) *querierConnections {
 }
 
 // querierIsAvailable returns true if the querier is registered to the querierConnections and is not shutting down.
-func (qcm *querierConnections) querierIsAvailable(querierID QuerierID) bool {
-	q := qcm.queriersByID[querierID]
+func (qc *querierConnections) querierIsAvailable(querierID QuerierID) bool {
+	q := qc.queriersByID[querierID]
 	return q != nil && !q.shuttingDown
 }
 
-// does not manage updating querierIDsSorted; that should be managed by the caller/whatever cares about it getting updated.
-func (qcm *querierConnections) addQuerierWorkerConn(conn *QuerierWorkerConn) (addQuerier bool) {
+// addQuerierWorkerConn is called when the queueBroker processes a querierWorkerOperation; it adds the querier-worker
+// connection, creating a new querier connection if we've never seen this querier before.
+func (qc *querierConnections) addQuerierWorkerConn(conn *QuerierWorkerConn) (addQuerier bool) {
 	if conn.IsRegistered() {
 		panic("received request to register a querier-worker which was already registered")
 	}
 
-	querier := qcm.queriersByID[conn.QuerierID]
+	querier := qc.queriersByID[conn.QuerierID]
 	if querier != nil {
 		querier.AddWorkerConn(conn)
 
@@ -48,13 +51,17 @@ func (qcm *querierConnections) addQuerierWorkerConn(conn *QuerierWorkerConn) (ad
 	// First connection from this querier.
 	newQuerierConns := &querierState{}
 	newQuerierConns.AddWorkerConn(conn)
-	qcm.queriersByID[conn.QuerierID] = newQuerierConns
+	qc.queriersByID[conn.QuerierID] = newQuerierConns
 
 	return true
 }
 
-func (qcm *querierConnections) removeQuerierWorkerConn(conn *QuerierWorkerConn, now time.Time) (removedQuerier bool) {
-	querier := qcm.queriersByID[conn.QuerierID]
+// removeQuerierWorkerConn removes a registered QuerierWorkerConn from an active querier. If the removed querier-worker
+// connection is the last active worker connection for the querier, it also deletes the querier connection, or records
+// the disconnection time so the querier can be forgotten if it does not establish any new querier-worker connections
+// before querierForgetDelay time has passed.
+func (qc *querierConnections) removeQuerierWorkerConn(conn *QuerierWorkerConn, now time.Time) (removedQuerier bool) {
+	querier := qc.queriersByID[conn.QuerierID]
 	if querier == nil || !querier.IsActive() {
 		panic("unexpected number of connections for querier")
 	}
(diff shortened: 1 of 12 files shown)


pkg/scheduler/queue/tenant_queues.go
```diff
@@ -190,17 +193,39 @@ func (qb *queueBroker) dequeueRequestForQuerier(
...
 
+// forgetDisconnectedQueriers removes all queriers which have had zero connections for longer than the forget delay.
+// Returns true if tenant-querier reshard was triggered.
 func (qb *queueBroker) forgetDisconnectedQueriers(now time.Time) (resharded bool) {
-	return qb.tenantQuerierAssignments.forgetDisconnectedQueriers(now)
+	queriersToRemove := qb.querierConnections.forgettableQueriers(now)
+	for _, querierID := range queriersToRemove {
+		qb.querierConnections.removeQuerier(querierID)
+	}
+	return qb.tenantQuerierAssignments.removeQueriers(queriersToRemove...)
```
francoposa, line 230: IMO we might have decomposed the steps a touch too much here - if we have no usage of `forgettableQueriers` where we do anything but remove each one, I think we should wrap it up into one interface:

instead of this
```golang
	queriersToRemove := qb.querierConnections.forgettableQueriers(now)
	for _, querierID := range queriersToRemove {
		qb.querierConnections.removeQuerier(querierID)
	}
	return qb.tenantQuerierAssignments.removeQueriers(queriersToRemove...)
```

we could have

```golang
forgottenQueriers := qb.querierConnections.removeForgettableQueriers(now)
return qb.tenantQuerierAssignments.removeQueriers(forgottenQueriers...)
```

So `queueBroker.forgetDisconnectedQueriers` is still serving this responsibility of "coordinating the two things that need done at the same time", but the `querierConnections` is still encapsulating the full step it is responsible for instead of making the caller do it

pkg/scheduler/queue/tenant_queues.go
```diff
@@ -190,17 +193,39 @@ func (qb *queueBroker) dequeueRequestForQuerier(
...
+		return qb.tenantQuerierAssignments.removeQueriers(conn.QuerierID)
+	}
+	return false
 }
 
+// notifyQuerierShutdown handles a graceful shutdown notification from a querier.
+// Returns true if tenant-querier reshard was triggered.
 func (qb *queueBroker) notifyQuerierShutdown(querierID QuerierID) (resharded bool) {
-	return qb.tenantQuerierAssignments.notifyQuerierShutdown(querierID)
+	if removedQuerier := qb.querierConnections.shutdownQuerier(querierID); removedQuerier {
```
francoposa, line 216: same note as below - let `shutdownQuerier` do the removing. There's no case at the moment where the caller should do anything else, so don't give them the chance.

francoposa: also if we have these two usages of `removeQuerier`, moved behind the respective `shutdownQuerier` and `removeForgettableQueriers` interfaces, there are no `removeQuerier` usages outside the `querier_worker_conn_manager`.

If we are not going to draw a package boundary and use public/private methods to prevent unwanted usage, we could still just delete `removeQuerier` to show that it's not something that should be done outside the context of the pre-prepared use cases of shutting down or forgetting queriers.

pkg/scheduler/queue/tenant_queues_test.go
```diff
@@ -916,3 +918,36 @@ func (n *Node) nodeCount() int {
...
 	return count
 }
+
+// getNextTenantForQuerier is a test utility, not intended for use by consumers of queueBroker
+//
+// The next tenant for the querier is obtained by rotating through the global tenant order
+// starting just after the last tenant the querier received a request for, until a tenant
+// is found that is assigned to the given querier according to the querier shuffle sharding.
+// A newly connected querier provides lastTenantIndex of -1 in order to start at the beginning.
+func (qb *queueBroker) getNextTenantForQuerier(lastTenantIndex int, querierID QuerierID) int {
```
francoposa, line 928: I have not bothered with this much myself, but maybe we can move toward just making the relevant type the first argument of a function:
`func getNextTenantForQuerier(qb *queueBroker, lastTenantIndex int, querierID QuerierID) int {`
instead of using methods and commenting them to say test utility only.

pkg/scheduler/queue/querier_connections.go
```diff
@@ -0,0 +1,217 @@
...
+	// place in graceful shutdown state; any queued requests to dispatch queries
+	// to this querier will receive error responses until all querier workers disconnect
+	querier.shuttingDown = true
+	return false
+}
+
+// forgettableQueriers removes all querier connections which no longer have any querier-worker connections and for whom
+// querierForgetDelay time has passed since the querier disconnected. It returns a slice of all querier IDs which were
+// removed.
+func (qc *querierConnections) forgettableQueriers(now time.Time) []QuerierID {
```
francoposa, line 119: can we rename to express that they were in fact removed/forgotten

pkg/scheduler/queue/queue_test.go
```diff
@@ -29,17 +29,14 @@ import (
...
 	return []struct {
 		name                      string
-		useMultiAlgoTreeQueue     bool
 		prioritizeQueryComponents bool
 	}{
-		{"legacy tree queue with prioritize query components disabled", false, false},
-		{"integrated tree queue with prioritize query components disabled", true, false},
-		{"integrated tree queue with prioritize query components enabled", true, true},
+		{"prioritize query components disabled", false},
+		{"prioritize query components enabled", true},
```
francoposa, line 39: :rocket:

pkg/scheduler/queue/tenant_querier_assignment.go
```diff
@@ -286,91 +163,27 @@ func (tqa *tenantQuerierAssignments) removeTenant(tenantID TenantID) {
...
-	if querier.shuttingDown || tqa.querierForgetDelay == 0 {
-		return tqa.removeQuerier(conn.QuerierID)
-	}
-
-	// No graceful shutdown has been notified yet, so we should track the current time
-	// so that we'll remove the querier as soon as we receive the graceful shutdown
-	// notification (if any) or once the threshold expires.
-	querier.disconnectedAt = now
-	return false
+// TODO (casie): Docstring
```
francoposa, line 166: outstanding TODO docstring

pkg/scheduler/queue/tenant_querier_assignment.go
```diff
@@ -286,91 +163,27 @@ func (tqa *tenantQuerierAssignments) removeTenant(tenantID TenantID) {
...
+	tqa.querierIDsSorted = append(tqa.querierIDsSorted, querierID)
+	sort.Sort(tqa.querierIDsSorted)
 }
 
-// removeQuerier deletes a querier from the tenant-querier assignments.
-// Returns true if tenant-querier reshard was triggered.
-func (tqa *tenantQuerierAssignments) removeQuerier(querierID QuerierID) (resharded bool) {
-	delete(tqa.queriersByID, querierID)
+// removeQueriers deletes an arbitrary number of queriers from the querier connection manager, and returns true if
+// recomputing tenant-querier sharding was necessary.
```
francoposa, line 174: imo "was necessary" leaves a little ambiguity about if it happened - let's just say if the tenant-querier sharding _was recomputed_.

pkg/scheduler/queue/tenant_queues_test.go
```diff
@@ -836,7 +836,9 @@ func (qb *queueBroker) makeQueuePathForTests(tenantID TenantID) QueuePath {
 }
 
 func isConsistent(qb *queueBroker) error {
-	if len(qb.tenantQuerierAssignments.querierIDsSorted) != len(qb.tenantQuerierAssignments.queriersByID) {
+	if len(qb.tenantQuerierAssignments.querierIDsSorted) != len(qb.querierConnections.queriersByID) {
+		fmt.Println(qb.tenantQuerierAssignments.querierIDsSorted)
+		fmt.Println(qb.querierConnections.queriersByID)
```
francoposa, line 841: we probably do not want to leave the prints in, right? If we do we should use a test logger

francoposa (approved):
approved but there are a couple outstanding docstring things :+1:

pkg/scheduler/queue/tenant_queues_test.go
chencs: oops

PR #9295, opened 2024-09-13
Alertmanager: Register grafana state size on sync

titolins:
## https://github.com/grafana/alerting-squad/issues/913
- Add grafana state size so we can implement proper limits

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -183,12 +183,19 @@ func (cfg *MultitenantAlertmanagerConfig) CheckExternalURL(alertmanagerHTTPPrefi
 }
 
 type multitenantAlertmanagerMetrics struct {
+	grafanaStateSize              *prometheus.GaugeVec
 	lastReloadSuccessful          *prometheus.GaugeVec
 	lastReloadSuccessfulTimestamp *prometheus.GaugeVec
 }
 
 func newMultitenantAlertmanagerMetrics(reg prometheus.Registerer) *multitenantAlertmanagerMetrics {
 	m := &multitenantAlertmanagerMetrics{}
+
+	m.grafanaStateSize = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "cortex",
+		Name:      "alertmanager_grafana_state_size_bytes",
+		Help:      "Size of the grafana alertmanager state.",
+	}, []string{"user"})
 
 	m.lastReloadSuccessful = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "cortex",
@@ -796,6 +803,7 @@ func (am *MultitenantAlertmanager) syncStates(ctx context.Context, cfg amConfig)
 		return err
 	}
 	userAM.usingGrafanaState.Store(true)
+	am.multitenantMetrics.grafanaStateSize.WithLabelValues(cfg.User).Set(float64(s.State.Size()))
 
 	// Delete state.
 	if err := am.store.DeleteFullGrafanaState(ctx, cfg.User); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -796,6 +803,7 @@ func (am *MultitenantAlertmanager) syncStates(ctx context.Context, cfg amConfig)
 		return err
 	}
 	userAM.usingGrafanaState.Store(true)
+	am.multitenantMetrics.grafanaStateSize.WithLabelValues(cfg.User).Set(float64(s.State.Size()))
```
santihernandezc, line 806: The problem with this approach is that the size of the Grafana state would only be registered when it's promoted and immediately before it's deleted (check line 809).

#### Alternative 1

We can register the size when a request to save a state comes in. Check [here](https://github.com/grafana/mimir/blob/841f2d1529dc84f5d38aa02880e61914abf8539b/pkg/alertmanager/api_grafana.go#L218), we could do something like:

```go
fullState := alertspb.FullStateDesc{State: protoState}
am.multitenantMetrics.grafanaStateSize.WithLabelValues(userID).Set(float64(fullState.Size()))
err = am.store.SetFullGrafanaState(r.Context(), userID, fullState)
```

We can set the size to 0 if no Grafana state is found during the state sync (check [here](https://github.com/grafana/mimir/blob/841f2d1529dc84f5d38aa02880e61914abf8539b/pkg/alertmanager/multitenant.go#L762-L766)) and if we get a request to delete the state and we succeed (check [here](https://github.com/grafana/mimir/blob/841f2d1529dc84f5d38aa02880e61914abf8539b/pkg/alertmanager/api_grafana.go#L245-L251)).

The problem is that we will always have different values in different instances. Queries to get the size of the latest Grafana state will be tricky.

#### Alternative 2

Call `am.store.GetFullGrafanaState(ctx, cfg.User)` at the top of [syncStates()](https://github.com/grafana/mimir/blob/841f2d1529dc84f5d38aa02880e61914abf8539b/pkg/alertmanager/multitenant.go#L739), use the size of the state to update the metric, or set it to 0 if no state is found.

Problem: Unnecessary DB reads, but IMO it's worth it. This approach is simpler and more accurate than the previous one.

santihernandezc (changes requested):

pkg/alertmanager/multitenant.go
titolins: Makes sense 👍 

Going for the second option then, I've moved the state fetch op to be the first thing on `syncStates`.

> or set it to 0 if no state is found.

I'm just not sure I agree with this though. There's an on-going effort to filter 0 value metrics due to their cost ([slack thread](https://raintank-corp.slack.com/archives/C028MCV4R7C/p1726234793242919)). I do see the cons regarding visualizing / querying the data without 0's, but we can also work around that with promql. Wdyt?

pkg/alertmanager/multitenant.go
```diff
@@ -737,6 +744,18 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 
 // syncStates promotes/unpromotes the Grafana state and updates the 'promoted' flag if needed.
 func (am *MultitenantAlertmanager) syncStates(ctx context.Context, cfg amConfig) error {
+	// fetching grafana state first so we can register its size independently of it being promoted or not
+	s, err := am.store.GetFullGrafanaState(ctx, cfg.User)
+	if err != nil {
+		if errors.Is(err, alertspb.ErrNotFound) {
+			// This is expected if the state was already promoted.
+			level.Debug(am.logger).Log("msg", "grafana state not found, skipping promotion", "user", cfg.User)
```
santihernandezc, line 752: The metric is reporting the size of the last state even after we delete it. until the AM instance is restarted. We could add this line when we get an `ErrNotFound` to fix the issue:

```go
am.multitenantMetrics.grafanaStateSize.DeleteLabelValues(cfg.User)
```

This way, we delete the metric for the user once the state is gone from the DB.

titolins: Ahh, good catch! Fixed now 👍

PR #9314, opened 2024-09-17
blockbuilder: more tests

narqo:
#### What this PR does

~~_This one seats atop https://github.com/grafana/mimir/pull/9199 for now_~~

Here we backport the test cases for how block-builder handles the out-of-order samples.

Also, the PR fixes a flaky `TestBlockBuilder_StartWithLookbackOnNoCommit` test, by making sure the test waits for the correct outcome.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder_test.go
+++ b/pkg/blockbuilder/blockbuilder_test.go
@@ -287,6 +287,121 @@ func TestBlockBuilder_WithMultipleTenants(t *testing.T) {
 	}
 }
 
+func TestBlockBuilder_WithOutOfOrderRecordsAndSamples(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithCancelCause(context.Background())
+	t.Cleanup(func() { cancel(errors.New("test done")) })
+
+	_, kafkaAddr := testkafka.CreateClusterWithoutCustomConsumerGroupsSupport(t, numPartitions, testTopic)
+
+	kafkaClient := mustKafkaClient(t, kafkaAddr)
+	kafkaClient.AddConsumeTopics(testTopic)
+
+	cfg, overrides := blockBuilderConfig(t, kafkaAddr)
+
+	bb, err := New(cfg, log.NewNopLogger(), prometheus.NewPedanticRegistry(), overrides)
+	require.NoError(t, err)
+
+	// We don't want to run the service here. Instead, the test cases below trigger and assert the consumption cycles explicitly.
+	require.NoError(t, bb.starting(ctx))
+	t.Cleanup(func() {
+		require.NoError(t, bb.stopping(nil))
+	})
+
+	const tenantID = "1"
+
+	bucketDir := path.Join(cfg.BlocksStorage.Bucket.Filesystem.Directory, tenantID)
+
+	cycleEndStartup := cycleEndAtStartup(time.Now(), cfg.ConsumeInterval, cfg.ConsumeIntervalBuffer)
+	cycleEnd := cycleEndStartup
+
+	var allSamples []mimirpb.Sample
+
+	// Out of order sample w.r.t. samples in last cycle. But for this cycle,
+	// the TSDB starts fresh. So in terms of the actual block building, it will be
+	// taken as in-order. Depending on if the out-of-order and in-order sample below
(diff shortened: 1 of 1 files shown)


pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -215,14 +208,14 @@ func TestBlockBuilder_StartWithLookbackOnNoCommit(t *testing.T) {
...
-		cortex_blockbuilder_consumer_lag_records{partition="1"} 0
-	`), "cortex_blockbuilder_consumer_lag_records"))
+	require.Eventually(t, func() bool {
+		return assert.NoError(t, promtest.GatherAndCompare(reg, strings.NewReader(`
+			# HELP cortex_blockbuilder_consumer_lag_records The per-topic-partition number of records, instance needs to work through each cycle.
+			# TYPE cortex_blockbuilder_consumer_lag_records gauge
+			cortex_blockbuilder_consumer_lag_records{partition="0"} 0
+			cortex_blockbuilder_consumer_lag_records{partition="1"} 0
+		`), "cortex_blockbuilder_consumer_lag_records"))
+	}, 30*time.Second, 100*time.Millisecond)
```
narqo, line 218: Note: here (and other similar changes), it's fine to wait that long. These assertions are terminal, so the test should not process if the outcome doesn't "eventually" happen.

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -287,6 +280,121 @@ func TestBlockBuilder_WithMultipleTenants(t *testing.T) {
...
+			allSamples[:len(allSamples)-2], // Don't expect the last two sample.
+			nil,
+			labels.MustNewMatcher(labels.MatchRegexp, "foo", ".*"),
+		)
+	})
+
+	t.Run("future record", func(t *testing.T) {
+		// The sample from above which was in-order but the kafka record was in future
+		// should get consumed in this cycle. The other sample that is still in the future should not be consumed.
+		cycleEnd = cycleEnd.Add(cfg.ConsumeInterval)
```
dimitarvdimitrov, line 368: doing this ourselves skips testing the BB logic which does it. Don't we want to test that too? 

Can we let the BB run until it has filled the bucket with some data (i.e. we expect 3 blocks; do something like `Eventually(func() {bucket.countBlocks() == 3})` or until it updates some metrics. Then we can assert on what each block contains with `tsdb.OpenBlock()` instead of `tsdb.Open`

Thoughts?

narqo: I didn't get the point, sorry. Here, what we do is, effectively, move the wall clock forward. It's not that we "skip" any of the block-builder's logic. We only point it to a portion of the partition, where the cycle's data represent what the test case tests.

We would get the same if we start block-builder and let it run for several hours. It would go over all the test data over the course of multiple cycle hours, and tested the data produced on each cycle.

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -287,6 +280,121 @@ func TestBlockBuilder_WithMultipleTenants(t *testing.T) {
 	}
 }
 
+func TestBlockBuilder_WithOutOfOrderRecordsAndSamples(t *testing.T) {
```
pracucci, line 283: [nit] This test assumes ConsumeInterval is 1h, which is the default, but it may change in the future. To keep this test stable, I would suggest to override ConsumeInterval to 1h in the code, and add a comment to explain that's an assumption of the test, regardless what the default value will be in the future.

Should we do the same for ConsumeIntervalBuffer?

pracucci (approved):
Nice test, I like it! 👏 I left a couple of minor comments. No need for me to re-review it.

pkg/blockbuilder/blockbuilder_test.go
dimitarvdimitrov: > We would get the same if we started the block-builder and let it run for several hours

yes, but we'd exercise the production code logic. Right now the logic for calculating `cycleEnd` is in both the tests and the prod code.

but your point about waiting for hours also makes sense. Maybe smaller blocks can solve this? like 2-3-second long blocks with a much smaller ConsumeInterval? A small test is better than no test, so i don't want to block on this

PR #9315, opened 2024-09-17
Register block-builder target in mimir

narqo:
#### What this PR does

~~_This one seats atop https://github.com/grafana/mimir/pull/9199 for now_~~

Here we register the `block-builder` target in mimir. The block-builder is still in an earlier experiment, thus don't expose any documentation or changelog for now.

Also, the PR addresses the discussion we had with @dimitarvdimitrov in https://github.com/grafana/mimir/pull/9199#discussion_r1773600398 about a potential edge-case for a partition that has offset gaps. Here we refactor the internals of the consumer loop replacing the use of remaining lag with the reference to the partition end, at the time of the cycle.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -110,8 +110,8 @@ func (b *BlockBuilder) stopping(_ error) error {
 func (b *BlockBuilder) running(ctx context.Context) error {
 	// Do initial consumption on start using current time as the point up to which we are consuming.
 	// To avoid small blocks at startup, we consume until the <consume interval> boundary + buffer.
-	cycleEnd := cycleEndAtStartup(time.Now(), b.cfg.ConsumeInterval, b.cfg.ConsumeIntervalBuffer)
-	err := b.nextConsumeCycle(ctx, cycleEnd)
+	cycleEndTime := cycleEndAtStartup(time.Now(), b.cfg.ConsumeInterval, b.cfg.ConsumeIntervalBuffer)
+	err := b.nextConsumeCycle(ctx, cycleEndTime)
 	if err != nil {
 		if errors.Is(err, context.Canceled) {
 			return nil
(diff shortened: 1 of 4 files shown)


narqo:
Close in favour of #9315

pkg/blockbuilder/blockbuilder.go
```diff
@@ -289,7 +288,7 @@ func partitionStateFromLag(logger log.Logger, lag kadm.GroupMemberLag, fallbackM
 
 // consumePartition consumes records from the given partition until the cycleEnd timestamp.
 // If the partition is lagging behind, it takes care of consuming it in sections.
-func (b *BlockBuilder) consumePartition(ctx context.Context, partition int32, state partitionState, cycleEnd time.Time) (err error) {
+func (b *BlockBuilder) consumePartition(ctx context.Context, partition int32, state partitionState, cycleEnd time.Time, cycleEndOffset int64) (err error) {
```
pracucci, line 291: [nit] No we have `cycleEndOffset`. WDYT about renaming `cycleEnd` to something like `cycleEndTime` to disambiguate? If you accept this comment, then I would also rename `sectionEnd` accordingly.

```suggestion
func (b *BlockBuilder) consumePartition(ctx context.Context, partition int32, state partitionState, cycleEndTime time.Time, cycleEndOffset int64) (err error) {
```

pracucci (approved):
I'm very glad you got rid of `remaining` offsets. LGTM.

> Also, the PR addresses the discussion we had with @dimitarvdimitrov in https://github.com/grafana/mimir/pull/9199#discussion_r1773600398 about a potential edge-case for a partition that has offset gaps.

Can we unit test it? Or is it too complicated because we can't create gaps in the mocked Kafka cluster?

pkg/mimir/mimir.go
```diff
@@ -123,6 +124,7 @@ type Config struct {
 	Worker           querier_worker.Config           `yaml:"frontend_worker"`
 	Frontend         frontend.CombinedFrontendConfig `yaml:"frontend"`
 	IngestStorage    ingest.Config                   `yaml:"ingest_storage"`
+	BlockBuilder     blockbuilder.Config             `yaml:"block_builder" doc:"hidden"`
```
dimitarvdimitrov, line 127: isn't this part of ingest storage? i mean we use the same kafka config and it's rightly coupled with kafka just like ingest storage

dimitarvdimitrov, line 127: why hidden? Ingest storage is just experimental

pracucci: I think it's fair having a root leve config. Ingest storage is the config shared between component. This config is block-builder specific.

narqo: I don't feel strong about it, but I am not comfortable exposing anything about block-builder to the user-visible helps. Block-builder is even more experimental than ingest-storage.

narqo: +1 to what Marco says. For me, ingest-storage is a package, that services like ingester, query-frontend, ruler, and now block-builder use. I.e. block-builder is not a feature of ingest-storage, it's the opposite 🤷

narqo:
> Can we unit test it? Or is it too complicated because we can't create gaps in the mocked Kafka cluster?

For now, I cannot think of a way to replicate that in a controlled manner, so will need some exploration. I will add it to the rest of the "more tests" PRs later.

PR #9317, opened 2024-09-17
Distributor: Add checkStartedMiddleware

leizor:
#### What this PR does

Add distributor middleware that checks if the underlying distributor service has started and if not, ~~blocks~~ rejects incoming requests until it has.

During normal startup this is not necessary because Mimir will not mark itself as ready until all its services have started.

However, if a distributor pod was already up and marked ready and then the distributor container restarted, k8s may still consider the pod ready before realizing that it is not and the pod may still end up receiving push requests.

In this state the push requests may end up being serviced incorrectly because other distributor middlewares expect their underlying services to already have started but they may not have been.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 * [CHANGE] Ruler: Removed `-ruler.drain-notification-queue-on-shutdown` option, which is now enabled by default. #9115
 * [CHANGE] Querier: allow wrapping errors with context errors only when the former actually correspond to `context.Canceled` and `context.DeadlineExceeded`. #9175
 * [CHANGE] Query-scheduler: Remove the experimental `-query-scheduler.use-multi-algorithm-query-queue` flag. The new multi-algorithm tree queue is always used for the scheduler. #9210
+* [CHANGE] Distributor: Add middleware that blocks until the distributor service has started. #9317
 * [FEATURE] Alertmanager: Added `-alertmanager.log-parsing-label-matchers` to control logging when parsing label matchers. This flag is intended to be used with `-alertmanager.utf8-strict-mode-enabled` to validate UTF-8 strict mode is working as intended. The default value is `false`. #9173
 * [FEATURE] Alertmanager: Added `-alertmanager.utf8-migration-logging-enabled` to enable logging of tenant configurations that are incompatible with UTF-8 strict mode. The default value is `false`. #9174
 * [FEATURE] Querier: add experimental streaming PromQL engine, enabled with `-querier.query-engine=mimir`. #8422 #8430 #8454 #8455 #8360 #8490 #8508 #8577 #8660 #8671 #8677 #8747 #8850 #8872 #8838 #8911 #8909 #8923 #8924 #8925 #8932 #8933 #8934 #8962 #8986 #8993 #8995 #9008 #9017 #9018 #9019 #9120 #9121 #9136 #9139 #9140 #9145 #9191 #9192 #9194 #9196 #9201 #9212 #9225 #9260 #9272 #9277 #9278 #9280
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 * [CHANGE] Ruler: Removed `-ruler.drain-notification-queue-on-shutdown` option, which is now enabled by default. #9115
 * [CHANGE] Querier: allow wrapping errors with context errors only when the former actually correspond to `context.Canceled` and `context.DeadlineExceeded`. #9175
 * [CHANGE] Query-scheduler: Remove the experimental `-query-scheduler.use-multi-algorithm-query-queue` flag. The new multi-algorithm tree queue is always used for the scheduler. #9210
-* [CHANGE] Distributor: Add middleware that blocks until the distributor service has started. #9317
+* [CHANGE] Distributor: Add middleware that rejects incoming requests until the distributor service has started. #9317
 * [FEATURE] Alertmanager: Added `-alertmanager.log-parsing-label-matchers` to control logging when parsing label matchers. This flag is intended to be used with `-alertmanager.utf8-strict-mode-enabled` to validate UTF-8 strict mode is working as intended. The default value is `false`. #9173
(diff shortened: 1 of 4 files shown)


pkg/distributor/distributor.go
```diff
@@ -1290,6 +1296,23 @@ func (d *Distributor) cleanupAfterPushFinished(rs *requestState) {
 	}
 }
 
+// checkStartedMiddleware blocks until the underlying distributor service is started.
+func (d *Distributor) checkStartedMiddleware(next PushFunc) PushFunc {
+	return func(ctx context.Context, pushReq *Request) error {
+		next, maybeCleanup := NextOrCleanup(next, pushReq)
+		defer maybeCleanup()
+
+		if d.isServiceStarted.Load() {
```
flxbk, line 1305: We already have `services.Service.State()` implemented by the distributor struct, would it make sense to use that instead?

pkg/distributor/distributor.go
```diff
@@ -781,7 +786,8 @@ func (d *Distributor) wrapPushWithMiddlewares(next PushFunc) PushFunc {
 	// The middlewares will be applied to the request (!) in the specified order, from first to last.
 	// To guarantee that, middleware functions will be called in reversed order, wrapping the
 	// result from previous call.
-	middlewares = append(middlewares, d.limitsMiddleware) // should run first because it checks limits before other middlewares need to read the request body
+	middlewares = append(middlewares, d.checkStartedMiddleware) // This middleware doesn't read the request body.
```
flxbk, line 789: Is there a reason not to put this middleware after the limits middleware?

pkg/distributor/distributor.go
```diff
@@ -100,6 +100,8 @@ type Distributor struct {
 	ingesterPool  *ring_client.Pool
 	limits        *validation.Overrides
 
+	isServiceStarted *atomic.Bool
```
pstibrany, line 103: We don't need this. We can do `d.State() == services.Running` check instead, similar to how ingesters do it: https://github.com/grafana/mimir/blob/c715b0a90bee797adb8c762e3c27405f44dbefd3/pkg/ingester/ingester.go#L3818-L3824

`d.State()` will switch to `Running` after `starting` method finishes successfully.

pkg/distributor/distributor.go
```diff
@@ -1290,6 +1296,23 @@ func (d *Distributor) cleanupAfterPushFinished(rs *requestState) {
...
+func (d *Distributor) checkStartedMiddleware(next PushFunc) PushFunc {
+	return func(ctx context.Context, pushReq *Request) error {
+		next, maybeCleanup := NextOrCleanup(next, pushReq)
+		defer maybeCleanup()
+
+		if d.isServiceStarted.Load() {
+			return next(ctx, pushReq)
+		}
+
+		if err := d.AwaitRunning(ctx); err != nil {
```
pstibrany, line 1309: Should we block push request, or reject it? By blocking here, we can pile up enough requests in the memory that distributor can already crash.

pkg/distributor/distributor.go
pstibrany: Or *inside* limits middleware? It's just one `if` check.

(Node that even if middleware doesn't read the request body, it doesn't mean that request body isn't already in memory. This is especially true when receiving request via httpgrpc -- entire request is in memory, just not decompressed yet.)

pkg/distributor/distributor.go
leizor: Yep, done in https://github.com/grafana/mimir/pull/9317/commits/aecfd7321693ace25b39a595ba80fb3a62d4fd41

pkg/distributor/distributor.go
leizor: Thanks, done in https://github.com/grafana/mimir/pull/9317/commits/aecfd7321693ace25b39a595ba80fb3a62d4fd41

pkg/distributor/distributor.go
leizor: I thought it made sense in front of the limits middleware because the limits middleware depends on runtime config to load distributor instance limits; runtime config is one of the services that might not have been properly initialized before the distributor begins handling push requests.

Looking more closely at the limits middleware though, I'm thinking we could still put this check after the rate limiting portion but before the request size check. I think it'd also make sense to change to rejecting instead of blocking, heeding your comment below:

> Should we block push request, or reject it? By blocking here, we can pile up enough requests in the memory that distributor can already crash.

pkg/distributor/distributor.go
leizor: Switched to reject in https://github.com/grafana/mimir/pull/9317/commits/010354af2db1e43763209e4c0ce8689a8c129991

pkg/distributor/distributor.go
leizor: I think I was wrong about where we're dependent on runtime config; I've moved the check to the beginning of the limits middleware in https://github.com/grafana/mimir/pull/9317/commits/010354af2db1e43763209e4c0ce8689a8c129991

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 * [CHANGE] Ruler: Removed `-ruler.drain-notification-queue-on-shutdown` option, which is now enabled by default. #9115
 * [CHANGE] Querier: allow wrapping errors with context errors only when the former actually correspond to `context.Canceled` and `context.DeadlineExceeded`. #9175
 * [CHANGE] Query-scheduler: Remove the experimental `-query-scheduler.use-multi-algorithm-query-queue` flag. The new multi-algorithm tree queue is always used for the scheduler. #9210
+* [CHANGE] Distributor: Add middleware that rejects incoming requests until the distributor service has started. #9317
```
flxbk, line 34: This is now not quite accurate anymore.

pstibrany, line 34:
```suggestion
* [CHANGE] Distributor: reject incoming requests until the distributor service has started. #9317
```

pstibrany (approved):
Thank you.

PR #9319, opened 2024-09-17
query-tee: Add an option to shift the query times for comparison

codesome:
#### What this PR does

Adds an option `proxy.shift-comparison-queries-by` to shift the query times for query-tee comparisons.

Sometimes `proxy.compare-skip-recent-samples` option may not be enough because if you want to set it to a higher duration (say few hours), it will completely skip comparing the queries for recent times, which in practice forms a big % of all queries (like recording rules, many most instant queries, etc).

With this change, the query-tee still does the query with original times for the preferred backend, but for comparison, it shifts the query times, and does another query to the preferred backend with the shifted times and only one shifted query to the secondary backend and uses those results for comparison.

Added `proxy.shift-comparison-sampling-ratio` to samples the queries for shifting the query times.

Draft PR because there are still some TODOs and need to write tests.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tools/querytee/proxy_endpoint.go
```diff
@@ -246,52 +244,140 @@ func (p *ProxyEndpoint) executeBackendRequests(req *http.Request, backends []Pro
...
-		p.metrics.proportionalDuration.WithLabelValues(p.route.RouteName).Observe(proportionalDurationDifference)
-		p.metrics.responsesComparedTotal.WithLabelValues(p.route.RouteName, string(result)).Inc()
+// shiftQueryRequest shifts the query times of the request for instant and range queries.
+// If there was any error, the error is just logged and a nil request is returned.
+func (p *ProxyEndpoint) shiftQueryRequest(req *http.Request, d time.Duration) (shiftedRequest *http.Request) {
+	if !querymiddleware.IsRangeQuery(req.URL.Path) && !querymiddleware.IsInstantQuery(req.URL.Path) {
+		return
+	}
+
+	codec := querymiddleware.NewPrometheusCodec(nil, 5*time.Minute, "protobuf")
```
dimitarvdimitrov, line 362: i think the comparator only works on json responses https://github.com/grafana/mimir/blob/14beb568f0093c6dc9bf12678b367f0dd14e378e/tools/querytee/response_comparator.go#L65-L76

codesome:
Added a unit test

codesome:
Tried it in a dev environment, weird that all shifted queries are being skipped for comparison because the response format is text/plain and not application/json

Screenshot 2024-09-20 at 10 14 17 PM

codesome:
I converted this to draft since there is the above bug that I need to fix.

codesome:
It is working locally with the docker-compose setup, so I have no clue why it gave issues in the dev environment. I will try running it again in dev next week and try to debug it there.

PR #9343, opened 2024-09-20
MQE: Add support for group aggregation 

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


charleskorn (approved):
LGTM. It'd be good to see some benchmarks for `group` and `count`, but this shouldn't block this PR - we can always come back to it after merging.

PR #9344, opened 2024-09-20
kafka replay speed: adjust batchingQueueCapacity

dimitarvdimitrov:
#### What this PR does

I made 2000 up when we were flushing individual series to the channel. Then 2000 might have made sense, but when flushing whole WriteRequests a capacity of 1 should be sufficient.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -469,7 +469,7 @@ type batchingQueue struct {
 func newBatchingQueue(capacity int, batchSize int) *batchingQueue {
 	return &batchingQueue{
 		ch:           make(chan flushableWriteRequest, capacity),
-		errCh:        make(chan error, capacity),
+		errCh:        make(chan error, capacity+1),
 		done:         make(chan struct{}),
 		currentBatch: flushableWriteRequest{WriteRequest: &mimirpb.WriteRequest{Timeseries: mimirpb.PreallocTimeseriesSliceFromPool()}},
 		batchSize:    batchSize,
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -31,7 +31,8 @@ import (
 // If we're still flushing batch `N`, batch `N+1` is enqueued for flushing, and we've just finished batching batch `N+2`,
 // then we're going too fast - in the time to flush `N` we've created `N+2` and `N+1`.
 // A buffer of 1 doesn't let us go too far ahead with batching, but still doesn't require synchronisation between batching and flushing.
-const batchingQueueCapacity = 1
+// If there's high variability in the time to flush or in the time to batch, then this buffer might need to be increased.
+const batchingQueueCapacity = 5
 
 type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -24,7 +24,14 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/storage/ingest/pusher.go
```diff
@@ -24,7 +24,14 @@ import (
...
 )
 
-const shardForSeriesBuffer = 2000 // TODO dimitarvdimitrov 2000 is arbitrary; the idea is that we don't block the goroutine calling PushToStorage while we're flushing. A linked list with a sync.Cond or something different would also work
+// batchingQueueCapacity controls how many batches can be enqueued for flushing.
+// We want to batch and flush in parallel. There's no point in batching faster than we can flush.
+// A buffer of 1 would allow to balance the time for batching and flushing 50/50.
+// For example, if we flush 1 batch/sec, then batching 2 batches/sec doesn't make us faster.
+// If we're still flushing batch `N`, batch `N+1` is enqueued for flushing, and we've just finished batching batch `N+2`,
+// then we're going too fast - in the time to flush `N` we've created `N+2` and `N+1`.
+// A buffer of 1 doesn't let us go too far ahead with batching, but still doesn't require synchronisation between batching and flushing.
```
gotjosh, line 33:
```suggestion
// batchingQueueCapacity controls how many batches can be enqueued for flushing.
// We don't want to push any batches in parallel and instead want to prepare the next one while the current one finishes, hence the buffer of 1.
// For example, if we flush 1 batch/sec, then batching 2 batches/sec doesn't make us faster.
// This is our initial assumption, and there's potential in testing with higher numbers if there's a high variability in flush times - assuming we can preserve the order of the batches. For now, we'll stick to 1.
```

gotjosh: nit: I don't think we need that many words convey meaning - in this case, I feel like less is more. I'm not fussed and you can pick (or not) any of my words apart.

pkg/storage/ingest/pusher_test.go
```diff
@@ -637,6 +638,42 @@ func TestParallelStorageShards_ShardWriteRequest(t *testing.T) {
...
+	}
+
+	// Close the queue to signal no more items will be added
+	err := queue.Close()
+	require.ErrorContains(t, err, "mock error")
+
+	wg.Wait()
+
+	// Ensure the queue is empty and no deadlock occurred
+	require.Len(t, queue.ch, 0)
```
gotjosh, line 674: We should probably ensure that all the errors were consumed, this is covered in other test but wouldn't hurt to add it here.

gotjosh (approved):
LGTM - my comments are nits.

pkg/storage/ingest/pusher.go
```diff
@@ -24,7 +24,12 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
-const shardForSeriesBuffer = 2000 // TODO dimitarvdimitrov 2000 is arbitrary; the idea is that we don't block the goroutine calling PushToStorage while we're flushing. A linked list with a sync.Cond or something different would also work
+// batchingQueueCapacity controls how many batches can be enqueued for flushing.
+// We don't want to push any batches in parallel and instead want to prepare the next one while the current one finishes, hence the buffer of 1.
+// For example, if we flush 1 batch/sec, then batching 2 batches/sec doesn't make us faster.
+// This is our initial assumption, and there's potential in testing with higher numbers if there's a high variability in flush times - assuming we can preserve the order of the batches. For now, we'll stick to 1.
```
dimitarvdimitrov, line 30:
```suggestion
// This is our initial assumption, and there's potential in testing with higher numbers if there's a high variability in flush times - assuming we can preserve the order of the batches. For now, we'll stick to 5.
```

PR #9349, opened 2024-09-20
kafka replay speed: move error handling closer to actual ingestion

dimitarvdimitrov:
#### What this PR does
Previously, we'd let error bubble-up and only take decisions on whether to abort the request or not at the very top (`pusherConsumer`). This meant that we'd potentially buffer more requests before we detect an error.

This change extracts error handling logic into a `Pusher` implementation: `clientErrorFilteringPusher`. This implementation logs client errors and then swallows them. We inject that implementation in front of the ingester. This means that the parallel storage implementation can abort ASAP instead of collecting and bubbling up the errors.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

gotjosh (approved):
LGTM

Issue #9052, opened 2024-08-19
Flaky TestIngester_PushToStorage_CircuitBreaker


narqo:
The `TestIngester_PushToStorage_CircuitBreaker` test [failed on the CI][1] (for an unrelated [PR](https://github.com/grafana/mimir/pull/9050)).

@duricanikolic could you have a look if that's something obvious.

```
--- FAIL: TestIngester_PushToStorage_CircuitBreaker (4.01s)
    --- FAIL: TestIngester_PushToStorage_CircuitBreaker/deadline_exceeded_with_initial_delay_enabled (2.00s)
        logger.go:25: 2024-08-19 17:28:50.993596764 +0000 UTC m=+6.589185212 level warn msg -blocks-storage.backend=filesystem is for development and testing only; you should switch to an external object store for production use or use a shared filesystem
        logger.go:25: 2024-08-19 17:28:50.996329099 +0000 UTC m=+6.591917547 level info msg TSDB idle compaction timeout set timeout 1h1m15.660357536s
        logger.go:25: 2024-08-19 17:28:50.999902208 +0000 UTC m=+6.595490657 level info msg opening existing TSDBs
        logger.go:25: 2024-08-19 17:28:51.003382648 +0000 UTC m=+6.598971096 level info component ingest_reader partition 0 msg starting consumption from partition start because no offset has been found start_offset -2 consumer_group ingester-zone-a-0
        logger.go:25: 2024-08-19 17:28:51.005191734 +0000 UTC m=+6.600780183 level info component ingest_reader partition 0 component kafka_client msg immediate metadata update triggered why querying metadata for consumer initialization
        logger.go:25: 2024-08-19 17:28:51.005773777 +0000 UTC m=+6.601362226 level info component ingest_reader partition 0 target_lag 2s max_lag 15s msg partition reader is starting to consume partition until target and max consumer lag is honored
        logger.go:25: 2024-08-19 17:28:51.006917386 +0000 UTC m=+6.602505834 level info component ingest_reader partition 0 component kafka_client msg assigning partitions why new assignments from direct consumer how assigning everything new, keeping current assignment input mimir[0{-2 e-1 ce0}]
        logger.go:25: 2024-08-19 17:28:51.008982868 +0000 UTC m=+6.604571316 level info component ingest_reader partition 0 target_lag 2s max_lag 15s msg partition reader found no records to consume because partition is empty partition_start_offset 0 last_produced_offset -1
[...]

PR #9351, opened 2024-09-20
TestIngester_PushToStorage_CircuitBreaker: increase initial delay

dimitarvdimitrov:
#### What this PR does

I believe there's a race between sending the first request and then collecting the metrics. It's possible that we collect the metrics longer than 200ms after the first request, at which point the CB has opened. I could reproduce https://github.com/grafana/mimir/issues/9052 by reducing the initialDelay to 10ms.

This PR increases it to 1 hour so that we're more sure that the delay hasn't expired when we're collecting the metrics.

Fixes https://github.com/grafana/mimir/issues/9052

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/circuitbreaker_test.go
+++ b/pkg/ingester/circuitbreaker_test.go
@@ -455,7 +455,7 @@ func TestIngester_PushToStorage_CircuitBreaker(t *testing.T) {
 				failureThreshold := 2
 				var initialDelay time.Duration
 				if initialDelayEnabled {
-					initialDelay = 200 * time.Millisecond
+					initialDelay = time.Hour
 				}
 				cfg.PushCircuitBreaker = CircuitBreakerConfig{
 					Enabled:                    true,
--- a/pkg/ingester/circuitbreaker_test.go
+++ b/pkg/ingester/circuitbreaker_test.go
@@ -398,7 +398,7 @@ func TestCircuitBreaker_FinishRequest(t *testing.T) {
 	}
 }
 
-func TestIngester_PushToStorage_CircuitBreaker(t *testing.T) {
+func TestIngester_IngestStorage_PushToStorage_CircuitBreaker(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/ingester/circuitbreaker_test.go
```diff
@@ -455,7 +455,7 @@ func TestIngester_PushToStorage_CircuitBreaker(t *testing.T) {
 				failureThreshold := 2
 				var initialDelay time.Duration
 				if initialDelayEnabled {
-					initialDelay = 200 * time.Millisecond
+					initialDelay = time.Hour
 				}
```
narqo, line 459: There is also `TestIngester_Push_CircuitBreaker_DeadlineExceeded` down there, which uses the same test pattern, where `initialDelay = 200 * time.Millisecond`. Should we, maybe, update it also, to keep the tests consistent. Otherwise, it makes it even more complicated to follow what those tests are actually testing.

narqo (approved):
Works for me. Although, I still don't fully understand what does the case of `initialDelayEnabled` tests.

Note also the comment about another one to update.

dimitarvdimitrov:
> Although, I still don't fully understand what does the case of `initialDelayEnabled` tests.

I think it basically checks that the CB doesn't open for the first N seconds after a failure. If I was writing it, I'd probably then assert that the CB _does_ open eventually; right now we just assert it stays closed

pkg/ingester/circuitbreaker_test.go
dimitarvdimitrov: yeah probably. I think one uses ingest storage, the other one doesn't

dimitarvdimitrov: pushed this https://github.com/grafana/mimir/pull/9351/commits/a299bcd6857573a73580285d071ba72a4b42604c

PR #9353, opened 2024-09-20
Alerting: Filter am zero value gauges

titolins:
## Dependencies
- https://github.com/grafana/dskit/pull/584

*NOTE*: Creating as draft as I'm not sure the dskit proposal is the right way to go


titolins:
doing it all in [here](https://github.com/grafana/mimir/pull/9359) since dskit changes were already merged

PR #9359, opened 2024-09-20
Alertmanager: Skip exposing some more metrics when they have zero value

titolins:
Filter 0 value tenant metrics as they are unnecessary

## Related PR's
- adapt dskit so we can filter gauges as well: https://github.com/grafana/dskit/pull/584


stevesg:
Nice one, we should put an entry in the CHANGELOG.md.

stevesg (approved):
Looks ready to me, good work on debugging the integration test.

pstibrany (approved):
🙈

tacole02 (approved):

PR #9362, opened 2024-09-22
Move the e2e histograms code to somewhere where it can be exported

zenador:
#### What this PR does

Minor refactor of integration/util.go to separate concerns more cleanly, moving the e2e histograms code to its own package, which will allow us to export it without exporting a lot of other unnecessary code.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

None as these are minor, non-user-facing changes.

### DIFF
--- a/integration/e2ehistograms/e2ehistograms.go
+++ b/integration/e2ehistograms/e2ehistograms.go
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: AGPL-3.0-only
-//go:build requires_docker
 
 package e2ehistograms
 
--- a/integration/e2ehistograms/e2ehistograms.go
+++ b/integration/e2ehistograms/e2ehistograms.go
@@ -1,0 +1,142 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//go:build requires_docker
+
+package e2ehistograms
+
+import (
+	"math/rand"
+	"time"
+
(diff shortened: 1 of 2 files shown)


integration/e2ehistograms/e2ehistograms.go
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//go:build requires_docker
```
narqo, line 2: Does this package still need this `go:build` directive?

narqo (approved):
(nit) On nitpicking side of things, maybe we can name the new package `e2eseries`. I don't think it's worth it to have a dedicated E2E package of a single *.go file per-series type, should we ever add more such helpers.

Works for me otherwise 🔥

integration/e2ehistograms/e2ehistograms.go
zenador: Good point, removed.

zenador:
Keeping the name for now, because it's quite unlikely that we'll have a new series type anytime soon, and we can easily rename at that point if necessary. Also we keep the e2e generators for floats somewhere else (e2e repo), and we have the histogram generators in this mimir repo because the e2e repo seems to be in use by non-metric products that don't have histograms, so `e2eseries` might be confusing as it doesn't include the float generators.

PR #9363, opened 2024-09-22
Generalise some of Mimir's query sharding code to be more reusable

zenador:
#### What this PR does

Generalise some of Mimir's query sharding code to be more reusable for other kinds of sharded queries.

Follow up to https://github.com/grafana/mimir/pull/8755

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

No docs as these are minor, non-user-facing changes.

### DIFF
--- a/pkg/frontend/querymiddleware/codec.go
+++ b/pkg/frontend/querymiddleware/codec.go
@@ -49,7 +49,9 @@ var (
 	allFormats        = []string{formatJSON, formatProtobuf}
 
 	// List of HTTP headers to propagate when a Prometheus request is encoded into a HTTP request.
-	prometheusCodecPropagateHeaders = []string{compat.ForceFallbackHeaderName, chunkinfologger.ChunkInfoLoggingHeader, api.ReadConsistencyOffsetsHeader}
+	clusterNameHeader                     = "X-Cluster-Name"
+	prometheusCodecPropagateHeaders       = []string{compat.ForceFallbackHeaderName, chunkinfologger.ChunkInfoLoggingHeader, api.ReadConsistencyOffsetsHeader, clusterNameHeader}
+	prometheusCodecLabelsPropagateHeaders = []string{clusterNameHeader}
 )
 
 const (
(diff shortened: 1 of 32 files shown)


pkg/frontend/querymiddleware/codec.go
```diff
@@ -755,6 +817,90 @@ func (c prometheusCodec) DecodeResponse(ctx context.Context, r *http.Response, _
 	return resp, nil
 }
 
+func (c prometheusCodec) DecodeLabelsQueryResponse(ctx context.Context, r *http.Response, lr LabelsQueryRequest, logger log.Logger) (Response, error) {
```
charleskorn, line 820: Should this be split in half, one half for labels and another for series responses?

Or is there a reason to keep them in one method like this?

(same applies for `EncodeLabelsQueryResponse`)

pkg/frontend/querymiddleware/querysharding.go
```diff
@@ -496,7 +508,7 @@ type annotationAccumulator struct {
 	infos    *sync.Map
 }
 
-func newAnnotationAccumulator() *annotationAccumulator {
+func NewAnnotationAccumulator() *annotationAccumulator { //nolint:revive
```
charleskorn, line 511: Should we just export `annotationAccumulator`? Seems weird to export `NewAnnotationAccumulator` but not the type it returns.

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -136,6 +137,19 @@ type MetricsQueryHandler interface {
...
 
+// LabelsHandlerFunc is like http.HandlerFunc, but for LabelsQueryHandler.
+type LabelsHandlerFunc func(context.Context, LabelsQueryRequest) (Response, error)
+
+// Do implements LabelsQueryHandler.
+func (q LabelsHandlerFunc) Do(ctx context.Context, req LabelsQueryRequest) (Response, error) {
+	return q(ctx, req)
+}
+
+// LabelsQueryHandler is like http.Handle, but specifically for Prometheus label names and values calls.
```
charleskorn, line 148: Typo?

```suggestion
// LabelsQueryHandler is like http.Handler, but specifically for Prometheus label names and values calls.
```

zenador: It was copied from MetricsQueryHandler, but it does seem to be a typo so I updated both,

pkg/frontend/querymiddleware/codec.go
zenador: I kept it as one because the LabelsQueryRequest is used for label names, label values and series because their request formats are all fairly similar and this function outputs a Response which works for all three, and I don't want to clutter this with more functions that do almost the same thing, and only differentiate between the labels and series in the formatter interface (since the response formats are different).

charleskorn (approved):

dimitarvdimitrov (commented):
i was catching up with the code written so far for this project and left a few comments. Do what you will with them 😄

PR #9366, opened 2024-09-23
query-tee: add equivalent errors for string expression for range queries

charleskorn:
#### What this PR does

This PR builds on https://github.com/grafana/mimir/pull/9143 to add another class of equivalent errors.

https://github.com/grafana/mimir/pull/9143

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


jhesketh (commented):
Could we also get this added to the list of tests?

charleskorn:
> Could we also get this added to the list of tests?

My intention is that we'll add more and more of these equivalence classes over time, most of which will be simple patterns like these. I don't think it's worth adding more test cases for these, provided the general functionality of error equivalence is tested, but open to other opinions.

jhesketh (approved):
Okay, I'm not blocking on it, but maybe it's worth having a more generic test that loads each errorEquivalenceClasses pair, extracts the matching string and checks if it outputs the correct result (if that makes sense)

charleskorn:
> Okay, I'm not blocking on it, but maybe it's worth having a more generic test that loads each errorEquivalenceClasses pair, extracts the matching string and checks if it outputs the correct result (if that makes sense)

This would be difficult to do as each pattern is a regex, not a string (although some of the patterns are just simple matches).

PR #9367, opened 2024-09-23
MQE: Add support for stddev

jhesketh:
With the upstream patch the edge case tester all passes.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/testdata/upstream/aggregators.test
```diff
@@ -199,14 +199,12 @@ eval instant at 50m ceil(0.004 * http_requests{group="production",job="api-serve
...
+eval instant at 50m stddev(http_requests)
+  {} 229.12878474779
 
-# Unsupported by streaming engine.
-# eval instant at 50m stddev by (instance)(http_requests)
-#   {instance="0"} 223.60679774998
-#   {instance="1"} 223.60679774998
+eval instant at 50m stddev by (instance)(http_requests)
+  {instance="0"} 223.60679774998
+  {instance="1"} 223.60679774998
```
charleskorn, line 207: Could you add a test case of our own that has multiple groups that have different output values?

It'd also be good to see a test with NaN and infinite input values.

pkg/streamingpromql/engine_test.go
```diff
@@ -1786,6 +1786,10 @@ func TestCompareVariousMixedMetrics(t *testing.T) {
 	for _, labels := range labelCombinations {
 		labelRegex := strings.Join(labels, "|")
 		// Aggregations
+		// TODO(jhesketh): Add stddev back ing.
+		// stddev is excluded until https://github.com/prometheus/prometheus/pull/14941 is merged
+		// fixing an inconsistency in the promql engine where if a native histogram is the first sample
+		// loaded, it is incorrectly treaded as a 0 float point.
```
charleskorn, line 1792:
```suggestion
		// TODO(jhesketh): Add stddev back in.
		// stddev is excluded until https://github.com/prometheus/prometheus/pull/14941 is merged
		// fixing an inconsistency in Prometheus' engine where if a native histogram is the first sample
		// loaded, it is incorrectly treated as a 0 float point.
```

pkg/streamingpromql/aggregations/stddev.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type StddevAggregationGroup struct {
+	floats     []float64
+	floatMeans []float64
+
+	// Keeps track of how many series we have encountered thus far for the group at this point
+	// This is necessary to do per point (instead of just counting the groups) as a series may have
+	// stale or non-existent values that are not added towards the count.
+	// We use float64 instead of uint64 so that we can reuse the float pool and don't have to retype on each division.
```
charleskorn, line 25: [nit] This might make this a little clearer (same applies for comment in `avg.go` above):

```suggestion
	// Keeps track of how many samples we have encountered thus far for the group at this point
	// This is necessary to do per point (instead of just counting the input series) as a series may have
	// stale or non-existent values that are not added towards the count.
	// We use float64 instead of uint64 so that we can reuse the float pool and don't have to retype on each division.
```

pkg/streamingpromql/engine_test.go
jhesketh: I don't own a cat. I've had coffee. I don't know why I type bad and such.

pkg/streamingpromql/testdata/upstream/aggregators.test
jhesketh: Fair point to call out cases for infinity. The stddev of anything with infinity should be infinity. However, Prometheus returns NaN (and so do we).

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -327,3 +327,18 @@ eval range from 0m to 1m step 1m count(series)
 
 eval range from 0m to 1m step 1m group(series)
   {} _ 1
+
+clear
+
+load 1m
```
charleskorn, line 333: Should we make this 6m instead to avoid lookbacks? 

The way the test data is constructed below, it looks like the goal is to test scenarios where there is only one input sample for a group (eg. the 5th step for group A).

Or is the intention to exercise lookbacks?

jhesketh: The intention was to exercise, but it could be worth doing for only one input series. I'll add another group with single points for that.

pkg/streamingpromql/testdata/upstream/aggregators.test
charleskorn: >The stddev of anything with infinity should be infinity. However, Prometheus returns NaN (and so do we).

That seems wrong - is it worth raising an issue with Prometheus to confirm the expected behaviour here?

jhesketh: For now I've extended the change in Prometheus to handle infinities consistently. With that change, and MQE, any stddev containing infinity is NaN (or undefined). https://github.com/prometheus/prometheus/pull/14941

pkg/streamingpromql/testdata/ours/aggregators.test
jhesketh: I added this check upstream that we'll eventually get. If I were to add this here I would have to comment it out until upstream is merged anyway.

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -338,7 +338,16 @@ load 1m
...
+# https://github.com/prometheus/prometheus/pull/14941 is merged.
+#	series{label="g", group="d"} NaN 3 0 9 NaN 10
+#	series{label="h", group="e"} -inf 3 0 9 inf 10
+
 eval range from 0 to 6m step 1m stddev by (group) (series)
   {group="a"} 4 3 2 7 7.5 26 26
   {group="b"} 0 NaN NaN 18 NaN 25 25
   {group="c"} NaN 0.5 1.5 2.5 NaN 30 30
+#  {group="d"} NaN 0 0 0 NaN 0 0
+#  {group="e"} NaN 0 0 0 NaN 0 0
```
charleskorn: Could we move this into a test in the `ours-only` directory?

charleskorn (approved):
🚀

PR #9371, opened 2024-09-23
MQE: fix `rate()` over native histograms where first point in range is a counter reset

charleskorn:
#### What this PR does

This PR fixes an issue in the implementation of `rate()` in MQE where it would calculate incorrect results if the first point in the range was marked as a counter reset.

`h1.DetectReset(h2)` (unsurprisingly) returns `true` if `h2` is marked as a counter reset, even if `h1` is `h2`.

If the first point in the range was marked as a counter reset, incorrect results would be generated in `histogramRate` because `accumulate()` was called including the first point in the range. `DetectReset()` would return true for the first point and so the first point would be added to `delta`.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/testdata/ours/native_histograms.test
```diff
@@ -147,6 +147,15 @@ eval range from 0 to 4m step 1m sum by (env) (rate(incr_histogram[5m]))
 
 clear
 
+# Test rate() with explicit counter reset information.
+load 1m
+  metric {{sum:3 count:4 buckets:[1 2 1] counter_reset_hint:reset}} {{sum:63 count:124 buckets:[100 4 20] counter_reset_hint:not_reset}}
```
jhesketh, line 152: can we also write a test that triggers this without a hint?

jhesketh (approved):

pkg/streamingpromql/testdata/ours/native_histograms.test
charleskorn: When you say "triggers this", are you referring to the first point being included incorrectly? If so, then I believe the answer is no: the only way `h1.DetectReset(h1)` can return true is if `h1.CounterResetHint` is set to `CounterReset`.

charleskorn:
Given the approval, I'm going to merge this - I can always come back and add further tests based on the discussion above if need be.

PR #9381, opened 2024-09-23
Add MimirKubernetesServiceEndpointsApproachingLimit and MimirKubernetesServiceEndpointsOverLimit alerts

pracucci:
#### What this PR does

Last week I learned that K8S service endpoints is limited. Kubernetes default is 100, but can be increased up to 1000. This is an issue for K8S services used by Mimir for service discovery. In this PR I propose to add a couple of alerts firing when is approaching, or has reached, such limit. The alert checks only services used by Mimir for service discovery, given for others it's not expected to be a real issue.

There's no Prometheus metric exposing the limit, so I added a mixin config option for that. The mixin default is 1000,  Kubernetes default is 100 and Kubernetes max is 1000. Why did I pick 1000 as mixin default? To try to avoid having such limit firing for people running Mimir at large scale as soon as they updated the alerts (assuming the already raised the Kubernetes limit to the max). I'm open to feedback tho.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,8 @@
 ### Mixin
 
 * [CHANGE] Dashboards: set default auto-refresh rate to 5m. #8758
+* [FEATURE] Alerts: add `MimirGossipMembersEndpointsOutOfSync` alert. #9347
+* [FEATURE] Alerts: add `MimirKubernetesServiceEndpointsApproachingLimit` and `MimirKubernetesServiceEndpointsOverLimit` alerts. Please configure `_config.kubernetes_endpoints_limit` in the mixin to the actual endpoints limit configured in your Kubernetes cluster. The mixin default is 1000, which is the upperbound configurabled in Kubernetes. #9381
 * [ENHANCEMENT] Dashboards: allow switching between using classic or native histograms in dashboards.
   * Overview dashboard: status, read/write latency and queries/ingestion per sec panels, `cortex_request_duration_seconds` metric. #7674 #8502 #8791
   * Writes dashboard: `cortex_request_duration_seconds` metric. #8757 #8791
@@ -146,7 +148,6 @@
 * [ENHANCEMENT] Dashboards: add 'Read path' selector to 'Mimir / Queries' dashboard. #8878
 * [ENHANCEMENT] Dashboards: add annotation indicating active series are being reloaded to 'Mimir / Tenants' dashboard. #9257
 * [ENHANCEMENT] Dashboards: limit results on the 'Failed evaluations rate' panel of the 'Mimir / Tenants' dashboard to 50 to avoid crashing the page when there are many failing groups. #9262
-* [FEATURE] Alerts: add `MimirGossipMembersEndpointsOutOfSync` alert. #9347
 * [BUGFIX] Dashboards: fix "current replicas" in autoscaling panels when HPA is not active. #8566
 * [BUGFIX] Alerts: do not fire `MimirRingMembersMismatch` during the migration to experimental ingest storage. #8727
 * [BUGFIX] Dashboards: avoid over-counting of ingesters metrics when migrating to experimental ingest storage. #9170
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -970,6 +970,30 @@ How to **investigate**:
        $.apps.v1.statefulSet.spec.template.metadata.withLabelsMixin({ [$._config.gossip_member_label]: 'false' }),
      ```
 
+
+### MimirKubernetesServiceEndpointsApproachingLimit
+
+This alert fires if a Kubernetes service used by Mimir for service discovery is approaching the endpoints limit.
+
+See [MimirKubernetesServiceEndpointsOverLimit](#MimirKubernetesServiceEndpointsOverLimit).
+
+
+### MimirKubernetesServiceEndpointsOverLimit
+
+This alert fires if a Kubernetes service used by Mimir for service discovery has reached the endpoints limit.
+
+How it **works**:
+
+- By default, the Kubernetes control plane creates and manages service endpoints to have no more than 100 endpoints each ([doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)). You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
+- Once a Kubernetes service endpoints list reached the limit, the endpoints are truncated to the limit.
+- Mimir uses some Kubernetes services for DNS-based service discovery. If the endpoints of one of such services is getting truncated, Mimir is unable to discover all pods.
+
+How to **fix** it:
+
+- Increase Kubernetes `--max-endpoints-per-slice kube-controller-manager` to 1000, if it's currently running with a lower value. Configure the limit in the alert to reflect the new setting.
+- Reduce the number of replicas of the affected service and vertically scale it up.
+
+
 ### EtcdAllocatingTooMuchMemory
 
 This can be triggered if there are too many HA dedupe keys in etcd. We saw this when one of our clusters hit 20K tenants that were using HA dedupe config. Raise the etcd limits via:
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -970,13 +970,11 @@ How to **investigate**:
        $.apps.v1.statefulSet.spec.template.metadata.withLabelsMixin({ [$._config.gossip_member_label]: 'false' }),
      ```
 
-
 ### MimirKubernetesServiceEndpointsApproachingLimit
 
 This alert fires if a Kubernetes service used by Mimir for service discovery is approaching the endpoints limit.
 
 See [MimirKubernetesServiceEndpointsOverLimit](#MimirKubernetesServiceEndpointsOverLimit).
-
 
 ### MimirKubernetesServiceEndpointsOverLimit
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -120,6 +120,8 @@
 ### Mixin
 
 * [CHANGE] Dashboards: set default auto-refresh rate to 5m. #8758
+* [FEATURE] Alerts: add `MimirGossipMembersEndpointsOutOfSync` alert. #9347
+* [FEATURE] Alerts: add `MimirKubernetesServiceEndpointsApproachingLimit` and `MimirKubernetesServiceEndpointsOverLimit` alerts. Please configure `_config.kubernetes_endpoints_limit` in the mixin to the actual endpoints limit configured in your Kubernetes cluster. The mixin default is 1000, which is the upperbound configurabled in Kubernetes. #9381
```
pstibrany, line 124: Is 1000 the default in Kubernetes?

Typo in "configurabled".

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+
+See [MimirKubernetesServiceEndpointsOverLimit](#MimirKubernetesServiceEndpointsOverLimit).
+
+### MimirKubernetesServiceEndpointsOverLimit
+
+This alert fires if a Kubernetes service used by Mimir for service discovery has reached the endpoints limit.
+
+How it **works**:
+
+- By default, the Kubernetes control plane creates and manages service endpoints to have no more than 100 endpoints each ([doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)). You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
```
pstibrany, line 985:
```suggestion
- By default, the `EndpointSlice` objects created and managed by Kubernetes control plane cannot have more than 100 endpoints ([doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)). You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+- Once a Kubernetes service endpoints list reached the limit, the endpoints are truncated to the limit.
+- Mimir uses some Kubernetes services for DNS-based service discovery. If the endpoints of one of such services is getting truncated, Mimir is unable to discover all pods.
```
pstibrany, line 987:
```suggestion
- Mimir uses some Kubernetes services for DNS-based service discovery. If the endpoints list of one of such services is getting truncated, Mimir is unable to discover all pods.
```

pstibrany (approved):

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+See [MimirKubernetesServiceEndpointsOverLimit](#MimirKubernetesServiceEndpointsOverLimit).
+
+### MimirKubernetesServiceEndpointsOverLimit
+
+This alert fires if a Kubernetes service used by Mimir for service discovery has reached the endpoints limit.
+
+How it **works**:
+
+- By default, the Kubernetes control plane creates and manages service endpoints to have no more than 100 endpoints each ([doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)). You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
+- Once a Kubernetes service endpoints list reached the limit, the endpoints are truncated to the limit.
```
narqo, line 986: Is this how `EndpointSlices` work really? The way I thought about the endpoint-slices since reading the [announcement][1] was that

1. the control plane creates endpoint-slices per service
2. (by default) one endpoint-slice can include up to 100 endpoints (pointers to a pod)
3. when the limit is reached, the control plane creates additional endpoint-slices
4. since then the expectation was that a service can scale to beyond 100K endpoints.

Refer to the original "[Scaling Kubernetes Networking With EndpointSlices][1]" post.

Is this the problem we've encountered somewhere?

[1]: https://kubernetes.io/blog/2020/09/02/scaling-kubernetes-networking-with-endpointslices/

pracucci: The problem is that the DNS only returns the first 1K, at least that's what we observed. So when you do DNS-based service discovery, you're limited to the first 1K.

> Is this the problem we've encountered somewhere?

It's related to another issue we've observed. I can share more details offline about it.

pstibrany: Does it make sense to check if individual endpoint slice has reached 100 limit, if DNS entry can return endpoints from more than 1 slice? Perhaps this alerts needs more investigation before merging.

CHANGELOG.md
tacole02, line 124:
```suggestion
* [FEATURE] Alerts: add `MimirKubernetesServiceEndpointsApproachingLimit` and `MimirKubernetesServiceEndpointsOverLimit` alerts. Configure `_config.kubernetes_endpoints_limit` in the mixin to the actual endpoints limit configured in your Kubernetes cluster. The mixin default is 1000, which is the upperbound configurabled in Kubernetes. #9381
```

tacole02: We try to avoid "please".

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+
+This alert fires if a Kubernetes service used by Mimir for service discovery is approaching the endpoints limit.
+
+See [MimirKubernetesServiceEndpointsOverLimit](#MimirKubernetesServiceEndpointsOverLimit).
+
+### MimirKubernetesServiceEndpointsOverLimit
+
+This alert fires if a Kubernetes service used by Mimir for service discovery has reached the endpoints limit.
+
+How it **works**:
```
tacole02, line 983:
```suggestion
How it works:
```

tacole02: We only use bold formatting for UI elements, not for emphasis.

docs/sources/mimir/manage/mimir-runbooks/_index.md
tacole02, line 985:
```suggestion
- By default, the Kubernetes control plane creates and manages service endpoints to have no more than 100 endpoints each. Refer to ([EndpointSlices
](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)) in the Kubernetes documentation. You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
tacole02, line 986:
```suggestion
- After a Kubernetes service endpoints list reaches the limit, the endpoints are truncated to the limit.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+
+- By default, the Kubernetes control plane creates and manages service endpoints to have no more than 100 endpoints each ([doc](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/)). You can configure this limit with the `--max-endpoints-per-slice kube-controller-manager` flag, up to a maximum of 1000.
+- Once a Kubernetes service endpoints list reached the limit, the endpoints are truncated to the limit.
+- Mimir uses some Kubernetes services for DNS-based service discovery. If the endpoints of one of such services is getting truncated, Mimir is unable to discover all pods.
+
+How to **fix** it:
```
tacole02, line 989:
```suggestion
How to fix it:
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -970,6 +970,27 @@ How to **investigate**:
...
+
+- Increase Kubernetes `--max-endpoints-per-slice kube-controller-manager` to 1000, if it's currently running with a lower value. Configure the limit in the alert to reflect the new setting.
```
tacole02, line 991:
```suggestion
- Increase the Kubernetes `--max-endpoints-per-slice kube-controller-manager` limit to 1000, if it's currently running with a lower value. Configure the limit in the alert to reflect the new setting.
```

tacole02: If `--max-endpoints-per-slice kube-controller-manager` is a limit, it's good to specify that here. Same if it's a flag, setting, etc.

tacole02 (approved):
Looks good! Thank you 😄

docs/sources/mimir/manage/mimir-runbooks/_index.md
narqo: Out of the curiosity, I tried to catch up with the Kubernetes details, and I feel the "How it works" text is confusing here. I don't think there should be a connection between the `--max-endpoints-per-slice kube-controller-manager` and the observed DNS issue. The "1000 entries" problem feels like a coincidence rather than a relation to the `EndpointSlices`. Could it be so that the truncation happened on the client's resolver level (see also [Kubernetes DNS at scale][1])?

[1]: https://gist.github.com/aojea/32aeaa86aacebcdd93596ecb70fcba4f

pracucci:
Starting from [Vladimir's comment](https://github.com/grafana/mimir/pull/9381#discussion_r1771147318), I investigate it a bit more and (a) the DNS resolution limit looks unrelated from K8S endpoints limit and (b) looks like the actual DNS resolution limit I'm empirically seeing through a test in one of our largest Mimir installations is actually lower than 1000 entries. We need more time to investigate it, but it's clearly that this PR is wrong.

PR #9385, opened 2024-09-23
Add integration test assertion to ensure cortex_distributor_replication_factor is not exported when ingest storage is enabled

pracucci:
#### What this PR does

Add integration test assertion to ensure cortex_distributor_replication_factor is not exported when ingest storage is enabled. This covers a regression we recently had when running the experimental ingest storage at Grafana Labs.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/integration/query_frontend_test.go
+++ b/integration/query_frontend_test.go
@@ -455,6 +455,12 @@ func runQueryFrontendTest(t *testing.T, cfg queryFrontendTestConfig) {
 	assertServiceMetricsPrefixes(t, Querier, querier)
 	assertServiceMetricsPrefixes(t, QueryFrontend, queryFrontend)
 	assertServiceMetricsPrefixes(t, QueryScheduler, queryScheduler)
+
+	// Ensure cortex_distributor_replication_factor is not exported when ingest storage is enabled
+	// because it's how we detect whether a Mimir cluster is running with ingest storage.
+	if flags["-ingest-storage.enabled"] == "true" {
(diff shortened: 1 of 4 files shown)


integration/query_frontend_test.go
```diff
@@ -455,6 +455,12 @@ func runQueryFrontendTest(t *testing.T, cfg queryFrontendTestConfig) {
 	assertServiceMetricsPrefixes(t, Querier, querier)
 	assertServiceMetricsPrefixes(t, QueryFrontend, queryFrontend)
 	assertServiceMetricsPrefixes(t, QueryScheduler, queryScheduler)
+
+	// Ensure cortex_distributor_replication_factor is not exported when ingest storage is enabled
+	// because it's how we detect whether a Mimir cluster is running with ingest storage.
+	if flags["-ingest-storage.enabled"] == "true" {
+		assertServiceMetricsNotMatching(t, "cortex_distributor_replication_factor", queryFrontend, queryScheduler, distributor, ingester, querier)
+	}
```
narqo, line 463: Do we have any tests that verify this metric presents otherwise?

narqo (approved):
The changes overall look good to me 🔥 

I'm not sure if it's worth it to add an assertion, to check the metric exists in the non-ingest-store mode.

integration/query_frontend_test.go
pracucci: Good point, we don't. Added in https://github.com/grafana/mimir/pull/9385/commits/0062f2d7cb7cb1f0db7a113575fc0ed0edd9b18c.

Issue #9386, opened 2024-09-23
Ruler: add support for caching rule group contents


56quarters:
#### Is your feature request related to a problem? Please describe.

Currently, the contents of rule groups for tenants are not cached. This means an API call to return all configuration for rule groups must perform a list call for object storage (this is sometimes cached) and then N additional object storage requests. For tenants with a large number of rule groups, this can result in a huge number of calls.

Most rule group configurations don't change very often. This makes them a good candidate for caching. However when they are changed, that needs to be reflected immediately in API results. So if we want to add caching for rule groups we also need a way to invalidate their cached contents.

#### Describe the solution you'd like

We should:
* [X] Add functionality to `CachingBucket` to invalidate cached object storage contents on mutations. ([PR](https://github.com/grafana/mimir/pull/9387))
* [x] Unify "direct" and "cached' ruler storage backends in the Ruler so that cache invalidation works correctly. ([PR](https://github.com/grafana/mimir/pull/9434))
* [x] Add further cache invalidation logic to prevent data races. ([#591](https://github.com/grafana/dskit/pull/591), [#9546](https://github.com/grafana/mimir/pull/9546), [#601](https://github.com/grafana/dskit/pull/601), [#9575](https://github.com/grafana/mimir/pull/9575))
* [x] Configure ruler storage to use caching for `Get` operations ([PR](https://github.com/grafana/mimir/pull/9595))
* [ ] [optional] Increase the cache TTL for rule groups cached via `Get` operations to something longer than a single evaluation interval

#### Describe alternatives you've considered

Doing nothing is simpler and doesn't require opening up the can of worms that is cache invalidation.

PR #9387, opened 2024-09-23
Invalidate bucket caches on object modification or deletion

56quarters:
#### What this PR does

Invalidate content, existence, and attributes cached for a particular object when the object is modified or deleted.

Part of #9386

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/tsdb/bucketcache/caching_bucket.go
+++ b/pkg/storage/tsdb/bucketcache/caching_bucket.go
@@ -159,19 +159,26 @@ func NewCachingBucket(bucketID string, bucketClient objstore.Bucket, cfg *Cachin
 	return cb, nil
 }
 
-// invalidate invalidates content, attributes, and existence caches for the given
-// object. Note that this is best-effort and errors invalidating the cache are ignored.
+// invalidate invalidates content, existence, and attribute caches for the given object.
+// Note that this is best-effort and errors invalidating the cache are ignored.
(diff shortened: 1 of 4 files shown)


pkg/storage/tsdb/bucketcache/caching_bucket.go
```diff
@@ -159,6 +159,38 @@ func NewCachingBucket(bucketID string, bucketClient objstore.Bucket, cfg *Cachin
 	return cb, nil
 }
 
+func (cb *CachingBucket) invalidate(ctx context.Context, name string) {
```
pracucci, line 162: [nit] I would add a comment to explain this is just a best effort. It suffers race conditions, and there's no guarantee that an upload/delete that was reported as failure wasn't actually successfully completed on the server side.

pracucci, line 162: Should we allow to conditionally enable invalidation, so that we enable it only for caching buckets where we need it? Not feeling strong about it.

pracucci (approved):

pkg/storage/tsdb/bucketcache/caching_bucket.go
56quarters: I added an option to `CachingBucketConfig::ConfigGet()`

Issue #5698, opened 2023-08-08
Active series tracker by label value


bboreham:
#### Is your feature request related to a problem? Please describe.

People want to count active series subdivided by some label, e.g. team, department, job.

#### Describe the solution you'd like

An option on ingesters to expose counts similar to current active series tracker, but based on mapping a single label value instead of a list of regexps.

For example it could be:

```
    -ingester.active-series-custom-tracker-label=job
```

This should only be used on labels with modest cardinality - hundreds not millions.
There should be a limit to avoid runaway memory consumption.

#### Describe alternatives you've considered

See discussion at #2420

#### Additional context

The current implementation of ActiveSeriesTracker does a linear search of regexps every time a series is added, and consumes 1.5KB per tracker per tenant. This is unreasonably expensive.

PR #9392, opened 2024-09-24
Poc cost attribution

ying-jeanne:
#### What this PR does
The prototype to support adding custom label to track active series/ samples received / discarded samples related to this ticket https://github.com/grafana/mimir/issues/5698. 

The cost attribution related metrics would be exported through a different endpoint.
 
Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/distributor/distributor.go
```diff
@@ -307,7 +307,10 @@ func (m *PushMetrics) deleteUserMetrics(user string) {
 }
 
 // New constructs a new Distributor
-func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Overrides, activeGroupsCleanupService *util.ActiveGroupsCleanupService, ingestersRing ring.ReadRing, partitionsRing *ring.PartitionInstanceRing, canJoinDistributorsRing bool, reg prometheus.Registerer, log log.Logger) (*Distributor, error) {
+func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Overrides,
+	activeGroupsCleanupService *util.ActiveGroupsCleanupService, costAttributionClenaupService *util.CostAttributionCleanupService,
+	ingestersRing ring.ReadRing, partitionsRing *ring.PartitionInstanceRing,
+	canJoinDistributorsRing bool, reg prometheus.Registerer, log log.Logger) (*Distributor, error) {
```
colega, line 313: Opinionated: I'd drop these to one per line, once we go multipline.
And I'd leave the return params on their own line.

pkg/distributor/distributor.go
```diff
@@ -342,6 +345,7 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		healthyInstancesCount: atomic.NewUint32(0),
 		limits:                limits,
 		HATracker:             haTracker,
+		costAttributionsvr:    costAttributionClenaupService,
```
colega, line 348:
```suggestion
		costAttributionSvc:    costAttributionCleanupService,
```

pkg/distributor/distributor.go
```diff
@@ -356,7 +360,8 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		receivedSamples: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "cortex_distributor_received_samples_total",
 			Help: "The total number of received samples, excluding rejected and deduped samples.",
-		}, []string{"user"}),
+		}, []string{"user", "attrib"}),
```
colega, line 363: As mentioned offline: can we use `group` just like in the existing feature for push errors tracking?

pkg/distributor/distributor.go
```diff
@@ -1638,15 +1648,25 @@ func tokenForMetadata(userID string, metricName string) uint32 {
 	return mimirpb.ShardByMetricName(userID, metricName)
 }
 
-func (d *Distributor) updateReceivedMetrics(req *mimirpb.WriteRequest, userID string) {
+func (d *Distributor) updateReceivedMetrics(req *mimirpb.WriteRequest, userID string, costAttributionLabel string) {
 	var receivedSamples, receivedExemplars, receivedMetadata int
+	costAttribution := make(map[string]int)
```
colega, line 1653: I'd initialize this to `max-cost-attribution-per-user` size

pkg/ingester/activeseries/active_series.go
```diff
@@ -210,6 +231,18 @@ func (c *ActiveSeries) ActiveWithMatchers() (total int, totalMatching []int, tot
...
 }
 
+func (c *ActiveSeries) ActiveByAttributionValue() map[string]uint32 {
+	total := map[string]uint32{}
+	for s := 0; s < numStripes; s++ {
+		c.stripes[s].mu.RLock()
+		defer c.stripes[s].mu.RUnlock()
+		for k, v := range c.stripes[s].costAttributionValues {
+			total[k] += v
+		}
```
colega, line 241: ⚠️ You're locking all the stripes until the end of the execution of the method.

```suggestion
		for k, v := range c.stripes[s].costAttributionValues {
			total[k] += v
		}
		c.stripes[s].mu.RUnlock()
```

pkg/ingester/activeseries/active_series.go
```diff
@@ -210,6 +231,18 @@ func (c *ActiveSeries) ActiveWithMatchers() (total int, totalMatching []int, tot
 	return
 }
 
+func (c *ActiveSeries) ActiveByAttributionValue() map[string]uint32 {
+	total := map[string]uint32{}
```
colega, line 235: This can also be initialized to max size

pkg/ingester/activeseries/active_series.go
```diff
@@ -47,9 +48,13 @@ type ActiveSeries struct {
 	matchers           *Matchers
 	lastMatchersUpdate time.Time
 
+	CostAttributionLabel string
```
colega, line 51: Does this need to be exported?

pkg/ingester/activeseries/active_series.go
```diff
@@ -61,8 +66,8 @@ type seriesStripe struct {
 	// Unix nanoseconds. Only used by purge. Zero = unknown.
 	// Updated in purge and when old timestamp is used when updating series (in this case, oldestEntryTs is updated
 	// without holding the lock -- hence the atomic).
-	oldestEntryTs atomic.Int64
-
+	oldestEntryTs                        atomic.Int64
+	costAttributionsvr                   *util.CostAttributionCleanupService
```
colega, line 70:
```suggestion
	costAttributionSvc                   *util.CostAttributionCleanupService
```

pkg/ingester/ingester.go
```diff
@@ -380,15 +382,20 @@ func newIngester(cfg Config, limits *validation.Overrides, registerer prometheus
 }
 
 // New returns an Ingester that uses Mimir block storage.
-func New(cfg Config, limits *validation.Overrides, ingestersRing ring.ReadRing, partitionRingWatcher *ring.PartitionRingWatcher, activeGroupsCleanupService *util.ActiveGroupsCleanupService, registerer prometheus.Registerer, logger log.Logger) (*Ingester, error) {
+func New(
+	cfg Config, limits *validation.Overrides,
+	ingestersRing ring.ReadRing, partitionRingWatcher *ring.PartitionRingWatcher,
+	activeGroupsCleanupService *util.ActiveGroupsCleanupService,
+	costAttributionCleanupService *util.CostAttributionCleanupService,
+	registerer prometheus.Registerer, logger log.Logger) (*Ingester, error) {
```
colega, line 390: Same here, can you drop these to separate lines? Otherwise it just looks like a cloud of params.

pkg/ingester/ingester.go
```diff
@@ -312,6 +312,8 @@ type Ingester struct {
 
 	activeGroups *util.ActiveGroupsCleanupService
 
+	costAttribution *util.CostAttributionCleanupService
```
colega, line 315: I'd keep the name consistent:
```suggestion
	costAttributionSvc *util.CostAttributionCleanupService
```

pkg/ingester/ingester.go
```diff
@@ -1386,6 +1417,14 @@ func (i *Ingester) pushSamplesToAppender(userID string, timeseries []mimirpb.Pre
 	var builder labels.ScratchBuilder
 	var nonCopiedLabels labels.Labels
 	for _, ts := range timeseries {
+		// The cost attribution value for the series
+		costLabel := i.limits.CostAttributionLabel(userID)
```
colega, line 1421: There's no need to call this for each timeseries
```suggestion
	// The cost attribution value for the series
	costLabel := i.limits.CostAttributionLabel(userID)
	for _, ts := range timeseries {
```

pkg/ingester/ingester.go
```diff
@@ -1386,6 +1417,14 @@ func (i *Ingester) pushSamplesToAppender(userID string, timeseries []mimirpb.Pre
...
+		var costAttrib string
+		// when cost attribution label is set
+		if costLabel != "" {
+			costAttrib = i.costAttribution.UpdateAttributionTimestamp(userID, validation.AttributionValue(costLabel, userID, ts.Labels), time.Now())
```
colega, line 1425: Here and everywhere the time should be the request's time, not `time.Now()`, for both consistency and saving some resources calling this (calling `time.Now()` has a cost)

pkg/util/cost_attribution.go
```diff
@@ -0,0 +1,185 @@
...
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/services"
+	"go.uber.org/atomic"
+)
+
+type CostAttribution struct {
```
colega, line 16: This looks very similar to `pkg/util/active_groups.go`, as mentioned offline: both should follow the same logic and use the same label.

pkg/util/validation/limits.go
```diff
@@ -281,7 +284,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
...
 
 	f.StringVar(&l.SeparateMetricsGroupLabel, "validation.separate-metrics-group-label", "", "Label used to define the group label for metrics separation. For each write request, the group is obtained from the first non-empty group label from the first timeseries in the incoming list of timeseries. Specific distributor and ingester metrics will be further separated adding a 'group' label with group label's value. Currently applies to the following metrics: cortex_discarded_samples_total")
-
+	f.StringVar(&l.CostAttributionLabel, "validation.cost-attribution-label", "", "Label used to define the cost attribution label. For each write request, the cost attribution is obtained from the first non-empty cost attribution label from the first timeseries in the incoming list of timeseries. Specific distributor and ingester metrics will be further separated adding a 'cost' label with cost label's value. Currently applies to the following metrics: cortex_discarded_samples_total")
```
colega, line 287:
```suggestion
	f.StringVar(&l.CostAttributionLabel, "validation.cost-attribution-label", "", "Label used to define the cost attribution label. For each write request, the cost attribution is obtained from the first non-empty cost attribution label from the first timeseries in the incoming list of timeseries. Specific distributor and ingester metrics will be further separated adding a 'cost' label with cost attribution's label's value. Applies to the following metrics: cortex_discarded_samples_total.")
```

Also: isn't the label called `attr`?

pkg/util/validation/separate_metrics.go
```diff
@@ -30,3 +30,17 @@ func GroupLabel(o *Overrides, userID string, timeseries []mimirpb.PreallocTimese
...
+		return ""
+	}
+
+	for _, label := range lbs {
+		if label.Name == attribLab {
+			return label.Value
+		}
+	}
+	return ""
+}
```
colega, line 46: Why does this method need `userID`? And why do we need this method? How is it different from a usual label value getter?

pkg/util/validation/separate_metrics.go
```diff
@@ -30,3 +30,17 @@ func GroupLabel(o *Overrides, userID string, timeseries []mimirpb.PreallocTimese
 
 	return ""
 }
+
+// AttributionLabel obtains the value of cost attribution label for tenant
+func AttributionValue(attribLab string, userID string, lbs []mimirpb.LabelAdapter) string {
+	if len(lbs) == 0 {
+		return ""
+	}
```
colega, line 38: This condition is redundant with the for loop.

pkg/ingester/activeseries/active_series.go
```diff
@@ -392,6 +426,14 @@ func (s *seriesStripe) findAndUpdateOrCreateEntryForSeries(ref storage.SeriesRef
 		numNativeHistogramBuckets: numNativeHistogramBuckets,
 	}
 
+	// here if we have a cost attribution label, we can split the serie count based on the value of the label
+	// we also set the reference to the value of the label in the entry, so when remove, we can decrease the counter accordingly
+	if s.costAttributionLabel != "" {
+		attributionValue := s.costAttributionsvr.UpdateAttributionTimestamp(s.userID, series.Get(s.costAttributionLabel), time.Now())
```
colega, line 432:
```suggestion
		attributionValue := s.costAttributionsvr.UpdateAttributionTimestamp(s.userID, series.Get(s.costAttributionLabel), time.UnixNano(nowNanos))
```

bboreham:
Glad to see this make some progress.  Ideally I would like to see this kind of implementation take over from usage groups, in simple cases. So it would be good to have the label names, etc., line up.

pkg/distributor/distributor.go
ying-jeanne: my issue is what if a user configured using group and cost attribution feature together? I should have also touching the same metrics `cortex_discarded_samples_total`, i think group can be miss leading when activate 2 features together.

pkg/ingester/activeseries/active_series.go
ying-jeanne: done

pkg/ingester/ingester.go
ying-jeanne: done

pkg/ingester/metrics.go
```diff
@@ -303,7 +303,7 @@ func newIngesterMetrics(
 		activeSeriesPerUser: promauto.With(activeSeriesReg).NewGaugeVec(prometheus.GaugeOpts{
 			Name: "cortex_ingester_active_series",
 			Help: "Number of currently active series per user.",
-		}, []string{"user"}),
+		}, []string{"user", "attrib"}),
```
Logiraptor, line 306: To follow the pattern set by GEL, we would create a new metric instead of adding to the existing one. 

1. This new metric would be exposed on a separate endpoint apart from `/metrics` called `/usage_metrics`
2. The label _name_ should be dynamic and match the configured name, instead of using a constant `attrib` name

pkg/mimir/mimir.go
```diff
@@ -108,6 +108,8 @@ type Config struct {
 	NoAuthTenant                    string                 `yaml:"no_auth_tenant" category:"advanced"`
 	ShutdownDelay                   time.Duration          `yaml:"shutdown_delay" category:"advanced"`
 	MaxSeparateMetricsGroupsPerUser int                    `yaml:"max_separate_metrics_groups_per_user" category:"experimental"`
+	MaxCostAttributionPerUser       int                    `yaml:"max_cost_attribution_per_user" category:"experimental"`
```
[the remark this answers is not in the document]
Logiraptor: This would also reduce some of the noise in the diff, since we don't need to pass the value down separately from the limits, which we already pass around everywhere.

pkg/util/cost_attribution.go
```diff
@@ -0,0 +1,185 @@
...
+	// if the user attribution is already exist and we are not in the cooldown period, we don't need to check the limit
+	_, exists := ca.timestampsPerUser[userID][attribution]
+	if exists {
+		return false
+	}
+
+	// if the user has reached the limit, we will set the cooldown period which is 20 minutes
+	maxReached := len(ca.timestampsPerUser[userID]) >= ca.maxAttributionPerUser
+	if maxReached {
+		ca.coolDownDeadline[userID].Store(time.Now().Add(20 * time.Minute).UnixNano())
```
colega, line 131: Cool down interval should be configurable.
Shouldn't it be the same as inactiveTimeout?

In that case, I think we can just delete all user labels here, because we know they all are going to expire during the cool down period.

pkg/util/validation/separate_metrics.go
ying-jeanne: removed the helper function.

pkg/distributor/distributor.go
colega: > my issue is what if a user configured using group and cost attribution feature together?

Do we have any customer who wants to configure these two (for me one) features with different labels? I feel like the _active groups_ thing is the predecessor feature to _attribution_.

ying-jeanne: I mostly agree with you, just active groups is used for tracking discarded samples with reasons, for the cost breakdown, we don't need reason label, and in order to have it, we will need to store a detailed counter per reason. I start thinking that have a completely new metrics exported directly to a new registry making more sense for the following reasons, I will list this in design doc:
- lower cardinality: the new exported metrics would have only userID and cost-attribution-label as labels, this should be much easier, and no breaking changes for active group feature, less pressure on current registry
- can have high scraping interval, more flexible
- having dynamic label that corresponding to user config instead of hard coded "group"
- align with how loki/tempo's way of approaching this

colega (commented):
Made a quick pass, left some comments, didn't finish reviewing everything because I have two big concerns right now:
- The current implementation will use quite a lot of memory resources in active series tracker, I left appropriate comments on that.
- I don't think that the current implementation will handle correctly a configuration change on the cost attribution label and back, i.e.:
  - Current label is `a`, series is created.
  - Label is changed to `b`, counters in the series stripes are removed
  - Label is changed back to `a`
  - Active series are purged, since label is the same as when it was created, the counters which are already zero are decremented and are going negative.

pkg/costattribution/caimpl/managerImpl.go
```diff
@@ -0,0 +1,158 @@
+package caimpl
```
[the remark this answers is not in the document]
ying-jeanne: to isolate the dependencies, all function needs to be called out side of package should lay in manager.go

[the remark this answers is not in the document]
colega: I personally find this approach strange and unnecessary. Why do we need to isolate that? We don't do that anywhere.

I would move everything into the `costattribution` package.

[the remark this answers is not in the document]
ying-jeanne: to avoid import cycle, some projects do this, and easy mocks with interface. 👍 since we don't use this in Mimir would remove it to cost attribution as requested.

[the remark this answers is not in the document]
colega: I would suggest building the mocks in a different package, so production code just needs its production package, but test packages would import the mock packages. If you use `mockery` to generate them, you can use `--outpkg` and `--output` flags for that. 

Otherwise you'd need to put your mocks in the production interface package, and import two packages from your prod package.

pkg/ingester/metrics.go
ying-jeanne: Thanks, this is fixed in current PR.

pkg/util/cost_attribution.go
ying-jeanne: added cooldown config

pkg/costattribution/caimpl/tracker.go
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package caimpl
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+type Tracker struct {
```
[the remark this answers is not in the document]
ying-jeanne: made private

pkg/costattribution/caimpl/tracker_group.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"sync"
+	"time"
+
+	"go.uber.org/atomic"
+
+	"github.com/grafana/mimir/pkg/util/validation"
+)
+
+type AttributionTrackerGroup struct {
```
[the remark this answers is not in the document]
ying-jeanne: changed to private

pkg/util/validation/limits.go
ying-jeanne: updated PR, now we are using the customerized label instead, i would resolved this.

pkg/distributor/distributor.go
```diff
@@ -105,7 +106,7 @@ type Distributor struct {
 	distributorsLifecycler *ring.BasicLifecycler
 	distributorsRing       *ring.Ring
 	healthyInstancesCount  *atomic.Uint32
-
+	costAttributionMng     costattribution.Manager
```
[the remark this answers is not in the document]
ying-jeanne: updated to costAttributionMgr

pkg/ingester/activeseries/active_series.go
```diff
@@ -394,6 +428,20 @@ func (s *seriesStripe) findAndUpdateOrCreateEntryForSeries(ref storage.SeriesRef
...
+	if s.costAttributionMng != nil && s.costAttributionMng.EnabledForUser(s.userID) {
+		isOutDated, attributionValue := s.costAttributionMng.UpdateAttributionTimestamp(s.userID, s.caLabel, series, time.Unix(0, nowNanos))
+		if isOutDated {
+			// if the label is outdated, we need to remove the reference to the old value
+			s.costAttributionValues = map[string]uint32{}
+			s.caLabel = s.costAttributionMng.GetUserAttributionLabel(s.userID)
+		}
+		s.costAttributionValues[attributionValue]++
+		e.attributionValue = s.caLabel
+		e.attributionValue = attributionValue
```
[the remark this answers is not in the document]
ying-jeanne: fixed

pkg/costattribution/caimpl/managerImpl.go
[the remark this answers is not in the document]
colega: I don't think we've used `mockery` so far, but I'm not telling you to use _mockery_, the tool you use for the mock is orthogonal to the mock placement.

PR #9396, opened 2024-09-24
Remove localQueueIndex from QueuingAlgorithms

chencs:
#### What this PR does
With the introduction of leaf-node designations in #8871, we no longer need to maintain local queues on all nodes, and we can instead use `isLeaf()` to govern tree traversal decisions. 

Note that this introduces a subtle difference to the round-robin child queue ordering (`Node.queueOrder`) because all enqueued children will be enqueued "behind" the current `queueOrder[queuePosition]` node instead of maintaining a special case. This _does not_ change dequeue order, and is a no-op from the perspective of a caller to the tree.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/scheduler/queue/tenant_querier_assignment.go
+++ b/pkg/scheduler/queue/tenant_querier_assignment.go
@@ -254,13 +254,13 @@ func (tqa *tenantQuerierAssignments) dequeueSelectNode(node *Node) *Node {
 	if node.isLeaf() || len(node.queueMap) == 0 {
 		return node
 	}
-
 	// can't get a tenant if no querier set
 	if tqa.currentQuerier == "" {
 		return nil
 	}
 
-	// tenantOrderIndex is set to the _last_ tenant we dequeued from; advance queue position for dequeue
+	// Before this point, tenantOrderIndex represents the _last_ tenant we dequeued from;
+	// here, we advance queue position to point to the index we _will_ dequeue from
 	tqa.tenantOrderIndex++
 	if tqa.tenantOrderIndex >= len(tqa.tenantIDOrder) {
 		tqa.tenantOrderIndex = 0
--- a/pkg/scheduler/queue/tenant_querier_assignment.go
+++ b/pkg/scheduler/queue/tenant_querier_assignment.go
@@ -12,6 +12,9 @@ import (
 	"github.com/grafana/mimir/pkg/util"
 )
 
+// Newly-connected queriers should start from this index in tenantQuerierAssignments.tenantIDOrder.
+const newQuerierTenantIndex = -1
+
 type queueTenant struct {
 	tenantID    TenantID
 	maxQueriers int
@@ -92,7 +95,6 @@ func newTenantQuerierAssignments() *tenantQuerierAssignments {
 		tenantQuerierIDs: map[TenantID]map[QuerierID]struct{}{},
 		tenantNodes:      map[string][]*Node{},
 		currentQuerier:   "",
-		tenantOrderIndex: localQueueIndex,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/scheduler/queue/tenant_querier_assignment.go
```diff
@@ -249,20 +251,19 @@ func (tqa *tenantQuerierAssignments) setup(dequeueArgs *DequeueArgs) {
 // Note that because we use the shared  tenantIDOrder and tenantOrderIndex to manage the queue, we functionally
 // ignore each Node's individual queueOrder and queuePosition.
 func (tqa *tenantQuerierAssignments) dequeueSelectNode(node *Node) *Node {
+	if node.isLeaf() || len(node.queueMap) == 0 {
```
francoposa, line 254: `len(node.queueMap) == 0` should never occur for a non-leaf node, correct? any empty paths would get deleted up until the nodes have children again

pkg/scheduler/queue/tenant_querier_assignment.go
```diff
@@ -249,20 +251,19 @@ func (tqa *tenantQuerierAssignments) setup(dequeueArgs *DequeueArgs) {
...
+		return node
+	}
+
 	// can't get a tenant if no querier set
 	if tqa.currentQuerier == "" {
 		return nil
 	}
 
-	// advance queue position for dequeue
+	// tenantOrderIndex is set to the _last_ tenant we dequeued from; advance queue position for dequeue
```
francoposa, line 263: "is set to" makes this a little unclear - it is set to before this line or this is the line that sets it?
"starts at" or "was set to" would help

pkg/scheduler/queue/tree_queueing_algorithms.go
```diff
@@ -47,43 +47,42 @@ type roundRobinState struct {
...
+	if len(parent.queueOrder) > 1 {
 		parent.queuePosition++ // keep position pointed at same element
 	}
 }
 
 // dequeueSelectNode returns the node at the node's queuePosition. queuePosition represents the position of
 // the next node to dequeue from, and is incremented in dequeueUpdateState.
 func (rrs *roundRobinState) dequeueSelectNode(node *Node) *Node {
-	if node.queuePosition == localQueueIndex {
+	if node.isLeaf() || len(node.queueOrder) == 0 {
```
francoposa, line 69: same note on len of queueOrder == 0 should not be possible

francoposa:
opened https://github.com/grafana/mimir/pull/9440 to demonstrate simplifying the node insertion/deletion logic in a way that also lets have an easier-to-grok queue node ordering particularly for tests

pkg/scheduler/queue/tenant_querier_assignment.go
chencs: Yeah, this code is _technically_ unreachable under normal conditions, since leaf nodes also get checked for earlier in dequeue. I was torn on whether to leave this check in or take it out, and I'm still not strongly opinionated either way; I just left it because it seemed like the _especially_ dummy-proof thing to do.

francoposa: I am fine with that; I lean more towards "remove the safeguards and investigate if it breaks" because dummy-proofing is sort of a never-ending rabbit hole, but since we are still in rollout we can leave for now.

pkg/scheduler/queue/tenant_querier_assignment.go
```diff
@@ -254,13 +254,13 @@ func (tqa *tenantQuerierAssignments) dequeueSelectNode(node *Node) *Node {
...
 	}
-
 	// can't get a tenant if no querier set
 	if tqa.currentQuerier == "" {
 		return nil
 	}
 
-	// tenantOrderIndex is set to the _last_ tenant we dequeued from; advance queue position for dequeue
+	// Before this point, tenantOrderIndex represents the _last_ tenant we dequeued from;
+	// here, we advance queue position to point to the index we _will_ dequeue from
```
francoposa, line 263: nice

francoposa (approved):

PR #9398, opened 2024-09-25
MQE: add support for comparison operations

charleskorn:
#### What this PR does

This PR adds support in MQE for comparison operations: `==`, `!=`, `>`, `<`, `>=` and `<=`.

This is supported with and without the `bool` modifier, and for all forms of binary operations: vector/vector, vector/scalar and scalar/scalar.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/operators/vector_scalar_binary_operation.go
```diff
@@ -67,13 +77,23 @@ func NewVectorScalarBinaryOperation(
...
-			return f(scalar, vectorF, nil, vectorH)
+			return f(vectorF, scalar, vectorH, nil)
+		}
+	} else if op.IsComparisonOperator() && !returnBool {
+		b.opFunc = func(scalar float64, vectorF float64, vectorH *histogram.FloatHistogram) (float64, *histogram.FloatHistogram, bool, error) {
+			_, _, ok, err := f(scalar, vectorF, nil, vectorH)
+
+			// We always want to return the value from the vector when we're doing a filter-style comparison.
+			//
+			// We deliberately ignore the histogram value as we need to treat it as if it was a float with value 0,
```
aknuds1, line 90: Nit :)
```suggestion
			// We deliberately ignore the histogram value as we need to treat it as if it were a float with value 0,
```

tacole02 (approved):
👍

jhesketh (approved):
LGTM, thanks! Hopefully the upstream bug doesn't happen too often confusing our comparison results.

PR #9401, opened 2024-09-25
Upgrade to google.golang.org/grpc v1.66.2 / modify certain protobuf messages to retain their unmarshaling buffer

aknuds1:
#### What this PR does

Upgrade google.golang.org/grpc from v1.65.0 to v1.66.2. The [performance regression](https://github.com/grpc/grpc-go/pull/7571) referenced from dskit's reversal to v1.65.0 is fixed in v1.66.2, and the _potential_ functional regression also [referenced in said reversal](https://github.com/grafana/loki/issues/14053) is Loki specific.

What we found out is that v1.66 fixes a serious performance bottleneck in conjunction with compression; the `decompress` function is [optimized through memory pooling](https://github.com/grpc/grpc-go/blob/d0bf90aeb9b5bdf4031d812dbb743b0eb616c7b2/rpc_util.go#L903-L904), whereas v1.65.0 would use `io.ReadAll` and cause a lot of allocations during decompression. As a result, we saw ingester CPU usage increase by ~40% when enabling gRPC compression. The increase is now much more modest, maybe ~7% based on observations.

In order to keep allowing for unsafe references to unmarshalling buffers in e.g. `LabelAdapter`, I had to introduce a custom gRPC unmarshalling hook (`mimirpb.CodecV2.Unmarshal`) plus a scheme for protobuf messages to keep a reference to their unmarshalling buffer (`mimirpb.UnmarshalerV2`). The underlying reason is that from v1.66 on, gRPC immediately recycles each buffer after unmarshalling, unless a reference is kept. This causes data races with e.g. `LabelAdapter` taking unsafe string references to the unmarshal buffer, unless, as mentioned, a buffer reference is kept with the root protobuf message.

Ideally, one should call `FreeBuffer` on protobuf messages keeping an unmarshalling buffer reference so the buffer can be given back to the gRPC pool. If this isn't done though, there also shouldn't be a memory leak since the buffer should be garbage collected.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


aknuds1:
Some tests are breaking for this upgrade, have to fix them.

aknuds1:
Tests currently pass, but that is through undoing some memory unsafe optimizations in the `mimirpb` package. We want to investigate whether the optimizations can safely be retained.

pkg/mimirpb/custom.go
```diff
@@ -8,8 +8,76 @@ import (
...
+		panic(fmt.Errorf("unrecognized message type %T", v))
+	}
+}
+
+// Unmarshal customizes gRPC unmarshalling.
+// If v implements the UnmarshalerV2 interface, its SetBuffer method is called with the unmarshalling buffer.
+func (c *CodecV2) Unmarshal(data mem.BufferSlice, v any) error {
+	vv := messageV2Of(v)
+	buf := data.MaterializeToBuffer(mem.DefaultBufferPool())
+	defer buf.Free()
```
pracucci, line 44: [nit] We decrease the buffer ref counter but then later we retain it. However, before retaining it we call `Ref()` again, so I think this code is correct (this defer release it in case we `return err` below) but it's a bit tricky to understand so I would add a comment to explain.

pkg/mimirpb/split_test.go
```diff
@@ -166,7 +166,7 @@ func TestSplitWriteRequestByMaxMarshalSize_WriteRequestHasChanged(t *testing.T)
 
 	// If the fields of WriteRequest have changed, then you will probably need to modify
 	// the SplitWriteRequestByMaxMarshalSize() implementation accordingly!
-	assert.ElementsMatch(t, []string{"Timeseries", "Source", "Metadata", "SkipLabelValidation", "skipUnmarshalingExemplars"}, fieldNames)
+	assert.ElementsMatch(t, []string{"Timeseries", "Source", "Metadata", "SkipLabelValidation", "skipUnmarshalingExemplars", "buffer"}, fieldNames)
```
pracucci, line 169: I think the most correct thing to do in `SplitWriteRequestByMaxMarshalSize()` is to retain a reference of the original `buffer` in each split `WriteRequest` (and by calling `bufferRef()` for each split `WriteRequest`). We could also mention it in the `SplitWriteRequestByMaxMarshalSize()` function doc.

Then where we call `SplitWriteRequestByMaxMarshalSize()` (which is `marshalWriteRequestToRecords()`) we should call the `FreeBuffer()` on each split request once we've done marshalling it.

The only annoying thing of this approach is that `SplitWriteRequestByMaxMarshalSize()` will just return the input `WriteRequest` is not splitting happened at all. We could slightly modify it to also return a `bool` to indicate if splitting happened, so that the caller will call `FreeBuffer()` only if splitting happened.

pkg/mimirpb/custom.go
aknuds1: Yep, when the deferred call executes, the reference count will have been incremented unless an error occurred.

pkg/mimirpb/custom.go
```diff
@@ -8,8 +8,76 @@ import (
...
+	"google.golang.org/protobuf/protoadapt"
 )
 
+func init() {
+	c := encoding.GetCodecV2(proto.Name)
+	encoding.RegisterCodecV2(&CodecV2{c})
+}
+
+// CodecV2 customizes gRPC unmarshalling.
+type CodecV2 struct {
```
colega, line 24: I would rename this codec to reflect it's purpose, something like: `BufferHolderCodec`

pkg/mimirpb/custom.go
```diff
@@ -8,8 +8,76 @@ import (
...
+	encoding.CodecV2
```
colega, line 25: I would not embed this here, but convert it to a field and implement the two methods missing.

Otherwise it's unclear what are the other things that we're inheriting from the orignal codec: for example here this means that our codec has the same `Name()` as the parent: is that ok?

aknuds1: It's definitely OK that `Name()` is the same as `encoding.CodecV2`, as it should replace it:

```golang
func init() {
	c := encoding.GetCodecV2(proto.Name)
	encoding.RegisterCodecV2(&CodecV2{c})
}
```

I think that embedding is good, because our implementation is supposed to only customize the `Unmarshal` method of the standard `CodecV2`.

pkg/mimirpb/custom.go
aknuds1: I think the name `CodecV2` makes complete sense, since it wraps the `codecV2` type in `grpc/encoding/proto` and it implements the `grpc/encoding.CodecV2` interface.

pkg/mimirpb/custom.go
```diff
@@ -8,8 +8,76 @@ import (
...
+		buf.Ref()
+		unmarshaler.SetBuffer(buf)
+	}
+
+	return nil
+}
+
+// UnmarshalerV2 is an interface for protobuf messages that keep unsafe references to the unmarshalling buffer.
+// Implementations of this interface should keep a reference to said buffer.
+type UnmarshalerV2 interface {
```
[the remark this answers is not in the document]
aknuds1: Makes sense, done.

pkg/mimirpb/custom.go
aknuds1: Done.

pkg/mimirpb/custom.go
pstibrany: > I think that embedding is good, because our implementation is supposed to only customize the Unmarshal method of the standard CodecV2.

We can still "forward" implementation of other methods to original implementation, if we use named field. It would just be more explicit.

aknuds1: Yes of course, it's possible to do the same without embedding. Help me understand though, why should we not use the embedding technique in this case? Are you against it in principle?

colega: > If this case isn't right for struct embedding, then I wonder which is?

I think struct embedding is valid in codebases that respect backwards incompatible changes. Given that the grpc package makes breaking changes in minor versions, it would be very useful, IMO, to detect that they suddenly added a new method to the CodecV2 interface that changes the behaviour in some drastic way.

> We can still "forward" implementation of other methods to original implementation, if we use named field. It would just be more explicit.

Yes, that's what I meant.

```go
type CodecV2 struct {
    wrapped encoding.CodecV2
}

func (c *CodecV2) Name() string { return c.wrapped.Name() }
```

Etc.

pkg/mimirpb/custom.go
colega: IMO, our codec, that does special things, should reflect in it's name what's the special thing it's doing. Their `codecV2` name is because it's the default implementation.

BTW, do we need ours to be exported?

pkg/mimirpb/split_test.go
colega: I'm not sure I agree on this. These are _internal things we do with the WriteRequest_, they shouldn't affect the outer lifecycle: we had a WriteRequest, we call FreeBuffer after using it (and _using it_ here implies splitting it into multiple ones).

Maybe we should clarify in the comment of `SplitWriteRequestByMaxMarshalSize` that the requests returned in the slice _might_ still retain references to the original WriteRequest.

pstibrany: I agree with you Oleg, and suggested adding the comment instead on internal slack thread as well.

pkg/mimirpb/custom.go
aknuds1: No longer embedding.

pkg/mimirpb/split_test.go
aknuds1: After reviewing the code, I agree w/ @colega and @pstibrany and opted to document that the `WriteRequests` returned by `SplitWriteRequestByMaxMarshalSize` retain references to fields in the original `WriteRequest`.`

pkg/distributor/query.go
```diff
@@ -266,6 +266,8 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 				return ingesterQueryResult{}, err
 			}
 
+			defer resp.FreeBuffer()
```
colega, line 269: This is a `for` loop on the stream, and we're getting multiple responses here, and we're retaining a reference to each one of them until the end of the execution. We should free the buffer as soon as we've used it.

pkg/ingester/client/custom.go
```diff
@@ -9,10 +9,51 @@ import (
...
+func (m *ExemplarQueryResponse) SetBuffer(buf mem.Buffer) {
+	m.buffer = buf
+}
+
+func (m *ExemplarQueryResponse) FreeBuffer() {
+	if m.buffer != nil {
+		m.buffer.Free()
+		m.buffer = nil
+	}
+}
```
colega, line 55: All the BufferHolder implementation are the same?

I think _this_ might be a good candidate for embedding. 

What if we defined `BufferHolder` as the implementation, not an interface, and just embedded it in each one of the messages?

pkg/distributor/query.go
aknuds1: Good catch @colega, I missed that. Fixed, PTAL.

pkg/ingester/client/custom.go
aknuds1: Could we leave this to another PR? It would be a pain to fix up the *.pb.expdiff files to do this, and there are relatively few types affected. I would instead make this improvement in the future if more types would need to be `BufferHolder`s.

colega: I think that embedding a type reduces the amount of code and complexity, and we shouldn't merge something more complex just because we don't have a proper workflow to update the expdiff files.

Maybe this is a good opportunity to create a `make` target that would properly update the expdiff files? I tried following the instructions that you left in our internal weekly meeting, and the results I got don't look good: https://github.com/grafana/mimir/compare/arve/upgrade-grpc...arve/upgrade-grpc-keep-buffer-reference

aknuds1: @colega OK, I revised to use an embedded `BufferHolder` from the *.pb.go files, PTAL.

colega (approved):
LGTM, thank you for addressing all my comments.

PR #9403, opened 2024-09-25
MQE: Tidy up test case

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/engine_test.go
```diff
@@ -1810,3 +1767,78 @@ func TestCompareVariousMixedMetrics(t *testing.T) {
 		}
 	}
 }
+
+func TestCompareVariousMixedMetricsBinaryOperations(t *testing.T) {
+	labelsToUse := []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o"}
```
charleskorn, line 1772: As far as I can tell, this is the same for each test case - could we move this to one place?

For bonus points: could we somehow generate this from the series definitions in `runMixedMetricsTests` or vice versa, so they never get out of sync?

pkg/streamingpromql/engine_test.go
```diff
@@ -1810,3 +1767,78 @@ func TestCompareVariousMixedMetrics(t *testing.T) {
...
+	labelCombinations = append(labelCombinations, testutils.Combinations(labelsToUse, 4)...)
+
+	expressions := []string{}
+
+	for _, labels := range labelCombinations {
+		labelRegex := strings.Join(labels, "|")
+		expressions = append(expressions, fmt.Sprintf(`rate(series{label=~"(%s)"}[45s])`, labelRegex))
+		expressions = append(expressions, fmt.Sprintf(`rate(series{label=~"(%s)"}[1m])`, labelRegex))
+		expressions = append(expressions, fmt.Sprintf(`avg(rate(series{label=~"(%s)"}[2m15s]))`, labelRegex))
+		expressions = append(expressions, fmt.Sprintf(`avg(rate(series{label=~"(%s)"}[5m]))`, labelRegex))
```
charleskorn, line 1840: I know we were using `avg` before, but I'm wondering if `sum` would be better: if we have a case where we drop a series, and that series just so happens to have the same value as the average, we won't notice it, whereas that difference will be visible in a sum.

pkg/streamingpromql/engine_test.go
jhesketh: This is needed because each set of tests may want to do different labels. I can deduplicate it a bit though, but I'm not aiming for bonus points ;-).

charleskorn (approved):

PR #9412, opened 2024-09-25
mixin: fix errors on autoscaling metrics after series churn 

dimitarvdimitrov:
#### What this PR does

The promQL joins would fail when the series churn. For example, when
there are labels added by scrapers or when the K8s exporter restarts
and gets a new `pod` label.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥

PR #9417, opened 2024-09-26
MQE: Add helper for idx calcs

jhesketh:
#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9418, opened 2024-09-26
MQE: Add extra comparison op edge case tests

jhesketh:
Related to https://github.com/grafana/mimir/pull/9398

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


jhesketh:
(I'll update CHANGELOG after https://github.com/grafana/mimir/pull/9417 merges)

pkg/streamingpromql/engine_test.go
```diff
@@ -1889,3 +1889,30 @@ func TestCompareVariousMixedMetricsRate(t *testing.T) {
...
+	// Test each label individually to catch edge cases in with single series
+	labelCombinations := testutils.Combinations(labelsToUse, 1)
+	// Generate combinations of 2 labels. (e.g., "a,b", "e,f", etc)
+	labelCombinations = append(labelCombinations, testutils.Combinations(labelsToUse, 2)...)
+
+	expressions := []string{}
+
+	for _, labels := range labelCombinations {
+		labelRegex := strings.Join(labels, "|")
+		for _, op := range []string{"==", "!=", ">", "<", ">=", "<="} {
```
charleskorn, line 1905: This will only test vector / scalar combinations. Should we also add some cases for vector / vector combinations?

jhesketh: Good idea, done.

Are we testing scalar / scalar cases anywhere? (they should't go here, but I only recall a few basic comparisons upstream)

charleskorn: >Are we testing scalar / scalar cases anywhere? (they should't go here, but I only recall a few basic comparisons upstream)

Yep, [here](https://github.com/grafana/mimir/blob/45a14c913b0433003b28d17cc1f958dcfe72fd14/pkg/streamingpromql/testdata/ours/binary_operators.test#L703-L720).

charleskorn (approved):

PR #4950, opened 2023-05-08
Add optional cache support to ruler storage

pracucci:
#### What this PR does
In this PR I propose to add an optional shared cache for the ruler storage. In particular, in this PR I'm caching the two list objects operation issued by the ruler (list tenants and list rule groups of a tenant). When the cache is enabled, the rules synching becomes eventually consistent: after a change, the next sync is not guaranteed to be strongly consistent (e.g. new rule groups may not be found) but the next-next sync is expect to align it.

Note to reviewers:
- Added ` -ruler-storage.cache.*` config options. It's safe to reuse an existing memcached or redis instance (_see below_).
- Renamed `mimir.Mimir.RulerStorage` to `mimir.Mimir.RulerDirectStorage` to clarify it's the storage client without caching, and introduced `mimir.Mimir.RulerCachedStorage`. It's required to have both of them on `mimir.Mimir` struct to let GEM access them.
- Introduced `bucketID` to specify an (optional) bucket identifier to avoid cache collisions when sharing blocks and ruler bucket caching in the same memcached/redis instance.

_I also manually tested this changes in the local dev environment and looks working as designed._

#### Which issue(s) this PR fixes or relates to

Part of #4941

#### Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #9434, opened 2024-09-26
Unify direct and caching `RuleStore`s in ruler

56quarters:
#### What this PR does

Instead of using two different `RuleStore` implementations within the Ruler, use a single caching implementation and selectively disable caching when required.

This change removes the "direct" `RuleStore` implementation from the Ruler's gRPC and HTTP API layers. Instead, the caching implementation is used for all calls. In cases where caching returning stale results would not be acceptable, the caching is disabled _just_ for that call.

This allows rule group contents to be safety cached with the understanding that it is safe to cache them because they will correctly invalidated when deleted or modified.

Part of #9386

#### Notes to reviewers

I had a bit of trouble getting my head around this change. I've found the following to be a good way of thinking about it:

* We allow `Bucket.Iter()` calls to be cached already but use a non-caching bucket in some places where that's not acceptable.
* This change removes the non-caching bucket and selectively disables caching where stale `Bucket.Iter()` results are not acceptable. The point of using a single bucket is to allow caches to be invalidated on mutations (uploads, deletes).
* This allows us to add caching to rule group contents in the future but does not configure caching for them yet.
* When we add caching to rule group contents, it will be unconditional because cached rule group contents can be easily invalidated when the rule group is deleted or modified. Invalidating `Bucket.Iter()` calls is not so easy which is why this change adds the ability to disable caching of them.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/development/mimir-microservices-mode/docker-compose.jsonnet
+++ b/development/mimir-microservices-mode/docker-compose.jsonnet
@@ -294,7 +294,7 @@ std.manifestYamlDoc({
 
   memcached:: {
     memcached: {
-      image: 'memcached:1.6.19-alpine',
+      image: 'memcached:1.6.28-alpine',
       ports: [
         '11211:11211',
       ],
@@ -303,7 +303,7 @@ std.manifestYamlDoc({
 
   memcached_exporter:: {
     'memcached-exporter': {
-      image: 'prom/memcached-exporter:v0.6.0',
+      image: 'prom/memcached-exporter:v0.14.4',
       command: ['--memcached.address=memcached:11211', '--web.listen-address=0.0.0.0:9150'],
     },
   },
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -38,7 +38,6 @@ import (
 	"github.com/grafana/mimir/pkg/mimirpb"
 	"github.com/grafana/mimir/pkg/ruler/rulespb"
 	"github.com/grafana/mimir/pkg/ruler/rulestore"
-	"github.com/grafana/mimir/pkg/storage/tsdb/bucketcache"
 	"github.com/grafana/mimir/pkg/util"
 	"github.com/grafana/mimir/pkg/util/grpcencoding/s2"
 	util_log "github.com/grafana/mimir/pkg/util/log"
@@ -314,13 +313,12 @@ type MultiTenantManager interface {
 type Ruler struct {
 	services.Service
 
-	cfg         Config
-	lifecycler  *ring.BasicLifecycler
-	ring        *ring.Ring
-	directStore rulestore.RuleStore
-	cachedStore rulestore.RuleStore
-	manager     MultiTenantManager
-	limits      RulesLimits
+	cfg        Config
+	lifecycler *ring.BasicLifecycler
+	ring       *ring.Ring
+	store      rulestore.RuleStore
+	manager    MultiTenantManager
+	limits     RulesLimits
 
 	metrics *rulerMetrics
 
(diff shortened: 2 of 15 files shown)


pkg/ruler/ruler.go
```diff
@@ -660,7 +652,12 @@ func (r *Ruler) listRuleGroupsToSyncForAllUsers(ctx context.Context, reason rule
 
 	// In order to reduce API calls to the object storage among all ruler replicas,
 	// we support lookup of stale data for a short period.
-	users, err := r.cachedStore.ListAllUsers(bucketcache.WithCacheLookupEnabled(ctx, cacheLookupEnabled))
+	var opts []rulestore.Option
+	if !cacheLookupEnabled {
+		opts = append(opts, rulestore.WithCacheDisabled())
+	}
+
+	users, err := r.store.ListAllUsers(ctx, opts...)
```
56quarters, line 660: Passing the cache/don't cache option here doesn't actually affect anything since we specifically configure tenant listing calls to not be cached but it doesn't hurt to make that explicit here in case we change the config in the future.

development/mimir-microservices-mode/docker-compose.jsonnet
```diff
@@ -294,7 +294,7 @@ std.manifestYamlDoc({
 
   memcached:: {
     memcached: {
-      image: 'memcached:1.6.19-alpine',
+      image: 'memcached:1.6.28-alpine',
```
56quarters, line 297: Unrelated but I needed to be able to `watch deletions` which was only available in a newer version and this matches what we run in production.

pracucci (commented):
I remember there was a reason why I picked the 2 cache implementations instead of a single one and selectively disable cache on a per request basis (I remember I considered that as an option but then I found an issue). Unfortunately I don't remember the issue. Try to look at nested calls. Don't want to block on this if you don't find a good reason why we should keep it disabled, but I remember it's something I considered back in time.

56quarters:
> I remember there was a reason why I picked the 2 cache implementations instead of a single one and selectively disable cache on a per request basis (I remember I considered that as an option but then I found an issue). Unfortunately I don't remember the issue. Try to look at nested calls. Don't want to block on this if you don't find a good reason why we should keep it disabled, but I remember it's something I considered back in time.

I have been looking at the original PR that split the rule store implementations (#4950) and reading some discussion from you and @pstibrany . From what I can tell, we ended up with the current design (split store implementations) because

* There was opposition to the original `RuleStore.WithCache()` design.
* The ruler API is expected to be consistent and reflect changes made immediately.
* The ruler itself is permitted to _sometimes_ return stale results. We specifically allow stale rule group list calls during periodic syncs from object storage or syncs triggered by the ruler ring changing.
* Cache invalidation is difficult and likely to have edge cases and so we wanted to avoid it.
* GEM has a `RuleStore` implementation for self-monitoring which is a consideration.

I believe this design satisfies most of our requirements:

> There was opposition to the original `RuleStore.WithCache()` design.

This change adds per-method options to disable caching when required, no `.WithCache()` needed.

> The ruler API is expected to be consistent and reflect changes made immediately.

The ruler API will usually reflect changes made immediately due to a combination of disabling caching for listing of rule groups and the way the invalidation added in #9387 works: We invalidate the cache of rule group contents _after_ writing the updated rule group to object storage successfully (or deleting it completely) and we invalidate the cache by deleting the entries. We don't attempt to modify the cache contents.

> The ruler itself is permitted to _sometimes_ return stale results. We specifically allow stale rule group list calls during periodic syncs from object storage or syncs triggered by the ruler ring changing.

We preserve the ability to disable caching for some ruler syncs but not others.

> Cache invalidation is difficult and likely to have edge cases and so we wanted to avoid it.

This PR does not start caching rule groups but I plan to cache them in a future PR. It's true that there is a potential race condition when the same rule group is being updated and read at the same time from different requests who's execution is interleaved. It's possible an object is not in cache, fetched from object storage by request A, updated in object storage by request B, invalidated in cache (no-op) by request B, and stored to the cache by request A (stale). This will be mitigated to some extent by the short TTL used for the ruler cache. I do not have a perfect solution for this but I think this acceptable enough as a tradeoff to reduce the number of object storage operations performed.

> GEM has a `RuleStore` implementation for self-monitoring which is a consideration.

The `RuleStore` in GEM for self-monitoring is pretty simple and shouldn't be hard to adapt to work with these changes. The only issue in GEM is the usual issue of duplicate API code because of support for the remote write ruler.

56quarters:
> > Cache invalidation is difficult and likely to have edge cases and so we wanted to avoid it.

> I do not have a perfect solution for this but I think this acceptable enough as a tradeoff to reduce the number of object storage operations performed.

Actually, I think this can be mostly mitigated with some [clever use](https://github.com/memcached/memcached/wiki/ProgrammingTricks#ghetto-central-locking) of the Memcached `add` command to prevent stale values from being cached immediately following an update to a rule group. With changes to dskit, this could be done entirely within the `CachingBucket` bucket implementation. Since this will require dskit changes I'd like to do it as a follow-up to this PR.

56quarters:
I've put a PR together that shows how the "locking" might work based on some changes to dskit:

https://github.com/grafana/mimir/pull/9516

pstibrany (approved):
Looks good, thank you. From what I can see, this is a refactoring without any change in logic (right?).

Issue #8697, opened 2024-07-12
Flaky TestPartitionReader_Commit


dimitarvdimitrov:
```
--- FAIL: TestPartitionReader_Commit (0.00s)
    --- FAIL: TestPartitionReader_Commit/commit_at_shutdown_doesn't_persist_if_we_haven't_consumed_any_records_since_startup (2.08s)
        logging.go:33: level info partition 1 msg starting consumption from partition start because no offset has been found start_offset -2 consumer_group test-group
        logging.go:33: level info partition 1 component kafka_client msg immediate metadata update triggered why querying metadata for consumer initialization
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader is starting to consume partition until target and max consumer lag is honored
        logging.go:33: level info partition 1 component kafka_client msg assigning partitions why new assignments from direct consumer how assigning everything new, keeping current assignment input test[1{-2 e-1 ce0}]
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader found no records to consume because partition is empty partition_start_offset 0 last_produced_offset -1
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader consumed partition and current lag is lower than configured target consumer lag last_consumed_offset -1 current_lag 0s
        logging.go:33: level debug partition 1 msg last commit offset successfully committed to Kafka offset 1
        logging.go:33: level info partition 1 msg stopping partition reader
        logging.go:33: level info partition 1 msg starting consumption from last consumed offset last_consumed_offset 1 start_offset 2 consumer_group test-group
        logging.go:33: level info partition 1 component kafka_client msg immediate metadata update triggered why querying metadata for consumer initialization
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader is starting to consume partition until target and max consumer lag is honored
        logging.go:33: level info partition 1 component kafka_client msg assigning partitions why new assignments from direct consumer how assigning everything new, keeping current assignment input test[1{2 e-1 ce0}]
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader is consuming records to honor target and max consumer lag partition_start_offset 0 last_produced_offset 2
        logging.go:33: level info partition 1 target_lag 2s max_lag 15s msg partition reader consumed partition and current lag is lower than configured target consumer lag last_consumed_offset 2 current_lag 12.882613ms
        reader_test.go:1700: 
[...]

PR #9436, opened 2024-09-26
ingest storage: proper shutdown of partitionCommitter

dimitarvdimitrov:
#### What this PR does

This `partitionCommitter` would be shut down via the services manager as soon as the service context is cancelled. This means that they shut down in parallel with the `PartitionReader`. The race comes when the `partitionCommitter` has already shut down while the `PartitionReader` is still processing some records. Then when the `PartitionReader` tries to `enqueueCommit`, that sets the atomic, but does not send this to Kafka.

As a result we may not always persist the latest commit to Kafka on shutdown.

Fixes https://github.com/grafana/mimir/issues/8697

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -131,7 +131,9 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 	if err != nil {
 		return errors.Wrap(err, "creating service manager")
 	}
-	err = services.StartManagerAndAwaitHealthy(ctx, r.dependencies)
+	// Use context.Background() because we want to stop all dependencies when the PartitionReader stops
+	// instead of stopping them when ctx is cancelled and while the PartitionReader is still running.
+	err = services.StartManagerAndAwaitHealthy(context.Background(), r.dependencies)
 	if err != nil {
 		return errors.Wrap(err, "starting service manager")
 	}


gotjosh (approved):
LGTM

pkg/storage/ingest/reader.go
```diff
@@ -131,7 +131,9 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 	if err != nil {
 		return errors.Wrap(err, "creating service manager")
 	}
-	err = services.StartManagerAndAwaitHealthy(ctx, r.dependencies)
+	// Use context.Background() because we want to stop all dependencies when the PartitionReader stops
+	// instead of stopping them when ctx is cancelled and while the PartitionReader is still running.
+	err = services.StartManagerAndAwaitHealthy(context.Background(), r.dependencies)
```
dimitarvdimitrov, line 136: the tradeoff here is that if we get interrupted during startup, we wouldn't respect that and we'd wait for the full startup to finish

not sure if I should try to solve that (maybe with something like `context.AfterFunc()`)

pracucci: If the `PartitionReader` gets terminated (context canceled) while running `PartitionReader.start()`, do we evern call `stopDependencies()` at all? If I remember correctly, `PartitionReader.stop()` will get called only if `PartitionReader.start()` has successfully terminated, that is if it has returned `nil`.

dimitarvdimitrov: we take care of it a few lines above

https://github.com/grafana/mimir/blob/01d32acfb8722f6a9ab37a294b5e35dfaf802206/pkg/storage/ingest/reader.go#L105-L110

gotjosh:
@pracucci should we wait on your to review this or can we proceed with the merge?

pkg/storage/ingest/reader.go
pracucci: Oh right, I missed it. LGTM Then.

pracucci (approved):

PR #9438, opened 2024-09-26
Alertmanager: Fix custom Grafana templates not defined

santihernandezc:
As we're using `[]io.Reader` for templates, we can read them only once. This is a problem because we use different structs to execute templates for Mimir and Grafana integrations, and both need the same templates. When we create the Grafana template struct, the templates have already been read, resulting in "template not defined" errors when trying to send notifications with custom templates.

Since we stopped reading templates from disk, there's no advantage in using an `io.Reader` besides using the interface  
`github.com/prometheus/alertmanager/template` expects in `template.Parse()`.

This PR replaces `[]io.Reader` for templates in favor of `[]TemplateDefinition`, converting them into an `io.Reader` when calling `template.Parse()`.

Fixes https://github.com/grafana/alerting-squad/issues/929

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -11,7 +11,6 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"path"
@@ -433,12 +432,7 @@ func clusterWait(position func() int, timeout time.Duration) func() time.Duratio
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, tmpls []alertingTemplates.TemplateDefinition, rawCfg string, tmplExternalURL *url.URL, staticHeaders map[string]string) error {
-	templates := make([]io.Reader, 0, len(tmpls))
-	for _, tmpl := range tmpls {
-		templates = append(templates, strings.NewReader(tmpl.Template))
-	}
-
-	tmpl, err := loadTemplates(templates, WithCustomFunctions(am.cfg.UserID))
+	tmpl, err := loadTemplates(tmpls, WithCustomFunctions(am.cfg.UserID))
 	if err != nil {
 		return err
 	}
@@ -468,7 +462,7 @@ func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, 
 		return d + waitFunc()
 	}
 
-	integrationsMap, err := am.buildIntegrationsMap(cfg.Global, conf.Receivers, tmpl, templates, staticHeaders)
+	integrationsMap, err := am.buildIntegrationsMap(cfg.Global, conf.Receivers, tmpl, tmpls, staticHeaders)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/alertmanager/alertmanager_template.go
```diff
@@ -95,7 +96,7 @@ func WithCustomFunctions(userID string) template.Option {
 
 // loadTemplates produces a template.Template from several in-memory template files.
 // It is adapted from FromGlobs in prometheus/alertmanager: https://github.com/prometheus/alertmanager/blob/9de8ef36755298a68b6ab20244d4369d38bdea99/template/template.go#L67-L95
-func loadTemplates(tmpls []io.Reader, options ...template.Option) (*template.Template, error) {
+func loadTemplates(tmpls []alertingTemplates.TemplateDefinition, options ...template.Option) (*template.Template, error) {
```
stevesg, line 99: [nit] I don't think the `grafana/alerting` types need to leak in here, I think this is cleaner.

```suggestion
func loadTemplates(tmpls []string, options ...template.Option) (*template.Template, error) {
```

pkg/alertmanager/alertmanager.go
```diff
@@ -433,9 +432,9 @@ func clusterWait(position func() int, timeout time.Duration) func() time.Duratio
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, tmpls []alertingTemplates.TemplateDefinition, rawCfg string, tmplExternalURL *url.URL, staticHeaders map[string]string) error {
-	templates := make([]io.Reader, 0, len(tmpls))
```
titolins, line 436: The problem is not really the `io.Reader`, that's just an interface. You could also keep the type signature and create a wrapper that implements `Reader` - e.g.:

```go
type unlimitedStringReader struct {
    s string
}

func (r *unlimitedStringReader) Read(p []byte) (n int, err error) {
    return strings.NewReader(r.s).Read(p)
}
```

pkg/alertmanager/alertmanager.go
```diff
@@ -433,9 +432,9 @@ func clusterWait(position func() int, timeout time.Duration) func() time.Duratio
...
+	templates := make([]string, 0, len(tmpls))
 	for _, tmpl := range tmpls {
-		templates = append(templates, strings.NewReader(tmpl.Template))
+		templates = append(templates, tmpl.Template)
```
titolins, line 437: Then here you could just do the below and probably (not tested 😅 ) not change anything else:
```suggestion
		templates = append(templates, &unlimitedStringReader{tmpl.Template})
```

titolins (commented):
Just added a possible different approach. Will try to test this later today before approving 👍

pkg/alertmanager/alertmanager.go
santihernandezc: Yes, but we don't need this level of abstraction.
We're currently using readers just because we need to pass readers down the pipeline, it's a relic from reading templates from disk. We can do perfectly fine using plain strings.

titolins: > it's a relic from reading templates from disk

Ahh, tks for the context. It does make more sense that we'd like to get rid of it. I'm testing it now and should approve in a bit 👍

titolins (approved):

rwwiv (approved):
It would be good to see a test that illustrates the fix, but changes LGTM

PR #9447, opened 2024-09-27
kafka replay speed: fix concurrent fetching concurrency transition

dimitarvdimitrov:
#### What this PR does

fixes

1. There's a race between `Add()`ing to the `WaitGroup` and then `Wait()`ing for it; it's possible that we `close(r.done)` and `Wait()`; then `Wait()` just returns immediately
2. If we immediately call `Update()` after `newConcurrentFetchers`, then `lastReturnedRecord` might stay as -1. So we'd resume from offset -1 🥴 instead of from the start or end of the partition
3. after `Update()` we should continue from `lastReturnedRecord+1` because we've already fetched and returned `lastReturnedRecord`; this isn't critical though

This is missing tests; WIP

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/reader.go
```diff
@@ -1340,6 +1339,7 @@ func handleKafkaFetchErr(err error, fw fetchWant, longBackoff waiter, partitionS
 	case errors.Is(err, kerr.OffsetOutOfRange):
 		// We're either consuming from before the first offset or after the last offset.
 		partitionStart, err := partitionStartOffset.CachedOffset()
+		logger = log.With(logger, "partition_start_offset", partitionStart, "start_offset", fw.startOffset, "end_offset", fw.endOffset)
```
dimitarvdimitrov, line 1342:
```suggestion
		logger = log.With(logger, "log_start_offset", partitionStart, "start_offset", fw.startOffset, "end_offset", fw.endOffset)
```

gotjosh (approved):
LGTM

gotjosh (approved):

PR #9449, opened 2024-09-27
fix: fix slice init length

cuishuang:
#### What this PR does

The intention here should be to initialize a slice with a capacity of len(am.templates) rather than initializing the length of this slice.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


56quarters (approved):
Nice catch, thank you!

PR #9452, opened 2024-09-27
kafka replay speed: upstream concurrent fetchers

dimitarvdimitrov:
This is the second of series of PRs to upstream the code for improving Kafka replay speed in the ingester.

In this PR I'm upstreaming the fetching code. The core of the change is in `concurrentFetchers`.

### franz-go fork

We need the changes in this PR https://github.com/twmb/franz-go/pull/803, so for now I'm running a fork

### `concurrentFetchers` Overview

* **Segmentation (fetchWant):** The fetcher divides the work into segments called fetchWants. Each fetchWant represents a range of records to be fetched, defined by a start and end offset. This segmentation allows for concurrent fetching of different parts of the topic partition.

* **Concurrent Fetching:** Multiple goroutines (defined by the concurrency parameter) work on fetching these segments simultaneously. Each goroutine runs the `run` method, which processes fetchWants from a channel. The results (`fetchResult`) of the fetch are send to another channel: `fetchWant.result`
  * Different configuration of concurrency and size of fetchWant yield different results
    * low concurrency and low number of records per fetch yields lower end-to-end latency
    * high concurrency and high number of records per fetch yields higher replay speed
  * Because of this we chose to have two settings: while we're catching up and after that. These are both controlled via two sets of flags: `.startup-fetch-concurrency & .startup-records-per-fetch` and `.ongoing-fetch-concurrency & .ongoing-records-per-fetch`

* **Multiple Attempts:** If a fetch attempt fails or doesn't retrieve all requested records, the fetcher will retry. The fetcher updates the start offset of the fetchWant based on the last successfully fetched record and continues until all requested records are retrieved.

* **Overfetching Risk:** The system risks overfetching due to limitations in the Kafka protocol:
   * The Kafka protocol doesn't allow specifying the number of records to fetch; only the number of bytes can be specified.
   * This means that a fetch might retrieve more records than intended, especially at the end of one fetchWant.
   * These extra records can overlap with the next fetchWant's range.
   * To handle this, concurrentFetchers tracks the last returned record offset (`lastReturnedRecord`). We use `lastReturnedRecord` to remove any records from a `fetchResult` that may have already been processed.

* **Ordering:** The fetcher ensures that segments are processed in order by:
   * Using a linked list (`pendingResults`) to keep track of fetch results channels in the order they were requested
   * Buffering results in `bufferedResult` and only sending them to `orderedFetches` channel when they're next in sequence
   * The `PollFetches` method, which consumers call to get records, receives from the `orderedFetches` channel, ensuring records are always returned in the correct order

* **Adaptive Fetching:** The system adapts the size of fetch requests based on previous results. It estimates the bytes per record and adjusts the `MaxBytes` parameter of fetch requests accordingly, trying to optimize the amount of data fetched in each request.

### Changes to existing code

Most of this PR introduces new code. However, there are some changes to existing behaviours

* `processNextFetchesUntilTargetOrMaxLagHonored`: the second attempt is now allowed more time. The reason for this is that the catchup speed when during stable state-consumption is slower. For example, this means that we can't reduce lag from 15s to 2s within 15s. In those cases leaving 30s is enough to catch up to below 2s lag.

### Testing

`concurrentFetchers` is now the default in unit tests in `pkg/storage/ingest`. This increases the test coverage significantly and didn't require us to write so many tests.

### Future work

* Backoff and retry getting offset and topic ID when creating `concurrentFetchers` - this would unblock making concurrent fetchers the default (some e2e tests fail if we can't resolve the topic the first time)
* dynamic concurrency based on an EWMA; remove the two sets of flags for before and after startup


pkg/storage/ingest/fetcher.go
```diff
@@ -0,0 +1,712 @@
...
+}
+
+func (fr *fetchResult) finishWaitingForConsumption() {
+	if fr.waitingToBePickedUpFromOrderedFetchesSpan == nil {
+		fr.waitingToBePickedUpFromOrderedFetchesSpan, fr.ctx = opentracing.StartSpanFromContext(fr.ctx, "fetchResult.noWaitingForConsumption")
+	}
+	fr.waitingToBePickedUpFromOrderedFetchesSpan.Finish()
+}
+
+// Merge merges other with an older fetchResult. mergedWith keeps most of the fields of fr and assumes they are more up to date then other's.
```
gotjosh, line 177:
```suggestion
// Merge merges other with an older fetchResult. Merge keeps most of the fields of fr and assumes they are more up-to-date than older.
```

pkg/storage/ingest/fetcher.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+	f.topicID = topics[topic].ID
+
+	f.wg.Add(1)
+	go f.start(ctx, startOffset, concurrency, recordsPerFetch)
+
+	return f, nil
+}
+
+func (r *concurrentFetchers) Update(ctx context.Context, concurrency, records int) {
```
gotjosh, line 288:
```suggestion
// Update implements fetcher
func (r *concurrentFetchers) Update(ctx context.Context, concurrency, records int) {
```

pkg/storage/ingest/fetcher.go
```diff
@@ -0,0 +1,712 @@
...
+	r.Stop()
+	r.done = make(chan struct{})
+
+	r.wg.Add(1)
+	go r.start(ctx, r.lastReturnedRecord+1, concurrency, records)
+}
+
+func (r *concurrentFetchers) PollFetches(ctx context.Context) (kgo.Fetches, context.Context) {
```
gotjosh, line 296:
```suggestion
// PollFetches implements fetcher
func (r *concurrentFetchers) PollFetches(ctx context.Context) (kgo.Fetches, context.Context) {
```

pkg/storage/ingest/fetcher.go
```diff
@@ -0,0 +1,712 @@
...
+
+	orderedFetches     chan fetchResult
+	lastReturnedRecord int64
+	startOffsets       *genericOffsetReader[int64]
+
+	// trackCompressedBytes controls whether to calculate MaxBytes for fetch requests based on previous responses' compressed or uncompressed bytes.
+	trackCompressedBytes bool
+}
+
+func (r *concurrentFetchers) Stop() {
```
gotjosh, line 221:
```suggestion
// Stop implements fetcher
func (r *concurrentFetchers) Stop() {
```

gotjosh: nit: should be move this method below the constructor?

pkg/storage/ingest/fetcher.go
```diff
@@ -0,0 +1,712 @@
...
+	f.topicID = topics[topic].ID
+
+	f.wg.Add(1)
+	go f.start(ctx, startOffset, concurrency, recordsPerFetch)
+
+	return f, nil
+}
+
+func (r *concurrentFetchers) Update(ctx context.Context, concurrency, records int) {
+	r.Stop()
```
gotjosh, line 289: nit: Can you add a task to not update concurrency if we are already at that one? With the change we made as we approach the desired lag, there's a chance that we're already at the specified concurrency and we've stopped and start the workers for no reason.

gotjosh (approved):
LGTM

My comments are all nits, I don't need to see this again and you can merge after you've decided which ones you want to pick.

PR #9454, opened 2024-09-27
kafka replay speed: upstream push sharding

dimitarvdimitrov:
### What this PR does

This is the third of series of PRs to upstream the code for improving Kafka replay speed in the ingester.

In this PR we upstream push sharding. The basic idea is that instead of ingesting each record from Kafka sequentially, we can shard each timeseries and ingest them in parallel.

_I'm submitting the PR, but the work was done jointly by @gotjosh and myself_

#### Basic breakdown

1. `pusherConsumer`
   - This component receives records from Kafka and pushes them to storage.
   - It uses a separate goroutine for unmarshaling records while pushing the current record to storage.
   - After unmarshalling each record is pushed to a `parallelStoragePusher`

2. `parallelStoragePusher`
   - Manages parallel pushing of WriteRequests to storage.
   - Tenant separation: It maintains a map of tenant IDs to `parallelStorageShards`. This ensures each tenant's data is processed independently and in parallel.
   - It also maintains one shard per write request source (Ruler, API). This means we have double the number of shards.
   - When closing, all shards are flushed because:
     a) It provides a complete unit (batch of records) that can be retried if necessary.
     b) It ensures all data from the batch is processed before moving to the next batch.

3. `parallelStorageShards`
   - Handles sharding of series for parallel processing for a single `(tenant, WriteRequest.Source)` pair
   - Sharding method: Each series is sharded based on the hash of its labels. The hashing function used is the same as [Prometheus TSDB's stripeSeries](https://github.com/grafana/mimir/blob/e6ec9ddf5766d4528f6d5844c5ad9447ff020cad/vendor/github.com/prometheus/prometheus/tsdb/head.go#L2027-L2035), to make best use of reduced lock contention.

4. `batchingQueue`
   - Manages batching and flushing of data for each shard.
   - Batching: Accumulates time series and metadata up to a specified batch size.
   - Flushing: When the batch size is reached or when explicitly closed, it pushes the batch to a channel for processing. The other end of that channel invokes the ingester pushing logic.

#### Error Handling:

Due to the concurrent nature of pushing, it is hard to immediately abort fetching. As a result errors are delayed and we may continue ingesting for a little bit until we detect the error.

The error handling has been extracted from `pusherConsumer` and implemented in the `parallelStoragePusher` and `sequentialStoragePusher`. This approach allows errors to be handled closer to their source. Instead of propagating lists of errors, only retriable critical errors that require aborting consumption are propagated. This simplifies error management and allows for more targeted retry strategies.

#### Cosmetic Changes

In `pusherConsumer.Consume`, the two goroutines for unmarshaling and ingestion have been inlined into the Consume function. There shouldn't be any functional changes there.

#### Metrics changes

We renamed `cortex_ingest_storage_reader_processing_time_seconds` to `cortex_ingest_storage_reader_records_processing_time_seconds` to be more consistent.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -6,141 +6,465 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand/v2"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/cancellation"
 	"github.com/grafana/dskit/middleware"
+	"github.com/grafana/dskit/multierror"
 	"github.com/grafana/dskit/user"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
 
 	"github.com/grafana/mimir/pkg/mimirpb"
 	util_log "github.com/grafana/mimir/pkg/util/log"
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+// batchingQueueCapacity controls how many batches can be enqueued for flushing.
+// We don't want to push any batches in parallel and instead want to prepare the next one while the current one finishes, hence the buffer of 1.
+// For example, if we flush 1 batch/sec, then batching 2 batches/sec doesn't make us faster.
+// This is our initial assumption, and there's potential in testing with higher numbers if there's a high variability in flush times - assuming we can preserve the order of the batches. For now, we'll stick to 5.
+// If there's high variability in the time to flush or in the time to batch, then this buffer might need to be increased.
+const batchingQueueCapacity = 5
+
 type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
+type PusherCloser interface {
+	PushToStorage(context.Context, *mimirpb.WriteRequest) error
+	// Calls to close are safe and will not be called concurrenctly.
+	Close() []error
+}
+
+// pusherConsumer receivers records from Kafka and pushes them to the storage.
+// Each time a batch of records is received from Kafka, we instantiate a new pusherConsumer, this is to ensure we can retry if necessary and know whether we have completed that batch or not.
 type pusherConsumer struct {
+	metrics *pusherConsumerMetrics
+	logger  log.Logger
+
+	kafkaConfig KafkaConfig
+
 	pusher Pusher
-
-	processingTimeSeconds prometheus.Observer
-	clientErrRequests     prometheus.Counter
-	serverErrRequests     prometheus.Counter
-	totalRequests         prometheus.Counter
-
-	fallbackClientErrSampler *util_log.Sampler // Fallback log message sampler client errors that are not sampled yet.
-	logger                   log.Logger
 }
 
-type parsedRecord struct {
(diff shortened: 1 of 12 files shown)


tacole02 (approved):

pkg/storage/ingest/pusher.go
```diff
@@ -6,141 +6,465 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/prometheus/model/labels"
 
 	"github.com/grafana/mimir/pkg/mimirpb"
 	util_log "github.com/grafana/mimir/pkg/util/log"
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+// batchingQueueCapacity controls how many batches can be enqueued for flushing.
+// We don't want to push any batches in parallel and instead want to prepare the next one while the current one finishes, hence the buffer of 1.
```
gotjosh, line 27:
```suggestion
// batchingQueueCapacity controls how many batches can be enqueued for flushing.
// We don't want to push any batches in parallel and instead want to prepare the next ones while the current one finishes, hence the buffer of 5.
```

pkg/storage/ingest/pusher.go
```diff
@@ -6,141 +6,465 @@ import (
...
+
 type Pusher interface {
 	PushToStorage(context.Context, *mimirpb.WriteRequest) error
 }
 
+type PusherCloser interface {
+	PushToStorage(context.Context, *mimirpb.WriteRequest) error
+	// Calls to close are safe and will not be called concurrenctly.
+	Close() []error
+}
```
gotjosh, line 41:
```suggestion
type PusherCloser interface {
	// PushToStorage pushes the write request to the storage.
	PushToStorage(context.Context, *mimirpb.WriteRequest) error
	// Close tells the PusherCloser that no more records are coming and it should flush any remaining records.
	Close() []error
}

```

pkg/storage/ingest/pusher.go
```diff
@@ -6,141 +6,465 @@ import (
...
+
+// pusherConsumer receivers records from Kafka and pushes them to the storage.
```
gotjosh, line 43:
```suggestion
// pusherConsumer receives records from Kafka and pushes them to the storage.
```

pkg/storage/ingest/pusher.go
```diff
@@ -6,141 +6,465 @@ import (
...
+		if err := p.shards[shard].AddMetadataToBatch(ctx, request.Source, request.Metadata[mdIdx]); err != nil {
+			return fmt.Errorf("encountered a non-client error when ingesting; this error was for a previous write request for the same tenant: %w", err)
 		}
+		shard++
+		shard %= p.numShards
 	}
+
+	// We might some data left in some of the queues in the shards, but they will be flushed eventually once Stop is called, and we're certain that no more data is coming.
+	// So far we didn't find any non-client errors that are worth aborting for.
+	// We'll call Close eventually and collect the rest.
```
gotjosh, line 345:
```suggestion
	// We might have some data left in some of the queues in the shards, but they will be flushed eventually once Stop is called, and we're certain that no more data is coming.
	// So far we didn't find any non-client errors that are worth aborting for.
	// We'll call Close eventually and collect the rest.
```

gotjosh (approved):
LGTM

PR #9482, opened 2024-10-01
MQE: fix incorrect query results or "found duplicate series for the match group" errors when binary operation has unsorted labels in `on`

charleskorn:
#### What this PR does

This PR fixes a bug in MQE's binary operation implementation where query results can be incorrect or incorrectly fail with "found duplicate series for the match group" errors.

`labels.Labels.BytesWithLabels` and `labels.Labels.BytesWithoutLabels` expect the list of label names they receive to be sorted. However, we weren't ensuring this was true when `BytesWithLabels` was called to generate the grouping key for a binary operation with `on`, so incorrect grouping keys were generated.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/testdata/ours/binary_operators.test
+++ b/pkg/streamingpromql/testdata/ours/binary_operators.test
@@ -98,6 +98,9 @@ eval range from 0 to 24m step 6m left_side - on(env, pod) right_side
   {env="test", pod="a"} -9 -18 -27
   {env="test", pod="b"} -36 -45 -54
 
+# Test the same thing again with the grouping labels in a different order.
+# (The implementation of binary operations relies on grouping labels being sorted in some places,
+# so this test exists to ensure this is done correctly.)
 eval range from 0 to 24m step 6m left_side - on(pod, env) right_side
   {env="prod", pod="a"} -63 -72 -81
   {env="test", pod="a"} -9 -18 -27
@@ -108,6 +111,9 @@ eval range from 0 to 24m step 6m left_side - ignoring(env, pod) right_side
   {group="bar"} -6 -15 -24
   {group="foo"} -69 -78 -87
 
+# Test the same thing again with the grouping labels in a different order.
+# (The implementation of binary operations relies on grouping labels being sorted in some places,
+# so this test exists to ensure this is done correctly.)
 eval range from 0 to 24m step 6m left_side - ignoring(pod, env) right_side
   {group="baz"} -33 -42 -51
   {group="bar"} -6 -15 -24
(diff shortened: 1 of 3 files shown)


pkg/streamingpromql/testdata/ours/binary_operators.test
```diff
@@ -98,11 +98,21 @@ eval range from 0 to 24m step 6m left_side - on(env, pod) right_side
   {env="test", pod="a"} -9 -18 -27
   {env="test", pod="b"} -36 -45 -54
 
+eval range from 0 to 24m step 6m left_side - on(pod, env) right_side
```
jhesketh, line 101: (nit) should add a comment here on the purpose of these tests and that the load order needs to be preserved

jhesketh (commented):
What do you think about having `BytesWithLabels` (etc) do the sort?

Granted, there are cases where we don't need to, and thus we don't want to add the overhead, so may we have a `sort` parameter on the function? This means we have to be explicit when using `BytesWithLabels` and hopefully avoid similar problems in the future

charleskorn:
> What do you think about having `BytesWithLabels` (etc) do the sort?
> 
> Granted, there are cases where we don't need to, and thus we don't want to add the overhead, so may we have a `sort` parameter on the function? This means we have to be explicit when using `BytesWithLabels` and hopefully avoid similar problems in the future

`BytesWithLabels` and `BytesWithoutLabels` is a method from Prometheus, so we'd have to convince them to add this parameter or change in behaviour. I doubt this would be accepted:

* adding the parameter is a breaking change to both methods
* most places that use either method call the method over and over again with the same set of labels like we do, so having `BytesWithLabels` or `BytesWithoutLabels` do the sorting would be very inefficient in comparison to having the caller do it once

pkg/streamingpromql/testdata/ours/binary_operators.test
charleskorn: >that the load order needs to be preserved

Not sure I follow this part - what are you referring to by "load order" here?

jhesketh (approved):
We can create a wrapper method to do the sorting if needed. But granted the number of times is an issue (and the complexity to only pass `sort=true` once probably outweighs the benefit).

pkg/streamingpromql/testdata/ours/binary_operators.test
jhesketh: the order the series are loaded in (above on line 88) could affect this test right? It's possible they are loaded in a way where the series are already sorted after the on/etc.

charleskorn: It doesn't matter what order the series are in - the only thing that matters is the order of the label names passed to `BytesWithLabels` or `BytesWithoutLabels`.

(and regardless of the order of the series in the `load`, they'll be retrieved in sorted order due to how chunks streaming works)

PR #9492, opened 2024-10-02
kafka replay speed: handle more errors

dimitarvdimitrov:
#### What this PR does

Putting in this PR mostly so we collect explicit behaviour when particular errors happen and avoid the log lines `received an error we're not prepared to handle; this shouldn't happen; please report this as a bug`

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -24,6 +25,11 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+const (
+	unknownBrokerFranzGoErrorString    = "unknown broker"
+	chosenBrokerDiedFranzGoErrorString = "the internal broker struct chosen to issue this request has died--either the broker id is migrating or no longer exists"
```
gotjosh, line 30:
```suggestion
	errUnknownBroker    = "unknown broker"
	errBrokerDied = "the internal broker struct chosen to issue this request has died--either the broker id is migrating or no longer exists"
```

gotjosh: nit: a bit more concise, I don't think the long name add anything if we're documtning their usage in the `case` loop. The prefix is just for consistency and easily grep in the file.

gotjosh (approved):
LGTM

Issue #9382, opened 2024-09-23
Release 2.14


narqo:
This issue tracks the progress of the Mimir 2.14 release.

### Publish the release candidate

- [x] Begin drafting the [release notes](https://github.com/grafana/mimir/blob/main/RELEASE.md#write-release-notes-document) (#9406)
  - Create the release notes PR targeting the main branch
  - This step shouldn't block from publishing release candidate
  - After the release notes PR is merged (which usually happen after RC is published), cherry-pick them into the release branch
- [x] Wait for any open PR we want to get merged before cutting the release candidate
  - We shouldn't wait for the open PRs beyond the scheduled release date
- [x] Update `CHANGELOG.md`
  - [x] Run `./tools/release/check-changelog.sh LAST-RELEASE-TAG...main` and add missing PRs to CHANGELOG
  - [x] Ensure CHANGELOG entries are [sorted by type](https://github.com/grafana/mimir/blob/main/docs/internal/contributing/README.md#changelog)
  - [x] Add a new section for the new release so that `## main / unreleased` is blank and at the top. The new section should say `## x.y.0-rc.0` (https://github.com/grafana/mimir/pull/9424).
- [x] Run `./tools/release/notify-changelog-cut.sh CHANGELOG.md`
- [x] Run `make mixin-screenshots` (#9425)
  - Before opening the PR, review all updated screenshots and ensure no sensitive data is disclosed
- [x] Create new release branch
  - [x] Create the branch
    ```bash
    git checkout r<xxx> # xxx is the latest weekly release
    git checkout -b release-<version>
    git push -u origin release-<version>
    ```
  - [x] Remove "main / unreleased" section from the CHANGELOG
  - [x] If a new minor or major version is being released, adjust the settings in the `renovate.json` configuration on the `main` branch by adding the new version.
        This way we ensure that dependency updates maintain the new version, as well as the latest two minor versions.
        For instance, if versions 3.0 and 2.10 are configured in `renovate.json`, and version 3.1 is being released,
        during the release process `renovate.json` should keep updated the following branches: `main`, `release-3.1`, `release-3.0` and `release-2.10`.
- [x] Publish the Mimir release candidate
  - [x] Update VERSION in the release branch and update CHANGELOG with version and release date.
    - Keep in mind this is a release candidate, so the version string in VERSION and CHANGELOG must end in `-rc.#`, where `#` is the release candidate number, starting at 0.
[...]

PR #6718, opened 2023-11-23
Implement series limit using ingester own series

pstibrany:
#### What this PR does

This PR implements limiting of user series by only using series owned by ingesters to check the limit.

How it works:

- After ingester opens all local TSDBs, ingester will wait until it's "ACTIVE" in the ring. At that point, ingester will read tokens from the ring to see which series it actually "owns", and ingester will update "owned series" for all open TSDBs. After that, ingester starts accepting push requests.

- When new series is created, we assume that it was assigned to this ingester because ingester "owns" it, and simply increase "owned series" for TSDB.

- When series is deleted, we don't update "owned series" counter for TSDB. Deletion only happens during compaction, so we will instead recompute "owned series" for TSDB after compaction has finished.
  
- We recompute "owned series" for TSDB periodically for each user
  - After compaction -- series may have been deleted.
  - After ring changes -- "owned series" may be different now. Ring is checked for updates periodically.
  - After shard size has changed -- even if there was no change in the global ring, users' subring is now possibly different.
  - When new TSDB is opened while ingester already runs. We set shard size immediately, but only update owned series during next iteration. That seems good-enough, because we can assume that TSDB was opened because ingester owns all incoming series, so there's no rush to compute token ranges for such TSDB.

- How user limits are affected:
  - Instead of "number of in-memory series", we limit on number of "owned series".
  - Changes in tenants's shard size are not reflected immediately, but only after updating token ranges and "owned series".
    Until this happens, old shard size is used for limits computation.

Credits: Original code was created by @pr00se, I continue in his work before he returns.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9496, opened 2024-10-02
ingester: promote series limit using own series to stable

narqo:
#### What this PR does

This feature was introduced in #6718, and we have been running it enabled in Grafana Cloud since. I suggest we promote it to stable in the next release. What do you think @pstibrany @pr00se?

Relates to #9382 (as a post-2.14 clean up)

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pstibrany (approved):
Thank you. Perhaps we could remove the flags as well?

I'm also not sure if we have provided documentation for this feature.

pr00se (changes requested):
I think this still applies:

https://github.com/grafana/mimir/pull/8695#issuecomment-2225829240

PR #9500, opened 2024-10-02
kafka replay speed: fall back to franz-go for ongoing fetching

dimitarvdimitrov:
#### What this PR does

Striking the right configuration for ongoing fetching depends a lot on the characteristics of cluster. franz-go is better at adaptive concurrency, so we fall back to it until we've implemented adaptive concurrency and/or records-per-fetch.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -285,10 +285,6 @@ func (r *PartitionReader) switchToOngoingFetcher(ctx context.Context) {
 			r.kafkaCfg.Topic: {r.partitionID: kgo.NewOffset().At(lastConsumed + 1)},
 		})
 	}
-
-	if r.kafkaCfg.StartupFetchConcurrency == 0 && r.kafkaCfg.OngoingFetchConcurrency > 0 {
-		panic("this shouldn't have been allowed through config validation; we can only use concurrent fetchers for ongoing fetches if they're also used for startup fetches")
-	}
 }
 
 func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver prometheus.Observer) error {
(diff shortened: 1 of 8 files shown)


pkg/storage/ingest/reader_test.go
```diff
@@ -1440,6 +1440,125 @@ func TestPartitionReader_ConsumeAtStartup(t *testing.T) {
 			})
 		}
 	})
+
+	t.Run("should read target lag and then consume more records after switching to 0 ongoing concurrency if position=start, startup_fetch_concurrency=2, ongoing_fetch_concurrency=0", func(t *testing.T) {
```
dimitarvdimitrov, line 1444: this is effectively a copy of another test case above but with configured startup and ongoing concurrency


```
	t.Run("should consume partition from start if position=start, and wait until target lag is honored, and then consume some records after lag is honored", func(t *testing.T) {
```

pkg/storage/ingest/reader.go
```diff
@@ -246,6 +246,51 @@ func (r *PartitionReader) run(ctx context.Context) error {
...
+		// Resume from the next unconsumed offset.
+		r.client.AddConsumePartitions(map[string]map[int32]kgo.Offset{
+			r.kafkaCfg.Topic: {r.partitionID: kgo.NewOffset().At(lastConsumed + 1)},
+		})
+	}
+
+	if r.kafkaCfg.StartupFetchConcurrency == 0 && r.kafkaCfg.OngoingFetchConcurrency > 0 {
+		panic("this shouldn't have been allowed through config validation; we can only use concurrent fetchers for ongoing fetches if they're also used for startup fetches")
+	}
+}
```
gotjosh, line 292: I don't think we need this if we're already validating at a configuration level.

gotjosh (approved):

pkg/storage/ingest/reader.go
dimitarvdimitrov: 👍 removed in https://github.com/grafana/mimir/pull/9500/commits/de335b8b7929d4fc33360ef92d6cd7eaa204d0b0

PR #9501, opened 2024-10-02
ingest storage: don't return an error when records are successfully ingested

dimitarvdimitrov:
#### What this PR does

This fixes a flaky `TestPartitionReader_Commit` because we'd shut down the reader right after receiving the records. That would return an error as if ingesting caused an error. The bug was introduced in #9454 where we did the opposite change. The opposite change isn't functionally necessary.

Related to #8697 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


gotjosh (approved):
LGTM

PR #9514, opened 2024-10-03
Add warmup counter to EWMA rate

### LINKED ISSUE #5443: Add warmup to EwmaRate
#### Describe the bug

The current `math.EwmaRate` implementation does not require a warmup period, allowing a rate to be available after a single sample is added, which can lead to unexpected results, as described in #5394. Let's move warmup support from that PR into `EwmaRate` itself, to avoid problems with other users of this type.

slimjim777:
#### What this PR does
Adds a warm-up period to the EWMA rate so that the rate for the first N samples is returned as 0. The approach mirrors the approach in [VividCortex](https://github.com/VividCortex/ewma/blob/master/README.md), although there the recommendation is to initialize the moving rate as the mean of the first 10 samples.

Fixes #5443

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


slimjim777:
@jhalterman I took a shot at #5443 and the implementation is in between the VividCortex approach (with a counter) and the current approach with using the sliding window time. This returns 0 during the first 9 ticks, whereas the sliding window time does that for the first 60s. 

I amended the existing test for `math.EWMARate` to expect 0 for the first 9 samples, but I suspect that there other tests that will fail as a result of this change.

slimjim777:
Amended the warm-up sample count to be 60 so it's aligned with the sliding window time of 60s.

jhalterman:
Hi @slimjim777 Thanks for the PR and apologies for the delay! 

My main question is, can we make the warmup size configurable? We could create an EWMA with warmup via something like `NewEWMARateWithWarmup`.

slimjim777:
> My main question is, can we make the warmup size configurable? We could create an EWMA with warmup via something like `NewEWMARateWithWarmup`.

I’ll add that

slimjim777:
> Hi @slimjim777 Thanks for the PR and apologies for the delay!
> 
> My main question is, can we make the warmup size configurable? We could create an EWMA with warmup via something like `NewEWMARateWithWarmup`.

I've added a `NewEWMARateWithWarmup` method. 

(It could have been done with adding the [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) approach or a builder pattern, but I kept it as a separate constructor)

PR #9515, opened 2024-10-03
query-tee: Add an option to skip samples for comparison before a given timestamp

codesome:
#### What this PR does

Similar to how we have `-proxy.compare-skip-recent-samples`, this PR adds `-proxy.compare-skip-samples-before` to skip samples for comparison that are before the given unix timestamp. It is useful for block builders where you don't want to compare beyond a certain point.

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -39,6 +39,7 @@ type ProxyConfig struct {
 	UseRelativeError                    bool
 	PassThroughNonRegisteredRoutes      bool
 	SkipRecentSamples                   time.Duration
+	SkipSamplesBefore                   int64
 	RequireExactErrorMatch              bool
 	BackendSkipTLSVerify                bool
 	AddMissingTimeParamToInstantQueries bool
@@ -65,6 +66,7 @@ func (cfg *ProxyConfig) RegisterFlags(f *flag.FlagSet) {
 	f.Float64Var(&cfg.ValueComparisonTolerance, "proxy.value-comparison-tolerance", 0.000001, "The tolerance to apply when comparing floating point values in the responses. 0 to disable tolerance and require exact match (not recommended).")
 	f.BoolVar(&cfg.UseRelativeError, "proxy.compare-use-relative-error", false, "Use relative error tolerance when comparing floating point values.")
 	f.DurationVar(&cfg.SkipRecentSamples, "proxy.compare-skip-recent-samples", 2*time.Minute, "The window from now to skip comparing samples. 0 to disable.")
+	f.Int64Var(&cfg.SkipSamplesBefore, "proxy.compare-skip-samples-before", 0, "Skip the samples before the given unix timestamp in seconds for comparison.")
 	f.BoolVar(&cfg.RequireExactErrorMatch, "proxy.compare-exact-error-matching", false, "If true, errors will be considered the same only if they are exactly the same. If false, errors will be considered the same if they are considered equivalent.")
 	f.BoolVar(&cfg.PassThroughNonRegisteredRoutes, "proxy.passthrough-non-registered-routes", false, "Passthrough requests for non-registered routes to preferred backend.")
 	f.BoolVar(&cfg.AddMissingTimeParamToInstantQueries, "proxy.add-missing-time-parameter-to-instant-queries", true, "Add a 'time' parameter to proxied instant query requests if they do not have one.")
--- a/tools/querytee/response_comparator.go
+++ b/tools/querytee/response_comparator.go
@@ -40,6 +40,7 @@ type SampleComparisonOptions struct {
 	Tolerance              float64
 	UseRelativeError       bool
 	SkipRecentSamples      time.Duration
+	SkipSamplesBefore      int64 // unix time in milliseconds
 	RequireExactErrorMatch bool
 }
 
@@ -220,7 +221,7 @@ func compareMatrix(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
 		return err
 	}
 
-	if allMatrixSamplesWithinRecentSampleWindow(expected, queryEvaluationTime, opts) && allMatrixSamplesWithinRecentSampleWindow(actual, queryEvaluationTime, opts) {
+	if allMatrixSamplesOutsideComparableWindow(expected, queryEvaluationTime, opts) && allMatrixSamplesOutsideComparableWindow(actual, queryEvaluationTime, opts) {
 		return nil
 	}
 
@@ -250,6 +251,19 @@ func compareMatrix(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
 }
 
 func compareMatrixSamples(expected, actual *model.SampleStream, queryEvaluationTime time.Time, opts SampleComparisonOptions) error {
+	expected.Values = trimBeginning(expected.Values, func(p model.SamplePair) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+	actual.Values = trimBeginning(actual.Values, func(p model.SamplePair) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+	expected.Histograms = trimBeginning(expected.Histograms, func(p model.SampleHistogramPair) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+	actual.Histograms = trimBeginning(actual.Histograms, func(p model.SampleHistogramPair) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+
 	expectedSamplesTail, actualSamplesTail, err := comparePairs(expected.Values, actual.Values, func(p1 model.SamplePair, p2 model.SamplePair) error {
 		err := compareSamplePair(p1, p2, queryEvaluationTime, opts)
 		if err != nil {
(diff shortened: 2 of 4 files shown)


tools/querytee/proxy.go
```diff
@@ -39,6 +39,7 @@ type ProxyConfig struct {
 	UseRelativeError                    bool
 	PassThroughNonRegisteredRoutes      bool
 	SkipRecentSamples                   time.Duration
+	SkipSamplesBefore                   int64
```
charleskorn, line 42: Is there a particular reason you chose to use a Unix epoch timestamp rather than say a `time.Time`?

tools/querytee/response_comparator.go
```diff
@@ -391,10 +430,17 @@ func compareVector(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
...
 		return nil
 	}
 
+	expected = filterSlice(expected, func(p *model.Sample) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+	actual = filterSlice(actual, func(p *model.Sample) bool {
+		return int64(p.Timestamp) < opts.SkipSamplesBefore
+	})
+
```
charleskorn, line 443: Mutating the slices like this might lead to confusing error messsages later on. For example if `SkipSamplesBefore` is set to 1000, and the sample for a series has timestamp 1001 on one side and the corresponding sample on the other side has timestamp 999, then we'll return a `expected metric {...} missing from actual response` or `expected X metrics but got Y` error below but in reality we've filtered out one side.

A similar issue applies to the matrix comparison above as well.

tools/querytee/proxy.go
codesome: Only reason I can think is I was lazy to write the time parsing lines :). Maybe it is a good idea to accept the time as a `RFC3339` string.

tools/querytee/response_comparator.go
codesome: Giving the right error message will make this comparison a little more complex where we compare the filtered out part as well. I am not sure if I want to spend more time in making this part of error message better at the moment because I haven't heard of requiring this option other than block builder at the moment. Maybe put it as a TODO right now?

tools/querytee/proxy.go
```diff
@@ -66,11 +66,20 @@ func (cfg *ProxyConfig) RegisterFlags(f *flag.FlagSet) {
...
 	f.BoolVar(&cfg.RequireExactErrorMatch, "proxy.compare-exact-error-matching", false, "If true, errors will be considered the same only if they are exactly the same. If false, errors will be considered the same if they are considered equivalent.")
 	f.BoolVar(&cfg.PassThroughNonRegisteredRoutes, "proxy.passthrough-non-registered-routes", false, "Passthrough requests for non-registered routes to preferred backend.")
 	f.BoolVar(&cfg.AddMissingTimeParamToInstantQueries, "proxy.add-missing-time-parameter-to-instant-queries", true, "Add a 'time' parameter to proxied instant query requests if they do not have one.")
 	f.Float64Var(&cfg.SecondaryBackendsRequestProportion, "proxy.secondary-backends-request-proportion", 1.0, "Proportion of requests to send to secondary backends. Must be between 0 and 1 (inclusive), and if not 1, then -backend.preferred must be set.")
+
+	var skipSamplesBefore string
+	f.StringVar(&skipSamplesBefore, "proxy.compare-skip-samples-before", "", "Skip the samples before the given time for comparison. The time must be in RFC3339 format.")
```
charleskorn, line 75: dskit's `flagext.Time` should be able to handle most of this parsing logic for you.

tools/querytee/response_comparator.go
charleskorn: Sure, I'm OK with adding a comment explaining that it might produce confusing error messages.

tools/querytee/proxy.go
codesome: Done

tools/querytee/response_comparator.go
codesome: I added a comment and also made the error message a little better - I added a note in the error message whenever we filtered them out.

codesome:
@charleskorn I improved the error message a little bit (unit tests should also reflect that) and also added a comment about it when we filter the samples in the vector.

tools/querytee/response_comparator.go
```diff
@@ -391,10 +430,36 @@ func compareVector(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
...
+	})
+	actual = filterSlice(actual, func(p *model.Sample) bool {
+		return p.Timestamp < opts.SkipSamplesBefore
+	})
+	eChanged, aChanged := len(expected) != eOrgLen, len(actual) != aOrgLen
+	defer func() {
+		if retErr != nil {
+			warning := ""
+			if eChanged && aChanged {
+				warning = " (also, some samples were filtered out from the expected and actual response)"
```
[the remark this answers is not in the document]
codesome: Right now we are only filtering for 'skip samples before' and not changing the behaviour for skip recent samples, so I will update the warning accordingly

codesome:
Should be all good now!

tools/querytee/response_comparator.go
```diff
@@ -495,6 +561,9 @@ func compareScalar(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
 }
 
 func compareSamplePair(expected, actual model.SamplePair, queryEvaluationTime time.Time, opts SampleComparisonOptions) error {
+	if expected.Timestamp < opts.SkipSamplesBefore && actual.Timestamp < opts.SkipSamplesBefore {
+		return nil
+	}
```
pracucci, line 566: [nit] It took me a while to understand why we do this check before `if expected.Timestamp != actual.Timestamp {` so I think it's worth a comment.

```suggestion
	// If the timestamp is before the configured SkipSamplesBefore then we don't even check if the timestamp is correct.
	// The reason is that the SkipSamplesBefore feature may be used to compare queries hitting a different storage and one of two storages has no historical data.
	if expected.Timestamp < opts.SkipSamplesBefore && actual.Timestamp < opts.SkipSamplesBefore {
		return nil
	}
	
```

tools/querytee/response_comparator.go
```diff
@@ -391,10 +430,36 @@ func compareVector(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
...
+		if retErr != nil {
+			warning := ""
+			if eChanged && aChanged {
+				warning = " (also, some samples were filtered out from the expected and actual response due to the 'skip samples before')"
+			} else if aChanged {
+				warning = " (also, some samples were filtered out from the actual response due to the 'skip samples before')"
+			} else if eChanged {
+				warning = " (also, some samples were filtered out from the expected response due to the 'skip samples before')"
+			}
+			retErr = fmt.Errorf("%w%s", retErr, warning)
```
pracucci, line 459: [nit]
```suggestion
			if warning != "" {
				retErr = fmt.Errorf("%w%s", retErr, warning)
			}
```

tools/querytee/response_comparator.go
```diff
@@ -391,10 +430,36 @@ func compareVector(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
...
-	if allVectorSamplesWithinRecentSampleWindow(expected, queryEvaluationTime, opts) && allVectorSamplesWithinRecentSampleWindow(actual, queryEvaluationTime, opts) {
+	if allVectorSamplesOutsideComparableWindow(expected, queryEvaluationTime, opts) && allVectorSamplesOutsideComparableWindow(actual, queryEvaluationTime, opts) {
 		return nil
 	}
 
+	// Warning: filtering samples can give out confusing error messages. For example if the actual response had
+	// matching series, but all sample timestamps were before SkipSamplesBefore, while expected response had samples
+	// after SkipSamplesBefore: instead of saying mismatch, it will instead say that some metrics is missing, because
+	// we filter them here.
+	eOrgLen, aOrgLen := len(expected), len(actual)
```
pracucci, line 441: Can we do all of this only `if opts.SkipSamplesBefore > 0`?

tools/querytee/response_comparator.go
```diff
@@ -345,6 +359,29 @@ func comparePairs[S ~[]M, M any](s1, s2 S, compare func(M, M) error) (S, S, erro
...
+			break
+		}
+	}
+	return s[i:]
+}
+
+// filterSlice returns a new slice with elements from s removed that satisfy skip().
+func filterSlice[S ~[]M, M any](s S, skip func(M) bool) S {
+	var i int
+	var res S
```
pracucci, line 376: Maybe you can preallocate this slice. I expect in most cases we don't filter out or we filter out a minority of samples.

tools/querytee/response_comparator.go
```diff
@@ -391,10 +430,36 @@ func compareVector(expectedRaw, actualRaw json.RawMessage, queryEvaluationTime t
...
+	})
+	actual = filterSlice(actual, func(p *model.Sample) bool {
+		return p.Timestamp < opts.SkipSamplesBefore
+	})
+	eChanged, aChanged := len(expected) != eOrgLen, len(actual) != aOrgLen
+	defer func() {
+		if retErr != nil {
+			warning := ""
+			if eChanged && aChanged {
+				warning = " (also, some samples were filtered out from the expected and actual response due to the 'skip samples before')"
```
pracucci, line 453: [nit] I would being even more explicit. Same comment applies to other warning messages:

```suggestion
				warning = " (some samples were filtered out from the expected and actual response due to the 'skip samples before'; if all samples have been filtered out, this could cause the check on the expected number of metrics to fail)"
```

pracucci (approved):
LGTM. I left few minor comments I would be glad if you could take a look at.

I think there a case where comparison will still fail. The `compareMatrix()` checks `len(expected) != len(actual)` as first thing. This means that if the number of series is different, it will just fail. However, the number of series may be different just because one of the series has all samples before `SkipSamplesBefore` and so it's not returned by one of the 2 backends you're comparing.

I don't think it's worth blocking this PR merging (you can already get some value with these changes), but I've the feeling you will have to re-iterate on this.

tools/querytee/response_comparator.go
```diff
@@ -523,6 +592,10 @@ func compareSampleValue(first, second float64, opts SampleComparisonOptions) boo
 }
 
 func compareSampleHistogramPair(expected, actual model.SampleHistogramPair, queryEvaluationTime time.Time, opts SampleComparisonOptions) error {
+	if expected.Timestamp < opts.SkipSamplesBefore {
```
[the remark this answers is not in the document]
codesome: We should, good catch!

tools/querytee/response_comparator.go
codesome: Does not really make a difference but I can make that change

codesome:
Follow up of this PR is here https://github.com/grafana/mimir/pull/9660

PR #9518, opened 2024-10-04
MQE: fix panic in `rate` or `increase` over native histograms with changing schema

charleskorn:
#### What this PR does

This PR fixes a panic in MQE when `rate` or `increase` is used over a range of native histograms that change schema multiple times.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/functions/rate_increase.go
+++ b/pkg/streamingpromql/functions/rate_increase.go
@@ -84,9 +84,9 @@ func histogramRate(isRate bool, step types.RangeVectorStepData, hHead []promql.H
 		emitAnnotation(annotations.NewNativeHistogramNotCounterWarning)
 	}
 
-	currentSchema := firstPoint.H.Schema
-	if lastPoint.H.Schema < currentSchema {
-		currentSchema = lastPoint.H.Schema
+	initialSchema := firstPoint.H.Schema
+	if lastPoint.H.Schema < initialSchema {
+		initialSchema = lastPoint.H.Schema
 	}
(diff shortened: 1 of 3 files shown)


pkg/streamingpromql/functions/rate_increase.go
```diff
@@ -114,7 +114,7 @@ func histogramRate(isRate bool, step types.RangeVectorStepData, hHead []promql.H
 				return histogram.ErrHistogramsIncompatibleSchema
 			}
 
-			if p.H.Schema < currentSchema {
+			if p.H.Schema < delta.Schema {
```
jhesketh, line 117: I think we can avoid doing this so often. If we track the min schema, we only need to copy delta to the min schema once it has finished accumulating points (and /maybe/ on line 108 before the add (I'm not as sure about there, would need to check).

jhesketh (approved):
Nice find. In checking why Prometheus' engine didn't suffer this I think we can optimise where we are copying the schema. Granted it's probably not a frequent operation.

pkg/streamingpromql/functions/rate_increase.go
charleskorn: Very possibly, but I'd like to keep the scope of this PR as small as possible, so I'll do that in a follow-up PR.

charleskorn: https://github.com/grafana/mimir/pull/9531

PR #9517, opened 2024-10-03
WIP: Split blocks tool

andyasp:
#### What this PR does

This is a work in progress.

We'd like to have a tool that splits multi-day blocks into day blocks. This can be helpful when migrating from other systems.

#### Which issue(s) this PR fixes or relates to

Fixes: N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #5410, opened 2023-07-04
Doc: mention possibility to reduce TTL for cached results when uploading blocks

pstibrany:
#### What this PR does

This PR updates "Configure TSDB block upload" article to mention possibilty to decrease TTL for cache results.

#### Checklist

- [na] Tests updated
- [x] Documentation added
- [na] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #9524, opened 2024-10-04
Block upload: Verify block ranges of uploaded blocks

pstibrany:
#### What this PR does

Verify that backfilled blocks do not cover too big time range or cross the boundary of the max time range used by compactor.

Larger blocks can be split using tool from https://github.com/grafana/mimir/pull/9517.

#### Checklist

- [x] Tests updated.
- [na] Documentation added.
- [na] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.


56quarters (approved):
These changes LGTM but it'd probably be nice to add some documentation for this in a follow up PR

pstibrany:
> LGTM. I'm curious if there's a future where we can knowingly accept a block of a larger duration, then do the splitting transparently at the verify step. The partial failures for that probably make it too complex.

I considered that too. It may be an option in the future, if there's a demand for it.

pstibrany:
I've updated my old PR with section on block validation and documented new checks there as well: https://github.com/grafana/mimir/pull/5410

PR #9527, opened 2024-10-04
Update embedded query in query sharding to include a separate map

zenador:
#### What this PR does

Adds a separate map to embedded query struct in query sharding so that other kinds of sharding can use extra info without needing to extract it through regular expressions which are less reliable.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

Not a user-facing change.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/embedded.go
+++ b/pkg/frontend/querymiddleware/astmapper/embedded.go
@@ -71,7 +71,7 @@ func (c jsonCodec) Decode(encoded string) (queries []EmbeddedQuery, err error) {
 	return embedded.Concat, nil
 }
 
-// VectorSquash reduces an AST into a single vector query which can be hijacked by a Queryable impl.
+// VectorSquash reduces multiple EmbeddedQueries into a single vector query which can be hijacked by a Queryable impl.
 // It always uses a VectorSelector as the substitution expr.
 // This is important because logical/set binops can only be applied against vectors and not matrices.
 func VectorSquasher(exprs ...EmbeddedQuery) (parser.Expr, error) {
--- a/pkg/frontend/querymiddleware/astmapper/embedded.go
+++ b/pkg/frontend/querymiddleware/astmapper/embedded.go
@@ -33,7 +33,19 @@ const (
 
 // EmbeddedQueries is a wrapper type for encoding queries
 type EmbeddedQueries struct {
-	Concat []string `json:"Concat"`
+	Concat []EmbeddedQuery `json:"Concat"`
+}
+
+type EmbeddedQuery struct {
+	Expr   string            `json:"Expr"`
+	Params map[string]string `json:"Params,omitempty"`
+}
+
+func NewEmbeddedQuery(expr string, params map[string]string) EmbeddedQuery {
+	return EmbeddedQuery{
+		Expr:   expr,
+		Params: params,
+	}
 }
 
 // JSONCodec is a Codec that uses JSON representations of EmbeddedQueries structs
@@ -41,7 +53,7 @@ var JSONCodec jsonCodec
 
 type jsonCodec struct{}
 
-func (c jsonCodec) Encode(queries []string) (string, error) {
+func (c jsonCodec) Encode(queries []EmbeddedQuery) (string, error) {
 	embedded := EmbeddedQueries{
 		Concat: queries,
 	}
@@ -49,7 +61,7 @@ func (c jsonCodec) Encode(queries []string) (string, error) {
 	return string(b), err
 }
 
-func (c jsonCodec) Decode(encoded string) (queries []string, err error) {
+func (c jsonCodec) Decode(encoded string) (queries []EmbeddedQuery, err error) {
 	var embedded EmbeddedQueries
 	err = json.Unmarshal([]byte(encoded), &embedded)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -26,11 +26,13 @@ func NewSharding(shardSummer ASTMapper) ASTMapper {
 	)
 }
 
-type Squasher = func(...parser.Expr) (parser.Expr, error)
+type Squasher = func(...EmbeddedQuery) (parser.Expr, error)
 
 type ShardLabeller interface {
 	GetLabelName() string
 	GetLabelValue(shard int) string
```
charleskorn, line 33: Do we still need these as part of the `ShardLabeller` interface if we have `GetLabelMatcher`?

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -489,12 +499,13 @@ func (summer *shardSummer) shardAndSquashAggregateExpr(expr *parser.AggregateExp
 		// Create the child expression, which runs the given aggregation operation
 		// on a single shard. We need to preserve the grouping as it was
 		// in the original one.
-		children = append(children, &parser.AggregateExpr{
+		var aggExpr parser.Expr = &parser.AggregateExpr{
```
charleskorn, line 502: [nit] Can we just do this?

```suggestion
		aggExpr := &parser.AggregateExpr{
```

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -542,12 +553,13 @@ func (summer *shardSummer) shardAndSquashBinOp(expr *parser.BinaryExpr) (parser.
 			return nil, err
 		}
 
-		children = append(children, &parser.BinaryExpr{
+		var binExpr parser.Expr = &parser.BinaryExpr{
```
charleskorn, line 556: [nit] Same as above, can we use this simpler form?

```suggestion
		binExpr := &parser.BinaryExpr{
```

charleskorn (commented):
Overall LGTM

pkg/frontend/querymiddleware/astmapper/sharding.go
zenador: True that they're not actually needed anymore with `GetLabelMatcher`, I just wanted to keep them because I thought it was clearer, but I can remove them if you think that'd be better.

pkg/frontend/querymiddleware/astmapper/embedded.go
```diff
@@ -62,14 +74,8 @@ func (c jsonCodec) Decode(encoded string) (queries []string, err error) {
...
 // This is important because logical/set binops can only be applied against vectors and not matrices.
-func VectorSquasher(exprs ...parser.Expr) (parser.Expr, error) {
-	// concat OR legs
-	strs := make([]string, 0, len(exprs))
-	for _, expr := range exprs {
-		strs = append(strs, expr.String())
-	}
-
-	encoded, err := JSONCodec.Encode(strs)
+func VectorSquasher(exprs ...EmbeddedQuery) (parser.Expr, error) {
```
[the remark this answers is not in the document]
zenador: Thanks, updated, do you think it's okay now?

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: If we only use `GetLabelMatcher` now, then I would remove the other methods.

pkg/frontend/querymiddleware/astmapper/instant_splitting_test.go
```diff
@@ -281,59 +281,59 @@ func TestInstantSplitter(t *testing.T) {
 		// Offset operator
 		{
 			in:                   `rate({app="foo"}[3m] offset 3m)`,
-			out:                  `sum without() (__embedded_queries__{__queries__="{\"Concat\":[\"increase({app=\\\"foo\\\"}[1m] offset 5m)\",\"increase({app=\\\"foo\\\"}[1m] offset 4m)\",\"increase({app=\\\"foo\\\"}[1m] offset 3m)\"]}"}) / 180`,
+			out:                  `sum without() (__embedded_queries__{__queries__="{\"Concat\":[{\"Expr\":\"increase({app=\\\"foo\\\"}[1m] offset 5m)\"},{\"Expr\":\"increase({app=\\\"foo\\\"}[1m] offset 4m)\"},{\"Expr\":\"increase({app=\\\"foo\\\"}[1m] offset 3m)\"}]}"}) / 180`,
```
dimitarvdimitrov, line 284: Won't this change the format that's being sent over the network between queriers and query-frontends? Would old queriers be able to work with new query-frontends or will queries fail?

_I thought i left this comment already the other day, but apparently I didn't 🦑_

zenador: Good point, I guess queriers and query-frontends must both be on the new version? @charleskorn WDYT?

charleskorn: No: this format is only used within query-frontends. `shardedQuerier` will receive the `EmbeddedQuery` and send the embedded expression (from `Expr`) to queriers. 

So queriers will receive exactly the same thing as before, so there should be no backwards or forwards compatibility issues.

charleskorn (approved):

PR #9531, opened 2024-10-07
MQE: only reduce resolution of delta in `rate` and `increase` once

charleskorn:
#### What this PR does

This PR modifies the behaviour of `rate` and `increase` to only reduce the resolution of the computed delta once, rather than every time the schema reduces in the input range.

This has no noticeable performance impact in our benchmarks as we don't exercise this case, as it rarely happens in practice.

Follow up to https://github.com/grafana/mimir/pull/9518#discussion_r1787164808

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/functions/rate_increase.go
```diff
@@ -114,8 +114,8 @@ func histogramRate(isRate bool, step types.RangeVectorStepData, hHead []promql.H
 				return histogram.ErrHistogramsIncompatibleSchema
 			}
 
-			if p.H.Schema < delta.Schema {
-				delta = delta.CopyToSchema(p.H.Schema)
+			if p.H.Schema < desiredSchema {
+				desiredSchema = p.H.Schema
```
jhesketh, line 118: do we need to copy delta to the desired schema at each counter reset (ie before adding the previous value)?

jhesketh (commented):
Thanks for following this up!

pkg/streamingpromql/functions/rate_increase.go
charleskorn: No - `Add` handles the case where the histograms have different schemas and will return a histogram with the lower of the two schemas (ie. the lower resolution).

jhesketh: Thanks, forgot about that.

jhesketh (approved):

PR #9532, opened 2024-10-07
MQE: Add support for label_replace

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -257,9 +257,9 @@ func TestNewInstantQuery_Strings(t *testing.T) {
 
 	q, err = prometheusEngine.NewInstantQuery(ctx, storage, nil, expr, time.Now())
 	require.NoError(t, err)
-	promethues := q.Exec(context.Background())
+	prometheus := q.Exec(context.Background())
 
-	testutils.RequireEqualResults(t, expr, promethues, mimir)
+	testutils.RequireEqualResults(t, expr, prometheus, mimir)
 }
 
 // This test runs the test cases defined upstream in https://github.com/prometheus/prometheus/tree/main/promql/testdata and copied to testdata/upstream.
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -61,7 +61,6 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
 
 	// These expressions are also unsupported, but are only valid as instant queries.
 	unsupportedInstantQueryExpressions := map[string]string{
-		"'a'":             "string value as top-level expression",
 		"metric{}[5m:1m]": "PromQL expression type *parser.SubqueryExpr for range vectors",
 	}
 
@@ -239,6 +238,28 @@ func TestNewRangeQuery_InvalidExpressionTypes(t *testing.T) {
 
 	_, err = engine.NewRangeQuery(ctx, nil, nil, `"thing"`, time.Now(), time.Now(), time.Second)
 	require.EqualError(t, err, "query expression produces a string, but expression for range queries must produce an instant vector or scalar")
+}
+
+func TestNewInstantQuery_Strings(t *testing.T) {
+	ctx := context.Background()
+	opts := NewTestEngineOpts()
+	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), log.NewNopLogger())
+	require.NoError(t, err)
+
+	prometheusEngine := promql.NewEngine(opts.CommonOpts)
+
+	storage := promqltest.LoadedStorage(t, ``)
+
+	expr := `"thing"`
+	q, err := mimirEngine.NewInstantQuery(ctx, storage, nil, expr, time.Now())
+	require.NoError(t, err)
+	mimir := q.Exec(context.Background())
+
+	q, err = prometheusEngine.NewInstantQuery(ctx, storage, nil, expr, time.Now())
+	require.NoError(t, err)
+	promethues := q.Exec(context.Background())
+
+	testutils.RequireEqualResults(t, expr, promethues, mimir)
 }
 
 // This test runs the test cases defined upstream in https://github.com/prometheus/prometheus/tree/main/promql/testdata and copied to testdata/upstream.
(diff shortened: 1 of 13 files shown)


pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -233,3 +233,42 @@ load 1m
 
 eval range from 0 to 7m step 1m -metric
   {} -1 -1 -1 -3 -3 _ -4 -4
+
+clear
+
+# label_replace() tests
```
charleskorn, line 239: Could you please add a test for the case where different series should merge into one after relabelling? Something like this:

```
load 1m
  series{label="a", idx="1"} 2 _
  series{label="a", idx="2"} _ 4

eval range from 0 to 1m step 1m label_replace(series, "idx", "replaced", "idx", ".*")
  series{label="a", idx="replaced"} 2 4
```

This should catch the missing `DeduplicateAndMerge`.

charleskorn (commented):
LGTM, thanks for working on this.

I'd be curious to see the benchmark results for `label_replace` in comparison to Prometheus' engine.

pkg/streamingpromql/testdata/ours/functions.test
jhesketh: This is fixed in this branch now, but has a related upstream problem https://github.com/prometheus/prometheus/issues/15114

pkg/streamingpromql/engine_test.go
```diff
@@ -241,6 +240,28 @@ func TestNewRangeQuery_InvalidExpressionTypes(t *testing.T) {
...
 
+func TestNewInstantQuery_Strings(t *testing.T) {
+	ctx := context.Background()
+	opts := NewTestEngineOpts()
+	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), log.NewNopLogger())
+	require.NoError(t, err)
+
+	prometheusEngine := promql.NewEngine(opts.CommonOpts)
+
+	storage := promqltest.LoadedStorage(t, ``)
```
charleskorn, line 251: It should be possible to drop this and pass `nil` to `NewInstantQuery` instead - the storage won't be used because there are no selectors.

pkg/streamingpromql/functions.go
```diff
@@ -141,6 +141,55 @@ func scalarToInstantVectorOperatorFactory(args []types.Operator, _ *limiting.Mem
...
+			return nil, fmt.Errorf("expected a string for 5th argument for label_replace, got %T", args[4])
+		}
+
+		f := functions.FunctionOverInstantVector{
+			SeriesDataFunc:           functions.PassthroughData,
+			SeriesMetadataFunc:       functions.LabelReplaceFactory(dstLabel, replacement, srcLabel, regex),
+			NeedsSeriesDeduplication: true,
+		}
+
+		var o types.InstantVectorOperator = operators.NewFunctionOverInstantVector(inner, memoryConsumptionTracker, f, expressionPosition)
```
charleskorn, line 187: [nit]

```suggestion
		o := operators.NewFunctionOverInstantVector(inner, memoryConsumptionTracker, f, expressionPosition)
```

pkg/streamingpromql/query.go
```diff
@@ -502,7 +522,15 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
 		} else {
 			q.result = &promql.Result{Value: q.populateMatrixFromScalarOperator(d)}
 		}
-
+	case parser.ValueTypeString:
+		if q.IsInstant() {
```
charleskorn, line 526: We don't bother checking this for range vectors above - I don't think this is necessary here, as it should be caught in `newQuery` above.

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -233,3 +233,68 @@ load 1m
...
+eval instant at 0m label_replace(floor(series), "__name__", "$1", "label", "(.*)")
+		a{label="a"} 1
+		b{label="b"} 2
+		c{label="c"} NaN
+
+# Label name is dropped by floor after it is set
+eval instant at 0m floor(label_replace(series, "__name__", "$1", "label", "(.*)"))
+		{label="a"} 1
+		{label="b"} 2
+		{label="c"} NaN
```
charleskorn, line 277: [nit] Indentation of the series is a bit off here, might be tabs instead of spaces or something like that

pkg/streamingpromql/engine_test.go
jhesketh: This is true for MQE, but Prometheus panic's without it. (And since this isn't a real scenario, I didn't dig deeper and just passed the storage into both engines).

pkg/streamingpromql/functions.go
jhesketh: This is needed because `NewFunctionOverInstantVector` returns `*FunctionOverInstantVector`

pkg/streamingpromql/engine_test.go
charleskorn: Weird, alright, let's leave it then.

pkg/streamingpromql/functions.go
charleskorn: Not sure I follow sorry - that sounds like it shouldn't be an issue, and making the change locally seems to work fine.

pkg/streamingpromql/query.go
charleskorn: Let's leave it for now.

pkg/streamingpromql/functions.go
jhesketh: really? It makes line 55 below incompatible:

``grafana/mimir/pkg/streamingpromql/functions.go:55:8: cannot use operators.NewDeduplicateAndMerge(o, memoryConsumptionTracker) (value of type *operators.DeduplicateAndMerge) as *operators.FunctionOverInstantVector value in assignment
```

charleskorn: I think you're looking at the wrong method, line 55 is `SingleInputVectorFunctionOperatorFactory` :) But I'll approve this to unblock you.

charleskorn (approved):

pkg/streamingpromql/functions.go
jhesketh: Whoops, thanks, I absolutely was. ETOOMANYFILETABSOPEN. Since it's still waiting to merge I pushed it onto here.

PR #9547, opened 2024-10-07
Validate only a single tenant ID passed to ruler API

56quarters:
#### What this PR does

This change extends an existing test to validate the ruler API endpoints only accept a single tenant ID, not multiple tenant IDs joined via `|` pipes.

N/A

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/api_test.go
+++ b/pkg/ruler/api_test.go
@@ -1614,7 +1614,7 @@ func TestAlertStateDescToPrometheusAlert(t *testing.T) {
 	})
 }
 
-func TestAPIRoutesCorrectlyHandleInvalidOrgID(t *testing.T) {
+func TestAPIRoutesCorrectlyHandleInvalidTenantID(t *testing.T) {
 	tcs := []struct {
 		route  string
 		method string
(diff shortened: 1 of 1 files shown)


pkg/ruler/api_test.go
```diff
@@ -1630,32 +1630,33 @@ func TestAPIRoutesCorrectlyHandleInvalidOrgID(t *testing.T) {
...
-		router.Path("/config/v1/rules/{namespace}").Methods(http.MethodPost).HandlerFunc(a.CreateRuleGroup)
-		router.Path("/config/v1/rules/{namespace}/{groupName}").Methods(http.MethodDelete).HandlerFunc(a.DeleteRuleGroup)
-		router.Path("/config/v1/rules/{namespace}").Methods(http.MethodDelete).HandlerFunc(a.DeleteNamespace)
-
-		req := requestFor(t, tc.method, "https://localhost:8080"+tc.route, nil, invalidOrgID)
-
-		w := httptest.NewRecorder()
-		router.ServeHTTP(w, req)
+		for _, tenantID := range []string{"", "user1|user2"} {
+			t.Run(fmt.Sprintf("method=%s, route=%s, userID=%s", tc.method, tc.route, tenantID), func(t *testing.T) {
```
narqo, line 1634: > `userID=%s", ... tenantID`

(nit)

1. the test is called `...HandleInvalidOrgID`
2. we define tenant IDs
3. we name a sub-test `userID=blah`

I know that it's a historical mess that we use `OrgID`, `TenantID` and `UserID` for the same thing. But, I wonder if we could keep those names scoped within one test function.

narqo (approved):
Works for me 🔥

_I don't know if we want to clean the org-tenant-user mess here._

PR #9563, opened 2024-10-08
Ruler: Add pagination for api/v1/rules

fayzal-g:
Adds the parameters `group_limit` and `group_next_token` to `api/v1/rules` based on https://github.com/prometheus/prometheus/pull/14017

### DIFF
--- a/docs/sources/mimir/references/http-api/index.md
+++ b/docs/sources/mimir/references/http-api/index.md
@@ -864,6 +864,8 @@ The `file`, `rule_group` and `rule_name` parameters are optional, and can accept
 
 The `exclude_alerts` parameter is optional. If set, it only returns rules and excludes active alerts.
 
+The `group_limit` and `group_next_token` parameters are optional. If `group_limit` is set, it will limit the number of rule groups returned in a single response. If the total number of rule groups exceeds this value, the response will contain a `groupNextToken`.
+This can be passed into subsequent requests via `group_next_token` to paginate over the remaining groups. The final response will not contain a token.
 For more information, refer to Prometheus [rules](https://prometheus.io/docs/prometheus/latest/querying/api/#rules).
 
 Requires [authentication](#authentication).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 * [ENHANCEMENT] Ingester: Disable the push circuit breaker when ingester is in read-only mode. #9760
 * [ENHANCEMENT] Ingester: Reduced lock contention in the `PostingsForMatchers` cache. #9773
 * [ENHANCEMENT] Storage: Allow HTTP client settings to be tuned for GCS and Azure backends via an `http` block or corresponding CLI flags. This was already supported by the S3 backend. #9778
+* [ENHANCEMENT] Ruler: Support `group_limit` and `group_next_token` parameters in the `<prometheus-http-prefix>/api/v1/rules` endpoint. #9563
 * [BUGFIX] Fix issue where functions such as `rate()` over native histograms could return incorrect values if a float stale marker was present in the selected range. #9508
 * [BUGFIX] Fix issue where negation of native histograms (eg. `-some_native_histogram_series`) did nothing. #9508
 * [BUGFIX] Fix issue where `metric might not be a counter, name does not end in _total/_sum/_count/_bucket` annotation would be emitted even if `rate` or `increase` did not have enough samples to compute a result. #9508
--- a/integration/e2emimir/client.go
+++ b/integration/e2emimir/client.go
@@ -718,7 +718,7 @@ func (c *Client) GetPrometheusRules(maxGroups int, token string) ([]*promv1.Rule
 		Status string `json:"status"`
 		Data   struct {
 			RuleGroups []*promv1.RuleGroup `json:"groups"`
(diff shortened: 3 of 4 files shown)


pkg/ruler/ruler.go
```diff
@@ -993,7 +995,31 @@ func (r *Ruler) GetRules(ctx context.Context, req RulesRequest) ([]*GroupStateDe
...
+	if req.MaxGroups > 0 {
+		slices.SortFunc(merged, func(a, b *GroupStateDesc) int {
+			nsCmp := strings.Compare(a.Group.Namespace, b.Group.Namespace)
+			if nsCmp != 0 {
+				return nsCmp
+			}
+
+			// If Namespaces are equal, check the group names
+			return strings.Compare(a.Group.Name, b.Group.Name)
+		})
```
stevesg, line 1010: Non-blocking nit: this might be over-engineering..... but I see a potential issue where someone updates the sort strategy in one place, and not the other (in api.go). Could we share them somehow? Define a "SortableGroup" interface or something?

pkg/ruler/ruler.go
```diff
@@ -1068,18 +1094,27 @@ func (r *Ruler) getLocalRules(ctx context.Context, userID string, req RulesReque
...
 
-		interval := group.Interval()
-
 		// The mapped filename is url path escaped encoded to make handling `/` characters easier
 		decodedNamespace, err := url.PathUnescape(strings.TrimPrefix(group.File(), prefix))
 		if err != nil {
 			return nil, errors.Wrap(err, "unable to decode rule filename")
 		}
+
+		if req.NextToken != "" && !foundToken {
```
stevesg, line 1109:
```suggestion
		// If a pagination token is provided, skip past groups until we reach the point of that token.
		if req.NextToken != "" && !foundToken {
```

pkg/ruler/ruler.go
```diff
@@ -1068,18 +1094,27 @@ func (r *Ruler) getLocalRules(ctx context.Context, userID string, req RulesReque
 	groupSet := makeStringFilterSet(req.RuleGroup)
 	ruleSet := makeStringFilterSet(req.RuleName)
 
+	foundToken := false
 	for _, group := range groups {
```
stevesg, line 1098: The rules are sorted at this point, right? From `GetRules`?

docs/sources/mimir/references/http-api/index.md
```diff
@@ -864,6 +864,8 @@ The `file`, `rule_group` and `rule_name` parameters are optional, and can accept
 
 The `exclude_alerts` parameter is optional. If set, it only returns rules and excludes active alerts.
 
+The `group_limit` and `group_next_token` parameters are optional. If `group_limit` is set, it will limit the number of rule groups returned in a single response. If the total number of rule groups exceeds this value, the response will contain a `groupNextToken`.
```
tacole02, line 867:
```suggestion
The `group_limit` and `group_next_token` parameters are optional. If `group_limit` is set, it limits the number of rule groups returned in a single response. If the total number of rule groups exceeds this value, the response contains a value of `groupNextToken`.
```

tacole02: We avoid using "will" in docs in favor of present tense.

docs/sources/mimir/references/http-api/index.md
```diff
@@ -864,6 +864,8 @@ The `file`, `rule_group` and `rule_name` parameters are optional, and can accept
...
+This can be passed into subsequent requests via `group_next_token` to paginate over the remaining groups. The final response will not contain a token.
```
tacole02, line 868:
```suggestion
You can pass this value into subsequent requests via the `group_next_token` parameter to paginate over the remaining groups. The final response doesn't contain a token.
```

tacole02 (approved):
LGTM 😄

integration/ruler_test.go
```diff
@@ -204,6 +205,138 @@ func TestRulerAPISingleBinary(t *testing.T) {
...
+	// Check the returned rules match the rules written
+	require.Len(t, returnedGroups, len(expectedGroups))
+	for i := 0; i < len(expectedGroups); i++ {
+		require.Equal(t, expectedGroups[i].Namespace, returnedGroups[i].Namespace)
+		require.Equal(t, expectedGroups[i].Group, returnedGroups[i].Group)
+	}
+
+	// Invalid max groups value
+	_, _, err = c.GetPrometheusRules(-1, "")
+	require.Error(t, err)
```
stevesg, line 337: We should check this returns 4xx, regressions of this type are quite common (returning 5xx instead of 4xx), so having it covered in the integration test is valuable.

stevesg (approved):
Nice! I only have one suggestion to the integration test, but assuming you fix that I don't need to review this again.

docs/sources/mimir/references/http-api/index.md
aknuds1: Should these two new parameters be added to the endpoint definition itself? I.e., `GET <prometheus-http-prefix>/api/v1/rules?type={alert|record}&file={}&rule_group={}&rule_name={}&exclude_alerts={true|false}`.

aknuds1 (approved):
LGTM, but see question please.

gotjosh (approved):
LGTM

PR #7175, opened 2024-01-19
Add support for ingesting out-of-order native histograms

fionaliao:
This PR adds OOO native histogram functionality to Mimir.

The core OOO logic is within prometheus and was synced to mimir-prometheus in https://github.com/grafana/mimir-prometheus/pull/703. This PR just adds a flag to Mimir for enabling out-of-order native histograms and some tests. 

#### Which issue(s) this PR fixes or relates to

Fixes #9421 

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [x] [`about-versioning.md`](/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9567, opened 2024-10-08
bugfix: When OOO native histograms are disabled it should be a client error

gotjosh:
#### What this PR does

When ingesters replay data from Kafka, if they receive a non-client error, they will continue to attempt to push the batch to the TSDB.

In this PR, I'm translating Native Histogram OOO errors into client errors to ensure that we can continue with the next set of write requests when replaying data.

This was introduced as part of https://github.com/grafana/mimir/pull/7175

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1739,8 +1739,7 @@ When `-ingester.error-sample-rate` is configured to a value greater than `0`, in
 ### err-mimir-native-histogram-ooo-disabled
 
 This non-critical error occurs when Mimir receives a write request that contains a sample that is a native histogram
-where another sample with a more recent timestamp has already been ingested and 
-`-ingester.ooo-native-histograms-ingestion-enabled` is set to `false`.
+where another sample with a more recent timestamp has already been ingested and `-ingester.ooo-native-histograms-ingestion-enabled` is set to `false`.
 
 {{< admonition type="note" >}}
 The series containing such samples are skipped during ingestion, and valid series within the same request are ingested.


pkg/ingester/ingester.go
```diff
@@ -1344,6 +1344,18 @@ func (i *Ingester) pushSamplesToAppender(userID string, timeseries []mimirpb.Pre
 			return true
 
 		// Map TSDB native histogram validation errors to soft errors.
+		case errors.Is(err, storage.ErrOOONativeHistogramsDisabled):
+			stats.invalidNativeHistogramCount++
```
gotjosh, line 1348: I'm not entirely sure this counts as an invalid native histogram. Technically, the histogram could have correct we just don't support accepting them.

pkg/ingester/ingester.go
```diff
@@ -1344,6 +1344,18 @@ func (i *Ingester) pushSamplesToAppender(userID string, timeseries []mimirpb.Pre
...
+			updateFirstPartial(i.errorSamplers.nativeHistogramValidationError, func() softError {
+				return newNativeHistogramValidationError(globalerror.NativeHistogramOOODisabled, err, model.Time(timestamp), labels)
+			})
+			return true
+		case errors.Is(err, storage.ErrNativeHistogramsDisabled):
```
gotjosh, line 1353: I was not able to replicate this failure in a test, is this handled gracefully somewhere else?

krajorama: yes, we don't even try to ingest them if they are disabled, so no error, this is not needed: see later:
```
if nativeHistogramsIngestionEnabled {
```

pkg/ingester/ingester.go
krajorama: Agree this is not invalid, rather `stats.sampleOutOfOrderCount++`

pkg/ingester/ingester_test.go
```diff
@@ -1915,6 +1916,77 @@ func TestIngester_Push(t *testing.T) {
 				cortex_ingester_tsdb_head_max_timestamp_seconds 0.01
 			`,
 		},
+		"should not fail if native histograms are disabled": {
```
krajorama, line 1919: There's a dedicated test already, but this is fine as well TestIngesterCanEnableIngestAndQueryNativeHistograms

gotjosh: Nice to know - I had this one already so decided to stick with it.

pkg/ingester/ingester.go
gotjosh: good shout, done in https://github.com/grafana/mimir/pull/9567/commits/911760a898e2a029d472577254bd2e764e3f0ca7

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1736,6 +1736,19 @@ The series containing such samples are skipped during ingestion, and valid serie
 When `-ingester.error-sample-rate` is configured to a value greater than `0`, invalid native histogram errors are logged only once every `-ingester.error-sample-rate` times.
 {{< /admonition >}}
 
+### err-mimir-native-histogram-ooo-disabled
+
+This non-critical error occurs when Mimir receives a write request that contains a sample that is a native histogram
+where another sample with a more recent timestamp has already been ingested.
```
fionaliao, line 1742:
```suggestion
This non-critical error occurs when Mimir receives a write request that contains a sample that is a native histogram
where another sample with a more recent timestamp has already been ingested and 
`-ingester.ooo-native-histograms-ingestion-enabled` is set to `false`.
```

fionaliao (approved):
LGTM

I'm having a look at why the [OOO integration test](https://github.com/grafana/mimir/blob/ab3e627b1da134f33c06dd09ee4402a2eaad68f1/integration/ooo_ingestion_test.go#L207-L212) didn't catch this issue. I think it's a problem with the test rather than the code, so I think that can be investigated separately.

krajorama (approved):

PR #9580, opened 2024-10-10
Skip stale postings in `/active_series`

flxbk:
#### What this PR does

When iterating over postings to build the response for `/active_series`, it's not guaranteed that all series references will still be valid at iteration time. We should not error when hitting a stale reference but just build a response that does not include the series referenced by the stale posting.

Thanks to @bboreham for pointing this out.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


ortuman (approved):
approved. do you think it's worth adding or modifying any test to validate this behavior?

PR #9582, opened 2024-10-10
bugfix: Stop the current fetcher before replacing it

gotjosh:
#### What this PR does

Otherwise, this causes a memory leak as we leave lingering around the resources we used to catch up as quickly as possible as the ingester was starting.

image

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


dimitarvdimitrov (approved):
whoo-hooo 🚀 nice find!

PR #9507, opened 2024-10-03
MQE: add support for `and`

charleskorn:
#### What this PR does

This PR adds support for the `and` binary operation.

Compared to Prometheus' engine, latency is up to 90% lower and peak memory consumption is up to 48% lower in our benchmarks:

```
                                                                       │  Prometheus   │                Mimir                │
                                                                       │    sec/op     │    sec/op     vs base               │
Query/a_1_and_b_1{l=~'.*[0-4]$'},_instant_query-10                        723.5µ ±  0%   717.8µ ±  1%        ~ (p=0.310 n=6)
Query/a_1_and_b_1{l=~'.*[0-4]$'},_range_query_with_100_steps-10           744.2µ ±  2%   734.0µ ±  4%        ~ (p=0.818 n=6)
Query/a_1_and_b_1{l=~'.*[0-4]$'},_range_query_with_1000_steps-10          1.269m ±  1%   1.130m ±  1%  -10.93% (p=0.002 n=6)
Query/a_100_and_b_100{l=~'.*[0-4]$'},_instant_query-10                    1.613m ± 11%   1.609m ±  2%        ~ (p=0.485 n=6)
Query/a_100_and_b_100{l=~'.*[0-4]$'},_range_query_with_100_steps-10       3.270m ±  2%   2.367m ± 10%  -27.62% (p=0.002 n=6)
Query/a_100_and_b_100{l=~'.*[0-4]$'},_range_query_with_1000_steps-10     17.321m ±  1%   8.414m ±  0%  -51.42% (p=0.002 n=6)
Query/a_2000_and_b_2000{l=~'.*[0-4]$'},_instant_query-10                  17.08m ±  1%   16.97m ±  2%        ~ (p=0.589 n=6)
Query/a_2000_and_b_2000{l=~'.*[0-4]$'},_range_query_with_100_steps-10     51.59m ±  4%   30.85m ±  1%  -40.20% (p=0.002 n=6)
Query/a_2000_and_b_2000{l=~'.*[0-4]$'},_range_query_with_1000_steps-10    392.2m ±  1%   146.9m ±  9%  -62.55% (p=0.002 n=6)
Query/a_1_and_b_1{l='notfound'},_instant_query-10                         232.3µ ±  2%   217.7µ ± 12%        ~ (p=0.065 n=6)
Query/a_1_and_b_1{l='notfound'},_range_query_with_100_steps-10            251.5µ ±  7%   218.5µ ±  6%  -13.12% (p=0.002 n=6)
[...]

PR #9588, opened 2024-10-11
MQE: don't reuse `FloatHistogram` instances for successive points in instant vector selectors

charleskorn:
#### What this PR does

This PR is an iteration on https://github.com/grafana/mimir/pull/8677. It changes the behaviour of instant vector selectors to not return the same `FloatHistogram` instance for successive points if they all select the same histogram.

This preserves the main optimisation of #8677 (avoiding creating copies of histograms in `sum`) without the complexity and associated bugs of needing to ensure that shared `FloatHistogram` instances are modified safely.

For example, I just found yet another issue related to this that was introduced in https://github.com/grafana/mimir/pull/9507, where shared `FloatHistogram` instances could be left in the unused portion of the histogram slice [here](https://github.com/grafana/mimir/blob/14bc627207af501e2a2b9fad17d0ea22ff76039b/pkg/streamingpromql/operators/and_unless_binary_operation.go#L299-L308) and then corrupted if the `HPoint` slice was later used for a range vector selector. 

I believe similar corruption could occur if a `HPoint` slice was used for an instant vector selector that returned the same `FloatHistogram` for successive points, and then that slice was reused for a range vector selector: the range vector selector would reuse the `FloatHistogram` instances when filling the buffer [here](https://github.com/grafana/mimir/blob/14bc627207af501e2a2b9fad17d0ea22ff76039b/pkg/streamingpromql/operators/range_vector_selector.go#L125-L126) and so all points would have the same value as the last histogram, rather than their expected values.

The performance impact of this change is minimal in our benchmarks: less than 1% degradation or less than 5% reduction in latency in all cases. In some cases things get slightly slower due to the extra copies created, and in some cases things get slightly faster due to the simpler logic. I think the simpler logic and far reduced likelihood of bugs make this change worthwhile.

In the real world, where lookback is more common, we'll only need to make copies (and therefore pay the price of those copies) if the same point is used for successive output samples, which would usually only happen if the query step is smaller than the scrape interval. Given most users run with 15s or 60s scrape intervals, and most queries tend to be either instant queries, I don't think this is likely to become a huge issue.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
   * `cortex_alertmanager_state_replication_failed_total`
   * `cortex_alertmanager_alerts`
   * `cortex_alertmanager_silences`
-* [FEATURE] Querier: add experimental streaming PromQL engine, enabled with `-querier.query-engine=mimir`. #9367 #9368 #9398 #9399 #9403 #9417 #9418 #9419 #9420 #9482 #9504 #9505 #9507 #9518 #9531 #9532 #9533 #9553
+* [FEATURE] Querier: add experimental streaming PromQL engine, enabled with `-querier.query-engine=mimir`. #9367 #9368 #9398 #9399 #9403 #9417 #9418 #9419 #9420 #9482 #9504 #9505 #9507 #9518 #9531 #9532 #9533 #9553 #9588
 * [FEATURE] Query-frontend: added experimental configuration options `query-frontend.cache-errors` and `query-frontend.results-cache-ttl-for-errors` to allow non-transient responses to be cached. When set to `true` error responses from hitting limits or bad data are cached for a short TTL. #9028
 * [FEATURE] gRPC: Support S2 compression. #9322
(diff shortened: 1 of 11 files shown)


jhesketh (approved):
I agree with the rationale of this. I do wonder though if we can measure the real-world impact. Should we have a metric on number of copied samples? We can see if that rises too high if we should reconsider reimplementing such an optimisation down the road (although it may be a bigger risk to do later)

charleskorn:
> Should we have a metric on number of copied samples?

We should be able to see this on profiles, including the continuous profiling we run on Mimir instances in Grafana Cloud.

jhesketh (approved):
Makes sense to me. Are we adequately testing the series merging to catch the latest clear?

charleskorn:
> Makes sense to me. Are we adequately testing the series merging to catch the latest clear?

I've added a test to ensure `clear` is called when we expect, and the existing tests should be sufficient to catch any other regressions.

PR #9592, opened 2024-10-11
Increase block-ranges-period for OOO native histogram integration test

fionaliao:
#### What this PR does

This is a follow up to https://github.com/grafana/mimir/pull/9567 - that PR spotted an issue where the OOO native histograms disabled error was not being mapped to a client error.

Prior to that PR, we already had an integration test that checked that trying to ingest OOO NH samples when the OOO NH flag was disabled returned a 400 and that test actually passed. The reason why it passed was that another 400 error being returned (out of bounds) - this is as `-blocks-storage.tsdb.block-ranges-period`, which is used to determine if a sample is out of bounds, is set to a very low value in tests. I've increased this value the out of bounds check passes and the OOO NH disabled error is hit instead.

Note that the test expectations haven't changed - we check OOO NH samples error with a 400, but the underlying error is now the OOO NH disabled error (I checked by running the test and looking at logs). The current integration test Mimir client doesn't return the response body so we can only access the status code. The bugfix PR (https://github.com/grafana/mimir/pull/9567) has a unit test that checks the specific error is returned. It's still useful to modify `-blocks-storage.tsdb.block-ranges-period` just in case in the future we accidentally start mapping the OOO NH error incorrectly again.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/integration/ooo_ingestion_test.go
+++ b/integration/ooo_ingestion_test.go
@@ -166,8 +166,9 @@ func TestOOOHistogramIngestionDisabled(t *testing.T) {
 		BlocksStorageFlags(),
 		BlocksStorageS3Flags(),
 		map[string]string{
-			"-ingester.out-of-order-time-window":            "10m",
-			"-ingester.native-histograms-ingestion-enabled": "true",
+			"-ingester.out-of-order-time-window":                "10m",
+			"-ingester.native-histograms-ingestion-enabled":     "true",
+			"-ingester.ooo-native-histograms-ingestion-enabled": "false",
 			// Default block-ranges-period for integration tests is 1m
 			// When OOO NH is disabled, all NH samples must be greater than or equal to minValidTime, otherwise an out
 			// of bounds error is returned. minValidTime which max sample time - (block-ranges-period[0]/2). Increase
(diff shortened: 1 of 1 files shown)


integration/ooo_ingestion_test.go
```diff
@@ -168,6 +168,12 @@ func TestOOOHistogramIngestionDisabled(t *testing.T) {
 		map[string]string{
 			"-ingester.out-of-order-time-window":            "10m",
 			"-ingester.native-histograms-ingestion-enabled": "true",
+			// Default block-ranges-period for integration tests is 1m
```
krajorama, line 171: nit: let's add `"-ingester.ooo-native-histograms-ingestion-enabled": "false"`  while we're at it. Just in case we change the default/remove the flag, make the test fail more explicitly.

krajorama (approved):
nice catch, approved with a nit

integration/ooo_ingestion_test.go
fionaliao: Updated in https://github.com/grafana/mimir/pull/9592/commits/aeb921c607493208bd9ae8d68f267dc448584c1e

krajorama (approved):

PR #9642, opened 2024-10-16
MQE: Fix iterating over scalar values in Clamp

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/functions/math.go
```diff
@@ -68,9 +68,25 @@ var Clamp InstantVectorSeriesFunction = func(seriesData types.InstantVectorSerie
...
+	minSampleIdx := 0
+	maxSampleIdx := 0
+
+	for _, data := range seriesData.Floats {
+		for data.T > minArg.Samples[minSampleIdx].T {
+			minSampleIdx++
+		}
+		for data.T > maxArg.Samples[maxSampleIdx].T {
+			maxSampleIdx++
+		}
```
charleskorn, line 86: If we had access to the `QueryTimeRange`, could we call `timeRange.PointIndex()` here? Might be a bit clearer than these loops, and I imagine a little faster too. 

We should also only need to do this once - given both sets of scalar data have points at every step, we should be able to use the same index for both.

jhesketh: I have another change which brings in `QueryTimeRange` into `InstantVectorSeriesFunction`, so that's a good idea for a followup if that's okay.

Yep, good point, we probably only need to track one index. I was wondering though, do you think we need any protections around this? eg to make sure we don't go out of range, or where the timestamps may not match. In theory those things should never happen, but this feels unsafe. Thoughts?

charleskorn: >Yep, good point, we probably only need to track one index. I was wondering though, do you think we need any protections around this? eg to make sure we don't go out of range, or where the timestamps may not match. In theory those things should never happen, but this feels unsafe. Thoughts?

The Go runtime will protect us against going out of range, and if the timestamps don't match then we have a bug somewhere else - the contract for `ScalarData` is that it contains a point for every step in the query. 

It's possible these invariants are violated, but there are a bunch of similar invariants we rely on in other places that aren't checked either, so I'm OK with not checking them here too.

charleskorn: >so that's a good idea for a followup if that's okay.

Sure, that's fine - let's get this and #9641 merged then come back to this.

charleskorn (approved):

Issue #6931, opened 2023-12-13
Cross-tenant queries produce cardinality estimation cache keys that are too long


chencs:
#### Describe the bug
When running cross-tenant queries with the query-frontend results cache enabled, the query caching middleware attempts to concatenate _all_ tenant IDs into the key ([code](https://github.com/grafana/mimir/blob/efa26e743027bf495c340a893939b2671ba05959/pkg/frontend/querymiddleware/split_and_cache.go#L173)). There's a 250-byte limit on memcached keys ([code](https://github.com/grafana/gomemcache/blob/7de47a8c3cb08c073e448e0a066e4f0e1a283928/memcache/memcache.go#L616)), so it's possible to have enough tenants (or long enough tenant names) that the key ends up exceeding that 250-byte limit, at which point gets will fail.

#### To Reproduce

1. Create multiple tenants whose names, concatenated, exceed 250 bytes in length.
2. Query across all tenants using [the wildcard tenant](https://grafana.com/blog/2021/08/26/whats-new-in-grafana-enterprise-metrics-1.5-per-tenant-usage-metrics-and-a-wildcard-tenant-for-queries/).
3. See malformed key logs on memcached fetches.

#### Expected behavior

Keys should be short enough that they don't exceed the 250-byte limit. Perhaps that means that cache keys should be created and looked up on a per-tenant basis?

56quarters:
One potential solution to this is to sort the tenant IDs (the parsing logic might already do this) and hash them and use the hash in the cache key instead of the tenant IDs directly.

dimitarvdimitrov:
the cache key is then hashed and that's used to key memcached. The full key is stored as part of the item to check for hash collisions, but it's not included verbatim in the cache key.

https://github.com/grafana/mimir/blob/efa26e743027bf495c340a893939b2671ba05959/pkg/frontend/querymiddleware/split_and_cache.go#L458

chencs:
[...]

PR #7043, opened 2024-01-03
Ensure cardinality estimate cache keys are valid

56quarters:
#### What this PR does

Ensure that cache keys generated from cardinality estimate middleware are less than 250 bytes in length by hashing the tenant IDs that are included in them. This is especially evident when using tenant federation where multiple tenant IDs in a request.

#### Which issue(s) this PR fixes or relates to

Fixes #6931

#### Checklist

- [X] Tests updated.
- [NA] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9644, opened 2024-10-16
Use CacheKeyGenerator implementation in error caching middleware

56quarters:
#### What this PR does

Use an implementation of the CacheKeyGenerator interface instead of bare function to generate cache keys for error caching. This allows downstream consumers (GEM) to inject custom logic for caching based on LBAC rules.

Related #9028

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/error_caching.go
+++ b/pkg/frontend/querymiddleware/error_caching.go
@@ -5,7 +5,6 @@ package querymiddleware
 import (
 	"context"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/go-kit/log"
@@ -26,7 +25,7 @@ const (
 	reasonNotCacheableError = "not-cacheable-api-error"
 )
 
-func newErrorCachingMiddleware(cache cache.Cache, limits Limits, shouldCacheReq shouldCacheFn, logger log.Logger, reg prometheus.Registerer) MetricsQueryMiddleware {
+func newErrorCachingMiddleware(cache cache.Cache, limits Limits, shouldCacheReq shouldCacheFn, keyGen CacheKeyGenerator, logger log.Logger, reg prometheus.Registerer) MetricsQueryMiddleware {
 	cacheLoadAttempted := promauto.With(reg).NewCounter(prometheus.CounterOpts{
 		Name: "cortex_frontend_query_error_cache_requests_total",
 		Help: "Number of requests that check the results cache for an error.",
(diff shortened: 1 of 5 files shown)


pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -85,7 +86,7 @@ func (e *errorCachingHandler) Do(ctx context.Context, request MetricsQueryReques
 	}
 
 	e.cacheLoadAttempted.Inc()
-	key := errorCachingKey(tenant.JoinTenantIDs(tenantIDs), request)
+	key := e.keyGen.QueryRequestError(ctx, tenant.JoinTenantIDs(tenantIDs), request)
```
chencs, line 89: I know this isn't a _change_, but since this is all relatively new, is this key going to be subject to the same cache key length issue described in #6931? Should we get ahead of it here?

56quarters: The "key" here isn't actually used directly, it's hashed and the base64 (I think) representation of the hashed key is used. The key is _also_ stored in the cached payload to handle hash collisions. I've been meaning to do something similar to resolve #6931

chencs: > it's hashed and the base64 (I think) representation of the hashed key is used

Was the blocking comment Marco raised [here](https://github.com/grafana/mimir/pull/7043#pullrequestreview-1803908171) specifically around hashing tenant IDs in isolation (as opposed to hashing the entire key)? I'm trying to understand why his concern doesn't extend to this/the results cache.

56quarters: Yes, it's a problem when only the hashed key is used. Because this includes the full key as part of the payload there's no chance of cache collisions.

pkg/frontend/querymiddleware/error_caching.go
```diff
@@ -85,7 +86,7 @@ func (e *errorCachingHandler) Do(ctx context.Context, request MetricsQueryReques
...
 	hashedKey := cacheHashKey(key)
 
 	if cachedErr := e.loadErrorFromCache(ctx, key, hashedKey, spanLog); cachedErr != nil {
```
56quarters, line 92: This method uses the `hashedKey` to fetch the value from cache and then compares the stored `key` to generated one.

chencs (approved):

PR #9650, opened 2024-10-16
Block-builder-scheduler initial structure

seizethedave:
#### What this PR does
Adds initial structure for `block-builder-scheduler` living at pkg/blockbuilder/scheduler. Adds a new `block-builder-scheduler` target to the binary. This target currently just connects to Kafka and records per-partition start and end offset gauge metrics. In the future it will compute jobs and assign them to block-builder workers.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/scheduler/config.go
```diff
@@ -0,0 +1,36 @@
...
+	BuilderConsumerGroup   string        `yaml:"builder_consumer_group"`
+	SchedulerConsumerGroup string        `yaml:"scheduler_consumer_group"`
+	KafkaMonitorInterval   time.Duration `yaml:"kafka_monitor_interval"`
+
+	// Config parameters defined outside the block-builder-scheduler config and are injected dynamically.
+	Kafka ingest.KafkaConfig `yaml:"-"`
+}
+
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.BuilderConsumerGroup, "block-builder-scheduler.builder-consumer-group", "block-builder", "The Kafka consumer group used by block-builders.")
```
pracucci, line 23: Not sure what this will be used for. I didn't expect block builders to have a consumer group at all, given all offsets should be tracked by the scheduler. I suggest to remove this for now, and re-introduce it in a later PR if it will be effectively used (if so, why?).

pkg/blockbuilder/scheduler/config.go
```diff
@@ -0,0 +1,36 @@
...
+	f.StringVar(&cfg.SchedulerConsumerGroup, "block-builder-scheduler.scheduler-consumer-group", "block-builder-scheduler", "The Kafka consumer group used by block-builder-scheduler.")
+	f.DurationVar(&cfg.KafkaMonitorInterval, "block-builder-scheduler.kafka-monitor-interval", 20*time.Second, "How frequently to monitor the Kafka partitions.")
```
pracucci, line 25: WDYT about scheduling interval? At the end it's how frequently the scheduler triggers to check if there's more work to do, no?

```suggestion
	f.DurationVar(&cfg.KafkaMonitorInterval, "block-builder-scheduler.scheduling-interval", 20*time.Second, "How frequently to monitor the Kafka partitions.")
```

pkg/blockbuilder/scheduler/config.go
```diff
@@ -0,0 +1,36 @@
...
+	SchedulerConsumerGroup string        `yaml:"scheduler_consumer_group"`
+	KafkaMonitorInterval   time.Duration `yaml:"kafka_monitor_interval"`
+
+	// Config parameters defined outside the block-builder-scheduler config and are injected dynamically.
+	Kafka ingest.KafkaConfig `yaml:"-"`
+}
+
+func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&cfg.BuilderConsumerGroup, "block-builder-scheduler.builder-consumer-group", "block-builder", "The Kafka consumer group used by block-builders.")
+	f.StringVar(&cfg.SchedulerConsumerGroup, "block-builder-scheduler.scheduler-consumer-group", "block-builder-scheduler", "The Kafka consumer group used by block-builder-scheduler.")
```
pracucci, line 24: Consider removing `scheduler-` prefix if ends up we'll not need the builder consumer groups too.

```suggestion
	f.StringVar(&cfg.SchedulerConsumerGroup, "block-builder-scheduler.consumer-group", "block-builder-scheduler", "The Kafka consumer group used by block-builder-scheduler.")
```

pkg/blockbuilder/scheduler/config.go
```diff
@@ -0,0 +1,36 @@
...
+	f.DurationVar(&cfg.KafkaMonitorInterval, "block-builder-scheduler.kafka-monitor-interval", 20*time.Second, "How frequently to monitor the Kafka partitions.")
+}
+
+func (cfg *Config) Validate() error {
+	if err := cfg.Kafka.Validate(); err != nil {
+		return err
+	}
+	if cfg.KafkaMonitorInterval <= 0 {
+		return fmt.Errorf("kafka_monitor_interval (%d) must be positive", cfg.KafkaMonitorInterval)
```
pracucci, line 33: > kafka_monitor_interval

We're used to either reference CLI flags (if so just define it in a const and reuse it both here and in RegisterFlags()) or just reference to it as a human readable setting (e.g. "The Kafka monitor interval ...")

pkg/blockbuilder/scheduler/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+func newSchedulerMetrics(reg prometheus.Registerer) schedulerMetrics {
+	return schedulerMetrics{
+		monitorPartitionsDuration: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+			Name: "cortex_blockbuilder_scheduler_monitor_partitions_duration_seconds",
+			Help: "Time spent monitoring partitions.",
+
+			NativeHistogramBucketFactor: 1.1,
+		}),
+		partitionStartOffsets: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "cortex_blockbuilder_scheduler_partition_start_offsets",
```
pracucci, line 25: Given it's 1 offset per partition, I would remove the final "s". Some comment applies to the end offset metric.

```suggestion
			Name: "cortex_blockbuilder_scheduler_partition_start_offset",
```

pkg/blockbuilder/scheduler/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+
+type schedulerMetrics struct {
+	monitorPartitionsDuration prometheus.Histogram
+	partitionStartOffsets     *prometheus.GaugeVec
+	partitionEndOffsets       *prometheus.GaugeVec
+}
+
+func newSchedulerMetrics(reg prometheus.Registerer) schedulerMetrics {
+	return schedulerMetrics{
+		monitorPartitionsDuration: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
```
pracucci, line 18: In this PR at the moment this metric measures the time it takes to fetch the offsets, but what's your longer term idea? If you will just track the offsets fetching timing with this metric, then just make it explicit (e.g. `cortex_blockbuilder_scheduler_fetch_offsets_duration_seconds`) otherwise if you plan to track the entire scheduling I would already name it accordingly (e.g. `cortex_blockbuilder_scheduler_scheduling_duration_seconds` ... or something better you can come up with).

pracucci (approved):
The skeleton LGTM! I left few minor comments. No need for me to re-review it.

pkg/blockbuilder/scheduler/config.go
seizethedave: Ah yes. This is something @codesome and I discussed: in the v1 of the scheduler, we'll just have the block-builder workers continue to checkpoint offsets to Kafka. For 2 reasons:

1. get v1 out quicker
2. less risky for backlogs: if there is a large backlog that takes 6 hours to consume, we'll get smaller section-granularity checkpoints rather than everything riding on that 6 hour job to complete.

(But it's mostly about reason number 1.) Let me know your thoughts.

pkg/blockbuilder/scheduler/metrics.go
seizethedave: Ah, yeah. This method will do all of:

1. fetch all necessary quantities from Kafka to make scheduling decisions. (Still have a couple more to add.)
2. update the schedule

Number 1 is the interesting part, timing-wise, but the timing covers both. Number 2 is going to be microseconds. To tie in your other good comment, this method will be renamed to something more like updateSchedule and the metric can be paired as `scheduling_duration`.

PR #9651, opened 2024-10-17
MQE: Add support for round

jhesketh:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -20,6 +20,7 @@ type InstantVectorFunctionOperatorFactory func(
 	memoryConsumptionTracker *limiting.MemoryConsumptionTracker,
 	annotations *annotations.Annotations,
 	expressionPosition posrange.PositionRange,
+	timeRange types.QueryTimeRange,
 ) (types.InstantVectorOperator, error)
 
 type ScalarFunctionOperatorFactory func(
@@ -37,7 +38,7 @@ type ScalarFunctionOperatorFactory func(
 //   - name: The name of the function
 //   - f: The function implementation
 func SingleInputVectorFunctionOperatorFactory(name string, f functions.FunctionOverInstantVector) InstantVectorFunctionOperatorFactory {
-	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange) (types.InstantVectorOperator, error) {
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, _ types.QueryTimeRange) (types.InstantVectorOperator, error) {
 		if len(args) != 1 {
 			// Should be caught by the PromQL parser, but we check here for safety.
 			return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", name, len(args))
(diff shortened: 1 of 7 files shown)


jhesketh:
(note that I'll update the point indexes in a separate change because I want discuss an implementation specific)

pkg/streamingpromql/testdata/ours-only/native_histograms.test
```diff
@@ -13,3 +13,9 @@ eval range from 0 to 5m step 1m clamp_min(mixed_metric, 2)
 
 eval range from 0 to 5m step 1m clamp_max(mixed_metric, 2)
 	{} _ 1 2 2
+
+# round ignores any histograms
+# Prometheus currently returns min instead of no value as per the documentation
```
charleskorn, line 18: What does 'min' refer to here?

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -436,3 +436,21 @@ eval range from 0 to 54m step 6m clamp_min(series, scalar(mins))
...
+  toNearest      1    _    3 4 0.1
+  series{a="b"}  -5.5 2.75 _ _ 6.87
+
+eval range from 0 to 24m step 6m round(series, scalar(toNearest))
+  series{a="b"}  -5   NaN  _ _ 6.9
+
+clear
+
+load 6m
+  toNearest      1    2     0.5   10    0.1   0.25    100       0.0001  5          1000      NaN _   _   NaN _ 2 2
```
charleskorn, line 452: What happens if the argument is negative?

pkg/streamingpromql/functions.go
```diff
@@ -251,6 +252,42 @@ func ClampMinMaxFunctionOperatorFactory(functionName string, isMin bool) Instant
...
+				// Should be caught by the PromQL parser, but we check here for safety.
+				return nil, fmt.Errorf("expected a scalar for 2nd argument for round, got %T", args[1])
+			}
+		} else {
+			toNearest = operators.NewScalarConstant(float64(1), timeRange, memoryConsumptionTracker, expressionPosition)
+		}
+
+		f := functions.FunctionOverInstantVector{
+			SeriesDataFunc: functions.Round,
+			// TODO(jhesketh): With the currently vendored prometheus, round does not consistently drop the __name__ label
```
charleskorn, line 281: Does not consistently drop the `__name__` label, or doesn't drop it at all?

Is there an issue link for this?

pkg/streamingpromql/testdata/ours-only/native_histograms.test
jhesketh: A typo.

pkg/streamingpromql/testdata/ours/functions.test
jhesketh: Added test. (Basically the way round is implemented makes it take the absolute value)

pkg/streamingpromql/functions.go
jhesketh: This is tricky. Against main, round does not return any __name__ (as evidence by the tests here https://github.com/prometheus/prometheus/pull/15176).

Against what we have vendored it does have the __name__. So the bug has seemingly since been fixed. I have not found an exact issue but assume it is related to the more recent delayed label name dropping work fixing something incidentally.

Also worth noting is that in the vendored version it has `DropName: true,`, so I'm not sure if that isn't being honoured somewhere (I haven't tracked down where). 

The only upstream tests of round are in the aggregations.test file, such as:
```
eval instant at 50m round(0.005 * http_requests{group="production",job="api-server"})
	{group="production", instance="0", job="api-server"} 1
	{group="production", instance="1", job="api-server"} 1
```
Which due to the multiplication has the resulting label dropped.

So it wouldn't surprise me if it went unnoticed upstream and also fixed as part of the label work (also unnoticed).

pkg/streamingpromql/testdata/ours-only/native_histograms.test
charleskorn: Looks like the same typo is earlier in the file as well, above the tests for `clamp`

pkg/streamingpromql/functions.go
charleskorn: Right, thanks for the explanation.

> Also worth noting is that in the vendored version it has `DropName: true,`, so I'm not sure if that isn't being honoured somewhere (I haven't tracked down where).

`DropName` is only used when the experimental new delayed `__name__` dropping is enabled, and this is disabled by default.


I would rephrase the comment slightly to make it clearer, something like this perhaps:

```suggestion
// TODO(jhesketh): With the version of Prometheus vendored at the time of writing, round does not drop the __name__ label, and this is verified by our tests.
//                 We match this behaviour for consistency, but this behaviour has changed in Prometheus 3.0, so we'll need to match that once prometheus 3.0 is vendored in.
```

charleskorn (approved):

PR #9659, opened 2024-10-17
[mixin/alerts]: Enable configuring job prefix for alerts to prevent clashes with metrics from Loki/Tempo.

### LINKED ISSUE #5260: PrometheusAlerts should filter by namespace (conflicts with cortex)
Since mimir / cortex share the same metric names, it's possible for say a Mimir alert to fire that is actually related to Cortex (and visa-versa). We've seen this on other grafana components too, such as Tempo.

There are many examples here: https://github.com/grafana/mimir/blob/main/operations/mimir-mixin-compiled/alerts.yaml

In our case, `MimirRulerTooManyFailedQueries` was firing but it was actually Cortex that was causing it (we are doing a migration).

---

I would propose that we filter these queries by namespace.

mtweten:
#### What this PR does
This enables the ability to configure a job prefix for alerts in the mimir-mixin. Currently, some alerts just use a regex that matches any job prefix (e.g. `.+/` or `.*/`). This can cause a problem because certain alerts will start triggering because they will match metrics for not just mimir, but also loki/tempo (if present).

As a specific example, `GossipMembersTooHigh` and `RingMembersMismatch` will start firing if Tempo metrics are being collected, since the current alerts will match not only on `job=mimir/ingester`, but also `job=tempo/ingester` 

Typically, users would configure this alert prefix to the namespace mimir is deployed to, e.g. "mimir/".

This is a non-impactful change as it's a new configuration option which defaults to the previous prefix, `.*/`. However, it does change GossipMembersTooHigh from the prefix `.+/` to `.*/`, but this shouldn't be relevant. Every other alert using this pattern was already using `.*/` for the job prefix.

Fixes #5260

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


56quarters (approved):
LGTM, thanks!

PR #9660, opened 2024-10-17
query-tee: Filter samples for matrix before checking number of series

codesome:
#### What this PR does

This is a follow up of https://github.com/grafana/mimir/pull/9515

It fixes the case where if a series was completely removed from matrix and if the remaining part would match, we would still fail comparison saying there is a mismatch in series.

Fixes #<issue number>

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/querytee/response_comparator_test.go
+++ b/tools/querytee/response_comparator_test.go
@@ -2192,6 +2192,19 @@ func TestCompareSamplesResponse(t *testing.T) {
 			skipSamplesBefore: 95 * 1000,
 		},
 		{
+			name: "fail when all series from expected are filtered out - float",
+			expected: json.RawMessage(`{
+							"status": "success",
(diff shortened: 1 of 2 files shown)


tools/querytee/response_comparator_test.go
```diff
@@ -2179,6 +2179,18 @@ func TestCompareSamplesResponse(t *testing.T) {
 						}`),
 			skipSamplesBefore: 95 * 1000,
 		},
+		{
+			name: "should not fail when some series is entirely skipped and rest matches - float",
```
charleskorn, line 2183: Could you please also add a test for the case where a series is entirely skipped in one response, but some samples are not skipped in the other response?

eg. if `skipSamplesBefore` is 95, then series A has samples at 90 and 93 in the expected response, and samples at 90, 93 and 100 in the actual response

codesome: Done! I also added warning to the error message like we are doing for vector responses.

charleskorn (approved):

PR #9661, opened 2024-10-17
kafka replay speed: dont send requests for higher than the high watermark

dimitarvdimitrov:
### What this PR does

This is related to a problem we have: high e2e latency when fetching at the head of the partition. The current theory is that WS introduces extra delay when requesting an offset which doesn't exist.

Currently, we have two sets of flags: one for startup and one for ongoing fetching. This PR enables us to have just a single configuration which determines the max concurrency. We will limit concurrency by never requesting an offset which doesn't exist.

#### Startup tests

I tested using the same concurrency and records-per-fetch settings for startup and ongoing fetching:

```
    'ingest-storage.kafka.startup-fetch-concurrency': '12',  
    'ingest-storage.kafka.startup-records-per-fetch': '1000',  
    'ingest-storage.kafka.ongoing-fetch-concurrency': '12',  
    'ingest-storage.kafka.ongoing-records-per-fetch': '1000',  
```

results looks pretty good:
1. cold replay speed remains high (variability is likely due to different data density in different partitions)
2. lag right after startup is on par with franz-go (previously our implementation struggled to get end-to-end latency below 2s and keep it there).
3. end-to-end latency while running is also on par with franz-go at ~400ms (first half of the screenshots)

| |  |
|--------|--------|
| Image | Image | 
|  Image | | 

#### Kafka outage tests

This also unlocks recovering from a Kafka outage. I tried to simulate that by scaling down the WS agents to from 40 replicas to 1. This ended up slowing down the path. The ingesters recovered with similar replay speed (admitedly the simulated "outage" was too short to be able to properly measure replay speed because the recovery was too fast)

| |  |
|--------|--------|
| Image | Image |

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -605,7 +605,7 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 		dispatchNextWant := chan fetchWant(nil)
 		if nextResult == nil || nextFetch.startOffset <= highWatermark.Load() {
 			// In Warpstream fetching past the end induced more delays than MinBytesWaitTime.
-			// So we dispatch a fetch for only if it's fetching an existing offset.
+			// So we dispatch a fetch only if it's fetching an existing offset.
 			// This shouldn't noticeably affect performance with Apache Kafka, after all franz-go only has a concurrency of 1 per partition.
 			//
 			// At the same time we don't want to reach a deadlock where the HWM is not updated and there are no fetches in flight.
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -21,6 +21,7 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 	"github.com/twmb/franz-go/pkg/kmsg"
 	"github.com/twmb/franz-go/plugin/kotel"
+	"go.uber.org/atomic"
 
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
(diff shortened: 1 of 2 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -587,13 +602,23 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 			// So we don't try to get new results from the fetchers.
 			refillBufferedResult = nil
 		}
+		dispatchNextWant := chan fetchWant(nil)
+		if nextResult == nil || nextFetch.startOffset <= highWatermark.Load() {
+			// In Warpstream fetching past the end induced more delays than MinBytesWaitTime.
+			// So we dispatch a fetch for only if it's fetching an existing offset.
```
pracucci, line 608: [nit]
```suggestion
			// So we dispatch a fetch only if it's fetching an existing offset.
```

pkg/storage/ingest/fetcher.go
```diff
@@ -500,7 +501,9 @@ func (r *concurrentFetchers) run(ctx context.Context, wants chan fetchWant, logg
 			if f.Err != nil {
 				w = handleKafkaFetchErr(f.Err, w, errBackoff, r.startOffsets, r.client, attemptSpan)
 			}
-
+			if hwm := f.HighWatermark; hwm >= 0 {
+				casHWM(highWatermark, hwm)
+			}
```
pracucci, line 506: Wouldn't be safer to do it only if `f.Err == nil`? However, `casHWM()` ensures that we always cas to the highest value, and in case there was an error and `f.HighWatermark` is the zero value, then it's probably not a big deal.

pracucci (approved):
I'm not super familar with this code. To my understand, the improvement done in this PR is to not dispatch a "want" (essentially a read from the partition) if there's already at least 1 in-flight read and the next fetch will start from an offset which is > than the last known HWM. In these regards, the change done LGTM 👍

pkg/storage/ingest/fetcher.go
dimitarvdimitrov: i wanted to always use the value from the response if it's there even if there's an error. It's not set if it's -1 (the library ensures this IIRC). And as you said a 0 isn't a big deal (we do fetches a couple of times a second, so an additional atomic load doesn't sound like a lot of overhead

PR #9662, opened 2024-10-17
Isolate tree internals into their own package

chencs:
#### What this PR does
This PR moves the `MultiQueuingAlgorithmTreeQueue` and implementations of `QueuingAlgorithm` into their own `tree` package. It also splits `tenantQuerierAssignments` into two types:
1. `TenantQuerierQueuingAlgorithm`, which contains a map of queriers assigned to tenant IDs, and implements `QueuingAlgorithm`.
1. `tenantQuerierAssignments`, which holds tenant information required to handle shuffle-sharding, and computes querier assignment across tenants. It holds a reference to `TenantQuerierQueuingAlgorithm`, updating the tenant-querier map when the assignments need recomputing.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/scheduler/queue/tree/multi_queuing_algorithm_tree_queue_test.go
+++ b/pkg/scheduler/queue/tree/multi_queuing_algorithm_tree_queue_test.go
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 
-package queue
+package tree
 
 import (
 	"fmt"
@@ -21,7 +21,6 @@ func Test_NewTree(t *testing.T) {
 	tests := []struct {
 		name      string
 		treeAlgos []QueuingAlgorithm
-		state     *tenantQuerierAssignments
 		expectErr bool
 	}{
 		{
@@ -30,7 +29,7 @@ func Test_NewTree(t *testing.T) {
 		},
 		{
 			name:      "create tenant-querier tree",
-			treeAlgos: []QueuingAlgorithm{&tenantQuerierAssignments{}},
+			treeAlgos: []QueuingAlgorithm{&TenantQuerierQueuingAlgorithm{}},
 		},
 		{
 			name:      "successfully create a tree with no queuing algorithms", // creates a single node with a queue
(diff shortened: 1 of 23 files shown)


chencs:
> But I do not really think splitting TQA is necessary. In terms of splitting tree queue out and treating it like its own package as if it's a 3rd-party library that we have imported, it's acceptable to me that the tree queue package would only have the QueueingAlgorithm interface (and maybe a simple generic implementation like the round robin) inside the package, then the interface can be implemented by types outside the package.

Conversely, a big part of the reason I decided to split TQA and move part of it into the `tree` package was because of the use of `tenantNodes`. It felt like a weird smell to me to leave pointers to tree-specific objects outside of the `tree` package. Maybe I'm overindexing on that smell, though.

pkg/scheduler/queue/tree/multi_queuing_algorithm_tree_queue_test.go
```diff
@@ -598,19 +598,19 @@ func Test_ChangeTenantQuerierAssignments(t *testing.T) {
...
+	tqa.tenantQuerierIDs["tenant-2"]["querier-3"] = struct{}{}
 	_, v = tree.Dequeue(&DequeueArgs{QuerierID: querier3, LastTenantIndex: newQuerierTenantIndex})
 	require.Equal(t, "query-3", v)
 
 	// during reshuffle, we only ever reassign tenant values, we don't assign an entirely new map value
-	// to TenantQuerierIDs. Reassign tenant-2 to an empty map value, and query-5 (tenant-3), which can be handled
+	// to tenantQuerierIDs. Reassign tenant-2 to an empty map value, and query-5 (tenant-3), which can be handled
 	// by any querier, should be dequeued,
-	tqa.TenantQuerierIDs["tenant-2"] = map[QuerierID]struct{}{}
+	tqa.tenantQuerierIDs["tenant-2"] = map[QuerierID]struct{}{}
```
francoposa: can these be done with the setters as well?

francoposa (approved):
LGTM!
minor naming I would lean towards just `shard` or `shards` over `sharding`.
Drop some length, keeping the meaning and not using the `-ing` seems a lot more more common in the codebase

PR #9666, opened 2024-10-18
distributor: fix pool buffer reuse logic when `distributor.max-request-pool-buffer-size` is set

ortuman:
#### What this PR does

This PR adapts the distributor code to remove the bucketed pool implementation in `util.RequestBuffers` when a maximum pool buffer size is defined.

Instead, the code for the default `util.bufferPool` type has been adapted to allow defining a maximum pool capacity.

The reason is that with the current implementatiom, bucketed buffers are used to decompress the incoming request acquiring the buffer based on the specified request size. However, since these buffers eventually end up being resized during the decompression process, they can later end up in a different bucket when returned to the pool, thus preventing subsequent reuse and leading to higher GC pressure.

This behavior has been observed in our dev environments when enabling the config parameter `-distributor.max-request-pool-buffer-size`.

Fixes n/a

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 * [BUGFIX] Fix issue where sharded queries could return annotations with incorrect or confusing position information. #9536
 * [BUGFIX] Fix issue where downstream consumers may not generate correct cache keys for experimental error caching. #9644
 * [BUGFIX] Fix issue where active series requests error when encountering a stale posting. #9580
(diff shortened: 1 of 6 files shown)


narqo (approved):
🔥 Works for me. I don't think there is a better way to benchmark the different approaches than against a real workload.

Issue #9672, opened 2024-10-18
Helm: add support for GEM's federation-frontend


dimitarvdimitrov:
This is an issue to track the work we're doing for improving the federation-frontend. Part of the work is to add support for it in the mimir-distributed Helm chart.

The idea is to be able to deploy the component on its own and connect it to another GEM cluster. This means being able to disable most of the components that the chart deploys while leaving the necessary ones (PSP, config ConfigMap, license secret, meta monitoring resources, ...)

```[tasklist]
### Tasks
- [x] Add initial support for deploying federation-frontend https://github.com/grafana/mimir/pull/9673
- [x] Add `enabled` flags to most of the unrelated components, so none of their resources are rendered (deployments, services, service monitor...) https://github.com/grafana/mimir/pull/9734
- [x] Conditional config: 1) optionally set the `distributor` endpoint in the gateway config when the distributor is enabled so the gateway doesn't crash without a distributor; 2) disable GEM's `instrumentation` when the distributor isn't deployed, so there aren't a lot of warning logs. https://github.com/grafana/mimir/pull/9734
- [x] Documentation for deploying federation-frontend on its own along with an example values file https://github.com/grafana/mimir/pull/9735
```

PR #9673, opened 2024-10-18
helm: Add GEM federation-frontend resources

dimitarvdimitrov:
#### What this PR does

This PR adds the federation-frontend resources to the Helm chart. The goal is to have the federation-frontend deployed on its own.

- Adds resources for the federation-frontend
  - Service
  - Deployment
  - PodDisruptionBudget
  - ServiceMonitor (if meta-monitoring is enabled)
- Adds an offline test with the current way of deploying a stand-alone federation-frontend: 
  - GEM gateway (to allow deploying the federation-frontend in an existing cluster)
  - admin-api, so that the GEM license is set up in case this is a standalone GEM cluster too (albeit with a single query component) 
  - minio to allow local testing for the admin-api
  - most other components are scaled to 0 replicas or disabled. My plan is to later add `enabled` toggles for all components so there are no unrelated components deployed

I've tested this locally with the `test-enterprise-federation-frontend-values.yaml` (+ license) by deploying a Mimir cluster and pointing the federation-frontend to it

Screenshot 2024-10-18 at 15 32 17

related to https://github.com/grafana/mimir/issues/9672

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥 Looks good to me.

(_Thinking out loud_) I'd maybe added a new validation in `validation.yaml`, that asserts `federation_frontend=true && enterprise=true`. But we seem to not do that generally, and I think it won't be too much of an obscure issue for a user, who might set randomly.

operations/helm/charts/mimir-distributed/ci/offline/test-enterprise-federation-frontend-values.yaml
```diff
@@ -0,0 +1,68 @@
...
+    auth:
+      type: trust
+    instrumentation:
+      enabled: false
+    federation:
+      proxy_targets:
+        - name: "mimir-distributed-1"
+          url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
+        - name: "mimir-distributed-2"
+          url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
```
charleskorn, line 28: This isn't necessarily something for this PR, but have you considered making this configurable as a value underneath the `federation_frontend` value? Given it will always need to be set when the federation-frontend is deployed, might make things a little simpler.

operations/helm/charts/mimir-distributed/templates/federation-frontend/federation-frontend-dep.yaml
```diff
@@ -0,0 +1,145 @@
...
+              mountPath: /etc/mimir
+            - name: license
+              mountPath: /license
+            - name: runtime-config
+              mountPath: /var/{{ include "mimir.name" . }}
+            - name: active-queries
+              mountPath: /active-query-tracker
+          {{- if .Values.federation_frontend.extraVolumeMounts }}
+            {{- toYaml .Values.federation_frontend.extraVolumeMounts | nindent 12 }}
+          {{- end }}
```
charleskorn, line 69: Should this be here a second time? It's also present on line 53.

operations/helm/charts/mimir-distributed/ci/offline/test-enterprise-federation-frontend-values.yaml
dimitarvdimitrov: Soo:


```yaml
federation_frontend:
  proxy_targets:
    - name: "mimir-distributed-1"
      url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
    - name: "mimir-distributed-2"
      url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
```

vs

```yaml
mimir:
  structuredConfig:
      federation:
        proxy_targets:
          - name: "mimir-distributed-1"
            url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
          - name: "mimir-distributed-2"
            url: "http://mimir-query-frontend.real-deployment.svc.cluster.local:8080/prometheus"
```


I spent about 20 min trying to get the validation to work ok so users can't provide both (because one is silently dropped then). It's surprisingly difficult and it's still not robust enough.

Do you think it will make a big enough difference? I'm erring on the side of not doing it.

operations/helm/tests/test-enterprise-federation-frontend-values-generated/mimir-distributed/templates/minio/create-bucket-job.yaml
```diff
@@ -4,44 +4,3 @@
 # however the hook won't be executed if helm install is run
```
charleskorn, line 4: I wonder if the reason this file is being generated is because the `if` is after the comment in https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/minio/create-bucket-job.yaml.

Have you tried moving the `{{- if .Values.minio.enabled }}` to the top of the file?

(not a big deal if we can't fix this easily)

charleskorn (approved):

operations/helm/tests/test-enterprise-federation-frontend-values-generated/mimir-distributed/templates/minio/create-bucket-job.yaml
dimitarvdimitrov: oh, wow i didn't think to actually look in the file

dimitarvdimitrov: should be fixed in https://github.com/grafana/mimir/pull/9673/commits/f9cd475715127c043ca5c22251208e22381abb79

operations/helm/charts/mimir-distributed/ci/offline/test-enterprise-federation-frontend-values.yaml
dimitarvdimitrov: > although I imagine the same validation problem occurs with all the other options we configure automatically.

the fact that this is a list makes it more complicated, the fact that there's a default empty list value. Otherwise we overlap the two configs if they're only nested objects without lists.

charleskorn: I think this is fine to leave as you have it.

dimitarvdimitrov:
> (_Thinking out loud_) I'd maybe added a new validation in `validation.yaml`, that asserts `federation_frontend=true && enterprise=true`. But we seem to not do that generally, and I think it won't be too much of an obscure issue for a user, who might set randomly.

just to follow-up: I did this in https://github.com/grafana/mimir/pull/9673/commits/ec08c6e9161b4651c59225c31c54b790150051a3

PR #9453, opened 2024-09-27
Mark Redis cache backend options as deprecated

56quarters:
#### Which issue(s) this PR fixes or relates to

Part of https://github.com/grafana/mimir-squad/issues/2414

Related #9441

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9677, opened 2024-10-18
jsonnet: Remove configuration to use Redis from jsonnet

56quarters:
#### What this PR does

Remove conditionals that allowed Redis to be set as a cache backend from jsonnet. Note that this does not remove support for Redis from Mimir as that will be done in a later release, this only affects jsonnet.

Related #9453

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥 As discussed on a side

> We don't plan to remove the flags from mimir until the next major release. But the jsonnet has no guarantee of backwards compatibility and we never had jsonnet to create redis sts/deployments. It only configures mimir to point at an existing redis setup

PR #9697, opened 2024-10-21
dashboards: add block max time delta panel

dimitarvdimitrov:
This helps identify when the compactor is falling behind

Screenshot 2024-10-21 at 15 54 54


seizethedave (approved):
🙇

PR #9699, opened 2024-10-21
Upstream rollout-operator zone-tracker settings.

seizethedave:
#### What this PR does

Configures rollout-operator to use zone-tracker mode, which uses a configmap to store metadata about statefulset scaling operations rather than as attributes on the statefulset being operated on. We found that the latter caused infinite recursion-type behavior in the K8S clusters of several cloud service providers.

Part of this configuration is granting rollout-operator RBAC permissions to manipulate configmaps.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 ### Jsonnet
 
 * [CHANGE] Remove support to set Redis as a cache backend from jsonnet. #9677
+* [CHANGE] Rollout-operator now stores scaling operation metadata in a Kubernetes configmap. #9699
 * [FEATURE] Add support to deploy distributors in multi availability zones. #9548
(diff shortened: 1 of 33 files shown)


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@
 ### Jsonnet
 
 * [CHANGE] Remove support to set Redis as a cache backend from jsonnet. #9677
+* [CHANGE] Rollout-operator now stores scaling operation metadata in a Kubernetes configmap. #9699
```
56quarters, line 66: Maybe elaborate on this a bit, why people would care, if they should expect things to work differently (I think the answer to this is "no"), etc.

56quarters (approved):
LGTM with one suggestion

PR #9706, opened 2024-10-22
dashbaords: include block level on queried blocks

dimitarvdimitrov:
#### Before

Screenshot 2024-10-22 at 10 36 45

#### After

Screenshot 2024-10-22 at 10 45 21

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥

jhalterman:
Do we care about only certain block levels? Maybe we could group the higher levels?

dimitarvdimitrov:
> Do we care about only certain block levels? Maybe we could group the higher levels?

Yeah, makes sense. There are sometimes up to 20 compaction levels. I grouped anything higher than L4. WDYT? 
Screenshot 2024-10-23 at 11 23 59

jhalterman (approved):

PR #9708, opened 2024-10-22
userTSDB: Avoid race condition when setting db

aknuds1:
#### What this PR does
Fix race condition in `pkg/ingester.userTSDB` when setting its `db` field.

```console
WARNING: DATA RACE
Read at 0x00c0153be140 by goroutine 76199:
  github.com/grafana/mimir/pkg/ingester.(*userTSDB).blocksToDelete()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/grafana/mimir/pkg/ingester/user_tsdb.go:364 +0x58
  github.com/grafana/mimir/pkg/ingester.(*userTSDB).blocksToDelete-fm()
      <autogenerated>:1 +0x51
  github.com/prometheus/prometheus/tsdb.(*DB).reloadBlocks()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/prometheus/prometheus/tsdb/db.go:1654 +0x49e
  github.com/prometheus/prometheus/tsdb.(*DB).run()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/prometheus/prometheus/tsdb/db.go:1149 +0x771
  github.com/prometheus/prometheus/tsdb.open.gowrap1()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/prometheus/prometheus/tsdb/db.go:1093 +0x4f

Previous write at 0x00c0153be140 by goroutine 5042:
  github.com/grafana/mimir/pkg/ingester.(*Ingester).createTSDB()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/grafana/mimir/pkg/ingester/ingester.go:2717 +0x2015
  github.com/grafana/mimir/pkg/ingester.(*Ingester).openExistingTSDB.func1()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/github.com/grafana/mimir/pkg/ingester/ingester.go:2818 +0x116
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/arve/Projects/grafana/backend-enterprise/vendor/golang.org/x/sync/errgroup/errgroup.go:78 +0x91
```

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/ingester/user_tsdb.go
```diff
@@ -92,6 +92,7 @@ type ownedSeriesState struct {
 }
 
 type userTSDB struct {
+	hasDB          atomic.Bool
```
colega, line 95: I think the fact that there's an atomic flag that is only checked sometimes (but not in most of the accesses) can create more issues in the future, as now it's unclear whether the access to `u.db` has to be synchronized.

This problem is local to the fact that `blocksToDelete` is being passed as a callback to prometheus/tsdb, so I would suggest just fixing that callback to be properly synchronized, like this:
```diff
diff --git pkg/ingester/ingester.go pkg/ingester/ingester.go
index 13b66bd64c..de9e9b743e 100644
--- pkg/ingester/ingester.go
+++ pkg/ingester/ingester.go
@@ -30,6 +30,7 @@ import (
        "github.com/grafana/dskit/ring"
        "github.com/grafana/dskit/services"
        "github.com/grafana/dskit/tenant"
+       "github.com/oklog/ulid"
        "github.com/pkg/errors"
        "github.com/prometheus/client_golang/prometheus"
        "github.com/prometheus/client_golang/prometheus/promauto"
@@ -2666,6 +2667,14 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
        }
        userDB.triggerRecomputeOwnedSeries(recomputeOwnedSeriesReasonNewUser)

+       userDBHasDB := atomic.NewBool(false)
+       blocksToDelete := func(blocks []*tsdb.Block) map[ulid.ULID]struct{} {
+               if !userDBHasDB.Load() {
+                       return nil
+               }
+               return userDB.blocksToDelete(blocks)
+       }
+
        oooTW := i.limits.OutOfOrderTimeWindow(userID)
        // Create a new user database
        db, err := tsdb.Open(udir, userLogger, tsdbPromReg, &tsdb.Options{
@@ -2680,7 +2689,7 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
                WALSegmentSize:                        i.cfg.BlocksStorageConfig.TSDB.WALSegmentSizeBytes,
                WALReplayConcurrency:                  walReplayConcurrency,
                SeriesLifecycleCallback:               userDB,
-               BlocksToDelete:                        userDB.blocksToDelete,
+               BlocksToDelete:                        blocksToDelete,
                EnableExemplarStorage:                 true, // enable for everyone so we can raise the limit later
                MaxExemplars:                          int64(i.limiter.maxExemplarsPerUser(userID)),
                SeriesHashCache:                       i.seriesHashCache,
@@ -2721,6 +2730,7 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
        }

        userDB.db = db
+       userDBHasDB.Store(true)
        // We set the limiter here because we don't want to limit
        // series during WAL replay.
        userDB.limiter = i.limiter
```



I

pkg/ingester/user_tsdb_test.go
```diff
@@ -103,6 +104,40 @@ func TestUserTSDB_acquireAppendLock(t *testing.T) {
...
+			defer wg.Done()
+
+			close(setDBStarted)
+			<-blockDeleteStarted
+			userDB.setDB(&tsdb.DB{})
+		}()
+
+		wg.Wait()
+	})
+}
```
colega, line 139: I don't think this test is useful. It's 30+ lines to test that we've actually set the bool, but it doesn't even check that we're calling `userDB.setDB(db)` instead of just doing `userDB.db = db`

pkg/ingester/user_tsdb.go
aknuds1: Thanks Oleg, applied your suggestion.

pkg/ingester/user_tsdb_test.go
aknuds1: For reference in case it wasn't clear, the test is for catching data races. I'll have to rethink it though, after applying your suggestion above.

colega (approved):

PR #9712, opened 2024-10-22
Allow querying against an expandable number of storages

leizor:
#### What this PR does

Currently the querier expects to run queries against two storages only:
  - Ingesters
  - Store gateways

The queriers know when each queryable was applicable and queries the appropriate queryables.

This commmit introduces the TimeRangeQueryable struct which is a wrapper around Queryable. It ties the Queryable to the logic that deems it applicable to an incoming query.

This, in conjunction with the newly-added AdditionalStorageQueryables property on the Mimir struct, enables the querier to query optionally against more storages than just the ingesters and store gateways in the future.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 * [ENHANCEMENT] Return server-side total bytes processed statistics as a header through query frontend. #9645
 * [ENHANCEMENT] Query-scheduler: Remove the experimental `query-scheduler.prioritize-query-components` flag. Request queues always prioritize query component dequeuing above tenant fairness. #9703
 * [ENHANCEMENT] Ingester: Emit traces for block syncing, to join up block-upload traces. #9656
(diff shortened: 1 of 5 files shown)


Logiraptor (approved):
Makes sense to me. Note this logic is in one more place: when we decide which query component queue the request should end up in. [source](https://github.com/grafana/mimir/blob/3117ce032d60d8d7789022428e83b9a35e5820bb/pkg/frontend/v2/frontend_scheduler_adapter.go#L96-L113).

I don't think you need to think about that in this PR, but something to keep in mind. If we want to be able to schedule queries differently depending on whether the new components are hit, we'll need to modify that code to understand they exist. It's not without some downsides though, since it would require increasing the minimum number of querier -> query-scheduler connections a bit. So I'd suggest returning to this based on performance testing later.

replay (approved):
Nice work, thx!

Would you mind adding a CHANGELOG entry please?

PR #9717, opened 2024-10-23
MQE: create slices with capacity for number of steps remaining, not all possible steps

charleskorn:
#### What this PR does

This PR is a small optimisation for instant vector selectors in MQE. 

Previously, instant vector selectors would always use a slice with capacity for the number of steps in the range query.

However, if a series only has samples beginning part-way through the query range, this is wasteful: we know the series can't have more samples than remain in the query range, so we don't need such a big slice.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


jhesketh (approved):
Nice find. This could be a decent amount for somebody querying over a year for a newly spawned metric.

PR #9664, opened 2024-10-18
MQE: subqueries

charleskorn:
#### What this PR does

This PR adds support for subqueries to MQE.

Our benchmarks show mild latency improvements in most cases over Prometheus' engine, and improvements in peak memory utilisation in most cases:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/streamingpromql/benchmarks
cpu: Apple M1 Pro
                                                                         │ Prometheus  │               Mimir                │
                                                                         │   sec/op    │   sec/op     vs base               │
Query/sum_over_time(a_1[10m:3m]),_instant_query-10                         156.0µ ± 4%   152.3µ ± 4%   -2.36% (p=0.015 n=6)
Query/sum_over_time(a_1[10m:3m]),_range_query_with_100_steps-10            165.6µ ± 2%   154.6µ ± 1%   -6.63% (p=0.002 n=6)
Query/sum_over_time(a_1[10m:3m]),_range_query_with_1000_steps-10           250.2µ ± 1%   245.1µ ± 2%   -2.03% (p=0.041 n=6)
Query/sum_over_time(a_100[10m:3m]),_instant_query-10                       1.116m ± 2%   1.075m ± 8%        ~ (p=0.065 n=6)
Query/sum_over_time(a_100[10m:3m]),_range_query_with_100_steps-10          1.882m ± 4%   1.691m ± 1%  -10.18% (p=0.002 n=6)
Query/sum_over_time(a_100[10m:3m]),_range_query_with_1000_steps-10         8.662m ± 0%   9.212m ± 1%   +6.36% (p=0.002 n=6)
Query/sum_over_time(a_2000[10m:3m]),_instant_query-10                      16.02m ± 1%   15.38m ± 1%   -3.99% (p=0.002 n=6)
Query/sum_over_time(a_2000[10m:3m]),_range_query_with_100_steps-10         30.11m ± 1%   26.48m ± 1%  -12.06% (p=0.002 n=6)
Query/sum_over_time(a_2000[10m:3m]),_range_query_with_1000_steps-10        160.5m ± 7%   171.7m ± 3%   +6.95% (p=0.015 n=6)
Query/sum_over_time(nh_1[10m:3m]),_instant_query-10                        202.3µ ± 1%   193.9µ ± 1%   -4.18% (p=0.002 n=6)
[...]

PR #9719, opened 2024-10-23
MQE: improve performance of subqueries when the parent query is a range query with many steps

charleskorn:
#### What this PR does

This PR is a follow-up PR to https://github.com/grafana/mimir/pull/9664.

This PR improves the performance of subqueries when the parent query is a range query with many steps.

It does this by changing the behaviour of the ring buffer implementations to search forwards through the buffer to find the end of the range when a subquery is being evaluated, rather than backwards. Searching backwards makes sense for range vector selectors, where there is expected to only be one point, if any, at the end of the buffer outside the range, but for subqueries in range queries it is expected that there will be many points at the end of the buffer outside the range, so it is usually faster to search from the beginning of the buffer.

I'd suggest reviewing each commit separately, as each builds on the previous one and incrementally improves things.

In benchmarking, peak memory consumption is unchanged, and latency is reduced for both subqueries and range vector selectors. For example:

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                                                                                       │ original.txt │              final.txt              │
                                                                                       │    sec/op    │    sec/op     vs base               │
Query/rate(a_1[1m]),_range_query_with_1000_steps/engine=Mimir-10                         230.2µ ±  1%   218.0µ ±  2%   -5.30% (p=0.002 n=6)
Query/rate(a_100[1m]),_range_query_with_1000_steps/engine=Mimir-10                       7.707m ±  1%   7.086m ±  2%   -8.07% (p=0.002 n=6)
Query/rate(a_2000[1m]),_range_query_with_1000_steps/engine=Mimir-10                      134.3m ±  1%   124.3m ±  1%   -7.43% (p=0.002 n=6)
Query/rate(a_2000[1m]),_range_query_with_10000_steps/engine=Mimir-10                      1.356 ±  6%    1.253 ±  5%   -7.56% (p=0.002 n=6)
Query/rate(nh_1[1m]),_range_query_with_1000_steps/engine=Mimir-10                        705.2µ ±  1%   649.9µ ±  1%   -7.85% (p=0.002 n=6)
Query/rate(nh_100[1m]),_range_query_with_1000_steps/engine=Mimir-10                      44.71m ±  1%   43.46m ±  1%   -2.79% (p=0.002 n=6)
Query/rate(nh_2000[1m]),_range_query_with_1000_steps/engine=Mimir-10                     853.5m ±  1%   843.9m ±  4%        ~ (p=0.310 n=6)
Query/rate(nh_2000[1m]),_range_query_with_10000_steps/engine=Mimir-10                     8.723 ±  3%    8.573 ±  1%   -1.73% (p=0.002 n=6)
Query/sum_over_time(a_2000[10m:3m]),_range_query_with_1000_steps/engine=Mimir-10         172.4m ±  2%   129.6m ±  3%  -24.80% (p=0.002 n=6)
Query/sum_over_time(nh_2000[10m:3m]),_range_query_with_1000_steps/engine=Mimir-10         1.155 ±  2%    1.101 ±  1%   -4.68% (p=0.002 n=6)
Query/sum(sum_over_time(a_100[10m:3m])),_range_query_with_1000_steps/engine=Mimir-10     9.652m ±  1%   7.456m ±  1%  -22.75% (p=0.002 n=6)
Query/sum(sum_over_time(a_2000[10m:3m])),_range_query_with_1000_steps/engine=Mimir-10    177.2m ±  1%   133.6m ±  1%  -24.56% (p=0.002 n=6)

```

https://github.com/grafana/mimir/pull/9664

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/types/fpoint_ring_buffer.go
+++ b/pkg/streamingpromql/types/fpoint_ring_buffer.go
@@ -80,8 +80,13 @@ func (b *FPointRingBuffer) Append(p promql.FPoint) error {
 // Set searchForwards to true if it is expected that there are many points with timestamp greater than maxT, and few points with
 // earlier timestamps.
 // Set searchForwards to false if it is expected that only a few of the points will have timestamp greater than maxT.
+// existing is an existing view instance for this buffer that is reused if provided. It can be nil.
 // The returned view is no longer valid if this buffer is modified (eg. a point is added, or the buffer is reset or closed).
-func (b *FPointRingBuffer) ViewUntil(maxT int64, searchForwards bool) FPointRingBufferView {
+func (b *FPointRingBuffer) ViewUntil(maxT int64, searchForwards bool, existing *FPointRingBufferView) *FPointRingBufferView {
+	if existing == nil {
+		existing = &FPointRingBufferView{buffer: b}
+	}
+
 	if searchForwards {
 		size := 0
 
@@ -89,7 +94,8 @@ func (b *FPointRingBuffer) ViewUntil(maxT int64, searchForwards bool) FPointRing
 			size++
 		}
 
-		return FPointRingBufferView{buffer: b, size: size}
+		existing.size = size
+		return existing
 	}
 
 	size := b.size
@@ -98,7 +104,8 @@ func (b *FPointRingBuffer) ViewUntil(maxT int64, searchForwards bool) FPointRing
 		size--
 	}
 
-	return FPointRingBufferView{buffer: b, size: size}
+	existing.size = size
+	return existing
 }
 
 // pointAt returns the point at index 'position'.
--- a/pkg/streamingpromql/types/fpoint_ring_buffer.go
+++ b/pkg/streamingpromql/types/fpoint_ring_buffer.go
@@ -161,6 +161,10 @@ type FPointRingBufferView struct {
 // FIXME: the fact we have to expose this is a bit gross, but the overhead of calling a function with ForEach is terrible.
 // Perhaps we can use range-over function iterators (https://go.dev/wiki/RangefuncExperiment) once this is not experimental?
 func (v FPointRingBufferView) UnsafePoints() (head []promql.FPoint, tail []promql.FPoint) {
+	if v.size == 0 {
+		return nil, nil
+	}
+
 	endOfHeadSegment := v.buffer.firstIndex + v.size
 
 	if endOfHeadSegment > len(v.buffer.points) {
--- a/pkg/streamingpromql/types/fpoint_ring_buffer.go
+++ b/pkg/streamingpromql/types/fpoint_ring_buffer.go
@@ -165,7 +165,7 @@ func (v FPointRingBufferView) UnsafePoints() (head []promql.FPoint, tail []promq
 
 	if endOfHeadSegment > len(v.buffer.points) {
 		// Need to wrap around.
-		endOfTailSegment := endOfHeadSegment % len(v.buffer.points)
+		endOfTailSegment := endOfHeadSegment - len(v.buffer.points)
 		endOfHeadSegment = len(v.buffer.points)
 		return v.buffer.points[v.buffer.firstIndex:endOfHeadSegment], v.buffer.points[0:endOfTailSegment]
 	}
--- a/pkg/streamingpromql/types/fpoint_ring_buffer.go
+++ b/pkg/streamingpromql/types/fpoint_ring_buffer.go
@@ -173,7 +173,7 @@ type FPointRingBufferView struct {
 
 // UnsafePoints returns slices of the points in this buffer view.
 // Either or both slice could be empty.
-// Callers must not modify the values in the returned slices or return them to a pool.
+// Callers must not modify the values in the returned slices nor return them to a pool.
 // Calling UnsafePoints is more efficient than calling CopyPoints, as CopyPoints will create a new slice and copy all
 // points into the slice, whereas UnsafePoints returns a view into the internal state of the buffer.
 // The returned slices are no longer valid if this buffer is modified (eg. a point is added, or the buffer is reset or closed).
--- a/pkg/streamingpromql/types/fpoint_ring_buffer.go
+++ b/pkg/streamingpromql/types/fpoint_ring_buffer.go
@@ -42,82 +42,6 @@ func (b *FPointRingBuffer) DiscardPointsBefore(t int64) {
 	}
 }
 
-// UnsafePoints returns slices of the points in this buffer, including only points with timestamp less than or equal to maxT.
-// Either or both slice could be empty.
-// Callers must not modify the values in the returned slices or return them to a pool.
-// Calling UnsafePoints is more efficient than calling CopyPoints, as CopyPoints will create a new slice and copy all
-// points into the slice, whereas UnsafePoints returns a view into the internal state of this buffer.
-// The returned slices are no longer valid if this buffer is modified (eg. a point is added, or the buffer is reset or closed).
-//
-// FIXME: the fact we have to expose this is a bit gross, but the overhead of calling a function with ForEach is terrible.
-// Perhaps we can use range-over function iterators (https://go.dev/wiki/RangefuncExperiment) once this is not experimental?
-func (b *FPointRingBuffer) UnsafePoints(maxT int64) (head []promql.FPoint, tail []promql.FPoint) {
-	size := b.size
-
-	for size > 0 && b.points[(b.firstIndex+size-1)%len(b.points)].T > maxT {
-		size--
-	}
-
-	endOfHeadSegment := b.firstIndex + size
-
-	if endOfHeadSegment > len(b.points) {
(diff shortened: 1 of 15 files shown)


pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -58,6 +61,11 @@ func (b *FPointRingBuffer) Append(p promql.FPoint) error {
 			return err
 		}
 
+		if !isPowerOfTwo(cap(newSlice)) {
```
jhesketh, line 64: Why not enforce this contract in the bucketing pools themselves. ie, don't allow a `factor` and set it to `2`. I think this will also allow us to make looking up the correct bucket for each Get+Put more efficient where we won't need to loop over the buckets.

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -183,60 +166,100 @@ func (b *FPointRingBuffer) Close() {
...
-	if b.size == 0 {
-		panic("Can't get first element of empty buffer")
+type FPointRingBufferView struct {
+	buffer *FPointRingBuffer
+	size   int
+}
+
+// UnsafePoints returns slices of the points in this buffer view.
+// Either or both slice could be empty.
+// Callers must not modify the values in the returned slices or return them to a pool.
```
jhesketh, line 176: (nit, I know this existed previously) could be read as "callers must return them to a pool".

```suggestion
// Callers must not modify the values in the returned slices nor return them to a pool.
```

pkg/streamingpromql/types/hpoint_ring_buffer.go
```diff
@@ -234,58 +203,105 @@ func (b *HPointRingBuffer) Close() {
...
-	if b.size == 0 {
-		panic("Can't get first element of empty buffer")
+type HPointRingBufferView struct {
+	buffer *HPointRingBuffer
+	size   int
+}
+
+// UnsafePoints returns slices of the points in this buffer view.
+// Either or both slice could be empty.
+// Callers must not modify the values in the returned slices or return them to a pool.
```
jhesketh, line 213: (nit, I know this existed previously) could be read as "callers must return them to a pool".

```suggestion
// Callers must not modify the values in the returned slices nor return them to a pool.
```

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -169,12 +145,19 @@ func (b *FPointRingBuffer) Release() {
 // s will be modified in place when the buffer is modified, and callers should not modify s after passing it off to the ring buffer via Use.
 // s will be returned to the pool when Close is called, Use is called again, or the buffer needs to expand, so callers
 // should not return s to the pool themselves.
+// s must have a capacity that is a power of two.
 func (b *FPointRingBuffer) Use(s []promql.FPoint) {
+	if !isPowerOfTwo(cap(s)) {
```
jhesketh, line 150: With my above suggestion, this is perhaps a little more risky to leave unenforced, so may be worth keeping here. However, in theory, so long as everything has come from the bucketed pool then we should always have the right power.

pkg/streamingpromql/types/ring_buffer_test.go
```diff
@@ -13,26 +13,30 @@ import (
...
-	UnsafePoints(maxT int64) (head []T, tail []T)
-	CopyPoints(maxT int64) ([]T, error)
-	LastAtOrBefore(maxT int64) (T, bool)
-	CountAtOrBefore(maxT int64) int
-	AnyAtOrBefore(maxT int64) bool
-	First() T
 	Reset()
 	Use(s []T)
 	Release()
+	ViewUntilForTesting(maxT int64, searchForwards bool) ringBufferView[T]
```
jhesketh, line 24: I know this comes from an earlier commit in the history, but I think you should drop `ViewUntilForTesting` and implement a ViewUntilSearchingForwards/Backwards

jhesketh (commented):
lgtm, some nice collection of optimisations.

I think we can optimise the bucketed pools somewhat by enforcing the factor to be 2. I can look at doing that as a followup if you prefer though.

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -150,6 +74,38 @@ func (b *FPointRingBuffer) Append(p promql.FPoint) error {
...
 }
 
+// ViewUntil returns a view into this buffer, including only points with timestamps less than or equal to maxT.
+// searchForwards is a hint used to optimise the search for points satisfying the maxT condition, ViewUntil will return the
+// same result regardless of teh value.
+// Set searchForwards to true if it is expected that there are many points with timestamp greater than maxT, and few points with
+// earlier timestamps.
+// Set searchForwards to false if it is expected that only a few of the points will have timestamp greater than maxT.
+// The returned view is no longer valid if this buffer is modified (eg. a point is added, or the buffer is reset or closed).
+func (b *FPointRingBuffer) ViewUntil(maxT int64, searchForwards bool) FPointRingBufferView {
```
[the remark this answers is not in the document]
jhesketh: nevermind, an early note adding in stepping through the commits.

pkg/streamingpromql/types/fpoint_ring_buffer.go
charleskorn: >this will also allow us to make looking up the correct bucket for each Get+Put more efficient where we won't need to loop over the buckets

Yep, I had the same idea 🙂 I'd like to leave that for a separate PR though, this PR is big enough already.

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -183,58 +139,90 @@ func (b *FPointRingBuffer) Close() {
...
-		if p.T <= maxT {
-			return count
-		}
+	return v.buffer.points[v.buffer.firstIndex]
+}
 
-		count--
+// Last returns the last point in this ring buffer view.
+// It returns false if the view is empty.
+func (v FPointRingBufferView) Last() (promql.FPoint, bool) {
```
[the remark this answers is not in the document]
charleskorn: Agreed, but I'd like to keep this for a separate PR as you suggest.

pkg/streamingpromql/types/ring_buffer_test.go
charleskorn: Done in f3ac4c3

pkg/streamingpromql/types/fpoint_ring_buffer.go
charleskorn: This check is cheap, and tracking down an issue related to this could be quite painful, so I'd prefer to keep it for now.

pkg/streamingpromql/types/ring_buffer_test.go
```diff
@@ -370,7 +370,7 @@ type fPointRingBufferWrapper struct {
 }
 
 func (w *fPointRingBufferWrapper) ViewUntilForTesting(maxT int64, searchForwards bool) ringBufferView[promql.FPoint] {
-	return w.ViewUntil(maxT, searchForwards)
+	return w.ViewUntil(maxT, searchForwards, nil)
```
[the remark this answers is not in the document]
charleskorn: Done: c399b1b

pkg/streamingpromql/types/fpoint_ring_buffer.go
charleskorn: Comment added in a91c8b1.

pkg/streamingpromql/types/ring_buffer_test.go
```diff
@@ -286,6 +286,51 @@ func TestRingBuffer_RemoveLastPoint(t *testing.T) {
...
+		require.NoError(t, buf.Append(promql.FPoint{T: 1, F: 100}))
+		require.NoError(t, buf.Append(promql.FPoint{T: 2, F: 200}))
+		require.NoError(t, buf.Append(promql.FPoint{T: 3, F: 300}))
+		require.NoError(t, buf.Append(promql.FPoint{T: 4, F: 400}))
+
+		view := buf.ViewUntilSearchingForwards(2, nil)
+		viewShouldHavePoints(t, view, promql.FPoint{T: 1, F: 100}, promql.FPoint{T: 2, F: 200})
+
+		// Test that reusing the view with ViewUntilSearchingForwards works correctly.
+		view = buf.ViewUntilSearchingForwards(1, view)
```
jhesketh, line 301: should we ignore the returned value here to confirm that `view` is being reused and not a new value is being returned into view (either that or checking the pointer values etc)

jhesketh (approved):
Just a nit on the test, not blocking on it :-).

lgtm :rocket:

pkg/streamingpromql/types/ring_buffer_test.go
charleskorn: Yep, fixed in af9b9fa

PR #9718, opened 2024-10-23
Revert "Upgrade to google.golang.org/grpc v1.66.2 / modify certain protobuf messages to retain their unmarshaling buffer"

charleskorn:
Reverts grafana/mimir#9401

Running `go test -v -run="TestBothEnginesReturnSameResultsForBenchmarkQueries/a_100,.*"` from `pkg/streamingpromql/benchmarks` fails roughly 50% of the time for me, and it looks like Prometheus' engine is getting corrupted label names and values, for example:

```
=== RUN   TestBothEnginesReturnSameResultsForBenchmarkQueries/2_*_a_100,_range_query_with_100_steps
    utils.go:61:
        	Error Trace:	/Users/charleskorn/Repositories/Grafana/mimir/pkg/streamingpromql/testutils/utils.go:61
        	            				/Users/charleskorn/Repositories/Grafana/mimir/pkg/streamingpromql/benchmarks/comparison_test.go:111
        	Error:      	Not equal:
        	            	expected: labels.Labels{labels.Label{Name:"\x03", Value:"\n"}}
        	            	actual  : labels.Labels{labels.Label{Name:"l", Value:"0"}}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -2,4 +2,4 @@
        	            	  (labels.Label) {
        	            	-  Name: (string) (len=1) "\x03",
        	            	-  Value: (string) (len=1) "\n"
        	            	+  Name: (string) (len=1) "l",
        	            	+  Value: (string) (len=1) "0"
        	            	  }
        	Test:       	TestBothEnginesReturnSameResultsForBenchmarkQueries/2_*_a_100,_range_query_with_100_steps
```

(the result from Prometheus' engine is the expected value in this test, MQE is the actual)

Another sample failure:

```
=== RUN   TestBothEnginesReturnSameResultsForBenchmarkQueries/a_100,_range_query_with_1000_steps
    utils.go:61:
        	Error Trace:	/Users/charleskorn/Repositories/Grafana/mimir/pkg/streamingpromql/testutils/utils.go:61
        	            				/Users/charleskorn/Repositories/Grafana/mimir/pkg/streamingpromql/benchmarks/comparison_test.go:111
        	Error:      	Not equal:
        	            	expected: labels.Labels{labels.Label{Name:"\x00(\x03@\n\x01\xd0\x02", Value:"\x00\xa0\xfd\x00("}, labels.Label{Name:"\xd0", Value:"4\x00"}}
[...]

PR #9721, opened 2024-10-23
Don't free buffers after reading query stream

colega:
#### What this PR does

While the write path has been carefully crafted to make sure that we don't keep references to the strings backed by arrays with data from grpc stream (because it would cause a memory leak), this doesn't seem to be the case in the query path, where some tests started failing after we started to reuse those backing arrays through the usage of memory buffers implemented in the new gRPC library.

This change reverts the buffer freeing, means that it won't be recycled (and will be garbage collected) to ensure data correctness, while we investigate where the data references are kept.

Follow up on https://github.com/grafana/mimir/pull/9401

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -269,7 +269,6 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 			if len(resp.Timeseries) > 0 {
 				for _, series := range resp.Timeseries {
 					if limitErr := queryLimiter.AddSeries(series.Labels); limitErr != nil {
-						resp.FreeBuffer()
 						return ingesterQueryResult{}, limitErr
 					}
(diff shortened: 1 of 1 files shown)


colega:
There was no changelog entry in https://github.com/grafana/mimir/pull/9401 so nothing to update here.

bboreham (approved):
Seems legit.  I guess we should take a look at the profiles once it gets into use.

narqo (approved):
🔥 Works for me

> where some tests started failing after we started to reuse those backing arrays through the usage of memory buffers implemented in the new gRPC library

For the history #9718 shows some details about which tests

colega:
It's also concerning that only those tests failed 😱

PR #9723, opened 2024-10-23
blockbuilder: Basic alerts

narqo:
#### What this PR does

This PR adds new alerts, that warns about possible scenarios, where the `block-builder` behaves abnormally. See the updated runbook for the details.

Note that I put the alert as warnings for now, since block-builder is still in its early experimental phase.

### DIFF
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1624,15 +1624,16 @@ How to **investigate**:
 
 - Check the block-builder logs to see what its pods have been busy with. The block-builder logs the `start consuming` and `done consuming` log messages, that mark per-partition conume-cycles. These log records include the details about the cycle, the Kafka topic's offsets, etc. Troubleshoot based on that.
 
-### MimirBlockBuilderLaging
+### MimirBlockBuilderLagging
 
-This alert fires when the block-builder instances report large lag values for the Kafka partitions.
+This alert fires when the block-builder instances report a large number of unprocessed records in the Kafka partitions.
 
 How it **works**:
 
-- When the block-builder starts a new consume cycle, it checks how many records the Kafka partition has in the backlog.
+- When the block-builder starts a new consume cycle, it checks how many records the Kafka partition has in the backlog. This number is tracked in the `cortex_blockbuilder_consumer_lag_records` metric.
 - The block-builder must consume and process these records into TSDB blocks.
-- If the block-builder reports high values in the lag, this could indicate that a block-builder instance cannot fully process and commit the Kafka records.
+- At the end of the processing, the block-builder commits the offset of the last fully processed record into Kafka.
+- If the block-builder reports high values in the lag, this could indicate that a block-builder instance cannot fully process and commit Kafka record.
 
 How to **investigate**:
 
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1618,11 +1618,11 @@ This alert fires when the block-builder stops reporting any processed cycles for
 How it **works**:
 
 - Block-builder periodically consumes a portion of the backlog from Kafka partition, and processes the consumed data into TSDB blocks. The block-buikder calls these periods "cycles".
-- If no cycles were processed during extended period of time, that can indicate that block-builder is stuck and cannot complete cycle processing.
+- If no cycles were processed during extended period of time, that can indicate that a block-builder instance is stuck and cannot complete cycle processing.
 
 How to **investigate**:
 
-- Check block-builder logs to see what block-builder is busy with, and troubleshoot based on that.
+- Check block-builder logs to see what its pods are busy with. Troubleshoot based on that.
 
 ### MimirBlockBuilderCompactAndUploadFailed
 
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1611,6 +1611,31 @@ How to **fix**:
 
   1. Once ingesters are stable, revert the temporarily config applied in the previous step.
 
+### MimirBlockBuilderNoCycleProcessing
+
+This alert fires when the block-builder stops reporting any processed cycles for an unexpectedly long time.
+
+How it **works**:
+
+- Block-builder periodically consumes a portion of the backlog from Kafka topic, and processes the consumed data into TSDB blocks. The block-buikder calls these periods "cycles".
+- If no cycles were processed during extended period of time, that can indicate that block-builder is stuck and cannot complete cycle processing.
+
+How to **investigate**:
+
+- Check block-builder logs to see what block-builder is busy with, and troubleshoot based on that.
+
+### MimirBlockBuilderCompactAndUploadFailed
+
+How it **works**:
+
+- Block-builder periodically consumes data from Kafka topic, and processes the consumed data into TSDB blocks.
+- It compacts and uploads the produced TSDB blocks to the object storage.
+- If block-builder encounters issues during compaction or uploading of the blocks if reports the failure metric, that triggers the alert.
+
+How to **investigate**:
+
+- Explore block-builder logs to check what errors are there.
+
 ## Errors catalog
 
 Mimir has some codified error IDs that you might see in HTTP responses or logs.
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1624,6 +1624,21 @@ How to **investigate**:
 
 - Check the block-builder logs to see what its pods have been busy with. The block-builder logs the `start consuming` and `done consuming` log messages, that mark per-partition conume-cycles. These log records include the details about the cycle, the Kafka topic's offsets, etc. Troubleshoot based on that.
 
+### MimirBlockBuilderLaging
+
+This alert fires when the block-builder instances report large lag values for the Kafka partitions.
+
+How it **works**:
+
+- When the block-builder starts a new consume cycle, it checks how many records the Kafka partition has in the backlog.
+- The block-builder must consume and process these records into TSDB blocks.
+- If the block-builder reports high values in the lag, this could indicate that a block-builder instance cannot fully process and commit the Kafka records.
+
+How to **investigate**:
+
+- Check if the per-partition lag, reported by the `cortex_blockbuilder_consumer_lag_records` metric, has been growing over the past hours.
+- Explore the block-builder logs for any errors reported while it processed the partition.
+
 ### MimirBlockBuilderCompactAndUploadFailed
 
 How it **works**:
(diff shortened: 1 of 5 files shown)


docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
 
   1. Once ingesters are stable, revert the temporarily config applied in the previous step.
 
+### MimirBlockBuilderNoCycleProcessing
+
+This alert fires when the block-builder stops reporting any processed cycles for an unexpectedly long time.
+
+How it **works**:
+
+- Block-builder periodically consumes a portion of the backlog from Kafka topic, and processes the consumed data into TSDB blocks. The block-buikder calls these periods "cycles".
```
pracucci, line 1620: [nit] Isn't partition more accurate here? It processes 1 partition at a time.

```suggestion
- Block-builder periodically consumes a portion of the backlog from Kafka partition, and processes the consumed data into TSDB blocks. The block-buikder calls these periods "cycles".
```

operations/mimir-mixin/alerts/ingest-storage.libsonnet
```diff
@@ -212,6 +212,36 @@
             message: '%(product)s {{ $labels.%(per_instance_label)s }} in %(alert_aggregation_variables)s Kafka client produce buffer utilization is {{ printf "%%.2f" $value }}%%.' % $._config,
           },
         },
+
+        // Alert if block-builder didn't process cycles in the past hour.
+        {
+          alert: $.alertName('BlockBuilderNoCycleProcessing'),
+          'for': '30m',
+          expr: |||
+            max by(%(alert_aggregation_labels)s) (histogram_count(increase(cortex_blockbuilder_consume_cycle_duration_seconds[1h]))) == 0
```
pracucci, line 221: Two comments here:
1. I would suggest alerting on a per-pod basis, so that we notice if a single block-builder is stuck: `max by(cluster, namespace, pod) (histogram_count(increase(cortex_blockbuilder_consume_cycle_duration_seconds[1h]))) == 0`
2. Doing `[1h]` for `30m`, or doing `[90m]` for `1m` is the same thing, so you can easily back test it. Try to run the query `max by(cluster, namespace, pod) (histogram_count(increase(cortex_blockbuilder_consume_cycle_duration_seconds[90m]))) == 0` both in dev and prod. You will see some single replicas stuck from time to time. I would like you to investigate the root cause before adding this alert, because if they're false positives we should find a way to fix the false positives, otherwise we should investigate the issue

pracucci (approved):
Pre-approving, but I would really like to see a per-pod alert for the stuck processing.

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+This alert fires when the block-builder stops reporting any processed cycles for an unexpectedly long time.
+
+How it **works**:
+
+- Block-builder periodically consumes a portion of the backlog from Kafka partition, and processes the consumed data into TSDB blocks. The block-buikder calls these periods "cycles".
+- If no cycles were processed during extended period of time, that can indicate that a block-builder instance is stuck and cannot complete cycle processing.
+
+How to **investigate**:
+
+- Check block-builder logs to see what its pods are busy with. Troubleshoot based on that.
```
tacole02, line 1625:
```suggestion
- Check the block-builder logs to see what its pods are busy with. Troubleshoot based on that.
```

tacole02: Can we specify "what its pods are busy with". This sounds a bit jargony and anthropomorphic to me. What do the pods actually do?

tacole02: Similarly, can we provide more detail into "troubleshoot based on that"? It feels a little open-ended.

I think clarifying the first point will help us to clarify the second.

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+
+### MimirBlockBuilderCompactAndUploadFailed
+
+How it **works**:
```
tacole02, line 1629:
```suggestion
How it works:
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+
+- Block-builder periodically consumes data from Kafka topic, and processes the consumed data into TSDB blocks.
```
tacole02, line 1631:
```suggestion
- The block-builder periodically consumes data from Kafka topic and processes the consumed data into TSDB blocks.
```

tacole02: Could we say either "a Kafka topic" or "the Kafka topic" depending on which is more accurate?

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+- It compacts and uploads the produced TSDB blocks to the object storage.
```
tacole02, line 1632:
```suggestion
- It compacts and uploads the produced TSDB blocks to object storage.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+- If block-builder encounters issues during compaction or uploading of the blocks if reports the failure metric, that triggers the alert.
```
tacole02, line 1633:
```suggestion
- If the block-builder encounters issues while compacting or uploading the blocks, it reports the failure metric, which then triggers the alert.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+
+How to **investigate**:
```
tacole02, line 1635:
```suggestion
How to investigate:
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
...
+
+- Explore block-builder logs to check what errors are there.
```
tacole02, line 1637:
```suggestion
- Explore the block-builder logs to check what errors are there.
```

tacole02 (approved):
Looks good! I left a few comments and suggestions.

codesome (commented):
We should probably also add a warning when the lag is too high when starting a consumption. We can choose some manual number for now and tweak it later or find a better way to alert on this. Perhaps 3-4M is a good number to start?

Recently, when we were treating OOO histograms as server error, we were lagging quite behind but the consumption was getting initiated and ending as soon as we hit the error. So probably the two warnings in this PR may not have alerted us on it.

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1611,6 +1611,31 @@ How to **fix**:
 
   1. Once ingesters are stable, revert the temporarily config applied in the previous step.
 
+### MimirBlockBuilderNoCycleProcessing
+
+This alert fires when the block-builder stops reporting any processed cycles for an unexpectedly long time.
+
+How it **works**:
```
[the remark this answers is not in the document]
narqo: I'm not sure, we export the runbooks to the docs site. Here I'm sticking to how all other alerts are structured in this document. Let's look at their formatting in bulk separately.

narqo:
@codesome could you take another look. I've added the `MimirBlockBuilderLaging` warning.

operations/mimir-mixin/alerts/ingest-storage.libsonnet
```diff
@@ -212,6 +212,51 @@
...
+            message: '%(product)s {{ $labels.%(per_instance_label)s }} in %(alert_aggregation_variables)s has not processed cycles in the past hour.' % $._config,
+          },
+        },
+
+        // Alert if block-builder per partition lag is higher than the threshhold.
+        {
+          alert: $.alertName('BlockBuilderLaging'),
+          'for': '1h',
+          expr: |||
+            max by(%(alert_aggregation_labels)s, %(per_instance_label)s) (max_over_time(cortex_blockbuilder_consumer_lag_records[60m])) > 4e6
```
seizethedave, line 236: This reports lag in terms of records, doesn't it? This seems less useful than seconds, because 4M records could be 10 seconds or 10 hours depending on the cell. What do you think?

operations/mimir-mixin/alerts/ingest-storage.libsonnet
```diff
@@ -212,6 +212,51 @@
...
+            severity: 'warning',
+          },
+          annotations: {
+            message: '%(product)s {{ $labels.%(per_instance_label)s }} in %(alert_aggregation_variables)s has not processed cycles in the past hour.' % $._config,
+          },
+        },
+
+        // Alert if block-builder per partition lag is higher than the threshhold.
+        {
+          alert: $.alertName('BlockBuilderLaging'),
```
seizethedave, line 233: "Lagging"

docs/sources/mimir/manage/mimir-runbooks/_index.md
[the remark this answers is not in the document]
tacole02: Sounds good!

operations/mimir-mixin/alerts/ingest-storage.libsonnet
codesome: We don't have a metric to measure lag in terms of time. You are true that 4M could mean any time range, but for now this should be a usable warning and migrate over to time based measurement with new metrics (maybe the scheduler can help in this).

codesome (commented):
Regarding `BlockBuilderLagging`:

If the partition was lagging during a cycle above 4M, but if it caught up and went below 4M by the next cycle, we will still send a warning for that for an hour because of the 60m lookback in the query. Because this alert gets active as soon as the first cycle starts (stays in pending), and firing for the entire duration of second cycle.

```bash
| <-1st cycle->    |<---2nd cycle-->
|------- 5M -------|------- 3M --------|           # lag metric
|                  |                   |
0m                 1h                  2h          # timeline
|---alert pending--|-------firing------| resolved  # alert state

```

If we assume that the cycle is 1hr long, we can probably just lookback 10m (i.e. `[10m]` in the query), and have the `for` period to be greater than lookback + 1hr, so >70m, so perhaps 75m? This is to avoid warning if during a cycle it went above 4M and in the next cycle it was below 4M again.

```bash
| <-1st cycle->    |<---2nd cycle-->
|------- 5M -------|------- 3M --------|  # lag metric
|                  |   |   |           |
0m                 1h  70m 75m         2h # timeline
|-----alert pending----|--inactive--->    # alert state

```

WDYT?

---

Edit:

Queries in action

60m
Screenshot 2024-10-28 at 10 20 23 PM


10m
Screenshot 2024-10-28 at 10 20 45 PM

operations/mimir-mixin/alerts/ingest-storage.libsonnet
seizethedave: Cool. Yeah, I think it can fairly easily.

PR #9724, opened 2024-10-23
MQE: Add support for changes function

lamida:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- n.a Documentation added.


pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,54 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+	StepFunc:               changes,
+}
+
+func changes(step types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
+	fHead, fTail := step.Floats.UnsafePoints(step.RangeEnd)
+
+	haveFloats := len(fHead) > 0 || len(fTail) > 0
+
+	if !haveFloats {
+		// PromQL engine doesn't support histogram for `changes` function yet,
```
jhesketh, line 384: (nit)
```suggestion
		// Prometheus' engine doesn't support histogram for `changes` function yet,
```

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,54 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+		// therefore we won't add that yet too.
+		return 0, false, nil, nil
+	}
+
+	return changesFloats(fHead, fTail), true, nil, nil
+}
+
+func changesFloats(head, tail []promql.FPoint) float64 {
```
jhesketh, line 392: (nit) probably doesn't need a separate function and can go inside changes above

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,54 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+			if current != prev && !(math.IsNaN(current) && math.IsNaN(prev)) {
+				changes++
+			}
+			prev = current
+		}
+	}
+
+	if len(tail) > 0 {
+		prev := tail[0].F
+		for _, sample := range tail[1:] {
```
jhesketh, line 412: rather than repeating this logic, set up a function you can pass head and tail into. Have a look at some other functions with `accumulate`

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,54 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+			current := sample.F
+			if current != prev && !(math.IsNaN(current) && math.IsNaN(prev)) {
+				changes++
+			}
+			prev = current
+		}
+	}
+
+	if len(tail) > 0 {
+		prev := tail[0].F
```
charleskorn, line 411: This isn't quite right - if the last sample in `head` and the first sample in `tail` are different, we won't increment `changes`, but we should.

Perhaps we could get the first point in the ring buffer by calling `step.Floats.First()`, then iterate through the points?

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,54 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+
+func changesFloats(head, tail []promql.FPoint) float64 {
+	changes := 0.0
+
+	if len(head) == 0 && len(tail) == 0 {
+		return 0
+	}
+
+	if len(head) > 0 {
+		prev := head[0].F
```
jhesketh, line 400: I'm not sure this is correct as it is a ring buffer you'll probably need to look at the last element of tail. Might need to add a test of this boundary

jhesketh (commented):
Nice work, thanks for that!

I think it's also worth putting this function through `TestCompareVariousMixedMetricsVectorSelectors` to test some edge cases (just add it to line 2310).

pkg/streamingpromql/operators/functions/range_vectors.go
charleskorn, line 384: [nit] We use the term "Prometheus' engine" to refer to it - "PromQL engine" is ambiguous because MQE is a PromQL engine too.
```suggestion
		// Prometheus' engine doesn't support histogram for `changes` function yet,
```

Might also be good to add a link to the Prometheus' issue tracking adding support for native histograms in `changes`, if there is one.

lamida: I tried to find but it seems there are no issue tracking for that.

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Updated in the latest commit

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Updated in the latest commit

lamida:
@jhesketh @charleskorn I have addressed all the comments. These are changes from the previous reviews:
- Handle ring-buffer wrap around.
- Moved the function logic from `range_vector.go` to `fpoint_ring_buffer.go` so that I can test ring-buffer wrap around easier. Let me know if anyone have concern.

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: Charles is likely referring to this one: https://github.com/prometheus/prometheus/issues/13934

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -237,6 +239,49 @@ func (b *FPointRingBuffer) AnyAtOrBefore(maxT int64) bool {
 	return b.points[b.firstIndex].T <= maxT
 }
 
+// ChangesAtOrBefore returns the number of changes between subsequent points in this ring buffer from the first
+// point to the point with timestamp less than or equal to maxT.
+func (b *FPointRingBuffer) ChangesAtOrBefore(maxT int64) (float64, bool) {
```
jhesketh, line 244: I think this is better fit into functions/range_vectors in terms of consistency. Can we not create a test in the functions package and pass it a constructed `types.RangeVectorStepData`?

pkg/streamingpromql/types/fpoint_ring_buffer.go
```diff
@@ -237,6 +239,49 @@ func (b *FPointRingBuffer) AnyAtOrBefore(maxT int64) bool {
...
+	if len(fHead) > 0 && len(fTail) > 0 {
+		// We don't need to compare the first point with itself, hence start from fHead[1:]
+		accumulate(fHead[1:])
+		accumulate(fTail)
+	}
+
+	if len(fHead) > 0 && len(fTail) == 0 {
+		// We don't need to compare the first point with itself, hence start from fHead[1:]
+		accumulate(fHead[1:])
+	}
```
jhesketh, line 281: This is consistent with how we use accumulate elsewhere. I don't think there is much overhead in range when the len is 0 vs using an if.
```suggestion
	// The points buffer is wrapped around, therefore we need to check changes starting from the buffer's head
	// and then continue to the tail.
	// We don't need to compare the first point with itself, hence start from fHead[1:]
	accumulate(fHead[1:])
	accumulate(fTail)
```

pkg/streamingpromql/types/fpoint_ring_buffer_test.go
```diff
@@ -0,0 +1,155 @@
...
+			name: "Changes on points that are changing before the time and its ring-buffer are wrapped around",
+			fields: fields{
+				firstIndex: 2,
+				size:       5,
+				points: []promql.FPoint{
+					{T: 7, F: 4},
+					{T: 9, F: 99},
+					{T: 1, F: 1},
+					{T: 3, F: 2},
+					{T: 5, F: 3},
```
jhesketh, line 128: Would be good to have a test where the value doesn't change at the wrap around

jhesketh (dismissed):

pkg/streamingpromql/types/fpoint_ring_buffer.go
lamida: > I think this is better fit into functions/range_vectors in terms of consistency

Actually some functions are implemented in `fpoint_ring_buffer.go` too like `CountAtOrBefore`, `LastAtOrBefore` and `AnyAtOrBefore`.

> Can we not create a test in the functions package and pass it a constructed types.RangeVectorStepData?
It is not possible to create the buffer to simulate wrap around because `points` and other fields are private. I prefer not to make them public just for testing.

Other than that, if we put the logic in `range_vectors.go` we are exposing the internal representation of the ring buffer in `range_vector.go`. Like on how iterating head and tail which I believe should be internal `fpoint_ring_buffer.go`. 

At the end I don't have any strong opinion and I am fine if there is a better argument.

lamida: > Can we not create a test in the functions package and pass it a constructed types.RangeVectorStepData?
It is not possible to create the buffer to simulate wrap around because the fields are private.

pkg/streamingpromql/types/fpoint_ring_buffer.go
lamida: You are right. We can simplify that. Updated in a7d04746f58fc416884ba3fbeb1f3536418eabfb

pkg/streamingpromql/types/fpoint_ring_buffer_test.go
lamida: Added in af5c86715873ccde85c07e5064897864dc473be6

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: That issue didn't mention changes. This is the comment that mentioned the need to add support ([code link](https://github.com/prometheus/prometheus/blob/7c7116fea8343795cae6da42960cacd0207a2af8/promql/functions.go#L1444)).

pkg/streamingpromql/types/fpoint_ring_buffer.go
charleskorn: I would favour this not being on the ring buffer type - `ChangesAtOrBefore` can efficiently be implemented by using `UnsafePoints`, whereas `CountAtOrBefore`, `LastAtOrBefore` and `AnyAtOrBefore` don't need the slices produced by `UnsafePoints` and so benefit from being implemented directly on the ring buffer type.

lamida: I will moved back to `range_vectors.go` then. I will figure out how the best to test that.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,43 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+		for _, sample := range points {
+			current := sample.F
+			if current != prev && !(math.IsNaN(current) && math.IsNaN(prev)) {
+				changes++
+			}
+			prev = current
+		}
+	}
+
+	accumulate(fHead)
```
charleskorn, line 407: Couldn't we skip the first point? Otherwise we'll just be comparing it to itself.

```suggestion
	accumulate(fHead[1:])
```

charleskorn (approved):
LGTM modulo suggestion below

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: Yes, I think this was already done in the previous revision?

jhesketh (commented):
lgtm too, but agree with Charles. Thanks Jon!

PR #9732, opened 2024-10-24
blockbuilder: bail when cycle section reached cycle end offset

narqo:
#### What this PR does

This PR fixes the following problem we observed in one of our testing environments:

1. a partition was scaled down (and became inactive)
1. block-builder started the next cycle for this partition
1. block-builder's cycle was lagging and it was chopped into sections
1. the second to last cycle section consumed the partition up to the `cycle_end_offset` (which was the partition's high watermark)
1. the last cycle section started consuming from the last commit (`commit_offset == cycle_end_offset`)

Because the partition was inactive, the consumer ended up in the dead-spiral. It kept polling the broker for records, without noticing that it was already standing at the high watermark.


pkg/blockbuilder/blockbuilder.go
```diff
@@ -332,6 +338,14 @@ func (b *BlockBuilder) consumePartitionSection(
 	// Thus, truncate the timestamp with ConsumptionInterval here to round the block's range.
 	blockEnd := sectionEndTime.Truncate(b.cfg.ConsumeInterval)
 
+	// If the last commit offset has already reached the offset, that marks the end of the cycle, bail. This can happen when the partition
+	// was scaled down (it's inactive now), but block-builder was lagging, and it chopped this cycle into sections. After a cycle section
```
seizethedave, line 342: Is it specifically when there was sufficient backlog to make >1 consumption section?

seizethedave: nit: I think this whole comment is adding a complex explanation to a simple situation: start==end means nothing to consume!

seizethedave (approved):
LGTM % optional reduction of commentary. 😄

pkg/blockbuilder/blockbuilder.go
narqo: Yes, exactly. Note that, though, the `nextConsumeCycle` checks that the partition's lag is none zero. So the part about having more than one cycle section is important to trigger the bug.

codesome (approved):

PR #9743, opened 2024-10-26
Ingester circuit breaker deactivate() function

pr00se:
#### What this PR does

Add a `deactivate()` method to the ingester circuit breakers to allow them to be disabled and re-enabled dynamically.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/circuitbreaker.go
+++ b/pkg/ingester/circuitbreaker.go
@@ -192,10 +192,12 @@ func (cb *circuitBreaker) activate() {
 		return
 	}
 	if cb.cfg.InitialDelay == 0 {
+		level.Info(cb.logger).Log("msg", "activating circuit breaker", "requestType", cb.requestType)
 		cb.active.Store(true)
 		return
 	}
 	time.AfterFunc(cb.cfg.InitialDelay, func() {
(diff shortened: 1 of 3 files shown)


pkg/ingester/circuitbreaker.go
```diff
@@ -188,17 +188,28 @@ func (cb *circuitBreaker) isActive() bool {
...
 		cb.active.Store(true)
+		return
 	}
 	time.AfterFunc(cb.cfg.InitialDelay, func() {
+		level.Info(cb.logger).Log("msg", "activating circuit breaker", "requestType", cb.requestType)
 		cb.active.Store(true)
 	})
 }
 
+func (cb *circuitBreaker) deactivate() {
```
duricanikolic, line 205: So this PR basically adds a support to deactivate circuit breakers, but it doesn't actually deactivate them anywhere?

pr00se: Correct -- I wanted to keep the changes small, so will add the invocation in another PR. The plan is to call it from https://github.com/grafana/mimir/blob/main/pkg/ingester/downscale.go when we change the read-only state.

duricanikolic (approved):
LGTM

PR #9756, opened 2024-10-28
Distributor.queryIngesterStream: Free gRPC buffers upon error

aknuds1:
#### What this PR does

In `Distributor.queryIngesterStream`, free gRPC buffers when an error occurs in the `queryIngester` closure that gets passed to `ring.DoMultiUntilQuorumWithoutSuccessfulContextCancellation`. No attempt is made to free the buffers though if querying is successful, as the combined query response should retain references to said buffers. I add a comment for future reference as to why we don't free buffers in the success case.

See #9721 for background, as it reverts the `Distributor.queryIngesterStream` gRPC buffer freeing introduced in #9401 (due to race conditions essentially, as the combined query response retains gRPC buffer references).

I've run the `TestBothEnginesReturnSameResultsForBenchmarkQueries` tests in pkg/streamingpromql/benchmarks with race detection enabled, and no problems are found.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/query.go
+++ b/pkg/distributor/query.go
@@ -229,7 +229,7 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 
 	// queryIngester MUST call cancelContext once processing is completed in order to release resources. It's required
 	// by ring.DoMultiUntilQuorumWithoutSuccessfulContextCancellation() to properly release resources.
-	queryIngester := func(ctx context.Context, ing *ring.InstanceDesc, cancelContext context.CancelCauseFunc) (ingesterQueryResult, error) {
+	queryIngester := func(ctx context.Context, ing *ring.InstanceDesc, cancelContext context.CancelCauseFunc) (result ingesterQueryResult, err error) {
 		log, ctx := spanlogger.NewWithLogger(ctx, d.log, "Distributor.queryIngesterStream")
 		cleanup := func() {
 			log.Span.Finish()
@@ -248,6 +248,10 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 
 				cleanup()
 			}
+
+			if err != nil {
+				result.freeBuffers()
+			}
 		}()
 
 		log.Span.SetTag("ingester_address", ing.Addr)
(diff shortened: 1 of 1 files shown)


pkg/distributor/query.go
```diff
@@ -277,20 +297,24 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 			} else if len(resp.Chunkseries) > 0 {
 				// Enforce the max chunks limits.
 				if err := queryLimiter.AddChunks(ingester_client.ChunksCount(resp.Chunkseries)); err != nil {
+					result.freeBuffers()
```
colega, line 300: We could use a single defer statement that checks the `err` and frees if it's not nil, wdyt? Then we don't have to care about making sure that all cases are covered.

aknuds1: Yeah, I thought one _could_ also use `defer`. If you prefer it, I can do it.

aknuds1: Done, PTAL.

pkg/distributor/query.go
```diff
@@ -249,15 +267,15 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
-
 		// Why retain the batches rather than iteratively build a single slice?
 		// If we iteratively build a single slice, we'll spend a lot of time copying elements as the slice grows beyond its capacity.
 		// So instead, we build the slice in one go once we know how many series we have.
 		var streamingSeriesBatches [][]labels.Labels
 		streamingSeriesCount := 0
 
 		for {
+			// XXX: Note that while we free responses' gRPC buffers on error, we don't do the same in case of success,
+			// as the combined response retains references to gRPC buffers.
```
colega, line 278: Now, thinking twice about this PR: does it make sense to have this extra code optimizing the performance of a non-happy case? 

I think we should fix (or at least evaluate if it's intended) the usage of pooled data instead.

aknuds1: The way I'm thinking about it is that this is a first, and easy step, before we fix also the happy path. It's difficult to say how often the `queryIngester` closure will fail in practice.

colega (approved):

PR #9759, opened 2024-10-28
continuous-test: make debugging easier in docker compose env

francoposa:
#### What this PR does

Converts the timestamps fed into the runner to be in UTC, since all other timestamps encountered when attaching a breakpoint are in UTC.

Also expands OOO window config so the catchup writes from read-write continuous test don't all fail when you bring it up when metamonitoring is on.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/development/mimir-microservices-mode/config/mimir.yaml
+++ b/development/mimir-microservices-mode/config/mimir.yaml
@@ -172,6 +172,9 @@ limits:
   query_sharding_total_shards: 16
   query_sharding_max_sharded_queries: 32
   ingestion_rate: 50000
+  # expanded OOO window makes it easier to run continuous-test
+  # otherwise catch-up writes are rejected when metamonitoring is on
+  out_of_order_time_window: 1h
   native_histograms_ingestion_enabled: true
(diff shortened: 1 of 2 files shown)


56quarters (approved):
LGTM. Is it correct that the UTC change is a no-op when the machine is already in UTC so this should only affect local dev?

francoposa:
> LGTM. Is it correct that the UTC change is a no-op when the machine is already in UTC so this should only affect local dev?

yes, I put my machine in UTC and this shows true:

```golang
func main() {

	timeNow := time.Now()

	timeNowUTC := timeNow.UTC()

	fmt.Println(timeNow)
	fmt.Println(timeNowUTC)
	fmt.Println(timeNow.Unix() == timeNowUTC.Unix())

}
```

```console
2024-10-28 21:19:16.85950292 +0000 UTC m=+0.000031090
2024-10-28 21:19:16.85950292 +0000 UTC
true
```

The two objects are not considered equal because of the different time zone but the Unix time is equal

PR #9760, opened 2024-10-28
Disable ingester push circuit breaker in read-only mode

pr00se:
#### What this PR does

We ran across an issue where the ingester's push circuit breaker was getting tripped when first starting, or when leaving read-only mode. The `--initial-delay` option allows us to add a delay before the circuit breaker starts measuring request times and making decisions. This is a solution for the cold start case, but does not help us when a running ingester is leaving read-only mode.

This PR builds on https://github.com/grafana/mimir/pull/9743 to disable the ingester push circuit breaker when entering read-only mode, and enable it when leaving read-only mode. It also only enables the push breaker at start time if the ingester isn't starting in read-only mode.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


jhalterman (approved):
LGTM

PR #9770, opened 2024-10-29
Close gzip readers after use

aknuds1:
#### What this PR does
I noticed that in most cases we don't close gzip readers after use, even though it's the [caller's responsibility](https://pkg.go.dev/compress/gzip#NewReader).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -18,6 +18,7 @@ import (
 	"github.com/grafana/dskit/httpgrpc"
 	"github.com/grafana/dskit/httpgrpc/server"
 	"github.com/grafana/dskit/middleware"
+	"github.com/grafana/dskit/runutil"
 	"github.com/grafana/dskit/tenant"
 	"github.com/pierrec/lz4/v4"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/distributor/otel.go
```diff
@@ -78,10 +79,10 @@ func OTLPHandler(
 			return httpgrpc.Errorf(http.StatusUnsupportedMediaType, "unsupported compression: %s. Only \"gzip\" or no compression supported", contentEncoding)
 		}
 
-		var decoderFunc func(io.ReadCloser) (req pmetricotlp.ExportRequest, uncompressedBodySize int, err error)
+		var decoderFunc func(io.Reader) (req pmetricotlp.ExportRequest, uncompressedBodySize int, err error)
```
aknuds1, line 82: This function shouldn't close the passed reader, so might as well use `io.Reader`.

bboreham:
> it's the caller's responsibility

Are you reading that from [here](https://github.com/grafana/mimir/blob/3e6b60e7f98feaeb028b7e5bc15cec4038645e48/vendor/github.com/klauspost/compress/gzip/gunzip.go#L91)?
It seems ambiguous to me whether it means the Reader passed in or the Reader returned.
I tend towards the former, given that `Close()` on the gzip decompressor doesn't do anything.

aknuds1:
> Are you reading that from [here](https://github.com/grafana/mimir/blob/3e6b60e7f98feaeb028b7e5bc15cec4038645e48/vendor/github.com/klauspost/compress/gzip/gunzip.go#L91)?

I'm reading it from [here](https://pkg.go.dev/compress/gzip#NewReader) (stdlib vs klauspost).

> It seems ambiguous to me whether it means the Reader passed in or the Reader returned.

To me it seems clear that it's referring to the returned `gzip.Reader`, especially since it [links](https://pkg.go.dev/compress/gzip#Reader) to that type.

AFAICT, the gzip compressor Close method doesn't do anything but return an eventual error, but that's an implementation detail and could change. Also, I don't think it's a bad idea to consistently log any returned error (as we already do in one call site), if we do have a logger.

aldernero (approved):

PR #9195, opened 2024-09-04
fix relabeling and compactor svc showing diff in argocd

KyriosGN0:
Signed-off-by: AvivGuiser <aviv.guiser@placer.ai>

#### What this PR does

fixes mimir chart showing constest diff due to server side changes that k8s makes but are not in the chart itself

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9774, opened 2024-10-30
fix alertmanager sts showing diff in argocd

daanschipper:
#### What this PR does

Same as #9195, but now also for AlertManager. Fixes mimir chart showing constest diff due to server side changes that k8s makes but are not in the chart itself.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
Works for me, thank you 🔥

PR #9782, opened 2024-10-30
query-frontend: Do not break connections by failing early when the query parameters are malformed

codesome:
#### What this PR does

Fail early on malformed queries in the query-frontend. Otherwise, they fail when we are trying to send the query to the scheduler, and can break the connection between the frontend and the scheduler.

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/v2/frontend.go
+++ b/pkg/frontend/v2/frontend.go
@@ -269,6 +269,14 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, req *httpgrpc.HTTPRequest)
 
 	retries := f.cfg.WorkerConcurrency + 1 // To make sure we hit at least two different schedulers.
 
+	// Validate if we can convert the request to scheduler request.
+	// If we don't check here, it can fail when trying to send to the scheduler, and break
+	// the connection between frontend and scheduler, failing all the queries that were on that connection.
+	_, err = f.schedulerWorkers.toSchedulerAdapter.extractAdditionalQueueDimensions(freq.ctx, freq.request, time.Now())
+	if err != nil {
+		return nil, nil, httpgrpc.Errorf(http.StatusBadRequest, "failed to convert to scheduler enqueue request: %s", err.Error())
+	}
+
 enqueueAgain:
 	spanLogger.DebugLog("msg", "enqueuing request")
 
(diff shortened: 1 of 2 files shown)


pkg/frontend/v2/frontend.go
```diff
@@ -269,6 +269,14 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, req *httpgrpc.HTTPRequest)
 
 	retries := f.cfg.WorkerConcurrency + 1 // To make sure we hit at least two different schedulers.
 
+	// Validate if we can convert the request to scheduler request.
+	// If we don't check here, it can fail when trying to send to the scheduler, and break
+	// the connection between frontend and scheduler, failing all the queries that were on that connection.
+	_, err = f.schedulerWorkers.toSchedulerAdapter.extractAdditionalQueueDimensions(freq.ctx, freq.request, time.Now())
```
francoposa, line 275: this will only parse time params for range, instant and labels queries.

cardinality, series, metadata, remote read, etc would not be validated still

the part of that func is

```golang
switch {
	case querymiddleware.IsRangeQuery(httpRequest.URL.Path), querymiddleware.IsInstantQuery(httpRequest.URL.Path):
		decodedRequest, err := a.codec.DecodeMetricsQueryRequest(httpRequest.Context(), httpRequest)
		if err != nil {
			return nil, err
		}
		minT := decodedRequest.GetMinT()
		maxT := decodedRequest.GetMaxT()

		return a.queryComponentQueueDimensionFromTimeParams(tenantIDs, minT, maxT, now), nil
	case querymiddleware.IsLabelsQuery(httpRequest.URL.Path):
		decodedRequest, err := a.codec.DecodeLabelsQueryRequest(httpRequest.Context(), httpRequest)
		if err != nil {
			return nil, err
		}
```

codesome: The current aim is to prevent connections from breaking. I updated the PR title to reflect it. We can revisit this part of checking all errors as a follow up. WDYT?

francoposa (approved):

codesome:
Closing in favour of https://github.com/grafana/mimir/pull/9786

PR #9783, opened 2024-10-30
Block-builder-scheduler: job management

seizethedave:
#### What this PR does

* Add a `jobQueue` type with support for assigning, completing, reassigning jobs. jobQueue strives to be general. It doesn't know anything about Kafka. It could be reused for something else with little effort.
* Add a time-based lease mechanism for jobs. Callers can extend the lease, and jobs with expired leases can be unassigned/reassigned.
* Add a job to the queue when a sufficient Kafka backlog has accumulated in a partition.
* Add stand-in methods that represent the eventual scheduler RPC handlers and test them.

Changes to Kafka workings:
* Update the Kafka job backlog detection logic. (Not the final state, but on the road to it.)
* Update consumer group configs to reflect that there will be **one** consumer group - the one whose offsets the scheduler exclusively manages. (Block-builder workers will just fetch, and won't update offsets.)

What's next?
* Periodically committing updated offsets to Kafka.
* Startup procedure that recovers initial job state from learned Kafka/blockbuilder information.
* RPC server interface

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/blockbuilder.go
```diff
@@ -203,7 +203,7 @@ func (b *BlockBuilder) nextConsumeCycle(ctx context.Context, cycleEndTime time.T
 			continue
 		}
 
-		state := partitionStateFromLag(b.logger, lag, b.fallbackOffsetMillis)
+		state := PartitionStateFromLag(b.logger, lag, b.fallbackOffsetMillis)
```
seizethedave, line 206: Note: I will follow up with moving many of these Kafka utilities to the scheduler package. Long-term, scheduler will be their sole user.

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -78,24 +86,91 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
...
 	}
-	endOffsets, err := admin.ListEndOffsets(ctx, s.cfg.Kafka.Topic)
+
+	oldTime := time.Now().Add(-s.cfg.ConsumeInterval)
+	oldOffsets, err := s.adminClient.ListOffsetsAfterMilli(ctx, oldTime.UnixMilli(), s.cfg.Kafka.Topic)
 	if err != nil {
-		level.Warn(s.logger).Log("msg", "failed to list end offsets", "err", err)
+		level.Warn(s.logger).Log("msg", "failed to obtain old offsets", "err", err)
+		return
 	}
```
seizethedave, line 115: Provisional. Change imminent. Skim heavily.

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -78,24 +86,91 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
...
+		for part, gl := range ps {
+			partStr := fmt.Sprint(part)
+			s.metrics.partitionStartOffset.WithLabelValues(partStr).Set(float64(gl.Start.Offset))
+			s.metrics.partitionEndOffset.WithLabelValues(partStr).Set(float64(gl.End.Offset))
+			s.metrics.partitionCommittedOffset.WithLabelValues(partStr).Set(float64(gl.Commit.At))
+		}
 	}
-	endOffsets, err := admin.ListEndOffsets(ctx, s.cfg.Kafka.Topic)
+
+	oldTime := time.Now().Add(-s.cfg.ConsumeInterval)
```
seizethedave, line 110: Provisional. Change imminent. Skim heavily.

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -78,24 +86,91 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
...
 
-	startOffsets.Each(func(o kadm.ListedOffset) {
-		s.metrics.partitionStartOffset.WithLabelValues(fmt.Sprint(o.Partition)).Set(float64(o.Offset))
-	})
-	endOffsets.Each(func(o kadm.ListedOffset) {
-		s.metrics.partitionEndOffset.WithLabelValues(fmt.Sprint(o.Partition)).Set(float64(o.Offset))
+	oldOffsets.Each(func(o kadm.ListedOffset) {
+		if l, ok := lag.Lookup(o.Topic, o.Partition); ok {
+			if l.Commit.At < o.Offset {
+				level.Info(s.logger).Log("msg", "partition ready", "p", o.Partition)
```
narqo, line 122: From the current block-builder's experience, having a log record, with the partitions, that were skipped due to no lag, was helpful to understand what had happened during an unexpected behaviour.

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -4,27 +4,34 @@ package scheduler
...
+	adminClient *kadm.Client
+	jobs        *jobQueue
 	cfg         Config
 	logger      log.Logger
 	register    prometheus.Registerer
 	metrics     schedulerMetrics
+
+	mu        sync.Mutex
+	committed kadm.Offsets
+	dirty     bool
```
narqo, line 34: To double-check: these fields are for the future updates, when the scheduler will take over managing of the commits, right? Maybe a todo comment here could clarify it?

narqo (approved):
I didn't do a detailed review. But the current state of the changes makes a good sense to me 🔥

pkg/blockbuilder/scheduler/scheduler.go
seizethedave: Thanks, that's correct. Will do!

seizethedave: I've just removed these bits as they're entirely half-baked noise in this PR.

pkg/blockbuilder/scheduler/config.go
codesome, line 1: We don't need a separate file for this. Let's put all this in scheduler.go

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -78,24 +80,84 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
...
-	})
-	endOffsets.Each(func(o kadm.ListedOffset) {
-		s.metrics.partitionEndOffset.WithLabelValues(fmt.Sprint(o.Partition)).Set(float64(o.Offset))
+	oldOffsets.Each(func(o kadm.ListedOffset) {
+		if l, ok := lag.Lookup(o.Topic, o.Partition); ok {
+			if l.Commit.At < o.Offset {
+				level.Info(s.logger).Log("msg", "partition ready", "p", o.Partition)
+
+				// The job is uniquely identified by {topic, partition, consumption start offset}.
+				jobID := fmt.Sprintf("%s/%d/%d", o.Topic, o.Partition, l.Commit.At)
```
codesome, line 119: With the aim of having only one job per partition at a time, I probably should not have the commit offset as part of the job ID.

pkg/blockbuilder/scheduler/jobs.go
```diff
@@ -0,0 +1,192 @@
...
+func (s *jobQueue) addOrUpdate(id string, spec jobSpec) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if j, ok := s.jobs[id]; ok {
+		// We can only update an unassigned job.
+		if j.assignee == "" {
+			j.spec = spec
+		}
+	} else {
```
codesome, line 67: nit: we can remove the `else` condition by having a `return` in the `if` block.

pkg/blockbuilder/scheduler/jobs.go
```diff
@@ -0,0 +1,192 @@
...
+func (a *jobSpec) less(b *jobSpec) bool {
+	return a.commitRecTs.Before(b.commitRecTs)
+}
+
+type jobHeap []*job
+
+// Implement the heap.Interface for jobHeap.
+func (h *jobHeap) Len() int           { return len(*h) }
+func (h *jobHeap) Less(i, j int) bool { return (*h)[i].spec.less(&(*h)[j].spec) }
+func (h *jobHeap) Swap(i, j int)      { (*h)[i], (*h)[j] = (*h)[j], (*h)[i] }
```
codesome, line 177: This should also work. Slices are already pointers.
```suggestion
func (h jobHeap) Len() int           { return len(h) }
func (h jobHeap) Less(i, j int) bool { return h[i].spec.less(&h[j].spec) }
func (h jobHeap) Swap(i, j int)      { h[i], h[j] = h[j], h[i] }
```

pkg/blockbuilder/scheduler/jobs.go
```diff
@@ -0,0 +1,192 @@
...
+		return errJobNotAssigned
+	}
+
+	j.leaseExpiry = time.Now().Add(s.leaseTime)
+	return nil
+}
+
+// completeJob completes the job with the given ID for the given worker,
+// removing it from the jobQueue.
+func (s *jobQueue) completeJob(jobID, workerID string) error {
```
codesome, line 107: In the context of the queue, "completing" a job may not make sense. We are technically just removing the job from the queue here.

```suggestion
func (s *jobQueue) removeJob(jobID, workerID string) error {
```

pkg/blockbuilder/scheduler/jobs.go
```diff
@@ -0,0 +1,192 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	j, ok := s.jobs[jobID]
+	if !ok {
+		return errJobNotFound
+	}
+	if j.assignee != workerID {
+		return errJobNotAssigned
+	}
```
codesome, line 124: I have one scenario in mind. Job J was assigned to BB1. Lease expires due to communication issue and job gets assigned to BB2 now, but BB1 is still doing the work. Now BB1 and BB2 completes the job. So I guess in that case we ignore the job completion of BB1 and only consider BB2. Right?

Small optimisation: when BBs are updating their job status, scheduler can verify if that job is assigned to that BB, and if not, ask BB to stop that job. Because scheduler will not consider their work for a "complete" status.

Scenarios like this might also affect our in-memory maintenance of committed offset and can drift from what's in kafka. So we should be careful when doing that, i.e. not get the commit offset down, and only re-commit if it is higher than previous commit. Basically scheduler syncs with kafka and keeps the higher among scheduler and kafka and commits where necessary.

pkg/blockbuilder/scheduler/config.go
```diff
@@ -11,18 +11,24 @@ import (
...
 }
 
 func (cfg *Config) RegisterFlags(f *flag.FlagSet) {
-	f.StringVar(&cfg.BuilderConsumerGroup, "block-builder-scheduler.builder-consumer-group", "block-builder", "The Kafka consumer group used by block-builders.")
-	f.StringVar(&cfg.SchedulerConsumerGroup, "block-builder-scheduler.scheduler-consumer-group", "block-builder-scheduler", "The Kafka consumer group used by block-builder-scheduler.")
+	f.StringVar(&cfg.BuilderConsumerGroup, "block-builder-scheduler.builder-consumer-group", "block-builder", "The Kafka consumer group used for getting/setting commmitted offsets.")
 	f.DurationVar(&cfg.SchedulingInterval, "block-builder-scheduler.scheduling-interval", 20*time.Second, "How frequently to recompute the schedule.")
+	f.DurationVar(&cfg.ConsumeInterval, "block-builder-scheduler.consume-interval", 1*time.Hour, "Interval between consumption cycles.")
+	f.DurationVar(&cfg.StartupObserveTime, "block-builder-scheduler.startup-observe-time", 25*time.Second, "How long to observe worker state before scheduling jobs.")
+	f.DurationVar(&cfg.JobLeaseTime, "block-builder-scheduler.job-lease-time", 2*time.Minute, "How long a job lease will live for before expiring.")
```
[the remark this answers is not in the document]
codesome: After reading through the code, I understand now. The term "lease" was confusing. Probably something along the lines of "job update timeout" might be a more user friendly flag name. Perhaps it can also be derived from `UpdateJobInterval`, for example `4 * UpdateJobInterval` based on current defaults.

codesome (approved):
Some initial review. I did not review scheduler.go since I believe it is in-progress. I have not reviewed tests as well assuming it will get updated with ongoing work.

I have approved it to unblock you. I plan to do a more through review of all the code once we have the core logic stabilized.

pkg/blockbuilder/scheduler/config.go
```diff
@@ -11,18 +11,24 @@ import (
...
+	f.DurationVar(&cfg.UpdateJobInterval, "block-builder-scheduler.update-job-interval", 30*time.Second, "How often clients should send job updates.")
```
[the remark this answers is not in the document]
seizethedave: Details are in the air, but I'm considering telling builders this info rather than having it configured in 2 services.

pkg/blockbuilder/scheduler/jobs.go
seizethedave: Yes on all counts. 

> I have one scenario in mind. Job J was assigned to BB1. Lease expires due to communication issue and job gets assigned to BB2 now, but BB1 is still doing the work. Now BB1 and BB2 completes the job. So I guess in that case we ignore the job completion of BB1 and only consider BB2. Right?

That is correct. This is even made tricky in the startup observation code because we're learning from the workers about the jobs. I'm adding an assignment generation number so we can decide the winner.

> Small optimisation: when BBs are updating their job status, scheduler can verify if that job is assigned to that BB, and if not, ask BB to stop that job. Because scheduler will not consider their work for a "complete" status.
> 
> Scenarios like this might also affect our in-memory maintenance of committed offset and can drift from what's in kafka. So we should be careful when doing that, i.e. not get the commit offset down, and only re-commit if it is higher than previous commit. Basically scheduler syncs with kafka and keeps the higher among scheduler and kafka and commits where necessary.

pkg/blockbuilder/scheduler/config.go
seizethedave: Can you convince me more? Scheduler.go is approaching 300LOC in my other branch and I'm trying to keep it somewhat lucid.

pkg/blockbuilder/scheduler/config.go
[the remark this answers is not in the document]
seizethedave: Hm, Lease is a common name for this feature, esp in task queues. So I'm not convinced. JobLeaseExpiry or something like that may be an improvement, though.

pkg/blockbuilder/scheduler/config.go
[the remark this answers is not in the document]
seizethedave: I'll remove it in the meantime. :)

pkg/blockbuilder/scheduler/jobs.go
seizethedave: True enough!

seizethedave: Actually, it's a fairly specific operation where it only removes a job that is in-progress and assigned to the given worker. So, I think the intention of completeJob rings more true.
As with all things, I'll keep it in mind as the clay gets pushed around.

pkg/blockbuilder/scheduler/scheduler.go
seizethedave: Thanks for this. Here's what I think:

The v1 guarantee of 1 job per partition (1JPP) is definitely something we want to uphold.

While I started with "topic/partition" IDs like you suggested, I've found including the offset makes a lot of sense because:
- It keeps logs about job progress/failure unambiguous and easy to correlate
- It simplifies our client communication model where clients repeatedly report historical job status
- It positions us well for future multi-job partitions

About the current code and how it guarantees 1JPP - yes, it's currently weak. It assumes offsets won't advance until the current job is complete. Given we have several services and multiple flavors of Kafka in play, that's not an assumption we should rely on.

Rather than using the jobQueue map key to enforce uniqueness of active partition consumption, I'm considering adding dedicated state in `BlockBuilderScheduler` to unambiguously track scheduled partitions. WDYT?

PR #9795, opened 2024-11-01
MQE: Add support for delta function

lamida:

Fixes #<issue number>

#### Checklist

- [X] Tests updated.
- na Documentation added.
- na [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2073,6 +2073,12 @@ func TestAnnotations(t *testing.T) {
 			expr: `sum(metric{type="histogram"})`,
 		},
 
+		"delta() native histogram unknown CounterResetHint": {
+			data:                       mixedFloatHistogramData,
+			expr:                       `delta(metric{type="histogram"}[3m])`,
+			expectedWarningAnnotations: []string{`PromQL warning: this native histogram metric is not a gauge: "metric" (1:7)`},
+		},
+
 		"stdvar() with only floats": {
 			data: mixedFloatHistogramData,
 			expr: `stdvar(metric{type="float"})`,
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2054,6 +2054,11 @@ func TestAnnotations(t *testing.T) {
 		metric{series="incompatible-custom-buckets"} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}} {{schema:-53 sum:1 count:1 custom_values:[2 3] buckets:[1]}} {{schema:-53 sum:5 count:4 custom_values:[5 10] buckets:[1 2 1]}}
     `
 
+	nativeHistogramsWithResetHintsMix := `
+		metric{reset_hint="unknown"} {{schema:0 sum:0 count:0}}+{{schema:0 sum:5 count:4 buckets:[1 2 1]}}x3
+		metric{reset_hint="gauge"} {{schema:0 sum:0 count:0 counter_reset_hint:gauge}}+{{schema:0 sum:5 count:4 buckets:[1 2 1] counter_reset_hint:gauge}}x3
+    `
+
 	testCases := map[string]annotationTestCase{
 		"sum() with float and native histogram at same step": {
 			data:                       mixedFloatHistogramData,
@@ -2074,9 +2079,13 @@ func TestAnnotations(t *testing.T) {
 		},
 
 		"delta() native histogram unknown CounterResetHint": {
-			data:                       mixedFloatHistogramData,
-			expr:                       `delta(metric{type="histogram"}[3m])`,
+			data:                       nativeHistogramsWithResetHintsMix,
+			expr:                       `delta(metric{reset_hint="unknown"}[3m])`,
 			expectedWarningAnnotations: []string{`PromQL warning: this native histogram metric is not a gauge: "metric" (1:7)`},
+		},
+		"delta() native histogram gauge CounterResetHint": {
+			data: nativeHistogramsWithResetHintsMix,
+			expr: `delta(metric{reset_hint="gauge"}[3m])`,
 		},
 
 		"stdvar() with only floats": {
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2364,6 +2364,29 @@ func TestRateIncreaseAnnotations(t *testing.T) {
 			expectedInfoAnnotations:    []string{},
 		}
 	}
+}
+
+func TestDeltaAnnotations(t *testing.T) {
+	nativeHistogramsWithGaugeResetHints := `
+		metric{series="mix-float-nh"} 10 {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:5 count:4 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}}
+		metric{series="mixed-exponential-custom-buckets"} {{schema:0 sum:1 count:1 buckets:[1]}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}} {{schema:0 sum:5 count:4 buckets:[1 2 1]}}
+	`
+
+	testCases := map[string]annotationTestCase{}
+	testCases["delta() over metric with mixed float and nh"] = annotationTestCase{
+		data: nativeHistogramsWithGaugeResetHints,
+		expr: fmt.Sprintf(`delta(metric{series="mix-float-nh"}[1m1s])`),
+		expectedWarningAnnotations: []string{
+			`PromQL warning: encountered a mix of histograms and floats for metric name "metric" (1:7)`,
+		},
+	}
+	testCases["delta() over metric with incompatible schema"] = annotationTestCase{
+		data: nativeHistogramsWithGaugeResetHints,
+		expr: fmt.Sprintf(`delta(metric{series="mixed-exponential-custom-buckets"}[1m1s])`),
+		expectedWarningAnnotations: []string{
+			`PromQL warning: vector contains a mix of histograms with exponential and custom buckets schemas for metric name "metric" (1:7)`,
+		},
+	}
 	runAnnotationTests(t, testCases)
 }
 
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2373,7 +2373,7 @@ func TestDeltaAnnotations(t *testing.T) {
 	`
 
 	testCases := map[string]annotationTestCase{}
-	testCases["delta() over metric with mixed float and nh"] = annotationTestCase{
+	testCases["delta() over series with mixed floats and native histograms"] = annotationTestCase{
 		data: nativeHistogramsWithGaugeResetHints,
 		expr: `delta(metric{series="mix-float-nh"}[1m1s])`,
 		expectedWarningAnnotations: []string{
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2801,7 +2801,7 @@ func TestCompareVariousMixedMetricsVectorSelectors(t *testing.T) {
 
 	for _, labels := range labelCombinations {
 		labelRegex := strings.Join(labels, "|")
-		for _, function := range []string{"rate", "increase", "changes", "resets", "deriv", "irate", "idelta"} {
+		for _, function := range []string{"rate", "increase", "changes", "resets", "deriv", "irate", "idelta", "delta"} {
 			expressions = append(expressions, fmt.Sprintf(`%s(series{label=~"(%s)"}[45s])`, function, labelRegex))
 			expressions = append(expressions, fmt.Sprintf(`%s(series{label=~"(%s)"}[1m])`, function, labelRegex))
 			expressions = append(expressions, fmt.Sprintf(`sum(%s(series{label=~"(%s)"}[2m15s]))`, function, labelRegex))
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2364,6 +2364,7 @@ func TestRateIncreaseAnnotations(t *testing.T) {
 			expectedInfoAnnotations:    []string{},
 		}
 	}
+	runAnnotationTests(t, testCases)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/streamingpromql/testdata/upstream/functions.test
charleskorn, line 1: Could you please add some tests for `delta` over native histograms?

It'd be good include native histograms with custom buckets, as well as scenarios that trigger the various annotations. (Have a look at our tests for `rate` and `increase` for some inspiration, and look at `TestAnnotations` in `engine_test.go`.)

charleskorn, line 1: Could you please also add some tests for the case where the value (float or histogram) increases and decreases during the range, to ensure that we're correctly ignoring resets?

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -135,18 +145,24 @@ func histogramRate(isRate bool, step types.RangeVectorStepData, hHead []promql.H
...
 		delta = delta.CopyToSchema(desiredSchema)
 	}
 
-	val := calculateHistogramRate(isRate, step.RangeStart, step.RangeEnd, rangeSeconds, firstPoint, lastPoint, delta, hCount)
-	return val, err
+	if isCounter {
+		val := calculateHistogramRate(isRate, step.RangeStart, step.RangeEnd, rangeSeconds, firstPoint, lastPoint, delta, hCount)
+		return val, err
+	}
+	return delta, err
```
charleskorn, line 162: I'm not sure this is quite right - [the docs for `delta`](https://prometheus.io/docs/prometheus/latest/querying/functions/#delta) say:

> `delta(v range-vector)` calculates the difference between the first and last value of each time series element in a range vector `v`, returning an instant vector with the given deltas and equivalent labels. The delta is extrapolated to cover the full time range as specified in the range vector selector...

I see two issues with the implementation here:

* we're not doing extrapolation
* we're considering resets in `delta`, but we want to consider just the first and last value

I wonder if it would be clearer and simpler for `delta` to not use `rate` above, and instead be implemented separately and call `calculateFloatRate` and `calculateHistogramRate` to perform extrapolation.

lamida: Actually I was looking at how prometheus engine was implementing delta. But let me check at this again and try to apply your suggestion.

chencs:

jhesketh (commented):
I agree with the separation of rate from delta. However, I also think this is how it's done upstream so I'm okay either way.

We should add some more tests to our tests. Specifically around native histograms and mixed data types. eg: what is the delta when the first and last sample are different types. What about having a NH in the middle of a vector of floats etc..

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -122,15 +132,20 @@ func histogramRate(isRate bool, hCount int, hHead []promql.HPoint, hTail []promq
 		return nil
 	}
 
-	err = accumulate(hHead)
-	if err != nil {
-		return nil, err
+	if isCounter {
```
charleskorn, line 135: Should `previousValue` and `accumulate` move inside this `if`? They seem to only be used here.

pkg/streamingpromql/testdata/upstream/functions.test
lamida: Added in 4b170e7e62c7fdafa9713f94664c1317e7a47056

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: This comment is not relevant anymore.

pkg/streamingpromql/engine_test.go
```diff
@@ -2073,6 +2078,16 @@ func TestAnnotations(t *testing.T) {
...
+		"delta() native histogram unknown CounterResetHint": {
+			data:                       nativeHistogramsWithResetHintsMix,
+			expr:                       `delta(metric{reset_hint="unknown"}[3m])`,
+			expectedWarningAnnotations: []string{`PromQL warning: this native histogram metric is not a gauge: "metric" (1:7)`},
+		},
+		"delta() native histogram gauge CounterResetHint": {
+			data: nativeHistogramsWithResetHintsMix,
+			expr: `delta(metric{reset_hint="gauge"}[3m])`,
+		},
+
```
charleskorn, line 2090: Could you please also add test cases where only the first or last points are not gauges?

pkg/streamingpromql/testdata/upstream/functions.test
charleskorn, line 1: There are a couple of new `delta` tests that were added in the Prometheus upgrade in https://github.com/grafana/mimir/pull/10383 - could you please rebase to include those changes and enable the new tests?

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -620,6 +630,50 @@ eval range from 0 to 8m step 1m irate(metric[3m1s])
...
+  {case="all Inf"} _ NaN NaN NaN NaN NaN
+  {case="all NaN"} _ NaN NaN NaN NaN NaN
+  {case="all floats with reset"} _ 9 0 6.033333333333333 -6.033333333333333 6.033333333333333 -6.033333333333333 6.033333333333333 0
+  {case="all nh up down"} _ {{schema:3 counter_reset_hint:gauge buckets:[4.5]}} {{schema:3 counter_reset_hint:gauge buckets:[1.5083333333333333]}} {{schema:3 counter_reset_hint:gauge buckets:[1.5083333333333333]}} {{schema:3 counter_reset_hint:gauge buckets:[-3.033333333333333]}}
+  {case="all nhcb up down"} _ {{schema:-53 custom_values:[5 10] counter_reset_hint:gauge buckets:[4.5]}} {{schema:-53 custom_values:[5 10] counter_reset_hint:gauge}} {{schema:-53 custom_values:[5 10] counter_reset_hint:gauge}} {{schema:-53 custom_values:[5 10] counter_reset_hint:gauge buckets:[-4.55]}}
+
+
+
+
+
```
charleskorn, line 676: [nit] The extra whitespace isn't necessary

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -272,3 +278,74 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+	fHead, fTail := step.Floats.UnsafePoints()
+	fCount := len(fHead) + len(fTail)
+
+	hHead, hTail := step.Histograms.UnsafePoints()
+	hCount := len(hHead) + len(hTail)
+
+	if fCount > 0 && hCount > 0 {
+		// We need either at least two histograms and no floats, or at least two floats and no histograms to calculate a delta.
+		// Otherwise, emit a warning and drop this sample.
+		emitAnnotation(annotations.NewMixedFloatsHistogramsWarning)
```
charleskorn, line 292: Could you please add a test for this case? You should be able to extend or borrow from `TestRateIncreaseAnnotations` in `engine_test.go` for this.

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -272,3 +278,74 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+
+	var lastPoint promql.HPoint
+	if len(hTail) > 0 {
+		lastPoint = hTail[len(hTail)-1]
+	} else {
+		lastPoint = hHead[len(hHead)-1]
+	}
+
+	if firstPoint.H.UsesCustomBuckets() != lastPoint.H.UsesCustomBuckets() {
+		return nil, histogram.ErrHistogramsIncompatibleSchema
```
charleskorn, line 338: Could you please add a test for this case? You should be able to extend or borrow from `TestRateIncreaseAnnotations` in `engine_test.go` for this.

pkg/streamingpromql/engine_test.go
lamida: Added in 19045f9064bc64395723d092d5abda653b678ab8

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Added in b9f89fb9dd2f586bc2a148cb3a04da2afce430b5

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Added in b9f89fb9dd2f586bc2a148cb3a04da2afce430b5

pkg/streamingpromql/testdata/upstream/functions.test
lamida: Rebased

pkg/streamingpromql/engine_test.go
```diff
@@ -2073,6 +2080,26 @@ func TestAnnotations(t *testing.T) {
...
+		"delta() native histogram unknown CounterResetHint": {
+			data:                       nativeHistogramsWithResetHintsMix,
+			expr:                       `delta(metric{reset_hint="unknown"}[3m])`,
+			expectedWarningAnnotations: []string{`PromQL warning: this native histogram metric is not a gauge: "metric" (1:7)`},
+		},
+		"delta() native histogram gauge CounterResetHint": {
+			data: nativeHistogramsWithResetHintsMix,
+			expr: `delta(metric{reset_hint="gauge"}[3m])`,
+		},
+		"delta() native histogram first gauge last unknown CounterResetHint": {
```
charleskorn, line 2092: [nit]

```suggestion
		"delta() with first point having gauge CounterResetHint and last point having unknown CounterResetHint": {
```

pkg/streamingpromql/engine_test.go
```diff
@@ -2073,6 +2080,26 @@ func TestAnnotations(t *testing.T) {
...
+			data:                       nativeHistogramsWithResetHintsMix,
+			expr:                       `delta(metric{reset_hint="gauge-unknown"}[3m])`,
+			expectedWarningAnnotations: []string{`PromQL warning: this native histogram metric is not a gauge: "metric" (1:7)`},
+		},
+		"delta() native histogram first unknown last gauge CounterResetHint": {
```
charleskorn, line 2097: [nit]

```suggestion
		"delta() with first point having unknown CounterResetHint and last point having gauge CounterResetHint": {
```

pkg/streamingpromql/engine_test.go
```diff
@@ -2337,6 +2364,29 @@ func TestRateIncreaseAnnotations(t *testing.T) {
...
+}
+
+func TestDeltaAnnotations(t *testing.T) {
+	nativeHistogramsWithGaugeResetHints := `
+		metric{series="mix-float-nh"} 10 {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:5 count:4 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}}
+		metric{series="mixed-exponential-custom-buckets"} {{schema:0 sum:1 count:1 buckets:[1]}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}} {{schema:0 sum:5 count:4 buckets:[1 2 1]}}
+	`
+
+	testCases := map[string]annotationTestCase{}
+	testCases["delta() over metric with mixed float and nh"] = annotationTestCase{
```
charleskorn, line 2376: [nit]

```suggestion
	testCases["delta() over series with mixed floats and native histograms"] = annotationTestCase{
```

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Note that sprintf removed in 350a0e100af8f0fa2d5e9a45cbda160973a5a897

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Note that sprintf removed in 350a0e100af8f0fa2d5e9a45cbda160973a5a897

pkg/streamingpromql/engine_test.go
```diff
@@ -2337,6 +2364,29 @@ func TestRateIncreaseAnnotations(t *testing.T) {
...
 	}
+}
+
+func TestDeltaAnnotations(t *testing.T) {
+	nativeHistogramsWithGaugeResetHints := `
+		metric{series="mix-float-nh"} 10 {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}} {{schema:-53 sum:5 count:4 custom_values:[5 10] buckets:[1] counter_reset_hint:gauge}}
+		metric{series="mixed-exponential-custom-buckets"} {{schema:0 sum:1 count:1 buckets:[1]}} {{schema:-53 sum:1 count:1 custom_values:[5 10] buckets:[1]}} {{schema:0 sum:5 count:4 buckets:[1 2 1]}}
+	`
+
+	testCases := map[string]annotationTestCase{}
```
charleskorn, line 2375: [nit] Given the test case names are known statically (unlike `TestRateIncreaseAnnotations`), this could be written something like this:

```golang
testCases := map[string]annotationTestCase{
    "delta() over series...": {
        ...
    },
}
```

pkg/streamingpromql/testdata/upstream/functions.test
```diff
@@ -243,10 +243,9 @@ load 5m
...
 
-# Unsupported by streaming engine.
-# eval instant at 20m delta(http_requests[20m])
-# 	{path="/foo"} 200
-# 	{path="/bar"} -200
+eval instant at 20m delta(http_requests[20m])
+	{path="/foo"} 200
+	{path="/bar"} -200
 
 # Unsupported by streaming engine.
```
charleskorn, line 250: Could you please enable the new `delta`-related tests below?

`tools/check-for-disabled-but-supported-mqe-test-cases` will help identify any test cases that might be able to be enabled.

pkg/streamingpromql/engine_test.go
lamida: Updated in 617cfeb9872a90178f8ee5df0cd63544243ba84c

charleskorn (approved):
🚀

pkg/streamingpromql/testdata/upstream/functions.test
lamida: Oh missed that. Updated in e72c50452be2e0fce088f586bac246cec291814b

Issue #5796, opened 2023-08-21
distributor might start serving writes with an empty HA tracker state


dimitarvdimitrov:
#### Describe the bug
 The distributor doesn't wait for the [HA tracker](https://grafana.com/docs/mimir/latest/configure/configure-high-availability-deduplication/) to complete initializing before starting to serve writes. This means that the distributor can start serving writes with an empty HA tracker state. An empty state will lead to accepting any sample from any HA replica.

#### Details

The HA tracker watches the KV store for changes in case any other distributor replica updates the state.

https://github.com/grafana/mimir/blob/9af647e73b2e75cdffc054770850317dac826697/pkg/distributor/ha_tracker.go#L230

The HA tracker also syncs its own state to the KV store periodically, which would trigger a WatchKey event and cause the tracker to update its tracker.

https://github.com/grafana/mimir/blob/9af647e73b2e75cdffc054770850317dac826697/pkg/distributor/ha_tracker.go#L279-L283

However, the first update of the KV store might happen as late as after one `UpdateTimeout`, which is 15 seconds by default. If there are only a few other distributors, then the KV store changes might come near the end of the 15 seconds, which can be enough for the distributor to start serving writes and accept a write from a non-primary replica.

#### How to fix

The distributor can list all clusters' keys upon startup and update its internal cache, similar to how it's done for each individual replica/cluster [here](https://github.com/grafana/mimir/blob/9af647e73b2e75cdffc054770850317dac826697/pkg/distributor/ha_tracker.go#L231-L263). This can be done as the `StartingFn` of the HA tracker service [here](https://github.com/grafana/mimir/blob/9af647e73b2e75cdffc054770850317dac826697/pkg/distributor/ha_tracker.go#L207) so the dependants of the HA tracker don't start before it has synced the state from the KV store.

NickAnge:
Hello @dimitarvdimitrov. I am planning to work on this issue as part of https://github.com/grafana/pir-actions/issues/94. So, I will investigate/implement the cache population in the context of memberlist kv store.  I will assign it to me, but please let me know if you want me to create a separate issue

PR #6991, opened 2023-12-24
Distributor might start serving writes with an empty HA tracke…

TimKotowski:
#### What this PR does
The distributor doesn't wait for the HA tracker to complete initializing before starting to serve writes. This means that the distributor can start serving writes with an empty HA tracker state. An empty state will lead to accepting any sample from any HA replica. Allow fix to preload our cluster cache before writing to replicas. Want to ensure we are writing to the leader replica only.
#### Which issue(s) this PR fixes or relates to

Closes https://github.com/grafana/mimir/issues/5796

#### Checklist

- [x] Tests updated.
- NA Documentation Written
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- NA [about-versioning.md](https://github.com/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9826, opened 2024-11-04
fix(distributor-ha): init cache on startup

NickAnge:
#### What this PR does
This PR adds the initialization of the local cache of the ha_tracker

The investigation if this is needed has been done in a separate PR https://github.com/grafana/mimir/pull/9853

Check [conclusion](https://github.com/grafana/mimir/pull/9853#issue-2643760579) for TLDR

Relates with: https://github.com/grafana/mimir/issues/5796

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/development/mimir-microservices-mode/config/mimir_override.yaml
+++ b/development/mimir-microservices-mode/config/mimir_override.yaml
@@ -1,0 +1,184 @@
+multitenancy_enabled: false
+
+distributor:
+  pool:
+    health_check_ingesters: true
+  ring:
+    kvstore:
+      store: memberlist
+
+ingester_client:
+  grpc_client_config:
+    # Configure the client to allow messages up to 100MB.
+    max_recv_msg_size: 104857600
+    max_send_msg_size: 104857600
+    grpc_compression: gzip
+
+ingester:
+  ring:
+    # We want to start immediately.
+    final_sleep: 0s
+    num_tokens: 512
+    kvstore:
+      store: memberlist
+
+# These memberlist options will be only used if memberlist is activated via CLI option.
+memberlist:
+  join_members:
+    - distributor-1:10000
+  rejoin_interval: 10s
+
+blocks_storage:
+  backend: s3
+
+  tsdb:
+    dir: /tmp/mimir-tsdb-ingester
+    # Note: this value is intentionally set low to create a faster feedback loop
+    # in development. However, setting this lower than 2m can cause the ruler's
+    # write requests to fail with out of bounds errors
+    block_ranges_period: ["2m"]
+    # retention_period must be larger than block_ranges_period and querier.query_store_after
+    retention_period: 15m
+    ship_interval: 1m
+
+    # Always use the PostingsForMatchers() cache in order to exercise it.
+    head_postings_for_matchers_cache_force: true
+    block_postings_for_matchers_cache_force: true
+    block_postings_for_matchers_cache_ttl: 1m
+
+  bucket_store:
+    sync_dir: /tmp/mimir-tsdb-querier
+    sync_interval: 1m
+    # ignore_blocks_within and sync_interval must be small enough for the store-gateways
+    # to discover & load new blocks shipped from the ingesters before they begin to be queried.
+    # With querier.query_store_after: 10m and sync_interval: 1m, anything larger than 2m causes issues.
+    # Slightly larger values for ignore_blocks_within work if sync_interval is reduced.
+    ignore_blocks_within: 2m
+
+    index_cache:
+    # Cache is configured via CLI flags. See docker-compose.jsonnet
+
+    chunks_cache:
+    # Cache is configured via CLI flags. See docker-compose.jsonnet
+
+    metadata_cache:
+    # Cache is configured via CLI flags. See docker-compose.jsonnet
+
+  s3:
+    endpoint:          minio:9000
+    bucket_name:       mimir-tsdb
+    access_key_id:     mimir
+    secret_access_key: supersecret
+    insecure:          true
+
+ruler:
+  ring:
+    heartbeat_period:   5s
+    heartbeat_timeout:  15s
+    kvstore:
+      store: memberlist
+
+  alertmanager_url: http://alertmanager-1:8031/alertmanager,http://alertmanager-2:8032/alertmanager,http://alertmanager-3:8033/alertmanager
+
+ruler_storage:
+  backend: s3
+
+  cache:
+  # Cache is configured via CLI flags. See docker-compose.jsonnet
+
+  s3:
+    bucket_name:       mimir-ruler
+    endpoint:          minio:9000
+    access_key_id:     mimir
+    secret_access_key: supersecret
+    insecure:          true
+
+alertmanager:
+  fallback_config_file: './config/alertmanager.yaml'
+  sharding_ring:
+    replication_factor: 3
+    heartbeat_period: 5s
+    heartbeat_timeout: 15s
+    kvstore:
+      store: memberlist
+
+alertmanager_storage:
+  backend: s3
+  s3:
+    bucket_name:       mimir-alertmanager
+    endpoint:          minio:9000
+    access_key_id:     mimir
+    secret_access_key: supersecret
+    insecure:          true
+
+compactor:
+  data_dir: "/tmp/mimir-compactor"
+  block_ranges: [ 2m, 4m, 8m, 16m ]
+  compaction_interval: 1m
+  compaction_concurrency: 2
+  cleanup_interval: 1m
+  tenant_cleanup_delay: 1m
+  sharding_ring:
+    kvstore:
+      store: memberlist
+
+store_gateway:
+  sharding_ring:
+    replication_factor: 1
+    heartbeat_period:   5s
+    heartbeat_timeout:  15s
+    wait_stability_min_duration: 0
+    kvstore:
+      store: memberlist
+
+frontend:
+  query_stats_enabled: true
+  parallelize_shardable_queries: true
+  cache_results: true
+  shard_active_series_queries: true
+
+  # Uncomment when using "dns" service discovery mode for query-scheduler.
+  # scheduler_address: "query-scheduler:9011"
+
+  results_cache:
+    # Cache is configured via CLI flags. See docker-compose.jsonnet
+    compression: snappy
+
+frontend_worker:
+  response_streaming_enabled: true
+
+  # Uncomment when using "dns" service discovery mode for query-scheduler.
+  # scheduler_address: "query-scheduler:9011"
+
+  # Uncomment to skip query-scheduler and enqueue queries directly in the query-frontend.
+  # frontend_address: "query-frontend:9007"
+
+querier:
+  # query_store_after must be smaller than blocks_storage.tsdb.retention_period
+  query_store_after: 10m
+
+query_scheduler:
+  # Change to "dns" to switch to query-scheduler DNS-based service discovery.
+  service_discovery_mode: "ring"
+
+limits:
+  # Limit max query time range to 31d
+  max_partial_query_length: 744h
+  max_global_exemplars_per_user: 5000
+  query_sharding_total_shards: 16
+  query_sharding_max_sharded_queries: 32
+  ingestion_rate: 50000
+  # expanded OOO window makes it easier to run continuous-test
+  # otherwise catch-up writes are rejected when metamonitoring is on
+  out_of_order_time_window: 1h
+  native_histograms_ingestion_enabled: true
+  cardinality_analysis_enabled: true
+  query_ingesters_within: 20m
+  # HA tracker configuration
+  accept_ha_samples: true
+  ha_cluster_label: ha_cluster
+  ha_replica_label: ha_replica
+  ha_max_clusters: 10
+
+runtime_config:
+  file: ./config/runtime.yaml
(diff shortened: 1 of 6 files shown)


pkg/distributor/ha_tracker.go
```diff
@@ -236,47 +273,58 @@ func (h *haTracker) loop(ctx context.Context) error {
...
+		if !ok {
+			return false
 		}
-
-		// Store the received information into our cache
-		h.electedLock.Lock()
-		h.updateCache(user, cluster, replica)
-		h.electedLock.Unlock()
-		h.electedReplicaPropagationTime.Observe(time.Since(timestamp.Time(replica.ReceivedAt)).Seconds())
+		h.processKVStoreEntry(key, replica)
```
aldernero, line 280: nice, putting this in a separate function makes the purpose of the `loop` much clearer 👍🏻

pkg/distributor/ha_tracker.go
```diff
@@ -236,47 +273,58 @@ func (h *haTracker) loop(ctx context.Context) error {
...
+	// Valid key would look like cluster/replica, and a key without a / such as `ring` would be invalid.
+	if len(segments) != 2 {
+		return
+	}
+
+	user := segments[0]
+	cluster := segments[1]
+
+	if replica.DeletedAt > 0 {
+		h.cleanupDeletedReplica(user, cluster)
```
aldernero, line 300: same comment as above, nice!

aldernero (approved):
LGTM

pkg/distributor/ha_tracker.go
NickAnge: Thank you. Most of this code came from this [PR](https://github.com/grafana/mimir/pull/6991/files). Just addressed comments and refactored some code 🙌

pkg/distributor/ha_tracker.go
```diff
@@ -212,10 +212,47 @@ func newHATracker(cfg HATrackerConfig, limits haTrackerLimits, reg prometheus.Re
...
 }
 
+func (h *haTracker) syncHATrackerStateOnStart(ctx context.Context) error {
+	if !h.cfg.EnableHATracker {
+		return nil
+	}
+	// haTracker holds HATrackerConfig with the prefix.
+	keys, err := h.client.List(ctx, "")
+	if err != nil {
+		return err
```
NickAnge, line 226: After running the local setup in [microservices mode ](https://github.com/grafana/mimir/tree/main/development/mimir-microservices-mode), I noticed that this particular piece of code makes the ruler, and queriers fail because of connection issue to consul. 

**Why ?** 
This happens because both `querier` and `ruler` have `distributor` as a dependency in the code. Which means that they initialize the service in both scenarios. 

**Before - init cache**
Before the initialization it was also  failing but when the distributor was in Running state which didn't make the ruler, querier initialization fail.

**After - Init Cache**
Now, the error occurs earlier when the distributor is in the Starting state, causing both ruler and querier to fail.

**How we can solve it ?**
1. Override the mimir.yaml for ruler, and querier to not initialize the ha_tracker : ✅ 
2. Provide connection to consul, which creates confusion since we can see in metrics the ha_tracker from both ruler and querier, thats why I rejected it ❌

pkg/distributor/ha_tracker.go
```diff
@@ -212,10 +212,47 @@ func newHATracker(cfg HATrackerConfig, limits haTrackerLimits, reg prometheus.Re
...
+	}
+
+	for i := 0; i < len(keys); i++ {
+		if ctx.Err() != nil {
+			return fmt.Errorf("syncing HA tracker state on startup: %w", context.Cause(ctx))
+		}
+
+		val, err := h.client.Get(ctx, keys[i])
+		if err != nil {
+			level.Warn(h.logger).Log("msg", "sync HA state on start: failed to get replica value", "key", keys[i], "err", err)
```
dimitarvdimitrov, line 241: do we want to fail startup here or just continue? The `WatchKey` would ignore an unparsable value. Swallowing the error would also make us resilient to random values in the KV store (maybe it's being shared between multiple applications and someone forgot to configure a prefix for Mimir)

pkg/distributor/ha_tracker.go
```diff
@@ -212,10 +212,47 @@ func newHATracker(cfg HATrackerConfig, limits haTrackerLimits, reg prometheus.Re
...
+			return err
+		}
+
+		desc, ok := val.(*ReplicaDesc)
+		if !ok {
+			level.Error(h.logger).Log("msg", "sync HA state on start: got invalid replica descriptor", "key", keys[i])
```
dimitarvdimitrov, line 247: again on adding the implications to the log line. This is less clear, but maybe mentioning that the value in the key is skipped should be enough.

pkg/distributor/ha_tracker_test.go
```diff
@@ -70,6 +70,53 @@ func checkReplicaTimestamp(t *testing.T, duration time.Duration, c *haTracker, u
 	})
 }
 
+func TestHATrackerCacheSyncOnStart(t *testing.T) {
+	cluster := "c1"
+	replica := "r1"
```
dimitarvdimitrov, line 75: nitpick to make these `const`s. Some editor color variables and constants differently for example

pkg/distributor/ha_tracker_test.go
```diff
@@ -70,6 +70,53 @@ func checkReplicaTimestamp(t *testing.T, duration time.Duration, c *haTracker, u
...
+	assert.NoError(t, err)
+
+	err = services.StopAndAwaitTerminated(context.Background(), c)
+	assert.NoError(t, err)
+
+	replicaTwo := "r2"
+	c, err = newHATracker(HATrackerConfig{
+		EnableHATracker:        true,
+		KVStore:                kv.Config{Mock: mock},
+		UpdateTimeout:          1000 * time.Second,
```
dimitarvdimitrov, line 106: the update timeouts are different. Is this intentional?

pkg/distributor/ha_tracker_test.go
```diff
@@ -70,6 +70,53 @@ func checkReplicaTimestamp(t *testing.T, duration time.Duration, c *haTracker, u
 	})
 }
 
+func TestHATrackerCacheSyncOnStart(t *testing.T) {
```
dimitarvdimitrov, line 73: just double-checking, but this test passes on `main` too. You know that, right? My understanding is that it's just asserting that the tracker works on distributor (re)starts without losing state, not the code you just added.

Can you think of a way to test the new code too? Maybe somehow making sure that given a running HA tracker and querying an already cached replica, the HA tracker does not query the KV store again upon checkReplica?

dimitarvdimitrov (commented):
Good job on the investigation! This looks like an improvement we could use 👍 I left mostly small suggestions and questions

development/mimir-microservices-mode/config/mimir_override.yaml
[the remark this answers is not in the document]
NickAnge: No, its not by mistake. This `mimir_override.yaml`, is used in rulers and queries in order to disable the ha_tracker. Reasoning explained [here](https://github.com/grafana/mimir/pull/9826#discussion_r1840176310)

I updated also the jsonnet file, which is the source of truth for the docker-compoase.yaml  ([088b23c](https://github.com/grafana/mimir/pull/9826/commits/088b23c1976d225e3654258284198286bdf9332d)) 

Let me know if it makes sense

pkg/distributor/ha_tracker.go
```diff
@@ -212,10 +212,47 @@ func newHATracker(cfg HATrackerConfig, limits haTrackerLimits, reg prometheus.Re
...
-	t.Service = services.NewBasicService(nil, t.loop, nil)
+	t.Service = services.NewBasicService(t.syncHATrackerStateOnStart, t.loop, nil)
 	return t, nil
 }
 
+func (h *haTracker) syncHATrackerStateOnStart(ctx context.Context) error {
+	if !h.cfg.EnableHATracker {
+		return nil
+	}
+	// haTracker holds HATrackerConfig with the prefix.
```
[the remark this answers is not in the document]
NickAnge: Thank you. Removed it at [253c56a](https://github.com/grafana/mimir/pull/9826/commits/253c56a2f0ce5cc09dcb48cc1eec2273eabdcf3a)

pkg/distributor/ha_tracker.go
NickAnge: Yeah makes sense. I have updated it at [253c56a](https://github.com/grafana/mimir/pull/9826/commits/253c56a2f0ce5cc09dcb48cc1eec2273eabdcf3a). Please let me know what you. think

pkg/distributor/ha_tracker.go
```diff
@@ -212,10 +212,47 @@ func newHATracker(cfg HATrackerConfig, limits haTrackerLimits, reg prometheus.Re
...
+	keys, err := h.client.List(ctx, "")
+	if err != nil {
+		return err
+	}
+
+	if len(keys) == 0 {
+		level.Warn(h.logger).Log("msg", "sync HA state on start: no keys for HA tracker prefix", "err", err)
```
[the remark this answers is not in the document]
NickAnge: Yeah makes sense. I have updated it at [253c56a](https://github.com/grafana/mimir/pull/9826/commits/253c56a2f0ce5cc09dcb48cc1eec2273eabdcf3a). Please let me know what you. think

pkg/distributor/ha_tracker.go
NickAnge: My thought here was that if we have received the key, we should have a respective value, otherwise we should fail 🤔 

However I think it makes sense to continue and not fail here for resilient purposes as you mentioned. I will updated it

pkg/distributor/ha_tracker_test.go
NickAnge: Yes, the test passes on the main branch as well. You are  right.

I'm struggling to find a way to isolate and test just this function 🤔. When a new ha_tracker is created, the sync* function gets called automatically, initializing the cache. Without this function, checkReplica would still update the cache.

This is why I mentioned that the addition doesn’t add much value for initializing the cache. It justs cause initialization to happen earlier. 

I'll keep looking for alternatives, but any suggestions are very welcome!

pkg/distributor/ha_tracker_test.go
NickAnge: Fixed [2e808ec](https://github.com/grafana/mimir/pull/9826/commits/2e808ec33b59399794bf624b2c16ea04fb8b0cf5). Thank you

pkg/distributor/ha_tracker_test.go
NickAnge: Fixed [2e808ec](https://github.com/grafana/mimir/pull/9826/commits/2e808ec33b59399794bf624b2c16ea04fb8b0cf5). Thank you

development/mimir-microservices-mode/config/mimir_override.yaml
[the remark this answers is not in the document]
dimitarvdimitrov: that's perfect 👌

pkg/distributor/ha_tracker_test.go
dimitarvdimitrov: some ideas below, _but i also don't want to block this PR on it. You can merge this and then follow-up with more tests or continue the discussion_

you can implement a decorator `kv.Client` and pass that the KV store instead of the consul one here: 

https://github.com/grafana/mimir/blob/c7e1819b5b5c622ad50317a06e091f1c20116f09/pkg/distributor/ha_tracker_test.go#L88


in your implementation you can delegate to a consul client to do the heavy-lifting of a KV-store, but note when calls are being made. For example, you can assert that after startup, no calls to `Get` or `CAS` are made (within the update timeout for example).

That way you assert that the warmup during startup was effective and we didn't go to the KV store because of a cold cache.

pkg/distributor/ha_tracker_test.go
```diff
@@ -70,6 +70,54 @@ func checkReplicaTimestamp(t *testing.T, duration time.Duration, c *haTracker, u
...
+
+	var c *haTracker
+	var err error
+	var now time.Time
+
+	codec := GetReplicaDescCodec()
+	kvStore, closer := consul.NewInMemoryClient(codec, log.NewNopLogger(), nil)
+	t.Cleanup(func() { assert.NoError(t, closer.Close()) })
+
+	mock := kv.PrefixClient(kvStore, "prefix")
```
dimitarvdimitrov, line 85: we don't need the preifxed client, do we?

dimitarvdimitrov (approved):
LGTM, feel free to merge if you want while we keep thinking about the test. Nice work 💪

pkg/distributor/ha_tracker_test.go
NickAnge: You are right. Fixed it [2e20498](https://github.com/grafana/mimir/pull/9826/commits/2e204988d140eefe7d1459758b85afffc4803882)

NickAnge: I am trying to create a mock client and followed the approach from other tests. However might not needed 🤔 . When I remove it though tests are failing. This code might be replaced at the next PR in favor of [this](https://github.com/grafana/mimir/pull/9826#discussion_r1841194320) .

pkg/distributor/ha_tracker_test.go
NickAnge: Related PR: https://github.com/grafana/mimir/pull/9905

pkg/distributor/ha_tracker_test.go
NickAnge: Related PR: https://github.com/grafana/mimir/pull/9905

PR #9828, opened 2024-11-04
fix marshaling error on invalid label value

replay:
This makes sure that when a label value fails validation due to invalid UTF-8 characters we don't return these invalid characters in the error, because including them can break the rendering of the error that should be returned.


andyasp (approved):
LGTM

PR #9838, opened 2024-11-05
Do not buffer records in the Kafka client when fetch concurrency is enabled

pracucci:
#### What this PR does

Few days ago, @gotjosh and I discovered that when the Kafka concurrent fetch is enabled (either during startup only or startup + ongoing) the Kafka client buffers records anyway internally. This is done automatically as soon as you configure `kgo.ConsumePartitions()`.

This is not a critical issue, but it's annoying because of a couple of reasons. When ongoing concurrency is enabled, these records buffered by the Kafka client will stay there forever. They consume a bit of memory (not much), and they also cause the `cortex_ingest_storage_reader_buffered_fetch_records_total` to continuously report these buffered records, like if some records are never processed by the ingester (these buffered records are not processed, but we don't even want to process them because when concurrent fetch is enabled we do manually issue fetch requests and we don't read records from Kafka client buffer).

In this PR I propose a change to avoid having the Kafka client buffering records when concurrent fetch is enabled. To avoid buffering I'm using `PauseFetchPartitions()` to pause the consumption. I've added some code comments to explain in details why we keep calling `AddConsumePartitions()` even if we're in control of the fetches (please take a look at the code).

Note to reviewers:
- Enable "hide whitespace changes" to simplify the tests review

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/development/mimir-ingest-storage/config/mimir.yaml
+++ b/development/mimir-ingest-storage/config/mimir.yaml
@@ -81,7 +81,6 @@ ruler_storage:
 
 query_scheduler:
   service_discovery_mode: "ring"
-  prioritize_query_components: true
 
 alertmanager:
   data_dir: /data/alertmanager
--- a/pkg/storage/ingest/reader.go
+++ b/pkg/storage/ingest/reader.go
@@ -153,10 +153,12 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 		r.consumedOffsetWatcher.Notify(lastConsumedOffset)
 	}
 
-	r.client, err = r.newKafkaReader(kgo.NewOffset().At(startOffset))
+	// Create a Kafka client without configuring any partition to consume (it will be done later).
+	r.client, err = NewKafkaReaderClient(r.kafkaCfg, r.metrics.kprom, r.logger)
 	if err != nil {
 		return errors.Wrap(err, "creating kafka reader client")
 	}
+
 	r.committer = newPartitionCommitter(r.kafkaCfg, kadm.NewClient(r.client), r.partitionID, r.consumerGroup, r.logger, r.reg)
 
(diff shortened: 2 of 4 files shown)


development/mimir-ingest-storage/config/mimir.yaml
```diff
@@ -81,7 +81,6 @@ ruler_storage:
 
 query_scheduler:
   service_discovery_mode: "ring"
-  prioritize_query_components: true
```
pracucci, line 84: Note to reviewers: this config option doesn't exist anymore.

pkg/storage/ingest/reader.go
```diff
@@ -266,25 +288,37 @@ func (r *PartitionReader) switchToOngoingFetcher(ctx context.Context) {
 			// This method has been called before, no need to switch the fetcher.
 			return
```
pracucci, line 289: @gotjosh I haven't touched this part of the code, but it's it a bug? Look above. We always call `r.fetcher.Stop()` and then we check if the fetcher has changed. If it hasn't changed we exit. Shouldn't we move the `Stop()` after the `if`?

gotjosh (approved):
LGTM

pkg/storage/ingest/reader.go
gotjosh: Checking.

gotjosh: It's technically not a bug because the method just no-ops [link](https://github.com/grafana/mimir/blob/main/pkg/storage/ingest/reader.go#L126) in the partition reader, but I think I agree it can be better structured as it can easily turn into a future bug if we decide to do something in that method now.

pracucci: I will fix it in a follow up PR to keep the change separated.

pracucci: Follow up PR: https://github.com/grafana/mimir/pull/9850

PR #9840, opened 2024-11-05
Alerting: add new type to support both primitive and string ints

titolins:
#### Context
- Helm doesn't handle big ints well - we should support using strings
- Some tenants are already hitting the new config limits we've added but now we can't increase them using configmaps because helm will convert them to scientific notation
- For flags this is not an issue
- I've merged this PR earlier, but it caused AM pods to crash since the go-yaml doesn't handle the string conversion automatically: https://github.com/grafana/deployment_tools/pull/189045
- `IntString` should allow us to use both a string or an integer
- Use that for config/state size limits
- Tested the flag parsing is working by running locally (we don't have a unit test for that setup and doing so requires a bit tinkering)
Screenshot 2024-11-05 at 19 26 47


titolins:
> Consider putting the new type here: https://github.com/grafana/dskit/tree/main/flagext

Ah, that's a lot better. Tks 🙏 

Converting this to a draft then, will adjust once the new type is added to dskit 👍

PR #9842, opened 2024-11-05
[ingester] Optionally delay circuit breaker activation until after traffic starts

pr00se:
#### What this PR does

Change the implementation of the `--initial-delay` option for ingester circuit breakers as follows:
- Start the activation delay with the first call to `tryAcquirePermit()` rather than the call to `activate()`. This lets us ignore the first `initial_delay` of traffic, regardless of how long after activation it arrives.
- If `activate()` is called during the activation delay, don't schedule another delayed activation
- If `deactivate()` is called during the activation delay, cancel any pending activation

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1295,9 +1295,9 @@ push_circuit_breaker:
   # CLI flag: -ingester.push-circuit-breaker.cooldown-period
   [cooldown_period: <duration> | default = 10s]
 
-  # (experimental) How long the circuit breaker should wait between an
-  # activation request and becoming effectively active. During that time both
-  # failures and successes will not be counted.
+  # (experimental) How long after an initial request an activated circuit
+  # breaker should wait before becoming effectively active. During this time
+  # neither failures nor successes will not be counted.
```
tacole02, line 1300:
```suggestion
  # (experimental) Duration, in seconds, after an initial request an activated circuit
  # breaker should wait before becoming effectively active. During this time,
  # neither failures nor successes are counted.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1332,9 +1332,9 @@ read_circuit_breaker:
   # CLI flag: -ingester.read-circuit-breaker.cooldown-period
   [cooldown_period: <duration> | default = 10s]
 
-  # (experimental) How long the circuit breaker should wait between an
-  # activation request and becoming effectively active. During that time both
-  # failures and successes will not be counted.
+  # (experimental) How long after an initial request an activated circuit
+  # breaker should wait before becoming effectively active. During this time
+  # neither failures nor successes will not be counted.
   # CLI flag: -ingester.read-circuit-breaker.initial-delay
```
tacole02, line 1338: Same suggestion as above here.

tacole02 (approved):
Thanks for updating these docs! I massaged the wording a little bit.

docs/sources/mimir/configure/configuration-parameters/index.md
pr00se: addressed in https://github.com/grafana/mimir/pull/9842/commits/24a3b0db894298d97e53aab419553c6d71074ec9

docs/sources/mimir/configure/configuration-parameters/index.md
pr00se: addressed in https://github.com/grafana/mimir/pull/9842/commits/24a3b0db894298d97e53aab419553c6d71074ec9

pkg/ingester/circuitbreaker.go
```diff
@@ -195,47 +203,54 @@ func (cb *circuitBreaker) tryRecordFailure(err error) bool {
...
+
 	level.Info(cb.logger).Log("msg", "deactivating circuit breaker", "requestType", cb.requestType)
-	cb.activating.Store(activationIdle)
-	cb.active.Store(false)
+	cb.activationNeeded.Store(false)
+	cb.state.Store(circuitBreakerInactive)
+
+	// cancel any pending activation
+	if cb.activationTimer != nil {
+		cb.activationTimer.Stop()
```
duricanikolic, line 250: I guess here we rely on the fact that within `time.AfterFunc()` we actually set `cb.activationTimer` to `nil`. 
Are we sure it can never happen that we reach this line while `time.AfterFunc()` is executing?

I personally prefer usage of `time.NewTimer()` because then `time.Stop()`'s result (`true` or `false`) clearly explains what happened. Here `false` means that the timer expired, and the function passed to `time.AfterFun()` was started in a separate go routine, but doesn't say anything about whether the function was completed or not. 

Anyway, if we opt for this approach, since it is crucial to execute `cb.activationTimer = nil`, I would put a comment on both lines that execute it (251 and 284) to highlight that those lines must not be deleted or moved from those places, because that would break the logic.

pkg/ingester/circuitbreaker_test.go
```diff
@@ -4,6 +4,7 @@ package ingester
 
 import (
 	"context"
+	"errors"
```
duricanikolic, line 7: Do you think it is feasible to add a test that would simulate what happens if we:
- at timestamp `ts`: schedule a cb activation with delay `3*d`
- at timestamp `ts + d`: deactivate the cb
- at timestamp `ts + 2d`: schedule another cb activation with delay `3*d`
- ensure that the cb is not active before timestamp `ts + 5d`?

duricanikolic (commented):
Left a couple of nits

pkg/ingester/circuitbreaker_test.go
pr00se: added in https://github.com/grafana/mimir/pull/9842/commits/c67e4e3402eb087e18114c4249aa5fef86b2f54e

duricanikolic (approved):
LGTM
Thank you @pr00se

PR #9844, opened 2024-11-06
Upgrade to latest mimir-prometheus with Prometheus 3.0 changes

aknuds1:
#### What this PR does
Upgrade to mimir-prometheus@fd7e39a8798e, which [syncs in Prometheus 3.0 changes](https://github.com/grafana/mimir-prometheus/pull/741).

NB: I add a `go-kit/log` to `slog` adapter, since Prometheus now uses the latter.

TODO:

- [x] Add to changelog (@charleskorn suggests referencing the [v3 migration guide](https://prometheus.io/docs/prometheus/3.0/migration/#promql))
- [x] Finalize [holt-winters alias](https://github.com/grafana/mimir-prometheus/pull/743)
- [x] Look over Mimir changes, to make sure they are sufficiently reviewed

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


bboreham:
To address one small set of test failures: https://github.com/grafana/mimir/pull/9854
(The number changes because Prometheus 3.0 changed a window like `rate ... [1m]` from including both ends to excluding the left-hand end (it is "left-open")).

pkg/util/log/slogadapter.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (h goKitHandler) WithGroup(name string) slog.Handler {
+	if name == "" {
+		return h
+	}
+
+	h.group = name
+	return h
+}
```
tjhop, line 98: I'm not sure what kind of tolerances are required for mimir, but I believe the current implementation may be a bit naive. For instance, the `WithGroup` docs on the slog.Handler state:
> 	// WithGroup returns a new Handler with the given group appended to
>	// the receiver's existing groups.

It is indeed possible to have more than 1 group; here's output from a contrived example ([playground](https://go.dev/play/p/HERFt3YH3ff)):
```bash
time=2009-11-10T23:00:00.000Z level=INFO msg=test g1.g2.hello=world
time=2009-11-10T23:00:00.000Z level=INFO msg=test g1.hello=world g1.hello=world
time=2009-11-10T23:00:00.000Z level=INFO msg=test g1.hello=world g1.g2.hello=world
time=2009-11-10T23:00:00.000Z level=INFO msg=test hello=world hello=world hello=world
```

However, attempting any similar log calls with a logger produced by `SlogFromGoKit()` panics in the implementation of `Handle()`

aknuds1: @tjhop yeah, admittedly I was uncertain about the `WithGroup` implementation, thanks for producing a failing example. Are you able to identify any problems with the other `goKitHandler` methods?

tjhop: Seems I might've been too hasty in reading the panic message, it actually seems the panic was related to me half-assing the mock logger calls when first playing with this. Better mocking out the same contrived log statements as that playground example, I get this from the handler:

```
level=info msg=test g2="[hello=world]"
level=info g1="[hello=world]" msg=test hello=world
level=info g1="[hello=world]" msg=test g2="[hello=world]"
level=info hello=world hello=world msg=test hello=world
```

A few thoughts:
- slog includes the time by default -- currently the implementation doesn't forward the timestamp from the record along into the kv pairs that get passed to `logger.Log()`
- ordering is a bit different, but given the purpose of structured logging, I'd argue it's fine
-  like we already acknowledged, grouping needs some adjustment to properly nest. this patch gets us a bit closer, but still needs further tweaking:
    ```
    ~/go/src/github.com/grafana/mimir (arve/upgrade-mimir-prometheus [ U ]) -> gd pkg/util/log/slogadapter.go
    diff --git a/pkg/util/log/slogadapter.go b/pkg/util/log/slogadapter.go
    index 66f084dce3..381a7cdcae 100644
    --- a/pkg/util/log/slogadapter.go
    +++ b/pkg/util/log/slogadapter.go
    @@ -93,6 +93,11 @@ func (h goKitHandler) WithGroup(name string) slog.Handler {
                    return h
            }

    -       h.group = name
    +       if h.group == "" {
    +               h.group = name
    +       } else {
    +               h.group = h.group + "." + name
    +       }
    +
            return h
     }
    ```

    ```
    level=info msg=test g1.g2="[hello=world]"
    level=info g1="[hello=world]" msg=test hello=world
    level=info g1="[hello=world]" msg=test g2="[hello=world]"
    level=info hello=world hello=world msg=test hello=world
    ```

tjhop: @aknuds1 I think I have something pretty functional -- mind if I PR against this branch?

aknuds1: @tjhop feel free :)

charleskorn:
MQE changes are in https://github.com/grafana/mimir/pull/9948.

aknuds1:
> MQE changes to unblock this are in https://github.com/grafana/mimir/pull/9948.

Thanks @charleskorn - included your changes there into this PR.

jhesketh (approved):
I've reviewed the changes to `pkg/streamingpromql` and those lgtm. I *have not reviewed* in detail the rest of the changes, but they also look good at a glance.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 * [CHANGE] Ingester: remove experimental flags `-ingest-storage.kafka.ongoing-records-per-fetch` and `-ingest-storage.kafka.startup-records-per-fetch`. They are removed in favour of `-ingest-storage.kafka.max-buffered-bytes`. #9906
 * [CHANGE] Ingester: Replace `cortex_discarded_samples_total` label from `sample-out-of-bounds` to `sample-timestamp-too-old`. #9885
 * [CHANGE] Ruler: the `/prometheus/config/v1/rules` does not return an error anymore if a rule group is missing in the object storage after been successfully returned by listing the storage, because it could have been deleted in the meanwhile. #9936
+* [CHANGE] Querier: The `.` pattern in regular expressions in PromQL matches newline characters. With this change regular expressions like `.*` match strings that include `\n`. To maintain the old behaviour, you will have to change regular expressions by replacing all `.` patterns with `[^\n]`, e.g. `foo[^\n]*`. #9844
```
colega, line 22: How about something like this?
(please also apply to the next changelog line)
```suggestion
* [CHANGE] Querier: The `.` pattern in regular expressions in PromQL matches newline characters. With this change regular expressions like `.*` match strings that include `\n`. To maintain the old behaviour, you will have to change regular expressions by replacing all `.` patterns with `[^\n]`, e.g. `foo[^\n]*`. This upgrades PromQL compatibility from Prometheus 2.0 to 3.0. #9844
```

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
...
+* [CHANGE] Querier: Lookback and range selectors are left open and right closed (previously left closed and right closed). This change affects queries when the evaluation time perfectly aligns with the sample timestamps. For example assume querying a timeseries with evenly spaced samples exactly 1 minute apart. Previously, a range query with `5m` would usually return 5 samples. But if the query evaluation aligns perfectly with a scrape, it would return 6 samples. Now, queries like this will always return 5 samples. #9844
```
colega, line 23: Would need a better wording from @tacole02, but I'd replace periods with commas around "but...":
```suggestion
* [CHANGE] Querier: Lookback and range selectors are left open and right closed (previously left closed and right closed). This change affects queries when the evaluation time perfectly aligns with the sample timestamps. For example assume querying a timeseries with evenly spaced samples exactly 1 minute apart. Previously, a range query with `5m` would usually return 5 samples, or 6 samples if the query evaluation aligns perfectly with a scrape. Now, queries like this will always return 5 samples. #9844
```

pkg/util/promqlext/promqlext.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+// ExtendPromQL enriches PromQL with Mimir extensions.
+func ExtendPromQL() {
+	// Keep an alias for users using holt_winters, even though dropped in Prometheus v3.
+	promql.FunctionCalls["holt_winters"] = promql.FunctionCalls["double_exponential_smoothing"]
+	parser.Functions["holt_winters"] = parser.Functions["double_exponential_smoothing"]
+	parser.Functions["holt_winters"].Experimental = false
```
colega, line 15: We're not removing `double_exponential_smoothing` from Mimir, so it's a new function for Mimir users, so it should be mentioned in the changelog.

pkg/util/promqlext/promqlext.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+// ExtendPromQL enriches PromQL with Mimir extensions.
+func ExtendPromQL() {
+	// Keep an alias for users using holt_winters, even though dropped in Prometheus v3.
+	promql.FunctionCalls["holt_winters"] = promql.FunctionCalls["double_exponential_smoothing"]
```
colega, line 13: Should we start a deprecation path for this function?

aknuds1: I don't know, I don't think we discussed it. Could you remind me of how to do so?

pkg/util/promqlext/promqlext.go
aknuds1: Added it to changelog, thanks.

pkg/util/promqlext/promqlext.go
aknuds1: Discussed offline that we'll discuss eventual deprecation after this PR.

colega: We have a list of deprecated features in `about-versioning.md`. Anyway, not blocking this PR on that.

aknuds1: Thanks @colega!

PR #9845, opened 2024-11-06
use dskit flagext.Bytes for sizes

titolins:
#### What this PR does
- Changes the type for the `alertmanager.max-grafana-state-size-bytes` and `alertmanager.max-grafana-config-size-bytes` flags / overrides to `dskit` [flagext.Bytes](https://github.com/grafana/dskit/blob/main/flagext/bytes.go#L13)
- This fixes the issue we have with big ints being converted to scientific notation, but also provides better readability
- We're aware this is a breaking change, check the linked deployment tools PR for how we're planning to release this

### DIFF
--- a/cmd/mimir/help-all.txt.tmpl
+++ b/cmd/mimir/help-all.txt.tmpl
@@ -251,9 +251,9 @@ Usage of ./cmd/mimir/mimir:
     	Maximum size of the Alertmanager configuration for a tenant. 0 = no limit.
   -alertmanager.max-dispatcher-aggregation-groups int
     	Maximum number of aggregation groups in Alertmanager's dispatcher that a tenant can have. Each active aggregation group uses single goroutine. When the limit is reached, dispatcher will not dispatch alerts that belong to additional aggregation groups, but existing groups will keep working properly. 0 = no limit.
-  -alertmanager.max-grafana-config-size-bytes int
+  -alertmanager.max-grafana-config-size-bytes value
     	Maximum size of the Grafana Alertmanager configuration for a tenant. 0 = no limit.
-  -alertmanager.max-grafana-state-size-bytes int
+  -alertmanager.max-grafana-state-size-bytes value
     	Maximum size of the Grafana Alertmanager state for a tenant. 0 = no limit.
   -alertmanager.max-recv-msg-size int
     	Maximum size (bytes) of an accepted HTTP request body. (default 104857600)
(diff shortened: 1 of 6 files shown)


pkg/util/validation/limits.go
```diff
@@ -375,9 +375,11 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 		l.NotificationRateLimitPerIntegration = NotificationRateLimitMap()
 	}
 	f.Var(&l.NotificationRateLimitPerIntegration, "alertmanager.notification-rate-limit-per-integration", "Per-integration notification rate limits. Value is a map, where each key is integration name and value is a rate-limit (float). On command line, this map is given in JSON format. Rate limit has the same meaning as -alertmanager.notification-rate-limit, but only applies for specific integration. Allowed integration names: "+strings.Join(allowedIntegrationNames, ", ")+".")
-	f.IntVar(&l.AlertmanagerMaxGrafanaConfigSizeBytes, AlertmanagerMaxGrafanaConfigSizeFlag, 0, "Maximum size of the Grafana Alertmanager configuration for a tenant. 0 = no limit.")
+	_ = l.AlertmanagerMaxGrafanaConfigSizeBytes.Set("0B")
```
santihernandezc, line 378: Just FYI, we can just pass "0". Either way should be fine.
```suggestion
	_ = l.AlertmanagerMaxGrafanaConfigSizeBytes.Set("0")
```

santihernandezc (commented):
Tested it, works as expected when strings are passed in.
I tried using ints without units, Mimir won't start. We have to keep in mind that this is a breaking change.

pkg/util/validation/limits.go
titolins: Nice, changed that now 👍

santihernandezc (approved):
LGTM

tacole02 (approved):
LGTM!

cmd/mimir/help-all.txt.tmpl
```diff
@@ -251,9 +251,9 @@ Usage of ./cmd/mimir/mimir:
     	Maximum size of the Alertmanager configuration for a tenant. 0 = no limit.
   -alertmanager.max-dispatcher-aggregation-groups int
     	Maximum number of aggregation groups in Alertmanager's dispatcher that a tenant can have. Each active aggregation group uses single goroutine. When the limit is reached, dispatcher will not dispatch alerts that belong to additional aggregation groups, but existing groups will keep working properly. 0 = no limit.
-  -alertmanager.max-grafana-config-size-bytes int
+  -alertmanager.max-grafana-config-size-bytes value
```
bboreham, line 254: Is "value" clear for end-users?

pkg/util/validation/limits_test.go
```diff
@@ -725,7 +725,6 @@ differentuser:
 
 	for name, tt := range tc {
 		t.Run(name, func(t *testing.T) {
-
```
bboreham, line 728: spurious change

bboreham (approved):
Code change looks ok; the help-string is a little weird.

cmd/mimir/help-all.txt.tmpl
titolins: good point, added a case to handle `flagext.Bytes` now. It does change the descriptor of all flags using it though

pkg/util/validation/limits_test.go
titolins: sry, formatter is a bit too invasive - added it back now 👍

cmd/mimir/help-all.txt.tmpl
titolins: Dropped the commit as discussed in slack. I'll create another PR and propose an appropriate format for that 👍

PR #9846, opened 2024-11-06
support fallback configuration for KEDA autoscaling

hobbsh:
#### What this PR does

Supports a fallback configuration for the KEDA autoscaling configuration in the `mimir-distributed` helm chart, so if/when the metrics endpoint being used to scale becomes unavailable, the ScaledObject will fallback to the configured replica count.

n/a

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -33,6 +33,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [ENHANCEMENT] Add the possibility to create a dedicated serviceAccount for the `alertmanager` component by setting `alertmanager.serviceAcount.create` to true in the values. #9781
 * [BUGFIX] Fix PVC template in AlertManager to not show diff in ArgoCD. #9774
 * [BUGFIX] Fix how `fullnameOverride` is reflected in generated manifests. #9564
+* [ENHANCEMENT] Add support for `fallback` in kedaAutoscaling settings. #9846
 
 ## 5.5.1
 
--- a/operations/helm/charts/mimir-distributed/templates/distributor/distributor-so.yaml
+++ b/operations/helm/charts/mimir-distributed/templates/distributor/distributor-so.yaml
@@ -17,6 +17,10 @@ spec:
       {{- end }}
   maxReplicaCount: {{ .Values.distributor.kedaAutoscaling.maxReplicaCount }}
   minReplicaCount: {{ .Values.distributor.kedaAutoscaling.minReplicaCount }}
+  {{- with .Values.distributor.kedaAutoscaling.fallback }}
+  fallback:
+    {{ .toYaml . | nindent 4 }}
+  {{- end }}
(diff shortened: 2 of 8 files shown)


jhesketh:
Thank you for the contribution!

This will need an entry in `operations/helm/charts/mimir-distributed/CHANGELOG.md`.

I'm also curious on your opinion if a fallback if preferable to just maintaining the current number of replicas given that a deployment is also protected by the minReplicas? Would it not be better to maintain the current replicas and alert when KEDA isn't able to get the metrics? Or is the intention to generally have a fallback end up scaling up a deployment?

hobbsh:
First off, thank you for the engagement!

> Would it not be better to maintain the current replicas and alert when KEDA isn't able to get the metrics?

In my experience, if the distributors aren't scaling, they will quickly OOM and then we get into a catch 22 where we need to scale but can't because metrics can't be retrieved. KEDA created `fallback` for exactly this reason so I would much prefer to use it rather than try to rely on minReplicas, because for one it would lead to additional overhead/cost to do it that way. We use this for some internal services and it works well, so it would be great to have this option for Mimir. So yes, the intention is to have `fallback` scale up the deployment if we have a metrics blip. Obviously, it would be great to have the autoscaling metrics coming from an external source, but we are quite resource strapped and that's not the easiest option.

> This will need an entry in operations/helm/charts/mimir-distributed/CHANGELOG.md

Done, thanks!

chencs:
The CHANGELOG has just been cut to prepare for the next release. Please rebase `main` and eventually move the CHANGELOG entry added / updated in this PR to the top of the operations/helm/charts/mimir-distributed/CHANGELOG.md document. Thanks!

dimitarvdimitrov:
Do I understand correctly that you're using the same Mimir cluster for autoscaling metrics? This has other negative implications too. So if this change is targeted to only so that, we would rather discourage that use case and document the recommendation to use an external prometheus/second Mimir cluster.

Is there another use case for having fallback configs?

hobbsh:
> Do I understand correctly that you're using the same Mimir cluster for autoscaling metrics? This has other negative implications too. So if this change is targeted to only so that, we would rather discourage that use case and document the recommendation to use an external prometheus/second Mimir cluster.
> 
> Is there another use case for having fallback configs?

@dimitarvdimitrov That's correct. We've switched to using Grafana Cloud for just the specific metrics that the distributors use to autoscale, and it has been great. However, we had to override the ScaledObject because specific auth configurations for KEDA are missing. It would be nice if that was supported in the chart - maybe I should submit a PR for that?

dimitarvdimitrov:
> However, we had to override the ScaledObject because specific auth configurations for KEDA are missing. It would be nice if that was supported in the chart - maybe I should submit a PR for that?

🙂 @bjorns163 is working on just this in https://github.com/grafana/mimir/pull/10266



And for the fallback config. We were thinking of improving the docs around autoscaling, so that we discourage using MImir itself for autoscaling Mimir. In the meantime I opened a PR to add docs for this https://github.com/grafana/mimir/pull/10732

PR #9853, opened 2024-11-08
test(distributor): verify need of cache sync with memberlist KV in ha_tracker

NickAnge:
#### What this PR does
In this PR we have added `TestHaTrackerWithMemberListAndWithoutCacheSyncStart`, in order to verify if the cache sync is needed when using [KV store with Memberlist](https://github.com/grafana/dskit/tree/main/kv/memberlist). 

##### Why
When a new KV is created in memberlist its performing the `fastJoinMembersOnStartup` ([Ref](https://github.com/grafana/dskit/blob/a6b453a88040c92083d1c733494e3b0f9bfa1746/kv/memberlist/memberlist_client.go#L552)), which should fill up the KV store based on the state of other Members in the cluster.

With this test we want to check if that function is enough to fill up the KV store of the ha_tracker and then the local cache should have the updated value.

##### Test Case

- Creates a basic `c` ha_tracker with memberlist KV store
- Invokes [checkReplica](https://github.com/grafana/mimir/blob/4195814f3816c48d0e92e14cb205c5b6d3214dbf/pkg/distributor/ha_tracker.go#L425) with r1, which should update the KV store and the local cache
- Creates a second ha_tracker with memberlist KV store
- **The fastJoinMembersOnStartup, should update the KV store state based on the other members, the c one.**
- checkReplica with r2, should fail with error `replicasDidNotMatchError`

### Conclusion

It seems that all the KV client operations are blocking until the underneath KV store is in running state ([Ref](https://github.com/grafana/dskit/blob/a6b453a88040c92083d1c733494e3b0f9bfa1746/kv/memberlist/memberlist_client.go#L111)), for that reason we are certain that when the first CAS operation is performed at [checkReplica](https://github.com/grafana/mimir/blob/e32b5e2531d40677f9793f8c72ed18aa79fcf50b/pkg/distributor/ha_tracker.go#L425). the fastJoinMembersOnStartup will have filled up the KV store and the cache will be updated.

In conclusion, the cache will have a valid state, even with Memberlist, before the first CAS operation. The main advantage of adding syncStart as a startingFn in ha_tracker is that it pre-populates the cache earlier. Additionally, the distributor and ha_tracker won’t be considered in a running state until the underlying KV store is fully operational

**By using the sync cache as the startingFn, we ensure that both the KV store is in a running state and the cache is initialized before ha_tracker enters the running state.**

Relates with  https://github.com/grafana/mimir/issues/5796

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -15,6 +15,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/grafana/dskit/kv/memberlist"
+
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gogo/protobuf/proto"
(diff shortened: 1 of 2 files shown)


pkg/distributor/ha_tracker.go
```diff
@@ -51,6 +52,78 @@ func NewReplicaDesc() *ReplicaDesc {
 	return &ReplicaDesc{}
 }
 
+// Merge merges other replica into this one. Returns a new replica that represents the change,
+// and can be sent out to other clients.
+func (r *ReplicaDesc) Merge(other memberlist.Mergeable, _ bool) (memberlist.Mergeable, error) {
```
NickAnge, line 57: FYI: Merge function is not aligning with the requirements of the interface, needs to re-iterate. As a first step I just wanted to just enable ha_tracker to use memberlist.

NickAnge:
Closing this, since its not needed anymore. It was used only for testing purposes

PR #9855, opened 2024-11-08
fix `MimirIngesterStuckProcessingRecordsFromKafka`

gotjosh:
#### What this PR does

The alert `MimirIngesterStuckProcessingRecordsFromKafka` relied on the metric `cortex_ingest_storage_reader_buffered_fetch_records_total ` provided by the Kafka client to identify wether we had stuck buffers or not.

Now that we've implemented concurrent fetching from Kafka and bypass the client's polling function we needed an equivalent metric when using concurrent fetching. This PR does that; In addition to that - the metric also takes the client's buffered records In case we do use a mixture of non-concurrent fetching and concurrent fetching.

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -307,6 +307,7 @@ func (r *concurrentFetchers) Stop() {
 	close(r.done)
 
 	r.wg.Wait()
+	r.bufferedFetchedRecords.Store(0)
 
 	level.Info(r.logger).Log("msg", "stopped concurrent fetchers", "last_returned_record", r.lastReturnedRecord)
 }
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -685,9 +685,9 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 			bufferedResult = result
 			readyBufferedResults = r.orderedFetches
 
-			// We increase the count of buffered records only for ordered records that we're sure
-			// will not be discarded later, so that we can get an accurate tracking of the records
-			// ready to be polled by PollFetches() but not polled yet.
+			// We increase the count of buffered records only for ordered records, so that we can
+			// get an accurate tracking of the records ready to be polled by PollFetches() but not
+			// polled yet. This mimics the similar tracking done by the franz-go library.
 			r.bufferedFetchedRecords.Add(int64(len(result.Records)))
 
 		case readyBufferedResults <- bufferedResult:
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -223,7 +223,12 @@ type concurrentFetchers struct {
 
 	minBytesWaitTime time.Duration
 
-	orderedFetches     chan fetchResult
+	// orderedFetches is a channel where we write fetches that are ready to be polled by PollFetches().
+	// Since all records must be polled in order, the fetches written to this channel are after
+	// ordering and "deduplication" (in case the same record is fetched multiple times from different
+	// routines).
+	orderedFetches chan fetchResult
+
 	lastReturnedRecord int64
 	startOffsets       *genericOffsetReader[int64]
 
@@ -304,10 +309,17 @@ func (r *concurrentFetchers) BufferedRecords() int64 {
 
 // Stop implements fetcher.
 func (r *concurrentFetchers) Stop() {
-	close(r.done)
+	// Ensure it's not already stopped.
+	select {
+	case _, ok := <-r.done:
+		if !ok {
+			return
+		}
+	default:
+	}
 
+	close(r.done)
 	r.wg.Wait()
-	r.bufferedFetchedRecords.Store(0)
 
 	level.Info(r.logger).Log("msg", "stopped concurrent fetchers", "last_returned_record", r.lastReturnedRecord)
 }
@@ -328,8 +340,6 @@ func (r *concurrentFetchers) PollFetches(ctx context.Context) (kgo.Fetches, cont
 	case <-ctx.Done():
 		return kgo.Fetches{}, ctx
 	case f := <-r.orderedFetches:
-		// At this point, we're guaranteed that the records are not going to be with us anymore.
-		r.bufferedFetchedRecords.Sub(int64(len(f.FetchPartition.Records)))
 		firstUnreturnedRecordIdx := recordIndexAfterOffset(f.Records, r.lastReturnedRecord)
 		r.recordOrderedFetchTelemetry(f, firstUnreturnedRecordIdx, waitStartTime)
 
@@ -514,7 +524,6 @@ func (r *concurrentFetchers) run(ctx context.Context, wants chan fetchWant, logg
 			attemptSpan.SetTag("attempt", attempt)
 
 			f := r.fetchSingle(ctx, w)
-			r.bufferedFetchedRecords.Add(int64(len(f.FetchPartition.Records)))
 			f = f.Merge(previousResult)
 			previousResult = f
 			if f.Err != nil {
@@ -602,18 +611,31 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 		go r.run(ctx, wants, logger, highWatermark)
 	}
 
+	// We need to make sure we don't leak any goroutine given that start is called within a goroutine.
+	defer r.wg.Done()
+
 	var (
 		nextFetch      = fetchWantFrom(startOffset, recordsPerFetch)
 		nextResult     chan fetchResult
 		pendingResults = list.New()
 
-		bufferedResult       fetchResult
-		readyBufferedResults chan fetchResult // this is non-nil when bufferedResult is non-empty
+		// bufferedResult is the next fetch that should be polled by PollFetches().
+		bufferedResult fetchResult
+
+		// readyBufferedResults channel gets continuously flipped between nil and the actual channel
+		// where PollFetches() reads from. This channel is nil when there are no ordered buffered
+		// records ready to be written to the channel where PollFetches(), and is non-nil when there
+		// are some ordered buffered records ready.
+		//
+		// It is guaranteed that this channel is non-nil when bufferedResult is non-empty.
+		readyBufferedResults chan fetchResult
 	)
 	nextFetch.bytesPerRecord = 10_000 // start with an estimation, we will update it as we consume
 
-	// We need to make sure we don't leak any goroutine given that start is called within a goroutine.
-	defer r.wg.Done()
+	// When the fetcher is stopped, buffered records are intentionally dropped. For this reason,
+	// we do reset the counter of buffered records to zero when this goroutine ends.
+	defer r.bufferedFetchedRecords.Store(0)
+
 	for {
 		refillBufferedResult := nextResult
 		if readyBufferedResults != nil {
@@ -657,10 +679,24 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 				continue
 			}
 			nextFetch = nextFetch.UpdateBytesPerRecord(result.fetchedBytes, len(result.Records))
+
+			// We have some ordered records ready to be sent to PollFetches(). We store the fetch
+			// result in bufferedResult, and we flip readyBufferedResults to the channel used by
+			// PollFetches().
 			bufferedResult = result
 			readyBufferedResults = r.orderedFetches
 
+			// We increase the count of buffered records only for ordered records that we're sure
+			// will not be discarded later, so that we can get an accurate tracking of the records
+			// ready to be polled by PollFetches() but not polled yet.
+			r.bufferedFetchedRecords.Add(int64(len(result.Records)))
+
 		case readyBufferedResults <- bufferedResult:
+			// We've written the records to the channel read by PollFetches(). Since the channel
+			// is not buffered, as soon as the write succeed it means PollFetches() has read it,
+			// so we can consider these records polled and decrease the buffered records counter.
+			r.bufferedFetchedRecords.Sub(int64(len(bufferedResult.Records)))
+
 			bufferedResult.finishWaitingForConsumption()
 			readyBufferedResults = nil
 			bufferedResult = fetchResult{}
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -225,8 +225,7 @@ type concurrentFetchers struct {
 
 	// orderedFetches is a channel where we write fetches that are ready to be polled by PollFetches().
 	// Since all records must be polled in order, the fetches written to this channel are after
-	// ordering and "deduplication" (in case the same record is fetched multiple times from different
-	// routines).
+	// ordering.
 	orderedFetches chan fetchResult
 
 	lastReturnedRecord int64
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -320,6 +320,10 @@ func (r *concurrentFetchers) Stop() {
 	close(r.done)
 	r.wg.Wait()
 
+	// When the fetcher is stopped, buffered records are intentionally dropped. For this reason,
+	// we do reset the counter of buffered records here.
+	r.bufferedFetchedRecords.Store(0)
+
 	level.Info(r.logger).Log("msg", "stopped concurrent fetchers", "last_returned_record", r.lastReturnedRecord)
 }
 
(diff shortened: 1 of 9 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -226,14 +229,8 @@ type concurrentFetchers struct {
 
 	// trackCompressedBytes controls whether to calculate MaxBytes for fetch requests based on previous responses' compressed or uncompressed bytes.
 	trackCompressedBytes bool
-}
-
-// Stop implements fetcher
```
gotjosh, line 231: This is just a move as is - these methods above the constructor seem to be oddly placed.

pkg/storage/ingest/fetcher.go
```diff
@@ -312,6 +324,10 @@ func (r *concurrentFetchers) PollFetches(ctx context.Context) (kgo.Fetches, cont
 	case <-ctx.Done():
 		return kgo.Fetches{}, ctx
 	case f := <-r.orderedFetches:
+		// At this point, we're guaranteed that the records are not going to be with us anymore.
+		tr := r.bufferedFetchedRecords.Sub(int64(len(f.FetchPartition.Records)))
+		level.Debug(r.logger).Log("msg", "remove buffered fetched records", "num_records", len(f.FetchPartition.Records), "total_records", tr)
```
pracucci, line 329: [nit] I typically recommend to not keep debug logs for hot path. They have a computational cost even when debug level is disabled.

_Same applies below._

pkg/storage/ingest/fetcher.go
```diff
@@ -48,6 +48,9 @@ type fetcher interface {
 
 	// Stop stops the fetcher.
 	Stop()
+
+	// BufferedRecords returns the number of records that have been fetched but not yet consumed.
+	BufferedRecords() float64
```
pracucci, line 53: [nit]Design wise, this should return `int64` given the number of buffered records is a discrete number. Then we convert to `float64` in the gauge func.

pkg/storage/ingest/fetcher.go
```diff
@@ -443,6 +459,9 @@ func (r *concurrentFetchers) parseFetchResponse(ctx context.Context, startOffset
 		fetchedBytes = sumRecordLengths(partition.Records)
 	}
 
+	nv := r.bufferedFetchedRecords.Add(int64(len(partition.Records)))
```
pracucci, line 462: Looks a bit weird incrementing it in the parse function. Have you considered incrementing it right after the call to `r.fetchSingle()` in `concurrentFetchers.run()`?

pkg/storage/ingest/reader.go
```diff
@@ -110,13 +113,14 @@ func newPartitionReader(kafkaCfg KafkaConfig, partitionID int32, instanceID stri
...
 		newConsumer:                           consumer,
 		consumerGroup:                         kafkaCfg.GetConsumerGroup(instanceID, partitionID),
-		metrics:                               newReaderMetrics(partitionID, reg),
 		consumedOffsetWatcher:                 newPartitionOffsetWatcher(),
 		concurrentFetchersMinBytesMaxWaitTime: defaultMinBytesMaxWaitTime,
 		logger:                                log.With(logger, "partition", partitionID),
 		reg:                                   reg,
 	}
 
+	r.metrics = newReaderMetrics(partitionID, reg, func() float64 { return r.BufferedRecords() })
```
pracucci, line 122: [nit] Following my suggestion to change `BufferedRecords()` to return an `int64`, then we would have to change this to:

```suggestion
	r.metrics = newReaderMetrics(partitionID, reg, func() float64 { return float64(r.BufferedRecords()) })
```

pkg/storage/ingest/reader.go
```diff
@@ -271,6 +294,9 @@ func (r *PartitionReader) run(ctx context.Context) error {
 // switchToOngoingFetcher switches to the configured ongoing fetcher. This function could be
 // called multiple times.
 func (r *PartitionReader) switchToOngoingFetcher(ctx context.Context) {
+	r.fetcherMtx.Lock()
+	defer r.fetcherMtx.Unlock()
```
pracucci, line 298: Here we may the lock for some long period of time, which would block metrics scraping. I've also seen the lock is not taken in `processNextFetches()` and `stopDependencies()` which makes it harder to reason about whether it's required or not. Since `r.client` is not set in the "constructure" but in `start()` to guarantee avoiding any race condition you would actually need a lock around it too.

We need to simplify this locking. I think that we just want to protect is from race conditions on the fetcher and client pointer. We don't need a locking to protect race conditions in the logic, just the pointer. So I think we could add `setClient()`, `getClient()`, `setFetcher()`, `getFetcher()`: these 4 functions just manipulate the fetcher/client pointer wrapped with a lock. Then each time we need to set/get the pointers, we call these functions. No other locking required in the callers.

pracucci (commented):
You improved an unit test to check that the value is 0 when done, but there's no unit test asserting that this value is actually tracked. Can we add a test running for the 4 concurrency variants that test the value is tracked to the buffered number of records? I think we can implement a mocked "pusher" that doesn't process records (it just hangs) so next records are buffered in the client, then we assert on the metric, and finally we release the pusher to end the test.

pkg/storage/ingest/fetcher.go
gotjosh: Oh, TIL.

gotjosh: Done in https://github.com/grafana/mimir/pull/9855/commits/43f6386c4e650ca5b22d651d4ba0a7cb60d9672d

pkg/storage/ingest/reader.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9855/commits/43f6386c4e650ca5b22d651d4ba0a7cb60d9672d

pkg/storage/ingest/fetcher.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9855/commits/43f6386c4e650ca5b22d651d4ba0a7cb60d9672d

pkg/storage/ingest/fetcher.go
gotjosh: I originally had it there - but changed it because it seemed semantically correct here to move it to the first place where we have actually records in memory.

I'll change it back because that makes my original hunch correct that right after `fetchSingle` it is the most logical place where one would go look.

gotjosh: Done in https://github.com/grafana/mimir/pull/9855/commits/17021962c99417bf2c1dcb6afc3d60d399156456

pkg/storage/ingest/reader.go
gotjosh: Thanks for the suggestion - I had the same concern but didn't have any better ideas that would not involve a moving too much code around. Your suggestion of the four functions makes sense.

pkg/storage/ingest/reader.go
```diff
@@ -154,14 +173,15 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
...
 	// Create a Kafka client without configuring any partition to consume (it will be done later).
-	r.client, err = NewKafkaReaderClient(r.kafkaCfg, r.metrics.kprom, r.logger)
+	client, err := NewKafkaReaderClient(r.kafkaCfg, r.metrics.kprom, r.logger)
 	if err != nil {
 		return errors.Wrap(err, "creating kafka reader client")
 	}
+	r.setClient(client)
 
-	r.committer = newPartitionCommitter(r.kafkaCfg, kadm.NewClient(r.client), r.partitionID, r.consumerGroup, r.logger, r.reg)
+	r.committer = newPartitionCommitter(r.kafkaCfg, kadm.NewClient(r.getClient()), r.partitionID, r.consumerGroup, r.logger, r.reg)
```
gotjosh, line 182: I think I don't need to rework these arguments as closures here (e.g. `func() *kgo.Client return { r.getClient() }`) as this only happens within `start` and from my understanding, `start`, `stop` and `run` are guaranteed to never concurrently - but please let me know if I'm wrong.

pkg/storage/ingest/reader.go
gotjosh: Done in https://github.com/grafana/mimir/pull/9855/commits/e14bba531244269d51dde0c0fc0fc47776219516

pracucci (approved):
Nice work! LGTM

pkg/storage/ingest/fetcher.go
```diff
@@ -496,6 +510,7 @@ func (r *concurrentFetchers) run(ctx context.Context, wants chan fetchWant, logg
 			attemptSpan.SetTag("attempt", attempt)
 
 			f := r.fetchSingle(ctx, w)
+			r.bufferedFetchedRecords.Add(int64(len(f.FetchPartition.Records)))
```
dimitarvdimitrov, line 513: if something calls `Stop()` before these records are pushed to `orderedFetches` then the atomic counter will always be positive. Should we decrement the counter if we give up on sending these records?

I'm thinking about the cases in this select which end up returning

```go
			select {
			case <-r.done:
				wantSpan.Finish()
				attemptSpan.Finish()
				close(w.result)
				return
			case w.result <- f:
				previousResult = fetchResult{}
			case <-ctx.Done():
			default:
				if w.startOffset >= w.endOffset {
					// We've fetched all we were asked for the whole batch is ready, and we definitely have to wait to send on the channel now.
					f.startWaitingForConsumption()
					select {
					case <-r.done:
						wantSpan.Finish()
						attemptSpan.Finish()
						close(w.result)
						return
					case w.result <- f:
						previousResult = fetchResult{}
					case <-ctx.Done():
					}
				}
			}
```

at this point you can also try to unify all the places which do cleanup actions

dimitarvdimitrov (commented):
can you explain why we need the mutex around the fetcher and the client?

pracucci:
> can you explain why we need the mutex around the fetcher and the client?

Client is set in the starting function but there's no guarantee metrics couldn't be scraped in the meanwhile, so race condition. Fetcher same, and also could be updated later if we change it between start and ongoing, so race condition.

pkg/storage/ingest/fetcher.go
pracucci: > if something calls Stop() before these records are pushed to orderedFetches then the atomic counter will always be positive

I thought about it too, but do we really care?

- If the stop is called at shutdown, we don't care.
- If the stop is called because we're moving from concurrent fetcher at startup to sequential fetchwe when ongoing, we don't care because the fetcher reference is trashed
- If the stop is called because we're updating the concurrent fetcher config (`Update()`) then we care... so what if we simply reset the buffer in `Stop()` after the `r.wg.Wait()` to keep it simple?

dimitarvdimitrov: > * If the stop is called because we're updating the concurrent fetcher config (`Update()`) then we care... so what if we simply reset the buffer in `Stop()` after the `r.wg.Wait()` to keep it simple?

yeah that's another option 👍

dimitarvdimitrov:
> > can you explain why we need the mutex around the fetcher and the client?
> 
> Client is set in the starting function but there's no guarantee metrics couldn't be scraped in the meanwhile, so race condition. Fetcher same, and also could be updated later if we change it between start and ongoing, so race condition.

😿 go guarantees atomic pointer swaps, so technically we should be safe. Was the race detector complaining?

pracucci:
> 😿 go guarantees atomic pointer swaps, so technically we should be safe. Was the race detector complaining?

Does it? So why `atomic.Pointer` exists if it's already atomic? 🤔

seizethedave:
Whole-word pointer operations are atomic, but their order w/ respect to surrounding code executing is not guaranteed without the atomic stuff. That's what the race detector complains about. Many of the whole-word atomic types boil down to using different assembly instructions that make sure dirty CPU cache entries are invalidated appropriately.

dimitarvdimitrov (approved):
if we can solve this with atomic.Pointer I'd prefer that to a mutex. An atomic forces you to use it in a synchronised way, whereas you have to _remember_ to use the mutex. But I'm aware it might be on the bike-shedding side of the spectrum, so you can merge as-is too

pkg/storage/ingest/fetcher.go
pracucci: I'm adding some more unit tests and I found an issue. The way we currently increase the number of records is wrong because the same record may be concurrently fetched by multiple routines. I'm working on a fix.

pracucci:
@dimitarvdimitrov I reworked this PR, doing the following changes:

1. Changed how buffered records are tracked. Now they track only the ordered records that are ready to be sent to PollFetches(). The counter is manipulated exclusively in the goroutine run by `start()`, which I think makes the logic easier to follow.
2. I've added some comments to explain a bit the logic in `start()` based on my understanding of how it works. Please double check it.
3. Improved unit tests.
4. Used atomic instead of a mutex to protect client/fetcher access

pkg/storage/ingest/fetcher.go
```diff
@@ -586,18 +610,31 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
 		go r.run(ctx, wants, logger, highWatermark)
 	}
 
+	// We need to make sure we don't leak any goroutine given that start is called within a goroutine.
+	defer r.wg.Done()
```
pracucci, line 614: Note to reviewers: moved here just to group it with `r.wg.Add()`.

pkg/storage/ingest/fetcher_test.go
```diff
@@ -357,26 +369,42 @@ func TestConcurrentFetchers(t *testing.T) {
...
 			}
 			assert.Equal(t, 10, consumedRecords)
 		}()
 
-		// Produce more records while processing is slow
-		for i := 5; i < 10; i++ {
-			produceRecord(ctx, t, client, topicName, partitionID, []byte(fmt.Sprintf("record-%d", i)))
-		}
+		// Slowly produce more records while processing is slow too. This increase the chances
+		// of progressive fetches done by the consumer.
```
pracucci, line 387: Note to reviewers: debugging this test I realised it wan't really simulating a slow processing of fetches because it was always fetching all 10 records in a single PollFetches() call.

pkg/storage/ingest/fetcher.go
```diff
@@ -657,10 +679,24 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
...
+			// We have some ordered records ready to be sent to PollFetches(). We store the fetch
+			// result in bufferedResult, and we flip readyBufferedResults to the channel used by
+			// PollFetches().
 			bufferedResult = result
 			readyBufferedResults = r.orderedFetches
 
+			// We increase the count of buffered records only for ordered records that we're sure
+			// will not be discarded later, so that we can get an accurate tracking of the records
+			// ready to be polled by PollFetches() but not polled yet.
+			r.bufferedFetchedRecords.Add(int64(len(result.Records)))
```
dimitarvdimitrov, line 691: it's more accurate if we track this in `.run()`. Technically the single fetcher (`run()`) can be blocked and trying to send its records to the result channel, but the `start()` goroutine is waiting to send to `orderedFetches`. In that case the records buffered within the `run()` are not counted.

I even think that most of the records will be buffered there because `bufferedResult` is only one fetch result.

pkg/storage/ingest/reader.go
```diff
@@ -80,10 +80,15 @@ type PartitionReader struct {
...
-	fetcher     fetcher
+	// client and fetcher are both start after PartitionReader creation. Fetcher could also be
+	// replaced during PartitionReader lifetime. To avoid concurrency issues with functions
+	// getting their pointers (e.g. BufferedRecords()) we use atomic / mutex to protect.
+	client atomic.Pointer[kgo.Client]
+
+	// We use a mutex for fetcher because it's an interface and we have to check against the
+	// nil case too.
+	fetcherMx sync.Mutex
+	fetcher   fetcher
```
dimitarvdimitrov, line 91: nitpick: can't we use a `atomic.Pointer[*fetcher]`? `*fetcher` will never be `nil`, but the `fetcher` it's pointing to might sometimes be

pkg/storage/ingest/fetcher.go
pracucci: > I'm adding some more unit tests and I found an issue. The way we currently increase the number of records is wrong because the same record may be concurrently fetched by multiple routines. I'm working on a fix.

Not true. It was my test that was not doing the assertion correctly. I'm getting back to the Josh implementation.

dimitarvdimitrov (approved):

dimitarvdimitrov (approved):
thanks!

PR #9858, opened 2024-11-10
MQE: Add support for deriv function

lamida:

Fixes #<issue number>

#### Checklist

- [X] Tests updated.
- n.a Documentation added.


pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,62 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+
+var Deriv = FunctionOverRangeVectorDefinition{
+	SeriesMetadataFunction: DropSeriesName,
+	StepFunc:               deriv,
+}
+
+func deriv(step types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
+	head, tail := step.Floats.UnsafePoints(step.RangeEnd)
+
+	if len(head) < 2 && len(tail) == 0 {
```
charleskorn, line 381: There's no guarantee the first two points will be in the head:

```suggestion
	if (len(head) + len(tail)) < 2 {
```

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,62 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+		return 0, false, nil, nil
+	}
+
+	slope, _ := linearRegression(append(head, tail...), head[0].T)
```
charleskorn, line 385: Rather than appending `head` and `tail` like this, what if we pass both to `linearRegression` and modify it to support two slices of points?

pkg/streamingpromql/testdata/upstream/functions.test
charleskorn, line 1: Could you please add tests that include:

* native histograms
* special cases like the range being all 0, the range all containing the same non-zero value, the range containing NaN, Inf, and -Inf

lamida: Sure I will add today

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: That's true! Thanks.

pkg/streamingpromql/testdata/upstream/functions.test
lamida: Added test in 185ea6473c5c106241dac6e4e03d6d468acc6d3e

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -460,3 +460,38 @@ load 6m
 
 eval range from 0 to 96m step 6m round(series, scalar(toNearest))
   {a="b"}  -5   2.7  0.5   20    10    3.25    0         1000.01 0          -1000000  NaN NaN NaN NaN _ _ 10
+
+# Testing deriv
```
charleskorn, line 464: [nit] Similar comment to https://github.com/grafana/mimir/pull/9859#discussion_r1839004339 applies here as well

charleskorn (approved):
LGTM modulo nit

pkg/streamingpromql/testdata/ours/functions.test
lamida: Refactored the test in 6fdb1f69ef738f595ba3d2b141afa0f85f65665c

PR #9859, opened 2024-11-10
MQE: Add support for resets function

lamida:

Fixes #<issue number>

#### Checklist

- [X] Tests updated.
- n.a Documentation added.


pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,56 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
 
 	return avgSoFar, nil
 }
+
+var Resets = FunctionOverRangeVectorDefinition{
+	SeriesMetadataFunction:         DropSeriesName,
+	StepFunc:                       resets,
+	NeedsSeriesNamesForAnnotations: true,
```
charleskorn, line 376: Is this correct? I don't think `resets` emits any annotations, so this shouldn't be needed.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,56 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+			for _, sample := range points {
+				current := sample.F
+				if current < prev {
+					resets++
+				}
+				prev = current
+			}
+
+		}
+		accumulate(fHead)
```
charleskorn, line 404: We should be able to skip the first element (same comment applies below for histogram case too):

```suggestion
		accumulate(fHead[1:])
```

pkg/streamingpromql/testdata/upstream/functions.test
charleskorn, line 1: Could you please add some tests that cover:

* native histograms
* native histograms with custom buckets (or a mixture of exponential and custom buckets)
* a mixture of floats and native histograms in the range

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -369,3 +369,56 @@ func avgHistograms(head, tail []promql.HPoint) (*histogram.FloatHistogram, error
...
+		accumulate(fTail)
+
+		return resets, true, nil, nil
```
charleskorn, line 407: I don't think this is quite right for the case where the range contains both floats and histograms, [the docs for `resets`](https://prometheus.io/docs/prometheus/latest/querying/functions/#resets) says both are considered:

>If the range vector contains a mix of float and histogram samples for the same series, counter resets are detected separately and their numbers added up. The change from a float to a histogram sample is not considered a counter reset. Each float sample is compared to the next float sample, and each histogram is comprared to the next histogram.

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: I added this because some of failed instance test in `TestCompareVariousMixedMetricsVectorSelectors` due to annotation. But I think it failed because something else. Removed in 54d583988a7ce8befe1a631f188d05d1bd22d114.

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Yes I just remembered that resets allow the mix. I will check this again and fix it. Thanks!

lamida: This has been fixed.

pkg/streamingpromql/testdata/upstream/functions.test
lamida: Added in 08264fee95a16490a354739ff23c780185490088

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -460,3 +460,68 @@ load 6m
 
 eval range from 0 to 96m step 6m round(series, scalar(toNearest))
   {a="b"}  -5   2.7  0.5   20    10    3.25    0         1000.01 0          -1000000  NaN NaN NaN NaN _ _ 10
+
+clear
+
+# Testing resets
+load 1m
+  simple_metric_all_same_no_reset{num="0"} 0 0 0 0 0 0 0 0
```
charleskorn, line 468: [nit] It might be clearer to change these series to use labels like these:

```
metric{case="all the same, no resets, value 0"}
metric{case="all the same, no resets, value 3"}
metric{case="all floats, no resets"}
metric{case="all floats, some resets"}
...
```

This would make the cases clearer given we can use full English phrases. Then the `eval`s below can be collapsed into one `eval range from 0 to X step Y resets(metric)` as well.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -375,3 +375,55 @@ func changes(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationF
...
+	SeriesMetadataFunction: DropSeriesName,
+	StepFunc:               resets,
+}
+
+func resets(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
+	fHead, fTail := step.Floats.UnsafePoints()
+	hHead, hTail := step.Histograms.UnsafePoints()
+
+	haveFloats := len(fHead) > 0 || len(fTail) > 0
+	haveHistograms := len(hHead) > 0 || len(hTail) > 0
```
charleskorn, line 389: [nit] The head slice will always be populated if there are any points, so we could drop the `len(xTail)` checks.

charleskorn (approved):
LGTM modulo nits

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Copied from `xOverTime` function 😆 . Such [as](https://github.com/grafana/mimir/blob/03fc64b8ab4b3230e1ac12344b1733a08453f995/pkg/streamingpromql/operators/functions/range_vectors.go#L139). So this means, we can drop tail check there too, right? Maybe we can do it in another PR.

charleskorn: Yep, we can drop it there too, and I'm happy for you to do that in another PR.

pkg/streamingpromql/testdata/ours/functions.test
lamida: Refactored the test in 647feb3c9f2e3a97d8c6ce2ecb523b8248052b58

PR #9874, opened 2024-11-12
MQE: fix `or` where both sides have series with the same labels

charleskorn:
#### What this PR does

This PR fixes a bug in MQE's implementation of `or` where it could return incorrect results if both sides contained series with the same labels, and the series on the left side did not have a value for all time steps.

For example, in the expression `min(series_1) or min(series_2)`, both sides have a single series with no labels. Previously, `or` would have returned two series with the labels `{}`, rather than correctly returning a single `{}` series.

I've opted to use `DeduplicateAndMerge` here, as it's the simplest solution that is already well-tested. A more sophisticated option would be to modify `OrBinaryOperation` to handle this case itself, but this would come at the cost of increased complexity, and it would largely replicate the behaviour and logic of `DeduplicateAndMerge`.

The same issue does not affect `and` or `unless`, as these operations always return only the left series.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
+++ b/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
@@ -242,6 +242,8 @@ func TestOrBinaryOperationSorting(t *testing.T) {
 		},
 		// OrBinaryOperation does not handle the case where both sides contain series with identical labels, and
 		// instead relies on DeduplicateAndMerge to handle merging series with identical labels.
+		// Given NewOrBinaryOperation wraps the OrBinaryOperation in a DeduplicateAndMerge, we can still test this
+		// here.
 		"same series on both sides, one series": {
 			leftSeries: []labels.Labels{
 				labels.FromStrings("series", "1", "group", "1"),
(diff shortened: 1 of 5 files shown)


pkg/streamingpromql/operators/binops/or_binary_operation_test.go
```diff
@@ -240,6 +240,55 @@ func TestOrBinaryOperationSorting(t *testing.T) {
 				labels.FromStrings("right", "6", "group", "3"),
 			},
 		},
+		// OrBinaryOperation does not handle the case where both sides contain series with identical labels, and
+		// instead relies on DeduplicateAndMerge to handle merging series with identical labels.
+		"same series on both sides, one series": {
```
jhesketh, line 245: What do you think about extending this in a new test that passes the expected output through DeduplicateAndMerge and ensures we have the expected final result?

jhesketh (approved):
lgtm, nice find

pkg/streamingpromql/operators/binops/or_binary_operation_test.go
charleskorn: I went one better in 755aa2d: I've changed `NewOrBinaryOperation` to always wrap the operator in `DeduplicateAndMerge`, so then we'll never forget to include it.

PR #9883, opened 2024-11-12
ingester: handle errors when getting topic ID

dimitarvdimitrov:
#### What this PR does

There are two errors returned and we were only handling one of them.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@
 * [BUGFIX] Ingester: Fix race condition in per-tenant TSDB creation. #9708
 * [BUGFIX] Ingester: Fix race condition in exemplar adding. #9765
 * [BUGFIX] Ingester: Fix race condition in native histogram appending. #9765
-* [BUGFIX] Ingester: fix but in concurrent fetching where a failure to list topics on startup would cause to use an invalid topic ID (0x00000000000000000000000000000000). #9883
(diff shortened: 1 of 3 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -288,7 +288,11 @@ func newConcurrentFetchers(
 	if !topics.Has(topic) {
 		return nil, fmt.Errorf("failed to find topic ID: topic not found")
 	}
-	f.topicID = topics[topic].ID
+	topicMetadata := topics[topic]
```
pracucci, line 291: I would feel more comfortable if we just check `topics.Error()`. In case of any error we just return and will be retried. WDYT?

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 * [BUGFIX] Ingester: Fix race condition in per-tenant TSDB creation. #9708
 * [BUGFIX] Ingester: Fix race condition in exemplar adding. #9765
 * [BUGFIX] Ingester: Fix race condition in native histogram appending. #9765
+* [BUGFIX] Ingester: fix but in concurrent fetching where a failure to list topics on startup would cause to use an invalid topic ID (0x00000000000000000000000000000000). #9883
```
narqo, line 70: (nit)

```suggestion
* [BUGFIX] Ingester: fix bug in concurrent fetching where a failure to list topics on startup would cause to use an invalid topic ID (0x00000000000000000000000000000000). #9883
```

narqo (approved):
🔥

PR #9884, opened 2024-11-12
Log an error if the Kafka partitions auto-creation fail with an error in the Kafka response

pracucci:
#### What this PR does

This is a minor thing, but could be useful to get insights if partitions auto-creation fail. Previously we were only logging an error if the partitions creation request fail, but it could also succeed and contain an error in the response. Now we do log the error in the latter case too.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/util.go
+++ b/pkg/storage/ingest/util.go
@@ -171,13 +171,23 @@ func setDefaultNumberOfPartitionsForAutocreatedTopics(cfg KafkaConfig, logger lo
 	defer adm.Close()
 
 	defaultNumberOfPartitions := fmt.Sprintf("%d", cfg.AutoCreateTopicDefaultPartitions)
-	_, err = adm.AlterBrokerConfigsState(context.Background(), []kadm.AlterConfig{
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov (approved):
thanks

pkg/storage/ingest/util.go
```diff
@@ -171,14 +171,24 @@ func setDefaultNumberOfPartitionsForAutocreatedTopics(cfg KafkaConfig, logger lo
...
 			Value: &defaultNumberOfPartitions,
 		},
 	})
 
+	// Check if any error has been returned as part of the response.
+	if err == nil {
+		for _, res := range responses {
+			if res.Err != nil {
+				err = res.Err
+				break
```
seizethedave, line 187: We're here looking at the sub-errors. What do you think about recording _all_ non-nil sub-errors into a multierr or something like that, rather than just the first?

pracucci: We issue an operation on a single resource, there shouldn't be more than 1 error.

PR #9888, opened 2024-11-12
query scheduler: optimize deleting empty tenants from rotation

francoposa:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
+++ b/pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
@@ -2,6 +2,8 @@
 
 package tree
 
+import "slices"
+
 type (
 	TenantID  string
 	QuerierID string
@@ -203,7 +205,7 @@ func (qa *TenantQuerierQueuingAlgorithm) dequeueSelectNode(node *Node) *Node {
 // dequeueUpdateState deletes the dequeued-from node from the following locations if it is empty:
 //   - parent's queueMap,
 //   - tenantNodes
-//   - tenantIDOrder iff there are no other nodes by the same name in tenantNodes. If the child is at the end of
+//   - tenantIDOrder if there are no other nodes by the same name in tenantNodes. If the child is at the end of
 //     tenantIDOrder, it is removed outright; otherwise, it is replaced with an empty ("") element
 //
 // dequeueUpdateState would normally also handle incrementing the queue position after performing a dequeue, but
@@ -222,27 +224,43 @@ func (qa *TenantQuerierQueuingAlgorithm) dequeueUpdateState(node *Node, dequeued
 	// delete from shared tenantNodes
 	for i, tenantNode := range qa.tenantNodes[childName] {
 		if tenantNode == dequeuedFrom {
-			qa.tenantNodes[childName] = append(qa.tenantNodes[childName][:i], qa.tenantNodes[childName][i+1:]...)
+			qa.tenantNodes[childName] = slices.Delete(qa.tenantNodes[childName], i, i+1)
+
 		}
-	}
 
-	// check tenantNodes; we only remove from tenantIDOrder if all nodes with this name are gone.
-	// If the removed child is at the _end_ of tenantIDOrder, we remove it outright; otherwise,
-	// we replace it with an empty string. Removal of elements from anywhere other than the end of the slice
-	// would re-index all tenant IDs, resulting in skipped tenants when starting iteration from the
-	// querier-provided LastTenantIndex.
-	removeFromSharedQueueOrder := len(qa.tenantNodes[childName]) == 0
+		// check tenantNodes; we only remove from tenantIDOrder if all nodes with this name are gone.
+		removeFromSharedQueueOrder := len(qa.tenantNodes[childName]) == 0
 
-	if removeFromSharedQueueOrder {
-		for idx, name := range qa.tenantIDOrder {
-			if name == childName {
-				qa.tenantIDOrder[idx] = string(emptyTenantID)
+		// As tenantIDOrder can be very long, removing tenants from the order can be performance-sensitive
+		// when clearing & deleting many tenant queues due to shutdown operations during rollouts or otherwise.
+		if removeFromSharedQueueOrder {
+
+			// First, just replace the tenantID with the emptyTenantID sentinel value.
+			for idx, name := range qa.tenantIDOrder {
+				if name == childName {
+					qa.tenantIDOrder[idx] = string(emptyTenantID)
+					// tenant ID should only appear once in the list
+					break
+				}
 			}
-		}
-		// clear all sequential empty elements from tenantIDOrder
-		lastElementIndex := len(qa.tenantIDOrder) - 1
-		for i := lastElementIndex; i >= 0 && qa.tenantIDOrder[i] == ""; i-- {
-			qa.tenantIDOrder = qa.tenantIDOrder[:i]
+
+			// Then, only shrink tenantIDOrder when we can remove emptyTenantID sentinel values from the end.
+			// Shrinking from the middle would re-index the tenant order and could cause us to skip tenants
+			// when starting iteration from the LastTenantIndex provided by a querier-worker.
+			//
+			// Because we wait to shrink tenantIDOrder until we can remove the last element from the end,
+			// there may be many consecutive emptyTenantIDs up to (last index - 1) waiting to be removed;
+			// count up the emptyTenantIDs at the end and remove all at once to avoid multiple slice resizes.
+			emptyTenantIDsAtEnd := 0
+			for i := len(qa.tenantIDOrder) - 1; i >= 0 && qa.tenantIDOrder[i] == ""; i-- {
+				emptyTenantIDsAtEnd++
+			}
+
+			qa.tenantIDOrder = slices.Delete(
+				qa.tenantIDOrder,
+				len(qa.tenantIDOrder)-emptyTenantIDsAtEnd,
+				len(qa.tenantIDOrder),
+			)
 		}
 	}
 }


pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
```diff
@@ -222,27 +224,43 @@ func (qa *TenantQuerierQueuingAlgorithm) dequeueUpdateState(node *Node, dequeued
 	// delete from shared tenantNodes
 	for i, tenantNode := range qa.tenantNodes[childName] {
 		if tenantNode == dequeuedFrom {
-			qa.tenantNodes[childName] = append(qa.tenantNodes[childName][:i], qa.tenantNodes[childName][i+1:]...)
+			qa.tenantNodes[childName] = slices.Delete(qa.tenantNodes[childName], i, i+1)
```
francoposa, line 227: not related to the perf issue but using slices.Delete in the other code showed a slight improvement over these append-spread approaches and is 100x easier to understand.

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
```diff
@@ -222,27 +224,43 @@ func (qa *TenantQuerierQueuingAlgorithm) dequeueUpdateState(node *Node, dequeued
...
+			// count up the emptyTenantIDs at the end and remove all at once to avoid multiple slice resizes.
+			emptyTenantIDsAtEnd := 0
+			for i := len(qa.tenantIDOrder) - 1; i >= 0 && qa.tenantIDOrder[i] == ""; i-- {
+				emptyTenantIDsAtEnd++
+			}
+
+			qa.tenantIDOrder = slices.Delete(
+				qa.tenantIDOrder,
+				len(qa.tenantIDOrder)-emptyTenantIDsAtEnd,
+				len(qa.tenantIDOrder),
```
charleskorn, line 262: Is there a reason why you're using `slices.Delete` here rather than reslicing with something like `qa.tenantIDOrder = qa.tenantIDOrder[:len(qa.tenantIDOrder)-emptyTenantIDsAtEnd]`? I'd expect that to be a little faster given we don't need some of the things that `slices.Delete` does like clearing out the deleted elements.

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
charleskorn: To clarify: did you benchmark this change as well? I'd expect these two cases to perform slightly differently given one is removing an element from the middle of the slice and the one below is removing elements from the end of the slice.

francoposa: no this change was not part of the benchmark - this is not a behavior change, just updating to use the new slices built-ins for the same purpose as before

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
francoposa: `slices.Delete` is just so much easier to understand when reading.
Re-benchmarking those two options shows:
* not using slices.Delete gets 61.53% speedup
* using it gets 61.06% speedup

I am not quite convinced that is a point of optimization we need here.

charleskorn: If the concern is readability, what if we introduce a `deleteFromEndOfSlice` method (with a better name than that 🙂)?

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
charleskorn: Could you please expand the benchmark to include this change? I'm not expecting a dramatic difference due to this change, but it'd be good to confirm.

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
francoposa: @charleskorn I have introduced this PR in order to accomplish this - essentially copying over the slices utils from Go stdlib as they were in 1.21 before the clearing of elements was introduced for the deletion-related utils

needs a rubber-stamp, but I got the go-ahead on the license header setup from Bryan Boreham https://github.com/grafana/mimir/pull/9894

francoposa: if adding that package is going to be a big deal I would prefer to just merge this as it is as the hot path is benchmarked.

pkg/scheduler/queue/tree/tenant_querier_queuing_algorithm.go
francoposa: I’ll need to create a separate benghmark suite since this is not a hot path. the slice never exceeds length 4.
Is this necessary?

francoposa:
@charleskorn 

I spent a couple hours expanding the benchmark as well as re-ran the existing one a few times.
Starting back at the existing benchmark, which does not test this patch:

```
-	qa.tenantNodes[childName] = append(qa.tenantNodes[childName][:i], qa.tenantNodes[childName][i+1:]...)
+	qa.tenantNodes[childName] = slices.Delete(qa.tenantNodes[childName], i, i+1)
```

I found that the bare truncation option does not stabilize to be faster than the `slices.Delete` one until we get up to 1000 benchmark iterations each of deleting 1000 tenants.
With benchmark iterations in the hundreds, the slices option often comes up faster.

---
Now with expanding the benchmark to include testing the tenantNode deletion, using `slices.Delete` in both places ends up _faster_ than not using it in either place, at both 100 iterations and 1000 iterations.
Overall % improvement is down very slightly, in part because I had to build the structs in the test loop which takes time, but the results hold that using `slices.Delete` is actually significantly faster when the full path is benchmarked.
```
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/scheduler/queue/tree
cpu: AMD Ryzen 9 PRO 6950H with Radeon Graphics     
BenchmarkFullTenantDeletion
BenchmarkFullTenantDeletion/delete_tenant_func_baseline
BenchmarkFullTenantDeletion/delete_tenant_func_baseline-16         	    1000	 229005856 ns/op
BenchmarkFullTenantDeletion/delete_tenant_func_breakEarlyShrinkOnce
BenchmarkFullTenantDeletion/delete_tenant_func_breakEarlyShrinkOnce-16         	    1000	  93047442 ns/op
BenchmarkFullTenantDeletion/delete_tenant_func_breakEarlyShrinkOnceNoSlices
BenchmarkFullTenantDeletion/delete_tenant_func_breakEarlyShrinkOnceNoSlices-16 	    1000	 101171564 ns/op
PASS
```
* **`slices.Delete` improvement: `1 - (93047442 / 229005856) = 0.5936896827651429` or 59.4% improvement**
* **without `slices.Delete` improvement: `1 - (101171564 / 229005856) = 0.5582140746654094` or 55.8% improvement**


full benchmark code:

```golang
package tree

import (
	"fmt"
	"math/rand"
	"slices"
	"strconv"
	"testing"
)

type fullTenantDeleteFunc func(tenantNodes map[string][]*Node, tenantIDOrder []string, tenant string) (map[string][]*Node, []string)

func baseline(tenantNodes map[string][]*Node, tenantIDOrder []string, tenant string) (map[string][]*Node, []string) {
	// delete from shared tenantNodes
	for i := range tenantNodes[tenant] {
		// only ever going to have a slice of length one here for simplicity
		// and best representation of the actual case we are benchmarking;
		// no need to check if this is the dequeuedFrom node
		//if tenantNode == dequeuedFrom {
		tenantNodes[tenant] = append(tenantNodes[tenant][:i], tenantNodes[tenant][i+1:]...)
		//}
	}

	for idx, name := range tenantIDOrder {
		if name == tenant {
			tenantIDOrder[idx] = string(emptyTenantID)
		}
	}

	// clear all sequential empty elements from tenantIDOrder
	lastElementIndex := len(tenantIDOrder) - 1
	for i := lastElementIndex; i >= 0 && tenantIDOrder[i] == ""; i-- {
		tenantIDOrder = tenantIDOrder[:i]
	}
	return tenantNodes, tenantIDOrder
}

func breakEarly(tenantNodes map[string][]*Node, tenantIDOrder []string, tenant string) (map[string][]*Node, []string) {
	// delete from shared tenantNodes
	for i := range tenantNodes[tenant] {
		// only ever going to have a slice of length one here for simplicity
		// and best representation of the actual case we are benchmarking;
		// no need to check if this is the dequeuedFrom node
		//if tenantNode == dequeuedFrom {
		tenantNodes[tenant] = append(tenantNodes[tenant][:i], tenantNodes[tenant][i+1:]...)
		//}
	}

	for idx, name := range tenantIDOrder {
		if name == tenant {
			tenantIDOrder[idx] = string(emptyTenantID)
			break
		}
	}

	// clear all sequential empty elements from tenantIDOrder
	lastElementIndex := len(tenantIDOrder) - 1
	for i := lastElementIndex; i >= 0 && tenantIDOrder[i] == ""; i-- {
		tenantIDOrder = tenantIDOrder[:i]
	}
	return tenantNodes, tenantIDOrder
}

func breakEarlyShrinkOnce(tenantNodes map[string][]*Node, tenantIDOrder []string, tenant string) (map[string][]*Node, []string) {
	// delete from shared tenantNodes
	for i := range tenantNodes[tenant] {
		// only ever going to have a slice of length one here for simplicity
		// and best representation of the actual case we are benchmarking;
		// no need to check if this is the dequeuedFrom node
		//if tenantNode == dequeuedFrom {
		tenantNodes[tenant] = slices.Delete(tenantNodes[tenant], i, i+1)
		//}
	}

	for idx, name := range tenantIDOrder {
		if name == tenant {
			tenantIDOrder[idx] = string(emptyTenantID)
			break
		}
	}

	emptyTenantIDsAtEnd := 0
	for i := len(tenantIDOrder) - 1; i >= 0 && tenantIDOrder[i] == ""; i-- {
		emptyTenantIDsAtEnd++
	}
	tenantIDOrder = slices.Delete(
		tenantIDOrder,
		len(tenantIDOrder)-emptyTenantIDsAtEnd,
		len(tenantIDOrder),
	)

	return tenantNodes, tenantIDOrder
}

func breakEarlyShrinkOnceNoSlices(tenantNodes map[string][]*Node, tenantIDOrder []string, tenant string) (map[string][]*Node, []string) {
	// delete from shared tenantNodes
	for i := range tenantNodes[tenant] {
		// only ever going to have a slice of length one here for simplicity
		// and best representation of the actual case we are benchmarking;
		// no need to check if this is the dequeuedFrom node
		//if tenantNode == dequeuedFrom {
		tenantNodes[tenant] = append(tenantNodes[tenant][:i], tenantNodes[tenant][i+1:]...)
		//}
	}

	for idx, name := range tenantIDOrder {
		if name == tenant {
			tenantIDOrder[idx] = string(emptyTenantID)
			break
		}
	}

	emptyTenantIDsAtEnd := 0
	for i := len(tenantIDOrder) - 1; i >= 0 && tenantIDOrder[i] == ""; i-- {
		emptyTenantIDsAtEnd++
	}
	tenantIDOrder = tenantIDOrder[:len(tenantIDOrder)-emptyTenantIDsAtEnd]

	return nil, tenantIDOrder
}

func BenchmarkFullTenantDeletion(b *testing.B) {
	tenantCount := 10000

	var testCases = []struct {
		name                string
		deleteFunc          fullTenantDeleteFunc
		tenantNodes         map[string][]*Node
		tenantRotationOrder []string
		tenantDeletionOrder []string
	}{
		{"baseline", baseline, nil, nil, nil},
		{"breakEarlyShrinkOnce", breakEarlyShrinkOnce, nil, nil, nil},
		{"breakEarlyShrinkOnceNoSlices", breakEarlyShrinkOnceNoSlices, nil, nil, nil},
	}

	makeStubs := func() (tenantNodes map[string][]*Node, tenantRotationOrder, tenantDeletionOrder []string) {
		tenantNodes = make(map[string][]*Node, tenantCount)
		tenantRotationOrder = make([]string, tenantCount)

		for i := range tenantRotationOrder {
			tenantRotationOrder[i] = strconv.Itoa(i)
			tenantNodes[strconv.Itoa(i)] = append(tenantNodes[strconv.Itoa(i)], &Node{})
		}

		tenantDeletionOrder = make([]string, tenantCount)
		copy(tenantDeletionOrder, tenantRotationOrder)
		rand.Shuffle(len(tenantDeletionOrder), func(i, j int) {
			tenantDeletionOrder[i], tenantDeletionOrder[j] = tenantDeletionOrder[j], tenantDeletionOrder[i]
		})

		return tenantNodes, tenantRotationOrder, tenantDeletionOrder
	}

	for _, testCase := range testCases {
		b.Run(fmt.Sprintf("delete_tenant_func_%s", testCase.name), func(b *testing.B) {
			for i := 0; i < b.N; i++ {
				tenantNodes, tenantRotationOrder, tenantDeletionOrder := makeStubs()
				testCase.tenantNodes = tenantNodes
				testCase.tenantRotationOrder = tenantRotationOrder
				testCase.tenantDeletionOrder = tenantDeletionOrder

				tenantDeleteIdx := 0
				for len(testCase.tenantRotationOrder) > 0 {
					//fmt.Println("len(testCase.tenantRotationOrder): ", len(testCase.tenantRotationOrder))
					testCase.tenantNodes, testCase.tenantRotationOrder = testCase.deleteFunc(
						testCase.tenantNodes, testCase.tenantRotationOrder, testCase.tenantDeletionOrder[tenantDeleteIdx],
					)
					tenantDeleteIdx++
				}
			}
		})
	}

}
```

charleskorn (approved):
Not going to block on the other changes, thanks for expanding the benchmarks!

PR #9891, opened 2024-11-13
ingester: shortcut fetchWants that have already been satisfied

dimitarvdimitrov:
#### What this PR does

After #9855 and the assertions in `pollFetchesAndAssertNoRecords` there were some flaky tests (`TestConcurrentFetchers/consume_from_end_and_update_immediately` in particular). This happened because some fetch would satisfy the future fetchWants. Future fetchWantes would only end up buffering records and then discarding them at PollFetches.

The tests are more obviously flaky if you make consuming records a bit more expensive, for example, but changing the logger with 

```diff
diff --git a/pkg/storage/ingest/fetcher_test.go b/pkg/storage/ingest/fetcher_test.go
index f96f27d5b4..b1d5df9aa9 100644
--- a/pkg/storage/ingest/fetcher_test.go
+++ b/pkg/storage/ingest/fetcher_test.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 	"testing"
 	"time"
@@ -1057,7 +1058,7 @@ func TestConcurrentFetchers(t *testing.T) {
 }
 
 func createConcurrentFetchers(ctx context.Context, t *testing.T, client *kgo.Client, topic string, partition int32, startOffset int64, concurrency, recordsPerFetch int) *concurrentFetchers {
-	logger := log.NewNopLogger()
+	logger := log.NewLogfmtLogger(os.Stdout)
 	reg := prometheus.NewPedanticRegistry()
 	metrics := newReaderMetrics(partition, reg, func() float64 { return 1 })
```

#### The fix

This PR shortcuts this, so the tests are less likely to be flaky. Now we can still make a request for a fetchWants that's been satisfied, but it's less likely and there will only be up to `concurrency` of these.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9892, opened 2024-11-13
kafka replay speed: add bytes limit for inflight fetch requests

dimitarvdimitrov:
### Problem

OOMs happen because the `concurrency * records-per-fetch` product is too large. We hold too many records in memory and don't consume them as fast as we fetch them. The goal is to have a setting to control the memory consumption of the process. In the future it may be scaled with the ingester's memory request for example.

### Proposal 
Instead of configuring records-per-fetch we should configure the maximum memory for all fetchers. We can work backwards to devise the number of records per `fetchWant` from the number of concurrent fetchers.

I propose to remove these flags
```
-ingest-storage.kafka.ongoing-records-per-fetch=30
-ingest-storage.kafka.startup-records-per-fetch=512
```
and replace them with a single flag

```
-ingest-storage.kafka.max-buffered-bytes=100000000 # 100MB
```

 Each `fetchWant` still needs to have a well defined `startOffset` and `endOffset`. The number of records of each fetchWant will be `($max_inflight_bytes / $startup_concurrency) / $bytes_per_record`.  We already track the average `bytes_per_record` of the last few fetches, but the value there can be volatile. Maybe we can also invest in a better tracking of average bytes.

#### What this PR does

This PR is the first step to doing the above: adding a limit for the inflight bytes.

```
-ingest-storage.kafka.max-buffered-bytes
```

 The limit works by controlling the `fetchWant`s. We don't dispatch a `fetchWant` if its MaxBytes would exceed the limit. The idea is that MaxBytes of the request are usually a good estimation of how much data we have to fetch. 

This does have a caveat that if estimation is bad, we'd continue and still risk OOMing.

I didn't add two separate limits for startup and ongoing because we're currently setting them both to the same value and the short/medium-term plan is to only have a single config option.

#### Next steps

The next step would be to remove the records per fetch config options altogether and only rely on max bytes.

#### Note to reviewers

This is based on https://github.com/grafana/mimir/pull/9891 because that PR fixes flaky tests. The changes aren't otherwise dependant.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3924,6 +3924,12 @@ kafka:
   # CLI flag: -ingest-storage.kafka.use-compressed-bytes-as-fetch-max-bytes
   [use_compressed_bytes_as_fetch_max_bytes: <boolean> | default = true]
 
+  # The maximum number of bytes for buffered responses to fetch requests. This
+  # limit is applied on the sum of all MaxBytes requests that are ongoing. 0 to
```
tacole02, line 3928:
```suggestion
  # limit applies to the sum of all ongoing MaxBytes requests. Set to 0 to
```

tacole02: I think we can use a more active voice here.

tacole02 (approved):
Thanks so much, @dimitarvdimitrov !

pkg/storage/ingest/fetcher.go
```diff
@@ -610,11 +628,48 @@ func casHWM(highWwatermark *atomic.Int64, newHWM int64) {
...
+func (w *inflightFetchWants) count() int {
+	return w.queue.Len()
+}
+
+func (w *inflightFetchWants) add(nextFetch fetchWant) {
+	w.bytes.Add(int64(nextFetch.MaxBytes()))
+	w.queue.PushBack(nextFetch)
+}
+
+func (w *inflightFetchWants) headResultsExhausted() {
```
pracucci, line 660: [nit] The `inflightFetchWants` is essentially a list. I would rename this function in terms or list operations (remove the first item) and how the actual action that triggered it (we've done consuming the first item).

```suggestion
func (w *inflightFetchWants) removeNextResult() {
```

pkg/storage/ingest/fetcher.go
```diff
@@ -610,11 +628,48 @@ func casHWM(highWwatermark *atomic.Int64, newHWM int64) {
...
+		return nil
+	}
+	return w.queue.Front().Value.(fetchWant).result
+}
+
+func (w *inflightFetchWants) count() int {
+	return w.queue.Len()
+}
+
+func (w *inflightFetchWants) add(nextFetch fetchWant) {
```
pracucci, line 655: [nit] Just to clarify it's added at the end.

```suggestion
func (w *inflightFetchWants) append(nextFetch fetchWant) {
```

pkg/storage/ingest/fetcher.go
```diff
@@ -610,11 +628,48 @@ func casHWM(highWwatermark *atomic.Int64, newHWM int64) {
 	}
 }
 
+type inflightFetchWants struct {
+	// queue is the list of all fetchResult of all inflight fetch operations. Pending results
+	// are ordered in the same order these results should be returned to PollFetches(), so the first one
+	// in the list is the next one that should be returned.
+	queue list.List
```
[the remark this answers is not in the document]
dimitarvdimitrov: at this point they're no longe results though 😕 they are the original `fetchWant`s. But i see now it's probably a better idea to call the field what it holds, not what it is. How about `wants list.List`?

pkg/storage/ingest/reader.go
```diff
@@ -1043,6 +1084,12 @@ func newReaderMetrics(partitionID int32, reg prometheus.Registerer, bufferedReco
 		lastConsumedOffset:               lastConsumedOffset,
 		kprom:                            NewKafkaReaderClientMetrics(component, reg),
 	}
+
+	m.Service = services.NewTimerService(100*time.Millisecond, nil, func(context.Context) error {
```
pracucci, line 1088: I can't see anyone starting/stopping this service.

pkg/storage/ingest/fetcher.go
```diff
@@ -654,23 +709,28 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
...
 		// It is guaranteed that this channel is non-nil when bufferedResult is non-empty.
+		//
+		// The idea is that we don't want to block the main loop when we have records ready to be consumed.
 		readyBufferedResults chan fetchResult
 	)
 	nextFetch.bytesPerRecord = 10_000 // start with an estimation, we will update it as we consume
 
 	for {
-		refillBufferedResult := nextResult
+		refillBufferedResult := inflight.peekNextResult()
```
pracucci, line 719: [nit] Can I suggest adding another comment to explain this?

```suggestion
		// refillBufferedResult is the channel of the next fetch result. This variable is valued (non-nil) only when
		// we're ready to actually read the result, so that we don't try to read the next result if we're not ready.
		refillBufferedResult := inflight.peekNextResult()
```

pkg/storage/ingest/fetcher_test.go
```diff
@@ -1054,12 +1061,176 @@ func TestConcurrentFetchers(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		_, clusterAddr := testkafka.CreateCluster(t, partitionID+1, topicName)
+		client := newKafkaProduceClient(t, clusterAddr)
+
+		// Create fetchers early to ensure we don't miss any records
+		fetchers := createConcurrentFetchers(ctx, t, client, topicName, partitionID, 0, concurrency, recordsPerFetch, maxInflightBytes)
+
+		waitForStableBufferedRecords := func() {
```
pracucci, line 1159: [nit] Since this function is used in 2 test cases, you could move it at the beginning of `TestConcurrentFetchers`.

pkg/storage/ingest/fetcher_test.go
```diff
@@ -1054,12 +1061,176 @@ func TestConcurrentFetchers(t *testing.T) {
...
+		t.Log("Buffered records stabilized")
+
+		assert.LessOrEqualf(t, fetchers.BufferedBytes(), int64(maxInflightBytes), "Should not buffer more than %d bytes of large records", maxInflightBytes)
+		// Consume all large records
+		consumedRecords := 0
+		for consumedRecords < largeRecordsCount {
+			fetches, _ := fetchers.PollFetches(ctx)
+			consumedRecords += fetches.NumRecords()
+		}
+
```
pracucci, line 1187: To ensure we really consumed all records, I suggest to add this check here:

```
pollFetchesAndAssertNoRecords(t, fetchers)
```

pracucci (approved):
Nice work and nice tests! Pre-approving, but please remember to start/stop the reader metrics services.

pkg/storage/ingest/reader.go
dimitarvdimitrov: it's started here 
https://github.com/grafana/mimir/blob/314647e93954141427fb861dee9e9dda6007a0a3/pkg/storage/ingest/reader.go#L219-L228

and then stopped here
https://github.com/grafana/mimir/blob/314647e93954141427fb861dee9e9dda6007a0a3/pkg/storage/ingest/reader.go#L280-L285

pkg/storage/ingest/fetcher.go
```diff
@@ -682,22 +742,12 @@ func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concu
...
-				// In case we previously exhausted pendingResults, we just created
-				nextResult = pendingResults.Front().Value.(chan fetchResult)
-				pendingResults.Remove(pendingResults.Front())
-			}
+			inflight.append(nextFetch)
 			nextFetch = nextFetch.Next(recordsPerFetch)
 
 		case result, moreLeft := <-refillBufferedResult:
 			if !moreLeft {
-				if pendingResults.Len() > 0 {
```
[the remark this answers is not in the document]
dimitarvdimitrov: that's right. Now instead of keeping `nextResult` as state we compute it on every iteration. The invariants from before still hold

PR #9894, opened 2024-11-13
introduce go 1.21 slices package behavior in utils

francoposa:
#### What this PR does

The information in the header explains:

> 
> Package slicesleak is intended to provide access to the Go stdlib slices package
> as it existed in Go 1.21, prior to the acceptance of Proposal 63393.
> 
> Proposal 63393, implemented in commit e21dc702d54e85381a97259db7deec710108279b
> changed the behavior of the slices package to avoid what can be memory leaks in some cases,
> by zeroing the slice elements discarded by Delete, DeleteFunc, Compact, CompactFunc, Replace.
> Clearing deleted slice elements has a performance overhead and may be undesirable
> in cases where the slice is expected to expand to reuse the memory again.
> 
> References:
> * Proposal: https://github.com/golang/go/issues/63393
> * Commit: https://github.com/golang/go/commit/e21dc702d54e85381a97259db7deec710108279b
> * Official Go Blog on changes: https://go.dev/blog/generic-slice-functions

We wanted to use some cleaner interfaces for the ugly standard slices deletes in this PR: https://github.com/grafana/mimir/pull/9888
but the post-go1.21 version of the slices implementation had a small but measurable impact on performance.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


jhesketh (commented):
I'm not sure how I feel about copying this without tests. I understand why, but if we want to maintain an alternative set of slice functions then I think we need some protections and need to consider writing our own tests for them.

Do we have benchmarks on the performance differences? Are there other alternatives that doesn't increase our maintainership burden.

PR #9897, opened 2024-11-13
Block builder scheduler startup: learn state from worker updates

seizethedave:
#### What this PR does

More progress on the block-builder-scheduler:

* Add a startup period where we're learning the state of the world. This includes:
   * committed offsets learned from Kafka
   * any job progress/completion updates we've learned from workers given `block-builder-scheduler.startup-observe-time`
* When the startup period if over, we compute the starting state from the observations and begin normal operation.
* Add a job assignment `epoch` mechanism to break ties when multiple workers report updates about the same job. Clients include the epoch in any job-related communications.

What's next?
* Flush updated offsets periodically back to Kafka.
* RPC + client layers.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -65,6 +77,37 @@ func (s *BlockBuilderScheduler) stopping(_ error) error {
 }
 
 func (s *BlockBuilderScheduler) running(ctx context.Context) error {
+	// The startup process for block-builder-scheduler entails learning the state of the world:
+	//  1. obtain an initial set of offset info from Kafka
+	//  2. listen to worker updates for a while to learn what the previous scheduler knew
+	// When both of those are complete, we transition from observation mode to normal operation.
```
narqo, line 83: (nit) What if we move these pieces of work into the `starting` handler? This may better separate the logic before and after the scheduler becomes ready to do the work.

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -65,6 +77,37 @@ func (s *BlockBuilderScheduler) stopping(_ error) error {
...
+		defer wg.Done()
+		lag, err := s.fetchLag(ctx)
+		if err != nil {
+			panic(err)
+		}
+		s.committed = commitOffsetsFromLag(lag)
+	}()
+	go func() {
+		defer wg.Done()
+		time.Sleep(s.cfg.StartupObserveTime)
```
narqo, line 98: I'm not sure how it will look in the later PRs, but maybe add a TODO here to bail from sleeping if the context is cancelled?

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -78,14 +121,31 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
...
+	if s.observationComplete {
+		return
+	}
+
+	s.jobs = newJobQueue(s.cfg.JobLeaseExpiry, s.logger)
+
+	if err := s.finalizeObservations(); err != nil {
+		level.Warn(s.logger).Log("msg", "failed to compute state from observations", "err", err)
+		// (what to do here?)
+	}
```
narqo, line 137: Should this error be a terminal state for whole scheduler app, or maybe it can retry here?

narqo (approved):
🔥 Looks good to me overall. Since it's still in the active dev phase, the suggestions I left shouldn't block.

pkg/blockbuilder/scheduler/scheduler.go
seizethedave: I suspect retrying won't change anything as it's some deterministic non-RPC code. It would really be a logic error that we'd want to fix post-haste. I think the scheduler should resume starting up and just log/increment metrics if this happens. Thanks for forcing me to think about it. :)

pkg/blockbuilder/scheduler/scheduler.go
seizethedave: I like the idea, hadn't considered it. The service needs to be "up" during this time to receive requests, but reading the BasicService code, that wouldn't interfere there. Done.

PR #9903, opened 2024-11-14
Reduce 'skipped records because already returned' log frequency

pracucci:
#### What this PR does

I was looking at the flaky test reported in https://github.com/grafana/mimir/pull/9891. I actually reproduce it enabling the logging (and also increasing the number of records produced/consumed by the test). I personally think the main problem is that we log 'skipped records because already returned' for every single record, which looks a bit too much to me in case there are many skipped records.

In this PR I propose to just log the summary of all skipped records, logging the count, and the min/max offsets. With this change I can't get the test being flaky anymore. I also measured `'skipped records because already returned'()` timing and now it's in the order of microserconds when running the tests locally (with log enabled).

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -383,15 +383,35 @@ func (r *concurrentFetchers) recordOrderedFetchTelemetry(f fetchResult, firstRet
 	level.Debug(r.logger).Log("msg", "received ordered fetch", "num_records", len(f.Records), "wait_duration", waitDuration)
 	r.metrics.fetchWaitDuration.Observe(waitDuration.Seconds())
 
-	doubleFetchedBytes := 0
+	var (
+		doubleFetchedBytes             = 0
+		skippedRecordsCount            = 0
+		firstSkippedRecordOffset int64 = -1
+		lastSkippedRecordOffset  int64 = -1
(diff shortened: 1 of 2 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -383,15 +383,35 @@ func (r *concurrentFetchers) recordOrderedFetchTelemetry(f fetchResult, firstRet
...
+				firstSkippedRecordOffset = record.Offset
+			}
 		}
 		r.tracer.OnFetchRecordUnbuffered(record, true)
 	}
 	r.metrics.fetchedDiscardedRecordBytes.Add(float64(doubleFetchedBytes))
+
+	if skippedRecordsCount > 0 {
+		spanlogger.FromContext(f.Records[0].Context, r.logger).DebugLog(
+			"msg", "skipped records because already returned",
```
dimitarvdimitrov, line 410:
```suggestion
			"msg", "skipped records because it is already returned",
```

dimitarvdimitrov (commented):
LGTM, thanks for fixing this

dimitarvdimitrov (approved):

PR #9906, opened 2024-11-14
kafka replay speed: remove records-per-fetch configuration

dimitarvdimitrov:
#### What this PR does

This is the next step after https://github.com/grafana/mimir/pull/9892. We want to control only the number of bytes for fetching because that's more indicative of the load we'd receive. It also allows us to fetch in smaller increments when records become big.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/storage/ingest/fetcher_test.go
```diff
@@ -834,76 +831,6 @@ func TestConcurrentFetchers(t *testing.T) {
 		pollFetchesAndAssertNoRecords(t, fetchers)
 	})
 
-	t.Run("staggered production with one less than multiple of concurrency and records per fetch", func(t *testing.T) {
```
dimitarvdimitrov, line 837: now that we don't do exact number of records this test becomes to a large extent non-deterministic. Its purpose was to test an off-by-one condition to the previous test, but achieving this with tests is really hard now because of byte estimations

pkg/storage/ingest/fetcher.go
```diff
@@ -694,8 +698,11 @@ func (w *inflightFetchWants) removeNextResult() {
 	w.wants.Remove(head)
 }
 
-func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concurrency, recordsPerFetch int) {
-	level.Info(r.logger).Log("msg", "starting concurrent fetchers", "start_offset", startOffset, "concurrency", concurrency, "recordsPerFetch", recordsPerFetch)
+func (r *concurrentFetchers) start(ctx context.Context, startOffset int64, concurrency int) {
+	const defaultBytesPerRecord = 10_000 // start with an estimation, we will update it as we consume
```
pracucci, line 702: [nit] Don't hate me. I think "initial" is more accurate than "default". This is not a fallback, it's just the initial value, before we can estimate.

```suggestion
	const initialBytesPerRecord = 10_000 // start with an estimation, we will update it as we consume
```

pkg/storage/ingest/fetcher.go
```diff
@@ -66,22 +66,27 @@ type fetchWant struct {
 	startOffset    int64 // inclusive
 	endOffset      int64 // exclusive
 	bytesPerRecord int
```
pracucci, line 68: [nit] You know I'm obsessed about calling the same data with the same exact name across the code based :D 
```suggestion
	estimatedBytesPerRecord int
```

pracucci (approved):
Nice work, LGTM!

tacole02 (approved):
Docs look good. Thanks, @dimitarvdimitrov !

pracucci:
I just noticed that a test hung in CI. Could be worth investigating if we introduced a regression (if so I can't see it just by looking at the code).

dimitarvdimitrov:
> I just noticed that a test hung in CI. Could be worth investigating if we introduced a regression (if so I can't see it just by looking at the code).

i'm totally beat today. i'll check tomorrow

dimitarvdimitrov:
> I just noticed that a test hung in CI. Could be worth investigating if we introduced a regression (if so I can't see it just by looking at the code).

oh boy. This test hangs when it fails for some reason. Locally, i see the assertion failing, but the test never terminates. In CI we only observe it never terminating.

The test failure was more complicated than I anticipated, but at least now I understand it. I've added comments to explain the asserted values now.

PR #9913, opened 2024-11-15
Run pruning middleware before query sharding middleware

zenador:
#### What this PR does

Moves pruning middleware to earlier in the flow, before query sharding middleware instead of after, as it makes more sense to do it earlier.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/roundtrip.go
+++ b/pkg/frontend/querymiddleware/roundtrip.go
@@ -395,20 +395,6 @@ func newQueryMiddlewares(
 		queryRangeMiddleware = append(queryRangeMiddleware, cfg.ExtraRangeQueryMiddlewares...)
 	}
 
-	if cfg.PrunedQueries {
-		pruneMiddleware := newPruneMiddleware(log)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
   * `cortex_alertmanager_silences`
 * [CHANGE] Distributor: Drop experimental `-distributor.direct-otlp-translation-enabled` flag, since direct OTLP translation is well tested at this point. #9647
 * [CHANGE] Ingester: Change `-initial-delay` for circuit breakers to begin when the first request is received, rather than at breaker activation. #9842
+* [CHANGE] Query-frontend: run the pruning middleware before the query sharding middleware instead of after. #9913
```
charleskorn, line 18: [nit] Probably makes sense to describe this in terms the average user would understand:
```suggestion
* [CHANGE] Query-frontend: apply query pruning before query sharding instead of after. #9913
```

charleskorn (approved):
LGTM modulo nit

PR #9919, opened 2024-11-15
kafka replay speed: don't trim fetchWants

dimitarvdimitrov:
#### What this PR does

I realized that trimming `fetchWant`s can end up discarding offsets in extreme circumstances.

### How it works

If the fetchWant is so big that its size would exceed 2GiB, then we trim it. We trim it by reducing the end offset. The idea is that the next fetchWant will pick up from where this one left off.

### How it can break

We trim the `fetchWant` in `UpdateBytesPerRecord` too. `UpdateBytesPerRecord` can be invoked in `concurrentFEtchers.run` after the `fetchWant` is dispatched. In that case the next `fetchWant` would have already been calculated. And we would end up with a gap.

### Did it break?

It's hard to tell, but it's very unlikely. To reach 2GiB we would have needed to have the estimation for bytes per record be 2 MiB (with 1000 records per fetch). While these large records are possible, they should be rare and our rolling average estimation for records size shouldn't reach it.

### Why we can do without `trimIfTooBig()`?

`MaxBytes()` already caps the value to MaxInt32, so we're compliant with the Kafka Protocol. 

https://github.com/grafana/mimir/blob/175bef494b0e1c13c6609a7e76d0e1c04c90cf6c/pkg/storage/ingest/fetcher.go#L97-L102

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -95,7 +95,7 @@ func (w fetchWant) Next() fetchWant {
 // It's capped at math.MaxInt32 to avoid overflow, and it'll always fetch a minimum of 1MB.
 func (w fetchWant) MaxBytes() int32 {
 	fetchBytes := w.expectedBytes()
-	if fetchBytes > math.MaxInt32 {
+	if fetchBytes > math.MaxInt32 || fetchBytes < 0 {
 		// This shouldn't happen because w should have been trimmed before sending the request.
 		// But we definitely don't want to request negative bytes by casting to int32, so add this safeguard.
 		return math.MaxInt32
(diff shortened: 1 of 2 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -126,18 +126,6 @@ func (w fetchWant) expectedBytes() int {
 	return int(overFetchBytesFactor * float64(w.estimatedBytesPerRecord*int(w.endOffset-w.startOffset)))
```
pracucci, line 126: [nit] I would make `expectedBytes()` return a `int64`, just in case we have some OSS users running on 32bit systems (don't think so, but...).

pracucci (approved):
I see the problem. Good job spotting it!

Do you think we can unit test `fetchWatch.MaxBytes()` and `UpdateBytesPerRecord()` for this specific case the fetch would be greater than int32 bytes? I know we removed the trim code, but could help us not reintroduce something similar in the future.

pkg/storage/ingest/fetcher.go
dimitarvdimitrov: 👍 https://github.com/grafana/mimir/pull/9919/commits/70fa210765ee7edfc97875266eb546240569b431

dimitarvdimitrov:
> Do you think we can unit test `fetchWatch.MaxBytes()` and `UpdateBytesPerRecord()` for this specific case the fetch would be greater than int32 bytes? I know we removed the trim code, but could help us not reintroduce something similar in the future.

added some tests here and had to adjust MaxBytes slightly because it could overflow sometimes https://github.com/grafana/mimir/pull/9919/commits/dcc1397eb1fcef76801419c8253301e6620715af

PR #9936, opened 2024-11-18
Ruler: do not return an error if a rule group is missing after founding it when listing

pracucci:
#### What this PR does

Over the weekend we've paged because of some ruler failures with log message:
```
an error occurred while loading 1 rule groups
```

After some investigation we've found that the issue is caused by concurrent `ListRules()` and other CRUD operations on rule groups config. If a rule group is deleted after being listed but before getting loaded, the `ListRules()` API will fail because of them missing rule group.

This behaviour was introduced to guarantee the API is strongly consistent. This is a property we still want to preserve. However, in Mimir we also assume that the object storage is strongly consistent (the API can't be strongly consistent anyway if the object storage isn't), so I don't think we have to error out if the object storage returns a 404 when loading a rule group (but we should still error out for any other type of error).

In this PR I propose to change the `/prometheus/config/v1/rules` behaviour, to not return an error anymore if a rule group is missing in the object storage after been successfully returned by listing the storage, because it could have been deleted in the meanwhile. It's important to note that the `missing` list only contains rule groups that failed to load because they don't exist in the object storage, and not because of other reasons.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -500,9 +500,29 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	if len(missing) > 0 {
-		// This API is expected to be strongly consistent, so it's an error if any rule group was missing.
-		http.Error(w, fmt.Sprintf("an error occurred while loading %d rule groups", len(missing)), http.StatusInternalServerError)
-		return
+		// This API is expected to be strongly consistent, but we expect the object storage to be strongly
+		// consistent too. This means that if a rule group existed when we listed the storage but doesn't exist
+		// (it's missing) when we load it, then it could have been deleted in the meanwhile.
+		//
+		// We don't want to consider this condition a failure, so we don't return an error in this case, but we
+		// just log a warning because it could be of interest for an operator to further investigate it.
+		level.Warn(a.logger).Log(
+			"msg", "list rules API skipped some rule groups, because missing when loading them after listing the storage (this could be due to rule groups deleted between listing the storage and getting rule groups content)",
+			"listed_rule_groups", len(rgs),
+			"missing_rule_groups", len(missing))
+
+		// Filter out missing rule groups, so they're not returned by the API (they haven't been loaded,
+		// so their content is empty).
+		tenantRuleGroups := map[string]rulespb.RuleGroupList{userID: rgs}
+		tenantRuleGroups = filterRuleGroupsByNotMissing(tenantRuleGroups, missing, a.logger)
+
+		var tenantFound bool
+		rgs, tenantFound = tenantRuleGroups[userID]
+		if !tenantFound {
+			// This should never happen, unless a bug.
+			http.Error(w, "an error occurred when filtering missing rule groups", http.StatusInternalServerError)
+			return
+		}
 	}
 
 	numRules := 0
--- a/pkg/ruler/api_test.go
+++ b/pkg/ruler/api_test.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/gorilla/mux"
+	"github.com/grafana/dskit/concurrency"
 	"github.com/grafana/dskit/services"
 	"github.com/grafana/dskit/test"
 	"github.com/grafana/dskit/user"
@@ -36,17 +37,20 @@ import (
 
 func TestRuler_ListRules(t *testing.T) {
 	const (
-		userID   = "user1"
-		interval = time.Minute
+		userID                      = "user1"
+		interval                    = time.Minute
+		skippedMissingRuleGroupsMsg = "list rules API skipped some rule groups, because missing when loading them after listing the storage"
 	)
 
 	testCases := map[string]struct {
-		requestPath        string
-		configuredRules    rulespb.RuleGroupList
-		missingRules       rulespb.RuleGroupList
-		expectedStatusCode int
-		expectedRules      map[string][]rulefmt.RuleGroup
-		expectedErr        string
+		requestPath          string
+		configuredRules      rulespb.RuleGroupList
+		missingRules         rulespb.RuleGroupList
+		expectedStatusCode   int
+		expectedRules        map[string][]rulefmt.RuleGroup
+		expectedErr          string
+		expectLogsContain    []string
+		expectLogsNotContain []string
 	}{
 		"should list all rule groups of an user if the namespace parameter is missing": {
 			requestPath: "/prometheus/config/v1/rules",
@@ -87,6 +91,7 @@ func TestRuler_ListRules(t *testing.T) {
 					}),
 				},
 			},
+			expectLogsNotContain: []string{skippedMissingRuleGroupsMsg},
 		},
 		"should list all rule groups of an user belonging to the input namespace": {
 			requestPath: "/prometheus/config/v1/rules/namespace1",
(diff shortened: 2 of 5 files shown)


pkg/ruler/api.go
```diff
@@ -500,9 +500,29 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
...
+		// This API is expected to be strongly consistent, but we expect the object storage to be strongly
+		// consistent too. This means that if a rule group existed when we listed the storage but doesn't exist
+		// (it's missing) when we load it, then it could have been deleted in the meanwhile.
+		//
+		// We don't want to consider this condition a failure, so we don't return an error in this case, but we
+		// just log a warning because it could be of interest for an operator to further investigate it.
+		level.Warn(a.logger).Log(
+			"msg", "list rules API skipped some rule groups, because missing when loading them after listing the storage (this could be due to rule groups deleted between listing the storage and getting rule groups content)",
+			"listed_rule_groups", len(rgs),
+			"missing_rule_groups", len(missing))
```
dimitarvdimitrov, line 512: do you want to also log which were the missing ones?

pkg/ruler/api.go
```diff
@@ -500,9 +500,29 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
...
+
+		// Filter out missing rule groups, so they're not returned by the API (they haven't been loaded,
+		// so their content is empty).
+		tenantRuleGroups := map[string]rulespb.RuleGroupList{userID: rgs}
+		tenantRuleGroups = filterRuleGroupsByNotMissing(tenantRuleGroups, missing, a.logger)
+
+		var tenantFound bool
+		rgs, tenantFound = tenantRuleGroups[userID]
+		if !tenantFound {
+			// This should never happen, unless a bug.
```
dimitarvdimitrov, line 522: what if all the rule groups were missing (e.g. just 1 rule group which was deleted between the listing and the GET?)

dimitarvdimitrov (approved):
SGTM

pkg/ruler/api.go
pracucci: Good job spotting it! It was a bug. Fixed here and added a unit test:
https://github.com/grafana/mimir/pull/9936/commits/7dd5c5a2f6c3a923a6f83495ed56ac4c4c2cf5ab

pkg/ruler/api.go
pracucci: Good point. I think logging all of them may be excessive: what if we log the first one to give an hint?
https://github.com/grafana/mimir/pull/9936/commits/7dd5c5a2f6c3a923a6f83495ed56ac4c4c2cf5ab

pkg/ruler/api.go
```diff
@@ -500,9 +500,31 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
...
+		// (it's missing) when we load it, then it could have been deleted in the meanwhile.
+		//
+		// We don't want to consider this condition a failure, so we don't return an error in this case, but we
+		// just log a warning because it could be of interest for an operator to further investigate it.
+		level.Warn(a.logger).Log(
+			"msg", "list rules API skipped some rule groups, because missing when loading them after listing the storage (this could be due to rule groups deleted between listing the storage and getting rule groups content)",
+			"listed_rule_groups", len(rgs),
+			"missing_rule_groups", len(missing),
+			// Logging all rule groups may excessive, but logging at least 1 may give some hints.
+			"first_missing_rule_group", fmt.Sprintf("user=%s namespace=%s group=%s", missing[0].User, missing[0].Namespace, missing[0].Name))
```
dimitarvdimitrov, line 514: i'm not sure about nesting logfmt kv pairs. It's harder to parse in loki (`first_missing_rule_group="user=123 namespace=x group=abc"`)

can we just have separate fields for each?
```suggestion
			"user", missing[0].User,
			"first_missing_rule_group_namespace", missing[0].Namespace,
			"first_missing_rule_group_name", missing[0].Name)
```

pkg/ruler/api.go
```diff
@@ -500,9 +500,31 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
...
-		// This API is expected to be strongly consistent, so it's an error if any rule group was missing.
-		http.Error(w, fmt.Sprintf("an error occurred while loading %d rule groups", len(missing)), http.StatusInternalServerError)
-		return
+		// This API is expected to be strongly consistent, but we expect the object storage to be strongly
+		// consistent too. This means that if a rule group existed when we listed the storage but doesn't exist
+		// (it's missing) when we load it, then it could have been deleted in the meanwhile.
+		//
+		// We don't want to consider this condition a failure, so we don't return an error in this case, but we
+		// just log a warning because it could be of interest for an operator to further investigate it.
+		level.Warn(a.logger).Log(
```
56quarters, line 509: This should be using `logger` which is a `SpanLogger`.

pkg/ruler/api.go
```diff
@@ -500,9 +500,31 @@ func (a *API) ListRules(w http.ResponseWriter, req *http.Request) {
...
+		// Filter out missing rule groups, so they're not returned by the API (they haven't been loaded,
+		// so their content is empty).
+		tenantRuleGroups := map[string]rulespb.RuleGroupList{userID: rgs}
+		tenantRuleGroups = filterRuleGroupsByNotMissing(tenantRuleGroups, missing, a.logger)
+
+		var tenantFound bool
+		rgs, tenantFound = tenantRuleGroups[userID]
+		if !tenantFound && len(missing) < len(rgs) {
+			// This should never happen, unless a bug.
+			http.Error(w, "an error occurred when filtering missing rule groups", http.StatusInternalServerError)
```
56quarters, line 525: Is it expected that we'll be able to debug this problem with the above logging statement? Can we add some details to the `SpanLogger`, `logger`?

pkg/ruler/api_test.go
```diff
@@ -184,6 +363,14 @@ func TestRuler_ListRules(t *testing.T) {
 			} else {
 				require.Contains(t, string(body), tc.expectedErr)
 			}
+
+			// Check logs.
+			for _, expect := range tc.expectLogsContain {
+				require.Contains(t, logs.String(), expect)
+			}
+			for _, expect := range tc.expectLogsNotContain {
+				require.NotContains(t, logs.String(), expect)
```
56quarters, line 372: Can we come up with another way to test this? I really don't like introducing dependencies on particular log messages.

56quarters (approved):
LGTM with a few suggestions.

pkg/ruler/api_test.go
pracucci: I need suggestions :)

pkg/ruler/api.go
pracucci: Very good point. Done in https://github.com/grafana/mimir/pull/9936/commits/776ffccf38dcc9c4a07a8e7bef38efc053da50df

pkg/ruler/api.go
pracucci: Done in https://github.com/grafana/mimir/pull/9936/commits/776ffccf38dcc9c4a07a8e7bef38efc053da50df

pkg/ruler/api.go
pracucci: I added an extra error log to signal this is a non expected condition, but at the end the information I've logged is the same as the previous log: https://github.com/grafana/mimir/pull/9936/commits/363be718f12b924690fab6492713f001dee86b6d

pkg/ruler/api_test.go
pracucci: I'm testing that we log it. I want to make sure we have a test :) I'm going to keep it. I see this kind of logs testing similar to metrics testing: just testing observability signals. Don't see it as a bad thing.

PR #9963, opened 2024-11-20
Fix data loss bug in the experimental ingest storage when a Kafka Fetch is split into multiple requests and some of them return an error

pracucci:
#### What this PR does

This PR fixed a data loss bug in the experimental ingest storage when a Kafka Fetch is split into multiple requests and some of them return an error.

**The bug**
Due to the `fetchResult.Merge()` we call in `concurrentFetchers.run()` we can merge records in a `kgo.Fetch` that also contains a `Err` (so it both has `Err` and some records). When this fetch is returned by `PollFetches()` to `PartitionReader`,  `PartitionReader` will filter out everything because of the `Err` set. We don't expect to both have an `Err` and some records set in the same `kgo.Fetch`, so I think `PartitionReader` is doing the right thing, and `concurrentFetchers` shouldn't mix `Err` with records.

**The fix**
A part from some small improvements / comments / new unit tests here and there, in this PR I address the issue doing two things:

1. `concurrentFetchers.parseFetchResponse()` checks for the error both at the `kmsg.FetchResponse` root and in the partition response
2. `concurrentFetchers.run()` doesn't merge `fetchResult` containing `Err`. To make the code a bit more robust and avoid mistakes in the future, I've explicitly introduced an `if` branch on the result from `r.fetchSingle()` with two different behaviours in case of err or no-err. I don't expect the logic to have been changed here, a part from not merging errors with records.

I intentionally introduced a couple of panics for invariants that, if violated, cause serious issues.

**The test**
I've locally run the new `TestPartitionReader_ShouldNotMissRecordsIfFetchRequestContainPartialFailuresWithConcurrentFetcherIsUsed` 100 times with race and it looks both stable and fast (thanks to the custom backoff config the test time dropped from 50s to < 2s):

```
go test -v -race -failfast -count=100 -run 'TestPartitionReader_ShouldNotMissRecordsIfFetchRequestContainPartialFailuresWithConcurrentFetcherIsUsed' ./pkg/storage/ingest
```

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -463,8 +463,6 @@ func (r *concurrentFetchers) recordOrderedFetchTelemetry(f fetchResult, firstRet
 // contain any record in case an error is also returned.
 //
 // If ctx is cancelled, fetchSingle will return an empty fetchResult without an error.
-// TODO unit test: If ctx is cancelled, fetchSingle will return an empty fetchResult without an error
-// TODO unit test: doesn't contain any record in case an error is also returned
 func (r *concurrentFetchers) fetchSingle(ctx context.Context, fw fetchWant) (fr fetchResult) {
 	defer func(fetchStartTime time.Time) {
 		fr.logCompletedFetch(fetchStartTime, fw)
@@ -489,7 +487,6 @@ func (r *concurrentFetchers) fetchSingle(ctx context.Context, fw fetchWant) (fr 
 		return newErrorFetchResult(ctx, r.partitionID, fmt.Errorf("fetching from kafka: %w", err))
 	}
 
-	// TODO guarantee no error and records at the same time.
 	return r.parseFetchResponse(ctx, fw.startOffset, resp)
 }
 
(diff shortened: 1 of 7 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -555,37 +622,42 @@ func sumRecordLengths(records []*kgo.Record) (sum int) {
...
+				// Keep non-mergeable fields from res, because the last response is the most updated one.
+				bufferedResult = res.Merge(bufferedResult)
 			}
-			if hwm := f.HighWatermark; hwm >= 0 {
-				casHWM(highWatermark, hwm)
-			}
-			if len(f.Records) == 0 {
+
+			if len(bufferedResult.Records) == 0 {
 				// Typically if we had an error, then there wouldn't be any records.
```
dimitarvdimitrov, line 661: this comment is now outdated.

i couldn't find a claim that an error means no records. Are we relying on the alert for offset gaps to validate this assumption?

dimitarvdimitrov (approved):
LGTM, only found one outdated comment

pkg/storage/ingest/fetcher.go
pracucci: > this comment is now outdated.

Commented updated here: https://github.com/grafana/mimir/pull/9963/commits/883e4ea9aeae2604303991da9a9465491b809da6

dimitarvdimitrov (approved):

PR #9921, opened 2024-11-15
kafka replay speed: add alert for when we miss records in Kafka

dimitarvdimitrov:
#### What this PR does

Adds an alert and metrics to detect when we have bugs.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #9972, opened 2024-11-21
ingest: move missed record tracking to PartitionReader

dimitarvdimitrov:
#### What this PR does

follow-up to #9921 

The idea is to catch gaps as close as possible to where we consume fetches. With the previous alert we wouldn't have caught the bug fixed in #9963

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -449,7 +449,6 @@ func instrumentGaps(gaps []offsetRange, records prometheus.Counter, logger log.L
 			"records_offset_gap_end_exclusive", gap.end,
 		)
 		records.Add(float64(gap.numOffsets()))
-		level.Error(logger).Log("msg", "found gap in records", "start", gap.start, "end", gap.end)
 	}
 }
 
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -468,7 +468,7 @@ func (g offsetRange) numOffsets() int64 {
 func findGapsInRecords(records kgo.Fetches, lastReturnedOffset int64) []offsetRange {
 	var gaps []offsetRange
 	records.EachRecord(func(r *kgo.Record) {
-		if r.Offset != lastReturnedOffset+1 {
+		if r.Offset > lastReturnedOffset+1 {
 			gaps = append(gaps, offsetRange{start: lastReturnedOffset + 1, end: r.Offset})
 		}
 		lastReturnedOffset = r.Offset
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -466,6 +466,10 @@ func (g offsetRange) numOffsets() int64 {
 
 func findGapsInRecords(records kgo.Fetches, lastSeenOffset int64) []offsetRange {
 	var gaps []offsetRange
+	if lastSeenOffset < 0 {
+		firstRecord := records.RecordIter().Next()
+		lastSeenOffset = firstRecord.Offset
+	}
 	records.EachRecord(func(r *kgo.Record) {
 		if r.Offset > lastSeenOffset+1 {
 			gaps = append(gaps, offsetRange{start: lastSeenOffset + 1, end: r.Offset})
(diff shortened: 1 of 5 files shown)


pkg/storage/ingest/fetcher.go
```diff
@@ -466,14 +465,14 @@ func (g offsetRange) numOffsets() int64 {
 	return g.end - g.start
 }
 
-func findGapsInRecords(records []*kgo.Record, lastReturnedOffset int64) []offsetRange {
+func findGapsInRecords(records kgo.Fetches, lastReturnedOffset int64) []offsetRange {
 	var gaps []offsetRange
-	for _, r := range records {
-		if r.Offset != lastReturnedOffset+1 {
+	records.EachRecord(func(r *kgo.Record) {
+		if r.Offset > lastReturnedOffset+1 {
```
dimitarvdimitrov, line 471: `!=` --> `>`  because i saw a test panic that we're trying to report a gap that's negative in value (adding negative value to prometheus counter). I think the only way this can happen is if we re-consme some records. This isn't necessarily a problem, so I'm changing the comparison here to only record missing records, not records already consumed

pracucci: Would be nice to code this scenario in the `findGapsInRecords()` unit test, or at least leave a comment in the code.

pkg/storage/ingest/reader.go
```diff
@@ -188,6 +189,7 @@ func (r *PartitionReader) start(ctx context.Context) (returnErr error) {
 		return err
 	}
 
+	r.lastSeenOffset = lastConsumedOffset
```
pracucci, line 192: [nit] I would add a comment here like:

```
// The lastConsumedOffset could be a special negative offset (e.g. partition start, or partition end).
```

pkg/storage/ingest/fetcher.go
```diff
@@ -466,14 +465,14 @@ func (g offsetRange) numOffsets() int64 {
 	return g.end - g.start
 }
 
-func findGapsInRecords(records []*kgo.Record, lastReturnedOffset int64) []offsetRange {
+func findGapsInRecords(records kgo.Fetches, lastReturnedOffset int64) []offsetRange {
```
pracucci, line 468: [nit] Just to keep consistency.

```suggestion
func findGapsInRecords(records kgo.Fetches, lastSeenOffset int64) []offsetRange {
```

pkg/storage/ingest/fetcher.go
```diff
@@ -466,14 +465,14 @@ func (g offsetRange) numOffsets() int64 {
...
 	var gaps []offsetRange
```
pracucci, line 469: I'm not sure it's correct detect as a gap the case this is the first fetch and `lastSeenOffset` is a special negative value (kafkaOffsetStart or kafkaOffsetEnd). I suggest:

```
if (lastSeenOffset < 0) {
  // If this is the first fetched records batch and we started reading from a special offset (e.g. kafkaOffsetStart or kafkaOffsetEnd)
  // then we don't want to report it as a gap.
  lastSeenOffset = <the offset of the first record>
}
```

pracucci: If you agree with this change, then please update the unit test too.

pkg/storage/ingest/reader.go
```diff
@@ -375,6 +377,10 @@ func (r *PartitionReader) processNextFetches(ctx context.Context, delayObserver
 	r.logFetchErrors(fetches)
 	fetches = filterOutErrFetches(fetches)
 
+	// instrument only after we've done all pre-processing of records. We don't expect the set of records to change beyond this point.
+	instrumentGaps(findGapsInRecords(fetches, r.lastSeenOffset), r.metrics.missedRecords, r.logger)
```
pracucci, line 381: I noticed noticed that we double log in `instrumentGaps()`. The 2nd log should be removed:

```diff
- level.Error(logger).Log("msg", "found gap in records", "start", gap.start, "end", gap.end)
```

pracucci (approved):
LGTM (modulo a couple of nits). Thanks!

pkg/storage/ingest/fetcher.go
dimitarvdimitrov: because we ignore negative gaps, this works without the change. I updated the test

pkg/storage/ingest/reader.go
dimitarvdimitrov: 😕 i have no idea how we ended up here; removed

PR #9976, opened 2024-11-21
fix(distributor-ha): Use a no-op HA tracker when not accepting writes

56quarters:
#### What this PR does

This changes how the HA tracker in the distributor is initialized. Instead of creating it unconditionally when a distributor is started, use a no-op version when the HA tracker is disabled or when running as part of component that does not accept writes but still uses distributor code like queriers and rulers.

Related #9826 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


development/mimir-microservices-mode/docker-compose.jsonnet
```diff
@@ -114,7 +114,7 @@ std.manifestYamlDoc({
         extraArguments:
           // Use of scheduler is activated by `-querier.scheduler-address` option and setting -querier.frontend-address option to nothing.
           if $._config.use_query_scheduler then '-querier.scheduler-address=query-scheduler:9011 -querier.frontend-address=' else '',
-        extraVolumes: ['./config/mimir_override.yaml:/mimir/config/mimir.yaml'],
+        extraVolumes: ['./config/mimir.yaml:/mimir/config/mimir.yaml'],
```
NickAnge, line 117: These can be removed, I think. By default it has the correct volumes ([Ref](https://github.com/grafana/mimir/blob/b8772e6a759d886dc454ae14c9b944c575254567/development/mimir-microservices-mode/docker-compose.jsonnet#L272)) , as part of the mimirService construtor

pkg/distributor/distributor.go
```diff
@@ -504,10 +496,28 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		ingestionRateStrategy = newGlobalRateStrategyWithBurstFactor(limits, d)
 	}
 
+	// If this isn't a real distributor that will be accepting writes or if the HA tracker is
+	// disabled, use a no-op implementation. We don't want to require it to be configured when
+	// it's not used due to being disabled or this being a component that doesn't use it (rulers,
+	// queriers).
+	var haTrackerImpl haTracker
+
+	if !canJoinDistributorsRing || !cfg.HATrackerConfig.EnableHATracker {
```
NickAnge, line 505: Shouldn't the `cfg.HATrackerConfig.EnableHATracker` be part of the HaTracker. Currently as it is if `cfg.HATrackerConfig.EnableHATracker=false`, the newHaTracker will not have any effect 🤔

pkg/distributor/distributor.go
```diff
@@ -504,10 +496,28 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
...
+		haTrackerImpl = newNopHaTracker()
```
NickAnge, line 506: I was thinking also if we should wrap the behaviour in the above if statement, to avoid duplicate check. wdyt ? 

```
if !canJoinDistributorsRing {
 requestRateStrategy = newInfiniteRateStrategy()
....
```

pkg/distributor/distributor.go
```diff
@@ -504,10 +496,28 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
...
+	}
+
 	d.requestRateLimiter = limiter.NewRateLimiter(requestRateStrategy, 10*time.Second)
 	d.ingestionRateLimiter = limiter.NewRateLimiter(ingestionRateStrategy, 10*time.Second)
 	d.distributorsLifecycler = distributorsLifecycler
 	d.distributorsRing = distributorsRing
+	d.HATracker = haTrackerImpl
 
 	d.activeUsers = util.NewActiveUsersCleanupWithDefaultValues(d.cleanupInactiveUser)
 	d.activeGroups = activeGroupsCleanupService
```
NickAnge, line 523: Can we also add a test for it in the distributor side ? Wdyt ?

development/mimir-microservices-mode/docker-compose.jsonnet
56quarters: Will do.

pkg/distributor/distributor.go
56quarters: No, this is on purpose. Having the `defaultHaTracker ` need to be functional when it's not enabled causes extra complication. The HA tracker is entirely internal to the distributor so we're using a no-op version when it's not wanted or needed. `newHaTracker()` should only ever be called when a functional HA tracker is needed.

> if we set cfg.HATrackerConfig.EnableHATracker=false, newHaTracker will always create a KV client etc

`newHaTracker()` is never called when `EnableHATracker` is `false`.

pkg/distributor/distributor.go
56quarters: I have this as its own block on purpose since the logic is slight different and the justification (comment) is different than the other block.

pkg/distributor/distributor.go
NickAnge: I see, make sense. Thanks for the explanation

pkg/distributor/distributor.go
NickAnge: I was thinking a UT in which a distributor was created with  `fg.HATrackerConfig.EnableHATracker` equal to `false`and a noOp tracker was created.

56quarters: Actually, it looks like this is already tested: https://github.com/grafana/mimir/blob/b8772e6a759d886dc454ae14c9b944c575254567/pkg/distributor/distributor_test.go#L878-L886

NickAnge (approved):
LGTM. Thank you

aldernero (approved):

PR #9983, opened 2024-11-21
block-builder: Add basic series limit

codesome:
#### What this PR does

This is a very naive solution to the limits before we have a better solution.

With this PR, we will allow up to the global series allowed for the user _per partition_, iff they global limit is under `block-builder.apply-global-series-under`. We do not apply any limits if the configured limit is over the flag value.

Few examples of it working:
Let's assume `block-builder.apply-global-series-under=10000`
1. If a user has global limit of 5k series, and is sharded between 3 partitions, then it will allow 5k series per partition. So technically this user can ingest up to 15k series.
2. If a user has global limit of 5k series, and is sharded only on 1 partition, then with 5k series per partition, the user will be limited to 5k series as intended.
3. If a user has global limit of 15k series, and no matter how many partitions they are sharded to, while ingesters will apply the proper limits, block builder will not apply any limits (meaning they can ingest anything).

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


pracucci (commented):
I only checked the config to see the overall idea of this change (which LGTM). I let rest of the team reviewing the actual code changes.

pkg/blockbuilder/tsdb.go
```diff
@@ -256,6 +259,15 @@ func (b *TSDBBuilder) newTSDB(tenant tsdbTenant) (*userTSDB, error) {
 	userID := tenant.tenantID
 	userLogger := util_log.WithUserID(userID, b.logger)
 
+	udb := &userTSDB{
+		userID: userID,
+	}
+
+	userLimit := b.limits.MaxGlobalSeriesPerUser(userID)
+	if userLimit <= b.applyGlobalSeriesLimitUnder {
```
seizethedave, line 267: I realize this is a temporary measure, but `Under` doesn't match with `<=`. It's really an `AtMost`. Perhaps `applyGlobalSeriesLimitMax`?

pkg/blockbuilder/tsdb.go
```diff
@@ -256,6 +259,15 @@ func (b *TSDBBuilder) newTSDB(tenant tsdbTenant) (*userTSDB, error) {
...
+		udb.maxGlobalSeries = userLimit
```
seizethedave, line 268: This new code is kinda mysterious on casual read. I think it needs a brief handrail comment about why we're selectively applying limits based on some threshold.

seizethedave (commented):
LGTM modulo commentary 😸

pkg/blockbuilder/tsdb.go
codesome: With the reword to `applyMaxGlobalSeriesPerUser...`, using `Max` as a suffix again sounds odd. How about `applyMaxGlobalSeriesPerUserBelow`? And put in comments that it is inclusive.

codesome:
@seizethedave I updated the PR. PTAL.

seizethedave (approved):

PR #9987, opened 2024-11-22
Alertmanager: add missing version to email receiver config

titolins:
Adds missing version to the email footer

Screenshot 2024-11-22 at 13 59 46

Fixes: https://github.com/grafana/alerting-squad/issues/873

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -486,6 +486,7 @@ func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, 
 		KeyFile:       cfg.Global.HTTPConfig.TLSConfig.KeyFile,
 		SkipVerify:    !cfg.Global.SMTPRequireTLS,
 		StaticHeaders: staticHeaders,
+		Version:       version.Version,
 	}
 	am.emailCfgMtx.Unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/alertmanager.go
```diff
@@ -486,6 +486,7 @@ func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig,
 		KeyFile:       cfg.Global.HTTPConfig.TLSConfig.KeyFile,
 		SkipVerify:    !cfg.Global.SMTPRequireTLS,
 		StaticHeaders: staticHeaders,
+		Version:       version.Version,
```
titolins, line 489: not sure if we'd like to add more info here (e.g. `1.2.3 (mimir)`)

santihernandezc: I think a better alternative would be to modify our default email templates ([txt](https://github.com/grafana/alerting/blob/main/receivers/templates/ng_alert_notification.txt), [html](https://github.com/grafana/alerting/blob/main/receivers/templates/ng_alert_notification.html)) to make the _"Sent by"_ part more flexible, otherwise, it will always say "Sent by Grafana v...".

One option could be having a `SentBy` field instead of a `Version` field. We could pass `"Grafana v <version>"` or `"Mimir v <version>"` depending on where we're executing the template.

Once we update the templates, we can update the alerting package in Grafana and Mimir and use the new field.

titolins:
superseded by https://github.com/grafana/alerting/pull/258

PR #9990, opened 2024-11-22
add probes for cache statefulsets - issue#9771 https://github.com/grafana/mimir/issues/9771

### LINKED ISSUE #9771: Add probes for cache statefulsets

I've encountered an operational challenge in monitoring and ensuring the health of the caches (result, chunks, index,metadata )StatefulSets efficiently. Currently, there's no straightforward method to integrate Kubernetes readiness and liveness probes directly through the Helm chart configurations. This makes it difficult to automatically manage the healthy state and readiness of our deployed services, impacting our ability to reliably scale and maintain our infrastructure.

I would like the Helm chart for the cache StatefulSets to support configuration options that enable the easy integration of Kubernetes probes.

An alternative solution was manually modifying the deployment templates post-Helm generation to include these probes, but this approach is not maintainable or scalable as it bypasses the advantages of using Helm for deployment configurations. Using pre-hooks in Helm to modify deployments post-deployment was also considered but dismissed for the same reasons.

Sabian-A:
#### What this PR does
add probes for cache statefulsets.
Fixes #9771

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/templates/memcached/_memcached-statefulset.tpl
+++ b/operations/helm/charts/mimir-distributed/templates/memcached/_memcached-statefulset.tpl
@@ -131,6 +131,26 @@ spec:
             {{- toYaml . | nindent 12 }}
             {{- end }}
 
+          {{- if $.ctx.Values.memcached.readinessProbe.enabled }}
+          readinessProbe:
+            tcpSocket:
+              port: {{ .port }}
+            initialDelaySeconds: {{ $.ctx.Values.memcached.readinessProbe.initialDelaySeconds }}
+            periodSeconds: {{ $.ctx.Values.memcached.readinessProbe.periodSeconds }}
+            successThreshold: {{ $.ctx.Values.memcached.readinessProbe.successThreshold }}
+            failureThreshold: {{ $.ctx.Values.memcached.readinessProbe.failureThreshold }}
+          {{- end }}
+      
+          {{- if $.ctx.Values.memcached.livenessProbe.enabled }}
+          livenessProbe:
+            tcpSocket:
+              port: {{ .port }}
+            initialDelaySeconds: {{ $.ctx.Values.memcached.livenessProbe.initialDelaySeconds }}
+            periodSeconds: {{ $.ctx.Values.memcached.livenessProbe.periodSeconds }}
+            successThreshold: {{ $.ctx.Values.memcached.livenessProbe.successThreshold }}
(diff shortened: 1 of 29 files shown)


operations/helm/charts/mimir-distributed/templates/memcached/_memcached-statefulset.tpl
```diff
@@ -131,6 +131,26 @@ spec:
...
+            tcpSocket:
+              port: {{ .port }}
+            initialDelaySeconds: {{ $.ctx.Values.memcached.readinessProbe.initialDelaySeconds }}
+            periodSeconds: {{ $.ctx.Values.memcached.readinessProbe.periodSeconds }}
+            successThreshold: {{ $.ctx.Values.memcached.readinessProbe.successThreshold }}
+            failureThreshold: {{ $.ctx.Values.memcached.readinessProbe.failureThreshold }}
+          {{- end }}
+      
+          {{- if $.ctx.Values.memcached.livenessProbe.enabled }}
+          livenessProbe:
```
56quarters, line 145: A liveness probe doesn't make sense for Memcached. A liveness probe means that a pod is broken and cannot recover except by being restarted. This isn't the case for Memcached, it should always be able to recover as long as it's still running.

56quarters (commented):
I'm sorry to change my mind but upon further consideration, I don't think these probes make sense for Memcached. 

It definitely shouldn't have a liveness probe and I'm leaning towards not having a readiness probe. Memcached is run as a statefulset with a headless service. When a readiness probe fails, Memcached would be removed from the headless service and Mimir would stop reading/writing to it. The only reasons the check would fail is hitting a connection limit. Hitting a connection limit is a transient condition and we wouldn't want to remove the cache pod from the service in that case since it will recover eventually.

Back to the original issue opened, I don't think there's any more information to be gained by readiness check:

> I specifically need the Readiness check because, in our GKE setup, the Pod Disruption Budget (PDB) relies on the readiness probe to determine if a pod is healthy and can handle traffic. Without the readiness check, the PDB has limited visibility into the pod’s true health status, only knowing that it’s up and running but not if it’s fully ready to serve requests.

Memcached doesn't have any work to do to be ready to handle traffic: it's either up and accepting connection or not. Thank you for the PR but I don't think it makes sense for Memcached.

Sabian-A:
@56quarters The readiness probe is mandatory for GKE to do out of the box upgrades and its needed to run mimir on GKE seamlessly so what do you suggest as a solution ?

PR #9997, opened 2024-11-22
Use resident set memory for the etcd memory alert.

seizethedave:
#### What this PR does

This changes the `EtcdAllocatingTooMuchMemory` alerts to use RSS memory rather than `working_set`.
The reason for this is etcd can use a **lot** of file cache memory for its file-backed database that will be dropped if it approaches the container's memory limit. This causes this alert to basically be scare-noise, at times. We had a situation where the Go process only had 50MiB of memory resident, but there was 700MiB of file cache memory.

RSS fits better with "AllocatingTooMuch" anyway.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


duricanikolic (approved):
LGTM

dimitarvdimitrov (approved):
makes sense, thanks

PR #9998, opened 2024-11-24
MQE: fix issue where subqueries could return series with no points

charleskorn:
#### What this PR does

This PR fixes an issue where MQE could return invalid results if a subquery was evaluated as the top-level expression for an instant query.

If a series had all of its samples filtered out, MQE would return a series with no samples, for example the query `(avg(foo) > 100)[10m:30s]` could return a result like this:

```json
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "env": "prod"
        }
      }
    ]
  }
}
```

It should instead not return the series at all:

```json
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
    ]
  }
}
```

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 * [CHANGE] Ingester: remove experimental flags `-ingest-storage.kafka.ongoing-records-per-fetch` and `-ingest-storage.kafka.startup-records-per-fetch`. They are removed in favour of `-ingest-storage.kafka.max-buffered-bytes`. #9906
 * [CHANGE] Ingester: Replace `cortex_discarded_samples_total` label from `sample-out-of-bounds` to `sample-timestamp-too-old`. #9885
 * [CHANGE] Ruler: the `/prometheus/config/v1/rules` does not return an error anymore if a rule group is missing in the object storage after been successfully returned by listing the storage, because it could have been deleted in the meanwhile. #9936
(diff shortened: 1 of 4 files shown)


jhesketh (commented):
lgtm

However do we have a test where this happens with a sub-query at not the top of the query. I believe it should be fine, but I think we should have a test for it.

charleskorn:
> However do we have a test where this happens with a sub-query at not the top of the query. I believe it should be fine, but I think we should have a test for it.

Do you mean something like `abs((metric > Inf)[20s:10s])`, or something like `((metric > Inf)[20s:10s])[30s:20s]`?

jhesketh:
> Do you mean something like last_over_time((metric > Inf)[20s:10s]), or something like ((metric > Inf)[20s:10s])[30s:20s]?

I meant the former, but why not both? ¯\_(ツ)_/¯

charleskorn:
> > Do you mean something like last_over_time((metric > Inf)[20s:10s]), or something like ((metric > Inf)[20s:10s])[30s:20s]?
> 
> I meant the former, but why not both? ¯_(ツ)_/¯

Added in [961f2e5](https://github.com/grafana/mimir/pull/9998/commits/961f2e57dc657e8427a7470da4ec7401ce0107d4)

jhesketh (approved):

jhesketh:
Thanks!

PR #10010, opened 2024-11-25
MQE: mangle native histograms in returned slices to catch use-after-return bugs + fix issues found

charleskorn:
#### What this PR does

This PR extends MQE's tests to aid in catching native histogram use-after-return bugs, and fixes some issues they've uncovered.

Now, during tests in the `streamingpromql` package, any histograms in slices returned to the `HPoint` or `FloatHistogram` pools will immediately be mutated. This ensures that any possible use-after-return bugs are triggered deterministically.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/types/limiting_pool.go
+++ b/pkg/streamingpromql/types/limiting_pool.go
@@ -29,12 +29,17 @@ const (
 )
 
 var (
+	// EnableManglingReturnedSlices enables mangling values in slices returned to pool to aid in detecting use-after-return bugs.
+	// Only used in tests.
+	EnableManglingReturnedSlices = false
+
 	FPointSlicePool = NewLimitingBucketedPool(
 		pool.NewBucketedPool(1, maxExpectedPointsPerSeries, pointsPerSeriesBucketFactor, func(size int) []promql.FPoint {
 			return make([]promql.FPoint, 0, size)
 		}),
 		FPointSize,
 		false,
+		nil,
 	)
 
 	HPointSlicePool = NewLimitingBucketedPool(
@@ -43,6 +48,10 @@ var (
 		}),
 		HPointSize,
 		false,
+		func(point promql.HPoint) promql.HPoint {
+			point.H = mangleHistogram(point.H)
+			return point
+		},
 	)
 
 	VectorPool = NewLimitingBucketedPool(
@@ -51,6 +60,7 @@ var (
 		}),
 		VectorSampleSize,
 		false,
+		nil,
 	)
 
 	Float64SlicePool = NewLimitingBucketedPool(
@@ -59,6 +69,7 @@ var (
 		}),
 		Float64Size,
 		true,
+		nil,
 	)
 
 	BoolSlicePool = NewLimitingBucketedPool(
@@ -67,6 +78,7 @@ var (
 		}),
 		BoolSize,
 		true,
+		nil,
 	)
 
 	HistogramSlicePool = NewLimitingBucketedPool(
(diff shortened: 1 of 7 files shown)


pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -75,9 +87,34 @@ var (
...
 )
 
+func mangleHistogram(h *histogram.FloatHistogram) *histogram.FloatHistogram {
+	if h == nil {
+		return nil
+	}
+
+	h.ZeroCount = 12345678
+	h.Count = 12345678
+	h.Sum = 12345678
```
jhesketh, line 101: (nit) it's *very* unlikely, but if we write a test with these values elsewhere the mangle won't work. Maybe for just one of these (eg Sum) we multiply it or similar.
```suggestion
	h.Sum = h.Sum * 9.87654321
```

pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -86,13 +123,15 @@ type LimitingBucketedPool[S ~[]E, E any] struct {
 	inner       *pool.BucketedPool[S, E]
 	elementSize uint64
 	clearOnGet  bool
+	mangle      func(E) E
 }
 
-func NewLimitingBucketedPool[S ~[]E, E any](inner *pool.BucketedPool[S, E], elementSize uint64, clearOnGet bool) *LimitingBucketedPool[S, E] {
+func NewLimitingBucketedPool[S ~[]E, E any](inner *pool.BucketedPool[S, E], elementSize uint64, clearOnGet bool, mangle func(E) E) *LimitingBucketedPool[S, E] {
```
jhesketh, line 129: we should add a quick test that `mangle func(E) E` is called when the setting is true/false etc.

charleskorn: Added in 26391e9.

pkg/streamingpromql/types/limiting_pool.go
charleskorn: I'm keen to keep this as simple as possible and to make the resulting histogram easy to identify. I think the likelihood of us using a histogram with these values elsewhere and also having a bug that is then hidden because of that is very, very unlikely, as you say.

pkg/streamingpromql/types/limiting_pool_test.go
```diff
@@ -190,6 +192,39 @@ func TestLimitingPool_ClearsReturnedSlices(t *testing.T) {
 	})
 }
 
+func TestLimitingPool_Mangling(t *testing.T) {
+	defer func() {
+		// Ensure we reset this back to the default state given it applies globally.
+		EnableManglingReturnedSlices = false
+	}()
```
jhesketh, line 199: (nit) In case we change where this is turned on in the future etc.

```suggestion
	currentEnableManglingReturnedSlices := EnableManglingReturnedSlices
	defer func() {
		// Ensure we reset this back to the default state given it applies globally.
		EnableManglingReturnedSlices = currentEnableManglingReturnedSlices
	}()
	EnableManglingReturnedSlices = false
```

jhesketh (approved):
lgtm, sans a suggestion

pkg/streamingpromql/types/limiting_pool.go
jhesketh: Fair enough, I don't hold this strongly.

pkg/streamingpromql/types/limiting_pool_test.go
charleskorn: Added in 7293a8b.

PR #10025, opened 2024-11-26
Distributor: Simplify OTel resource attribute promotion pluggability

jesusvazquez:
Simplify pluggable logic for getting OTel resource attributes to promote.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -170,6 +170,9 @@ func OTLPHandler(
 		}
 		addSuffixes := limits.OTelMetricSuffixesEnabled(tenantID)
 		enableCTZeroIngestion := limits.OTelCreatedTimestampZeroIngestionEnabled(tenantID)
+		if resourceAttributePromotionConfig == nil {
+			resourceAttributePromotionConfig = limits
+		}
 		promoteResourceAttributes := resourceAttributePromotionConfig.PromoteOTelResourceAttributes(tenantID)
 
 		pushMetrics.IncOTLPRequest(tenantID)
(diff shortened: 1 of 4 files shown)


pkg/api/api.go
```diff
@@ -269,7 +269,7 @@ func (a *API) RegisterDistributor(d *distributor.Distributor, pushConfig distrib
 		a.cfg.SkipLabelCountValidationHeader, limits, pushConfig.RetryConfig, d.PushWithMiddlewares, d.PushMetrics, a.logger,
 	), true, false, "POST")
 	a.RegisterRoute(OTLPPushEndpoint, distributor.OTLPHandler(
-		pushConfig.MaxOTLPRequestSize, d.RequestBufferPool, a.sourceIPs, limits, pushConfig.OTelResourceAttributePromotionConfig,
+		pushConfig.MaxOTLPRequestSize, d.RequestBufferPool, a.sourceIPs, limits, limits,
```
jesusvazquez, line 272: @aknuds1 Arve is this correct? I understand that if there is no override for a user it will fallback to `pushConfig.OTelResourceAttributePromotionConfig`

aknuds1: At least it looks weird to me to have to pass `limits` twice. Checking.

jesusvazquez: This is because limits implements the interface that returns the `OTelResourceAttributePromotionConfig`

colega:
Can we please reopen this PR and apply this change?

I don't understand why we have an interface, that should serve for the most basic application of polymorphism through the [strategy pattern](https://www.digitalocean.com/community/tutorials/strategy-design-pattern-in-java-example-tutorial), but instead we have a wonky `if` clause and decide whether we're using `resourceAttributePromotionConfig` or something else instead.

This code doesn't look nice right now.

aknuds1:
@colega I can make a PR that doesn't require this `if`.

colega:
@aknuds1 I would appreciate.

aknuds1:
@colega yes, still on my mind, have focused on the limits operator in the meantime.

PR #10026, opened 2024-11-26
frontend: update query pruning

krajorama:
#### What this PR does

The current method of excluding/including sub query results in PromQL by comparing to -Inf or +Inf is no longer valid after https://github.com/prometheus/prometheus/pull/15245 due to comparison of native histograms to a float with < or > result in Jeanette's warning, not empty set.

To ease migration away from the old wrong logic, I'm adding the new logic first.

The new method uses logical AND operation to intersect the sub query with either a const vector() or an empty vector(). E.g.

```
subquery and on() (vector(1)==1)
```

which becomes:

```
subquery
```

and conversely 

```
subquery and on() (vector(-1)==1)
```
becomes
```
(vector(-1)==1)
```

#### Note

We cannot just drop `(vector(-1)==1)` altogether , because of this example:
```
(other_subquery) and (subquery and on() (vector(-1)==1))
```
removal of the whole right side would allow results from the `other_subquery` which is not correct. Solving this is out of scope.

Fixes N/A

#### Checklist

- [X] Tests updated.
- N/A Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 * [ENHANCEMENT] Ingester: Add `-blocks-storage.tsdb.bigger-out-of-order-blocks-for-old-samples` to build 24h blocks for out-of-order data belonging to the previous days instead of building smaller 2h blocks. This reduces pressure on compactors and ingesters when the out-of-order samples span multiple days in the past. #9844 #10033 #10035
 * [ENHANCEMENT] Distributor: allow a different limit for info series (series ending in `_info`) label count, via `-validation.max-label-names-per-info-series`. #10028
 * [ENHANCEMENT] Ingester: do not reuse labels, samples and histograms slices in the write request if there are more entries than 10x the pre-allocated size. This should help to reduce the in-use memory in case of few requests with a very large number of labels, samples or histograms. #10040
+* [ENHANCEMENT] Query-Frontend: prune `<subquery> and on() (vector(x)==y)` style queries and no longer prune `<subquery> < -Inf`. Triggered by https://github.com/prometheus/prometheus/pull/15245. #10026
 * [BUGFIX] Fix issue where functions such as `rate()` over native histograms could return incorrect values if a float stale marker was present in the selected range. #9508
 * [BUGFIX] Fix issue where negation of native histograms (eg. `-some_native_histogram_series`) did nothing. #9508
 * [BUGFIX] Fix issue where `metric might not be a counter, name does not end in _total/_sum/_count/_bucket` annotation would be emitted even if `rate` or `increase` did not have enough samples to compute a result. #9508
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 * [ENHANCEMENT] Ingester: Add `-blocks-storage.tsdb.bigger-out-of-order-blocks-for-old-samples` to build 24h blocks for out-of-order data belonging to the previous days instead of building smaller 2h blocks. This reduces pressure on compactors and ingesters when the out-of-order samples span multiple days in the past. #9844 #10033 #10035
 * [ENHANCEMENT] Distributor: allow a different limit for info series (series ending in `_info`) label count, via `-validation.max-label-names-per-info-series`. #10028
 * [ENHANCEMENT] Ingester: do not reuse labels, samples and histograms slices in the write request if there are more entries than 10x the pre-allocated size. This should help to reduce the in-use memory in case of few requests with a very large number of labels, samples or histograms. #10040
-* [ENHANCEMENT] Query-Frontend: prune `<subquery> and on() (vector(x)==y)` style queries and no longer prune `<subquery> < -Inf`. Triggered by https://github.com/prometheus/prometheus/pull/15245. #10026
+* [ENHANCEMENT] Query-Frontend: prune `<subquery> and on() (vector(x)==y)` style queries and stop pruning `<subquery> < -Inf`. Triggered by https://github.com/prometheus/prometheus/pull/15245. #10026
 * [BUGFIX] Fix issue where functions such as `rate()` over native histograms could return incorrect values if a float stale marker was present in the selected range. #9508
 * [BUGFIX] Fix issue where negation of native histograms (eg. `-some_native_histogram_series`) did nothing. #9508
 * [BUGFIX] Fix issue where `metric might not be a counter, name does not end in _total/_sum/_count/_bucket` annotation would be emitted even if `rate` or `increase` did not have enough samples to compute a result. #9508
--- a/pkg/frontend/querymiddleware/astmapper/pruning.go
+++ b/pkg/frontend/querymiddleware/astmapper/pruning.go
@@ -51,8 +51,8 @@ func (pruner *queryPruner) MapExpr(expr parser.Expr) (mapped parser.Expr, finish
 		// "and on()", return the right hand side.
 		return e.RHS, false, nil
 	}
-	// The right hand side is const no empty, so the whole expression is just the
-	// left side.
+	// The right hand side is const and not empty, so the whole expression is
+	// just the left side.
 	return e.LHS, false, nil
 }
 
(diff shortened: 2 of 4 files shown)


pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
```diff
@@ -0,0 +1,179 @@
...
+			`(vector(3) == 4.5)`,
+		},
+		{
+			`(avg(rate(foo[1m]))) and on() (vector(5.5) == 5.5)`,
+			`(avg(rate(foo[1m])))`,
+		},
+		{
+			// "and on()" is not on top level, "or" has lower precedence.
+			`(avg(rate(foo[1m]))) and on() (vector(0) == 1) or avg(rate(bar[1m]))`,
+			`(vector(0) == 1) or avg(rate(bar[1m]))`,
```
charleskorn, line 154: Couldn't we simplify this further, to `avg(rate(bar[1m]))`?

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		{
+			// "and on()" is not on top level, due to left-right associativity.
+			`(avg(rate(foo[1m]))) and on() (vector(0) == 1) and avg(rate(bar[1m]))`,
+			`(vector(0) == 1) and avg(rate(bar[1m]))`,
```
charleskorn, line 159: Couldn't we simplify this further, to `vector(0) == 1`?

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
```diff
@@ -0,0 +1,179 @@
...
+			`((((-1 * -Inf) < avg(rate(foo[4m]))) unless avg(rate(foo[3m]))) and avg(rate(foo[2m]))) or avg(rate(foo[1m]))`,
+			`((((-1 * -Inf) < avg(rate(foo[4m]))) unless avg(rate(foo[3m]))) and avg(rate(foo[2m]))) or avg(rate(foo[1m]))`,
+		},
+		{
+			`avg(rate(foo[1m])) or avg(rate(bar[1m]))`,
+			`avg(rate(foo[1m])) or avg(rate(bar[1m]))`,
+		},
+		{ // The const expression is on the wrong side.
+			`(vector(0) == 1) and on() (avg(rate(foo[1m])))`,
+			`(vector(0) == 1) and on() (avg(rate(foo[1m])))`,
```
charleskorn, line 124: Same as below - I think this can be simplified to `vector(0) == 1`.

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		{ // Matching on labels.
+			`(avg(rate(foo[1m]))) and on(id) (vector(0) == 1)`,
+			`(avg(rate(foo[1m]))) and on(id) (vector(0) == 1)`,
```
charleskorn, line 128: Couldn't this be simplified to `vector(0) == 1`? The right side will never produce a result, so we'll never return anything from the left side.

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+		{ // Not "on" expression.
+			`(avg(rate(foo[1m]))) and ignoring() (vector(0) == 1)`,
+			`(avg(rate(foo[1m]))) and ignoring() (vector(0) == 1)`,
```
charleskorn, line 132: Same as above - the right side will never produce a result.

charleskorn (commented):
I think we can further prune the expression in some cases:

beorn7 (commented):
Looks good at first glance. (I assume you don't want a code-level review from me, just that the idea is correct.)

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
krajorama: That's more complicated. The main point is to avoid loading chunks and `(vector(0)==1)` doesn't load chunks, so should be very  efficient.
We can add that logic in a new PR, seems to be independent from this optimization, wdyt?
But having two algorithms raises the question of whether they need to be run repeatedly to optimize away everything, so it gets even more complicated.

krajorama: Note that removing `(vector(0)==1)` from an OR expression is different from removing it from an AND expression, which is what I mean by more complicated . Outside my scope.

colega: Is there a test for this? We should document that in a test so nobody comes tomorrow and "optimizes further".

charleskorn: >having two algorithms raises the question of whether they need to be run repeatedly to optimize away everything, so it gets even more complicated.

If we optimise the leaf node expressions first and work our way back up the tree, can't we do both `and` and `or` in one pass?

But if you want to limit this to `and` and come back to `or` later, that's fine by me, please just make this clear with a test or a comment as Oleg suggests.

krajorama: adding both

krajorama: well, actually I had 3 tests on this already so just improved the comments

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@
 * [ENHANCEMENT] Ingester: Add `-blocks-storage.tsdb.bigger-out-of-order-blocks-for-old-samples` to build 24h blocks for out-of-order data belonging to the previous days instead of building smaller 2h blocks. This reduces pressure on compactors and ingesters when the out-of-order samples span multiple days in the past. #9844 #10033 #10035
 * [ENHANCEMENT] Distributor: allow a different limit for info series (series ending in `_info`) label count, via `-validation.max-label-names-per-info-series`. #10028
 * [ENHANCEMENT] Ingester: do not reuse labels, samples and histograms slices in the write request if there are more entries than 10x the pre-allocated size. This should help to reduce the in-use memory in case of few requests with a very large number of labels, samples or histograms. #10040
+* [ENHANCEMENT] Query-Frontend: prune `<subquery> and on() (vector(x)==y)` style queries and no longer prune `<subquery> < -Inf`. Triggered by https://github.com/prometheus/prometheus/pull/15245. #10026
```
tacole02, line 80: Nit: `no longer prune` >> `stop pruning`

pkg/frontend/querymiddleware/astmapper/pruning.go
```diff
@@ -33,185 +31,89 @@ func (pruner *queryPruner) MapExpr(expr parser.Expr) (mapped parser.Expr, finish
...
+	lIsConst, lValue = pruner.isNumber(lhs)
+	if !lIsConst {
+		return false, false
 	}
-
-	return expr
+	rIsVector, rValue := pruner.isConstVector(rhs)
+	if !rIsVector {
+		return false, false
+	}
```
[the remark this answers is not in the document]
krajorama: updated, ptal

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
krajorama: out of scope

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
krajorama: Out of scope, not a useful expression in the first place . Also true is it had `vector(1)`.

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
krajorama: out of scope

pkg/frontend/querymiddleware/astmapper/pruning_new_test.go
krajorama: out of scope

tacole02 (approved):

charleskorn (approved):

PR #10029, opened 2024-11-26
helm: don't quote container command args passed via *.extraArgs values

narqo:
#### What this PR does

This PR fixes how the chart's templates handle `extraArgs` values. Right now the chart fails to render the manifests if the passed flag contains a JSON value. E.g. flags like `-ruler.max-rule-groups-per-tenant-by-namespace`, etc accept a JSON string. If a user sets the values via the `ruler.extraArgs` value — similar to how we use it in Jsonnet — the chart fails with an obscure error:

```
Error: YAML parse error on mimir-distributed/templates/ruler/ruler-dep.yaml: error converting YAML to JSON: yaml: line 61: did not find expected '-' indicator
```

_Note: I tried fairly hard to find a way how to handle that in better way but failed. After looking around for how other popular Helm charts solve it didn't help either. Everyone seem to just unquote the command args._

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov:
another idea is to use something like `printf "-{{key}}={{value}}" | quote`

but i also can't think of a good reason why these values were quoted in the first place - are we introducing a breaking change by removing the quotes?

narqo:
> another idea is to use something like `printf "-{{key}}={{value}}" | quote`

I tried `{{ printf "-%s=%v" key value | quote }}` but `printf` converts the value to its string representation. This breaks values like `nil` (and maybe others). I didn't find a knob to force `printf` to output the value as-is.

dimitarvdimitrov (approved):
LGTM

PR #10030, opened 2024-11-26
Ruler: Add support for arbitrary extra URL parameters to Alertmanager client's OAuth config

alexweav:
#### What this PR does

This PR allows you to specify arbitrary URL parameters to send alongside OAuth token requests, when the Alertmanager client is configured in OAuth mode.

Some OAuth providers require extra fields here that are not normally part of the spec - for example, `audience` is a common one. The Go stdlib followed a similar convention for such fields: https://github.com/golang/oauth2/issues/216 where arbitrary extra URL values are allowed.

n/a

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/util/validation/limits_map.go
+++ b/pkg/util/validation/limits_map.go
@@ -16,8 +16,12 @@ type LimitsMap[T float64 | int | string] struct {
 }
 
 func NewLimitsMap[T float64 | int | string](validator func(k string, v T) error) LimitsMap[T] {
+	return NewLimitsMapWithData(make(map[string]T), validator)
+}
+
+func NewLimitsMapWithData[T float64 | int | string](data map[string]T, validator func(k string, v T) error) LimitsMap[T] {
 	return LimitsMap[T]{
-		data:      make(map[string]T),
+		data:      data,
 		validator: validator,
 	}
 }
--- a/pkg/util/validation/limits_map.go
+++ b/pkg/util/validation/limits_map.go
@@ -73,6 +73,7 @@ func (m LimitsMap[T]) updateMap(newMap map[string]T) error {
 		}
 	}
 
+	clear(m.data)
 	for k, v := range newMap {
 		m.data[k] = v
 	}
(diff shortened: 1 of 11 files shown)


pkg/util/validation/limits_map.go
```diff
@@ -9,15 +9,19 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// LimitsMap is a generic map that can hold either float64 or int as values.
-type LimitsMap[T float64 | int] struct {
+// LimitsMap is a flag.Value implementation that looks like a generic map, holding float64s, ints, or strings as values.
+type LimitsMap[T float64 | int | string] struct {
```
alexweav, line 13: So far, the only place where we really use map flags is in Limits, that's why this was called LimitsMap. But, nothing in this file is really specific to limits.

This new use case isn't really a limit, but I felt it gives a better configuration experience to be consistent with existing flags here, rather than invent a second map format that behaves differently.

I've kept the name `LimitsMap` and package as i feel it's _close enough_ to being descriptive and would muddy the diff to include renames/moves. I also think there's a valid case for moving this to `dskit`'s `flagext` package in the future anyway, so it didn't make sense to do a big rename yet.

I'm happy to change the name, or explore moving this to dskit, if reviewers would prefer. Or, this can be left to a subsequent PR.

pkg/util/validation/limits_map.go
```diff
@@ -73,6 +81,7 @@ func (m LimitsMap[T]) updateMap(newMap map[string]T) error {
 		}
 	}
 
+	clear(m.data)
```
alexweav, line 84: drive-by bugfix i happened to notice. added corresponding test.

tacole02 (approved):
👍

francoposa (approved):
LGTM - I think putting it here is probably the right route for now.

I will say that even though it's a minor bit of code I don't think we need to be in the habit of validating something like these endpoint param as `validateOAuth2EndpointParam` does. Empty query param values are valid for HTTP, and since this config is for for connecting to/through some external service, it's not really our concern to tell the operator what is or is not valid.

PR #10031, opened 2024-11-27
MQE: fix incorrect peak memory consumption value reported for benchmarks run on Linux

charleskorn:
#### What this PR does

This PR fixes an issue where the peak memory consumption of benchmarks run on Linux is off by a factor of 1024x.

I haven't added a changelog entry given this is not a user-facing change.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/benchmark-query-engine/main.go
+++ b/tools/benchmark-query-engine/main.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


lamida (approved):

tools/benchmark-query-engine/main.go
```diff
@@ -339,11 +340,22 @@ func (a *app) runTestCase(name string, printBenchmarkHeader bool) error {
...
 		}
 	}
 
 	return nil
 }
+
+func maxRSSInBytes(usage *syscall.Rusage) int64 {
+	switch runtime.GOOS {
+	case "Linux":
+		return usage.Maxrss * 1024 // Maxrss is returned in kilobytes on Linux.
```
jhesketh, line 355: I'm confused by this comment.. If it's returned in kilobytes, why multiply it by 1024? Wouldn't you divide it?

jhesketh (changes requested):

tools/benchmark-query-engine/main.go
charleskorn: `usage.Maxrss` is given in kilobytes, but we want to return a value in bytes from `maxRSSInBytes`.

For example, if `usage.Maxrss` is 1, then that represents 1 KB, or 1024 bytes.

jhesketh: Right, that makes complete sense on re-reading, I'm not sure what I was thinking. I'll blame covid-brain. Sorry for the hold up!

jhesketh (approved):

tools/benchmark-query-engine/main.go
charleskorn: No worries 🙂

PR #10032, opened 2024-11-27
Remove "track_sizes" option from Memcached in jsonnet and helm

56quarters:
#### What this PR does

The option tracks the size of items in the cache in a histogram. We don't use this information for anything or monitor it in any way so remove it.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
Is this adding a lot of extra cardinality?

56quarters:
> Is this adding a lot of extra cardinality?

It's not actually exposed as Prometheus metrics, just the `stats sizes` Memcached command. It's not enabled by default since there's some overhead but the counts are actually lossy because they don't use locking or atomics.

PR #10036, opened 2024-11-28
[r314] Put an hard-coded cap on the pooled TimeSeries labels, samples and histograms slices

pracucci:
#### What this PR does

**DO NOT MERGE**. This is just an experiment.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/timeseries.go
+++ b/pkg/mimirpb/timeseries.go
@@ -19,11 +19,14 @@ import (
 )
 
 const (
-	minPreallocatedTimeseries         = 100
-	minPreallocatedLabels             = 20
-	minPreallocatedSamplesPerSeries   = 10
-	minPreallocatedExemplarsPerSeries = 1
-	maxPreallocatedExemplarsPerSeries = 10
+	minPreallocatedTimeseries          = 100
+	minPreallocatedLabels              = 20
+	maxPreallocatedLabels              = 100
+	minPreallocatedSamplesPerSeries    = 10
+	maxPreallocatedSamplesPerSeries    = 50
+	maxPreallocatedHistogramsPerSeries = 50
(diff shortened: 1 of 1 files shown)


NickAnge (approved):

duricanikolic (approved):
LGTM

pkg/mimirpb/timeseries.go
```diff
@@ -19,11 +19,14 @@ import (
...
-	minPreallocatedTimeseries         = 100
-	minPreallocatedLabels             = 20
-	minPreallocatedSamplesPerSeries   = 10
-	minPreallocatedExemplarsPerSeries = 1
-	maxPreallocatedExemplarsPerSeries = 10
+	minPreallocatedTimeseries          = 100
+	minPreallocatedLabels              = 20
+	maxPreallocatedLabels              = 100
+	minPreallocatedSamplesPerSeries    = 10
+	maxPreallocatedSamplesPerSeries    = 50
```
colega, line 26: Not sure if this is too aggressive, it won't make a difference if this is 10x of that in terms of memory usage, but I also don't have any data to support this.


(If I was writing this code, I'd do 1024 here or something like that)

colega (approved):

pkg/mimirpb/timeseries.go
pracucci: In the Prometheus world. We typically expect 1 sample per series. When remote write is catching up we could get multiple samples. For example, assuming 4DPM, 50 samples = 11 minutes of data. It still looks a lot coming from a single WriteRequest, considering how WAL is replayed.

What's the legit case you have in mind for 1K samples for 1 series in 1 request?

pracucci:
See https://github.com/grafana/mimir/pull/10040 instead.

PR #10038, opened 2024-11-28
query-scheduler: replace context.Cancelled with custom error

dimitarvdimitrov:
context cancelled is less informative

@colega asked if this can be a real context cancellation. I think it can't because:

so the cancellation of the real request is tracked a few lines down here

https://github.com/grafana/mimir/blob/f8155dda490b0efd35c8a8e43d715c07d2bcb159/pkg/scheduler/scheduler.go#L543-L549

the error we get in the changes in this PR is controlled by the querier. The querier variable is the thing which gRPC passes to us  when a querier makes a connection to the query-scheduler.

https://github.com/grafana/mimir/blob/f8155dda490b0efd35c8a8e43d715c07d2bcb159/pkg/scheduler/scheduler.go#L421-L423

And then the querier would cancel the connection on its end only after waiting for all queries to finish here (code is complicated; [here is the entrypoint](https://github.com/grafana/mimir/blob/9b742c1b701f6c0690b12e4aaa25663b4b498299/pkg/querier/worker/processor_manager.go#L74-L84) for all that)

so if we get a cancellation where i made a change it’s either a slim race condition with a querier shutting down right after we’ve selected it to do some work or the querier crashed while serving our query

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/grafana/dskit/cancellation"
 	"github.com/grafana/dskit/grpcclient"
+	"github.com/grafana/dskit/grpcutil"
 	"github.com/grafana/dskit/httpgrpc"
 	"github.com/grafana/dskit/middleware"
 	"github.com/grafana/dskit/ring"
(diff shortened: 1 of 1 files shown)


narqo (approved):
So we still want to reply to the frontend with a server error (5xx) but only to hide the fact of the context cancellation. Is this the problem we're solving here?

Providing the above, the change works for me 🔥

dimitarvdimitrov:
> So we still want to reply to the frontend with a server error (5xx) but only to hide the fact of the context cancellation. Is this the problem we're solving here?

yes. We want to propagate a more descriptive error. It also leaves the knowledge of what a context cancellation means in the code, so someone doesn't have to dig for it again

PR #10041, opened 2024-11-28
Jsonnet: add ingest_storage_kafka_client_id_settings config option

pracucci:
#### What this PR does

Adding a generic `ingest_storage_kafka_client_id_settings` config option to Jsonnet, and then having both `ingest_storage_kafka_producer_client_id_settings` and `ingest_storage_kafka_consumer_client_id_settings` inherit from it. This is useful if you want to set a client ID for every Kafka client, both producer and consumer.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir/ingest-storage.libsonnet
+++ b/operations/mimir/ingest-storage.libsonnet
@@ -47,8 +47,9 @@
 
   // The default Kafka client ID settings to use for producers and consumers.
   // These key-value settings get serialised into a comma-separated string.
-  ingest_storage_kafka_producer_client_id_settings:: {},
-  ingest_storage_kafka_consumer_client_id_settings:: {},
+  ingest_storage_kafka_client_id_settings:: {},
+  ingest_storage_kafka_producer_client_id_settings:: $.ingest_storage_kafka_client_id_settings,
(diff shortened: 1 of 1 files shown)


operations/mimir/ingest-storage.libsonnet
```diff
@@ -47,8 +47,9 @@
 
   // The default Kafka client ID settings to use for producers and consumers.
   // These key-value settings get serialised into a comma-separated string.
-  ingest_storage_kafka_producer_client_id_settings:: {},
-  ingest_storage_kafka_consumer_client_id_settings:: {},
+  ingest_storage_kafka_client_id_settings:: {},
+  ingest_storage_kafka_producer_client_id_settings:: $.ingest_storage_kafka_client_id_settings,
```
julienduchesne, line 51: Should be `$._config.ingest_storage_kafka_client_id_settings` or `self.ingest_storage_kafka_client_id_settings`? 🤔

duricanikolic (approved):
LGTM

operations/mimir/ingest-storage.libsonnet
julienduchesne: nvm, I see similar references in the file. Seems like `$.` gets the references within the same object also

julienduchesne (approved):

PR #10043, opened 2024-11-28
ruler: Register custom discovery config

aknuds1:
#### What this PR does

With Prometheus 3, it's required to register custom `discovery.Config` implementations, otherwise they [cannot be marshaled](https://github.com/prometheus/prometheus/blob/main/discovery/registry.go#L244). This currently causes Grafana Enterprise Metrics test failures in PRs to re-vendor Mimir.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/service_discovery.go
+++ b/pkg/ruler/service_discovery.go
@@ -22,6 +22,10 @@ const (
 	mechanismName = "dns_sd"
 )
 
+func init() {
+	discovery.RegisterConfig(dnsServiceDiscovery{})
+}
+
 type dnsServiceDiscovery struct {
 	refreshMetrics discovery.RefreshMetricsInstantiator
 


pkg/ruler/service_discovery.go
```diff
@@ -22,6 +22,10 @@ const (
 	mechanismName = "dns_sd"
 )
 
+func init() {
+	discovery.RegisterConfig(dnsServiceDiscovery{})
```
fionaliao, line 26:
```suggestion
	discovery.RegisterConfig(&dnsServiceDiscovery{})
```

I think this should be a pointer to the struct as per the other sd structs registered in prometheus

aknuds1: I don't think it's necessary since `dnsServiceDiscovery`'s methods don't take a pointer. If a pointer were necessary, it wouldn't have compiled.

fionaliao (approved):

colega (approved):

PR #10049, opened 2024-11-29
MQE: don't create copies of native histogram during binary operations

charleskorn:
#### What this PR does

This PR removes creating copies of native histogram during binary operations in MQE.

This dramatically improves the performance of binary operations involving native histograms, for example:

```
                                                                   │ original.txt  │              after.txt              │
                                                                   │    sec/op     │    sec/op     vs base               │
Query/nh_1_/_a_1,_instant_query/engine=Mimir-10                        325.0µ ± 5%   330.7µ ±  4%        ~ (p=0.310 n=6)
Query/nh_1_/_a_1,_range_query_with_100_steps/engine=Mimir-10           396.0µ ± 3%   380.3µ ±  2%   -3.96% (p=0.004 n=6)
Query/nh_1_/_a_1,_range_query_with_1000_steps/engine=Mimir-10         1155.5µ ± 3%   967.7µ ±  2%  -16.26% (p=0.002 n=6)
Query/nh_100_/_a_100,_instant_query/engine=Mimir-10                    5.499m ± 2%   5.426m ±  3%        ~ (p=0.394 n=6)
Query/nh_100_/_a_100,_range_query_with_100_steps/engine=Mimir-10       11.92m ± 5%   10.69m ±  4%  -10.36% (p=0.002 n=6)
Query/nh_100_/_a_100,_range_query_with_1000_steps/engine=Mimir-10      62.75m ± 1%   53.45m ±  1%  -14.81% (p=0.002 n=6)
Query/nh_2000_/_a_2000,_instant_query/engine=Mimir-10                  98.82m ± 3%   98.13m ±  2%        ~ (p=0.485 n=6)
Query/nh_2000_/_a_2000,_range_query_with_100_steps/engine=Mimir-10     205.1m ± 5%   179.9m ±  2%  -12.27% (p=0.002 n=6)
Query/nh_2000_/_a_2000,_range_query_with_1000_steps/engine=Mimir-10     1.169 ± 4%    1.058 ±  5%   -9.47% (p=0.002 n=6)
Query/nh_1_/_2,_instant_query/engine=Mimir-10                          186.3µ ± 2%   192.7µ ±  6%        ~ (p=0.240 n=6)
Query/nh_1_/_2,_range_query_with_100_steps/engine=Mimir-10             253.6µ ± 2%   232.6µ ±  4%   -8.28% (p=0.002 n=6)
Query/nh_1_/_2,_range_query_with_1000_steps/engine=Mimir-10            910.2µ ± 6%   746.1µ ±  2%  -18.03% (p=0.002 n=6)
Query/nh_100_/_2,_instant_query/engine=Mimir-10                        4.667m ± 1%   4.633m ±  1%        ~ (p=0.093 n=6)
Query/nh_100_/_2,_range_query_with_100_steps/engine=Mimir-10          10.354m ± 2%   8.979m ±  3%  -13.28% (p=0.002 n=6)
Query/nh_100_/_2,_range_query_with_1000_steps/engine=Mimir-10          57.01m ± 1%   48.25m ±  2%  -15.36% (p=0.002 n=6)
Query/nh_2000_/_2,_instant_query/engine=Mimir-10                       85.64m ± 1%   86.18m ±  2%        ~ (p=0.394 n=6)
Query/nh_2000_/_2,_range_query_with_100_steps/engine=Mimir-10          180.4m ± 1%   157.6m ±  1%  -12.62% (p=0.002 n=6)
Query/nh_2000_/_2,_range_query_with_1000_steps/engine=Mimir-10        1060.0m ± 2%   853.4m ±  1%  -19.49% (p=0.002 n=6)
```

This has no noticeable impact on binary operations only on floats, and has no noticeable impact on peak memory consumption.

I also took this as an opportunity to move the implementation of the various binary operations out of `one_to_one_vector_vector_binary_operation.go`. Because of this, I'd suggest reviewing each commit separately.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


charleskorn:
This is ready for a review, but will require changes after https://github.com/grafana/mimir/pull/10119 is merged.

pkg/streamingpromql/operators/binops/vector_vector_binary_operation.go
```diff
@@ -671,7 +682,7 @@ var arithmeticAndComparisonOperationFuncs = map[parser.ItemType]binaryOperationF
 		}
 
 		if hlhs != nil && hrhs != nil {
-			res, err := hlhs.Copy().Add(hrhs)
+			res, err := hlhs.Add(hrhs)
```
jhesketh, line 685: The FIXME a few lines above can be removed

jhesketh (commented):
lgtm, but will approve once reworked with the group changes.

I haven't looked closely, but do we have good tests to that would fail if we weren't nil'ing values before returning to the pool appropriately? (Just thinking how we can better test this).

I also imagine this would not work if we every revisited trying not to copy histograms on lookbacks. It might be worth adding a comment somewhere as such.

pkg/streamingpromql/types/data.go
```diff
@@ -47,9 +47,9 @@ func (i *InstantVectorSeriesDataIterator) Reset(data InstantVectorSeriesData) {
 // It returns the next point with the lowest timestamp.
 // If h is not nil, the value is a histogram, otherwise it is a float.
 // If no more values exist ok is false.
-func (i *InstantVectorSeriesDataIterator) Next() (t int64, f float64, h *histogram.FloatHistogram, ok bool) {
+func (i *InstantVectorSeriesDataIterator) Next() (t int64, f float64, h *histogram.FloatHistogram, hIndex int, ok bool) {
```
jhesketh, line 50: What do you think about keeping this method as just returning the next values.

Then have a couple of extra methods to `getCurrentHIndex` / `getCurrentFIndex` ?

I feel like that would be neater and keep the methods clearer on their uses.

charleskorn: The only places that use `Next` also need the histogram sample index, so adding an extra method for this seems unnecessary and possibly error prone (eg. imagine something called `Next` but used the previous value from `getCurrentHIndex`).

charleskorn:
> I haven't looked closely, but do we have good tests to that would fail if we weren't nil'ing values before returning to the pool appropriately? (Just thinking how we can better test this).

Yep, the mangling of native histograms when slices of `HPoint` are returned to the pool catches any cases of this. (The tests I added in this PR were all failing before I added the `nil`-ing, as well as the equivalent cases for one-to-one matching and vector/scalar cases as well.)

> I also imagine this would not work if we every revisited trying not to copy histograms on lookbacks. It might be worth adding a comment somewhere as such.

I'm not sure we'd ever go back to this, and a bunch of stuff will now break if we ever do this - the mixed metrics gauntlet you added should catch these, but if we ever revisit this we're going to have to check a lot of different places carefully. So I'm not sure it's worth adding a comment.

pkg/streamingpromql/types/data.go
jhesketh: Fair enough :+1:

jhesketh, line 50: We should describe the hIndex in the docstring

jhesketh (commented):
lgtm, thanks

pkg/streamingpromql/types/data.go
charleskorn: Done: d6eaad2

jhesketh (approved):

PR #10051, opened 2024-11-29
[r314] Update mimir-prometheus to include prometheus#15426

colega:
Backports https://github.com/prometheus/prometheus/pull/15426 to r314

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -282,7 +282,7 @@ require (
 )
 
 // Using a fork of Prometheus with Mimir-specific changes.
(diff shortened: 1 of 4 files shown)


NickAnge (approved):
LGTM

Just curious: why don't we also bring the MemPostings fields documentation?

colega:
This brings changes from mimir-prometheus which were added and deployed (in GEM) before I created the Prometheus PR and addressed the docs feedback there. (This also doesn't bring the latest SymbolTable changes from there).

PR #10053, opened 2024-11-29
OTLP: Handle StartTimeUnixNano with 'quiet zero' NaNs

jesusvazquez:
This PR updates mimir-prometheus dependency which brings the changes from:
- https://github.com/grafana/mimir-prometheus/pull/766


colega (commented):
Can you edit the title to explain what are we bringing here?

Does this require a changelog entry? Is it a user-facing change?

jesusvazquez:
Title updated. 

This is a change that keeps previous existing behaviour of adding zero's on otlp traffic based on startimeunixnano's value but avoids duplicate values for timestamp on subsequent samples. The flag that enables this is experimental at the moment `otel_created_timestamp_zero_ingestion_enabled`. Would you still add a note on the changelog?

colega (approved):

PR #10081, opened 2024-12-02
query-tee: add per-backend request headers

dimitarvdimitrov:
#### What this PR does

One of my backends needs a specific header so that it returns "correct" results.

Currently we don't have a way to add per-backend configuration. This PR introduces a new flag which takes JSON and allows to configure each backend separately. It will be used like so

```
-backend.config='{"mimir.com": {"request_headers": "X-Some-Header": ["value1", "value2"]}}'
```

I'm open to suggestions in case this is not a route we want to go on.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -206,6 +206,11 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
 		return nil, errors.New("at least 1 backend is required")
 	}
 
+	err := validateBackendConfig(p.backends, cfg.parsedBackendConfig)
+	if err != nil {
+		return nil, fmt.Errorf("validating external backend configs: %w", err)
+	}
+
 	// If the preferred backend is configured, then it must exist among the actual backends.
 	if cfg.PreferredBackend != "" {
 		exists := false
@@ -231,6 +236,22 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
 	}
 
 	return p, nil
+}
+
+func validateBackendConfig(backends []ProxyBackendInterface, config map[string]*BackendConfig) error {
+	for configuredBackend := range config {
+		backendExists := false
+		for _, actualBacked := range backends {
+			if actualBacked.Name() == configuredBackend {
+				backendExists = true
+				break
+			}
+		}
+		if !backendExists {
+			return fmt.Errorf("configured backend %s does not exist in the list of actual backends", configuredBackend)
+		}
+	}
+	return nil
 }
 
 func (p *Proxy) Start() error {
--- a/tools/querytee/proxy.go
+++ b/tools/querytee/proxy.go
@@ -23,6 +23,7 @@ import (
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/server"
 	"github.com/grafana/dskit/spanlogger"
+	"github.com/grafana/regexp"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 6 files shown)


charleskorn:
What if we used a YAML config file instead? Putting JSON into a CLI flag feels a little awkward.

dimitarvdimitrov:
> What if we used a YAML config file instead? Putting JSON into a CLI flag feels a little awkward.

that will make setting this up a bit hard. I'm thinking about the additional jsonnet to create a config map and mount it. But I'm not opposed, so I'll draft that

dimitarvdimitrov:
> > What if we used a YAML config file instead? Putting JSON into a CLI flag feels a little awkward.
> 
> that will make setting this up a bit hard. I'm thinking about the additional jsonnet to create a config map and mount it. But I'm not opposed, so I'll draft that


see 3ce31d1637717e9889e4a30644ab1effa6a01b8f

tools/querytee/proxy.go
```diff
@@ -153,7 +187,15 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
 			preferred = preferredIdx == idx
 		}
 
-		p.backends = append(p.backends, NewProxyBackend(name, u, cfg.BackendReadTimeout, preferred, cfg.BackendSkipTLSVerify))
+		backendCfg := cfg.parsedBackendConfig[name]
```
charleskorn, line 190: It'd also be good to check that the config file doesn't contain config for backends that haven't been configured - this would help catch typos in the config file.

tools/querytee/proxy.go
```diff
@@ -153,7 +187,15 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
...
+		if backendCfg == nil {
+			backendCfg = cfg.parsedBackendConfig[strconv.Itoa(idx)]
```
charleskorn, line 192: Why are we looking up the config by index as well as by name?

charleskorn:
Something for a future PR, not this one: would be good to move some of the config to the config file as well (eg. configuring the proportion of requests sent to a secondary backend per backend, rather than for all secondary backends).

dimitarvdimitrov:
> configuring the proportion of requests sent to a secondary backend per backend, rather than for all secondary backends

I thought about this and I'm not sure if it's compatible with results comparisons: we need the same sampling rate for all backends and the primary in order to compare results. I think it doesn't work for us.

Let's keep discussing though. I'll merge this PR so we have something in the weekly next week

tools/querytee/proxy.go
dimitarvdimitrov: so we can test. we do it for the primary selector too

tools/querytee/proxy.go
dimitarvdimitrov: done in https://github.com/grafana/mimir/pull/10081/commits/9533198bf1362dd152128694ad9d7241f70ef31b

tools/querytee/proxy.go
```diff
@@ -188,6 +238,22 @@ func NewProxy(cfg ProxyConfig, logger log.Logger, routes []Route, registerer pro
 	return p, nil
 }
 
+func validateBackendConfig(backends []ProxyBackendInterface, config map[string]*BackendConfig) error {
+	for configuredBackend := range config {
+		backendExists := false
+		for _, actualBacked := range backends {
```
zenador, line 244: How many backends do we normally expect to have? Should we convert `backends []ProxyBackendInterface` into a map first so we only loop through both `backends` and `config` once each instead of having a nested for loop, or is it expected to be small enough not to matter?

tools/querytee/proxy_backend_test.go
```diff
@@ -84,7 +84,7 @@ func Test_ProxyBackend_createBackendRequest_HTTPBasicAuthentication(t *testing.T
 				orig.Header.Set("X-Scope-OrgID", testData.clientTenant)
 			}
 
-			b := NewProxyBackend("test", u, time.Second, false, false)
+			b := NewProxyBackend("test", u, time.Second, false, false, BackendConfig{})
```
zenador, line 87: [nit] Do you think we should have a `defaultBackendConfig()` instead (we can set it to return `BackendConfig{}` for now) or is it unlikely that we'll ever need to set some values as default?

zenador (approved):
LGTM in general, approving first to unblock since we want to merge this soon and Charles has already reviewed it

tools/querytee/proxy.go
dimitarvdimitrov: this only happens once on startup and backend are usually below 10. I'm not conned about perf here

PR #10083, opened 2024-12-02
Add UsageTracker gRPC service definition

pracucci:
#### What this PR does

Add UsageTracker gRPC service.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/usagetracker/usagetrackerpb/usagetracker.proto
```diff
@@ -0,0 +1,29 @@
...
+service UsageTracker {
+  rpc TrackSeries(TrackSeriesRequest) returns (TrackSeriesResponse);
+}
+
+message TrackSeriesRequest {
+  // The tenant owning the series.
+  string userID = 1;
+
+  // The hashes of the series to track.
+  repeated int64 seriesHashes = 2;
```
colega, line 23: Can this be `uint64`?

colega, line 23: Looks good as a first approach, but given the horrible preformance of grpc unmarshaling of slices (it doesn't build the slice of a correct length), we may want to switch to just a []byte and read that from the service implementation in a streamed fashion.

That can be done in a future version though.

pracucci: I agree. Maybe we can keep this for a production version, rather in the prototype. Looks something we know how to optimize.

pracucci: Done in https://github.com/grafana/mimir/pull/10083/commits/98e96f4fb4a47b03beab5e7ff7111e9cb02e9104

colega (approved):

PR #10084, opened 2024-12-02
query-frontend: propagate headers on labels requests

dimitarvdimitrov:
#### What this PR does

We try to copy the headers from the internal request representation, but we never transfer it from the HTTP request to that representation  in the first place. This PR fixes that. 

https://github.com/grafana/mimir/blob/97f031939ce511e2d93a6e80634d9c5d9a5d8f90/pkg/frontend/querymiddleware/codec.go#L731-L741

Technically the list of headers to propagate is empty by default in Mimir, so there wasn't a bug for Mimir users, hence, not adding a changelog entry.

https://github.com/grafana/mimir/blob/451d592d6cd65171da4741564745ef105c1c475e/pkg/frontend/querymiddleware/codec.go#L53

> __Note to reviewers__: reviewing without whitespace is easier

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/codec.go
+++ b/pkg/frontend/querymiddleware/codec.go
@@ -328,12 +328,6 @@ func (c prometheusCodec) decodeRangeQueryRequest(r *http.Request) (MetricsQueryR
 		return nil, apierror.New(apierror.TypeBadData, err.Error())
 	}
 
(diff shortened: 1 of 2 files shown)


pracucci (approved):
LGTM!

PR #10088, opened 2024-12-02
Add UsageTrackerClient to track series

pracucci:
#### What this PR does

Add UsageTrackerClient to track series. It will be used by distributor in a follow up PR.

UsageTrackerClient features:
- Preferred zone
- Hedge to other zone instances after 100ms (configurable) or on failure

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/usagetracker/usagetrackerclient/client.go
+++ b/pkg/usagetracker/usagetrackerclient/client.go
@@ -1,0 +1,232 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package usagetrackerclient
+
+import (
+	"context"
+	"flag"
+	"math/rand/v2"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/concurrency"
+	"github.com/grafana/dskit/crypto/tls"
+	"github.com/grafana/dskit/ring"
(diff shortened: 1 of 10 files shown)


pkg/usagetracker/usagetrackerclient/client.go
```diff
@@ -0,0 +1,232 @@
...
+	)
+
+	// Series hashes are 64bit but the hash ring tokens are 32bit, so we truncate
+	// hashes to 32bit to get keys to lookup in the ring.
+	keys := make([]uint32, len(series))
+	for i, hash := range series {
+		keys[i] = uint32(hash)
+	}
+
+	err := ring.DoBatchWithOptions(ctx, ring.WriteNoExtend, c.partitionBatchRing, keys,
```
colega, line 111: Should this be this?
```suggestion
	err := ring.DoBatchWithOptions(ctx, trackSeriesOp, c.partitionBatchRing, keys,
```
(It's the same value, but not sure if this is semantically more correct)

pracucci: >Should this be this?

Yes, definitely. It was a mistake using `ring.WriteNoExtend` here.

colega (approved):

PR #7026, opened 2024-01-02
query-frontend: remove duplicate defaulting of instant query `time` param

dimitarvdimitrov:
After #6985 the default value of the time parameter (i.e. time.Now()) is set in the PrometheusCodec. This PR removes the default set in an HTTP middleware.

With this change the time parameter in the HTTP request is no longer populated if the client didn't provide it. Instead, it will only be populated in the decoded `querymiddleware.Request`. The HTTP parameter was referenced in a few places
* [DecodeInstantQueryTimeParams](https://github.com/grafana/mimir/blob/9fd2b19a61acec59336d677a50270806dcc4482c/pkg/frontend/querymiddleware/codec.go#L295) which now also falls back to the default (after #6985). This is also used for partial queries, but those shouldn't observe an unset parameter (since the parent query had the parameter defaulted to `time.Now()`).
* invocations of [ParseTimeParam](https://github.com/grafana/mimir/blob/bcb5c345e03a7c8d0ff0a2e0985be448a332dd7d/pkg/util/time.go#L40). One of them is in DecodeInstantQueryTimeParams, the other two are for label values requests, which have their own logic for defaulting the parameters.

PR #5524, opened 2023-07-17
Fix cardinality and label names/values requests handling on POST requests

pracucci:
#### What this PR does
We found a bug introduced in #5212 and #5426 which caused POST requests returning wrong results. The problem is that, when the cardinality and/or label names/values result cache is enabled, the caching middleware consumes the `http.Request` `Body` and so the body will be empty when later read to forward it to the query-scheduler or querier.

In this PR I propose a fix. The fix consists in consuming the Body and then replacing it with the original content. It's not a new approach, in fact we were already doing it in `Handler.ServeHTTP()` so I've re-used that code, moving it to a common utility function.

Note to reviewers:
- The modified integration and unit tests fail with the previous buggy version

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [x] Tests updated
- [ ] Documentation added
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #3599, opened 2022-12-01
Add default "time" param to parsed form (if any)

colega:
#### What this PR does

If the form was already parsed by a downstream middleware, we need to add it to that form too, otherwise it won't be available to upstream by just adding it to the url query.

This implement's @pstibrany's suggestion https://github.com/grafana/mimir/pull/3588#discussion_r1036930450 and copies all the tests from that PR.

#### Which issue(s) this PR fixes or relates to

Will make https://github.com/grafana/mimir/pull/3561 easier to review.

#### Checklist

- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #10093, opened 2024-12-02
Early Request Validation in Query Frontend

francoposa:
#### What this PR does

This PR introduces early request validation for the majority of request types supported by the query-frontend.
This specifically does not introduce any new request parsing or validation logic, but it ensures that the validation that does exist is called in the first roundtripper, before any other work is done _and_ that a standardized 400 Bad Request APIError is returned.

The tests are largely focused on making sure we get a status 400 API Error when we expect one - each request validator roundtripper relies on existing validation codecs which have their own extensive test suites.

**Done so far** are the ones that have clearly defined validation code already in Mimir:
* instant query
* range query
* label names query
* label values query
* series query
* label names cardinality
* label values cardinality
* active series cardinality

Still to be done, likely saved for a subsequent PR are:
* metadata
* query exemplars
* active native histogram metrics cardinality requests

In previous incidents we have seen unexpected issues when parsing an invalid requests errors out in a part of the stack which is not set up to  propagate the errors properly back up the stack and prevent breaking connections or crashing goroutines.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/roundtrip.go
+++ b/pkg/frontend/querymiddleware/roundtrip.go
@@ -293,11 +293,17 @@ func newQueryTripperware(
 			next = newReadConsistencyRoundTripper(next, ingestStorageTopicOffsetsReader, limits, log, metrics)
 		}
 
-		// Look up cache as first thing.
+		// Look up cache as first thing after validation.
 		if cfg.CacheResults {
 			cardinality = newCardinalityQueryCacheRoundTripper(c, cacheKeyGenerator, limits, cardinality, log, registerer)
 			labels = newLabelsQueryCacheRoundTripper(c, cacheKeyGenerator, limits, labels, log, registerer)
 		}
+
+		// Validate the request before any processing.
+		queryrange = NewMetricsQueryRequestValidationRoundTripper(codec, queryrange)
+		instant = NewMetricsQueryRequestValidationRoundTripper(codec, instant)
+
+		labels = NewLabelsQueryRequestValidationRoundTripper(codec, labels)
 
 		return RoundTripFunc(func(r *http.Request) (*http.Response, error) {
 			switch {
--- a/pkg/frontend/querymiddleware/roundtrip.go
+++ b/pkg/frontend/querymiddleware/roundtrip.go
@@ -273,6 +273,7 @@ func newQueryTripperware(
 		activeSeries := next
 		activeNativeHistogramMetrics := next
 		labels := next
+		series := next
 
 		if cfg.ShardActiveSeriesQueries {
 			activeSeries = newShardActiveSeriesMiddleware(activeSeries, cfg.UseActiveSeriesDecoder, limits, log)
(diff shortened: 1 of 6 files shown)


pkg/frontend/querymiddleware/roundtrip_test.go
```diff
@@ -186,26 +186,6 @@ func TestTripperware_InstantQuery(t *testing.T) {
 		}, res)
 	})
 
-	t.Run("specific time param with form being already parsed", func(t *testing.T) {
```
francoposa, line 189: these tests are for [old code](https://github.com/grafana/mimir/pull/3599) that was [removed](https://github.com/grafana/mimir/pull/7026) as it is more robustly handled in the prometheus codec and associated utils.

The tests continued to pass by chance that we weren't using this particular request parsing middleware twice.

We have moved away from trying to go back and fix the situation if the request body is consumed by our own middleware. Instead everything that reads a request body just needs to use `ParseRequestFormWithoutConsumingBody` introduced in https://github.com/grafana/mimir/pull/5524.

francoposa:
at this point I have added roundtrippers which call on the existing validation code for:
* instant query
* range query
* label names query
* label values query
* series query
* label names cardinality
* label values cardinality
* active series cardinality

this leaves only metadata, query exemplars, and active native histogram metrics requests still TODO for adding validation at the top of the roundtripper stack

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -289,16 +290,24 @@ func newQueryTripperware(
...
 		if cfg.CacheResults {
 			cardinality = newCardinalityQueryCacheRoundTripper(c, cacheKeyGenerator, limits, cardinality, log, registerer)
 			labels = newLabelsQueryCacheRoundTripper(c, cacheKeyGenerator, limits, labels, log, registerer)
 		}
 
+		// Validate the request before any processing.
+		queryrange = NewMetricsQueryRequestValidationRoundTripper(codec, queryrange)
+		instant = NewMetricsQueryRequestValidationRoundTripper(codec, instant)
+		labels = NewLabelsQueryRequestValidationRoundTripper(codec, labels)
+		series = NewLabelsQueryRequestValidationRoundTripper(codec, series)
```
francoposa, line 308: Series Request validation is also supported by the codec for Label Names and Label Values Requests - a separate PR exists to update the naming for the codecs to express this

56quarters (approved):
_You are in a maze of handlers, middlewares, roundtrippers, and tripperwares, all alike._

This change LGTM and is a definite improvement.

However, it feels weird that we're parsing/decoding requests multiple times. There's some existing functionality that adds `QueryDetails` to the context of a request. WDYT about doing something similar, adding the parsed request from the validation roundtrippers to the context of each request so that every subsequent roundtripper doesn't need to repeat the parsing process in a follow-up PR? It's fine if the answer to that is "that'd just make things more complicated"

francoposa:
> However, it feels weird that we're parsing/decoding requests multiple times. There's some existing functionality that adds `QueryDetails` to the context of a request. WDYT about doing something similar, adding the parsed request from the validation roundtrippers to the context of each request so that every subsequent roundtripper doesn't need to repeat the parsing process in a follow-up PR? It's fine if the answer to that is "that'd just make things more complicated"

@56quarters I will scratch that out and see how it looks - the multiple parse is definitely something we want to get away from and this context approach could be a good band-aid in the mean time.
Getting the parsing done early in the roundtripper stack like this PR does is one step towards getting away from multiple parse.

There's still a bit of work to get to a point where we can have roundtrippers that can take a `MetricsQueryRequest` or `LabelsSeriesQueryRequest` as an argument.
Right now the roundtrippers that handle several request types they do so by taking `http.Request` and switch on the request type by matching on URL patterns, meaning they would not support getting passed only a `MetricsQueryRequest` or `LabelsSeriesQueryRequest` directly in the signature.
We'd either need to apply a generics or split the signatures so there's different roundtripper paths for each parsed request type, which requires some level of accounting for which aspects of a request each roundtripper needs access to.

PR #10095, opened 2024-12-03
tracing: finish OTEL bridge span only once

dimitarvdimitrov:
#### Problem

The OTEL spec allows to close the span multiple times and that `End` should be idempotent. I think the jaeger implementation doesn't have the same guarantees, so we have to handle this in out bridge implementation.

Currently, we allow to call `Finish` multiple times and we end up reporting the span multiple times: once for exhausting the HTTP request body with `io.EOF` and once for `Close()`. For more details see https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6391

#### The OTEL spec

https://github.com/open-telemetry/opentelemetry-specification/blob/50027a1036746dce293ee0a8592639f131fc1fb8/specification/trace/api.md#end

> Implementations SHOULD ignore all subsequent calls to End and any other Span methods, i.e. the Span becomes non-recording by being ended (there might be exceptions when Tracer is streaming events and has no mutable state associated with the Span).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/tracing/tracing_test.go
+++ b/pkg/util/tracing/tracing_test.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"math/rand"
 	"testing"
+	"time"
 
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/log"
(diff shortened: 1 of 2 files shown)


pkg/util/tracing/tracing_test.go
```diff
@@ -235,3 +236,66 @@ func TestOpenTelemetryTracerBridge_Start(t *testing.T) {
...
+		t.Run(testName, func(t *testing.T) {
+			// Setup mock
+			m := &TracingSpanMock{}
+			m.On("Finish").Return()
+			m.On("FinishWithOptions", mock.Anything).Return()
+
+			// Create bridge
+			s := NewOpenTelemetrySpanBridge(m, nil)
+
+			// Call End multiple times
```
aknuds1, line 283: [Nit] For clarity:
```suggestion
			// Call End potentially multiple times
```

pkg/util/tracing/tracing_test.go
```diff
@@ -235,3 +236,66 @@ func TestOpenTelemetryTracerBridge_Start(t *testing.T) {
...
+				}
+			}
+
+			// Assert expectations
+			if testData.finishCalls > 0 {
+				m.AssertNumberOfCalls(t, "Finish", testData.finishCalls)
+			}
+			if testData.finishOpts > 0 {
+				m.AssertNumberOfCalls(t, "FinishWithOptions", testData.finishOpts)
+			}
```
aknuds1, line 298: Isn't it more sensible to also assert when 0 calls should be made?
```suggestion
			m.AssertNumberOfCalls(t, "Finish", testData.finishCalls)
			m.AssertNumberOfCalls(t, "FinishWithOptions", testData.finishOpts)
```

aknuds1 (commented):
Please see comments.

pkg/util/tracing/tracing_test.go
dimitarvdimitrov: yeah that makes sense 👍

aknuds1 (approved):

ying-jeanne (approved):
thanks for fixing it, lgtm.

PR #10096, opened 2024-12-03
Alerting: add `SentBy` information to email footer

titolins:
- The remote AM is not adding the build version to the email settings as it should.
- https://github.com/grafana/mimir/pull/9987 was doing that, but the result is misleading since in that case the notification is not really coming from Grafana
- We've added a new configuration parameter to the email sender in https://github.com/grafana/alerting/pull/258
- Now we should update the alerting library and use `SentBy` appropriately

Fixes https://github.com/grafana/alerting-squad/issues/873


santihernandezc (approved):
LGTM, tested it locally using Mailhog.
Please move the SNS changes to a separate PR 🙏

titolins:
> Please move the SNS changes to a separate PR 🙏

Done 👍

aldernero (approved):
LGTM

PR #10101, opened 2024-12-03
ingest: manually create Kafka topic instead of configuring num.partitions

dimitarvdimitrov:
### What this PR does

This PR does two things:

#### 1. Manually creates the Kafka topic instead of relying on kgo autocreation

Here’s why kgo autocreation doesn’t work for us.
* autocreation can only happen on metadata requests (for example `kgo.Client.ForceMetadataRefresh()` and `kadm.Client.Metadata()` ). There is a field on the request called `AllowAutoTopicCreation`
* kgo would set `AllowAutoTopicCreation` in the request only when doing `kgo.Client.ForceMetadataRefresh`, but not when doing `kadm.Client.Metadata`
* the field makes sense only if the metadata request contains some list of topics
* the metadata request can only contain a list of topics if the `kgo.Client` has seen them before - either by consuming from them or producing to them

In other words, for the topic to be created we need the distributor to push records or the ingester to consume records. With concurrent fetching we can't consume until we have a topic. The distributor waits for the ingester to become ready (and register itself in the rings) before producing. This created a cycling dependency between the ingester and distributor.

Because of this we now issue an explicit `CreateTopics` request.

**As a result we no longer configure num.partitions in the Kafka cluster**

#### 2. Enables ingestion and fetching concurrency by default in e2e tests

This is one of the last steps before making these the defaults in Mimir.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/util.go
```diff
@@ -94,10 +96,6 @@ func commonKafkaClientOptions(cfg KafkaConfig, metrics *kprom.Metrics, logger lo
 		}),
 	}
 
-	if cfg.AutoCreateTopicEnabled {
-		opts = append(opts, kgo.AllowAutoTopicCreation())
-	}
```
pracucci, line 99: I think removing this cause another subtle change. I think `commonKafkaClientOptions()` is used by the generic `NewKafkaReaderClient()` and `NewKafkaWriterClient()`, so they're used for other topics too (e.g. block-builder). 

In this PR we both remove this code and we also changed the meaning of `auto-create-topic-default-partitions` (which doesn't change the Kafka default anymore, but it's just the number of partitions `createTopic()` will set), which means it will be a breaking change for other clients.

Probably the safest change would be preserving `setDefaultNumberOfPartitionsForAutocreatedTopics()` behaviour, preserving `kgo.AllowAutoTopicCreation()`, but force a topic creation too in our clients.

Alternative, I think a relatively safe change may be move `createTopic()` invokation to `NewKafkaReaderClient()` and `NewKafkaWriterClient()`.

Makefile
```diff
@@ -478,6 +478,12 @@ lint: check-makefiles
 		"google.golang.org/grpc/metadata.{FromIncomingContext}=google.golang.org/grpc/metadata.ValueFromIncomingContext" \
 		./pkg/... ./cmd/... ./integration/...
 
+	# We don't use topic auto-creation because we don't control the num.partitions.
+	# As a result the topic can be created with the wrong number of partitions.
+	faillint -paths \
+		"github.com/twmb/franz-go/pkg/kgo.{AllowAutoTopicCreation}" \
+		./pkg/... ./cmd/... ./integration/...
```
pracucci, line 485: Maybe worth to add tools/ too. kafkatool currently has the config option, but I don't think it should.

pkg/storage/ingest/util.go
pracucci: Discussed offline. Agree with Dimitar's change.

pracucci (approved):
LGTM! Thanks for your patience 🙏

PR #9985, opened 2024-11-22
Query Frontend: Update Server-Timing format

tinitiuset:
#### What this PR does

This PR updates `Server-Timing` format to better comply with the standard described [here](https://www.w3.org/TR/server-timing/).

The standard does not talk about values, only about duration. But I feel like it is a good idea to follow the same pattern if we return values in this header.

#### Which issue(s) this PR fixes or relates to

There's no specific issue, but `bytes_processed` was [introduced recently](https://github.com/grafana/mimir/pull/9645) by @madhu-reddy-peram 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10103, opened 2024-12-03
Query Frontend: Expose samples_processed in Server-Timing header

tinitiuset:
#### What this PR does

This PR will make the Query frontend expose samples_processed in Server-Timing header. It is complementary to #9985 and [this](https://github.com/grafana/dskit/pull/619) in the effort of measuring throughput on Mimir.

This was worked on by @krajorama before #7966, but I decided to open a different PR as the idea has evolved overtime.

Not related to an specific issue.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pracucci (commented):
I have a comment about the naming. "total samples" may mean "total raw samples processed" or "total number of samples returned by the query". The cost of a query is a function of the actual raw samples processed, not returned. If that's what you're going to track (as I would expect) then we may consider renaming it to "samples processed" (to keep it consistent with "bytes processed" that we already have).

pkg/querier/stats_renderer.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"context"
+
+	promql_stats "github.com/prometheus/prometheus/util/stats"
+	prom_api "github.com/prometheus/prometheus/web/api/v1"
+
+	"github.com/grafana/mimir/pkg/querier/stats"
+)
+
+func StatsRenderer(ctx context.Context, s *promql_stats.Statistics, param string) promql_stats.QueryStats {
```
pracucci, line 14: This looks PromQL engine specific. Be aware we're migrating to Mimir Query Engine (MQE) and we need the tracking to work for MQE too. You can talk to @charleskorn about it.

charleskorn: This is something that is not currently supported by MQE.

Is your goal to measure the total number of samples read from ingesters and store-gateways for the query?

pracucci: > Is the goal to measure the total number of samples read from ingesters and store-gateways for the query?

I would say this one.

tinitiuset:
@pracucci 's feedback is acknowledged and MQE support will be added in another PR.

pkg/querier/stats/stats.go
charleskorn, line 1: It'd be good to expand the various tests in `stats_test.go` to cover the new field here.

pkg/querier/stats_renderer_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			q := &mockSampleAndChunkQueryable{
+				queryableFn: func(_, _ int64) (storage.Querier, error) {
+					return &mockQuerier{
+						selectFn: func(_ context.Context, _ bool, _ *storage.SelectHints, _ ...*labels.Matcher) storage.SeriesSet {
+							return series.NewConcreteSeriesSetFromUnsortedSeries(seriesList)
+						},
+					}, nil
+				},
+			}
```
charleskorn, line 68: [nit] Something that could simplify this: instead of creating a mock queryable, use a `promqltest.LoadedStorage` with ten series, and use different expressions for each test case to limit the number of series (and therefore samples) selected.

This would also allow you to move most of the setup code out of the `t.Run`, as you could share the same queryable and API. This would make the `t.Run` more focused on the specific test case.

charleskorn (commented):
LGTM, but I'd like to see the `Stats` tests expanded to cover the new `SamplesProcessed` field.

dimitarvdimitrov (approved):
LGTM, mod Charles and Marco's comments

chencs:

pkg/querier/stats/stats.go
tinitiuset: Thanks for noticing it!

pkg/querier/stats_renderer_test.go
tinitiuset: Thank you for the idea 😄.

tinitiuset:
Branch got out of sync with main. I also updated a test that was making other tests fail because it was enabling UTF-8 support.

pkg/querier/stats/stats.proto
```diff
@@ -36,4 +36,7 @@ message Stats {
   google.protobuf.Duration queue_time = 9 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
   // The time spent at the frontend encoding the query's final results. Does not include time spent serializing results at the querier.
   google.protobuf.Duration encode_time = 10 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
+  // TotalSamples represents the total number of samples scanned while evaluating a query.
+  // This value is taken from the Prometheus engine github.com/prometheus/prometheus/util/stats/query_stats.go
```
dimitarvdimitrov, line 40: comment is a bit inaccurate, right? This will also come from MQE. I don't think you should say where it's coming from - that can change with time and outdate the comment.

pkg/querier/stats_renderer_test.go
```diff
@@ -0,0 +1,141 @@
...
+			expr:                 `start_series{}`,
+			expectedTotalSamples: 6,
+		},
+		"end_series": {
+			expr:                 `end_series{}`,
+			expectedTotalSamples: 6,
+		},
+		"sparse_series": {
+			expr:                 `sparse_series{}`,
+			expectedTotalSamples: 5 + 4,
```
dimitarvdimitrov, line 58: shouldn't this be 2 because there are only two samples

```
			sparse_series 0 _ _ _ _ _ _ 7 _ _ _
```

pkg/querier/stats/stats.proto
tinitiuset: Makes sense, thanks for noticing.

pkg/querier/stats_renderer_test.go
dimitarvdimitrov: i wonder if it's not because of the [lookback period of 5 minutes](https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness). Can you quickly verify that it is? It's configurable on the promQL engine config via `LookbackDelta`

tinitiuset: Good call, I tried different `LookbackDelta` values and it's consistent with the samples returned.

PR #10107, opened 2024-12-03
Distributor: Delay simulation on ingestion

gotjosh:
#### What this PR does

With Sigyn, we have a much higher latency in ingestion - somewhere around 5x. Ths can produce adverse side affects on remote write clients struggling.

Still needs:

- [x] limits test
- [x] possibly an end to end test

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -1401,17 +1401,35 @@ func NextOrCleanup(next PushFunc, pushReq *Request) (_ PushFunc, maybeCleanup fu
 
 // Push is gRPC method registered as client.IngesterServer and distributor.DistributorServer.
 func (d *Distributor) Push(ctx context.Context, req *mimirpb.WriteRequest) (*mimirpb.WriteResponse, error) {
+	start := time.Now()
 	pushReq := NewParsedRequest(req)
 	pushReq.AddCleanup(func() {
 		mimirpb.ReuseSlice(req.Timeseries)
 	})
 
 	pushErr := d.PushWithMiddlewares(ctx, pushReq)
+
+	// TODO(gotjosh): Is this better as a middleware?
+	d.MaybeDelayIngestion(ctx, start)
+
 	if pushErr == nil {
 		return &mimirpb.WriteResponse{}, nil
 	}
 	handledErr := d.handlePushError(ctx, pushErr)
 	return nil, handledErr
+}
+
+func (d *Distributor) MaybeDelayIngestion(ctx context.Context, start time.Time) {
+	// Produce an artificial delay before we respond back to the client if configured.
+	userID, err := tenant.TenantID(ctx) // Log tenant ID if available.
+	if err != nil {
+		level.Warn(d.log).Log("msg", "failed to get tenant ID while trying to delay ingestion", "err", err)
+		return
+	}
+
+	// Target delay - time spent processing the request.
+	// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a non-zero value.
+	time.Sleep(d.limits.DistributorArtificialIngestionDelay(userID) - time.Since(start))
 }
 
 func (d *Distributor) handlePushError(ctx context.Context, pushErr error) error {
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -834,7 +834,7 @@ func (d *Distributor) wrapPushWithMiddlewares(next PushFunc) PushFunc {
 		next = middlewares[ix](next)
 	}
 
-	// The delay middleware must take into account total runtime of all other middlewares and the push func, hence why we wrap all others..
+	// The delay middleware must take into account total runtime of all other middlewares and the push func, hence why we wrap all others.
 	return d.outerMaybeDelayMiddleware(next)
 
 }
@@ -1438,7 +1438,6 @@ func (d *Distributor) Push(ctx context.Context, req *mimirpb.WriteRequest) (*mim
 	})
 
 	pushErr := d.PushWithMiddlewares(ctx, pushReq)
-
 	if pushErr == nil {
 		return &mimirpb.WriteResponse{}, nil
 	}
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -182,7 +182,14 @@ type Distributor struct {
 
 	// partitionsRing is the hash ring holding ingester partitions. It's used when ingest storage is enabled.
 	partitionsRing *ring.PartitionInstanceRing
+
+	// For testing.
+	sleep func(time.Duration)
+	now   func() time.Time
 }
+
+func defaultSleep(d time.Duration) { time.Sleep(d) }
+func defaultNow() time.Time        { return time.Now() }
 
 // OTelResourceAttributePromotionConfig contains methods for configuring OTel resource attribute promotion.
 type OTelResourceAttributePromotionConfig interface {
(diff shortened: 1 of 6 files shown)


pkg/util/validation/limits.go
```diff
@@ -280,6 +281,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.BoolVar(&l.OTelMetricSuffixesEnabled, "distributor.otel-metric-suffixes-enabled", false, "Whether to enable automatic suffixes to names of metrics ingested through OTLP.")
 	f.BoolVar(&l.OTelCreatedTimestampZeroIngestionEnabled, "distributor.otel-created-timestamp-zero-ingestion-enabled", false, "Whether to enable translation of OTel start timestamps to Prometheus zero samples in the OTLP endpoint.")
 	f.Var(&l.PromoteOTelResourceAttributes, "distributor.otel-promote-resource-attributes", "Optionally specify OTel resource attributes to promote to labels.")
+	f.Var(&l.DistributorArtificialIngestionDelay, "distributor.artificial-ingestion-latency", "Target ingestion delay. If set to a non-zero value, the distributor will artificially delay ingestion time-frame by the specified duration by computing the difference between actual ingestion and the target. There is no delay on actual ingestion of samples, it is only the response back to the client.")
```
56quarters, line 284: Can you rename this to remove the "Distributor" prefix of the configuration setting and variable? It's already named to indicate it's part of the ingestion path and we don't do this with other settings.

pkg/distributor/distributor.go
```diff
@@ -1401,19 +1401,37 @@ func NextOrCleanup(next PushFunc, pushReq *Request) (_ PushFunc, maybeCleanup fu
...
+func (d *Distributor) MaybeDelayIngestion(ctx context.Context, start time.Time) {
+	// Produce an artificial delay before we respond back to the client if configured.
+	userID, err := tenant.TenantID(ctx) // Log tenant ID if available.
+	if err != nil {
+		level.Warn(d.log).Log("msg", "failed to get tenant ID while trying to delay ingestion", "err", err)
+		return
+	}
+
+	// Target delay - time spent processing the request.
+	// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a non-zero value.
```
56quarters, line 1431:
```suggestion
	// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a negative value.
```

pkg/distributor/distributor.go
```diff
@@ -1401,19 +1401,37 @@ func NextOrCleanup(next PushFunc, pushReq *Request) (_ PushFunc, maybeCleanup fu
...
 func (d *Distributor) Push(ctx context.Context, req *mimirpb.WriteRequest) (*mimirpb.WriteResponse, error) {
+	start := time.Now()
 	pushReq := NewParsedRequest(req)
 	pushReq.AddCleanup(func() {
 		mimirpb.ReuseSlice(req.Timeseries)
 	})
 
 	pushErr := d.PushWithMiddlewares(ctx, pushReq)
+
+	// TODO(gotjosh): Is this better as a middleware?
```
56quarters, line 1412: I think this is probably correct that it would be better as a middleware.

pkg/distributor/distributor_test.go
```diff
@@ -8004,3 +8004,59 @@ func TestCheckStartedMiddleware(t *testing.T) {
...
+			}
+
+			// Measure the time taken by MaybeDelayIngestion
+			start := time.Now()
+			ctx := user.InjectOrgID(context.Background(), tc.userID)
+			distributor.MaybeDelayIngestion(ctx, tc.startTime)
+			elapsed := time.Since(start)
+
+			// Assert the sleep time is as expected (allowing a small margin for runtime delays).
+			require.InDelta(t, tc.expectedSleep, elapsed, float64(100*time.Millisecond))
```
56quarters, line 8059: I really really don't want to add more unit tests that depend on timing. Can we introduce a mock clock and assert it's called with the correct duration?

pkg/util/validation/limits.go
gotjosh: Are you certain? I've introduced a few recently in the ruler and they all have them - I thought it was more of a "introduce it in the new ones so that we maybe one day reach consistency". We have so many flags in this struct that I think the prefix is beneficial.

pkg/distributor/distributor.go
gotjosh: This was not up for review yet - it the end the middleware was the way to go for other reasons that I realise after testing 😄

pkg/distributor/distributor.go
```diff
@@ -1192,6 +1203,26 @@ func (d *Distributor) metricsMiddleware(next PushFunc) PushFunc {
...
+	return func(ctx context.Context, pushReq *Request) error {
+		start := d.now()
+		// Execute the whole middleware chain.
+		err := next(ctx, pushReq)
+
+		userID, userErr := tenant.TenantID(ctx) // Log tenant ID if available.
+		if userErr == nil {
+			// Target delay - time spent processing the middleware chain including the push.
+			// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a negative value.
+			d.sleep(d.limits.DistributorArtificialIngestionDelay(userID) - d.now().Sub(start))
```
pracucci, line 1217: I would recommend applying a jitter to be more realistic. You can do it like this:

```
if delay := d.limits.DistributorArtificialIngestionDelay(userID) - d.now().Sub(start); delay > 0 {
  d.sleep(util.DurationWithJitter(delay, 0.1))
}
```

The code above applies a 10% jitter (both positive and negative). It must take in input positive values, reason why I'm checking delay > 0.

pkg/util/validation/limits.go
pracucci: I don't feel strong on this, but an option to follow nick's suggestion is to rename the field to `ingestion_artificial_delay` (and CLI to `-distributor.ingestion-artificial-delay`). It would be consistent with `IngestionTenantShardSize`.

pracucci (approved):
Nice work!

pkg/util/validation/limits.go
56quarters: I prefer Marco's suggestion to use `ingestion_artificial_delay`. I don't like including "distributor" in the setting name since it's an implementation detail.

pkg/distributor/distributor.go
```diff
@@ -437,6 +444,8 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		}),
 
 		PushMetrics: newPushMetrics(reg),
+		now:         defaultNow,
```
56quarters, line 447: Style comment so feel free to ignore, instead of defining default functions above you could just set them here. e.g. 
```suggestion
		sleep:       time.Sleep,
		now:         time.Now,
```

pkg/distributor/distributor.go
```diff
@@ -182,8 +182,15 @@ type Distributor struct {
 
 	// partitionsRing is the hash ring holding ingester partitions. It's used when ingest storage is enabled.
 	partitionsRing *ring.PartitionInstanceRing
+
+	// For testing.
```
56quarters, line 186: Can you expand this comment a little bit? e.g.
```suggestion
	// For testing functionality that relies on timing without having to sleep in unit tests.
```

pkg/distributor/distributor.go
gotjosh: Done in https://github.com/grafana/mimir/pull/10107/commits/945dfb844564683d41a87fd4f290e626007805cd

pkg/util/validation/limits.go
gotjosh: Done in https://github.com/grafana/mimir/pull/10107/commits/945dfb844564683d41a87fd4f290e626007805cd

PR #10110, opened 2024-12-03
copyblocks: support copying between tenants

andyasp:
#### What this PR does

Adds the option `user-mapping` to `copyblocks` to allow copying between tenants. Previously it was always assumed that blocks would be copied to a different bucket under the same tenant.

For example, passing `--user-mapping="foo:bar"` would map the blocks from tenant `foo` to the destination tenant `bar`.

Something I did not touch in this PR was the copy marker naming. It is of the form `markers/blockID-copied-target-bucket`. Now that different tenants can be copied to the marker can clash if you copy data from one tenant to multiple tenants in the same destination bucket. That seemed like a weird use case so I deferred to YAGNI.

N/A

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.

### DIFF
--- a/tools/copyblocks/README.md
+++ b/tools/copyblocks/README.md
@@ -13,6 +13,7 @@ The currently supported services are Amazon Simple Storage Service (S3 and S3-co
 - Include or exclude users from having blocks copied (`--enabled-users` and `--disabled-users`)
 - Configurable minimum block duration (`--min-block-duration`) and (`--skip-no-compact-block-duration-check`) to target blocks that are not awaiting compaction
 - Configurable time range (`--min-time` and `--max-time`) to only copy blocks inclusively within a provided range
+- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` would copy source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If a mapping is not provided for a tenant it is assumed to be identical to the source tenant.
 - Log what would be copied without actually copying anything with `--dry-run`
 
 ## Running
--- a/tools/copyblocks/README.md
+++ b/tools/copyblocks/README.md
@@ -13,7 +13,7 @@ The currently supported services are Amazon Simple Storage Service (S3 and S3-co
 - Include or exclude users from having blocks copied (`--enabled-users` and `--disabled-users`)
 - Configurable minimum block duration (`--min-block-duration`) and (`--skip-no-compact-block-duration-check`) to target blocks that are not awaiting compaction
 - Configurable time range (`--min-time` and `--max-time`) to only copy blocks inclusively within a provided range
-- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` would map source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If a mapping is not provided for a tenant it is assumed to be identical to the source tenant.
+- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` maps source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If you don't provide a mapping for a tenant, it is assumed to be identical to the source tenant.
 - Log what would be copied without actually copying anything with `--dry-run`
 
 ## Running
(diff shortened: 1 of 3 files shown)


tools/copyblocks/README.md
```diff
@@ -13,6 +13,7 @@ The currently supported services are Amazon Simple Storage Service (S3 and S3-co
 - Include or exclude users from having blocks copied (`--enabled-users` and `--disabled-users`)
 - Configurable minimum block duration (`--min-block-duration`) and (`--skip-no-compact-block-duration-check`) to target blocks that are not awaiting compaction
 - Configurable time range (`--min-time` and `--max-time`) to only copy blocks inclusively within a provided range
+- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` would copy source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If a mapping is not provided for a tenant it is assumed to be identical to the source tenant.
```
andyasp, line 16:
```suggestion
- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` would map source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If a mapping is not provided for a tenant it is assumed to be identical to the source tenant.
```
Changing the wording here from `copy` to `map` to hopefully reduce confusion between this and `--enabled-users/--disabled-users`.

tools/copyblocks/README.md
```diff
@@ -13,6 +13,7 @@ The currently supported services are Amazon Simple Storage Service (S3 and S3-co
 - Include or exclude users from having blocks copied (`--enabled-users` and `--disabled-users`)
 - Configurable minimum block duration (`--min-block-duration`) and (`--skip-no-compact-block-duration-check`) to target blocks that are not awaiting compaction
 - Configurable time range (`--min-time` and `--max-time`) to only copy blocks inclusively within a provided range
+- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` would map source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If a mapping is not provided for a tenant it is assumed to be identical to the source tenant.
```
tacole02, line 16:
```suggestion
- Copy blocks between tenants with `--cross-tenant-mapping`. For instance, `--cross-tenant-mapping="tenant1:tenant2,tenant3:tenant4"` maps source blocks from `tenant1` to `tenant2` and source blocks from `tenant3` to tenant `tenant4`. If you don't provide a mapping for a tenant, it is assumed to be identical to the source tenant.
```

tacole02 (approved):

tools/copyblocks/main.go
```diff
@@ -297,7 +329,7 @@ func copyBlocks(ctx context.Context, cfg config, logger log.Logger, m *metrics)
 			m.blocksCopied.Inc()
 			level.Info(logger).Log("msg", "block copied successfully")
 
-			err = uploadCopiedMarkerFile(ctx, sourceBucket, tenantID, blockID, destBucket.Name())
+			err = uploadCopiedMarkerFile(ctx, sourceBucket, sourceTenantID, blockID, destBucket.Name())
```
narqo, line 332: (nit) Maybe we can add a note about the behaviour around markers, just it was clear from the code

```suggestion
			// Note that markers copying ignores the tenants mapping. This may not work if copying blocks to multiple destination tenants.
			err = uploadCopiedMarkerFile(ctx, sourceBucket, sourceTenantID, blockID, destBucket.Name())
```

tools/copyblocks/main.go
```diff
@@ -59,6 +62,7 @@ func (c *config) registerFlags(f *flag.FlagSet) {
 	f.DurationVar(&c.copyPeriod, "copy-period", 0, "How often to repeat the copy. If set to 0, copy is done once, and the program stops. Otherwise, the program keeps running and copying blocks until it is terminated.")
 	f.Var(&c.enabledUsers, "enabled-users", "If not empty, only blocks for these users are copied.")
 	f.Var(&c.disabledUsers, "disabled-users", "If not empty, blocks for these users are not copied.")
+	f.Var(&c.crossTenantMapping, "cross-tenant-mapping", "A comma-separated list of (source tenant):(destination tenant). If a source tenant is not mapped then its destination tenant is assumed to be identical.")
```
narqo, line 65: > cross-tenant-mapping

(nit) Maybe it's better to stick to "users" on the inputs level? The command has `--enabled-users` and `--disabled-users` flags. So it feels confusing if one doesn't know that those are the same. Can we name the flag `--user-mapping`?

narqo (approved):
Left a couple of nits, but the changes overall look good 🔥

tools/copyblocks/main.go
andyasp: Nice catch, thanks. I changed it to `--user-mapping` and updated the documentation to reflect that.

tools/copyblocks/main.go
andyasp: I reworded it a bit and added the comment:
```
// Note that only the blockID and destination bucket are considered in the copy marker.
// If multiple tenants in the same destination bucket are copied to from the same source tenant the markers will currently clash.
```

PR #10118, opened 2024-12-04
Block-builder: pull jobs from scheduler

seizethedave:
#### What this PR does

This adds a "pull" mode to block-builder so that if it is configured with a scheduler at startup, it will live its life in pull-mode, obtaining and completing jobs from a block-builder-scheduler service.

Pull-mode tests are largely duplicated from existing tests. This is temporary as the prior tests will be deleted once we're relying on the scheduler.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/blockbuilder.go
```diff
@@ -104,11 +166,68 @@ func (b *BlockBuilder) starting(context.Context) (err error) {
...
+			Partition: jobSpec.Partition,
+			At:        jobSpec.StartOffset,
+			Metadata:  "{}", // FIXME. This and other fields need to be plumbed in via the scheduler RPC layer.
+		},
+		CommitRecordTimestamp: jobSpec.CommitRecTs,
+		LastSeenOffset:        jobSpec.LastSeenOffset,
+		LastBlockEnd:          jobSpec.LastBlockEndTs,
+	}
+
+	cycleEndTime := jobSpec.LastBlockEndTs // ???
```
codesome, line 222: This is something the scheduler can send as part of the spec. Also, the current implementation of consumePartition() stops consuming after it has hit cycleEndTime. But we should update it to end consuming when it hits the cycleEndOffset perhaps.

codesome: At the scheduler side, cycleEndTime can be the time when the job was freshly created.

seizethedave: OK, cool. I'd like to address this comment in #10105 where I'm attempting to rectify all of the Kafka administrivia bits.

seizethedave: Actually, I went ahead and added it.

narqo (approved):
🔥 Overall, looks good to me. I left two small comments, but I don't have strong feelings about them.

pkg/blockbuilder/config_test.go
```diff
@@ -108,3 +108,13 @@ func blockBuilderConfig(t *testing.T, addr string) (Config, *validation.Override
 
 	return cfg, overrides
 }
+
+func blockBuilderPullModeConfig(t *testing.T, addr string) (Config, *validation.Overrides) {
```
codesome, line 112: Let's move this to the other test file where it is being used.

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -303,6 +486,116 @@ func TestBlockBuilder_ReachHighWatermarkBeforeLastCycleSection(t *testing.T) {
...
+	require.NoError(t, err)
+
+	require.NoError(t, services.StartAndAwaitRunning(ctx, bb))
+	t.Cleanup(func() {
+		require.NoError(t, services.StopAndAwaitTerminated(ctx, bb))
+	})
+
+	// Wait for both jobs to complete.
+	require.Eventually(t, func() bool {
+		return bb.jobIteration.Load() >= 2
```
codesome, line 579: Shall we also check for completed counts with scheduler.counts()? Should completed jobs be 2?

codesome (approved):
Nice work! Only small comments.

For future: I see a lot of code duplication in the test code, both in new and old tests, mainly in setting up kafka, BB, and something around adding samples. We should be able to deduplicate a lot of it so that tests are easier to read.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -104,11 +170,65 @@ func (b *BlockBuilder) starting(context.Context) (err error) {
...
+				return nil
+			}
+			level.Error(b.logger).Log("msg", "failed to get job", "err", err)
+			continue
+		}
+
+		if _, err := b.consumeJob(ctx, key, spec); err != nil {
+			level.Error(b.logger).Log("msg", "failed to consume job", "job_id", key.Id, "epoch", key.Epoch, "err", err)
+			continue
+		}
```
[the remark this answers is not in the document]
seizethedave: Yeah, I think I recall discussing this in a design document. Scheduler will know when it has failed as it won't receive an update within X seconds. We can always enhance this by adding failure info to the UpdateJob RPC. Initially I'm just keeping it barebones.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -61,23 +83,67 @@ func New(
...
+		nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// nolint:staticcheck // grpc.Dial() has been deprecated; we'll address it before upgrading to gRPC 2.
+	conn, err := grpc.Dial(b.cfg.SchedulerConfig.Address, dialOpts...)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
seizethedave: Ah yes, that's a good call. When the shutdown signal comes, there will (probably) be unflushed completions in the scheduler client. We should flush these, as well as close the client.

pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -303,6 +486,116 @@ func TestBlockBuilder_ReachHighWatermarkBeforeLastCycleSection(t *testing.T) {
 	)
 }
 
+// When there are no records to consume before the last cycle section, the whole cycle should bail.
+func TestBlockBuilder_ReachHighWatermarkBeforeLastCycleSection_PullMode(t *testing.T) {
+	// FIXME: I'm currently trying to understand why uncommenting this line causes tests to fail when all run together:
+	// t.Parallel()
```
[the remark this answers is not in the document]
seizethedave: You're onto something. When I run the entire file's tests at once with this `t.Parallel` uncommented, I get this test and a random other one failing with messages like:
```
        	Error Trace:	/Users/davidgrant/dev/mimir/pkg/blockbuilder/blockbuilder_test.go:469
        	Error:      	Condition never satisfied
        	Test:       	TestBlockBuilder_ReachHighWatermarkBeforeLastCycleSection
        	Messages:   	expected kafka commits
```

I took a bunch of traces and it turns out that when you crank up the parallelism all file I/O and other syscalls get monumentally slowed down on OS X, but not Linux. So I think this is fine for now.

PR #10130, opened 2024-12-04
UsageTracker: don't CAS when tracking series

colega:
When we're tracking series, the timestamp we're setting is `now` truncated to minutes. It's VERY unlikely that some other goroutine has updated this timestamp to something newer already.

Maybe there was a newer timestamp set from a snapshot, in which case we should overwrite it, as we cleanup based on *our* wall clock anyway.

### DIFF
--- a/pkg/usagetracker/tracker_store.go
+++ b/pkg/usagetracker/tracker_store.go
@@ -271,8 +271,7 @@ func (shard *tenantShard) trackSeries(series []uint64, now minutes, totalTenantS
 			continue
 		}
 
-		// CAS it if it's higher.
-		casIfGreater(now, ts)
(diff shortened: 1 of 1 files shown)


pracucci (approved):

colega:
```
~/w/github.com/grafana/mimir on  usage-tracker-dont-cas-when-tracking 17:52:18
$ go test ./pkg/usagetracker/...
?   	github.com/grafana/mimir/pkg/usagetracker/usagetrackerpb	[no test files]
ok  	github.com/grafana/mimir/pkg/usagetracker	0.395s
ok  	github.com/grafana/mimir/pkg/usagetracker/usagetrackerclient	8.641s
```

PR #10131, opened 2024-12-04
UsageTracker: limit series preallocation

colega:
Don't preallocate MaxUint64/128 if tenant has no limit set.

### DIFF
--- a/pkg/usagetracker/tracker_store.go
+++ b/pkg/usagetracker/tracker_store.go
@@ -18,6 +18,7 @@ import (
 )
 
 const shards = 128
+const noLimit = math.MaxUint64
 
 // trackerStore holds the core business logic of the usage-tracker abstracted in a testable way.
 // trackerStore should not depend on wall clock: time.Now() should be always injected as a parameter,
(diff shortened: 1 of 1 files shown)


pkg/usagetracker/tracker_store.go
```diff
@@ -135,7 +136,13 @@ func (t *trackerStore) getOrCreateTenantShard(userID string, shard uint8, limit
 		return m
 	}
 
-	m = &tenantShard{series: make(map[uint64]*atomic.Int32, int(limit/shards))}
+	var series map[uint64]*atomic.Int32
+	if limit == noLimit {
+		series = make(map[uint64]*atomic.Int32)
```
pracucci, line 141: Instead of not preallocating at all, should we use a large number? If a tenant has no limit, probably its number of series is pretty large.

pkg/usagetracker/tracker_store.go
```diff
@@ -271,7 +278,8 @@ func (shard *tenantShard) trackSeries(series []uint64, now minutes, totalTenantS
 			continue
 		}
 
-		ts.Store(int32(now))
+		// CAS it if it's higher.
+		casIfGreater(now, ts)
```
pracucci, line 282: Is this intentional?

pkg/usagetracker/tracker_store.go
```diff
@@ -282,7 +290,8 @@ func (shard *tenantShard) trackSeries(series []uint64, now minutes, totalTenantS
 			s := pending[0]
 			ts, ok := shard.series[s]
 			if ok {
-				ts.Store(int32(now))
+				// Was created in the meantime.
+				casIfGreater(now, ts)
```
pracucci, line 294: Is this intentional?

pracucci (approved):

pkg/usagetracker/tracker_store.go
colega: Removed from this PR.

pkg/usagetracker/tracker_store.go
colega: Removed from this PR.

PR #10132, opened 2024-12-04
UsageTracker: process created series events

colega:
This adds a method to process series creation events, also a test that checks that it's compatible with the events produced.

### DIFF
--- a/pkg/usagetracker/tracker_store.go
+++ b/pkg/usagetracker/tracker_store.go
@@ -50,7 +50,7 @@ type limiter interface {
 
 // events provides an abstraction to publish usage-tracker events.
 type events interface {
-	publishCreatedSeries(ctx context.Context, tenantID string, series []uint64) error
+	publishCreatedSeries(ctx context.Context, tenantID string, series []uint64, timestamp time.Time) error
 }
 
 func newTrackerStore(idleTimeout time.Duration, logger log.Logger, l limiter, ev events) *trackerStore {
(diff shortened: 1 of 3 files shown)


pkg/usagetracker/tracker_store.go
```diff
@@ -300,16 +341,54 @@ func (shard *tenantShard) trackSeries(series []uint64, now minutes, totalTenantS
...
-	rejected = append(rejected, pending...)
+	// Whatever is still missing is rejected (if any)
+	rejected = append(rejected, missing...)
 	return created, rejected
 }
 
+// processCreatedSeriesEvent creates the series coming from an event from Kafka,
+// i.e. series that were created by a different instance.
+// This does not check the limits, as we prioritize staying consistent across replicas over enforcing limits.
+// Timestamp is not updated if series exists already.
```
pracucci, line 357: Shouldn't be updated with the highest value between the two? I guess it shouldn't happen, but never know.

pkg/usagetracker/tracker_store.go
```diff
@@ -300,16 +341,54 @@ func (shard *tenantShard) trackSeries(series []uint64, now minutes, totalTenantS
...
+func (shard *tenantShard) processCreatedSeriesEvent(series []uint64, timestamp minutes, totalTenantSeries *atomic.Uint64) {
+	// missing contains the series that have to be created.
+	// As we advance through series, we reuse the same slice to avoid allocations.
+	missing := series[:0]
+
+	// First try to update the series that already exist, the ones that don't exist are moved to missing.
```
pracucci, line 363: [nit] The comment says "update" but we're not doing any update here. We're just looking for missing ones.

pracucci (approved):
Nice work! This was easy to review.

pkg/usagetracker/tracker_store.go
colega: It could happen but we don't really care as we'll update when we process the snapshot later.

PR #10137, opened 2024-12-05
Init usage-tracker Kafka client

pracucci:
#### What this PR does

Init usage-tracker Kafka client. There are 2 different configs for writer and reader, because we'll need to address different brokers. The config is inherited by `ingest.KafkaConfig` that contains a bunch of extra stuff we don't need. To productionise the usage-tracker we'll have to decouple the basic Kafka client config from the rest used to configure the ingest storage, but this shouldn't be a blocker for the prototype.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/usagetracker/tracker.go
+++ b/pkg/usagetracker/tracker.go
@@ -64,6 +64,9 @@ func (c *Config) Validate() error {
 	if err := c.SnapshotsStorage.Validate(); err != nil {
 		return err
 	}
+	if c.IdleTimeout <= 0 || c.IdleTimeout > time.Hour {
+		return fmt.Errorf("invalid usage-tracker idle timeout %q, should be greater than 0 and less than 1 hour", c.IdleTimeout)
+	}
 
 	return nil
 }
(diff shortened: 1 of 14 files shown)


pkg/usagetracker/tracker.go
```diff
@@ -38,25 +46,49 @@ func (c *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 
 	c.InstanceRing.RegisterFlags(f, logger)
 	c.PartitionRing.RegisterFlags(f)
+	c.EventsStorage.RegisterFlags(f)
 	c.SnapshotsStorage.RegisterFlagsWithPrefixAndDefaultDirectory("usage-tracker.snapshot-storage.", "usagetrackersnapshots", f)
 
 	f.DurationVar(&c.IdleTimeout, "usage-tracker.idle-timeout", 20*time.Minute, "The time after which series are considered idle and not active anymore. Must be greater than 0 and less than 1 hour.")
 }
 
+func (c *Config) Validate() error {
```
colega, line 55: We can also validate the IdleTimeout < 1h here now (instead of the constructor)

colega (approved):

pkg/usagetracker/tracker.go
pracucci: Done in https://github.com/grafana/mimir/pull/10137/commits/80b18184698a4112644e7d459899dc6db9eaca64

PR #10143, opened 2024-12-05
UsageTracker: publish and consume SeriesCreated events

colega:
This should publish the SeriesCreated event to Kafka, and consume it from there.

### DIFF
--- a/pkg/usagetracker/tracker.go
+++ b/pkg/usagetracker/tracker.go
@@ -42,7 +42,11 @@ type Config struct {
 	EventsStorage    EventsStorageConfig `yaml:"events_storage"`
 	SnapshotsStorage bucket.Config       `yaml:"snapshots_storage"`
 
-	IdleTimeout time.Duration `yaml:"idle_timeout"`
+	IdleTimeout                           time.Duration `yaml:"idle_timeout"`
+	CreatedSeriesEventsMaxPending         int           `yaml:"max_pending_created_series_events"`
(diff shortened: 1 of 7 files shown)


pkg/usagetracker/tracker.go
```diff
@@ -50,6 +58,10 @@ func (c *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 	c.SnapshotsStorage.RegisterFlagsWithPrefixAndDefaultDirectory("usage-tracker.snapshot-storage.", "usagetrackersnapshots", f)
 
 	f.DurationVar(&c.IdleTimeout, "usage-tracker.idle-timeout", 20*time.Minute, "The time after which series are considered idle and not active anymore. Must be greater than 0 and less than 1 hour.")
+	f.IntVar(&c.CreatedSeriesEventsMaxPending, "usage-tracker.created-series-events-max-pending", 1000, "Maximum number of pending created series events waiting to be published.")
```
pracucci, line 61: I think we could be much more generous with the default here.

pracucci (approved):
Nice job! When productionizing this (if we will) we may consider encoding 1 batch of events in 1 record (we've seen less but larger records are more efficient in franz-go).

colega:
> Nice job! When productionizing this (if we will) we may consider encoding 1 batch of events in 1 record (we've seen less but larger records are more efficient in franz-go).

Lets see what's the rate of the requests here first.

PR #10145, opened 2024-12-05
Distributor: remove samples with duplicated timestamps within a single timeseries

duricanikolic:
#### What this PR does
This PR introduces an additional request validation at distributor level in order to remove all float and histogram samples containing a duplicated timestamp. Namely, for each timeseries in a request, distributor iterates all its float and histogram samples, and as soon as a sample with a timestamp that has already been visited is found, the corresponding sample is discarded from the timeseries. Once the validation is completed, the request without discarded samples is forwarded to ingesters. 

Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`.

Benchmarks for the following usecases have been added (all these inputs have been applied to `distributor.validateSeries()`:
- a timeseries containing one float sample
- a timeseries containing one histogram sample
- a timeseries containing one float sample and one histogram sample
- a timeseries containing two float samples without timestamp duplications
- a timeseries containing two histogram samples without timestamp duplications
- a timeseries containing two float samples and two histogram samples without timestamp duplications
- a timeseries containing 80K float samples, all with the same timestamp, such that 40K have the same value, and 40K different values,
- a timeseries containing 80K histogram samples, all with the same timestamp, such that 40K have the same histogram, and 40K different histograms,
- a timeseries containing 80K float samples and 80K histogram samples, all with the same timestamp, such that 40K float samples and 40K histogram samples have the same value, while the remaining float and histogram samples have different values.

Comparison of `go test -bench BenchmarkDistributor_SampleDuplicateTimestamp -run=^# -count=6 -benchmem`
 of the benchmarks on `main` and the PR brench produced the following results:
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/distributor
cpu: Apple M2
                                                                                     │ withoutDeduplication.txt │  withDeduplication.txt   │
                                                                                     │         sec/op           │     sec/op     vs base   │
one_timeseries_with_one_sample-8                                                                  28.00n ±  3%    30.51n ±   3%   +8.93% (p=0.002 n=6)
one_timeseries_with_one_histogram-8                                                               28.66n ± 16%    31.55n ±   5%        ~ (p=0.056 n=6)
one_timeseries_with_one_sample_and_one_histogram-8                                                33.07n ± 14%    35.51n ±   4%        ~ (p=0.093 n=6)
one_timeseries_with_two_samples-8                                                                 32.36n ±  1%    63.52n ±   7%  +96.29% (p=0.002 n=6)
one_timeseries_with_80_000_samples_with_duplicated_timestamps-8                                   423.4µ ± 82%    339.7µ ±  94%        ~ (p=0.132 n=6)
one_timeseries_with_80_000_histograms_with_duplicated_timestamps-8                               12.686m ± 51%    6.996m ± 192%        ~ (p=0.310 n=6)
one_timeseries_with_two_histograms-8                                                              48.98n ± 27%    71.58n ±   9%  +46.13% (p=0.002 n=6)
one_timeseries_with_two_samples_and_two_histograms-8                                              62.76n ±  8%   106.85n ±   9%  +70.25% (p=0.002 n=6)
one_timeseries_with_80_000_samples_and_80_000_histograms_with_duplicated_timestamps-8            14.402m ± 44%    6.554m ±  45%  -54.49% (p=0.004 n=6)
geomean                                                                                           1.806µ          1.854µ          +2.66%

                                                                                     │ withoutDeduplication.txt │    withDeduplication.txt    │
                                                                                     │         B/op             │     B/op       vs base      │
one_timeseries_with_one_sample-8                                                                 0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_one_histogram-8                                                              0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_one_sample_and_one_histogram-8                                               0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_two_samples-8                                                                0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_80_000_samples_with_duplicated_timestamps-8                                  38.50 ± 77%     1646.00 ±  2%  +4175.32% (p=0.002 n=6)
one_timeseries_with_80_000_histograms_with_duplicated_timestamps-8                             1.107Ki ± 51%     2.197Ki ± 54%    +98.46% (p=0.002 n=6)
one_timeseries_with_two_histograms-8                                                             0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_two_samples_and_two_histograms-8                                             0.000 ±  0%       0.000 ±  0%          ~ (p=1.000 n=6) ¹
one_timeseries_with_80_000_samples_and_80_000_histograms_with_duplicated_timestamps-8          1.267Ki ± 42%     3.569Ki ±  7%   +181.80% (p=0.002 n=6)
geomean                                                                                                      ²                    +83.78%               ²
¹ all samples are equal
² summaries must be >0 to compute geomean
                                                                                     │ withoutDeduplication.txt │ withDeduplication.txt │
                                                                                     │        allocs/op         │  allocs/op   vs base  │
one_timeseries_with_one_sample-8                                                                 0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_one_histogram-8                                                              0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_one_sample_and_one_histogram-8                                               0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_two_samples-8                                                                0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_80_000_samples_with_duplicated_timestamps-8                                  0.000 ±  0%        6.000 ±   0%  ? (p=0.002 n=6)
one_timeseries_with_80_000_histograms_with_duplicated_timestamps-8                               11.50 ± 48%        12.00 ± 108%  ~ (p=0.567 n=6)
one_timeseries_with_two_histograms-8                                                             0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_two_samples_and_two_histograms-8                                             0.000 ±  0%        0.000 ±   0%  ~ (p=1.000 n=6) ¹
one_timeseries_with_80_000_samples_and_80_000_histograms_with_duplicated_timestamps-8            13.50 ± 41%        14.00 ±  21%  ~ (p=0.853 n=6)
geomean                                                                                                      ²                    ?               ²
¹ all samples are equal
² summaries must be >0 to compute geomean

```

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/distributor/distributor.go
```diff
@@ -739,70 +740,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+func (d *Distributor) validateSeries(nowt time.Time, ts *mimirpb.PreallocTimeseries, userID, group string, skipLabelValidation, skipLabelCountValidation bool, minExemplarTS, maxExemplarTS int64) (bool, error) {
 	if err := validateLabels(d.sampleValidationMetrics, d.limits, userID, group, ts.Labels, skipLabelValidation, skipLabelCountValidation); err != nil {
-		return err
+		return true, err
 	}
 
 	now := model.TimeFromUnixNano(nowt.UnixNano())
 
-	for _, s := range ts.Samples {
+	timestamps := make(map[int64]struct{})
```
pracucci, line 753: Some suggestions:

1. To make this code a bit more readable, I would move the validation of a series samples and histograms to a dedicated function (e.g. `validateSeriesSamples()` and `validateSeriesHistograms()`). I don't feel strong about this, but I'm just noting `validateSeries()` function is getting quite big and a bit messy.
2. Most requests have 1 sample per series. To keep performance unchanged for the happy case, I would skip the duplicated samples check at all if there's only 1 series (so that we don't allocate `map[int64]struct{}` at all). Moving the logic to a dedicated function will make it easier to circuit break (e.g. `if len(rs.Samples) == 1 { return validateSample() }`)
3. In most cases, we don't expected duplicated samples for the same timestamp. For this reason, I would pre-allocate `make(map[int64]struct{})` to reduce memory re-allocations. We could pre-allocated to the `min(len(ts.Samples), 100))` where 100 is a reasonable max non-duplicate samples we expect in a single request)

pkg/distributor/distributor_test.go
```diff
@@ -1467,6 +1467,244 @@ func TestDistributor_Push_HistogramValidation(t *testing.T) {
...
+	const (
+		metricName         = "series"
+		ts1        int64   = 10
+		ts2        int64   = 20
+		ts3        int64   = 30
+		ts4        int64   = 40
+		v1         float64 = 1
+		v2         float64 = 2
+		v3         float64 = 3
+		v4         float64 = 4
```
pracucci, line 1480: [nit] What's the purpose of defining constants of constants? Can't we just use the actual values in the test? :)

pkg/distributor/distributor_test.go
```diff
@@ -1467,6 +1467,244 @@ func TestDistributor_Push_HistogramValidation(t *testing.T) {
...
+			b.ResetTimer()
+			for n := 0; n < b.N; n++ {
+				b.StopTimer()
+				timeseries := tc.setup()
+				b.StartTimer()
+				for i, ts := range timeseries {
+					shouldRemove, err := ds[0].validateSeries(now, &ts, "user", "test-group", true, true, 0, 0)
+					require.False(b, shouldRemove)
+					if len(tc.expectedErrors) == 0 {
+						require.NoError(b, err)
```
pracucci, line 1692: Don't use the testify library (so `require` and `assert`) at all in benchmarks. It slow down the benchmark and make your results skered. Instead, you should do simple checks like:

```
if err != nil {
  b.Fatal(err)
}
```

pkg/distributor/distributor_test.go
```diff
@@ -1467,6 +1467,244 @@ func TestDistributor_Push_HistogramValidation(t *testing.T) {
...
+		metricName              = "series"
+		testSize                = 80_000
+		numberOfDifferentValues = 40_000
+	)
+	labels := []string{labels.MetricName, metricName, "job", "job", "service", "service"}
+
+	now := mtime.Now()
+	timestamp := now.UnixMilli()
+
+	testCases := map[string]struct {
```
pracucci, line 1589: You tested with extreme edge cases, but I would like to also see a benchmark for the happy path
- 1 sample per series, no duplicated samples
- 2 samples per series, no duplicated samples

Then, run this benchmark both against `main` and this PR (run each benchmark at least 6 times). And then compare the results with `benchcmp` (please post the updated benchmark in the PR description).

pkg/distributor/distributor_test.go
```diff
@@ -1467,6 +1467,244 @@ func TestDistributor_Push_HistogramValidation(t *testing.T) {
...
+					} else {
+						if tc.expectedErrors[i] == nil {
+							require.NoError(b, err)
+						} else {
+							require.Error(b, err)
+							require.Equal(b, tc.expectedErrors[i], err)
+						}
+					}
+				}
+				assert.Equal(b, tc.expectedSamples, timeseries)
```
pracucci, line 1702: Don't run these type of assertions in the benchmark. Just measure performance in benchmark, and move assertions to unit tests. Why? To get accurate measurements, you don't want CPU and memory allocations being inflated by assertions.

pkg/distributor/distributor_test.go
```diff
@@ -1467,6 +1467,244 @@ func TestDistributor_Push_HistogramValidation(t *testing.T) {
...
+	require.Len(b, regs, 1)
+
+	for name, tc := range testCases {
+		b.Run(name, func(b *testing.B) {
+			b.ReportAllocs()
+			b.ResetTimer()
+			for n := 0; n < b.N; n++ {
+				b.StopTimer()
+				timeseries := tc.setup()
+				b.StartTimer()
```
pracucci, line 1687: I'm not an expert on micro benchmarking in golang, but I'm not sure that this is a zero-sum game. The problem is that there's a stop-of-the-world for each `StartTimer()` and `StopTimer()`, and calling it in a tight loop may actually get measurements a bit skewed. What I typically do instead is to prepare `b.N` requests inside `b.Run()` then I call `b.ResetTimer()` and I run all `b.N` tests, using all the previously prepared requests.

pkg/distributor/distributor_test.go
duricanikolic: I have removed the constants

pkg/distributor/distributor.go
duricanikolic: I have applied all your suggestions. The first one made the code a bit more complicated, but I tried to make it as much readable as possible.

duricanikolic:
I have fixed most of the findings. The only missing thing are improved benchmarks, that I am going to fix tomorrow.

chencs:

pkg/distributor/distributor_test.go
duricanikolic: I have removed usage of the testify library from this benchmark.

pkg/distributor/distributor_test.go
duricanikolic: I have added the following benchmarks:
- 1 sample per series, no duplicated samples
- 1 histogram per series, no duplicated samples
- 1 sample and 1 histogram per series, no duplicated samples
- 2 samples per series, no duplicated samples
- 2 histograms per series, no duplicated samples
- 2 samples and 2 histograms per series, no duplicated samples

pkg/distributor/distributor_test.go
duricanikolic: I have removed this

pkg/distributor/distributor_test.go
duricanikolic: Thank you, this was a great tip.

pkg/distributor/distributor.go
```diff
@@ -737,40 +738,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+	timestamps := make(map[int64]struct{}, min(len(ts.Histograms), 100))
+	currPos := 0
+	duplicatesFound := false
 	histogramsUpdated := false
-	for i := range ts.Histograms {
-		updated, err := validateSampleHistogram(d.sampleValidationMetrics, now, d.limits, userID, group, ts.Labels, &ts.Histograms[i])
+	for idx := range ts.Histograms {
+		if _, ok := timestamps[ts.Histograms[idx].Timestamp]; ok {
+			// A sample with the same timestamp has already been validated, so we skip it.
+			duplicatesFound = true
```
pracucci, line 810: Can be simplified:

```suggestion
			histogramsUpdated = true
```

pkg/distributor/distributor.go
```diff
@@ -737,40 +738,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+		}
+		if updated {
+			ts.HistogramsUpdated()
+		}
+		return nil
 	}
 
+	timestamps := make(map[int64]struct{}, min(len(ts.Histograms), 100))
+	currPos := 0
+	duplicatesFound := false
```
pracucci, line 805: Can be simplified: `duplicatesFound` can be removed.

pkg/distributor/distributor.go
```diff
@@ -737,40 +738,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+		timestamps[ts.Histograms[idx].Timestamp] = struct{}{}
+		updated, err := validateSampleHistogram(d.sampleValidationMetrics, now, d.limits, userID, group, ts.Labels, &ts.Histograms[idx])
 		if err != nil {
 			return err
 		}
 		histogramsUpdated = histogramsUpdated || updated
+
+		if currPos != idx {
+			ts.Histograms[currPos] = ts.Histograms[idx]
+		}
```
pracucci, line 823: Can be simplified. I've run a benchmark and it's not slower:

```suggestion
		ts.Histograms[currPos] = ts.Histograms[idx]
```

pkg/distributor/distributor.go
```diff
@@ -737,40 +738,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
 	}
+
 	if histogramsUpdated {
 		ts.HistogramsUpdated()
 	}
 
+	if duplicatesFound {
+		ts.Histograms = ts.Histograms[:currPos]
+		ts.HistogramsUpdated()
+	}
```
pracucci, line 834: Can be simplified:

```suggestion
	if histogramsUpdated {
		ts.Histograms = ts.Histograms[:currPos]
		ts.HistogramsUpdated()
	}
```

pkg/distributor/distributor.go
```diff
@@ -737,40 +738,111 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
...
+		}
+
+		timestamps[s.TimestampMs] = struct{}{}
 		if err := validateSample(d.sampleValidationMetrics, now, d.limits, userID, group, ts.Labels, s); err != nil {
 			return err
 		}
+
+		if currPos != idx {
+			ts.Samples[currPos] = s
+		}
```
pracucci, line 771: Can be simplified. I've run a benchmark and it's not slower:

```suggestion
		ts.Samples[currPos] = s
```

pracucci (approved):
Nice work, LGTM! I just left few suggestions to simplify a bit the code.

The tiny regression you see in the benchmark for the 1 sample/histogram case is due to the extra function call. In absolute terms they're negligible (few nanos) so I'm not concerned.

bboreham:
Prior to this we would ignore repeated timestamps with the same value; now we report them as an error.
Was that intentional?

PR #10147, opened 2024-12-05
Fix TestTrackerStore_Cleanup_Concurrency

colega:
One of the assumptions of trackerStore is that cleanup() is called at least once in each idleTimeout: this is needed to make sure that "minutes" timestamp doesn't get too old and starts looking as a newer one.

We didn't ensure this in the test, so weird stuff was happening.

Now we increase the timestamp one minute each time, and we set idle timeout to 1 minute, so we force the race conditions.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (approved):
I trust you ;) Rubberstamping

PR #10148, opened 2024-12-05
UsageTracker: register store as prometheus collector

colega:
This will export the number of activer series tracker for each tenant.

### DIFF
--- a/pkg/usagetracker/tracker_store_collector.go
+++ b/pkg/usagetracker/tracker_store_collector.go
@@ -1,0 +1,23 @@
+package usagetracker
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var _ prometheus.Collector = &trackerStore{}
+
+var trackerStoreMetricDesc = prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


pkg/usagetracker/tracker_store_collector.go
```diff
@@ -0,0 +1,23 @@
+package usagetracker
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var _ prometheus.Collector = &trackerStore{}
+
+var trackerStoreMetricDesc = prometheus.NewDesc(
+	"cortex_usagetracker_active_series",
```
pracucci, line 8: For when we'll productionise it (if we will) we should standardize metric prefix. My code has `cortex_usage_tracker`.

pracucci (approved):

pkg/usagetracker/tracker_store_collector.go
colega: I'll push a commit now so we don't forget.

PR #1971, opened 2022-05-31
Influx push handler

colega:
#### What this PR does

Adds an http handler that accepts writes in the Influx [Line protocol](https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/). This code is brought from https://github.com/grafana/influx2cortex with some minor adjustments.

#### Which issue(s) this PR fixes or relates to

None yet, internal discussion.

#### Checklist

- [x] Tests updated
- [ ] End-to-end tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #10153, opened 2024-12-06
Add influx push endpoint to mimir

alexgreenbank:
#### What this PR does

Add influx endpoint to Mimir.

Based on original code from https://github.com/grafana/mimir/pull/1971/ but brought up to date for modern Distributor framework.

MVP. Experimantal/hidden feature so no documentation yet.

#### Checklist

- [x] Tests updated.
- [ ] Documentation.


pkg/api/api.go
```diff
@@ -268,6 +269,10 @@ func (a *API) RegisterDistributor(d *distributor.Distributor, pushConfig distrib
 		pushConfig.MaxRecvMsgSize, d.RequestBufferPool, a.sourceIPs, a.cfg.SkipLabelNameValidationHeader,
 		a.cfg.SkipLabelCountValidationHeader, limits, pushConfig.RetryConfig, d.PushWithMiddlewares, d.PushMetrics, a.logger,
 	), true, false, "POST")
+	// TODO(alexg): hidden behind a featureflag or experimental config option?
+	a.RegisterRoute(InfluxPushEndpoint, distributor.InfluxHandler(
```
colega, line 273: I don't think a feature flag for this is needed. We can just state in the docs (about-versioning.md) that the endpoint is experimental.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+	// TODO(alexg): one argument for splitting up the decoding and conversion is to facilitate granular timings
+	// right now since ParseInfluxLineReader() does both
+	// The otel version decodes the whole input and then processes it, the existing Influx code parses each line as it
+	// decodes it.
+	level.Debug(spanLogger).Log(
+		"msg", "decodeAndConvert complete",
+		"bytesRead", bytesRead,
+	)
+	if err != nil {
+		level.Error(logger).Log("err", err.Error())
```
colega, line 43: Does this logger have all the required context? Can you also add some context here about what was going on when this happened? I'm scared of finding this log:
```
ts=2024-12-17 err="unexpected EOF"
```

Also, nit, `.Error()` call is not needed, just pass `err`.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+	"context"
+	"errors"
+	"net/http"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/dskit/grpcutil"
+	"github.com/grafana/dskit/httpgrpc"
+	"github.com/grafana/dskit/middleware"
+	io2 "github.com/influxdata/influxdb/v2/kit/io"
```
colega, line 15:
```suggestion
	influxio "github.com/influxdata/influxdb/v2/kit/io"
```

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/mimir/pkg/distributor/influxpush"
+	"github.com/grafana/mimir/pkg/mimirpb"
+	"github.com/grafana/mimir/pkg/util"
+	utillog "github.com/grafana/mimir/pkg/util/log"
+	"github.com/grafana/mimir/pkg/util/spanlogger"
+)
+
+func parser(ctx context.Context, r *http.Request, maxSize int, _ *util.RequestBuffers, req *mimirpb.PreallocWriteRequest, logger log.Logger) error {
```
colega, line 25: This function lives in `pkg/distributor`, I would say it's a little bit pretentious to take the name `parser` for this :D 

How about `influxRequestParser`?

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+		}
+
+		supplier := func() (*mimirpb.WriteRequest, func(), error) {
+			rb := util.NewRequestBuffers(requestBufferPool)
+			var req mimirpb.PreallocWriteRequest
+
+			if err := parser(ctx, r, maxRecvMsgSize, rb, &req, logger); err != nil {
+				// TODO(alexg): Do we even need httpgrpc here?
+				// Check for httpgrpc error, default to client error if parsing failed
```
colega, line 94: I don't see any httpgrpc errors being returned by `parser`

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+	)
+	if err != nil {
+		level.Error(logger).Log("err", err.Error())
+		// TODO(alexg): need to pass on the http.StatusBadRequest
+		// http.Error(w, err.Error(), http.StatusBadRequest)
+		return err
+	}
+
+	// Sigh, a write API optimisation needs me to jump through hoops.
+	pts := make([]mimirpb.PreallocTimeseries, 0, len(ts))
```
colega, line 50: You should use `mimirpb.PreallocTimeseriesSliceFromPool()` instead of creating a new slice every time.

Also, IMO it would make sense to change `influxpush.ParseInfluxLineReader` to return `[]PreallocTimeseries` instead of `[]Timesries` because that's what we deal with later.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+				w.WriteHeader(statusClientClosedRequest)
+				return
+			}
+			if errors.Is(err, io2.ErrReadLimitExceeded) {
+				// TODO(alexg): One thing we have seen in the past is that telegraf clients send a batch of data
+				// if it is too big they should respond to the 413 below, but if a client doesn't understand this
+				// it just sends the next batch that is even bigger. In the past this has had to be dealt with by
+				// adding rate limits to drop the payloads.
+				level.Warn(logger).Log("msg", "request too large", "err", err)
+				// TODO(alexg): max size and bytes received in error?
```
colega, line 123: This would definitely help customer to debug.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+				w.WriteHeader(http.StatusRequestEntityTooLarge)
+				return
+			}
+			// From: https://github.com/grafana/influx2cortex/blob/main/pkg/influx/errors.go
+
+			var (
+				httpCode int
+				errorMsg string
+			)
```
colega, line 132: While the code below isn't extremely complex, it's still a couple of `if`s, so I'd write here some sane defaults.
```suggestion
			httpCode := http.StatusInternalServerError
			errorMsg := "unknown error"
```

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+					httpCode = http.StatusServiceUnavailable
+				} else {
+					httpCode = errorCauseToHTTPStatusCode(distributorErr.Cause(), false)
+				}
+			}
+			if httpCode != 202 {
+				// This error message is consistent with error message in Prometheus remote-write handler, and ingester's ingest-storage pushToStorage method.
+				msgs := []interface{}{"msg", "detected an error while ingesting Influx metrics request (the request may have been partially ingested)", "httpCode", httpCode, "err", err}
+				if httpCode/100 == 4 {
+					// TODO(alexg): what is this?
```
colega, line 152: Logs with `insight=true` are visible to our Grafana Cloud customers.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+					msgs = append(msgs, "insight", true)
+				}
+				level.Error(logger).Log(msgs...)
+			}
+			if httpCode < 500 {
+				level.Info(logger).Log("msg", errorMsg, "response_code", httpCode, "err", tryUnwrap(err))
+			} else if httpCode >= 500 {
```
colega, line 159:
```suggestion
			} else {
```

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+				// This error message is consistent with error message in Prometheus remote-write handler, and ingester's ingest-storage pushToStorage method.
+				msgs := []interface{}{"msg", "detected an error while ingesting Influx metrics request (the request may have been partially ingested)", "httpCode", httpCode, "err", err}
+				if httpCode/100 == 4 {
+					// TODO(alexg): what is this?
+					msgs = append(msgs, "insight", true)
+				}
+				level.Error(logger).Log(msgs...)
+			}
+			if httpCode < 500 {
+				level.Info(logger).Log("msg", errorMsg, "response_code", httpCode, "err", tryUnwrap(err))
```
colega, line 158: Why do we need to `tryUnwrap`? Wrapping errors provides details about what went wrong, this is literally "tryToRemoveDetails".

pkg/distributor/influx_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"bytes"
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-kit/log"
+	io2 "github.com/influxdata/influxdb/v2/kit/io"
```
colega, line 13:
```suggestion
	influxio "github.com/influxdata/influxdb/v2/kit/io"
```

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return nil, dataLen, fmt.Errorf("problem reading body: %s", err)
+	}
+
+	points, err := models.ParsePointsWithPrecision(data, time.Now().UTC(), precision)
+	if err != nil {
+		return nil, dataLen, fmt.Errorf("can't parse points: %s", err)
+	}
+	a, b := writeRequestFromInfluxPoints(points)
+	return a, dataLen, b
```
colega, line 57:
```suggestion
	ts, err := writeRequestFromInfluxPoints(points)
	return ts, dataLen, err
```

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func writeRequestFromInfluxPoints(points []models.Point) ([]mimirpb.TimeSeries, error) {
+	// Technically the same series should not be repeated. We should put all the samples for
+	// a series in single client.Timeseries. Having said that doing it is not very optimal and the
+	// occurrence of multiple timestamps for the same series is rare. Only reason I see it happening is
+	// for backfilling and this is not the API for that. Keeping that in mind, we are going to create a new
+	// client.Timeseries for each sample.
+
+	var returnTs []mimirpb.TimeSeries
```
colega, line 67: This should be allocated to the correct expected size, but also as I commented above, consider returning `[]mimirpb.PreallocTimeseries` here, and taking it from a pool here.

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+// analog of invalidChars = regexp.MustCompile("[^a-zA-Z0-9_]")
+func replaceInvalidChars(in *string) {
+	for charIndex, char := range *in {
+		charInt := int(char)
+		if !((charInt >= 97 && charInt <= 122) || // a-z
+			(charInt >= 65 && charInt <= 90) || // A-Z
+			(charInt >= 48 && charInt <= 57) || // 0-9
+			charInt == 95) { // _
+
+			*in = (*in)[:charIndex] + "_" + (*in)[charIndex+1:]
```
colega, line 156: This allocates a new string (hopefully just one, compiler should be clever) for each one of the invalid characters. How about you modify the bytes slice before transforming it to a string?

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+		})
+	}
+
+	return returnTs, nil
+}
+
+// analog of invalidChars = regexp.MustCompile("[^a-zA-Z0-9_]")
+func replaceInvalidChars(in *string) {
+	for charIndex, char := range *in {
+		charInt := int(char)
```
colega, line 150: Why do you need it to be an int?

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+		if !((charInt >= 97 && charInt <= 122) || // a-z
```
colega, line 151: If you use 'a' instead of 97 (is it correct?) you won't need the comment and the code will be less prone to have bugs.

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+			(charInt >= 65 && charInt <= 90) || // A-Z
+			(charInt >= 48 && charInt <= 57) || // 0-9
+			charInt == 95) { // _
+
+			*in = (*in)[:charIndex] + "_" + (*in)[charIndex+1:]
+		}
+	}
+	// prefix with _ if first char is 0-9
+	if int((*in)[0]) >= 48 && int((*in)[0]) <= 57 {
```
colega, line 160: Should check first that `*in` isn't empty, otherwise it would panic.

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,180 @@
...
+		})
+		for _, tag := range tags {
+			key := string(tag.Key)
+			if key == "__name__" || key == internalLabel {
+				continue
+			}
+			replaceInvalidChars(&key)
+			lbls = append(lbls, mimirpb.LabelAdapter{
+				Name:  key,
+				Value: string(tag.Value),
```
colega, line 128: We can do `yoloString` here instead of `string()`, as the push code is crafted to avoid keeping references to the strings from `LabelAdapter`

pkg/distributor/influx.go
```diff
@@ -0,0 +1,177 @@
...
+
+	ts, bytesRead, err := influxpush.ParseInfluxLineReader(ctx, r, maxSize)
+	// TODO(alexg): one argument for splitting up the decoding and conversion is to facilitate granular timings
+	// right now since ParseInfluxLineReader() does both
+	// The otel version decodes the whole input and then processes it, the existing Influx code parses each line as it
+	// decodes it.
+	level.Debug(spanLogger).Log(
+		"msg", "decodeAndConvert complete",
+		"bytesRead", bytesRead,
+	)
```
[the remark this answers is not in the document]
alexgreenbank: Fixed in next push.

pkg/distributor/influx.go
alexgreenbank: Agreed, fixed in next push.

pkg/distributor/influx.go
alexgreenbank: Agreed, fixed in next push.

pkg/distributor/influx.go
alexgreenbank: Agreed, fixed in next push.

pkg/distributor/influx.go
alexgreenbank: Oops, should have got rid of that one. Thanks. Done.

pkg/distributor/influx.go
alexgreenbank: Aha, TIL! Will add it back in as that could be useful.

pkg/distributor/influx.go
alexgreenbank: Done.

pkg/distributor/influx_test.go
alexgreenbank: Done.

pkg/distributor/influxpush/parser.go
alexgreenbank: Done.

pkg/distributor/influx.go
alexgreenbank: Agreed. Done!

pkg/api/api.go
alexgreenbank: OK, done.

pkg/distributor/influx.go
alexgreenbank: Ah, `httpgprc` is used to smuggle both the http status code and the error message out of the `supplier()` function. I've removed the misleading comment.

pkg/distributor/influx.go
alexgreenbank: Done. Bubbled `bytesRead` down from the parsing function so that it is available for both metrics/histograms and for logging here.

pkg/distributor/influxpush/parser.go
alexgreenbank: The values come from the influx library which won't allow an empty string to be created, but I've added an extra check just in case. Also added some extra tests to ensure the assumption about the influx library holds.

pkg/distributor/influxpush/parser.go
alexgreenbank: Moved to literal chars instead of ASCII codes.

pkg/distributor/influx.go
alexgreenbank: Done. Fixed in latest push.

alexgreenbank:
OK, comments left to deal with
- [ ] end2end test for happy path 
- [ ] TryUnwrap
- [ ] work on byteslice rather than chars
- [ ] `yoloString`

pkg/distributor/influx.go
alexgreenbank: Legacy from the original code: https://github.com/grafana/influx2cortex/blob/main/pkg/influx/errors.go

Removed now.

pkg/distributor/influxpush/parser.go
alexgreenbank: Agreed. Done in latest commit.

pkg/distributor/influxpush/parser.go
alexgreenbank: Interesting! Fixed as per suggestion. I take it no other strings are guaranteed not to be referenced and so it can't be used anywhere else?

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,192 @@
...
+		lbls = append(lbls, mimirpb.LabelAdapter{
+			Name:  internalLabel, // An internal label for tracking active series
+			Value: "influx",
+		})
+		for _, tag := range tags {
+			key := string(tag.Key)
+			if key == "__name__" || key == internalLabel {
+				continue
+			}
+			replaceInvalidChars(&key)
```
colega, line 128: We create a string from the `tag.Key` slice (first allocation), then we pass it as a pointer to `replaceInvalidChars` which allocates a slice from that string (another allocation), and then we write into that pointer the string we allocate (third allocation) from that bytes slice.

Why don't we just do `key := replaceInvalidChars(tag.Key)`, and make `replaceInvalidChars` just modify the incoming bytes slice and then return it?

pkg/api/api.go
```diff
@@ -256,6 +256,7 @@ func (a *API) RegisterRuntimeConfig(runtimeConfigHandler http.HandlerFunc, userL
 
 const PrometheusPushEndpoint = "/api/v1/push"
 const OTLPPushEndpoint = "/otlp/v1/metrics"
+const InfluxPushEndpoint = "/api/v1/influx/push"
```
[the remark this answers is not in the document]
alexgreenbank: Yes, sadly it's burned into many of the agents. I'd love the prefix to be `/influx` but I don't think it is possible.

I will make one final check though.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,154 @@
...
+	level.Debug(spanLogger).Log("msg", "decodeAndConvert complete", "bytesRead", bytesRead)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to parse Influx push request", "err", err)
+		return bytesRead, err
+	}
+
+	level.Debug(spanLogger).Log(
+		"msg", "Influx to Prometheus conversion complete",
+		"metric_count", len(pts),
+	)
```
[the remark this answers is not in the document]
alexgreenbank: OK, done in next commit.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,154 @@
...
+func influxRequestParser(ctx context.Context, r *http.Request, maxSize int, _ *util.RequestBuffers, req *mimirpb.PreallocWriteRequest, logger log.Logger) (int, error) {
+	spanLogger, ctx := spanlogger.NewWithLogger(ctx, logger, "Distributor.InfluxHandler.decodeAndConvert")
+	defer spanLogger.Span.Finish()
+
+	spanLogger.SetTag("content_type", r.Header.Get("Content-Type"))
+	spanLogger.SetTag("content_encoding", r.Header.Get("Content-Encoding"))
+	spanLogger.SetTag("content_length", r.ContentLength)
+
+	pts, bytesRead, err := influxpush.ParseInfluxLineReader(ctx, r, maxSize)
+	level.Debug(spanLogger).Log("msg", "decodeAndConvert complete", "bytesRead", bytesRead)
```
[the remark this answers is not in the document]
alexgreenbank: OK, done in next commit.

pkg/distributor/influxpush/parser.go
alexgreenbank: Done as agreed...

alexgreenbank:
All checks passed. Blimey.

tacole02 (approved):
Looks good!

pkg/api/api.go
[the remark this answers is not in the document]
colega: Okay!

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,188 @@
...
+			// If the field name is not "value" then we append it to the name, fixing chars as we go
+			name += "_" + string(replaceInvalidChars([]byte(field)))
+		}
+
+		tags := pt.Tags()
+		lbls := make([]mimirpb.LabelAdapter, 0, len(tags)+2) // An additional one for __name__, and one for internal label
+		lbls = append(lbls, mimirpb.LabelAdapter{
+			Name:  labels.MetricName,
+			Value: name,
+		})
```
colega, line 118:
```suggestion
		name := replaceInvalidChars(pt.Name())
		if field != "value" {
			// If the field name is not "value" then we append it to the name, fixing chars as we go
			name = append([]byte{'_'}, name...)
		}

		tags := pt.Tags()
		lbls := make([]mimirpb.LabelAdapter, 0, len(tags)+2) // An additional one for __name__, and one for internal label
		lbls = append(lbls, mimirpb.LabelAdapter{
			Name:  labels.MetricName,
			Value: yoloString(name),
		})
```

pkg/distributor/influxpush/parser.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		returnTs = append(returnTs, mimirpb.PreallocTimeseries{TimeSeries: &ts})
+	}
+
+	return returnTs, nil
+}
+
+// analog of invalidChars = regexp.MustCompile("[^a-zA-Z0-9_]")
+func replaceInvalidChars(bSlice []byte) []byte {
```
colega, line 151:
```suggestion
// replaceInvalidChars is the analog of invalidChars = regexp.MustCompile("[^a-zA-Z0-9_]")
// This function may modify the input slice.
func replaceInvalidChars(bSlice []byte) []byte {
```

pkg/distributor/influxpush/parser.go
alexgreenbank: This didn't quite work. If the initial results of `pt.Name()` was modified (e.g. with a `_` byte prepended) then subsequent calls to `pt.Name()` could return incorrect results.

I've rejigged it to make a copy of the slice returned by `pt.Name()` and then append as necessary to that, keeping it as a `[]byte` rather than converting to a string.

pkg/distributor/influx.go
```diff
@@ -0,0 +1,194 @@
...
+				level.Warn(logger).Log("msg", "push request canceled", "err", err)
+				w.WriteHeader(statusClientClosedRequest)
+				return
+			}
+			if errors.Is(err, influxio.ErrReadLimitExceeded) {
+				// TODO(alexg): One thing we have seen in the past is that telegraf clients send a batch of data
+				// if it is too big they should respond to the 413 below, but if a client doesn't understand this
+				// it just sends the next batch that is even bigger. In the past this has had to be dealt with by
+				// adding rate limits to drop the payloads.
+				level.Warn(logger).Log("msg", "request too large", "err", err, "bytesRead", bytesRead, "maxMsgSize", maxRecvMsgSize)
```
colega, line 112: Is this really a TODO? What is pending to be done here?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## main / unreleased
 
+* [FEATURE] Distributor: Add experimental Influx handler. #10153
```
colega, line 5: Please move this to the Grafana Mimir section below.

alexgreenbank: Done.

pkg/distributor/influx.go
alexgreenbank: Removed the TODO. This is more an operational issue, something that I plan on taking a look at during the dev/ops testing and will add it to the epic.

When the `telegraf` sends a payload that is too big it just tries it again a bit later with even more data, and so the problem never seems to fix itself.

We need to documented method/runbook for dealing with this situation (which is easy to trigger given we can control the maximum payload size we accept) otherwise it never resolves itself. We need to test how the `telegraf` agent responds to different return codes, or come up with another plan.

For now the code does the right thing and logs the error and rejects the payload.

colega (approved):

PR #8146, opened 2024-05-15
Introduce global, per-tenant flags and interval to evaluation threshold to control rule evaluation concurrency

gotjosh:
#### What this PR does

This change introduces two flags that control if rule evaluation concurrently.

First, you have a flag that controls the total amount of rules that can be running concurrently at any given time per ruler replica with:

`-ruler.max-global-rule-evaluation-concurrency`

Then, this is paired with `-ruler.max-concurrent-rule-evaluations-per-tenant` to control the amount of rules a single tenant is allowed to have concurrently. By default, this is `4`. However, the behaviour is disabled by default because `-ruler.max-global-rule-evaluation-concurrency` is set to `0` by default.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [x] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

NB: I couldn't think of a way to test this without incurring in a significant effort to set a test it, but I'm happy to spend the time if we think it's worth it.

PR #10157, opened 2024-12-06
Rule Concurrency: Prevent flapping of concurrency

julienduchesne:
#### What this PR does

Iterates on https://github.com/grafana/mimir/pull/8146

The `isGroupAtRisk` function only uses the group's last evaluation time as a metric 
However, if the concurrency of the group causes the group's eval time to lower to less than the threshold, this will flap between enabling concurrency and disabling it on every run

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/rule_concurrency.go
+++ b/pkg/ruler/rule_concurrency.go
@@ -5,6 +5,7 @@ package ruler
 import (
 	"context"
 	"sync"
+	"time"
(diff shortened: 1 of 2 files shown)


julienduchesne:
https://github.com/grafana/mimir/pull/10189 is better

PR #10158, opened 2024-12-06
Revert "Early Request Validation in Query Frontend"

francoposa:
Reverts grafana/mimir#10093 - incompatibility did not get caught because it was not rebased


francoposa:
rolled forward with  https://github.com/grafana/mimir/pull/10159 instead

PR #10185, opened 2024-12-09
fix(ha_tracker): track when there is a change during merge of ReplicaDesc Components

NickAnge:
## Issue

We observed that the number of `gossipMessages` for the `ha_tracker key` in the memberList KVStore was significantly higher compared to other keys. For example, after 5 minutes of running Mimir locally with 2 distributors and 2 Prom-HA instances, we had:

**Keys**
- ha_tracker: ~24k messages
- alertmanagers/alertmanager: ~987 messages (the next highest)

This is causing several issues:
- A high number of messages being broadcasted between Mimir components, requiring more bandwidth.
- Unnecessary changes in the versions of the keys (ha_tracker).

Screenshot 2024-12-09 at 14 24 06

## Solution

We have observed that this is happening because of how the Merge function of ReplicaDesc was behaving. In order to check if there was a change or not it was comparing the two mergeable objects. However some gossip messages can have the same values and not have any change. So the *r != *other

For that reason we introduced a boolean variable to check if there was change. This reduced by a lot the number of public gossip messages it is shown below

Screenshot 2024-12-09 at 14 24 45

Screenshot 2024-12-09 at 14 24 56

Related: https://github.com/grafana/mimir-squad/issues/2653

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@
 * [BUGFIX] Querier: Fix stddev+stdvar aggregations to treat Infinity consistently. #9844
 * [BUGFIX] Ingester: Chunks could have one unnecessary zero byte at the end. #9844
 * [BUGFIX] OTLP receiver: Preserve colons and combine multiple consecutive underscores into one when generating metric names in suffix adding mode (`-distributor.otel-metric-suffixes-enabled`). #10075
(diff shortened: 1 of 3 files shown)


dimitarvdimitrov (approved):
LGTM, nice job spotting it. I think i remember the comment which suggested we remove `changed` 👀 😄

PR #10189, opened 2024-12-09
Rule Concurrency: Prevent flapping of concurrency

julienduchesne:
#### What this PR does

Iterates on https://github.com/grafana/mimir/pull/8146

The `isGroupAtRisk` function only uses the group's last evaluation time as a metric 
However, if the concurrency of the group causes the group's eval time to lower to less than the threshold, this will flap between enabling concurrency and disabling it on every run

In this PR, a condition is added to also sum up the last evaluation time of each rule to compare against the threshold

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,12 @@
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
 * [ENHANCEMENT] Distributor: OTLP receiver now converts also metric metadata. See also https://github.com/prometheus/prometheus/pull/15416. #10168
 * [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145
-* [CHANGE] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
+* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
 * [BUGFIX] Ruler: fix indeterminate rules being always run concurrently (instead of never) when `-ruler.max-independent-rule-evaluation-concurrency` is set. https://github.com/prometheus/prometheus/pull/15560 #10258
 * [BUGFIX] PromQL: Fix various UTF-8 bugs related to quoting. https://github.com/prometheus/prometheus/pull/15531 #10258
-* [BUGFIX] Ruler: Prevent flapping (enabled/disabled) of the concurrency feature #10189
+* [BUGFIX] Ruler: Fixed an issue when using the experimental `-ruler.max-independent-rule-evaluation-concurrency` feature, where if a rule group was eligible for concurrency, it would flap between running concurrently or not based on the time it took after running concurrently. #9726 #10189
 
 ### Mixin
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
 * [ENHANCEMENT] Distributor: OTLP receiver now converts also metric metadata. See also https://github.com/prometheus/prometheus/pull/15416. #10168
 * [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145
-* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
+* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evaluation duration of a rule group regardless of concurrency #10189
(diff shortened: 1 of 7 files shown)


pkg/ruler/rule_concurrency.go
```diff
@@ -190,9 +190,19 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
...
+	runtimeThreshold := interval * c.thresholdRuleConcurrency / 100
 
-	return lastEvaluation >= interval*c.thresholdRuleConcurrency/100
+	if lastEvaluation := group.GetEvaluationTime().Seconds(); lastEvaluation >= runtimeThreshold {
+		return true
+	}
+
+	// Sum up the total duration of all rules in the group to determine if the group is at risk if rules were all run sequentially.
+	var ruleTotalDuration float64
+	for _, rule := range group.Rules() {
```
gotjosh, line 201:
```suggestion
	for _, r := range group.Rules() {
```

or something similar to avoid shadowing the rule package.

gotjosh (commented):
This looks good given it fixes the immediately problem -- However, I'd prefer if we have 2 fields for the runtime - one without concurrency and one with concurrency that we'd only update once the group finishes evaluating. 

I think this will make the runtimes more deterministic and it'll also help us see the difference between expected runtime vs actual runtime (if we expose it as a metric).

The complex part is that we need to do that in Prometheus.

gotjosh:
We also discussed offline, and we're going to implement this change directly in Prometheus. We agreed that we should two things:

1. Keep an atomic with the total runtime of the rules of the group (not how long the group actually took the evaluate) then we can use that directly in order to determine wether this group is eligible for concurrency or not.
2. Add a gauge pretty similar to `prometheus_rule_group_last_duration_seconds` so that we can actually tell when a rule is taking advantage of concurrency.

gotjosh:
Can you please add a changelog entry and make sure you map the newly introduced metric in https://github.com/grafana/mimir/blob/main/pkg/ruler/manager_metrics.go?

julienduchesne:
> Can you please add a changelog entry and make sure you map the newly introduced metric in https://github.com/grafana/mimir/blob/main/pkg/ruler/manager_metrics.go?

Done!

pkg/ruler/manager_metrics.go
```diff
@@ -156,6 +164,7 @@ func (m *ManagerMetrics) Collect(out chan<- prometheus.Metric) {
 	data.SendSumOfGaugesPerTenant(out, m.GroupInterval, "prometheus_rule_group_interval_seconds", dskit_metrics.WithLabels("rule_group"))
 	data.SendSumOfGaugesPerTenant(out, m.GroupLastEvalTime, "prometheus_rule_group_last_evaluation_timestamp_seconds", dskit_metrics.WithLabels("rule_group"))
 	data.SendSumOfGaugesPerTenant(out, m.GroupLastDuration, "prometheus_rule_group_last_duration_seconds", dskit_metrics.WithLabels("rule_group"))
+	data.SendSumOfGaugesPerTenant(out, m.GroupLastRuleDurationSum, "cortex_prometheus_rule_group_last_rule_duration_sum_seconds", dskit_metrics.WithLabels("rule_group"))
```
gotjosh, line 167:
```suggestion
	data.SendSumOfGaugesPerTenant(out, m.GroupLastRuleDurationSum, "prometheus_rule_group_last_rule_duration_sum_seconds", dskit_metrics.WithLabels("rule_group"))
```

gotjosh: You need the original metric name here so that we can gather the data from it - you've already renamed it in the collector up top.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## main / unreleased
+
+### Grafana Mimir
+
+* [CHANGE] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
```
gotjosh, line 7:
```suggestion
* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
```

gotjosh: Technically an enhancement.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+* [BUGFIX] Ruler: Prevent flapping (enabled/disabled) of the concurrency feature #10189
```
gotjosh, line 8:
```suggestion
* [BUGFIX] Ruler: Fixed an issue when using the experimental `-ruler.max-independent-rule-evaluation-concurrency` feature, where if a rule group was eligible for concurrency, it would flap between running concurrently or not based on the time it took after running concurrently. #9726 #10189
```

go.mod
```diff
@@ -285,7 +285,8 @@ require (
 )
 
 // Using a fork of Prometheus with Mimir-specific changes.
-replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v0.0.0-20241210170917-0a0a41616520
+// TODO: THIS IS A BRANCH, REVERT TO MAIN BEFORE MERGING THIS PR
+replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v0.0.0-20241213165041-26afe3d921cd
```
gotjosh, line 289: Please remember to update mimir prometheus before merging this PR - update it in a separate PR.

gotjosh (approved):
LGTM

Approving so that you don't need another review from me, please make sure you update mimir prometheus before merging this PR.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 * [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
+* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evalation duration of a rule group regardless of concurrency #10189
```
tacole02, line 13:
```suggestion
* [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evaluation duration of a rule group regardless of concurrency #10189
```

julienduchesne: Thanks!

PR #10196, opened 2024-12-09
UsageTracker: no groups experiment

colega:
Why try to find something somewhere in a group if we can jump directly into that place in the group? This increases the chances of finding the right element, but might be just more expensive because of the slice bounds check or whatever?


colega:
I think this is worth testing, but closing until we start working on this again.

PR #10201, opened 2024-12-10
Chore: use strconv.ParseInt() with 32bit when parsing 32bit numbers

pracucci:
#### What this PR does

We have a static security analysis tool which is complaining with "Potential Integer overflow made by strconv.Atoi result conversion to int16/32" in few places. In this PR I'm changing such parsing to use `strconv.ParseInt()` instead.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/metadata_handler.go
+++ b/pkg/querier/metadata_handler.go
@@ -74,8 +74,8 @@ func NewMetadataHandler(m MetadataSupplier) http.Handler {
 
 		metric := r.FormValue("metric")
 		req := &client.MetricsMetadataRequest{
-			Limit:          int32(limit),
-			LimitPerMetric: int32(limitPerMetric),
+			Limit:          limit,
+			LimitPerMetric: limitPerMetric,
 			Metric:         metric,
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/querier/metadata_handler.go
```diff
@@ -48,24 +48,28 @@ type metadataErrorResult struct {
 // Mimir for a given tenant. It is kept and returned as a set.
 func NewMetadataHandler(m MetadataSupplier) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		limit := -1
+		limit := int32(-1)
```
pracucci, line 51: Note to reviewers: changed the "base" data type to be `int32` so that casting now is always to bigger ones (except for the value returned by `strconv.ParseInt()` which is a explicitly parsed as 32bit).

pkg/querier/metadata_handler.go
```diff
@@ -48,24 +48,30 @@ type metadataErrorResult struct {
...
 				util.WriteJSONResponse(w, metadataErrorResult{Status: statusError, Error: "limit_per_metric must be a number"})
 				return
 			}
+
+			limitPerMetric = int32(parsed)
 		}
+
 		metric := r.FormValue("metric")
 		req := &client.MetricsMetadataRequest{
 			Limit:          int32(limit),
```
colega, line 77: So this is a noop casting now? Can we remove it?

PR #4443, opened 2023-03-09
math: Convert Min and Max to be generic

aknuds1:
#### What this PR does
Convert `pkg/util/math.Min` and `pkg/util/math.Max` to be generic, so they can operate on any arguments fulfilling `constraints.Ordered`. Also removing non-generic min and max implementations.

#### Checklist

- [na] Tests updated
- [na] Documentation added
- [na] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #10202, opened 2024-12-10
Replace min/max helpers with built-in min/max

Juneezee:
#### What this PR does

Follow-up of https://github.com/grafana/mimir/pull/4443#issuecomment-1462317691.

We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max

#### Checklist

- [N/A] Tests updated.
- [N/A] Documentation added.
- [N/A] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.


narqo (approved):
Works for me 🔥

PR #10208, opened 2024-12-11
Ops: Also set server.grpc-max-send-msg-size-bytes on ruler-queriers

jhesketh:
This should match what the ruler-query-frontend is configured for.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir-tests/test-ruler-remote-evaluation-generated.yaml
+++ b/operations/mimir-tests/test-ruler-remote-evaluation-generated.yaml
@@ -1161,6 +1161,7 @@ spec:
         - -query-frontend.results-cache.memcached.timeout=500ms
         - -query-frontend.scheduler-address=ruler-query-scheduler-discovery.default.svc.cluster.local.:9095
         - -runtime-config.file=/etc/mimir/overrides.yaml
+        - -server.grpc-max-recv-msg-size-bytes=104857600
         - -server.grpc-max-send-msg-size-bytes=104857600
         - -server.grpc.keepalive.max-connection-age=30s
         - -server.grpc.keepalive.max-connection-age-grace=5m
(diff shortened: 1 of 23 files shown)


operations/mimir/ruler-remote-evaluation.libsonnet
```diff
@@ -70,6 +74,7 @@
 
       // The ruler-query-frontend sends the query response back to the ruler.
       'server.grpc-max-send-msg-size-bytes': $._config.ruler_remote_evaluation_max_query_response_size_bytes,
+      'ruler.client.grpc-max-recv-msg-size': $._config.ruler_remote_evaluation_max_query_response_size_bytes,
```
charleskorn, line 77: Queriers call an RPC on query-frontends to send the result, so we need to use the `server` config for this:

```suggestion
      'server.grpc-max-recv-msg-size-bytes': $._config.ruler_remote_evaluation_max_query_response_size_bytes,
```

We should probably also add a comment above to separate this from the comment about ruler-query-frontend sending the response to rulers.

operations/mimir/ruler-remote-evaluation.libsonnet
```diff
@@ -33,7 +33,11 @@
     $.querier_args +
     $.querierUseQuerySchedulerArgs(rulerQuerySchedulerName) + {
       'querier.max-concurrent': $._config.ruler_querier_max_concurrency,
-    },
+    } + if !useRulerQueryFrontend then {} else {
+    // The ruler-querier-frontend send a query request to the ruler-query
+    'server.grpc-max-send-msg-size-bytes': $._config.ruler_remote_evaluation_max_query_response_size_bytes,
```
charleskorn, line 38: This should be this flag I believe:

```suggestion
    'querier.frontend-client.grpc-max-send-msg-size': $._config.ruler_remote_evaluation_max_query_response_size_bytes,
```

charleskorn (approved):

operations/mimir-tests/test-ruler-remote-evaluation-generated.yaml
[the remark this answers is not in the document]
jhesketh: That's already set to the 104857600 value. I'm not sure where tbh.

PR #10188, opened 2024-12-09
Vendor update mimir-prometheus at af2a1cb10c89de496cd4309ac532624b34112d74

krajorama:
#### What this PR does

This is upstream at
https://github.com/prometheus/prometheus/tree/af2a1cb10c89de496cd4309ac532624b34112d74

Changes listed in
https://github.com/grafana/mimir-prometheus/pull/787

Fixed query sharding test `TestQuerySharding_FunctionCorrectness` failing due to https://github.com/prometheus/prometheus/pull/15479.

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10212, opened 2024-12-11
chore(otel_test): rework test to prep for keep id resource attributes

krajorama:
#### What this PR does

Rework the test to be more explicit and support testing for https://github.com/prometheus/prometheus/pull/15448 that's coming next.

Related to #10188 
Related to https://github.com/prometheus/prometheus/pull/15448

#### Checklist

- [X] Tests updated.
- N/A Documentation added.
- N/A `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.

### DIFF
--- a/pkg/distributor/otel_test.go
+++ b/pkg/distributor/otel_test.go
@@ -53,11 +53,11 @@ func TestOTelMetricsToTimeSeries(t *testing.T) {
 		"service.instance.id": "service ID",
 		"existent-attr":       "resource value",
 		// This one is for testing conflict with metric attribute.
-		"metric-attr": "resource value conflict",
+		"metric-attr": "resource value",
 		// This one is for testing conflict with auto-generated job attribute.
-		"job": "resource value conflict",
+		"job": "resource value",
 		// This one is for testing conflict with auto-generated instance attribute.
-		"instance": "resource value conflict",
+		"instance": "resource value",
 	}
 
 	md := pmetric.NewMetrics()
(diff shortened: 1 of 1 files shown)


jesusvazquez (approved):

pkg/distributor/otel_test.go
```diff
@@ -50,41 +49,16 @@ func TestOTelMetricsToTimeSeries(t *testing.T) {
 	}, []string{tenantID, "group"})
 	resourceAttrs := map[string]string{
 		"service.name":        "service name",
+		"service.namespace":   "service namespace",
 		"service.instance.id": "service ID",
 		"existent-attr":       "resource value",
 		// This one is for testing conflict with metric attribute.
-		"metric-attr": "resource value",
+		"metric-attr": "resource value conflict",
```
aknuds1, line 56: Is there any practical reason for changing the resource attribute values? I mean, the conflicts are on the attribute names, so I don't think the values need to signal any conflict? The "resource value" part already takes care of signaling that the value comes from a resource attribute.

I also think it ends up giving the wrong impression in test expectations, as it looks as if there's a conflict in `target_info` when there is none.

aknuds1 (commented):
Generally looks good, but I don't agree with the changes to the resource attribute values, since I think they give the wrong impression.

pkg/distributor/otel_test.go
krajorama: Ack, reverted that bit, it wasn't obvious to me what was going on, my change was making it a bit more explicit, but I see your point.

aknuds1: Thanks, that's the rationale I suspected :)

aknuds1 (approved):
LGTM, thanks

PR #10228, opened 2024-12-12
lazyquerier: deep copy hints

dimitarvdimitrov:
#### What this PR does

The promQL engine can modify the grouping that it passes to us after calling. First the LazyQuerier is called in [`populateSeries`](https://github.com/grafana/mimir/blob/76ab843e3bb3d6517fcd1ee47e919bd9d107ef0d/vendor/github.com/prometheus/prometheus/promql/engine.go#L953-L976), and then the grouping is changed (via sorting) when evaluating the [aggregate expression](https://github.com/grafana/mimir/blob/76ab843e3bb3d6517fcd1ee47e919bd9d107ef0d/vendor/github.com/prometheus/prometheus/promql/engine.go#L1571-L1573)
#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (approved):

pkg/storage/lazyquery/lazyquery_test.go
```diff
@@ -0,0 +1,48 @@
...
+	typ := originalVal.Type()
+	for i := 0; i < typ.NumField(); i++ {
+		originalField := originalVal.Field(i)
+		copiedField := copiedVal.Field(i)
+
+		// Check if values are equal
+		require.Equal(t, originalField.Interface(), copiedField.Interface(), "Field %s has different values", typ.Field(i).Name)
+
+		// For reference types, ensure they point to different memory
+		if originalField.Kind() == reflect.Slice ||
```
zenador, line 40: [nit] Can we simplify this a bit, like use a match statement and/or merge with the condition in the nested if?

zenador (approved):
LGTM aside from small nit

Issue #10067, opened 2024-12-02
Mimir Query Enginee


jhesketh:
We're building a new query engine to improve the peak memory consumption of queriers.

The work has already progressed and is fairly mature. A good description of the new engine is found here:
https://github.com/grafana/mimir/blob/main/pkg/streamingpromql/README.md

This issue is to track the PRs and large number of changes involved. It will be an index instead of maintaining a list of PRs in the CHANGELOG.

# Main / Unreleased

* #9367 
* #9368
* #9398 
* #9399 
* #9403 
* #9417 
* #9418 
* #9419 
* #9420 
* #9482 
* #9504
* #9505 
* #9507 
* #9518 
* #9531 
* #9532 
* #9533 
* #9553 
* #9558 
* #9588 
* #9589 
* #9639 
* #9641 
* #9642 
* #9651 
* #9664 
* #9681 
* #9717 
* #9719 
* #9724 
* #9874 
* #9929 
* #9998 
* #10007 
* #10010 
* #10046 
* #10047 
* #10048 
* #10050

# 2.14.0

```
* [FEATURE] Querier: add experimental streaming PromQL engine, enabled with `-querier.query-engine=mimir`. #8422 #8430 #8454 #8455 #8360 #8490 #8508 #8577 #8660 #8671 #8677 #8747 #8850 #8872 #8838 #8911 #8909 #8923 #8924 #8925 #8932 #8933 #8934 #8962 #8986 #8993 #8995 #9008 #9017 #9018 #9019 #9120 #9121 #9136 #9139 #9140 #9145 #9191 #9192 #9194 #9196 #9201 #9212 #9225 #9260 #9272 #9277 #9278 #9280 #9281 #9342 #9343 #9371 #9859 #9858
```

# 2.13.0

```
* [FEATURE] Querier: add experimental streaming PromQL engine, enabled with `-querier.promql-engine=mimir`. #7693 #7898 #7899 #8023 #8058 #8096 #8121 #8197 #8230 #8247 #8270 #8276 #8277 #8291 #8303 #8340 #8256 #8348
```

PR #10232, opened 2024-12-13
MQE: track number of processed samples in each query

### LINKED ISSUE #10138: Idea: Measure processed samples on MQE

In the effort of trying to measure throughput in Mimir, we've realized `MQE` does not make processed samples data available as the `PromQL` engine does.

Make `MQE` count samples loaded from storage and expose it to `mimir-stats` so we can get it to a response header for each query.

I'm happy to listen to any alternatives. The objective is to measure throughput in the best possible way.

Related work on measuring Throughput in Mimir has been done in #10103, #9985, #7966 and [here](https://github.com/grafana/dskit/pull/619).

Small (<= 1 month dev)

_No response_

### Proposer?

_No response_

charleskorn:
#### What this PR does

This PR adds support for tracking the number of samples processed in a query evaluated by MQE.

Resolves #10138

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2824,6 +2824,7 @@ func TestQueryStats(t *testing.T) {
 			start_series  0 1 _ _ _ _ _ _ _ _ _
 			end_series    _ _ _ _ _ 5 6 7 8 9 10
 			sparse_series 0 _ _ _ _ _ _ 7 _ _ _
+			stale_series  0 1 2 3 4 5 stale 7 8 9 10
 			nan_series    NaN NaN NaN NaN NaN NaN NaN NaN NaN NaN NaN
 			native_histogram_series {{schema:0 sum:2 count:4 buckets:[1 2 1]}} {{sum:2 count:4 buckets:[1 2 1]}}
 	`)
@@ -2869,6 +2870,10 @@ func TestQueryStats(t *testing.T) {
 			expr:                 `sparse_series{}`,
 			expectedTotalSamples: 5 + 4, // 5 for first point at T=0, and 4 for second point at T=7
 		},
+		"instant vector selector with stale marker": {
+			expr:                 `stale_series{}`,
+			expectedTotalSamples: 10, // Instant vector selectors ignore stale markers.
+		},
 
 		"raw range vector selector with single point": {
 			expr:                 `dense_series[45s]`,
@@ -2900,6 +2905,10 @@ func TestQueryStats(t *testing.T) {
 		"range vector selector where range overlaps previous step's range": {
 			expr:                 `sum_over_time(dense_series{}[1m30s])`,
 			expectedTotalSamples: 21, // Each step except the first selects two points.
+		},
+		"range vector selector with stale marker": {
+			expr:                 `count_over_time(stale_series{}[1m30s])`,
+			expectedTotalSamples: 19, // Each step except the first selects two points. Range vector selectors ignore stale markers.
 		},
 
 		"expression with multiple selectors": {
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2635,14 +2635,14 @@ func runMixedMetricsTests(t *testing.T, expressions []string, pointsPerSeries in
 				q, err := prometheusEngine.NewRangeQuery(context.Background(), storage, nil, expr, start, end, tr.interval)
 				require.NoError(t, err)
 				defer q.Close()
-				expectedResults := q.Exec(context.Background())
+				prometheusResults := q.Exec(context.Background())
 
 				q, err = mimirEngine.NewRangeQuery(context.Background(), storage, nil, expr, start, end, tr.interval)
 				require.NoError(t, err)
 				defer q.Close()
 				mimirResults := q.Exec(context.Background())
 
-				testutils.RequireEqualResults(t, expr, expectedResults, mimirResults, skipAnnotationComparison)
+				testutils.RequireEqualResults(t, expr, prometheusResults, mimirResults, skipAnnotationComparison)
 			})
 		}
 	}
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -2824,6 +2824,8 @@ func TestQueryStats(t *testing.T) {
 			start_series  0 1 _ _ _ _ _ _ _ _ _
 			end_series    _ _ _ _ _ 5 6 7 8 9 10
 			sparse_series 0 _ _ _ _ _ _ 7 _ _ _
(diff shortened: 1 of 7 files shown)


tinitiuset (approved):
Thank you for being so quick on this. Looks good to me.

pkg/streamingpromql/operators/selectors/instant_vector_selector.go
```diff
@@ -119,6 +120,8 @@ func (v *InstantVectorSelector) NextSeries(ctx context.Context) (types.InstantVe
 			continue
 		}
 
+		v.Stats.TotalSamples++
```
jhesketh, line 123: Have we benchmarked this change?

One alternate would be to do `v.Stats.TotalSamples += v.Selector.TimeRange.EndT / v.Selector.TimeRange.IntervalMilliseconds` (probably with `floor` and an offset for the start or similar)

Or is this done here to catch stats for errored queries? I'm not sure if there would be value in that.

If not, alternatively we could look at the `len` of the returned data.

pkg/streamingpromql/types/stats.go
```diff
@@ -0,0 +1,18 @@
...
+// Provenance-includes-location: https://github.com/prometheus/prometheus/blob/main/util/stats/query_stats.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Prometheus Authors
+
+package types
+
+// QueryStats tracks statistics about the execution of a single query.
+//
+// It is not safe to use this type from multiple goroutines simultaneously.
+type QueryStats struct {
```
jhesketh, line 11: Why create our own struct instead of using stats.QuerySamples and passing that to the appropriate operators?

It would also make it easier to implement `TotalSamplesPerStep` if we want to support that too (which I think we do)

pkg/streamingpromql/engine_test.go
```diff
@@ -2807,3 +2807,112 @@ func TestCompareVariousMixedMetricsComparisonOps(t *testing.T) {
...
+
+	start := timestamp.Time(0)
+	end := start.Add(10 * time.Minute)
+
+	storage := promqltest.LoadedStorage(t, `
+		load 1m
+			dense_series  0 1 2 3 4 5 6 7 8 9 10
+			start_series  0 1 _ _ _ _ _ _ _ _ _
+			end_series    _ _ _ _ _ 5 6 7 8 9 10
+			sparse_series 0 _ _ _ _ _ _ 7 _ _ _
```
jhesketh, line 2826: should have some NH's to test too

Also some stale and NaN's etc.

tinitiuset: While addressing this I found out NH were not being counted to samples as PromQL does. I have adjusted the calculation.

pkg/streamingpromql/operators/selectors/instant_vector_selector.go
charleskorn: > If not, alternatively we could look at the `len` of the returned data.

This is a much better idea than what I did 🙂 I've changed this in 31cefce.

pkg/streamingpromql/engine_test.go
charleskorn: I've added a test case for stale markers in 14a0b7c.

pkg/streamingpromql/engine_test.go
```diff
@@ -2807,3 +2807,112 @@ func TestCompareVariousMixedMetricsComparisonOps(t *testing.T) {
...
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			prometheusCount := runQueryAndGetTotalSamples(t, prometheusEngine, testCase.expr, testCase.isInstantQuery)
+			require.Equal(t, testCase.expectedTotalSamples, prometheusCount, "invalid test case: expected samples does not match value from Prometheus' engine")
+
+			mimirCount := runQueryAndGetTotalSamples(t, mimirEngine, testCase.expr, testCase.isInstantQuery)
+			require.Equal(t, testCase.expectedTotalSamples, mimirCount)
```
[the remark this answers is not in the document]
charleskorn: This is currently difficult due to the optimisation in  https://github.com/prometheus/prometheus/pull/14097, as Prometheus' engine sometimes skips loading data for histograms if it's not needed. MQE does not yet have the same optimisation, so there are some expected differences in the total sample count from the two engines in some cases. 

Given the tests in `TestQueryStats`, and the fact the statistics are informational and may differ between engines in the future due to other optimisations, I'm tempted to leave this as-is.

Thoughts?

pkg/streamingpromql/types/stats.go
charleskorn: > Why create our own struct instead of using stats.QuerySamples and passing that to the appropriate operators?

Given we don't support anything other than `TotalSamples` in MQE, I wanted to make this clear in the code by using a struct that only had a field for `TotalSamples`.

> It would also make it easier to implement `TotalSamplesPerStep` if we want to support that too (which I think we do)

I don't think we want to do this unless there's a specific need for it - per-step stats are considered experimental and disabled by default in Prometheus, and are not possible to enable in Mimir as far as I can see. The docs for this also state that the value for each step should be the same as if the query was run as an instant query, so anyone who wanted this information could run the query as an instant query for the step(s) they're interested in.

pkg/streamingpromql/operators/selectors/instant_vector_selector.go
charleskorn: >Have we benchmarked this change?

With the latest changes in this PR, there is no statistically significant difference with this PR compared to `main`.

pkg/streamingpromql/engine_test.go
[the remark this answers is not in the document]
jhesketh: Happy to leave it out if that's the case.

It might be interesting to see some much larger queries/time ranges etc to see if we are returning consistent results. We could perhaps use the same data generated from the benchmarks to create some large queries (of just floats since NH will be different). Then have a flag to `RequireEqualResults` to compare them etc.

This isn't a blocker.

jhesketh (commented):
I don't really like how we have to iterate through the buffer vs counting the samples on filling the buffer, but otherwise lgtm. Is it possible to see a benchmark of range-vectors of histograms before approving?

charleskorn:
> I don't really like how we have to iterate through the buffer vs counting the samples on filling the buffer, but otherwise lgtm. Is it possible to see a benchmark of range-vectors of histograms before approving?

Benchmarks show there is no statistically significant difference with this PR compared to `main` - the extra work is insignificant compared to everything else that needs to be done.

jhesketh (approved):
Thanks for checking :rocket:

PR #10235, opened 2024-12-13
MQE: fix issues with vector/vector binary comparsion operations

charleskorn:
#### What this PR does

This PR fixes two issues with the implementation of comparison operations between two vectors without the `bool` modifier in MQE:

* Queries would return incorrect results if the left side contained series with different metric names, and all results would be merged into a single series rather than returned in separate series for each metric name
* Queries would incorrectly fail with a conflict error if multiple series on the left side matched the same series on the right side, those left side series had points at the same timestamp, and one or none of those points remained after applying the comparison operation

Changes in latency and peak memory consumption of affected benchmarks are within the margin of error of our benchmarks.

I'd recommend reviewing each commit separately.

#10067

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/benchmarks/benchmarks.go
+++ b/pkg/streamingpromql/benchmarks/benchmarks.go
@@ -159,6 +159,9 @@ func TestCases(metricSizes []int) []BenchCase {
 			Expr: "nh_X / a_X",
 		},
 		{
+			Expr: "a_X == b_X",
+		},
+		{
 			Expr: "2 * a_X",
 		},
 		{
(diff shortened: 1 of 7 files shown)


jhesketh (approved):
lgtm, but I am trying to think if we could add some extra tests around this and/or the annotation?

charleskorn:
> lgtm, but I am trying to think if we could add some extra tests around this and/or the annotation?

Josh and I discussed this offline - this was a misunderstanding about some of the error messages.

PR #10259, opened 2024-12-16
feat(ruler): add flags for the query frontend retries

julienduchesne:
Currently, Mimir is hardcoded at 3 retries and 100ms as a minimum retry time. Could be useful to configure this

- [x] Tests updated.
- [ ] Documentation added.


chencs:
Should the config flags be experimental for now?

dimitarvdimitrov (commented):
I assume you're adding those flags so you can experiment with different retry strategies. I'm also leaning towards making these experimental and removing the once we know what's the better setting.

tacole02 (approved):

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2987,6 +2987,12 @@ Usage of ./cmd/mimir/mimir:
     	Override the default minimum TLS version. Allowed values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13
   -ruler.query-frontend.grpc-client-config.tls-server-name string
     	Override the expected name on the server certificate.
+  -ruler.query-frontend.max-retries int
+    	[experimental] Maximum number of retries for a single request. (default 3)
```
dimitarvdimitrov, line 2991: this is actually max attempts, right? So it includes the first try. That's my understanding from reading the dskit code. I think we should clarify the CLI flag then.

dimitarvdimitrov (approved):
left a suggest about the wording on the max retries, but otherwise LGTM, I don't need to see it again

PR #10261, opened 2024-12-17
MQE: Fix panic in loading too many samples

jhesketh:
#### What this PR does

This PR solves two bugs:

### Bug 1:

BucketedPool would create pools in powers of two up to a set `maxSize`. That is, if `maxSize` isn't a power of two itself,
then the maximum bucket would be less than the maxSize.

However, when requesting a slice from the pool we were only checking against the maxSize, and not whether a bucket existed for that size. Now, instead, calculate the bucketIndex and check if that exists before using it.

This would cause panic's in range vectors of more than 65,536 samples.

This problem was introduced in https://github.com/grafana/mimir/pull/10199

### Bug 2:

If a requested slice size exceeded the size of the pools it would just be returned as the requested size. This caused a bug in the buffers which require the slices to be powers of two.

This PR ensures that the returned slice is always of power two. This means we will return a slice greater than the `MaxExpectedPointsPerSeries`, but it won't be from a pool (and similarly won't be returned to a pool). We are protected by the `MemoryConsumptionTracker` for any unreasonable requests.

This was introduced in https://github.com/grafana/mimir/issues/9664

Originally this would cause panic's in sub-queries selecting more than 65,536 samples, but since https://github.com/grafana/mimir/pull/10199 it returns just errors.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/types/limiting_pool.go
+++ b/pkg/streamingpromql/types/limiting_pool.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	MaxExpectedPointsPerSeries = 100_000 // There's not too much science behind this number: 100000 points allows for a point per minute for just under 70 days.
+	MaxExpectedPointsPerSeries = 131_072 // There's not too much science behind this number: 100,000 points allows for a point per minute for just under 70 days. Then we use the next power of two.
 
 	// Treat a native histogram sample as equivalent to this many float samples when considering max in-memory bytes limit.
 	// Keep in mind that float sample = timestamp + float value, so 5x this is equivalent to five timestamps and five floats.
--- a/pkg/util/pool/bucketed_pool.go
+++ b/pkg/util/pool/bucketed_pool.go
@@ -54,23 +54,10 @@ func (p *BucketedPool[T, E]) Get(size int) T {
 		return nil
 	}
 
-	// Enforce Go's maximum slice size limitation.
-	// Go slices are limited to (1<<31)-1 elements due to runtime constraints.
-	const maxSliceSize = (1 << 31) - 1
-	if size > maxSliceSize {
-		panic(fmt.Sprintf("Cannot allocate slice with size (%d) as it exceeds Go's maximum slice size (%d)", size, maxSliceSize))
-	}
-
 	bucketIndex := bits.Len(uint(size - 1))
 
 	// If bucketIndex exceeds the number of available buckets, return a slice of the next power of two.
 	if bucketIndex >= len(p.buckets) {
-		// Check for wrap around case when bucketIndex equals or exceeds the bit size limit.
-		if bucketIndex >= bits.UintSize {
-			// This should not occur due to the above slice size check.
-			panic(fmt.Sprintf("Cannot allocate slice with size (%d) as it exceeds the system limit", size))
-		}
-
 		nextPowerOfTwo := 1 << bucketIndex
 		return p.make(nextPowerOfTwo)
 	}
--- a/pkg/util/pool/bucketed_pool.go
+++ b/pkg/util/pool/bucketed_pool.go
@@ -51,11 +51,11 @@ func (p *BucketedPool[T, E]) Get(size int) T {
 		return nil
 	}
 
-	if uint(size) > p.maxSize {
+	bucketIndex := bits.Len(uint(size - 1))
+	if bucketIndex >= len(p.buckets) {
 		return p.make(size)
 	}
 
-	bucketIndex := bits.Len(uint(size - 1))
 	s := p.buckets[bucketIndex].Get()
 
 	if s == nil {
--- a/pkg/util/pool/bucketed_pool.go
+++ b/pkg/util/pool/bucketed_pool.go
@@ -44,6 +44,7 @@ func NewBucketedPool[T ~[]E, E any](maxSize uint, makeFunc func(int) T) *Buckete
 // Get returns a new slice with capacity greater than or equal to size.
 // If no bucket large enough exists, a slice larger than the requested size
 // of the next power of two is returned.
+// Get guarantees the resulting slice always has a capacity in power of twos.
 func (p *BucketedPool[T, E]) Get(size int) T {
 	if size < 0 {
 		panic(fmt.Sprintf("BucketedPool.Get with negative size %v", size))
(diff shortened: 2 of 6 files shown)


pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	MaxExpectedPointsPerSeries = 100_000 // There's not too much science behind this number: 100000 points allows for a point per minute for just under 70 days.
+	MaxExpectedPointsPerSeries = 131_072 // There's not too much science behind this number: 100,000 points allows for a point per minute for just under 70 days. Then we use the next power of two.
```
pracucci, line 16: I recommend a unit test just to ensure this number is a power of 2.

jhesketh:
(this will need backporting to r321)

pracucci (approved):

pkg/util/pool/bucketed_pool.go
```diff
@@ -51,11 +53,12 @@ func (p *BucketedPool[T, E]) Get(size int) T {
 		return nil
 	}
 
-	if uint(size) > p.maxSize {
-		return p.make(size)
+	bucketIndex := bits.Len(uint(size - 1))
+	if bucketIndex >= len(p.buckets) {
+		nextPowerOfTwo := 1 << bucketIndex
```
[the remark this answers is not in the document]
jhesketh: Good catch. I think this would only occur where the requested `size` exceeds the largest slice possible anyway. I'll add a check.

pkg/util/pool/bucketed_pool.go
```diff
@@ -51,11 +54,27 @@ func (p *BucketedPool[T, E]) Get(size int) T {
 		return nil
 	}
 
-	if uint(size) > p.maxSize {
-		return p.make(size)
+	// Enforce Go's maximum slice size limitation.
+	// Go slices are limited to (1<<31)-1 elements due to runtime constraints.
+	const maxSliceSize = (1 << 31) - 1
```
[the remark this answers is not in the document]
jhesketh: Right, makes sense. I think given we panic in both situations it's functionally the same anyway.

pkg/util/pool/bucketed_pool.go
```diff
@@ -76,8 +82,14 @@ func (p *BucketedPool[T, E]) Put(s T) {
 	}
 
 	bucketIndex := bits.Len(size - 1)
+	if bucketIndex >= len(p.buckets) {
+		return // Ignore slices larger than the largest bucket
+	}
```
charleskorn, line 87: Instead of this, should we require that `maxSize` is a power of two in `NewBucketedPool`? I think that would be easier to reason about and less confusing for someone using this type. 

For example, imagine someone created a pool with `maxSize` set to 1000. However, only slices with capacity up to 512 will be pooled, which is likely not what they intended.

pkg/util/pool/bucketed_pool.go
```diff
@@ -76,8 +82,14 @@ func (p *BucketedPool[T, E]) Put(s T) {
...
+
+	// Ignore slices that do not align to the current power of 2
```
charleskorn, line 89: [nit] 

```suggestion
	// Ignore slices with capacity that is not a power of 2
```

pkg/util/pool/bucketed_pool.go
```diff
@@ -42,6 +42,9 @@ func NewBucketedPool[T ~[]E, E any](maxSize uint, makeFunc func(int) T) *Buckete
 }
 
 // Get returns a new slice with capacity greater than or equal to size.
+// If no bucket large enough exists, a slice larger than the requested size
+// of the next power of two is returned.
+// Get guarantees the resulting slice always has a capacity in power of twos.
```
charleskorn, line 47: I think this could be a little clearer:

```suggestion
// The resulting slice always has a capacity that is a power of two.
// If size is greater than maxSize, then a slice is still returned, however it may not be drawn from a pool.
```

pkg/util/pool/bucketed_pool_test.go
```diff
@@ -120,5 +120,60 @@ func TestBucketedPool_PutSliceLargerThanMaximum(t *testing.T) {
...
 	require.NotSame(t, &s1[0], &s2[0])
-	require.Equal(t, 101, cap(s2))
+	require.Equal(t, 128, cap(s2))
+}
+
+func TestBucketedPool_GetSizeCloseToMax(t *testing.T) {
+	maxSize := 100000
+	pool := NewBucketedPool(uint(maxSize), makeFunc)
+
+	// Request a size that triggers the last bucket boundary.
```
charleskorn, line 130: [nit]

```suggestion
	// Request a slice with size that will be drawn from the last bucket in the pool.
```

pkg/util/pool/bucketed_pool_test.go
```diff
@@ -120,5 +120,60 @@ func TestBucketedPool_PutSliceLargerThanMaximum(t *testing.T) {
...
+	s := pool.Get(86401)
+
+	// Check that we still get a slice with the correct size.
+	require.Equal(t, 131072, cap(s))
+	require.Len(t, s, 0)
+}
+
+func TestBucketedPool_AlwaysReturnsPowerOfTwoCapacities(t *testing.T) {
```
charleskorn, line 138: Isn't this covered by `TestBucketedPool_HappyPath`?

pkg/util/pool/bucketed_pool_test.go
```diff
@@ -120,5 +120,60 @@ func TestBucketedPool_PutSliceLargerThanMaximum(t *testing.T) {
...
+		pool.Put(slice)
+	}
+}
+
+func TestBucketedPool_PutSizeCloseToMax(t *testing.T) {
+	maxSize := 100000
+	pool := NewBucketedPool(uint(maxSize), makeFunc)
+
+	// Create a slice with capacity that triggers the upper edge case
+	s := make([]int, 0, 65_000) // 86401 is close to maxSize but not aligned to power of 2
```
charleskorn, line 168: These two comments aren't clear to me: it's not clear why 86401 is mentioned in the second comment, and "upper edge case" isn't clear either. If this is intended to test the case where we're putting a slice into the last bucket in the pool and `maxSize` is not a power of two, then perhaps something along those lines would be clearer?

pkg/util/pool/bucketed_pool.go
```diff
@@ -51,11 +54,14 @@ func (p *BucketedPool[T, E]) Get(size int) T {
 		return nil
 	}
 
-	if uint(size) > p.maxSize {
-		return p.make(size)
+	bucketIndex := bits.Len(uint(size - 1))
+
+	// If bucketIndex exceeds the number of available buckets, return a slice of the next power of two.
```
charleskorn, line 59: [nit]

```suggestion
	// If the requested size is larger than the size of the largest bucket, return a slice of the next power of two greater than or equal to size.
```

charleskorn (commented):
Thanks for finding and fixing these issues!

pkg/util/pool/bucketed_pool.go
jhesketh: We should do both. It's possible for somebody to pass in something unexpected.

PR #9733, opened 2024-10-24
POC: Cost Attribution Proposal 2

ying-jeanne:
#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10269, opened 2024-12-17
Ingester/Distributor: Add support for exporting cost attribution metrics

ying-jeanne:
#### What this PR does
This is the follow up of https://github.com/grafana/mimir/pull/9733, 

The PR intent to export extra attributed metrics in distributor and ingester, in order to get sample received, sample discarded and active_series attributed by cost attribution label.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -455,6 +455,18 @@ overrides_exporter:
...
 [timeseries_unmarshal_caching_optimization_enabled: <boolean> | default = true]
+
+# (experimental) Time interval at which inactive cost attributions are evicted
+# from the counter, ensuring they are not included in the cost attribution
+# cardinality per user limit.
+# CLI flag: -cost-attribution.eviction-interval
+[cost_attribution_eviction_interval: <duration> | default = 20m]
+
+# (experimental) Defines a custom path for the registry. When specified, Mimir
+# will expose cost attribution metrics through this custom path, if not
```
tacole02, line 466:
```suggestion
# exposes cost attribution metrics through this custom path. If not
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -455,6 +455,18 @@ overrides_exporter:
...
+# specified, cost attribution metrics won't be exposed.
```
tacole02, line 467:
```suggestion
# specified, cost attribution metrics aren't exposed.
```

tacole02: We avoid using future tense in the docs.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3569,6 +3581,33 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -querier.active-series-results-max-size-bytes
 [active_series_results_max_size_bytes: <int> | default = 419430400]
 
+# (experimental) Defines labels for cost attribution, applied to metrics like
```
tacole02, line 3584:
```suggestion
# (experimental) Defines labels for cost attribution. Applies to metrics like
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3569,6 +3581,33 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# cortex_distributor_attributed_received_samples_total. Set to an empty string
+# to disable. Example: 'team,service' will produce metrics such as
```
tacole02, line 3586:
```suggestion
# cortex_distributor_attributed_received_samples_total. To disable, set to an empty string.
# For example, 'team,service' produces metrics such as
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3569,6 +3581,33 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+
+# (experimental) Maximum cardinality of cost attribution labels allowed per
+# user.
+# CLI flag: -validation.max-cost-attribution-cardinality-per-user
+[max_cost_attribution_cardinality_per_user: <int> | default = 10000]
+
+# (experimental) Cooldown period for cost attribution labels. Specifies the
+# duration the cost attribution remains in overflow before attempting a reset.
+# If the cardinality remains above the limit after this period, the system will
+# stay in overflow mode and extend the cooldown. Setting this value to 0
```
tacole02, line 3604:
```suggestion
# If the cardinality remains above the limit after this period, the system
# stays in overflow mode and extends the cooldown. Setting this value to 0
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3569,6 +3581,33 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# disables the cooldown, causing the system to continuously check whether the
+# cardinality has dropped below the limit. A reset will occur once the
+# cardinality falls below the limit.
```
tacole02, line 3607:
```suggestion
# cardinality has dropped below the limit. A reset occurs when the
# cardinality falls below the limit.
```

tacole02 (approved):
Thanks for updating the docs! I left a few suggestions.

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/services"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/grafana/mimir/pkg/util/validation"
+)
+
+const (
+	TrackerLabel       = "tracker"
+	TenantLabel        = "tenant"
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit d2eab6b54e76aa986689d8ab2a83b31e15201cea

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	m.Service = services.NewBasicService(nil, m.running, nil).WithName("cost attribution manager")
+	if err := reg.Register(m); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (m *Manager) running(ctx context.Context) error {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 1f392821a3e24ac01813682d28c44c71c62e7814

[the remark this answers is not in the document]
ying-jeanne: addressed 9b4337d14400f3af48fcfa0c0620867213b4c00f

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (m *Manager) EnabledForUser(userID string) bool {
+	if m == nil {
+		return false
+	}
+	return len(m.limits.CostAttributionLabels(userID)) > 0
+}
+
+func (m *Manager) TrackerForUser(userID string) *Tracker {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 9b4337d14400f3af48fcfa0c0620867213b4c00f

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+	if !m.EnabledForUser(userID) {
+		return nil
+	}
+
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	if tracker, exists := m.trackersByUserID[userID]; exists {
```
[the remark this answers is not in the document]
ying-jeanne: use fine locking in commit 1a523e17ab478102992f5b68deea52fad81a28c9

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (m *Manager) Collect(out chan<- prometheus.Metric) {
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+	for _, tracker := range m.trackersByUserID {
+		tracker.Collect(out)
+	}
+}
+
+func (m *Manager) Describe(chan<- *prometheus.Desc) {
```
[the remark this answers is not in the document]
ying-jeanne: updated comments in commit f10f7872610c468ffcf2130a9a7f7c4834e6f529

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (m *Manager) deleteUserTracker(userID string) {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit cc0e93985a6bac4e3924febaf185c54b17f41b2b

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+	m.mtx.RUnlock()
+
+	for _, userID := range userIDs {
+		if !m.EnabledForUser(userID) {
+			m.deleteUserTracker(userID)
+			continue
+		}
+
+		invalidKeys := m.inactiveObservationsForUser(userID, deadline)
+		cat := m.TrackerForUser(userID)
```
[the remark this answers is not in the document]
ying-jeanne: fixed in commit c020be0eaab34cb5e966c29dd7d376623153ea45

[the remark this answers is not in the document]
ying-jeanne: addressed in commit 71e46669757b40c6c5e9281e4abb97d8a9d4da4b

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+		cat.UpdateMaxCardinality(maxCardinality)
+	}
+
+	cooldown := int64(m.limits.CostAttributionCooldown(userID).Seconds())
+	if cooldown != cat.CooldownDuration() {
+		cat.UpdateCooldownDuration(cooldown)
+	}
+
+	return cat.InactiveObservations(deadline)
+}
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 9dd101ba552d981cd57a7b4d0909194c2bd4a66e

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,172 @@
...
+
+		if cat != nil && cat.cooldownUntil != nil && cat.cooldownUntil.Load() < deadline {
+			if len(cat.observed) <= cat.MaxCardinality() {
+				cat.state = OverflowComplete
+				m.deleteUserTracker(userID)
+			} else {
+				cat.cooldownUntil.Store(deadline + cat.cooldownDuration)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 7d4ea9acbb03ffd7e5372493508903adb3e65614

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	receivedSample   *atomic.Float64
+	discardSamplemtx sync.Mutex
+	discardedSample  map[string]*atomic.Float64
+	totalDiscarded   *atomic.Float64
+}
+
+type Tracker struct {
+	userID                         string
+	caLabels                       []string
+	caLabelMap                     map[string]int
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 67546662291c7bdff3db6f51b6dd91ec78b25bf7

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return t.maxCardinality
+}
+
+func (t *Tracker) CooldownDuration() int64 {
+	if t == nil {
+		return 0
+	}
+	return t.cooldownDuration
+}
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit fffc5b3544710c9510829bee0087e499990047f1

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	tracker.activeSeriesPerUserAttribution = prometheus.NewDesc("cortex_ingester_attributed_active_series",
+		"The total number of active series per user and attribution.", append(trackedLabels, TenantLabel),
+		prometheus.Labels{TrackerLabel: defaultTrackerName})
+	tracker.failedActiveSeriesDecrement = prometheus.NewDesc("cortex_ingester_attributed_active_series_failure",
+		"The total number of failed active series decrement per user and tracker.", []string{TenantLabel},
+		prometheus.Labels{TrackerLabel: defaultTrackerName})
+	return tracker
+}
+
+func (t *Tracker) CompareCALabels(currentLabels []string) bool {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 2cf8c3e488ef2414031387fc3525dedeecb054ae

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	if len(t.caLabels) != len(currentLabels) {
+		return false
+	}
+	for _, v := range currentLabels {
+		if _, exists := t.caLabelMap[v]; !exists {
+			return false
+		}
+	}
+	return true
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit f994034d4d41b67476b9add02d7e86df18c4ad16

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+var bufferPool = sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Buffer)
+	},
+}
+
+func (t *Tracker) cleanupTrackerAttribution(key string) {
+	if t == nil {
+		return
+	}
```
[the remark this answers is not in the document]
ying-jeanne: address in b060c093e6befcbc4302cd75438cbcbd5202a331, make most of private function no longer check nil when not necessary.

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+		if i > 0 {
+			buf.WriteRune(sep)
+		}
+		buf.WriteString(value)
+	}
+
+	t.obseveredMtx.Lock()
+	defer t.obseveredMtx.Unlock()
+
+	t.updateOverflow(buf.String(), ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason)
```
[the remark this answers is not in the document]
ying-jeanne: address in commit the function split 5cc0b5df279d8065a722384aad9f917a7002e9fe

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// handleObservation updates or creates a new stream observation in the 'observed' map.
+func (t *Tracker) handleObservation(stream string, ts int64, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64, reason *string) {
```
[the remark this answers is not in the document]
ying-jeanne: renamed in key in previous commits

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	if o, known := t.observed[stream]; known && o.lastUpdate != nil {
+		// Update the timestamp if needed
+		if o.lastUpdate.Load() < ts {
+			o.lastUpdate.Store(ts)
+		}
```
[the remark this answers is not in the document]
ying-jeanne: what about out of order samples?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+
+const (
+	Normal TrackerState = iota
+	Overflow
+	OverflowComplete
+)
+
+const sep = rune(0x80)
+
+type Observation struct {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 116a69edebea6bc1ed2156ab49c06201c6bfa82b

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	lastUpdate       *atomic.Int64
+	activeSerie      *atomic.Float64
+	receivedSample   *atomic.Float64
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 9c30445a730191b1c38d8cd11ad93b3573d775ea

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	discardSamplemtx sync.Mutex
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 88ef49e2c3af96eb4a775e3fa58128289b29c32e

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	userID                         string
+	caLabels                       []string
+	caLabelMap                     map[string]int
+	maxCardinality                 int
+	activeSeriesPerUserAttribution *prometheus.Desc
+	receivedSamplesAttribution     *prometheus.Desc
+	discardedSampleAttribution     *prometheus.Desc
+	failedActiveSeriesDecrement    *prometheus.Desc
+	overflowLabels                 []string
+	obseveredMtx                   sync.RWMutex
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 130636a1feb5e425cd4993479e6e591f95a8ba72

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	t.maxCardinality = limit
+}
+
+func (t *Tracker) UpdateCooldownDuration(cooldownDuration int64) {
+	if t == nil {
+		return
+	}
+	t.cooldownDuration = cooldownDuration
+}
```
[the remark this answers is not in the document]
ying-jeanne: if we create a new tracker, we would trigger activeseries recount, I would add a mutex

pkg/costattribution/tracker_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_GetCALabels(t *testing.T) {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit b701ba7885fd2c843fe6fea7175526e2f70a0e2d

pkg/distributor/allcase.txt
```diff
@@ -0,0 +1,90 @@
+goos: darwin
```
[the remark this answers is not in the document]
ying-jeanne: removed in commit dccd9c8669aa690c0400bc48cea8950511084c4f

docs/sources/mimir/configure/configuration-parameters/index.md
ying-jeanne: all doc updates addressed in eebd0288658ce4e2c9447d89f5efc8f96e0878ee

pkg/costattribution/manager.go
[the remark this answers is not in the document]
colega: Looks great, thanks

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,321 @@
...
+		discardedSampleMtx: sync.Mutex{},
+	}
+	if discardedSampleIncrement > 0 && reason != nil {
+		t.observed[string(key)].discardedSampleMtx.Lock()
+		t.observed[string(key)].discardedSample[*reason] = *atomic.NewFloat64(discardedSampleIncrement)
+		t.observed[string(key)].discardedSampleMtx.Unlock()
+	}
+}
+
+func (t *Tracker) shouldDelete(deadline int64) bool {
```
colega, line 299: I would rename this method to "recovered from overflow". The fact that it should be deleted is a decision of the Manager.

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,321 @@
...
+	if t.cooldownUntil != nil && t.cooldownUntil.Load() < deadline {
```
colega, line 300: You can't check whether `t.cooldownUntil` without holding the mutex here.

pkg/costattribution/tracker.go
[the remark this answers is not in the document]
ying-jeanne: addressed in 83865039abbf656ee3f1ba9b487cb8b96090336f

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+			o.discardSamplemtx.Lock()
+			o.discardedSample[*reason] = atomic.NewFloat64(discardedSampleIncrement)
+			o.discardSamplemtx.Unlock()
+		}
+	} else if len(t.observed) < t.maxCardinality*2 {
+		// If the ts is negative, it means that the method is called from DecrementActiveSeries, when key doesn't exist we should ignore the call
+		// Otherwise create a new observation for the stream
+		if ts >= 0 {
+			t.createNewObservation(stream, ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason)
+		}
```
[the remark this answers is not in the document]
ying-jeanne: addressed in d8f1e9b01e4a38a1a667d6679657521418a6c600

pkg/costattribution/tracker.go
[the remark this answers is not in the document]
ying-jeanne: I should note that this means we shouldn't permit self-service adjustments for maximum cardinality and cooldown settings.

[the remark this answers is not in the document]
colega: I think that if we allow customers to self-service these limits, these limits stop making sense.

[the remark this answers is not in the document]
ying-jeanne: addressed in b9efb943d9b16e4806d7e794ddae4b425668dc3c

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,162 @@
...
+		m.deleteTracker(userID)
+		return nil
+	}
+
+	newTrackedLabels := m.limits.CostAttributionLabels(userID)
+
+	// sort the labels to ensure the order is consistent
+	sort.Slice(newTrackedLabels, func(i, j int) bool {
+		return newTrackedLabels[i] < newTrackedLabels[j]
+	})
```
colega, line 124: You can probably do `slices.Sort(newTrackedLabels)`

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if t == nil {
+		m.deleteTracker(userID)
+		return nil
+	}
+
+	newTrackedLabels := m.limits.CostAttributionLabels(userID)
+
+	// sort the labels to ensure the order is consistent
+	sort.Slice(newTrackedLabels, func(i, j int) bool {
```
colega, line 122: Is it safe to modify the slice returned by `m.limits.CostAttributionLabels(userID)`?

pkg/distributor/validate.go
```diff
@@ -238,15 +239,17 @@ func newExemplarValidationMetrics(r prometheus.Registerer) *exemplarValidationMe
 // validateSample returns an err if the sample is invalid.
 // The returned error may retain the provided series labels.
 // It uses the passed 'now' time to measure the relative time of the sample.
-func validateSample(m *sampleValidationMetrics, now model.Time, cfg sampleValidationConfig, userID, group string, ls []mimirpb.LabelAdapter, s mimirpb.Sample) error {
+func validateSample(m *sampleValidationMetrics, now model.Time, cfg sampleValidationConfig, userID, group string, ls []mimirpb.LabelAdapter, s mimirpb.Sample, cat *costattribution.Tracker) error {
 	if model.Time(s.TimestampMs) > now.Add(cfg.CreationGracePeriod(userID)) {
 		m.tooFarInFuture.WithLabelValues(userID, group).Inc()
+		cat.IncrementDiscardedSamples(mimirpb.FromLabelAdaptersToLabels(ls), 1, reasonTooFarInFuture, now.Time())
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit a37e6deb52de09f5cdddba3573bbbe9bf279a39c

pkg/distributor/distributor.go
```diff
@@ -1820,6 +1832,7 @@ func (d *Distributor) updateReceivedMetrics(req *mimirpb.WriteRequest, userID st
 	for _, ts := range req.Timeseries {
 		receivedSamples += len(ts.TimeSeries.Samples) + len(ts.TimeSeries.Histograms)
 		receivedExemplars += len(ts.TimeSeries.Exemplars)
+		d.costAttributionMgr.TrackerForUser(userID).IncrementReceivedSamples(mimirpb.FromLabelAdaptersToLabels(ts.Labels), float64(receivedSamples), mtime.Now())
```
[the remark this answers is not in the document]
ying-jeanne: the 2nd point addressed in commit a37e6deb52de09f5cdddba3573bbbe9bf279a39c, for the batch update, it would be addressed in a following commit

pkg/distributor/validate.go
```diff
@@ -392,14 +400,16 @@ func removeNonASCIIChars(in string) (out string) {
 
 // validateLabels returns an err if the labels are invalid.
 // The returned error may retain the provided series labels.
-func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userID, group string, ls []mimirpb.LabelAdapter, skipLabelValidation, skipLabelCountValidation bool) error {
+func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userID, group string, ls []mimirpb.LabelAdapter, skipLabelValidation, skipLabelCountValidation bool, cat *costattribution.Tracker, ts time.Time) error {
 	unsafeMetricName, err := extract.UnsafeMetricNameFromLabelAdapters(ls)
 	if err != nil {
+		cat.IncrementDiscardedSamples(mimirpb.FromLabelAdaptersToLabels(ls), 1, reasonMissingMetricName, ts)
```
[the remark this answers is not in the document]
ying-jeanne: in distributor it is called with numSamples += len(ts.Samples) + len(ts.Histograms) for some of the cases.

pkg/ingester/activeseries/active_series.go
```diff
@@ -84,50 +89,55 @@ type seriesEntry struct {
...
 
-func (c *ActiveSeries) CurrentConfig() asmodel.CustomTrackersConfig {
-	c.matchersMutex.RLock()
-	defer c.matchersMutex.RUnlock()
-	return c.matchers.Config()
+func (c *ActiveSeries) CurrentConfig() (asmodel.CustomTrackersConfig, *costattribution.Tracker) {
+	c.configMutex.RLock()
+	defer c.configMutex.RUnlock()
+	return c.matchers.Config(), c.cat
 }
```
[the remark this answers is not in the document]
ying-jeanne: removed in 211b3a2ca5a0ecab5665b462647436adc95b97f9

pkg/ingester/activeseries/active_series.go
```diff
@@ -449,9 +460,21 @@ func (s *seriesStripe) purge(keepUntil time.Time) {
...
 	oldest := int64(math.MaxInt64)
+	buf := labels.NewScratchBuilder(128)
 	for ref, entry := range s.refs {
 		ts := entry.nanos.Load()
 		if ts < keepUntilNanos {
+			// cost attribution is enabled, if it's not nil, we need to decrement the active series count, otherwise means received error when get idx,
+			// we need to increment the active series failure count.
+			if s.cat != nil {
+				if idx == nil {
+					s.cat.IncrementActiveSeriesFailure(1)
```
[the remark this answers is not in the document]
ying-jeanne: addressed in f697e6f59d728da39f120bf2b464639264820f1f

pkg/costattribution/tracker.go
ying-jeanne: addressed in fe8a1e5565b1cc7b506fe31a64db5ec01b7a31e4

pkg/costattribution/tracker.go
ying-jeanne: addressed in fe8a1e5565b1cc7b506fe31a64db5ec01b7a31e4

pkg/costattribution/manager.go
ying-jeanne: addressed in 8b5836f3ac546913dad152c0f70cb71fd6745518

pkg/costattribution/manager.go
ying-jeanne: modify the copy instead, addressed in 888d8b06231e3e67d1cb7ccf38bac0b4275a95c9

pkg/ingester/ingester.go
```diff
@@ -3243,7 +3273,12 @@ func (i *Ingester) compactBlocksToReduceInMemorySeries(ctx context.Context, now
 		}
 
 		// Purge the active series so that the next call to Active() will return the up-to-date count.
-		db.activeSeries.Purge(now)
+		idx, err := db.Head().Index()
+		if err != nil {
+			level.Warn(i.logger).Log("msg", "failed to get the index of the TSDB head", "user", userID, "err", err)
+			idx = nil
+		}
```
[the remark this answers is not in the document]
ying-jeanne: PR in mimir-prometheus https://github.com/grafana/mimir-prometheus/pull/811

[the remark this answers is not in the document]
ying-jeanne: vendor and update in commit 4706bde40e735e38c882d52827fcf94e34f1bdb3

pkg/ingester/activeseries/active_series.go
```diff
@@ -502,6 +525,18 @@ func (s *seriesStripe) remove(ref storage.SeriesRef) {
 	}
 
 	s.active--
+	if s.cat != nil {
+		if idx == nil {
```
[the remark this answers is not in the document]
ying-jeanne: vendor and update in commit https://github.com/grafana/mimir/commit/4706bde40e735e38c882d52827fcf94e34f1bdb3

pkg/ingester/activeseries/active_series.go
```diff
@@ -449,9 +460,21 @@ func (s *seriesStripe) purge(keepUntil time.Time) {
...
 	for ref, entry := range s.refs {
 		ts := entry.nanos.Load()
 		if ts < keepUntilNanos {
+			// cost attribution is enabled, if it's not nil, we need to decrement the active series count, otherwise means received error when get idx,
+			// we need to increment the active series failure count.
+			if s.cat != nil {
+				if idx == nil {
+					s.cat.IncrementActiveSeriesFailure(1)
+				} else if err := idx.Series(ref, &buf, nil); err != nil {
+					s.cat.IncrementActiveSeriesFailure(1)
```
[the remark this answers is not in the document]
ying-jeanne: vendor and update in commit https://github.com/grafana/mimir/commit/4706bde40e735e38c882d52827fcf94e34f1bdb3

pkg/ingester/activeseries/active_series.go
```diff
@@ -449,9 +460,21 @@ func (s *seriesStripe) purge(keepUntil time.Time) {
...
 
 	oldest := int64(math.MaxInt64)
+	buf := labels.NewScratchBuilder(128)
 	for ref, entry := range s.refs {
 		ts := entry.nanos.Load()
 		if ts < keepUntilNanos {
+			// cost attribution is enabled, if it's not nil, we need to decrement the active series count, otherwise means received error when get idx,
+			// we need to increment the active series failure count.
+			if s.cat != nil {
+				if idx == nil {
```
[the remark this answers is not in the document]
ying-jeanne: addressed as the other comment for idx

pkg/ingester/activeseries/active_series_test.go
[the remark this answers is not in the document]
ying-jeanne: addressed in 17b64a911cdccb0a1d5c1b2fd6eb6837caf1321a

pkg/ingester/activeseries/active_series.go
```diff
@@ -449,9 +460,21 @@ func (s *seriesStripe) purge(keepUntil time.Time) {
 	s.activeMatchingNativeHistogramBuckets = resizeAndClear(len(s.activeMatchingNativeHistogramBuckets), s.activeMatchingNativeHistogramBuckets)
 
 	oldest := int64(math.MaxInt64)
+	buf := labels.NewScratchBuilder(128)
 	for ref, entry := range s.refs {
 		ts := entry.nanos.Load()
 		if ts < keepUntilNanos {
+			// cost attribution is enabled, if it's not nil, we need to decrement the active series count, otherwise means received error when get idx,
+			// we need to increment the active series failure count.
```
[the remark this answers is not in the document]
ying-jeanne: removed

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+		hashBuffer:              make([]byte, 0, 1024),
+		cooldownDuration:        int64(cooldown.Seconds()),
+		logger:                  logger,
+		overflowLabels:          overflowLabels,
+		totalFailedActiveSeries: atomic.NewFloat64(0),
+	}
+
+	tracker.discardedSampleAttribution = prometheus.NewDesc("cortex_discarded_attributed_samples_total",
+		"The total number of samples that were discarded per attribution.",
+		append(trackedLabels, TenantLabel, "reason"),
```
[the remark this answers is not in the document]
ying-jeanne: addressed in a1910441078ae1d1f44c3081e104e1fa29d42a7a

pkg/distributor/validate_test.go
```diff
@@ -60,6 +60,7 @@ func (vm validateMetadataCfg) MaxMetadataLength(_ string) int {
 }
 
 func TestValidateLabels(t *testing.T) {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 2bb1845fbf4bf1b8ca7d2518fe199149b76b87e2

pkg/mimir/modules.go
```diff
@@ -644,6 +648,18 @@ func (t *Mimir) initActiveGroupsCleanupService() (services.Service, error) {
 	return t.ActiveGroupsCleanup, nil
 }
 
+func (t *Mimir) initCostAttributionService() (services.Service, error) {
+	// The cost attribution service is only initilized if the custom registry path is provided.
+	if t.Cfg.CostAttributionRegistryPath != "" {
+		reg := prometheus.NewRegistry()
+		var err error
+		t.CostAttributionManager, err = costattribution.NewManager(3*time.Minute, time.Minute, t.Cfg.CostAttributionEvictionInterval, util_log.Logger, t.Overrides, reg)
```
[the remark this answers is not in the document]
ying-jeanne: active group is using 3*time.Minute as a fixed cron interval, i am thinking not to introduce too many parameters, how about make only export interval a configurable?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	t.totalFailedActiveSeries.Add(value)
+}
+
+func (t *Tracker) updateCounters(lbls labels.Labels, ts int64, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64, reason *string) {
+	labelValues := make([]string, len(t.caLabels))
+	lbls.Range(func(l labels.Label) {
+		if idx, ok := t.caLabelMap[l.Name]; ok {
+			labelValues[idx] = l.Value
+		}
```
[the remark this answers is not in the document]
ying-jeanne: benchmark shows slice is much better performing for a metrics with 40 labels lookup in 2 cost attribution labels, changed in commit ddd507de8d548cd93d264a88d21a3d06c6abb43a

pkg/mimir/modules.go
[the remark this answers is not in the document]
ying-jeanne: removed unused parameter b27e379ccc711b6b194f69db7a5d1a0c82452fe3

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+
+func (t *Tracker) updateCountersCommon(
+	extractValues func() []string,
+	ts int64,
+	activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64,
+	reason *string,
+	createIfDoesNotExist bool,
+) {
+	// Extract label values
+	labelValues := extractValues()
```
colega, line 232: Why are we passing down a callback here instead of just passing the `labelValues []string` to this method?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+		receivedSample:     *atomic.NewFloat64(receivedSampleIncrement),
+		discardedSample:    map[string]atomic.Float64{},
+		discardedSampleMtx: sync.Mutex{},
+	}
+	if discardedSampleIncrement > 0 && reason != nil {
+		t.observed[string(key)].discardedSampleMtx.Lock()
+		t.observed[string(key)].discardedSample[*reason] = *atomic.NewFloat64(discardedSampleIncrement)
+		t.observed[string(key)].discardedSampleMtx.Unlock()
+	}
+}
```
colega, line 334: Since we're creating a new observation here, I would pass a `key string` here, allocated once.

I'm not sure if this method isn't allocating 4 strings here (not a big deal, since it's not a hot path, but still..)

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	defer t.observedMtx.Unlock()
+
+	// Update observations and state
+	t.updateObservations(buf.Bytes(), ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason, createIfDoesNotExist)
+	t.updateState(ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement)
+}
+
+// updateObservations updates or creates a new observation in the 'observed' map.
+func (t *Tracker) updateObservations(key []byte, ts int64, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64, reason *string, createIfDoesNotExist bool) {
+	if o, known := t.observed[string(key)]; known && o.lastUpdate != nil {
```
colega, line 258: How can `o.lastUpdate` be nil?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+			} else {
+				o.discardedSample[*reason] = *atomic.NewFloat64(discardedSampleIncrement)
+			}
+			o.discardedSampleMtx.Unlock()
+		}
+	} else if len(t.observed) < t.maxCardinality*2 && createIfDoesNotExist {
+		// When createIfDoesNotExist is false, it means that the method is called from DecrementActiveSeries, when key doesn't exist we should ignore the call
+		// Otherwise create a new observation for the key
+		t.createNewObservation(key, ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason)
+	}
```
colega, line 280: I would move this to the beginning of the method to un-indent the most common path.
i.e.
```go
o, known := t.observed[string(key)]
if !known {
	if len(t.observed) < t.maxCardinality*2 && createIfDoesNotExist {
		// When createIfDoesNotExist is false, it means that the method is called from DecrementActiveSeries, when key doesn't exist we should ignore the call
		// Otherwise create a new observation for the key
		t.createNewObservation(key, ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason)
	}
	return
}
// ...
```

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	t.updateCountersWithLabelAdapter(lbs, now.Unix(), 0, value, 0, nil, true)
+}
+
+func (t *Tracker) IncrementActiveSeriesFailure() {
+	if t == nil {
+		return
+	}
+	t.totalFailedActiveSeries.Add(1)
+}
```
colega, line 189: I'm still not convinced this belongs here.

What did fail here? `costattribution.Tracker` didn't fail. It's the caller (`activeseries.ActiveSeries`) who failed to retrieve labels for a series it should know.
So it's the caller who should register the failure.

`costattribution.Tracker` wasn't even called here, so it can't fail.

pkg/distributor/distributor.go
```diff
@@ -1820,6 +1832,7 @@ func (d *Distributor) updateReceivedMetrics(req *mimirpb.WriteRequest, userID st
 	for _, ts := range req.Timeseries {
 		receivedSamples += len(ts.TimeSeries.Samples) + len(ts.TimeSeries.Histograms)
 		receivedExemplars += len(ts.TimeSeries.Exemplars)
+		d.costAttributionMgr.Tracker(userID).IncrementReceivedSamples(ts.Labels, float64(receivedSamples), mtime.Now())
```
colega, line 1835:
```suggestion
	cat := d.costAttributionMgr.Tracker(userID)
	for _, ts := range req.Timeseries {
		receivedSamples += len(ts.TimeSeries.Samples) + len(ts.TimeSeries.Histograms)
		receivedExemplars += len(ts.TimeSeries.Exemplars)
		cat.IncrementReceivedSamples(ts.Labels, float64(receivedSamples), mtime.Now())
```

go.mod
```diff
@@ -285,7 +285,7 @@ require (
 )
 
 // Using a fork of Prometheus with Mimir-specific changes.
-replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v0.0.0-20241219104229-b50052711673
+replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v0.0.0-20241224134504-460b7be5bce8
```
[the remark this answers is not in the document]
ying-jeanne: I would vendor main when I got it merged, just to make the code compile now.

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+	return m, nil
+}
+
+func (m *Manager) iteration(_ context.Context) error {
+	return m.purgeInactiveAttributionsUntil(time.Now().Add(-m.inactiveTimeout).Unix())
+}
+
+func (m *Manager) EnabledForUser(userID string) bool {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit 679f2cc02d02ec43420278880d8371f699e5795b

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,161 @@
...
+		m.deleteTracker(userID)
+		return nil
+	}
+
+	t := m.Tracker(userID)
+
+	lbls := m.limits.CostAttributionLabels(userID)
+
+	newTrackedLabels := make([]string, len(lbls))
+	copy(newTrackedLabels, lbls)
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 66accc9512287f763306180e478d5d90a301c074

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+
+const (
+	Normal TrackerState = iota
+	Overflow
+)
+
+const sep = rune(0x80)
+
+type observation struct {
+	lastUpdate         *atomic.Int64
```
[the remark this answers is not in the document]
ying-jeanne: Since it is optional, otherwise I can consider 0 as "not present". Do we prefer always instance than pointer because of the dereference?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	activeSerie        atomic.Float64
+	receivedSample     atomic.Float64
+	discardedSampleMtx sync.Mutex
+	discardedSample    map[string]atomic.Float64
```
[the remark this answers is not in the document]
ying-jeanne: i discovered it in this commit, would make it a map of string and pointer of atomic.Float64 here instead. https://github.com/grafana/mimir/commit/2bb1845fbf4bf1b8ca7d2518fe199149b76b87e2#diff-03f007124c4ba9d1016122f95863471c4a373baff94b02808f491768078851b0,

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	// Transition to overflow mode if maximum cardinality is exceeded.
+	previousState := t.state
+	if t.state == Normal && len(t.observed) > t.maxCardinality {
+		t.state = Overflow
+		// Initialize the overflow counter.
+		t.overflowCounter = &observation{
+			lastUpdate:     atomic.NewInt64(ts),
+			activeSerie:    *atomic.NewFloat64(0),
+			receivedSample: *atomic.NewFloat64(0),
+			totalDiscarded: *atomic.NewFloat64(0),
```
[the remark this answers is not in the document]
ying-jeanne: addressed in commit f4a4efde884013a8e52669a12fbd1ff16c62e9f6

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	failedActiveSeriesDecrement    *prometheus.Desc
+	overflowLabels                 []string
+	observed                       map[string]*observation
+	observedMtx                    sync.RWMutex
+	cooldownUntil                  int64
+	hashBuffer                     []byte
+	state                          TrackerState
+	overflowCounter                *observation
+	totalFailedActiveSeries        *atomic.Float64
+	cooldownDuration               int64
```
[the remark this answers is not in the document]
ying-jeanne: addressed in f90ac0ef07d184a1d08e1ca0496a349b457f14fc

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+	labels                         []string
+	maxCardinality                 int
+	activeSeriesPerUserAttribution *prometheus.Desc
+	receivedSamplesAttribution     *prometheus.Desc
+	discardedSampleAttribution     *prometheus.Desc
+	failedActiveSeriesDecrement    *prometheus.Desc
+	overflowLabels                 []string
+	observed                       map[string]*observation
+	observedMtx                    sync.RWMutex
+	cooldownUntil                  int64
```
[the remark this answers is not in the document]
ying-jeanne: addressed in f90ac0ef07d184a1d08e1ca0496a349b457f14fc

pkg/costattribution/tracker.go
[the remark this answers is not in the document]
ying-jeanne: addressed in 1ab89c582f7ddafc3643995127151ed1f941c145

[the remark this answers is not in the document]
colega: How can you have an observation that was never updated?

pkg/costattribution/tracker.go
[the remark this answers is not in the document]
ying-jeanne: addressed in 23b32cff274b9737494ae18db6178b0161314f87

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+
+	"github.com/grafana/mimir/pkg/mimirpb"
+)
+
+type TrackerState int
+
+const (
+	Normal TrackerState = iota
+	Overflow
+)
```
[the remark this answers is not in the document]
ying-jeanne: I think has a state is more clear what we mean, if you have a strong feeling with boolean I can change.

pkg/costattribution/tracker.go
ying-jeanne: addressed in 7a60c7d7c63f5a8dd5385e2a0af8e3644b46b529, because I thought i was smart

pkg/costattribution/tracker.go
ying-jeanne: addressed in 0287bf6cad53e0d4f32fdce6f47f44298efee148

pkg/costattribution/tracker.go
ying-jeanne: it can't, after changed to int64 instead of pointer

pkg/costattribution/tracker.go
ying-jeanne: when activeseries.ActiveSeries failed to get index, it is calling this method to increment failure, would you prefer ingester do this job? the advantage of having it here is it goes to cost attribution prometheus registry instead of ops registry.

pkg/costattribution/tracker.go
ying-jeanne: addressed in 9c4c2dfa32e4961b412e0d91eb391e9184b80c03

pkg/distributor/distributor.go
ying-jeanne: addressed in f8f2a49360e0028d56562b6da033fc0dd5c8f63a

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,371 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/model/labels"
+	"go.uber.org/atomic"
+
+	"github.com/grafana/mimir/pkg/mimirpb"
+)
+
+type TrackerState int
```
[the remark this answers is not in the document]
ying-jeanne: changed to private

pkg/costattribution/tracker.go
colega: When `activeseries.ActiveSeries` fails to do something, it should increment its failed metric, not someone else's (this component's), because this component didn't fail.

pkg/costattribution/tracker.go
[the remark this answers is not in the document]
ying-jeanne: yes sir we can. hecho aqui fa62ee1afa9d410fe37b85a47bba6db1b36d2f54

pkg/costattribution/tracker.go
ying-jeanne: make ingester work in commit 1b0fb005cc0ce0c144f2ca83e21e428c8e70aa03

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,365 @@
...
+
+func (t *Tracker) updateCounters(lbls labels.Labels, ts time.Time, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64, reason *string, createIfDoesNotExist bool) {
+	labelValues := make([]string, len(t.labels))
+	for idx, cal := range t.labels {
+		labelValues[idx] = lbls.Get(cal)
+		if labelValues[idx] == "" {
+			labelValues[idx] = missingValue
+		}
+	}
+	key := t.hashLabelValues(labelValues)
```
ying-jeanne, line 223: I am not so sure that get a key string everytime is a very good idea, but it did make the code less complex

colega: It simplifies the code at the cost of one allocation per each series received. Please let's revert to the previous approach where we used a pooled byte slice to perform this lookup.

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+func (t *Tracker) IncrementReceivedSamples(req *mimirpb.WriteRequest, now time.Time) {
+	if t == nil {
+		return
+	}
+
+	dict := make(map[string]int)
+	for _, ts := range req.Timeseries {
+		lvs := t.extractLabelValuesFromLabelAdapater(ts.Labels)
+		dict[t.hashLabelValues(lvs)] += len(ts.TimeSeries.Samples) + len(ts.TimeSeries.Histograms)
+	}
```
colega, line 173: This is the hottest path on our application, we should optimize it as much as possible.
Why do we need to build a new data structure (which escapes to heap) with holds `[]mimirpb.LabelAdapter` slices that escape to heap, which create a string that escapes to heap just to put it into a `dict` that I don't think should be a map, even if we need _some_ data structure.

Can we just extract the labelValues byte slices, recycled from a pool, and process each one separately in the loop below?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	t.updateCounters(lbs, time.Time{}, -1, 0, 0, nil, false)
+}
+
+func (t *Tracker) Collect(out chan<- prometheus.Metric) {
+
+	if t.isOverflow.Load() {
+		out <- prometheus.MustNewConstMetric(t.activeSeriesPerUserAttribution, prometheus.GaugeValue, t.overflowCounter.activeSerie.Load(), t.overflowLabels[:len(t.overflowLabels)-1]...)
+		out <- prometheus.MustNewConstMetric(t.receivedSamplesAttribution, prometheus.CounterValue, t.overflowCounter.receivedSample.Load(), t.overflowLabels[:len(t.overflowLabels)-1]...)
+		out <- prometheus.MustNewConstMetric(t.discardedSampleAttribution, prometheus.CounterValue, t.overflowCounter.totalDiscarded.Load(), t.overflowLabels...)
```
colega, line 129: Why are we doing the sub-slicing here to the length of the slice? That sounds like a noop.

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	if !t.isOverflow.Load() && len(t.observed) > t.maxCardinality {
+		// Make sure that we count current overflow only when state is switched to overflow from normal.
+		previousOverflow = t.isOverflow.Swap(true)
+		if !previousOverflow {
+			// Initialize the overflow counter.
+			t.overflowCounter = &observation{}
+
+			// Aggregate active series from all keys into the overflow counter.
+			for _, o := range t.observed {
+				if o != nil {
```
colega, line 283: How can `o` be nil?

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+			// Otherwise create a new observation for the key
+			t.createNewObservation(key, ts, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement, reason)
+		}
+		return
+	}
+
+	o.lastUpdate.Store(ts)
+	if activeSeriesIncrement != 0 {
+		o.activeSerie.Add(activeSeriesIncrement)
+	}
```
colega, line 254: We didn't check the overflow here, so we're incremeting something that isn't being used anymore, which means that the overflow number is wrong.

If we want to keep the overflow number correct, we need to handle these race conditions (and I don't think it will be easy).

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+func (t *Tracker) updateState(ts time.Time, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64) {
+	previousOverflow := true
+	t.observedMtx.RLock()
+	// Transition to overflow mode if maximum cardinality is exceeded.
+	if !t.isOverflow.Load() && len(t.observed) > t.maxCardinality {
+		// Make sure that we count current overflow only when state is switched to overflow from normal.
+		previousOverflow = t.isOverflow.Swap(true)
+		if !previousOverflow {
+			// Initialize the overflow counter.
+			t.overflowCounter = &observation{}
```
colega, line 279: There should be some kind of concurrency coordination here on setting this property.

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+			t.cooldownUntil = ts.Add(t.cooldownDuration)
+		}
+	}
+	t.observedMtx.RUnlock()
+
+	if t.isOverflow.Load() {
+		// if already in overflow mode, update the overflow counter. If it was normal mode, the active series are already applied.
+		if previousOverflow && activeSeriesIncrement != 0 {
+			t.overflowCounter.activeSerie.Add(activeSeriesIncrement)
```
colega, line 295: `t.overflowCounter` can be nil here.

pkg/costattribution/tracker.go
ying-jeanne: Since the discarded sample metric includes an additional "reason" label compared to the other two, it is the only one that contains the full set of overflow labels. The other two metrics will have one fewer label (i.e., len(overflow)-1).

pkg/costattribution/tracker.go
ying-jeanne: as paired, addressed in this commit a2ffe5a893c9a9eef26160c760cc6e57e50a3f1d

pkg/costattribution/tracker.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+}
+
+// createNewObservation creates a new observation in the 'observed' map.
+func (t *Tracker) createNewObservation(key string, ts int64, activeSeriesIncrement, receivedSampleIncrement, discardedSampleIncrement float64, reason *string) {
+	t.observedMtx.Lock()
+	defer t.observedMtx.Unlock()
+	if _, exists := t.observed[key]; exists {
+		return
+	}
```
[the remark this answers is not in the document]
ying-jeanne: fixed in f28d67212d212e1c43b623a47ec887b09d679f82

pkg/costattribution/tracker.go
ying-jeanne: removed in f28d67212d212e1c43b623a47ec887b09d679f82

pkg/costattribution/tracker.go
ying-jeanne: as paired, it is addressed in a2ffe5a893c9a9eef26160c760cc6e57e50a3f1d and f28d67212d212e1c43b623a47ec887b09d679f82

pkg/costattribution/tracker.go
ying-jeanne: this logic is gone in f28d67212d212e1c43b623a47ec887b09d679f82

pkg/costattribution/tracker.go
ying-jeanne: remove the pointer, putting a instance instead.

pkg/distributor/distributor.go
```diff
@@ -746,13 +749,15 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica
 // Returns an error explaining the first validation finding.
 // May alter timeseries data in-place.
 // The returned error may retain the series labels.
+
```
colega, line 752:
```suggestion
```

pkg/util/validation/limits.go
```diff
@@ -797,6 +814,22 @@ func (o *Overrides) SeparateMetricsGroupLabel(userID string) string {
 	return o.getOverridesForUser(userID).SeparateMetricsGroupLabel
 }
 
+func (o *Overrides) CostAttributionLabels(userID string) []string {
+	return o.getOverridesForUser(userID).CostAttributionLabels
+}
+
+func (o *Overrides) MaxCostAttributionLabelsPerUser(userID string) int {
+	return o.getOverridesForUser(userID).MaxCostAttributionLabelsPerUser
+}
```
colega, line 823: Is this a user setting? I can we make a fixed upper bound to this?

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,227 @@
...
+type Manager struct {
+	services.Service
+	logger log.Logger
+	limits *validation.Overrides
+	reg    *prometheus.Registry
+
+	mstx                   sync.RWMutex
+	sampleTrackersByUserID map[string]*SampleTracker
+	inactiveTimeout        time.Duration
+	cleanupInterval        time.Duration
```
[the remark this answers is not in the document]
ying-jeanne: I wanted to put it near sampleTracker since they are not used by activeSeriesTracker. I can move them to the common area.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## main / unreleased
 
+* [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269
```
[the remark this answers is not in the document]
ying-jeanne: addressed in 68410c8128c495897c224a11d2c8d2e22ee42069

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,227 @@
...
+
+	matx                   sync.RWMutex
+	activeTrackersByUserID map[string]*ActiveSeriesTracker
```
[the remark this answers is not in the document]
ying-jeanne: renamed stmtx and atmtx represent sample tracker mutex and active tracker mutex in commit 68410c8128c495897c224a11d2c8d2e22ee42069

pkg/costattribution/manager.go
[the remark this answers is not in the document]
ying-jeanne: moved in commit 68410c8128c495897c224a11d2c8d2e22ee42069

pkg/util/validation/limits.go
ying-jeanne: capped to 4 in this commit and test updated 68410c8128c495897c224a11d2c8d2e22ee42069

development/mimir-microservices-mode/config/mimir.yaml
```diff
@@ -183,6 +185,11 @@ limits:
   ha_cluster_label: ha_cluster
   ha_replica_label: ha_replica
   ha_max_clusters: 10
+  
+  cost_attribution_labels: "container"
```
colega, line 189: This should be an array, right? Not a single string.

ying-jeanne: it would be string seperated by comma, would update the config to make it more clear.

ying-jeanne: updated and tested in fbb2fac8a1b1f52af2fabf1ccdda683592ed22f7

colega: This is YAML, why are we using a comma separated string here? I also see it being defined as a slice (comma separated string slice) type in go. Are you sure this is working correctly? I don't really see any tests in that commit that would suggest that this is not parsed as a single `container,instance` label.

ying-jeanne: i ran the service config in order to tested it, this test should covered it https://github.com/grafana/mimir/pull/10269/files#diff-8f7093a5156899c67bddcb606deac6ee65300ffa0e52497045243995b1a18740R1081, i am checking whether I can add more tests

ying-jeanne: I think i would just remove it, i use this to test in local, it is not part of the commit.

ying-jeanne: image
the parse works on this config

pkg/costattribution/active_tracker.go
```diff
@@ -0,0 +1,193 @@
...
+	overflowSince                  atomic.Int64
+	overflowCounter                atomic.Int64
+	cooldownDuration               time.Duration
+	logger                         log.Logger
+}
+
+func newActiveSeriesTracker(userID string, trackedLabels []string, limit int, cooldownDuration time.Duration, logger log.Logger) *ActiveSeriesTracker {
+	orderedLables := slices.Clone(trackedLabels)
+	slices.Sort(orderedLables)
+
```
colega, line 36: We can remove this, because the constructor isn't exported and the manager already clones and sorts the labels before calling this.

colega:
```suggestion
```

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,308 @@
...
+	overflowSince              atomic.Int64
+	overflowCounter            observation
+	cooldownDuration           time.Duration
+	logger                     log.Logger
+}
+
+func newSampleTracker(userID string, trackedLabels []string, limit int, cooldown time.Duration, logger log.Logger) *SampleTracker {
+	orderedLables := slices.Clone(trackedLabels)
+	slices.Sort(orderedLables)
+
```
colega, line 47: We can also remove this for same reason.
```suggestion
```

ying-jeanne: it is called from 2 places, I moved the order function to the other caller to avoid double ordering 5c26a0b4dcf3040586b33dafe91fa403c8be0245

pkg/costattribution/active_tracker.go
ying-jeanne: addressed in 5c26a0b4dcf3040586b33dafe91fa403c8be0245, moved the order to another caller.

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,305 @@
...
+	for _, m := range prometheusMetrics {
+		out <- m
+	}
+}
+
+func (st *SampleTracker) IncrementDiscardedSamples(lbs []mimirpb.LabelAdapter, value float64, reason string, now time.Time) {
+	if st == nil {
+		return
+	}
+	st.updateCountersWithLabelAdapter(lbs, now, 0, value, &reason)
```
colega, line 127: Nit: I don't think this method call adds value: that method is 6 lines long and it's not called anywhere else. Just put its contents here.

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (st *SampleTracker) IncrementReceivedSamples(req *mimirpb.WriteRequest, now time.Time) {
+	if st == nil {
+		return
+	}
+
+	// We precompute the cost attribution per request before update Observations and State to avoid frequently update the atomic counters
+	dict := make(map[string]int)
```
colega, line 136:
```suggestion
	// We precompute the cost attribution per request before update Observations and State to avoid frequently update the atomic counters.
	// This is based on the assumption that usually a single WriteRequest will have samples that belong to the same or few cost attribution groups.
	dict := make(map[string]int)
```

pkg/costattribution/sample_tracker.go
ying-jeanne: removed here 73a988160668f510d1b2d89866427fc3fe47d5c8

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,302 @@
...
+		o.lastUpdate.Store(ts.Unix())
+		o.receivedSample.Add(receivedSampleIncrement)
+		if discardedSampleIncrement > 0 && reason != nil {
+			o.discardedSampleMtx.Lock()
+			if _, ok := o.discardedSample[*reason]; ok {
+				o.discardedSample[*reason].Add(discardedSampleIncrement)
+			} else {
+				o.discardedSample[*reason] = atomic.NewFloat64(discardedSampleIncrement)
+			}
+			o.discardedSampleMtx.Unlock()
```
colega, line 209: Either we switch `discardedSampleMtx` to a RWMutex (and we call `.Add()` while holding just the `RLock`, and we grab the `Lock` for creation, with a second check), or we switch the map to use non-atomic values (because why we need them atomic while we're holding a mutex?)

Given that the amount of reasons is most likely small, I would go for a RWMutex.

If we don't do this, then a customer that is discarding lots of samples would be causing lock contention here.

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,302 @@
...
+		st.observed[key].discardedSampleMtx.Lock()
+		st.observed[key].discardedSample[*reason] = atomic.NewFloat64(discardedSampleIncrement)
+		st.observed[key].discardedSampleMtx.Unlock()
+	}
+}
+
+func (st *SampleTracker) recoveredFromOverflow(deadline time.Time) bool {
+	st.observedMtx.RLock()
+	if st.overflowSince.Load() > 0 && time.Unix(st.overflowSince.Load(), 0).Add(st.cooldownDuration).Before(deadline) {
+		if len(st.observed) <= st.maxCardinality {
```
colega, line 273: Previously we checked:
```go
		// if it is not known, we need to check if the max cardinality is exceeded
		if len(st.observed) >= st.maxCardinality {
			st.overflowSince.Store(ts.Unix())
		}
```
So _equals_ meant we're in overflow, but now _equals_ means we've recovered from overflow. This is going to cause continous flapping there.

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,302 @@
...
+		st.overflowCounter.receivedSample.Add(receivedSampleIncrement)
+		if discardedSampleIncrement > 0 && reason != nil {
+			st.overflowCounter.totalDiscarded.Add(discardedSampleIncrement)
+		}
+		st.observedMtx.RUnlock()
+		return
+	}
+
+	o, known := st.observed[key]
+	if known && st.overflowSince.Load() == 0 {
```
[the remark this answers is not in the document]
ying-jeanne: updated in commit 6d27724c3424fdb1c226b1304d59743824216c8f

pkg/costattribution/manager.go
```diff
@@ -0,0 +1,236 @@
...
+	for _, userID := range userIDs {
+		st, at := m.updateTracker(userID)
+		if st == nil && at == nil {
+			continue
+		}
+
+		invalidKeys := st.inactiveObservations(deadline)
+		for _, key := range invalidKeys {
+			st.cleanupTrackerAttribution(key)
+		}
```
colega, line 223: This is very inefficient because you have to take a write mutex for each observation to cleanup.

OTOH, you're not checking again the observations before cleaning them up.

I would suggest you to delegate this logic to the tracker:

```suggestion
		st.cleanupInactiveObservations(deadline)
```

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,305 @@
...
+	labels                     []string
+	maxCardinality             int
+	receivedSamplesAttribution *prometheus.Desc
+	discardedSampleAttribution *prometheus.Desc
+	overflowLabels             []string
+	observed                   map[string]*observation
+	observedMtx                sync.RWMutex
+	// overflowSince is also protected by observedMtx, it is set when the max cardinality is exceeded
+	overflowSince    time.Time
+	overflowCounter  observation
```
colega, line 40: Everything that is protected by a mutex should go under a mutex in a separate group, to make it obvious.

pkg/costattribution/sample_tracker.go
ying-jeanne: addressed here 5ac64f5df5a2f8442c7a93b3a124d854e8bebb1f

pkg/costattribution/sample_tracker.go
ying-jeanne: changed the condition to strict less than maxcardinality in the mean time e5cda41a140083522dffcee0aad1cab3ff73d415

pkg/costattribution/manager.go
ying-jeanne: addressed here 41e9f477ba11a30db851e89fd8524b2dab0b39b4

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,305 @@
...
+	defer st.observedMtx.Unlock()
+	delete(st.observed, key)
+}
+
+func (st *SampleTracker) Collect(out chan<- prometheus.Metric) {
+	// We don't know the performance of out receiver, so we don't want to hold the lock for too long
+	var prometheusMetrics []prometheus.Metric
+	st.observedMtx.RLock()
+
+	if !st.overflowSince.IsZero() {
```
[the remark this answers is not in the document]
ying-jeanne: addressed in this commit cb5512bd0f3e765422d0e19164de3c23369552f2

pkg/costattribution/sample_tracker.go
ying-jeanne: addressed in this commit cb5512bd0f3e765422d0e19164de3c23369552f2 for both trackers.

pkg/costattribution/active_tracker.go
```diff
@@ -0,0 +1,188 @@
...
+		return
+	}
+	at.observedMtx.RUnlock()
+
+	at.observedMtx.RLock()
+	if !at.overflowSince.IsZero() {
+		at.overflowCounter.Dec()
+		return
+	}
+	defer at.observedMtx.RUnlock()
```
colega, line 146: This should be before the `if`. We're not unlocking in the previous return statement.

pkg/costattribution/sample_tracker.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+		return
+	}
+
+	// create a new observation
+	st.observed[key] = &observation{
+		lastUpdate:         *atomic.NewInt64(ts.Unix()),
+		discardedSample:    make(map[string]*atomic.Float64),
+		receivedSample:     *atomic.NewFloat64(receivedSampleIncrement),
+		discardedSampleMtx: sync.RWMutex{},
```
colega, line 269:
```suggestion
```

pkg/costattribution/active_tracker.go
ying-jeanne: addressed here 80b64dc69133148805220b4d7638daaf95389676

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
 * [CHANGE] Query-frontend: Add `topic` label to `cortex_ingest_storage_strong_consistency_requests_total`, `cortex_ingest_storage_strong_consistency_failures_total`, and `cortex_ingest_storage_strong_consistency_wait_duration_seconds` metrics. #10220
 * [CHANGE] Ruler: cap the rate of retries for remote query evaluation to 170/sec. This is configurable via `-ruler.query-frontend.max-retries-rate`. #10375 #10403
+* [CHANGE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269
```
colega, line 12: Sorry for a late comment, but this should be a FEATURE.

PR #10277, opened 2024-12-18
Show warnings for lazySeriesSet

zenador:
#### What this PR does

Fixes a bug where warning/info annotations are missing when we use lazy querier.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Distributor: Add native histogram support for `electedReplicaPropagationTime` metric in ha_tracker. #10264
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
+* [BUGFIX] Querier: show warning/info annotations in some cases where they were missing (if a lazy querier was used). #10277
```
dimitarvdimitrov, line 15: this also affects the query-frontend, right?

dimitarvdimitrov (approved):

CHANGELOG.md
zenador: Actually `NewLazyQueryable` seems only used in the frontend, I guess it should just be changed to that instead of querier?

PR #10284, opened 2024-12-19
ingester: always log CPU samples for CPU limiter

dimitarvdimitrov:
#### What this PR does

I don't see a reason why we might disable this. The memory footprint is miniscule and the CPU usage is minimal too. The samples give better granularity on what event in the past 60s might have triggered this: "did CPU increase just now or has it been increasing since 30s ago"

Since I think it's harmless, we also don't need a flag for it. The flag is experimental so we can just drop it without notice.

I'm using the occasion to also log the CPU samples both when enabling and disabling the CPU limiter

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * [CHANGE] Distributor: OTLP and push handler replace all non-UTF8 characters with the unicode replacement character `\uFFFD` in error messages before propagating them. #10236
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
+* [CHANGE] Ingester: remove experimental flag `-ingester.log-utilization-based-limiter-cpu-samples`. #10284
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [CHANGE] Distributor: OTLP and push handler replace all non-UTF8 characters with the unicode replacement character `\uFFFD` in error messages before propagating them. #10236
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
+* [CHANGE] Ingester: remove experimental flag `-ingester.log-utilization-based-limiter-cpu-samples`. #10284
```
duricanikolic, line 9: Did we have a CHANGELOG entry when this flag was introduced? If not, I think that this one can also be removed.

dimitarvdimitrov: ah, nice
```suggestion
```

tacole02 (approved):
Thank you!

PR #10285, opened 2024-12-19
`make generate-otlp`: check that tools are installed

colega:
#### What this PR does

Saves some time to the next developer who has to run `make generate-otlp` and isn't familiar with the tools required.

### DIFF
--- a/pkg/distributor/otlp/generate.sh
+++ b/pkg/distributor/otlp/generate.sh
@@ -3,8 +3,11 @@
 
 set -euo pipefail
 
+command -v gopatch >/dev/null 2>&1 || { echo "Please install gopatch. Run 'go install github.com/uber-go/gopatch@latest' or visit https://github.com/uber-go/gopatch for more info."; exit 1; }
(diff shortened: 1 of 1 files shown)


pkg/distributor/otlp/generate.sh
```diff
@@ -3,8 +3,11 @@
 
 set -euo pipefail
 
+command -v gopatch >/dev/null 2>&1 || { echo "Please install gopatch. Run 'go install github.com/uber-go/gopatch@latest' or visit https://github.com/uber-go/gopatch for more info."; exit 1; }
+command -v goimports >/dev/null 2>&1 || { echo "Please install goimports. Run 'go install golang.org/x/tools/cmd/goimports@latest' or visit https://pkg.go.dev/golang.org/x/tools/cmd/goimports for more info."; exit 1; }
+
 # Use GNU sed on MacOS falling back to `sed` everywhere else
-SED=sed
+SED="sed"
```
colega, line 10: Note: making shellcheck happy here.

NickAnge (approved):

56quarters (approved):
Seems fine to me but could we run this in a build container that has everything installed already the same way with do with tests?

colega:
> Seems fine to me but could we run this in a build container that has everything installed already the same way with do with tests?

Would be great, but totally out of scope for me right now (I even forgot about this PR)

PR #9156, opened 2024-08-31
update prometheus no client golang at grafana/mimir-prometheus@fdf902d

krajorama:
#### What this PR does

Update mimir-prometheus at https://github.com/grafana/mimir-prometheus/commit/fdf902dd68d9ae0da99356ce6729d052f867a926

The remote read client interface has changed in https://github.com/prometheus/prometheus/pull/11379
I've updated the users and also added native histogram support to backfill.

Keeps client_golang version as it has some bugs.
https://github.com/prometheus/client_golang/issues/1605
https://github.com/prometheus/client_golang/issues/1607

PR #10286, opened 2024-12-19
fix(mimirtool): Make remote-read actually return data

julienduchesne:
Since https://github.com/grafana/mimir/pull/9156 probably, all the remote read commands return no data

I added a missing error return using `timeseries.Err()` and got this error: `mimirtool: error: chunkedReader: message size exceeded the limit 0 bytes; got: 361 bytes, try --help` which lead me to find that we need to pass in `ChunkedReadLimit` on the client

I set it at a default 1 MiB and configurable through a flag

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirtool/commands/remote_read.go
+++ b/pkg/mimirtool/commands/remote_read.go
@@ -16,6 +16,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -46,9 +47,10 @@ type RemoteReadCommand struct {
 	readTimeout time.Duration
 	tsdbPath    string
 
-	selector string
-	from     string
-	to       string
+	selector      string
+	from          string
+	to            string
+	readSizeLimit uint64
 }
 
 func (c *RemoteReadCommand) Register(app *kingpin.Application, envVars EnvVarNames) {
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov (approved):
1MiB is also what mimir uses by default https://github.com/grafana/mimir/blob/7485d55dd74cff604e99d3277c1adc938ba5977e/pkg/querier/remote_read.go#L37-L40

SGTM, but maybe having something like 10-100MiB won't hurt so mimirtool is interoperable with other implementations that might choose slightly higher values.

dimitarvdimitrov:
I think the limit before the PR you linked was 50MiB, so we can restore that 


```
// DefaultChunkedReadLimit is the default value for the maximum size of the protobuf frame client allows.
// 50MB is the default. This is equivalent to ~100k full XOR chunks and average labelset.
const DefaultChunkedReadLimit = 5e+7
```

pkg/mimirtool/commands/remote_read.go
```diff
@@ -91,7 +95,7 @@ func (c *RemoteReadCommand) Register(app *kingpin.Application, envVars EnvVarNam
 			Default(now.Format(time.RFC3339)).
 			StringVar(&c.to)
 		cmd.Flag("read-size-limit", "Maximum number of bytes to read.").
-			Default(strconv.Itoa(int(math.Pow(1024, 2)))). // 1MiB
+			Default(strconv.Itoa(DefaultChunkedReadLimit)). // 1MiB
```
dimitarvdimitrov, line 98:
```suggestion
			Default(strconv.Itoa(DefaultChunkedReadLimit)).
```

dimitarvdimitrov (approved):

PR #10316, opened 2024-12-29
move one mimirpb custom type to separate package

francoposa:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (commented):
I'm missing the context of this change. However, instead of calling the package `mimirpb_custom` have you considered naming the package based on the actual content, e.g. `labelspb`?

francoposa:
> I'm missing the context of this change. However, instead of calling the package `mimirpb_custom` have you considered naming the package based on the actual content, e.g. `labelspb`?

@pracucci it is not yet fleshed out - it appears making this a draft did not stop it from tagging a lot of people for review.

This PR is just part of an exploration of what it would take to move off of `gogoproto` - specifically here I am looking at removing the need for the custom type extension and still getting everything to compile & import correctly.

As far as package construction, we can look at specifics when this gets closer to an actual in project, but the `mimirpb` package is the only which exports custom proto types - specifically this `LabelAdapter`, `PreallocatingMetric`, `PreallocTimeseries`, and `UnsafeByteSlice`.
I just moved `LabelAdapter` here to prove out the process, but I figured it would be least disruptive but to keep those together especially as they all reference / build on each other.

PR #10325, opened 2024-12-30
Limiter: avoid stop-the-world pause when reading heap size

bboreham:
#### What this PR does

Use a Go metric instead of calling `runtime.ReadMemStats`. Empirically this is much faster.

```
                  │ before.txt  │             after.txt              │
                  │   sec/op    │   sec/op     vs base               │
CombinedScanner-4   25.58µ ± 2%   12.41µ ± 3%  -51.48% (p=0.002 n=6)
```

In a trace from a large ingester; the stop-the-world pause was more like 1-5ms; see example in image.
(and this is called every 1 second)

image

#### Checklist

- NA Tests updated.
- NA Documentation added.


aknuds1 (approved):
LGTM, thanks!

dimitarvdimitrov (approved):
SGTM

PR #10326, opened 2025-01-01
MQE: Add support for irate and idelta function 

lamida:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -533,3 +533,61 @@ func linearRegression(head, tail []promql.FPoint, interceptTime int64) (slope, i
...
+
+		// We need at least two samples to calculate irate or idelta.
+		if len(fHead)+len(fTail) < 2 {
+			return 0, false, nil, nil
+		}
+
+		var lastSample promql.FPoint
+		var previousSample promql.FPoint
+
+		// If tails has more than two samples, we should use the last two samples from tail.
```
jhesketh, line 559: (nit)
```suggestion
		// If tail has more than two samples, we should use the last two samples from tail.
```

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -533,3 +533,61 @@ func linearRegression(head, tail []promql.FPoint, interceptTime int64) (slope, i
...
+	SeriesMetadataFunction: DropSeriesName,
+	StepFunc:               irate(true),
+}
+
+var Idelta = FunctionOverRangeVectorDefinition{
+	SeriesMetadataFunction: DropSeriesName,
+	StepFunc:               irate(false),
+}
+
+func irate(isRate bool) RangeVectorStepFunction {
```
jhesketh, line 547: (nit)
```suggestion
func irateIdelta(isRate bool) RangeVectorStepFunction {
```

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -533,3 +533,61 @@ func linearRegression(head, tail []promql.FPoint, interceptTime int64) (slope, i
...
+	return func(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
+		fHead, fTail := step.Floats.UnsafePoints()
+
+		// We need at least two samples to calculate irate or idelta.
+		if len(fHead)+len(fTail) < 2 {
```
jhesketh, line 552: (nit) `len(fTail)` (and head) are used a lot below. Consider moving to a variable.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -533,3 +533,61 @@ func linearRegression(head, tail []promql.FPoint, interceptTime int64) (slope, i
...
+	StepFunc:               irate(true),
+}
+
+var Idelta = FunctionOverRangeVectorDefinition{
+	SeriesMetadataFunction: DropSeriesName,
+	StepFunc:               irate(false),
+}
+
+func irate(isRate bool) RangeVectorStepFunction {
+	return func(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
```
jhesketh, line 548: (nit)
```suggestion
	return func(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
		// Histograms are ignored
```

jhesketh (commented):
Suggestion for some tests:
 - Something with a reset counter
 - Previous sample is a Histogram, so it needs to look back further (if that's consistent with promql)

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Addressed in 5383c9cc0dcf10fd046b00314d929a42f753b263

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Addressed in 5383c9cc0dcf10fd046b00314d929a42f753b263

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: Suggest to move the lenTail/lenHead to above this line so you can use it here too

lamida:
> Suggestion for some tests:
> 
> * Something with a reset counter
> * Previous sample is a Histogram, so it needs to look back further (if that's consistent with promql)

More tests are added in edaf8243189fb11d01d550c07bb7ca9c24484659

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Ahh you are right. Moved up in 1b1f3591afcf6d12df817ae972a4fb5e0cc4e6c9

jhesketh (approved):
LGTM, thanks for that.

We could add a test (outside of the testdata tests) to check the head/tail boundaries of the ring buffer, but I'm happy as it is.

PR #10329, opened 2025-01-02
tools/copyblocks: Disable pprof endpoint, to satisfy gosec's G108 rule

aknuds1:
#### What this PR does

Disable pprof endpoint in tools/copyblocks, to satisfy gosec's [G108 rule](https://github.com/securego/gosec?tab=readme-ov-file#available-rules). Also enable gosec for said rule.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,7 +11,6 @@ linters:
     - errorlint
     - forbidigo
     - loggercheck
-    - staticcheck
     - gosec
 
 linters-settings:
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -11,6 +11,8 @@ linters:
     - errorlint
     - forbidigo
     - loggercheck
+    - staticcheck
```
andyasp, line 14: [Documentation](https://golangci-lint.run/usage/linters/#enabled-by-default) says `staticcheck` is enabled by default which makes me think this isn't necessary.

andyasp (approved):

.golangci.yml
aknuds1: @andyasp yes, it's not necessary per se. I just figured it's good to be explicit about enabling it.

andyasp: Either way works for me. It only jumped out at me because I didn't know why it was included and went looking.

aknuds1: Removed it.

PR #10340, opened 2025-01-03
mimir-distributed: allow components to override their container image

### LINKED ISSUE #10237: Allow override image of individual components in Helm chart
Right now the `mimir-distributed` Helm chart only has a global configuration, that defines the container image, all Mimir components use (ref [`image` and `enterprise.image` in the default values][1]).

To be on par with Mimir Jsonnet bundle, the chart must provide a way to override image for an individual component (e.g. `ingester`, `querier`, `store-gateway`, etc).

Multiple use-cases for that are coming from the Grafana's internal experience:

1. When backporting a hotfix into a weekly release, we may want to avoid re-deploying the whole system, and only (temporary) update one of the components
2. Similarly, during testing, only a subset of components (usually single component under testing) may need to be overridden with a private development copy

[1]: https://github.com/grafana/mimir/blob/3af22a865221e493426ddbbf841fe55bf640662a/operations/helm/charts/mimir-distributed/values.yaml

narqo:
#### What this PR does

The PR updates the Helm chart to allow an individual component to override its container image reference via the component's local values. E.g.

```yaml
# values.yaml

ingester:
  image:
    repository: registry.example/mimir-ingester
    tag: blah-blah
```

The order in which the values take precedence is:

1. take the component's local image or tag if defined
1. take the image defined in the enterprise section if enterprise is enabled
1. take the globally-defined image otherwise

Fixes #10237

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 ## main / unreleased
 
 * [CHANGE] Memcached: Update to Memcached 1.6.34. #10318
+* [ENHANCEMENT] Add support for individual components to override their container image. #10340
```
dimitarvdimitrov, line 33: i've been trying to make the changelog enough for anyone to get the gist of a change and know how to use it. Can you give an example of overriding the image for a component? Something like `ingester.image.tag` and `ingester.image.repository`. Or explain that there's an `image` object under all component objects now

operations/helm/charts/mimir-distributed/templates/_helpers.tpl
```diff
@@ -50,10 +50,20 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
-Calculate image name based on whether enterprise features are requested
+Build mimir image reference based on whether enterprise features are requested. The component local values always take precedence.
+Params:
+  ctx = . context
+  component = component name (optional)
```
dimitarvdimitrov, line 56: `componentSectionFromName` errors out if the component doesn't exist, so i think `component` isn't optional

dimitarvdimitrov (commented):
can you also include the keys in the values.yaml? Since we still want to use the top-level `image`, then you can make the new `<component>.image` be a comment. Something like the TSC examples 


https://github.com/grafana/mimir/blob/fc8af05a33dbe0965a7db4b4d3c0d89b0da60cf5/operations/helm/charts/mimir-distributed/values.yaml#L873-L883

that's basically the only documentation we have of helm chart features (plus the changelog, which is barely documentation)

operations/helm/charts/mimir-distributed/CHANGELOG.md
narqo: I've rephrased the entry with more details about the change.

narqo:
🆙 I've addressed the comments. PHAL

operations/helm/scripts/build.sh
```diff
@@ -83,7 +83,7 @@ function generate_manifests() {
   helm template "${ARGS[@]}" 1>/dev/null
   cp -r "${INTERMEDIATE_OUTPUT_DIR}" "${OUTPUT_DIR}"
   rm "${OUTPUT_DIR}/${CHART_NAME}/templates/values-for-rego-tests.yaml"
-  find "${OUTPUT_DIR}/${CHART_NAME}/templates" -type f -print0 | xargs -0 "${SED}" -E -i -- "/^\s+(checksum\/(alertmanager-fallback-)?config|(helm.sh\/)?chart|app.kubernetes.io\/version|image: \"grafana\/(mimir|mimir-continuous-test|enterprise-metrics)):/d"
+  find "${OUTPUT_DIR}/${CHART_NAME}/templates" -type f -print0 | xargs -0 "${SED}" -E -i -- "/^\s+(checksum\/(alertmanager-fallback-)?config|(helm.sh\/)?chart|app.kubernetes.io\/version|image: \"?grafana\/(mimir|mimir-continuous-test|enterprise-metrics)):/d"
```
narqo, line 86: This one fixes the build script, so it searched both quoted and not-quoted reference to the default image:

```
image: "grafana/mimir:
````

and

```
image: grafana/mimir:
```

(note the quote `"`)

That is, we don't want the generated manifests to have the default image, so we didn't have to re-generate them everytime the default image changes (e.g. on every weekly release).

File: operations/helm/charts/mimir-distributed/values.yaml [hunk 4]
```diff
@@ -551,6 +551,11 @@ alertmanager:
   #   E.g. if 'replicas' is set to 4 and there are 3 zones, then 4/3=1.33 and after rounding up it means 2 pods per zone are started.
   replicas: 1
 
+  # -- Allows to override the container image of the alertmanager component.
```
dimitarvdimitrov, line 554: ok, i may be getting annoying, but can you mention that this takes precedence over the root-level `image` and `enterprise.image` sections?


```suggestion
  # -- Allows to override the container image of the alertmanager component from the root-level `image` and `enterprise.image` sections.
```

something along those lines. You should be able to do a find-and-replace of the string.

dimitarvdimitrov (approved):
values.yaml being the only interface to the chart and the primary source of docs, I try to keep them as explicit as possible. You could also mention that per-component settings exist in the docs for the `image` and `enterprise.image` sections.

But I don't need to see this again. Thanks!

PR #10353, opened 2025-01-06
Extract delta from rate function

lamida:
This PR is trying to address suggestion in https://github.com/grafana/mimir/pull/9795 to extract delta function from rate. The diff will show how we have many more line additions if we extract the logic from rate.

How this different with #9795 
- We created a new step function called `delta` 
- Delta has no reset handling
- Delta will call `calculateFloatRate` and `calculateHistogramRate` to do extrapolation

My concern with extracting delta from rate is we create many line duplications, such as most of lines in delta(), floatDelta() and histogramDelta(). This is reflected from 118 line additions and 36 line removal in this PR diffs.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/functions/rate_increase.go
+++ b/pkg/streamingpromql/operators/functions/rate_increase.go
@@ -16,21 +16,21 @@ import (
 )
 
 var Rate = FunctionOverRangeVectorDefinition{
-	StepFunc:                       rate(true, true),
+	StepFunc:                       rate(true),
 	SeriesValidationFuncFactory:    rateSeriesValidator,
 	SeriesMetadataFunction:         DropSeriesName,
 	NeedsSeriesNamesForAnnotations: true,
 }
 
 var Increase = FunctionOverRangeVectorDefinition{
-	StepFunc:                       rate(true, false),
+	StepFunc:                       rate(false),
 	SeriesValidationFuncFactory:    rateSeriesValidator,
 	SeriesMetadataFunction:         DropSeriesName,
 	NeedsSeriesNamesForAnnotations: true,
 }
 
 var Delta = FunctionOverRangeVectorDefinition{
-	StepFunc:                       rate(false, false),
+	StepFunc:                       delta(),
 	SeriesMetadataFunction:         DropSeriesName,
 	NeedsSeriesNamesForAnnotations: true,
 }
@@ -39,8 +39,7 @@ var Delta = FunctionOverRangeVectorDefinition{
 // It calculates the rate (allowing for counter resets if isCounter is true for rate and increase),
 // extrapolates if the first/last sample is close to the boundary, and returns
 // the result as either per-second (if isRate is true) or overall (this is increase function).
-// If isCounter and isRate are both false, this function will calculate the delta.
-func rate(isCounter, isRate bool) RangeVectorStepFunction {
+func rate(isRate bool) RangeVectorStepFunction {
 	return func(step *types.RangeVectorStepData, rangeSeconds float64, emitAnnotation types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
 		fHead, fTail := step.Floats.UnsafePoints()
 		fCount := len(fHead) + len(fTail)
@@ -57,12 +56,12 @@ func rate(isCounter, isRate bool) RangeVectorStepFunction {
 
 		if fCount >= 2 {
 			// TODO: just pass step here? (and below)
-			val := floatRate(isCounter, isRate, fCount, fHead, fTail, step.RangeStart, step.RangeEnd, rangeSeconds)
+			val := floatRate(isRate, fCount, fHead, fTail, step.RangeStart, step.RangeEnd, rangeSeconds)
 			return val, true, nil, nil
 		}
 
 		if hCount >= 2 {
-			val, err := histogramRate(isCounter, isRate, hCount, hHead, hTail, step.RangeStart, step.RangeEnd, rangeSeconds, emitAnnotation)
+			val, err := histogramRate(isRate, hCount, hHead, hTail, step.RangeStart, step.RangeEnd, rangeSeconds, emitAnnotation)
 			if err != nil {
 				err = NativeHistogramErrorToAnnotation(err, emitAnnotation)
 				return 0, false, nil, err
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
 		lastCheckedMetricName = metricName
 	}
 }
+
+func delta() RangeVectorStepFunction {
+	return func(step *types.RangeVectorStepData, rangeSeconds float64, emitAnnotation types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
```
charleskorn, line 285: I think this can be simplified to this - there's no need for the wrapper function given we don't need different functionality like we do for `rate` and `increase`:

```suggestion
func delta(step *types.RangeVectorStepData, rangeSeconds float64, emitAnnotation types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
```

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+			return 0, false, val, nil
+		}
+
+		return 0, false, nil, nil
+	}
+}
+
+func floatDelta(fCount int, fHead []promql.FPoint, fTail []promql.FPoint, rangeStart int64, rangeEnd int64, rangeSeconds float64) float64 {
+	firstPoint := fHead[0]
+	fHead = fHead[1:]
```
charleskorn, line 320: [nit] We don't need to do this for `delta`.

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+		lastPoint = fHead[len(fHead)-1]
+	}
+
+	delta := lastPoint.F - firstPoint.F
+	return calculateFloatRate(false, false, rangeStart, rangeEnd, rangeSeconds, firstPoint, lastPoint, delta, fCount)
+}
+
+func histogramDelta(hCount int, hHead []promql.HPoint, hTail []promql.HPoint, rangeStart int64, rangeEnd int64, rangeSeconds float64, emitAnnotation types.EmitAnnotationFunc) (*histogram.FloatHistogram, error) {
+	firstPoint := hHead[0]
+	hHead = hHead[1:]
```
charleskorn, line 335: [nit] We don't need to do this for `delta`.

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+	}
+
+	if firstPoint.H.CounterResetHint == histogram.GaugeType || lastPoint.H.CounterResetHint == histogram.GaugeType {
+		emitAnnotation(annotations.NewNativeHistogramNotCounterWarning)
+	}
+
+	desiredSchema := firstPoint.H.Schema
+	if lastPoint.H.Schema < desiredSchema {
+		desiredSchema = lastPoint.H.Schema
+	}
```
charleskorn, line 351: [nit] This could be simplified to:

```suggestion
	desiredSchema := min(firstPoint.H.Schema, lastPoint.H.Schema)
```

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+	var lastPoint promql.HPoint
+	if len(hTail) > 0 {
+		lastPoint = hTail[len(hTail)-1]
+	} else {
+		lastPoint = hHead[len(hHead)-1]
+	}
+
+	if firstPoint.H.CounterResetHint == histogram.GaugeType || lastPoint.H.CounterResetHint == histogram.GaugeType {
+		emitAnnotation(annotations.NewNativeHistogramNotCounterWarning)
+	}
```
charleskorn, line 346: Is this correct? We emit a "not a gauge" warning further down if either point is not a gauge.

(we might need to add more test cases to `TestAnnotations` in `engine_test.go`)

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+
+	desiredSchema := firstPoint.H.Schema
+	if lastPoint.H.Schema < desiredSchema {
+		desiredSchema = lastPoint.H.Schema
+	}
+
+	usingCustomBuckets := firstPoint.H.UsesCustomBuckets()
+	if lastPoint.H.UsesCustomBuckets() != usingCustomBuckets {
```
charleskorn, line 354: [nit] This could be simplified to:

```suggestion
	if firstPoint.H.UsesCustomBuckets() != lastPoint.H.UsesCustomBuckets() {
```

pkg/streamingpromql/operators/functions/rate_increase.go
```diff
@@ -289,3 +280,94 @@ func rateSeriesValidator() RangeVectorSeriesValidationFunction {
...
+	if err != nil {
+		return nil, err
+	}
+	if firstPoint.H.CounterResetHint != histogram.GaugeType || lastPoint.H.CounterResetHint != histogram.GaugeType {
+		emitAnnotation(annotations.NewNativeHistogramNotGaugeWarning)
+	}
+
+	if delta.Schema != desiredSchema {
+		delta = delta.CopyToSchema(desiredSchema)
+	}
```
charleskorn, line 369: I don't think this is necessary - the `Sub` and `CopyToSchema` calls above should cover this.

charleskorn (commented):
I'm OK with the extra duplication this introduces - I think this makes the logic of `delta` much clearer.

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Added one test case in 56f67de31963295f8ec49aeef265e35032c7e9f0. Still figuring out how to set `CounterResetHint` from promql test.

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Removed in 175a67fa65d78b98b9951336ae161984fb8ebd01

lamida:
@charleskorn I have addressed almost all of your suggestion. I will merge this to #9795 and we can continue the discussion there.

pkg/streamingpromql/operators/functions/rate_increase.go
lamida: Found the way to set reset hint by setting something like `counter_reset_hint:gauge`. Added the test in https://github.com/grafana/mimir/pull/9795 in commit 4b170e7e62c7fdafa9713f94664c1317e7a47056

PR #10367, opened 2025-01-07
jsonnet: pass downscale-http-port from config

narqo:
#### What this PR does

The jsonnet bundle hardcodes the `prepare-downscale-http-port: 80` in all components, that use autoscaling and downscaling via the rollout-operator. This doesn't match the default HTTP port the bundle sets for the `*-http-metrics` container port, which is `:8080`.

Note we haven't noticed this previously, because internally all mimir components are run with `$._config.server_http_port=80`.

Fixes n/a

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


56quarters (approved):
Nice find

PR #10372, opened 2025-01-08
MQE: Tidy up pool size handling

jhesketh:
Primarily from feedback on https://github.com/grafana/mimir/pull/10261

Rather than allowing an arbitrary maxSize in pools, require them to be a power of two since that is what happens internally anyway.

We keep the checks and protections around requiring it to be a power of two.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/pool/bucketed_pool.go
+++ b/pkg/util/pool/bucketed_pool.go
@@ -84,6 +84,7 @@ func (p *BucketedPool[T, E]) Put(s T) {
 
 	bucketIndex := bits.Len(size - 1)
 	if bucketIndex >= len(p.buckets) {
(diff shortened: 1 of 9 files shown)


pkg/util/pool/bucketed_pool.go
```diff
@@ -86,11 +87,15 @@ func (p *BucketedPool[T, E]) Put(s T) {
 		return // Ignore slices larger than the largest bucket
```
charleskorn, line 87: Should we note here that we expect this will never happen given the `maxSize` check a few lines above and the fact that `maxSize` should be a power of two?

charleskorn (approved):
LGTM, thanks for addressing my feedback1

PR #10379, opened 2025-01-08
Exclude G301 gosec

dimitarvdimitrov:
#### What this PR does

> G301: Expect directory permissions to be 0750 or less

##### Summary from an internal discussion:

Mimir's permissions model relies on the OS umask to control file permissions for directories. The umask on Unix/Linux systems (typically 0022 or 0002) ensures directories created with os.ModePerm will have appropriate permissions (0755 or 0775). This design decision allows operators to customize permissions via umask while maintaining compatibility with existing deployments.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -71,9 +71,11 @@ linters-settings:
       - G109
       - G112
       - G114
-      - G301
       - G302
       - G401
+
+    excludes:
+      - G301 # Relies on system umask to restrict directory permissions, preserving operator flexibility
   
   gocritic:
     disable-all: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -71,6 +71,7 @@ linters-settings:
       - G109
       - G112
       - G114
+      - G301
       - G302
       - G401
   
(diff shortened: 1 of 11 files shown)


narqo (commented):
Shouldn't we mention these changes in the changelog?

Suppose someone runs mimir in a linux box under the `nobody` (`65534:65534`). After these changes, they won't be able to access the data on disk from another user, right?

dimitarvdimitrov:
Fair point, I added a changelog entry too

narqo (approved):
🔥 works for me

tacole02 (approved):

56quarters:
> > 
> 
> This PR ignores existing violations of this rule and will enforce it for future occurences.

Why are we enforcing this? My entire point is that the rule is wrong. We shouldn't be enforcing it on existing cases or in the future.

56quarters (approved):

narqo (approved):
🔥🔥 I like the updated version better

PR #10382, opened 2025-01-08
Allow dynamic replication set of TSDB blocks on store-gateways

### LINKED ISSUE #9944: Increase replication factor for some blocks on store-gateways
Each block owned by store-gateways is replicated to three store-gateways. When there are many queries that touch a particular block this can result in unbalanced CPU usage between store-gateways leading to higher costs. By far, recent blocks are queried more than older blocks.

From an internal cluster, we see that most queries only touch the most recent data:

*    ~92% of Select() calls that hit store-gateways touch data from the last 25h
*   ~50% of Select() calls that hit store-gateways touch data from the last 73h
*   Less than 1% of Select() calls that hit store-gateways touch data older than 28d ago
*   Less than 0.1% of Select() calls that hit store-gateways touch data older than 30d ago

In order to support spreading load for more recent blocks to more store-gateways, we should introduce the ability to override the configured replication factor (three by default) to something higher. The mechanism for picking overridden replication factor may be configurable or may be based on a variety of factors. 

This issue proposed to add the ability to override the replication factor and default behavior or doing this based on the age or duration of blocks and iterating on the exact behavior in further PRs.

_copied from an internal issue and discussion_

56quarters:
#### What this PR does

This change adds a setting to allow a subset of TSDB blocks to be replicated to more than the configured replication factor (three by default) on store-gateways. 

The `DynamicReplication` interface in `pkg/storegateway/dynamic_replication.go` is the core of this change. This new bit of logic determines if a block is eligible to be replicated over more than the default number of store-gateways. It is shared between store-gateways (which must do the replication) and queriers (which need to know which store-gateways to query). It defaults to only doing this for blocks that have samples within the last 25h (configurable). Most of the rest of the changes are using block metadata in places where we previously only used block IDs (`ulid.ULID`).

Assumptions made in this change:

* Default / dynamic replication is driven only by how recent the block is and if the setting is on, nothing fancy.
* Dynamic replication means replication to the entire user sub-ring. There are no degrees of extra replication.
* The threshold for dynamic replication is a global (not per-tenant) setting.

Fixes #9944

Related grafana/dskit#632

#### Notes for reviewers

The first commit is the main change to concentrate on which introduces an interface `DynamicReplication` and implementations which are shared between queriers and store-gateways. 

The second commit is changes to the `mimir-microservices-mode` docker-compose setup to make testing store-gateway changes easier.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


dimitarvdimitrov (commented):
I'm not sure what would happen when a block just crosses the 25h boundary. What if store-gateways drop it before the querier realizes that the block is that old (or the querier made the decision for that 1m ago for example)? You pointed out we retry a fixed 3 times. It's not impossible we retry the block on other store-gateways that have dropped it.

Maybe we should fix the retries before merging this PR. Charles suggested retrying as much as the replication factor (recent blocks we retry "NumInstances" times and others only 3 times)

56quarters:
> I'm not sure what would happen when a block just crosses the 25h boundary. What if store-gateways drop it before the querier realizes that the block is that old (or the querier made the decision for that 1m ago for example)? You pointed out we retry a fixed 3 times. It's not impossible we retry the block on other store-gateways that have dropped it.
> 
> Maybe we should fix the retries before merging this PR. Charles suggested retrying as much as the replication factor (recent blocks we retry "NumInstances" times and others only 3 times)

Good point. We could increase the number of retries but I'd rather solve this with some sort of "buffer time" like we do with the upload grace period. Maybe queriers could stop trying to query a block at `25h - $sync_interval * 3`  or something similar.

pkg/querier/blocks_store_replicated_set.go
```diff
@@ -97,31 +101,34 @@ func (s *blocksStoreReplicationSet) stopping(_ error) error {
...
 }
 
-func (s *blocksStoreReplicationSet) GetClientsFor(userID string, blockIDs []ulid.ULID, exclude map[ulid.ULID][]string) (map[BlocksStoreClient][]ulid.ULID, error) {
-	blocks := make(map[string][]ulid.ULID)
+func (s *blocksStoreReplicationSet) GetClientsFor(userID string, blocks bucketindex.Blocks, exclude map[ulid.ULID][]string) (map[BlocksStoreClient][]ulid.ULID, error) {
+	blocksByAddr := make(map[string][]ulid.ULID)
 	instances := make(map[string]ring.InstanceDesc)
 
 	userRing := storegateway.GetShuffleShardingSubring(s.storesRing, userID, s.limits)
+	expandedReplicationOption := ring.WithReplicationFactor(userRing.InstancesCount())
```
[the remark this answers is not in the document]
56quarters: "dynamic replication" ?

pkg/storegateway/expanded_replication.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (e *MaxTimeExpandedReplication) EligibleForSync(b ReplicatedBlock) bool {
+	return e.isMaxTimeInWindow(b, e.now())
+}
+
+func (e *MaxTimeExpandedReplication) EligibleForQuerying(b QueryableReplicatedBlock) bool {
+	now := e.now()
+	uploadedDelta := now.Sub(b.GetUploadedAt())
+	return uploadedDelta > e.gracePeriod && e.isMaxTimeInWindow(b, now)
```
56quarters, line 106: This grace period is for the time the block was uploaded but I think we also need one for the max time? If a block was uploaded hours ago, we still want it to be synced by the store-gateways before the queriers start expecting store-gateways to have it.

56quarters: Fixed via 450dda17db9b00035717da5e2fac05676ab21cbe

pkg/storegateway/expanded_replication.go
```diff
@@ -0,0 +1,110 @@
...
+	return maxTimeDelta <= e.maxTime
+}
+
+func (e *MaxTimeExpandedReplication) EligibleForQuerying(b QueryableReplicatedBlock) bool {
+	now := e.now()
+	uploadedDelta := now.Sub(b.GetUploadedAt())
+	maxTimeDelta := now.Sub(b.GetMaxTime())
+	// To be eligible for querying a block must:
+	// * Have been uploaded more than `gracePeriod` ago since we need to allow store-gateways
+	//   to sync recently uploaded blocks.
```
charleskorn, line 106: Why do we need to check the upload grace period here? Won't this be covered by the logic in `BlocksConsistency`?

56quarters: You're right that removing this wouldn't cause issues with queries failing or correctness. However, if we expand the replication set without the grace period we'd end up querying more store-gateways than we would otherwise that may not have the block. So by taking the grace period into account we avoid a little extra work for recently uploaded blocks. I can remove this in the interest of keeping the code simpler.

pkg/storegateway/expanded_replication.go
```diff
@@ -0,0 +1,100 @@
...
+	return maxTimeDelta <= e.maxTime
+}
+
+func (e *MaxTimeExpandedReplication) EligibleForQuerying(b ReplicatedBlock) bool {
+	now := e.now()
+	maxTimeDelta := now.Sub(b.GetMaxTime())
+	// To be eligible for querying a block must have a max time within `maxTime-gracePeriod` since
+	// we need to allow store-gateways to sync blocks that have recently become eligible for expanded
+	// replication.
+	return maxTimeDelta <= (e.maxTime - e.gracePeriod)
```
charleskorn, line 99: I'm wondering if we should apply the grace period as an additional amount of time the block is kept synced for, rather than reducing the amount of time the block is queried for - I think this matches more with what I'd expect given how the config is set up.

For example, let's say the expanded replication threshold is 25h, and the grace period is 15m.

If we apply the grace period as a reduction on the querying period, then we'll only query the block if it's within the last 24h45m, but it'll be available from store-gateways for 25h.

If we apply the grace period as an extension on the sync period, then we'll query the block if it's within the last 25h but it'll be available from store-gateways for 25h15m. This (to me at least) feels more in line with what I'd expect based on setting the threshold to 25h.

charleskorn (approved):
LGTM modulo suggestion about how we handle the grace period

pkg/storegateway/expanded_replication.go
56quarters: Fixed in https://github.com/grafana/mimir/pull/10382/commits/48d3d73772ea4301a3a37b4543f8f44d197846b1

pkg/storegateway/gateway.go
```diff
@@ -173,7 +179,17 @@ func newStoreGateway(gatewayCfg Config, storageCfg mimir_tsdb.BlocksStorageConfi
 		return nil, errors.Wrap(err, "create ring client")
 	}
 
-	shardingStrategy = NewShuffleShardingStrategy(g.ring, lifecyclerCfg.ID, lifecyclerCfg.Addr, limits, logger)
+	var dynamicReplication DynamicReplication = NewNopDynamicReplication()
+	if gatewayCfg.DynamicReplication.Enabled {
+		dynamicReplication = NewMaxTimeDynamicReplication(
+			gatewayCfg.DynamicReplication.MaxTimeThreshold,
+			// Exclude blocks which have recently become eligible for expanded replication, in order to give
+			// enough time to store-gateways to discover and load them (3 times the sync interval)
```
charleskorn, line 187: I think this comment is out of date - shouldn't it be something like "retain blocks for longer to ensure store-gateways don't unload them before queriers stop querying them"?

56quarters: :facepalm: Yep, I updated a comment in one place but not there. Will fix.

pkg/storegateway/expanded_replication.go
dimitarvdimitrov: > However, if we expand the replication set without the grace period we'd end up querying more store-gateways than we would otherwise that may not have the block

after the first iteration of the consistency retries, the consistency tracker would [exclude](https://github.com/grafana/mimir/blob/608d11dbc4ee1edd67e47ef6a5e7559f61a34032/pkg/querier/blocks_consistency_checker.go#L90-L97) the recently uploaded blocks from `missingBlocks` so we won't be querying them anymore because the [constructor of `BlocksConsistency` excluded them](https://github.com/grafana/mimir/blob/608d11dbc4ee1edd67e47ef6a5e7559f61a34032/pkg/querier/blocks_consistency_checker.go#L53-L58). But we'd still end up querying them once. Maybe what we should do is [reassign `knownBlocks` after we've created the `consistencyTracker`](https://github.com/grafana/mimir/blob/a979e2b34f638528ca48f781ae08b0aceed675b5/pkg/querier/blocks_store_queryable.go#L540-L545). 

That looks somewhat tangential to this though.

dimitarvdimitrov (approved):
one tiny comment on naming, but otherwise, this looks good 🚀

pkg/storegateway/expanded_replication.go
```diff
@@ -50,24 +57,42 @@ func NewNopExpandedReplication() *NopExpandedReplication {
...
 
 // MaxTimeExpandedReplication is an ExpandedReplication implementation that determines
 // if a block is eligible for expanded replication based on how recent its MaxTime (most
-// recent sample) is.
+// recent sample) is. An upload grace period can optionally be used to ensure that blocks
+// are synced to store-gateways before they are expected to be available by queriers.
 type MaxTimeExpandedReplication struct {
-	maxTime time.Duration
-	now     func() time.Time
+	maxTime     time.Duration
```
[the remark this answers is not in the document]
56quarters: > nit on naming: this is a duration, not a time

`maxTimeDelta` ?

> I thought max time refers to the timestamp in the block meta. WDYT about naming this filed something similar to oldestReplicated ?

The field comes from block meta in the bucket index but the value is from the samples in the block. From `bucketindex.Block`:

```go
// Block holds the information about a block in the index.
type Block struct {
    // <snip>

    // MinTime and MaxTime specify the time range all samples in the block are in (millis precision).
    MinTime int64 `json:"min_time"`
    MaxTime int64 `json:"max_time"`
    
    // <snip>
}
```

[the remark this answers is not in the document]
dimitarvdimitrov: i see. Then `maxTimeDelta` looks good.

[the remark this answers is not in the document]
dimitarvdimitrov: `maxTimeThreshold` is also good 👍

charleskorn (approved):
🚀

PR #10384, opened 2025-01-09
MQE: Add support to disable functions+aggregations by name

jhesketh:
Related to https://github.com/grafana/mimir/issues/10067


pkg/streamingpromql/config.go
```diff
@@ -5,32 +5,35 @@ package streamingpromql
...
 
+	"github.com/grafana/dskit/flagext"
 	"github.com/prometheus/prometheus/promql"
 )
 
 type EngineOpts struct {
-	CommonOpts     promql.EngineOpts
-	FeatureToggles FeatureToggles
+	CommonOpts promql.EngineOpts
+	MQEOpts    MQEOpts
```
charleskorn, line 14: How about we call this something like `Features`?

I was imagining that most MQE-specific config would go on the `EngineOpts` struct directly, like `Pedantic` below, and we keep feature flags in their own struct.

pkg/streamingpromql/config.go
```diff
@@ -40,17 +43,20 @@ var EnableAllFeatures = FeatureToggles{
...
+	f.BoolVar(&t.EnableVectorScalarBinaryComparisonOperations, "querier.mimir-query-engine.enable-vector-scalar-binary-comparison-operations", true, "Enable support for binary comparison operations between a vector and a scalar in the Mimir query engine. Only applies if the MQE is in use.")
+	f.BoolVar(&t.EnableVectorVectorBinaryComparisonOperations, "querier.mimir-query-engine.enable-vector-vector-binary-comparison-operations", true, "Enable support for binary comparison operations between two vectors in the Mimir query engine. Only applies if the MQE is in use.")
+
+	f.Var(&t.DisabledAggregations, "querier.mimir-query-engine.disabled-aggregations", "Comma-separated list of aggregations to disable. Queries using a disabled aggregation will fall back to Prometheus' query engine. Only applies if MQE is in use.")
+	f.Var(&t.DisabledFunctions, "querier.mimir-query-engine.disabled-functions", "Comma-separated list of function names to disable. Queries using a disabled function will fall back to Prometheus' query engine. Only applies if MQE is in use.")
```
charleskorn, line 61: [nit] Fallback will only occur if fallback is enabled, and also applies to all of the flags above, but we don't mention it above. I don't think we need to mention it in each flag description - perhaps if we reword the description here to something like "Comma-separated list of function names to disable support for" that will make the behaviour clear without having to mention fallback?

pkg/streamingpromql/engine_test.go
```diff
@@ -58,132 +58,142 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
...
 	for expression, expectedError := range unsupportedExpressions {
 		t.Run(expression, func(t *testing.T) {
-			requireQueryIsUnsupported(t, featureToggles, expression, expectedError)
+			requireQueryIsUnsupported(t, mqeOpts, expression, expectedError)
 		})
 	}
 }
 
-func TestUnsupportedPromQLFeaturesWithFeatureToggles(t *testing.T) {
+func TestUnsupportedPromQLFeaturesWithMQEOpts(t *testing.T) {
```
charleskorn, line 66: [nit] `TestUnsupportedPromQLFeaturesWithFeatureToggles` seems like a clearer name to me.

pkg/streamingpromql/operators/aggregations/common.go
```diff
@@ -35,3 +37,43 @@ var AggregationGroupFactories = map[parser.ItemType]AggregationGroupFactory{
 //
 // Invalid combinations include exponential and custom buckets, and histograms with incompatible custom buckets.
 var invalidCombinationOfHistograms = &histogram.FloatHistogram{}
+
+func IsAggregationDisabled(op parser.ItemType, disabledAggregations []string) bool {
+	for _, aggName := range disabledAggregations {
+		switch strings.ToLower(aggName) {
```
charleskorn, line 43: Doing `strings.ToLower` on every item in `disabledAggregations` every time we encounter an aggregation seems expensive. What if we converted the list of disabled aggregations to their corresponding `ItemType` in `NewEngine`? This would also give us an opportunity to reject any invalid aggregation names to catch typos etc.

Another optimisation would be to sort the list of disabled `ItemTypes` in `NewEngine` and then do [a binary search](https://pkg.go.dev/slices#BinarySearch) over the list to check if the aggregation is present.

pkg/streamingpromql/query.go
```diff
@@ -310,10 +313,11 @@ func (q *Query) convertToInstantVectorOperator(expr parser.Expr, timeRange types
...
 func (q *Query) convertFunctionCallToInstantVectorOperator(e *parser.Call, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
-	// Handle special toggles for classic histograms
-	if !q.engine.featureToggles.EnableHistogramQuantileFunction {
-		if e.Func.Name == "histogram_quantile" {
-			return nil, compat.NewNotSupportedError(fmt.Sprintf("'%s' function", e.Func.Name))
+	for _, fName := range q.engine.mqeOpts.DisabledFunctions {
+		// e.Func.Name is already validated by the parser. Meaning we don't need to check if the function name
+		// is real before checking if it is disabled.
+		if e.Func.Name == fName {
+			return nil, compat.NewNotSupportedError(fmt.Sprintf("'%s' function is disabled", e.Func.Name))
```
charleskorn, line 320: Using `slices.Contains` would make the intention clearer here, and if we sort `DisabledFunctions`, then we could use `slices.BinarySearch` to make this search faster as well.

charleskorn, line 320: In all other cases, we don't differentiate between a feature being unsupported entirely and a feature being disabled by a feature flag:

```suggestion
		if e.Func.Name == fName {
			return nil, compat.NewNotSupportedError(fmt.Sprintf("'%s' function", e.Func.Name))
```

pkg/streamingpromql/query.go
```diff
@@ -310,10 +313,11 @@ func (q *Query) convertToInstantVectorOperator(expr parser.Expr, timeRange types
 }
 
 func (q *Query) convertFunctionCallToInstantVectorOperator(e *parser.Call, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
```
charleskorn, line 315: Do we need to do the same thing for functions over range vectors?

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1496,47 +1496,54 @@ mimir_query_engine:
...
-  # if the MQE is in use.
-  # CLI flag: -querier.mimir-query-engine.enable-one-to-many-and-many-to-one-binary-operations
-  [enable_one_to_many_and_many_to_one_binary_operations: <boolean> | default = true]
+  # (experimental) Enable support for binary comparison operations between two
+  # vectors in the Mimir query engine. Only applies if the MQE is in use.
+  # CLI flag: -querier.mimir-query-engine.enable-vector-vector-binary-comparison-operations
+  [enable_vector_vector_binary_comparison_operations: <boolean> | default = true]
+
+  # (experimental) Comma-separated list of aggregations to disable. Queries
+  # using a disabled aggregation will fall back to Prometheus' query engine.
```
tacole02, line 1537:
```suggestion
  # using a disabled aggregation fall back to Prometheus' query engine.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1496,47 +1496,54 @@ mimir_query_engine:
...
+  # Only applies if MQE is in use.
+  # CLI flag: -querier.mimir-query-engine.disabled-aggregations
+  [disabled_aggregations: <string> | default = ""]
+
+  # (experimental) Comma-separated list of function names to disable. Queries
+  # using a disabled function will fall back to Prometheus' query engine. Only
```
tacole02, line 1543:
```suggestion
  # using a disabled function fall back to Prometheus' query engine. Only
```

tacole02 (approved):
Looks good. I removed a few instances of "will", as we try to avoid future tense in docs.

pkg/streamingpromql/query.go
jhesketh: I'm not concerned about the overhead here as it's just a few operations per query in comparison to the data we're processing. That said, it's easy enough so will make the optimisations.

jhesketh: I feel like this is a little more explicitly disabled so the log line is helpful

pkg/streamingpromql/query.go
jhesketh: Function over range vectors are done here. The output is just instant vectors.

charleskorn: 🤦 Of course they are, ignore me 🙂

pkg/streamingpromql/engine.go
```diff
@@ -44,12 +47,28 @@ func NewEngine(opts EngineOpts, limitsProvider QueryLimitsProvider, metrics *sta
 		return nil, errors.New("enabling delayed name removal not supported by Mimir query engine")
 	}
 
+	// Sort DisabledFunctions to optimise lookups
+	sort.Strings(opts.Features.DisabledFunctions)
```
charleskorn, line 51: `slices.Sort` is preferred as of Go 1.22 (`sort.Strings` just calls `slices.Sort` now)

pkg/streamingpromql/engine.go
```diff
@@ -44,12 +47,28 @@ func NewEngine(opts EngineOpts, limitsProvider QueryLimitsProvider, metrics *sta
...
+		item, ok := aggregations.GetAggregationItemType(agg)
+		if !ok {
+			return nil, errors.New(fmt.Sprintf("disabled aggregation '%s' does not exist", agg))
+		}
+		opts.Features.DisabledAggregationsItems = append(opts.Features.DisabledAggregationsItems, item)
+	}
+	// No point sorting DisabledAggregations earlier, as ItemType ints are not in order.
+	sort.Slice(opts.Features.DisabledAggregationsItems, func(i, j int) bool {
+		return opts.Features.DisabledAggregationsItems[i] < opts.Features.DisabledAggregationsItems[j]
+	})
```
charleskorn, line 64: This should be able to be simplified to `slices.Sort(opts.Features. DisabledAggregationsItems)`.

pkg/streamingpromql/config.go
```diff
@@ -5,32 +5,37 @@ package streamingpromql
...
+
+	DisabledAggregations      flagext.StringSliceCSV `yaml:"disabled_aggregations" category:"experimental"`
+	DisabledAggregationsItems []parser.ItemType
+	DisabledFunctions         flagext.StringSliceCSV `yaml:"disabled_functions" category:"experimental"`
 }
 
-// EnableAllFeatures enables all features supported by MQE, including experimental or incomplete features.
-var EnableAllFeatures = FeatureToggles{
+// MQEAllFeatures enables all features supported by MQE, including experimental or incomplete features.
+var MQEAllFeatures = Features{
```
charleskorn, line 38: Should we leave this as `EnableAllFeatures` now that the struct is called `Features`?

pkg/streamingpromql/operators/aggregations/common.go
```diff
@@ -35,3 +37,27 @@ var AggregationGroupFactories = map[parser.ItemType]AggregationGroupFactory{
 //
 // Invalid combinations include exponential and custom buckets, and histograms with incompatible custom buckets.
 var invalidCombinationOfHistograms = &histogram.FloatHistogram{}
+
+func GetAggregationItemType(aggregation string) (parser.ItemType, bool) {
+	// The parser key is not exported, but the item types are. It's safe to assume the keys will continue to map their names onto their values.
+	var key = map[string]parser.ItemType{
+		// Aggregators.
```
charleskorn, line 44: Is this comment necessary?

pkg/streamingpromql/operators/aggregations/common.go
```diff
@@ -35,3 +37,27 @@ var AggregationGroupFactories = map[parser.ItemType]AggregationGroupFactory{
 //
 // Invalid combinations include exponential and custom buckets, and histograms with incompatible custom buckets.
 var invalidCombinationOfHistograms = &histogram.FloatHistogram{}
+
+func GetAggregationItemType(aggregation string) (parser.ItemType, bool) {
+	// The parser key is not exported, but the item types are. It's safe to assume the keys will continue to map their names onto their values.
```
charleskorn, line 42: I'm not sure I follow this comment - are you trying to say that it's safe to assume that (for example) the string `sum` will continue to represent `parser.SUM`?

pkg/streamingpromql/query.go
charleskorn: Not sure I follow sorry - isn't it just as explicit as any other feature disabled with a feature flag?

pkg/streamingpromql/operators/aggregations/common.go
jhesketh: Yes.

pkg/streamingpromql/query.go
jhesketh: The others are "enable" which default to `true`. This explicitly disables something, which is useful to know in the logs imo.

pkg/streamingpromql/engine.go
```diff
@@ -44,12 +47,26 @@ func NewEngine(opts EngineOpts, limitsProvider QueryLimitsProvider, metrics *sta
 		return nil, errors.New("enabling delayed name removal not supported by Mimir query engine")
 	}
 
+	// Sort DisabledFunctions to optimise lookups
```
charleskorn, line 50: I would make this comment (and the one below) even clearer: we must sort these two slices because we use a binary search over them later on.

pkg/streamingpromql/engine_test.go
```diff
@@ -58,132 +58,153 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
...
+		features.DisabledAggregations = []string{"sum", "avg", "MAX"}
 
-		requireQueryIsUnsupported(t, featureToggles, "metric{} + on() group_left() other_metric{}", "binary expression with many-to-one matching")
-		requireQueryIsUnsupported(t, featureToggles, "metric{} + on() group_right() other_metric{}", "binary expression with one-to-many matching")
+		requireQueryIsUnsupported(t, features, "avg by (label) (metric{})", "'avg' aggregation disabled")
+		requireQueryIsUnsupported(t, features, "max(metric{})", "'max' aggregation disabled")
+		requireQueryIsUnsupported(t, features, "SUM(metric{})", "'sum' aggregation disabled")
+	})
+
+	t.Run("wrong aggregation name disabled", func(t *testing.T) {
```
charleskorn, line 183: [nit]

```suggestion
	t.Run("unknown aggregation name disabled", func(t *testing.T) {
```

pkg/streamingpromql/operators/aggregations/common.go
charleskorn: How about rephrasing this to something like this then?

```suggestion
	// The aggregation names are not exported, but their item types are. It's safe to assume the names will not change.
```

pkg/streamingpromql/operators/aggregations/common.go
```diff
@@ -35,3 +37,27 @@ var AggregationGroupFactories = map[parser.ItemType]AggregationGroupFactory{
...
+	// (ie, "avg" will be parser.AVG).
+	var key = map[string]parser.ItemType{
```
charleskorn, line 44: There's no need to create this map each time `GetAggregationItemType` is called - what if we move this outside this method?

pkg/streamingpromql/query.go
```diff
@@ -310,11 +313,10 @@ func (q *Query) convertToInstantVectorOperator(expr parser.Expr, timeRange types
...
-	// Handle special toggles for classic histograms
-	if !q.engine.featureToggles.EnableHistogramQuantileFunction {
-		if e.Func.Name == "histogram_quantile" {
-			return nil, compat.NewNotSupportedError(fmt.Sprintf("'%s' function", e.Func.Name))
-		}
+	// e.Func.Name is already validated by the parser. Meaning we don't need to check if the function name
+	// is real before checking if it is disabled.
```
charleskorn, line 317: [nit]

```suggestion
	// e.Func.Name is already validated and canonicalised by the parser. Meaning we don't need to check if the function name
	// refers to a function that exists, nor normalise the casing etc. before checking if it is disabled.
```

pkg/streamingpromql/query.go
jhesketh: I still think it's worthwhile, especially to differentiate between a function that is disabled verses one that we haven't got any support for (ie is missing from `instantVectorFunctionOperatorFactories`).

Not a hill I care for though if you want to veto this that's fine :-).

charleskorn: At present, we don't differentiate between those cases, and I'd prefer to remain consistent with what we already have.

charleskorn (approved):
LGTM, thanks for addressing my feedback!

PR #10383, opened 2025-01-08
Bump to latest mimir-prometheus

julienduchesne:
Version synced here: https://github.com/grafana/mimir-prometheus/pull/818

The only problem I had was with this PR from prometheus: https://github.com/prometheus/prometheus/pull/15686. I had to fix some tests and implement the same fix in our MQE `deriv` function: https://github.com/grafana/mimir/pull/10383/commits/cd4912e913fd792bb32b7d56008b257d7610a728

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10400, opened 2025-01-10
Update to latest mimir-prometheus

julienduchesne:
- Implement new rule concurrency interface from https://github.com/prometheus/prometheus/pull/15681
- MQE: Pull new functions test and fix _over_time functions
    Same fix as https://github.com/prometheus/prometheus/pull/15711 and similar to what was done during the last sync: https://github.com/grafana/mimir/pull/10383

### DIFF
--- a/pkg/ruler/rule_concurrency.go
+++ b/pkg/ruler/rule_concurrency.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/rules"
@@ -115,6 +116,7 @@ func NewMultiTenantConcurrencyController(logger log.Logger, maxGlobalConcurrency
 // NewTenantConcurrencyControllerFor returns a new rules.RuleConcurrencyController to use for the input tenantID.
 func (c *MultiTenantConcurrencyController) NewTenantConcurrencyControllerFor(tenantID string) rules.RuleConcurrencyController {
 	return &TenantConcurrencyController{
+		logger:                  log.With(c.logger, "tenant", tenantID),
 		slotsInUse:              c.metrics.SlotsInUse.WithLabelValues(tenantID),
 		attemptsStartedTotal:    c.metrics.AttemptsStartedTotal.WithLabelValues(tenantID),
 		attemptsIncompleteTotal: c.metrics.AttemptsIncompleteTotal.WithLabelValues(tenantID),
@@ -132,6 +134,7 @@ func (c *MultiTenantConcurrencyController) NewTenantConcurrencyControllerFor(ten
 // TenantConcurrencyController is a concurrency controller that limits the number of concurrent rule evaluations per tenant.
 // It also takes into account the global concurrency limit.
 type TenantConcurrencyController struct {
+	logger                   log.Logger
 	tenantID                 string
 	thresholdRuleConcurrency float64 // Percentage of the rule interval at which we consider the rule group at risk of missing its evaluation.
 
@@ -156,18 +159,6 @@ func (c *TenantConcurrencyController) Done(_ context.Context) {
 
 // Allow tries to acquire a slot from the concurrency controller.
 func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Group, rule rules.Rule) bool {
-	// To allow a rule to be executed concurrently, we need 3 conditions:
-	// 1. The rule group must be at risk of missing its evaluation.
-	// 2. The rule must not have any rules that depend on it.
-	// 3. The rule itself must not depend on any other rules.
-	if !c.isGroupAtRisk(group) {
-		return false
-	}
-
-	if !isRuleIndependent(rule) {
-		return false
-	}
-
 	// Next, try to acquire a global concurrency slot.
 	c.attemptsStartedTotal.Inc()
 	if !c.globalConcurrency.TryAcquire(1) {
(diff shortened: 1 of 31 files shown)


pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -67,17 +67,23 @@ func presentOverTime(step *types.RangeVectorStepData, _ float64, _ types.EmitAnn
...
 
-func maxOverTime(step *types.RangeVectorStepData, _ float64, _ types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
+func maxOverTime(step *types.RangeVectorStepData, _ float64, emitAnnotation types.EmitAnnotationFunc) (float64, bool, *histogram.FloatHistogram, error) {
 	head, tail := step.Floats.UnsafePoints()
 
 	if len(head) == 0 && len(tail) == 0 {
 		return 0, false, nil, nil
 	}
 
+	hHead, hTail := step.Histograms.UnsafePoints()
```
charleskorn, line 82: This could be simplified to use `step.Histograms.Any()`.

(same applies below and to the change for `deriv` in https://github.com/grafana/mimir/pull/10383/files)

charleskorn (commented):
MQE changes LGTM modulo suggestion to use `Any()`

pkg/streamingpromql/operators/functions/range_vectors.go
julienduchesne: Much nicer. Done!

jhesketh (approved):

pkg/ruler/rule_concurrency.go
```diff
@@ -187,6 +178,78 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
...
+		}
+		remainingRules[r] = info
+	}
+	if len(firstBatch) == 0 {
+		// There are no rules without dependencies.
+		// Fall back to sequential evaluation.
+		level.Info(logger).Log("msg", "No rules without dependencies found, falling back to sequential rule evaluation. This may be due to indeterminate rule dependencies.")
+		return nil
+	}
+	order := []rules.ConcurrentRules{firstBatch}
```
gotjosh, line 214:
```suggestion
	result := []rules.ConcurrentRules{firstBatch}
```

nit: this feels like a better name

pkg/ruler/rule_concurrency.go
```diff
@@ -187,6 +178,78 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
...
+			// We can't evaluate them concurrently.
+			// Fall back to sequential evaluation.
+			level.Warn(logger).Log("msg", "Cyclic rule dependencies detected, falling back to sequential rule evaluation")
+			return nil
+		}
+
+		order = append(order, batch)
+	}
+
+	level.Info(logger).Log("msg", "Batched rules into concurrent blocks", "rules", len(g.Rules()), "batches", len(order))
```
gotjosh, line 248: Can we add a debug line that tells us exactly how the batches look like?

pkg/ruler/rule_concurrency.go
```diff
@@ -187,6 +178,78 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
...
+	}
+
+	logger := log.With(c.logger, "group", g.Name())
+
+	type ruleInfo struct {
+		ruleIdx                 int
+		unevaluatedDependencies map[rules.Rule]struct{}
+	}
+	remainingRules := make(map[rules.Rule]ruleInfo)
+	firstBatch := rules.ConcurrentRules{}
```
gotjosh, line 194:
```suggestion
        // This batch holds the rules that have no dependencies and will be run first.
	noDependencyBatch := rules.ConcurrentRules{}
```

pkg/ruler/rule_concurrency.go
```diff
@@ -187,6 +178,78 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
...
 
+func (c *TenantConcurrencyController) SplitGroupIntoBatches(_ context.Context, g *rules.Group) []rules.ConcurrentRules {
+	if !c.isGroupAtRisk(g) {
+		// If the group is not at risk, we can evaluate the rules sequentially.
+		return nil
+	}
+
+	logger := log.With(c.logger, "group", g.Name())
+
+	type ruleInfo struct {
```
gotjosh, line 189: I think you can simplify this whole function and get rid of this struct directly by:

```go
func (c *TenantConcurrencyController) SplitGroupIntoBatches(_ context.Context, g *rules.Group) []rules.ConcurrentRules {
	if !c.isGroupAtRisk(g) {
		return nil
	}

	logger := log.With(c.logger, "group", g.Name())
	remainingRules := make(map[rules.Rule]int)
	firstBatch := rules.ConcurrentRules{}

	for i, r := range g.Rules() {
		if r.NoDependencyRules() {
			firstBatch = append(firstBatch, i)
		} else {
			remainingRules[r] = i
		}
	}

	if len(firstBatch) == 0 {
		level.Info(logger).Log("msg", "No rules without dependencies found, falling back to sequential rule evaluation.")
		return nil
	}

	order := []rules.ConcurrentRules{firstBatch}

	for len(remainingRules) > 0 {
		previousBatch := order[len(order)-1]
		for _, idx := range previousBatch {
			rule := g.Rules()[idx]
			for _, dependent := range rule.DependentRules() {
				delete(remainingRules, dependent)
			}
		}

		var batch rules.ConcurrentRules
		for r, idx := range remainingRules {
			if len(r.DependencyRules()) == 0 {
				batch = append(batch, idx)
				delete(remainingRules, r)
			}
		}

		if len(batch) == 0 {
			level.Warn(logger).Log("msg", "Cyclic rule dependencies detected, falling back to sequential rule evaluation")
			return nil
		}

		order = append(order, batch)
	}

	level.Info(logger).Log("msg", "Batched rules into concurrent blocks", "rules", len(g.Rules()), "batches", len(order))
	return order
}
```

gotjosh (commented):
LGTM, but please take a look at my comments.

pkg/ruler/rule_concurrency.go
```diff
@@ -187,6 +178,78 @@ func (c *TenantConcurrencyController) Allow(_ context.Context, group *rules.Grou
 	return false
 }
 
+func (c *TenantConcurrencyController) SplitGroupIntoBatches(_ context.Context, g *rules.Group) []rules.ConcurrentRules {
```
gotjosh, line 181:
```suggestion
// SplitGroupIntoBatches splits the group into batches of rules that can be evaluated concurrently.
// It tries to batch rules that have no dependencies together and rules that have dependencies in separate batches.
// Returning no batches means that the group should be evaluated sequentially.
func (c *TenantConcurrencyController) SplitGroupIntoBatches(_ context.Context, g *rules.Group) []rules.ConcurrentRules {
```

pkg/ruler/rule_concurrency.go
julienduchesne: Here, `if len(r.DependencyRules()) == 0` will never be true for a rule that has dependencies, because we don't modify the dependency rules attribute of rules.

Also, instead of modifying the dependencies of rules directly, they should be copied to another slice, so that we don't mutate the rules themselves. If we mutate the rules, this function can only run once

pkg/ruler/rule_concurrency.go
julienduchesne: Done! Looks like this:
```
{"batches":[[0,4,8,12,16,20,24,28,32,36,37,38,58],[39,42,18,30,56,33,41,45,6,9,10,21,26,46,13,17,29,1,51,55,14,52,34,40,2,5,22,25],[23,15,19,31,11,27,7,35,3],[43,44,54,49,50,57,47,48,53]],"component":"concurrency-controller","group":"test-group","level":"debug","msg":"Batched rules into concurrent blocks","tenant":"user1"}
```

gotjosh (approved):

PR #10426, opened 2025-01-13
Add `gocritic/dupImport` check

julienduchesne:
`gocritic` is too noisy to add as a whole, but we can add checks when we find non-controversial things that could've easily been fixed


pracucci (approved):
👍

PR #10430, opened 2025-01-13
Distributor: don't return errors when discarding samples with duplicated timestamps

duricanikolic:
#### What this PR does
In #10145 we introduced a request validation at distributor level that, for each object of type `mimirpb.PreallocTimeseries` in a request, iterates all its float and histogram samples, and as soon as a sample with a timestamp that has already been visited is found, discards the corresponding sample from the timeseries. Once the validation is completed, the request without discarded samples is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`.
One of the problems introduced by #10145 is that if a distributor discards at least one sample due to a duplicated timestamp, it also returns an error reporting the name of the timeseries, and the number of discarded samples. This behaviour introduces a confusion since
- on the one hand, the first sample with a certain timestamp that reaches a distributor is propagated to ingesters for further processing, and might be successfully ingested, 
- while on the other hand, all other samples belonging to the same timeseries and with the same timestamp as the first one will be rejected, and will return an error, making the request both successful and unsuccessful.

In order to remove this ambiguity, we decided to:
- Keep discarding samples with duplicated timestamps within the same `mimirpb.PreallocTimeseries` object on distributor side.
- Keep tracing the discarded samples in the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`.
- Not to return errors reporting the timeseries name and the number of discarded samples.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * [CHANGE] Ruler: cap the rate of retries for remote query evaluation to 170/sec. This is configurable via `-ruler.query-frontend.max-retries-rate`. #10375 #10403
 * [ENHANCEMENT] Query Frontend: Return server-side `samples_processed` statistics. #10103
 * [ENHANCEMENT] Distributor: OTLP receiver now converts also metric metadata. See also https://github.com/prometheus/prometheus/pull/15416. #10168
-* [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145 #10430
+* [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by `cortex_discarded_samples_total` metrics with the reason `sample_duplicate_timestamp`. #10145 #10430
 * [ENHANCEMENT] Ruler: Add `cortex_prometheus_rule_group_last_rule_duration_sum_seconds` metric to track the total evaluation duration of a rule group regardless of concurrency #10189
 * [ENHANCEMENT] Distributor: Add native histogram support for `electedReplicaPropagationTime` metric in ha_tracker. #10264
 * [ENHANCEMENT] Ingester: More efficient CPU/memory utilization-based read request limiting. #10325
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 * [CHANGE] Ruler: cap the rate of retries for remote query evaluation to 170/sec. This is configurable via `-ruler.query-frontend.max-retries-rate`. #10375 #10403
 * [ENHANCEMENT] Query Frontend: Return server-side `samples_processed` statistics. #10103
 * [ENHANCEMENT] Distributor: OTLP receiver now converts also metric metadata. See also https://github.com/prometheus/prometheus/pull/15416. #10168
-* [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145
+* [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by the `cortex_discarded_samples_total` metrics with reason `sample_duplicate_timestamp`. #10145 #10430
```
tacole02, line 13:
```suggestion
* [ENHANCEMENT] Distributor: discard float and histogram samples with duplicated timestamps from each timeseries in a request before the request is forwarded to ingesters. Discarded samples are tracked by `cortex_discarded_samples_total` metrics with the reason `sample_duplicate_timestamp`. #10145 #10430
```

tacole02 (approved):

pkg/distributor/distributor.go
```diff
@@ -899,7 +899,7 @@ func (d *Distributor) validateSeries(nowt time.Time, ts *mimirpb.PreallocTimeser
 	if deduplicatedSamplesAndHistograms > 0 {
 		d.sampleValidationMetrics.duplicateTimestamp.WithLabelValues(userID, group).Add(float64(deduplicatedSamplesAndHistograms))
 		unsafeMetricName, _ := extract.UnsafeMetricNameFromLabelAdapters(ts.Labels)
-		return false, fmt.Errorf(duplicateTimestampMsgFormat, deduplicatedSamplesAndHistograms, unsafeMetricName)
+		level.Warn(d.log).Log("msg", "samples with duplicated timestamps have been discarded", "series", fmt.Sprintf("'%.200s'", unsafeMetricName), "samples discarded", deduplicatedSamplesAndHistograms)
```
pracucci, line 902: I'm concerned we may log too much. This is per-series. In the worst case, this would be logged for every series in every write request, killing distributors.

duricanikolic: Hi @pracucci ,
We could then either remove logs or try to sample them, like we already do with ingesters, where in some cases we log only 10% of actual logs. WDYT?

duricanikolic: Hm... we actually don't sample logs but errors. So if we opt for sampling, we need to keep throwing errors, but log only a percentage of them.

duricanikolic: We agreed to remove logs for the time being, and to introduce rate limited logs in a follow-up PR.

Issue #10350, opened 2025-01-06
Flaky `TestHaTrackerWithMemberList`


dimitarvdimitrov:
Failed on a helm PR:

```
--- FAIL: TestHaTrackerWithMemberList (1.02s)
    ha_tracker_test.go:310: expected <nil>, got replicas did not match: r1 != r2
level=debug msg=Get key=prefixuser/cluster wait_index=0
level=debug msg="Get - not found" key=prefixuser/cluster
level=debug msg=CAS key=prefixuser/cluster modify_index=0 value="\"\\x15P\\n\\x05first\\x10告\\xd7\\xc32 告\\xd7\\xc32\""
level=debug msg=Get key=prefixuser/cluster wait_index=0
level=debug msg=Get key=prefixuser/cluster modify_index=2 value="\"\\x15P\\n\\x05first\\x10告\\xd7\\xc32 告\\xd7\\xc32\""
level=debug msg=CAS key=prefixuser/cluster modify_index=2 value="\"\\x18\\\\\\n\\x06second\\x10\\xf5ߊ\\xd7\\xc32 \\xf5ߊ\\xd7\\xc32(\\x01\""
level=debug msg=Get key=prefixuser/cluster wait_index=0
level=debug msg=Get key=prefixuser/cluster modify_index=3 value="\"\\x18\\\\\\n\\x06second\\x10\\xf5ߊ\\xd7\\xc32 \\xf5ߊ\\xd7\\xc32(\\x01\""
level=debug msg=CAS key=prefixuser/cluster modify_index=3 value="\"\\x17X\\n\\x05first\\x10\\x9d\\xa6\\x8b\\xd7\\xc32 \\x85\\xae\\x8b\\xd7\\xc32(\\x02\""
2025/01/06 09:11:52 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/01/06 09:11:52 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/01/06 09:11:52 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/01/06 09:11:52 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/01/06 09:11:52 label __name__ is overwritten. Check if Prometheus reserved labels are used.
level=info msg="server listening on addresses" http=127.0.0.1:38001 grpc=127.0.0.1:43873
[...]

PR #10436, opened 2025-01-14
ha_tracker_test: Increase test polling duration to 10s

NickAnge:
#### What this PR does

It seems that 2 seconds was not enough to prevent the test from failing again, so I would like to increase it to 10 seconds to address edge cases that require more time. 

Relates https://github.com/grafana/mimir/issues/10350

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥 Works for me

NickAnge:
Thanks @narqo . Could you please merge it because I do not have permissions ? Thanks in advance

PR #10445, opened 2025-01-15
Rewrite subqueries that have the same range and resolution

56quarters:
#### What this PR does

Fix queries that have the same range and resolution since this will only return a single point when using Prometheus 3 selectors, left open and right closed (compared to Prometheus 2 left closed and right closed).

Internal incident

#### Notes for reviewers

These names were the first things I thought of, please feel free to suggest better ones!

#### Checklist

- [X] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/prom2_range_compat.go
+++ b/pkg/frontend/querymiddleware/prom2_range_compat.go
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package querymiddleware
+
+import (
+	"context"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/tenant"
+	"github.com/prometheus/prometheus/promql/parser"
+
+	apierror "github.com/grafana/mimir/pkg/api/error"
+	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
+	"github.com/grafana/mimir/pkg/util/spanlogger"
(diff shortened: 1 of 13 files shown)


chencs (approved):
LGTM. Thanks for doing this!

ywwg (approved):
lgtm for the rewrites we want to do

pkg/frontend/querymiddleware/prom2_range_compat.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	rewrittenQuery := rewritten.String()
+	if origQuery != rewrittenQuery {
+		spanLog.DebugLog(
+			"msg", "modified subquery for compatibility with Prometheus 3 range selectors",
+			"original", origQuery,
+			"rewritten", rewrittenQuery,
+		)
+		r, _ = r.WithExpr(rewritten)
```
charleskorn, line 60: Might be good to add an annotation to the response when we rewrite the query, so the person writing the query can be nudged to adjust it.

Might also be useful to add a metric for the number of rewritten queries so we can know when it's safe to turn rewriting off.

charleskorn (commented):
Two small suggestions:

pkg/frontend/querymiddleware/prom2_range_compat.go
56quarters: Opened an issue to take care of this in another PR. Thanks!

PR #10446, opened 2025-01-15
chore: Fix gosec g401 in a handful of spots

alexweav:
#### What this PR does

Solves or suppresses existing violations of gosec g401 (none were used in a vulnerable way). This allows for enabling g401 on the codebase in the future.

n/a

#### Checklist
<already covered by tests, does not need docs or changelog>

### DIFF
--- a/pkg/ruler/mapper.go
+++ b/pkg/ruler/mapper.go
@@ -6,7 +6,7 @@
 package ruler
 
 import (
-	"crypto/md5"
+	"bytes"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -156,11 +156,9 @@ func (m *mapper) writeRuleGroupsIfNewer(groups []rulefmt.RuleGroup, filename str
 		if err != nil {
 			return false, err
 		}
-		newHash := md5.New()
-		currentHash := md5.New()
 
 		// bailout if there is no update
-		if string(currentHash.Sum(current)) == string(newHash.Sum(d)) {
+		if bytes.Equal(current, d) {
 			return false, nil
 		}
 	}
--- a/pkg/util/shard.go
+++ b/pkg/util/shard.go
@@ -24,7 +24,7 @@ var (
 // ShuffleShardSeed returns seed for random number generator, computed from provided identifier.
 func ShuffleShardSeed(identifier, zone string) int64 {
 	// Use the identifier to compute an hash we'll use to seed the random.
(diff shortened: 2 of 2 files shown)


pkg/ruler/mapper.go
```diff
@@ -156,11 +156,9 @@ func (m *mapper) writeRuleGroupsIfNewer(groups []rulefmt.RuleGroup, filename str
 		if err != nil {
 			return false, err
 		}
-		newHash := md5.New()
-		currentHash := md5.New()
 
 		// bailout if there is no update
-		if string(currentHash.Sum(current)) == string(newHash.Sum(d)) {
+		if bytes.Equal(current, d) {
```
alexweav, line 161: A quick benchmark on a medium-sized rule group file says a direct bytes compare is inconsequential in terms of timing:

```
BenchmarkRuleGroupBytesCompare-16                 	  237231	     50473 ns/op
BenchmarkRuleGroupBytesCompareAlreadyEqual-16     	  266635	     44484 ns/op
BenchmarkRuleGroupMD5Compare-16                   	  226880	     52006 ns/op
BenchmarkRuleGroupMD5CompareAlreadyEqual-16       	  263240	     44985 ns/op
BenchmarkRuleGroupSHA256Compare-16                	  228654	     51549 ns/op
BenchmarkRuleGroupSHA256CompareAlreadyEqual-16    	  262724	     44853 ns/op
```
IMO `bytes.Equal` is more expressive, the code says what it's trying to do.

pkg/util/shard.go
```diff
@@ -24,7 +24,7 @@ var (
 // ShuffleShardSeed returns seed for random number generator, computed from provided identifier.
 func ShuffleShardSeed(identifier, zone string) int64 {
 	// Use the identifier to compute an hash we'll use to seed the random.
-	hasher := md5.New()
+	hasher := md5.New()               //nolint:gosec
```
alexweav, line 27: The seed is consumed in a way that it's expected to be equal for the same tenant across different query frontends:
```
// seed for shuffle sharding of queriers; computed from tenantID only,
// and is therefore consistent between different frontends.
```
Changing it means this assumption is broken during the rollout, which could have unintended side effects. Otherwise it's only used internally to seed in a way that's stable based on tenant ID, we do not rely on specifically a cryptographic hash here.

seizethedave (approved):

Issue #10023, opened 2024-11-25
Idea: Cache subqueries the same way as range queries


julienduchesne:
### What is the problem you are trying to solve?

When querying an aggregated subquery as an instant query:

```promql
max_over_time({{query}}[7d:1h])
````

or the same ``{{query}}` as a range query, there is a dramatic different in performance. In my own use case, the aggregated instant query takes around 1m50s and the range query takes ~1-5s, while they are querying the same data (minus a max operation which should take microseconds)

### Which solution do you envision (roughly)?

This difference in runtime is due to the fact that range queries are incrementally cached, while subqueries (as part of instant queries) are not cached at all (see https://github.com/grafana/mimir/issues/3815)

Using the range query cache for subqueries could lead to performance improvements, especially for ruler queries which often do aggregations over subqueries and are all instant queries

### How long do you think this would take to be developed?

Not sure

julienduchesne:
Note that, following investigation, this is pretty tricky to do because we cannot arbitrarily inject results into promql queries. The range queries can be easily split and cached because this is done at the root level of the query. It is split by time intervals and recomposed at the end.

The instant query subqueries are embedded within the queries. That means that to inject data, one would need to do something similar to the sharded querying done in range queries where a special selector replaces the actual query: https://github.com/grafana/mimir/blob/main/pkg/frontend/querymiddleware/astmapper/embedded.go#L15-L32. It _may_ be possible but I'm not entirely sure

julienduchesne:
This is now possible through this experimental feature: https://github.com/grafana/mimir/pull/10460

Future improvements will come with the Mimir Query Engine (streamingpromql) but that's a while down the line

PR #10460, opened 2025-01-16
frontend: Add option to "spin off" subqueries as actual range queries

julienduchesne:
Issue: https://github.com/grafana/mimir/issues/10023

**Description:**

The feature was developed by inspiring myself from the query sharding feature. The queries are mapped into either downstream queries or subqueries. Both types of queries are run and the results are fed back into prometheus' engine and the result is calculated in the frontend.

The way it optimizes subqueries, is that range queries are passed back to the frontend and all of the range queries optimizations are applied (query sharding, splitting + caching)

**Configuration:**

This is a new feature that is completely isolated within a new middleware so it shouldn't affect current functionality of the frontend.
For safety, it requires two configurations to be enabled:
- `--query-frontend.spin-off-instant-subqueries-to-url=<url>` on the frontend. This should be set to the URL of the frontend for optimal performance. The range queries are load balanced across frontends
- `instant_queries_with_subquery_spin_off` in tenant configs. These are regexp patterns that allow us to match individual queries (or all of them). This will allow us to opt-in queries to enable the feature gradually

**Performance impact:**

The AST mapper only selects queries that are susceptible to be improved, others are just passed on to the next middleware.
For the queries that are improved, results can be up to 50x faster. When a query is selected, the worst cases I've seen are ~equal or a bit better in performance to unmodified queries.
Further tests will be done and the mapper may be improved to detect cases that aren't optimal

**PromQL results impact:**

None detected from all the tests I've done. The results were always identical.


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1723,6 +1723,12 @@ results_cache:
 # CLI flag: -query-frontend.use-active-series-decoder
 [use_active_series_decoder: <boolean> | default = false]
 
+# (experimental) If set, subqueries in instant queries will be spun off as range
```
tacole02, line 1726:
```suggestion
# (experimental) If set, subqueries in instant queries are spun off as range
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1723,6 +1723,12 @@ results_cache:
...
+# queries and sent to the given URL. Also requires
+# `instant_queries_with_subquery_spin_off` to be set for the tenant.
```
tacole02, line 1728:
```suggestion
# queries and sent to the given URL. This parameter also requires you to set
# `instant_queries_with_subquery_spin_off` for the tenant.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3599,6 +3605,11 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -query-frontend.prom2-range-compat
 [prom2_range_compat: <boolean> | default = false]
 
+# (experimental) List of regexp patterns matching instant queries. Subqueries
```
tacole02, line 3608:
```suggestion
# (experimental) List of regular expression patterns matching instant queries. Subqueries
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3599,6 +3605,11 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# within those instant queries will be spun off as range queries to optimize
```
tacole02, line 3609:
```suggestion
# within those instant queries are spun off as range queries to optimize
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3599,6 +3605,11 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# their execution.
```
tacole02, line 3610: Could we say `performance` instead of `execution`?

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,191 @@
...
+		return selector, false, nil
+	}
+
+	switch e := expr.(type) {
+	case *parser.Call:
+		if len(e.Args) == 0 {
+			return expr, false, nil
+		}
+		lastArgIdx := len(e.Args) - 1
+		if sq, ok := e.Args[lastArgIdx].(*parser.SubqueryExpr); ok {
```
chencs, line 81: Why do we only look at the last arg/expression here?

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,191 @@
...
+
+type subquerySpinOffMapper struct {
+	ctx             context.Context
+	defaultStepFunc func(rangeMillis int64) int64
+
+	logger log.Logger
+	stats  *SubquerySpinOffMapperStats
+}
+
+// NewSubqueryExtractor creates a new instant query mapper.
```
chencs, line 34: comment fn name not consistent with actual fn name

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, false, err
+	}
+
+	// Immediately clone the expr to avoid mutating the original
+	expr, err = cloneExpr(expr)
+	if err != nil {
+		return nil, false, err
+	}
+
+	downstreamQuery := func(expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
```
chencs, line 61: I _think_ I understand what this function is doing and how it's different from the `__subquery_spinoff__` selector, but it took me a minute. It could be helpful to readers in the future to add a comment here.

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,191 @@
...
+				VectorSelector: selector,
+				Range:          sq.Range,
+			}
+			m.stats.AddSpunOffSubquery()
+			return e, true, nil
+		}
+
+		return downstreamQuery(expr)
+	default:
+		// If there's no subquery in the children, we can just
```
chencs, line 137: Missing end of sentence?

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -0,0 +1,305 @@
...
+	qry, err := newQuery(ctx, req, s.engine, lazyquery.NewLazyQueryable(queryable))
+	if err != nil {
+		level.Warn(spanLog).Log("msg", "failed to create new query from splittable request", "err", err)
+		return nil, apierror.New(apierror.TypeBadData, err.Error())
+	}
+
+	res := qry.Exec(ctx)
+	extracted, err := promqlResultToSamples(res)
+	if err != nil {
+		level.Warn(spanLog).Log("msg", "failed to execute split instant query", "err", err)
```
chencs, line 228: Is "split instant query" a little confusing (since we also will split instant queries by interval)?

chencs (commented):
Overall, I think this makes sense. I think it is a first for Mimir to see `Queryable` implemented in the frontend, so it's worth getting another person to comment on that pattern.

julienduchesne:
> Overall, I think this makes sense. I think it is a first for Mimir to see `Queryable` implemented in the frontend, so it's worth getting another person to comment on that pattern.
c
Query sharding does the same

pkg/frontend/querymiddleware/spin_off_subqueries.go
julienduchesne: Copy pasta :|

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
julienduchesne: Added a docstring on the function

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -377,6 +381,20 @@ func newQueryMiddlewares(
 		newStepAlignMiddleware(limits, log, registerer),
 	)
 
+	if spinOffURL := cfg.SpinOffInstantSubqueriesToURL; spinOffURL != "" {
+		// Spin off subqueries to a remote URL (or localhost)
+		spinOffQueryHandler, err := newSpinOffQueryHandler(codec, log, spinOffURL)
+		if err != nil {
+			level.Error(log).Log("msg", "failed to create spin off query handler", "error", err)
+		} else {
+			queryInstantMiddleware = append(
```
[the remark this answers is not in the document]
julienduchesne: A few reasons:
- I saw limited improvement from range queries
- The implementation was also much more complex. I think that drilling down into range queries (and even `instant -> range -> subquery in range`) would be a better fit to be done directly in MQE directly
- Our biggest cells have a much higher ruler (instant) query rate than on the regular query path

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			return storage.ErrSeriesSet(errors.Wrap(err, "failed to parse subquery range"))
+		}
+		queryStep, err := time.ParseDuration(stepStr)
+		if err != nil {
+			return storage.ErrSeriesSet(errors.Wrap(err, "failed to parse subquery step"))
+		}
+
+		end := q.req.GetEnd()
+		// Align query to absolute time by querying slightly into the future.
```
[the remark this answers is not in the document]
julienduchesne: The `TestSubquerySpinOff_Correctness` tests this on every subquery that runs this way, because absolute time is _never_ aligned

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,202 @@
...
+	switch e := expr.(type) {
+	case *parser.Call:
+		if len(e.Args) == 0 {
+			return expr, false, nil
+		}
+		lastArgIdx := len(e.Args) - 1
+		// The last argument will typically contain the subquery in an aggregation function
+		// Examples: last_over_time(<subquery>[5m:]) or quantile_over_time(0.5, <subquery>[5m:])
+		if sq, ok := e.Args[lastArgIdx].(*parser.SubqueryExpr); ok {
+			// Filter out subqueries with offsets, not supported yet
```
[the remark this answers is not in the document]
julienduchesne: Good catch. I skip them now (`@ end()`, `@ start()` and `@ <timestamp>`)

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -0,0 +1,189 @@
...
+		// Split queries into multiple smaller queries if they have more than 10000 datapoints
+		rangeStart := start
+		var rangeQueries []MetricsQueryRequest
+		for {
+			var rangeEnd int64
+			if remainingPoints := (end - start) / step; remainingPoints > 10000 {
+				rangeEnd = start + 10000*step
+			} else {
+				rangeEnd = end
+			}
```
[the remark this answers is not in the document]
julienduchesne: This value comes from the codec, it doesn't encode/decode more than 11000 points, I've changed this to use a constant instead, and I added a correctness test that checks that this behaviour works correctly, I had a mistake in my loop (which explains some results from my hackathon)

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -0,0 +1,189 @@
...
+			headers := q.req.GetHeaders()
+			headers = append(headers,
+				&PrometheusHeader{Name: "Content-Type", Values: []string{"application/json"}},
```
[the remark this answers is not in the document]
julienduchesne: I've removed this header, the codec adds the appropriate headers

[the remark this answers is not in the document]
julienduchesne: I've removed this header, the codec adds the appropriate headers

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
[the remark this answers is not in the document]
julienduchesne: Done!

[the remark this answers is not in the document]
julienduchesne: I've added a bunch of query end (== start because they're instant queries) times to the correctness tests (-10m, -33s, no offset, +33s, +10m, +1h), so this should cover this in all cases

pkg/frontend/querymiddleware/limits.go
```diff
@@ -116,6 +116,10 @@ type Limits interface {
 
 	// IngestStorageReadConsistency returns the default read consistency for the tenant.
 	IngestStorageReadConsistency(userID string) string
+
+	// InstantQueriesWithSubquerySpinOff returns a list of regexp patterns of instant queries that can be optimized by spinning off range queries.
+	// If the list is empty, the feature is disabled.
+	InstantQueriesWithSubquerySpinOff(userID string) []string
```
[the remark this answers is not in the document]
julienduchesne: I'd like to leave it as-is because it might otherwise get complicated. I want to gradually add more queries to it, without modifying the current inclusions, the slice works better for that.

For the regexp matcher, I use prometheus' `FastRegexMatcher` (https://github.com/prometheus/prometheus/blob/7bc11dcb0664/pkg/labels/regexp.go#L29), just like the blocked queries feature. It has a cache: https://github.com/grafana/mimir-prometheus/pull/465

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -110,62 +110,64 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
...
+		var queryOffset time.Duration
+		if offsetStr == "" {
+			queryOffset = 0
+		} else if queryOffset, err = time.ParseDuration(offsetStr); err != nil {
+			return storage.ErrSeriesSet(errors.Wrap(err, "failed to parse subquery offset"))
+		}
 
-		end := q.req.GetEnd()
-		// Align query to absolute time by querying slightly into the future.
+		end := q.req.GetEnd() - queryOffset.Milliseconds()
```
charleskorn, line 120: Might be worth adding a note here to make it clear that this would need to change if we want to support range queries in addition to instant queries.

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -110,62 +110,64 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
...
+		// Subqueries are aligned on absolute time, while range queries are aligned on relative time.
```
charleskorn, line 121: [nit] This seems clearer to me:

```suggestion
		// Subqueries are always aligned to absolute time in PromQL, so we need to make the same adjustment here for correctness.
```

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -110,62 +110,64 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
...
+		// To match the same behavior, we can query slightly into the future.
+		// The extra data points aren't used because the subquery is aggregated with an _over_time-style function.
```
charleskorn, line 123: Is this statement correct? Isn't the adjustment below small enough that it only introduces an additional point, and this is filtered out during evaluation because the PromQL engine (Prometheus' or MQE) will only include points from `start` to `start+queryRange`? 

Alternatively, we could use the same time range logic MQE uses, which avoids querying into the future: https://github.com/grafana/mimir/blob/266a393379b2c981a83557c5d66e56c97251ffeb/pkg/streamingpromql/query.go#L384-L398

pkg/frontend/querymiddleware/spin_off_subqueries_test.go
```diff
@@ -29,13 +28,11 @@ import (
...
-		numStaleSeries           = 100
-		numConvHistograms        = 1000
-		numStaleConvHistograms   = 100
-		histogramBuckets         = []float64{1.0, 2.0, 4.0, 10.0, 100.0, math.Inf(1)}
-		numNativeHistograms      = 1000
-		numStaleNativeHistograms = 100
+		numSeries      = 1000
+		numStaleSeries = 100
+		samplesStart   = time.Now().Add(-2 * 24 * time.Hour)
+		samplesEnd     = time.Now().Add(30 * time.Minute)
```
charleskorn, line 34: It's possible that `time.Now()` returns slightly different results between these two lines being evaluated - perhaps it would be safer to define `samplesEnd` in terms of `samplesStart`, or store "now" once and use it in both places.

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
```diff
@@ -0,0 +1,207 @@
...
+	switch e := expr.(type) {
+	case *parser.Call:
+		if len(e.Args) == 0 {
+			return expr, false, nil
+		}
+		lastArgIdx := len(e.Args) - 1
+		// The last argument will typically contain the subquery in an aggregation function
+		// Examples: last_over_time(<subquery>[5m:]) or quantile_over_time(0.5, <subquery>[5m:])
+		if sq, ok := e.Args[lastArgIdx].(*parser.SubqueryExpr); ok {
+			// @ is not supported
```
charleskorn, line 93: [not blocking] I don't think it would be too complex to support `@` - it effectively changes the end timestamp used in the query range calculations.

pkg/frontend/querymiddleware/roundtrip.go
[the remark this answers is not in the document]
charleskorn: Makes sense, might be adding a comment explaining that we're deliberately only doing this for instant queries.

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -0,0 +1,315 @@
...
+
+	matched := false
+	for _, tenantID := range tenantIDs {
+		patterns := s.limits.InstantQueriesWithSubquerySpinOff(tenantID)
+
+		for _, pattern := range patterns {
+			matcher, err := labels.NewFastRegexMatcher(pattern)
+			if err != nil {
+				level.Error(spanLog).Log("msg", "failed to compile regex pattern", "pattern", pattern, "err", err)
+				continue
```
charleskorn, line 139: Should we just return an error in this case, rather than logging it and continuing? Otherwise it could be quite difficult to uncover the fact that the pattern is invalid.

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -0,0 +1,315 @@
...
+			}
+		}
+
+		if matched {
+			break
+		}
+	}
+
+	if !matched {
+		spanLog.DebugLog("msg", "subquery spin off is disabled for this query")
```
charleskorn, line 154:
```suggestion
		spanLog.DebugLog("msg", "expression did not match any configured subquery spin off patterns, so subquery spin off is disabled for this query")
```

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		level.Warn(spanLog).Log("msg", "failed to parse query", "err", err)
+		s.metrics.spinOffSkipped.WithLabelValues(subquerySpinoffSkippedReasonParsingFailed).Inc()
+		return nil, apierror.New(apierror.TypeBadData, DecorateWithParamName(err, "query").Error())
+	}
+
+	spinOffQuery, err := mapper.Map(expr)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) && ctx.Err() == nil {
+			level.Error(spanLog).Log("msg", "timeout while spinning off subqueries, please fill in a bug report with this query, falling back to try executing without spin-off", "err", err)
```
charleskorn, line 176: [nit] In some places we use "spin off" and in others we use "spin-off" - would be good to be consistent about this

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -0,0 +1,315 @@
...
+	// Override the URL with the configured range query URL (only parts that are set).
+	if s.rangeQueryURL.Scheme != "" {
+		httpReq.URL.Scheme = s.rangeQueryURL.Scheme
+	}
+	if s.rangeQueryURL.Host != "" {
+		httpReq.URL.Host = s.rangeQueryURL.Host
+	}
+	if s.rangeQueryURL.Path != "" {
+		httpReq.URL.Path = s.rangeQueryURL.Path
+	}
```
charleskorn, line 294: This behaviour seems a bit confusing to me - why not just always use the values that have been configured? 

For example, if the configured URL is `http://foo`, then we'll always preserve the path from the original request, but this may not be what the user intended - they may want to target the root of `foo`.

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -0,0 +1,191 @@
...
+			if remainingPoints := (end - rangeStart) / step; remainingPoints > maxResolutionPoints {
+				rangeEnd = rangeStart + maxResolutionPoints*step
+			} else {
+				rangeEnd = end
+			}
+			headers := q.req.GetHeaders()
+			headers = append(headers,
+				&PrometheusHeader{Name: "X-Mimir-Spun-Off-Subquery", Values: []string{"true"}},
+				&PrometheusHeader{Name: "Content-Type", Values: []string{"application/x-www-form-urlencoded"}},
+			) // Downstream is the querier, which is HTTP req.
```
charleskorn, line 145: This comment seems out of place to me - is it still relevant?

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -0,0 +1,191 @@
...
+			rangeQueries = append(rangeQueries, newRangeRequest)
+			if rangeEnd == end {
+				break
+			}
+			rangeStart = rangeEnd // Move the start to the end of the previous range.
+		}
+
+		streams := make([][]SampleStream, len(rangeQueries))
+		for idx, req := range rangeQueries {
+			resp, err := q.upstreamRangeHandler.Do(ctx, req)
```
charleskorn, line 156: Will these requests run concurrently? If not, seems like it might be worthwhile to do so.

julienduchesne: Reason: I don't want to unexpectedly blow up the query path

The case where the steps will go over 11000 is when querying far into the past, since even with a 30s resolution, 11000 steps is a 92h range, so most queries are supported with a single range query

If we always have only one inflight query, there's no risk of killing the query path, because if someone does an aggregation for a whole year, then it could potentially spawn dozens of range queries. By doing one at a time, we'll eventually time out the query instead of killing Mimir.

pkg/frontend/querymiddleware/astmapper/subquery_spin_off.go
julienduchesne: I haven't seen it used much overall. I'll keep it out for now

pkg/frontend/querymiddleware/spin_off_subqueries.go
julienduchesne: Indeed. I was scared of invalid regexp blocking the whole query path, but it's actually pretty hard to have an invalid regexp in golang. Should be fine erroring out

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
julienduchesne: Nice! The MQE logic is better than mine. I replaced it

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
chencs: Did these kinds of queries (the ones that would time out when spun-off) timeout as instant queries as well?

chencs (approved):
I'm satisfied for this to go in as experimental (with one non-blocking question left)

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
julienduchesne: The top level instant query would timeout first, cancelling the context, it would then stop both downstream instant queries, and spun off range queries from being sent

PR #10462, opened 2025-01-16
partitionOffsetClient: Add topic label to metrics

leizor:
Add `topic` as a constant label for partitionOffsetClient's metrics. Without this, trying to use the partitionOffsetClient for multiple topics at once results in a panic.


replay (approved):
thx!

PR #10467, opened 2025-01-17
Upgrade mimir-prometheus

charleskorn:
#### What this PR does

This PR upgrades mimir-prometheus to include https://github.com/grafana/mimir-prometheus/pull/820.

I haven't added a changelog entry given this is a minor tracing improvement.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/vendor/github.com/prometheus/prometheus/tsdb/postings_for_matchers_cache.go
+++ b/vendor/github.com/prometheus/prometheus/tsdb/postings_for_matchers_cache.go
@@ -141,6 +141,9 @@ type postingsForMatcherPromise struct {
 	// callers contexts get canceled.
 	callersCtxTracker *contextsTracker
 
+	// Keep track of the time this promise was evaluated, so we can understand the age of this cache entry in traces.
+	evaluatedAt time.Time
+
 	// The result of the promise is stored either in cloner or err (only of the two is valued).
 	// Do not access these fields until the done channel is closed.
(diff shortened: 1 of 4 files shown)


vendor/github.com/prometheus/prometheus/tsdb/postings_for_matchers_cache.go
```diff
@@ -241,7 +250,7 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 
 	sizeBytes := int64(len(key) + size.Of(promise))
 
-	c.onPromiseExecutionDone(ctx, key, c.timeNow(), sizeBytes, promise.err)
+	c.onPromiseExecutionDone(ctx, key, ts, sizeBytes, promise.err)
```
pracucci, line 253: This subtle change could be quite impactful. It was done on purpose reading `c.timeNow()` at this point, to make the cache TTL lasting longer. The problem is that if you're running a very complex index lookup that takes few seconds, by the time you reach this point, the TTL of 10s may be either already expired or close to expire, effectively invalidating the benefits of the cache.

charleskorn: Good catch, I'll revert this and find another way to do this.

PR #10484, opened 2025-01-20
frontend: Allow blocking raw http requests

julienduchesne:
#### What this PR does

This is a setting that can be set in user overrides like this:
```
user1:
  blocked_requests:
    - path: /api/v1/series
      method: GET
      query_params:
        foo: bar
```

Or a combination of these. Each entry is an AND condition

`BlockedRequest` could be expanded to support headers, body content, etc, but we don't need these feature right now.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/frontend/querymiddleware/request_blocker.go
```diff
@@ -0,0 +1,84 @@
...
+		for _, blockedRequest := range blockedRequests {
+			if blockedPath := blockedRequest.Path; blockedPath != "" && blockedPath != r.URL.Path {
+				continue
+			}
+
+			if blockedMethod := blockedRequest.Method; blockedMethod != "" && blockedMethod != r.Method {
+				continue
+			}
+
+			if blockedParams := blockedRequest.QueryParams; len(blockedParams) > 0 {
```
pracucci, line 62: Food for thought. For most query endpoints we support POST too. For example, for instant and range queries Grafana issue POST requests instead of GET. For `/series` from Grafana metrics browser it looks like Grafana issues GET requests.

A follow up on this work (not in this PR) could be extending the support to parse POST body too, similar to what we do in `pkg/frontend/transport/handler.go` using `util.ParseRequestFormWithoutConsumingBody()`.

pkg/frontend/querymiddleware/request_blocker.go
```diff
@@ -0,0 +1,84 @@
...
+				query := r.URL.Query()
+				blockedByParams := false
+				for key, blockedValue := range blockedParams {
+					if query.Get(key) == blockedValue {
```
pracucci, line 66: Have you considered the value being a regexp instead, to make it more flexible? For queries blocker we have a flag to set whether the value is a regex or not, maybe we can do something similar here as well. Having a regexp could use useful to block patterns.

pracucci (approved):
Very nice work! LGTM. Please remember a CHANGELOG entry before merging. I left a couple of comments, but they're non blocking.

pkg/frontend/querymiddleware/request_blocker.go
julienduchesne: Yep! Nothing stops us from adding support for headers, body, etc with this format

pkg/frontend/querymiddleware/request_blocker.go
julienduchesne: Done! Made the thing a struct instead so that both are supported, the same way as blocked queries

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * [ENHANCEMENT] Query-frontend: include more information about read consistency in trace spans produced when using experimental ingest storage. #10412
 * [ENHANCEMENT] Ingester: Hide tokens in ingester ring status page when ingest storage is enabled #10399
 * [ENHANCEMENT] Ingester: add `active_series_additional_custom_trackers` configuration, in addition to the already existing `active_series_custom_trackers`. The `active_series_additional_custom_trackers` configuration allows you to configure additional custom trackers that get merged with `active_series_custom_trackers` at runtime. #10428
+* [ENHANCEMENT] Query-frontend: Allow blocking raw http requests with the `blocked_requests` configuration. Requests can be blocked based on their path, method
```
pracucci, line 28: Looks changelog has been truncated here.

julienduchesne: I have fixed it here: https://github.com/grafana/mimir/pull/10497

PR #10485, opened 2025-01-20
allow setting different annotations for headless service in the distributor

KyriosGN0:
#### What this PR does

this PR allows users to specify annotation to the distributor headless service alone (meaning without also setting them on the actual clusterIP service)

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -32,6 +32,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [CHANGE] Memcached: Update to Memcached 1.6.34. #10318
 * [ENHANCEMENT] Minio: update subchart to v5.4.0. #10346
 * [ENHANCEMENT] Individual mimir components can override their container images via the *.image values. The component's image definitions always override the values set in global `image` or `enterprise.image`. #10340
+* [ENHANCEMENT] allow setting different annotations for headless service in the distributor #10485
 * [BUGFIX] Fix calculation of `mimir.siToBytes` and use floating point arithmetics. #10044
 
 ## 5.6.0
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -32,7 +32,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [CHANGE] Memcached: Update to Memcached 1.6.34. #10318
 * [ENHANCEMENT] Minio: update subchart to v5.4.0. #10346
 * [ENHANCEMENT] Individual mimir components can override their container images via the *.image values. The component's image definitions always override the values set in global `image` or `enterprise.image`. #10340
-* [ENHANCEMENT] allow setting different annotations for headless service in the distributor #10485
+* [ENHANCEMENT] Allow setting different annotations for headless service in the distributor #10485
 * [BUGFIX] Fix calculation of `mimir.siToBytes` and use floating point arithmetics. #10044
 
 ## 5.6.0
--- a/operations/helm/charts/mimir-distributed/templates/distributor/distributor-svc-headless.yaml
+++ b/operations/helm/charts/mimir-distributed/templates/distributor/distributor-svc-headless.yaml
@@ -10,7 +10,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
   annotations:
-    {{- toYaml .Values.distributor.service.annotations | nindent 4 }}
+    {{- toYaml .Values.distributor.service.headlessAnnotations | nindent 4 }}
   namespace: {{ .Release.Namespace | quote }}
 spec:
   type: ClusterIP
--- a/operations/helm/charts/mimir-distributed/values.yaml
+++ b/operations/helm/charts/mimir-distributed/values.yaml
@@ -853,6 +853,7 @@ distributor:
 
   service:
     annotations: {}
+    headlessAnnotations: {}
     labels: {}
     # -- https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/
     internalTrafficPolicy: Cluster


operations/helm/charts/mimir-distributed/templates/distributor/distributor-svc-headless.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
   annotations:
-    {{- toYaml .Values.distributor.service.annotations | nindent 4 }}
+    {{- toYaml .Values.distributor.service.headlessAnnotations | nindent 4 }}
```
narqo, line 13: This is a breaking change. The setups that are currently in production, will stop working after they update to this one.

narqo (commented):
Appreciate the PR. As I mentioned in the comment below, the current changes seem to break the backwards compatibility.

It's clear what the changes here are doing, but I don't understand the problem that is being solved. That is, what is the use-cases where one needs a special set of annotations only in one distributor's service?

KyriosGN0:
@narqo for example we want to set the distributor service as load balancer with external-dns annotations, because both the headless service loads the same annotations (i.e. `external-dns.alpha.kubernetes.io/hostname`), external dns gives that record 2 IPs, one which point to the load balancer service and the other to the headless service (which is cluster ip) this causes us to sometimes get connection error when hitting that ip through dns 

i understand this a breaking change, is there a special procedure i need to follow in order to properly announce it (assuming it is ok withe that maintainers)

narqo:
So the issue is that you need to segregate the annotations of the clusterIP and the headless services, to be able to expose the former to an external LB? The `gateway` component is what the chart comes with for this purpose. This deploys an API gateway (the Mimir open-source uses nginx), whose service you can expose to the LB. This may solve it for you without needing for a breaking change in the chart.

In future major version of the chart we should revisit the layout of the components and their values. For now, though, we're restricting ourselves to avoid any breaking changes. If you can think of a way to solve what you were trying to archive w/o breaking the existing code, we can iterate on that.

KyriosGN0:
@narqo i can wrap it with another boolean value to opt in to this behavior, is that acceptable ? if not i'll open an issue and will wait until the next major version

operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [CHANGE] Memcached: Update to Memcached 1.6.34. #10318
 * [ENHANCEMENT] Minio: update subchart to v5.4.0. #10346
 * [ENHANCEMENT] Individual mimir components can override their container images via the *.image values. The component's image definitions always override the values set in global `image` or `enterprise.image`. #10340
+* [ENHANCEMENT] allow setting different annotations for headless service in the distributor #10485
```
tacole02, line 35:
```suggestion
* [ENHANCEMENT] Allow setting different annotations for headless service in the distributor #10485
```

tacole02 (approved):

dimitarvdimitrov:
I'd opt for doing it in a non-breaking manner because the next major release is still not planned. Can we for example make the `headlessAnnotations` be in addition to `annotations` instead of a replacement?

KyriosGN0:
@dimitarvdimitrov this doesn't solve the use case that i have (and we don't want to use nginx, we want to use the native services and ingress)
i can wrap it in a flag to opt in to the behavior, if not i'll wait until the next major release

dimitarvdimitrov:
I'm thinking of users who are already setting `annotations`. In your case you will set `annotations: {}`. `headlessAnnotations` will still be added to the headless service _only_. So the rendered manifests will be the same as the ones that the current version of this PR would create. Am I missing something? 👀

KyriosGN0:
@dimitarvdimitrov from what i understand it might be problemtic to users who relay on the both the headless service and clusterIP service having the same annotations, even with the my current changes they can achive this behavior, they can just use the same annotations in headlessAnnotations


hoping im making sense 🙏

dimitarvdimitrov:
I'd like to summarize to make sure we're on the same page. I see two distinct use cases:

1. **Existing users**: 
   - They currently have two services (regular and headless) with identical annotations
   - Both services use `distributor.service.annotations`
   - They do not want to change annotations on their services
   - They need a zero-migration path when upgrading without modifying their values files

2. **Users needing differentiated annotations (like yourself)**:
   - Need to set different annotations for the headless service versus the regular service
   - Want to prevent connection errors when hitting services through DNS
   - The annotations of the headless service must be a superset of the annotations of the regular service

My proposed solution:
- For the non-headless (regular) service: Keep using `distributor.service.annotations` (no change)
- For the headless service: Use a merged combination of both `distributor.service.annotations` AND `distributor.service.headlessAnnotations`

This way:
- The normal service retains its original annotations (backward compatibility)
- The headless service gets the base annotations plus any additional headless-specific annotations

This approach would maintain backward compatibility while adding the new functionality, avoiding a breaking change for existing users.

### Example:

**Values file:**
```yaml
distributor:
  service:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-type: nlb
      service.beta.kubernetes.io/aws-load-balancer-internal: "true"
    headlessAnnotations:
      service.beta.kubernetes.io/aws-load-balancer-healthcheck-port: "8090"
      external-dns.alpha.kubernetes.io/hostname: "headless-only.example.com"
```

**Rendered regular service:**
```yaml
apiVersion: v1
kind: Service
metadata:
  name: mimir-distributor
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
spec:
  type: ClusterIP
  # ... rest of service spec
```

**Rendered headless service:**
```yaml
apiVersion: v1
kind: Service
metadata:
  name: mimir-distributor-headless
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
    service.beta.kubernetes.io/aws-load-balancer-healthcheck-port: "8090"
    external-dns.alpha.kubernetes.io/hostname: "headless-only.example.com"
spec:
  type: ClusterIP
  clusterIP: None
  # ... rest of service spec
```

As you can see, the headless service receives both the base annotations and the headless-specific annotations, while the regular service remains unchanged.


Will this behaviour work for you?

KyriosGN0:
@dimitarvdimitrov the behavior i want is a bit different, i want to able to add annotations only to the regular while keeping the headless one without any annotations.

dimitarvdimitrov:
what if you set the annotation to `null` in the `headlessServiceAnnotations`? Would that unset them for the headless service?


@narqo already asked this: what's the reason you can't use the gateway nginx service? The ingress resource and the service there are meant for accepting requests at the "edge."

I also want to explain the reason for pushback - other services do not have this exclusive behaviour. In addition to that, if we have headless nad regular service have distinct annotations config, then some existing installations need to go through a migration (change their values file while upgrading), which I want to avoid.

KyriosGN0:
@dimitarvdimitrov it might work, but not sure about it, will try to implement your suggestion and test it out

the reason is that we don't like nginx, we use loki and tempo and in both cases we don't use the nginx (as much as we can) 
it works for us in tempo since tempo already support this across services which have headless services (i guess from much earlier in the chart's history) 
i understand the pushback :)

KyriosGN0:
@dimitarvdimitrov i checked (i merged the to yaml dicts and then passed it to toYaml), it doesn't work since it needs to be at least 
worth noting that users can still have the old behavior with my changes in this PR, (they can still set the same annotations in both dicts) but that's still a breaking change of course

dimitarvdimitrov:
this still looks nice a niche use case. I think a breaking change will inconvenience more people than it will help. 

Other ideas I can think of to solve your problem are to 1) post-process the YAML manifests and remove the annotations yourself. 2) Adjust the annotations to something invalid on the headless service, so they effectively don't do anything. 3) create another service resource with the correct annotations in the `extraObjects` key in the values file. The `app.kubernetes.io/component: distributor` label on pods doesn't change between releases

KyriosGN0:
@dimitarvdimitrov thanks for the consideration, we will proceed with option 3

PR #10486, opened 2025-01-20
MQE time related functions

lamida:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/functions.go
```diff
@@ -61,6 +61,37 @@ func SingleInputVectorFunctionOperatorFactory(name string, f functions.FunctionO
...
 
+func TimeFunctionOperatorFactory(name string, f functions.FunctionOverInstantVectorDefinition) InstantVectorFunctionOperatorFactory {
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+		if len(args) > 1 {
+			// Should be caught by the PromQL parser, but we check here for safety.
+			return nil, fmt.Errorf("expected 1 or 0 argument for %s, got %v", name, len(args))
+		}
+
+		var inner types.InstantVectorOperator
+		var ok bool
```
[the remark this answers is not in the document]
lamida: Updated in c18826a1f4dc5a8e5efc65f181a8979073129056

pkg/streamingpromql/operators/functions/times.go
```diff
@@ -0,0 +1,69 @@
...
+
+func timeWrapperFunc(f func(t time.Time) float64) InstantVectorSeriesFunction {
+	return func(seriesData types.InstantVectorSeriesData, _ []types.ScalarData, tr types.QueryTimeRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker) (types.InstantVectorSeriesData, error) {
+		if len(seriesData.Floats)+len(seriesData.Histograms) == 0 {
+			fp := promql.FPoint{
+				F: f(time.Unix(tr.StartT, 0).UTC()),
+			}
+			seriesData.Floats = append(seriesData.Floats, fp)
+			return seriesData, nil
+		}
```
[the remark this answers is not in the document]
lamida: Yes apparently that is not needed. Removed in 822b75310c28939304f787df3944f5e32974cdeb.

pkg/streamingpromql/operators/functions/times.go
```diff
@@ -0,0 +1,69 @@
...
+		if len(seriesData.Floats) > 0 {
+			for i := range seriesData.Floats {
+				t := time.Unix(int64(seriesData.Floats[i].F), 0).UTC()
+				seriesData.Floats[i].F = f(t)
+			}
+			return seriesData, nil
+		}
+
+		// we don't do time based function on histograms
+		types.HPointSlicePool.Put(seriesData.Histograms, memoryConsumptionTracker)
```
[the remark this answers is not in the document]
lamida: Moved in 05bd2868207bcbaa88780c1a17f703942aae5464

pkg/streamingpromql/testdata/upstream/functions.test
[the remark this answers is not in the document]
lamida: Added some tests in `ours/functions.test`.

pkg/streamingpromql/operators/functions/times.go
```diff
@@ -62,7 +54,6 @@ func timeWrapperFunc(f func(t time.Time) float64) InstantVectorSeriesFunction {
 		}
 
 		// we don't do time based function on histograms
-		types.HPointSlicePool.Put(seriesData.Histograms, memoryConsumptionTracker)
 		seriesData.Histograms = nil
```
charleskorn, line 57: This needs to be done before we return `seriesData` above as well.

In fact, we could now drop the `if len(seriesData.Floats) > 0` check entirely, which would remove the early return as well.

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -682,3 +682,143 @@ eval range from 0 to 7m step 1m last_over_time(some_metric_count[3m])
...
+
+eval range from 0 to 7m step 1m day_of_week(histogram_sample)
+
+eval range from 0 to 7m step 1m day_of_year(histogram_sample)
+
+eval range from 0 to 7m step 1m days_in_month(histogram_sample)
+
+eval range from 0 to 7m step 1m hour(histogram_sample)
+
+eval range from 0 to 7m step 1m  minute(histogram_sample)
```
charleskorn, line 786: [nit]

```suggestion
eval range from 0 to 7m step 1m minute(histogram_sample)
```

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -682,3 +682,143 @@ eval range from 0 to 7m step 1m last_over_time(some_metric_count[3m])
...
+
+eval range from 0 to 7m step 1m  month(histogram_sample)
```
charleskorn, line 788: [nit]

```suggestion
eval range from 0 to 7m step 1m month(histogram_sample)
```

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -682,3 +682,143 @@ eval range from 0 to 7m step 1m last_over_time(some_metric_count[3m])
...
+# 1485907200: February 1st 2017 not in leap year.
+# 1672493820: 2022-12-31 13:37:00 365th day in non-leap year.
+load 1m
+  some_metric{case="series1"} 0+1136239445x7
+  another_metric{case="series1"} 0+536254375x7
+  some_metric{case="series2"} 0+1672493820x7
+  some_metric{case="gap1"} _ 1136239445 _ 1485907200 _ 1672493820
+  some_metric{case="gap2"} 1136239445 _ _ 1485907200 _ _ 1672493820
+  some_metric{case="histogram-gap"} _ {{count:5}} _ {{count:5}}
+  some_metric{case="histogram-float"} 1136239445x7 {{count:5}} 1485907200 {{count:5}}
```
charleskorn, line 805: Did you intend to have seven floats before the histogram values?

```suggestion
  some_metric{case="histogram-float"} 1136239445 {{count:5}} 1485907200 {{count:5}}
```

pkg/streamingpromql/functions.go
```diff
@@ -61,6 +61,37 @@ func SingleInputVectorFunctionOperatorFactory(name string, f functions.FunctionO
 	}
 }
 
+func TimeFunctionOperatorFactory(name string, f functions.FunctionOverInstantVectorDefinition) InstantVectorFunctionOperatorFactory {
```
charleskorn, line 64: Given this is only used by `TimeTransformationFunctionOperatorFactory`, I would move this inside that function - I don't think the separation is useful, and they're always used together.

lamida: Should be inlined in 16843e7ab802f2662b2f5bde70dabc8bb4cdb2ce

pkg/streamingpromql/operators/functions/times.go
lamida: I moved this up now, but not inside defer block. Because putting them in defer block causing the pedantic memory pool test to fail complaining negative allocation.

pkg/streamingpromql/operators/time.go
```diff
@@ -0,0 +1,58 @@
...
+	samples, err := types.FPointSlicePool.Get(s.TimeRange.StepCount, s.MemoryConsumptionTracker)
+
+	if err != nil {
+		return types.ScalarData{}, err
+	}
+
+	samples = samples[:s.TimeRange.StepCount]
+
+	for step := 0; step < s.TimeRange.StepCount; step++ {
+		t := s.TimeRange.StartT + int64(step)*s.TimeRange.IntervalMilliseconds
```
charleskorn, line 44: [nit] This could be simplified to:

```suggestion
		t := s.TimeRange.IndexTime(step)
```

pkg/streamingpromql/operators/time.go
```diff
@@ -0,0 +1,58 @@
...
+func (s *Time) GetValues(_ context.Context) (types.ScalarData, error) {
+	samples, err := types.FPointSlicePool.Get(s.TimeRange.StepCount, s.MemoryConsumptionTracker)
+
+	if err != nil {
+		return types.ScalarData{}, err
+	}
+
+	samples = samples[:s.TimeRange.StepCount]
+
+	for step := 0; step < s.TimeRange.StepCount; step++ {
```
charleskorn, line 43: [nit] This could be simplified to:
```suggestion
	for step := range s.TimeRange.StepCount {
```

pkg/streamingpromql/functions.go
```diff
@@ -76,6 +76,42 @@ func InstantVectorTransformationFunctionOperatorFactory(name string, seriesDataF
...
+			inner, ok = args[0].(types.InstantVectorOperator)
+			if !ok {
+				// Should be caught by the PromQL parser, but we check here for safety.
+				return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", name, args[0])
+			}
+
+		} else {
+			// if the argument is not provided, it will default to vector(time())
+			inner = scalars.NewScalarToInstantVector(operators.NewTime(timeRange, memoryConsumptionTracker, expressionPosition), expressionPosition)
+		}
```
charleskorn, line 104: [nit] This could be simplified to a structure like:

```
if len(args) == 0 {
	...
} else if len(args) == 1 {
	...
} else {
	return nil, fmt.Errorf("expected 0 or 1 argument for %s, got %v", name, len(args))
}
```

charleskorn (approved):
LGTM modulo nits

charleskorn (commented):
Also just noticed there are a bunch of test cases in `pkg/streamingpromql/testdata/upstream/at_modifier.test` that can now be enabled (eg. those starting around line 200) - could you please enable these?

(`tools/check-for-disabled-but-supported-mqe-test-cases` should find them all for you)

pkg/streamingpromql/functions.go
lamida: Updated in 1b91d5671bd1e9324188e6baaac968ceecb06590

pkg/streamingpromql/testdata/upstream/name_label_dropping.test
charleskorn, line 1: These test cases aren't actually supported by MQE - we haven't yet implemented the experimental delayed name dropping feature, and don't intend to any time soon.

lamida: Reverted in 2a4762a5f9d2d15dc926609845fa66639d51256d . I added this because `tools/check-for-disabled-but-supported-mqe-test-cases` said it is supported, although it was saying it for non-times function.

PR #10505, opened 2025-01-23
influx: fix endpoint path to match existing proxy

alexgreenbank:
#### What this PR does

Following on from https://github.com/grafana/mimir/pull/10153/ this PR changes the endpoint to match that of the [influx2cortex code](https://github.com/grafana/influx2cortex/blob/main/pkg/influx/api.go#L30) used as the basis for the influx write proxy.

This moves it from `/api/v1/influx/push` to `/api/v1/push/influx/write` (note change of order of `influx` and `push`, as well as the extra `/write` suffix).

Experimental feature, no further CHANGELOG or documentation updates required.


aknuds1 (approved):
LGTM

PR #10506, opened 2025-01-23
block-builder: Add an option to consume a partition completely till cycle end offset with no partially region

codesome:
#### What this PR does

Currently we have "last seen" and "commit" points, and the records between can be partially consumed. This PR adds an experimental `block-builder.no-partially-consumed-region` where we do not have this partially consumed region, instead entire records until cycle end time goes inside a block.

When flag is enabled, we also reduce the buffer time from 15 mins to 5 mins. Which means, the consumption will technically be from `X:05` to `(X+1):05`. The extra 5 mins is to allow for any remote write delay. Hard coded for now for experiments.

This flag is an intermediate state, meaning after running tests with this option, we will only keep one of them and get rid of the flag. The wish is to keep this new version since it makes scheduler design and parallelisation easier. If this one wins, we can also get rid of "last seen offset" from the commit information.

#### Checklist

- [X] Tests updated.
- ~[ ] Documentation added.~
- ~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.~


pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -889,14 +839,9 @@ func TestBlockBuilder_RetryOnTransientErrors(t *testing.T) {
...
 	require.Eventually(t, func() bool { return kafkaCommits.Load() >= 1 }, 50*time.Second, 100*time.Millisecond, "expected kafka commits")
 
-	bucketDir := path.Join(cfg.BlocksStorage.Bucket.Filesystem.Directory, "1")
-	db, err := tsdb.Open(bucketDir, promslog.NewNopLogger(), nil, nil, nil)
-	require.NoError(t, err)
-	t.Cleanup(func() { require.NoError(t, db.Close()) })
-	compareQuery(t,
-		db,
-		producedSamples,
-		nil,
```
seizethedave, line 899: Cool, you had mentioned this, but I didn't quite see the blatant duplication. 👍

pkg/blockbuilder/tsdb.go
```diff
@@ -133,12 +133,12 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record, lastBlockMax
...
 
 		for _, s := range ts.Samples {
-			if s.TimestampMs >= blockMax {
+			if !processEverything && s.TimestampMs >= blockMax {
 				// We will process this sample in the next cycle.
 				allSamplesProcessed = false
 				continue
 			}
-			if recordAlreadyProcessed && s.TimestampMs < lastBlockMax {
+			if !processEverything && recordAlreadyProcessed && s.TimestampMs < lastBlockMax {
```
seizethedave, line 141: As I understand it the first time we restart blockbuilders with this flag enabled, we will produce a small amount of duplicate records.  (records that had `recordAlreadyProcessed` from the previous cycle, but we're now ignoring those.) I do not see an avenue where we'd _skip_ any records on the first run. Am I right?

seizethedave (approved):
LGTM%comment!

pkg/blockbuilder/tsdb.go
codesome: Yes, migrating from old the new way will create some duplicate samples in blocks which is harmless since compactor will merge them together.

PR #10509, opened 2025-01-24
Mimir: Add attributed prefix to labels of cost attribution metrics

ying-jeanne:

Users can set cost attribution labels for cluster and namespace, which may conflict with existing metrics that include a cluster label. To prevent this, cost attribution labels are prefixed, and the prefix can be removed later in the pipeline using recording rules.

Similar approach is implemented in Loki. 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Grafana Mimir
 
 * [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269
+* [CHANGE] Ingester/Distributor: Prefix cost attribution metric labels with `attributed_` to prevent collisions with system cluster and namespace labels. #10509
```
colega, line 8: This is not needed because this new feature isn't included in any release yet.

pkg/costattribution/manager.go
```diff
@@ -21,6 +21,7 @@ const (
 	defaultTrackerName = "cost-attribution"
 	missingValue       = "__missing__"
 	overflowValue      = "__overflow__"
+	usagePrefix        = "attributed"
```
colega, line 24: Nit, but I would expect the prefix to be `attributed_`.

colega (approved):

CHANGELOG.md
ying-jeanne: addressed here 97a5d77efb0c7b63dd34d6fa8b5d2216d7334373

pkg/costattribution/manager.go
ying-jeanne: addressed in 97a5d77efb0c7b63dd34d6fa8b5d2216d7334373

pkg/costattribution/active_tracker.go
```diff
@@ -34,6 +34,14 @@ type ActiveSeriesTracker struct {
 	overflowCounter atomic.Int64
 }
 
+func addLabelsPrefix(labels []string) []string {
+	out := make([]string, 0, len(labels))
+	for _, l := range labels {
+		out = append(out, strings.Join([]string{usagePrefix, l}, ""))
```
colega, line 40: This should be just:
```suggestion
		out = usagePrefix + l
```

pkg/costattribution/manager.go
```diff
@@ -21,6 +21,7 @@ const (
 	defaultTrackerName = "cost-attribution"
 	missingValue       = "__missing__"
 	overflowValue      = "__overflow__"
+	usagePrefix        = "attributed_"
```
colega, line 24: I'm not sure why is this called "usagePrefix", but I know naming is hard.
Maybe it could be `exportedAttributedLabelPrefix`?

PR #10513, opened 2025-01-24
Shuffle User Order in UsersScanner

dmwilson-grafana:
If there is a rollout, or any kind of graceful shutdown of compactors, they terminate ongoing block cleanup and bucket index update jobs. This can lead to the bucket index not being updated for an extended period if jobs are terminated back-to-back. 

#### What this PR does

This PR addresses the above by shuffling the order of users submitted to `BlocksCleaner` by `UsersScanner`. When users are always processed in lexicographic order, successive restarts can cause read path failures for tenants with stale indexes. By randomizing this order, we reduce the probability that these indexes become stale. 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
-  [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.


pkg/storage/tsdb/users_scanner_test.go
```diff
@@ -50,11 +51,14 @@ func TestUsersScanner_ScanUsers_ShouldReturnUsersForWhichOwnerCheckOrTenantDelet
 	s := NewUsersScanner(bucketClient, isOwned, log.NewNopLogger())
 	actual, deleted, err := s.ScanUsers(context.Background())
 	require.NoError(t, err)
-	assert.Equal(t, expected, actual)
```
56quarters, line 53: You can use `assert.ElementsMatch` in this case.

pkg/storage/tsdb/users_scanner_test.go
```diff
@@ -63,5 +67,32 @@ func TestUsersScanner_ScanUsers_ShouldNotReturnPrefixedUsedByMimirInternals(t *t
 	s := NewUsersScanner(bucketClient, AllUsers, log.NewNopLogger())
 	actual, _, err := s.ScanUsers(context.Background())
 	require.NoError(t, err)
-	assert.Equal(t, []string{"user-1", "user-2"}, actual)
+	for _, userID := range users {
+		assert.Contains(t, actual, userID)
+	}
+}
+
+func TestUsersScanner_ScanUsers_ShouldReturnRandomizedOrder(t *testing.T) {
```
56quarters, line 75: I'm not sure what this test is testing. It doesn't seem like it's ensuring a random order.

56quarters (commented):
LGTM with some small suggestions. I'll let @seizethedave approve

pkg/storage/tsdb/users_scanner_test.go
dmwilson-grafana: Without the shuffle, the order of the input (`users`) would be the same as the order of `actual`. On L94, I check that we have the same set of users (as you mentioned, this can be changed to `assert.ElementsMatch`) and on L97 I check that it's actually been shuffled , i.e. `users` != `actual`.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_misses_total`
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_requests_total`
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_skips_total`
+* [ENHANCEMENT] Compactor: Shuffle order of users in `UsersScanner` before submitting to `BlocksCleaner`. #10513
```
seizethedave, line 38: Can this briefly say "why" in addition to "how"? So that unfamiliar readers understand why they might upgrade, and so on. :)

pkg/storage/tsdb/users_scanner.go
```diff
@@ -46,6 +47,10 @@ func (s *UsersScanner) ScanUsers(ctx context.Context) (users, markedForDeletion
 		return nil, nil, err
 	}
 
+	rand.Shuffle(len(users), func(i, j int) {
```
seizethedave, line 50: I know we've changed course on this a time or two, but `ScanUsers` is a general purpose accessor function used by many services, whereas this shuffle behavior is particular to the needs of compactor. I actually think we can move the shuffle to `runCleanup` after we call `refreshOwnedUsers`, with a brief handrail comment noting why we'd do such a thing.

seizethedave (commented):
Thanks, this is looking good. Let me know what you think about these comments.

seizethedave (approved):
LGTM, thanks!

PR #10523, opened 2025-01-27
MQE Absent function

lamida:
#### What this PR does

Implement absent function. Added a new Operator: `Absent`. We also need to modify `InstantVectorFunctionOperatorFactory` to pass `parser.Expressions` object needed to evaluate the labels of the absent argument.

#### Benchmark

Step to create the benchmark:
- Add test in `pkg/streamingpromql/benchmarks/benchmarks.go`
- `cd tools/benchmark-query-engine`
- `go run . -bench=Query/absent -count=6 | tee absent-benchmark.txt `
- ` ./compare.sh absent-benchmark.txt`
- `rm absent-benchmark.txt`

```
goos: darwin
goarch: arm64
cpu: Apple M1 Max
                                                           │  Prometheus   │                Mimir                │
                                                           │    sec/op     │    sec/op     vs base               │
Query/absent(a_1),_instant_query-10                           144.0µ ±  4%   144.2µ ±  8%        ~ (p=0.699 n=6)
Query/absent(a_1),_range_query_with_100_steps-10              154.3µ ±  5%   147.4µ ±  2%   -4.44% (p=0.004 n=6)
Query/absent(a_1),_range_query_with_1000_steps-10             250.2µ ± 12%   199.0µ ±  3%  -20.47% (p=0.002 n=6)
Query/absent(a_100),_instant_query-10                         752.1µ ±  5%   731.9µ ±  3%   -2.69% (p=0.015 n=6)
Query/absent(a_100),_range_query_with_100_steps-10            1.347m ±  3%   1.250m ±  1%   -7.21% (p=0.002 n=6)
Query/absent(a_100),_range_query_with_1000_steps-10           6.264m ±  1%   5.454m ±  1%  -12.93% (p=0.002 n=6)
Query/absent(a_2000),_instant_query-10                       10.416m ±  2%   9.895m ±  9%        ~ (p=0.065 n=6)
Query/absent(a_2000),_range_query_with_100_steps-10           21.42m ± 13%   19.24m ± 11%  -10.16% (p=0.015 n=6)
Query/absent(a_2000),_range_query_with_1000_steps-10         146.07m ±  1%   97.98m ±  3%  -32.92% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_instant_query-10                     127.0µ ± 15%   117.8µ ±  9%   -7.25% (p=0.026 n=6)
Query/absent(a_1_>_Inf),_range_query_with_100_steps-10        176.2µ ±  3%   128.1µ ±  6%  -27.30% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_1000_steps-10       578.7µ ±  4%   190.6µ ±  4%  -67.07% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_instant_query-10                   768.5µ ±  7%   734.2µ ±  6%   -4.46% (p=0.041 n=6)
Query/absent(a_100_>_Inf),_range_query_with_100_steps-10      1.500m ±  3%   1.305m ±  2%  -13.03% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_1000_steps-10     7.671m ±  4%   5.881m ±  1%  -23.34% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_instant_query-10                  10.49m ± 12%   10.05m ±  2%   -4.19% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_100_steps-10     23.52m ±  9%   20.16m ±  2%  -14.28% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_1000_steps-10    171.2m ±  5%   106.4m ±  5%  -37.86% (p=0.002 n=6)
geomean                                                       2.370m         1.929m        -18.58%

                                                           │  Prometheus   │                Mimir                │
                                                           │     B/op      │     B/op      vs base               │
Query/absent(a_1),_instant_query-10                           22.23Ki ± 0%   20.85Ki ± 0%   -6.22% (p=0.002 n=6)
Query/absent(a_1),_range_query_with_100_steps-10              25.65Ki ± 0%   21.65Ki ± 0%  -15.58% (p=0.002 n=6)
Query/absent(a_1),_range_query_with_1000_steps-10             51.70Ki ± 0%   26.31Ki ± 0%  -49.12% (p=0.002 n=6)
Query/absent(a_100),_instant_query-10                         194.9Ki ± 0%   150.5Ki ± 0%  -22.78% (p=0.002 n=6)
Query/absent(a_100),_range_query_with_100_steps-10            253.7Ki ± 0%   206.9Ki ± 0%  -18.44% (p=0.002 n=6)
Query/absent(a_100),_range_query_with_1000_steps-10           697.5Ki ± 0%   632.3Ki ± 0%   -9.35% (p=0.002 n=6)
Query/absent(a_2000),_instant_query-10                        3.536Mi ± 0%   2.577Mi ± 0%  -27.11% (p=0.002 n=6)
Query/absent(a_2000),_range_query_with_100_steps-10           4.660Mi ± 0%   3.679Mi ± 0%  -21.04% (p=0.002 n=6)
Query/absent(a_2000),_range_query_with_1000_steps-10          12.80Mi ± 1%   11.85Mi ± 1%   -7.49% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_instant_query-10                     25.20Ki ± 0%   21.42Ki ± 0%  -15.02% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_100_steps-10        56.87Ki ± 0%   22.16Ki ± 0%  -61.03% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_1000_steps-10      338.08Ki ± 0%   26.59Ki ± 0%  -92.14% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_instant_query-10                   198.0Ki ± 0%   153.3Ki ± 0%  -22.54% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_100_steps-10      285.2Ki ± 0%   209.7Ki ± 0%  -26.49% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_1000_steps-10     984.6Ki ± 0%   631.3Ki ± 0%  -35.88% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_instant_query-10                  3.542Mi ± 0%   2.655Mi ± 0%  -25.03% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_100_steps-10     4.694Mi ± 0%   3.758Mi ± 0%  -19.95% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_1000_steps-10    13.05Mi ± 1%   11.89Mi ± 1%   -8.90% (p=0.002 n=6)
geomean                                                       473.8Ki        314.8Ki       -33.56%

                                                           │  Prometheus  │               Mimir                │
                                                           │  allocs/op   │  allocs/op   vs base               │
Query/absent(a_1),_instant_query-10                            401.0 ± 0%    347.0 ± 0%  -13.47% (p=0.002 n=6)
Query/absent(a_1),_range_query_with_100_steps-10               511.0 ± 0%    354.0 ± 0%  -30.72% (p=0.002 n=6)
Query/absent(a_1),_range_query_with_1000_steps-10             1437.0 ± 0%    380.0 ± 0%  -73.56% (p=0.002 n=6)
Query/absent(a_100),_instant_query-10                         2.413k ± 0%   2.252k ± 0%   -6.67% (p=0.002 n=6)
Query/absent(a_100),_range_query_with_100_steps-10            2.830k ± 0%   2.661k ± 0%   -5.97% (p=0.002 n=6)
Query/absent(a_100),_range_query_with_1000_steps-10           6.344k ± 0%   5.275k ± 0%  -16.85% (p=0.002 n=6)
Query/absent(a_2000),_instant_query-10                        40.87k ± 0%   38.80k ± 0%   -5.08% (p=0.002 n=6)
Query/absent(a_2000),_range_query_with_100_steps-10           47.06k ± 0%   46.87k ± 0%   -0.42% (p=0.002 n=6)
Query/absent(a_2000),_range_query_with_1000_steps-10         100.04k ± 0%   98.94k ± 0%   -1.10% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_instant_query-10                      462.0 ± 0%    361.0 ± 0%  -21.86% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_100_steps-10         970.5 ± 0%    368.0 ± 0%  -62.08% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_1000_steps-10       5499.0 ± 0%    394.0 ± 0%  -92.84% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_instant_query-10                   2.474k ± 0%   2.269k ± 0%   -8.29% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_100_steps-10      3.289k ± 0%   2.679k ± 0%  -18.55% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_1000_steps-10    10.406k ± 0%   5.293k ± 0%  -49.14% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_instant_query-10                  40.93k ± 0%   38.82k ± 0%   -5.15% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_100_steps-10     47.50k ± 0%   46.88k ± 0%   -1.31% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_1000_steps-10   104.09k ± 0%   98.99k ± 0%   -4.90% (p=0.002 n=6)
geomean                                                       6.007k        4.035k       -32.82%

                                                           │  Prometheus   │                Mimir                │
                                                           │       B       │      B        vs base               │
Query/absent(a_1),_instant_query-10                           67.13Mi ± 1%   66.93Mi ± 1%        ~ (p=0.240 n=6)
Query/absent(a_1),_range_query_with_100_steps-10              67.01Mi ± 2%   66.65Mi ± 1%        ~ (p=0.686 n=6)
Query/absent(a_1),_range_query_with_1000_steps-10             64.52Mi ± 2%   65.21Mi ± 1%   +1.07% (p=0.041 n=6)
Query/absent(a_100),_instant_query-10                         62.32Mi ± 1%   61.69Mi ± 1%        ~ (p=0.093 n=6)
Query/absent(a_100),_range_query_with_100_steps-10            62.62Mi ± 1%   62.02Mi ± 2%        ~ (p=0.132 n=6)
Query/absent(a_100),_range_query_with_1000_steps-10           64.91Mi ± 1%   61.78Mi ± 1%   -4.83% (p=0.002 n=6)
Query/absent(a_2000),_instant_query-10                        63.10Mi ± 1%   63.23Mi ± 1%        ~ (p=0.786 n=6)
Query/absent(a_2000),_range_query_with_100_steps-10           71.98Mi ± 2%   62.62Mi ± 1%  -13.00% (p=0.002 n=6)
Query/absent(a_2000),_range_query_with_1000_steps-10         127.16Mi ± 3%   68.32Mi ± 2%  -46.27% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_instant_query-10                     67.08Mi ± 2%   67.46Mi ± 1%        ~ (p=0.193 n=6)
Query/absent(a_1_>_Inf),_range_query_with_100_steps-10        65.09Mi ± 1%   67.28Mi ± 2%   +3.36% (p=0.002 n=6)
Query/absent(a_1_>_Inf),_range_query_with_1000_steps-10       63.39Mi ± 1%   65.66Mi ± 1%   +3.57% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_instant_query-10                   62.13Mi ± 1%   62.17Mi ± 1%        ~ (p=1.000 n=6)
Query/absent(a_100_>_Inf),_range_query_with_100_steps-10      62.62Mi ± 1%   61.56Mi ± 1%   -1.68% (p=0.002 n=6)
Query/absent(a_100_>_Inf),_range_query_with_1000_steps-10     65.15Mi ± 0%   61.96Mi ± 0%   -4.89% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_instant_query-10                  63.38Mi ± 3%   63.34Mi ± 2%        ~ (p=0.732 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_100_steps-10     71.33Mi ± 1%   62.75Mi ± 1%  -12.03% (p=0.002 n=6)
Query/absent(a_2000_>_Inf),_range_query_with_1000_steps-10   129.14Mi ± 3%   67.95Mi ± 2%  -47.39% (p=0.002 n=6)
geomean                                                       70.26Mi        64.32Mi        -8.45%

```

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -79,9 +79,9 @@ func InstantVectorTransformationFunctionOperatorFactory(name string, seriesDataF
 	return SingleInputVectorFunctionOperatorFactory(name, f)
 }
 
-func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, innerExpressions parser.Expressions) (types.InstantVectorOperator, error) {
+func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, argExpressions parser.Expressions) (types.InstantVectorOperator, error) {
 	functionName := "absent"
-	if len(args) != 1 && len(innerExpressions) != 1 {
+	if len(args) != 1 && len(argExpressions) != 1 {
 		// Should be caught by the PromQL parser, but we check here for safety.
 		return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", functionName, len(args))
 	}
@@ -91,7 +91,7 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 		return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 	}
 
-	var o types.InstantVectorOperator = functions.NewAbsent(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
+	var o types.InstantVectorOperator = functions.NewAbsent(inner, argExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 	return o, nil
 }
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -81,7 +81,7 @@ func InstantVectorTransformationFunctionOperatorFactory(name string, seriesDataF
 
 func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, innerExpressions parser.Expressions) (types.InstantVectorOperator, error) {
 	functionName := "absent"
-	if len(args) != 1 {
+	if len(args) != 1 && len(innerExpressions) != 1 {
 		// Should be caught by the PromQL parser, but we check here for safety.
 		return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", functionName, len(args))
 	}
@@ -91,7 +91,6 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 		return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 	}
 
-	// TODO check innerExpressions length too
 	var o types.InstantVectorOperator = operators.NewAbsent(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 	return o, nil
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -91,7 +91,7 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 		return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 	}
 
-	var o types.InstantVectorOperator = functions.NewAbsentOperator(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
+	var o types.InstantVectorOperator = functions.NewAbsent(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 	return o, nil
 }
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -91,7 +91,7 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 		return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 	}
 
-	var o types.InstantVectorOperator = operators.NewAbsentOperator(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
+	var o types.InstantVectorOperator = functions.NewAbsentOperator(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 	return o, nil
 }
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -91,7 +91,7 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 		return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 	}
 
-	var o types.InstantVectorOperator = operators.NewAbsent(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
+	var o types.InstantVectorOperator = operators.NewAbsentOperator(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 	return o, nil
 }
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -24,7 +24,7 @@ type InstantVectorFunctionOperatorFactory func(
 	annotations *annotations.Annotations,
 	expressionPosition posrange.PositionRange,
 	timeRange types.QueryTimeRange,
-	innerExpressions parser.Expressions,
+	argExpressions parser.Expressions,
 ) (types.InstantVectorOperator, error)
 
 type ScalarFunctionOperatorFactory func(
--- a/pkg/streamingpromql/functions.go
+++ b/pkg/streamingpromql/functions.go
@@ -24,6 +24,7 @@ type InstantVectorFunctionOperatorFactory func(
 	annotations *annotations.Annotations,
 	expressionPosition posrange.PositionRange,
 	timeRange types.QueryTimeRange,
+	innerExpressions parser.Expressions,
 ) (types.InstantVectorOperator, error)
 
 type ScalarFunctionOperatorFactory func(
@@ -41,7 +42,7 @@ type ScalarFunctionOperatorFactory func(
 //   - name: The name of the function
 //   - f: The function implementation
 func SingleInputVectorFunctionOperatorFactory(name string, f functions.FunctionOverInstantVectorDefinition) InstantVectorFunctionOperatorFactory {
-	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 		if len(args) != 1 {
 			// Should be caught by the PromQL parser, but we check here for safety.
 			return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", name, len(args))
@@ -78,19 +79,21 @@ func InstantVectorTransformationFunctionOperatorFactory(name string, seriesDataF
 	return SingleInputVectorFunctionOperatorFactory(name, f)
 }
 
-func AbsentFunctionOperatorFactory(name string, innerExpr parser.Expr) InstantVectorFunctionOperatorFactory {
-	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+func AbsentFunctionOperatorFactory() InstantVectorFunctionOperatorFactory {
+	functionName := "absent"
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, innerExpressions parser.Expressions) (types.InstantVectorOperator, error) {
 		if len(args) != 1 {
 			// Should be caught by the PromQL parser, but we check here for safety.
-			return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", name, len(args))
+			return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", functionName, len(args))
 		}
 		inner, ok := args[0].(types.InstantVectorOperator)
 		if !ok {
 			// Should be caught by the PromQL parser, but we check here for safety.
-			return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", name, args[0])
+			return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", functionName, args[0])
 		}
 
-		var o types.InstantVectorOperator = operators.NewAbsent(inner, innerExpr, timeRange, expressionPosition, memoryConsumptionTracker)
+		// TODO check innerExpressions length too
+		var o types.InstantVectorOperator = operators.NewAbsent(inner, innerExpressions[0], timeRange, expressionPosition, memoryConsumptionTracker)
 
 		return o, nil
 	}
@@ -102,7 +105,7 @@ func TimeTransformationFunctionOperatorFactory(name string, seriesDataFunc funct
 		SeriesMetadataFunction: functions.DropSeriesName,
 	}
 
-	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 		var inner types.InstantVectorOperator
 		if len(args) == 0 {
 			// if the argument is not provided, it will default to vector(time())
@@ -156,7 +159,7 @@ func FunctionOverRangeVectorOperatorFactory(
 	name string,
 	f functions.FunctionOverRangeVectorDefinition,
 ) InstantVectorFunctionOperatorFactory {
-	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+	return func(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 		if len(args) != 1 {
 			// Should be caught by the PromQL parser, but we check here for safety.
 			return nil, fmt.Errorf("expected exactly 1 argument for %s, got %v", name, len(args))
@@ -178,7 +181,7 @@ func FunctionOverRangeVectorOperatorFactory(
 	}
 }
 
-func PredictLinearFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+func PredictLinearFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 	f := functions.PredictLinear
 
 	if len(args) != 2 {
@@ -207,7 +210,7 @@ func PredictLinearFactory(args []types.Operator, memoryConsumptionTracker *limit
 	return o, nil
 }
 
-func QuantileOverTimeFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+func QuantileOverTimeFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 	f := functions.QuantileOverTime
 
 	if len(args) != 2 {
@@ -236,7 +239,7 @@ func QuantileOverTimeFactory(args []types.Operator, memoryConsumptionTracker *li
 	return o, nil
 }
 
-func scalarToInstantVectorOperatorFactory(args []types.Operator, _ *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, _ types.QueryTimeRange) (types.InstantVectorOperator, error) {
+func scalarToInstantVectorOperatorFactory(args []types.Operator, _ *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, _ types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 	if len(args) != 1 {
 		// Should be caught by the PromQL parser, but we check here for safety.
 		return nil, fmt.Errorf("expected exactly 1 argument for vector, got %v", len(args))
@@ -251,7 +254,7 @@ func scalarToInstantVectorOperatorFactory(args []types.Operator, _ *limiting.Mem
 	return scalars.NewScalarToInstantVector(inner, expressionPosition), nil
 }
 
-func LabelJoinFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+func LabelJoinFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, _ parser.Expressions) (types.InstantVectorOperator, error) {
 	// It is valid for label_join to have no source label names. ie, only 3 arguments are actually required.
 	if len(args) < 3 {
 		// Should be caught by the PromQL parser, but we check here for safety.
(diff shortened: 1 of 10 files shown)


pkg/streamingpromql/operators/absent.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+// AbsentOperator is an operator that implements the absent() function.
+type AbsentOperator struct {
+	timeRange                types.QueryTimeRange
+	innerExpr                parser.Expr
+	inner                    types.InstantVectorOperator
+	expressionPosition       posrange.PositionRange
+	absentCount              int
```
charleskorn, line 27: If I'm understanding your intention correctly, would `innerSeriesCount` be a clearer name?

pkg/streamingpromql/operators/absent.go
charleskorn, line 1: I don't think this is correct - I think we're missing a few test cases that might flush out the correct behaviour:
* instant query over a selector that selects multiple series, each with a point at the query time
* instant query over an expression that could produce multiple series, but doesn't have any points (eg. `absent(metric_with_many_series > Inf)`)
* range query over a selector that selects a single series that does not have points at every time step in the range
* same as above, but with points at every time step in the range
* range query over a selector that selects multiple series that together don't have points at every time step in the range
* same as above, but where each time step has a sample in at least one series
* range query over an expression that could produce multiple series, but doesn't have any points (eg. `absent(metric_with_many_series > Inf)`)

pkg/streamingpromql/operators/absent.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/promql/parser/posrange"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+// AbsentOperator is an operator that implements the absent() function.
+type AbsentOperator struct {
```
charleskorn, line 22: [nit] We generally haven't used an `Operator` suffix for operators - I think `Absent` would be sufficient.

pkg/streamingpromql/operators/absent.go
charleskorn, line 1: [nit] This probably belongs in the `functions` package given it implements a function.

pkg/streamingpromql/functions.go
```diff
@@ -22,6 +23,7 @@ type InstantVectorFunctionOperatorFactory func(
 	annotations *annotations.Annotations,
 	expressionPosition posrange.PositionRange,
 	timeRange types.QueryTimeRange,
+	innerExpressions parser.Expressions,
```
charleskorn, line 26: [nit] Perhaps `argumentExpressions` would be a clearer name? Or `argExpressions` to maintain symmetry with `args`?

pkg/streamingpromql/operators/absent.go
lamida: Thanks for the detailed feedback to ensure the correctness. I will add the test cases to ensure the correct behaviour.

pkg/streamingpromql/operators/absent.go
lamida: This is count how often the series is not there or being `absent`. But let me double check again according to the feedback in https://github.com/grafana/mimir/pull/10523#discussion_r1931695665 to ensure the correctness.

lamida:
@charleskorn I have addressed all of your comments from the previous review. Please help to do another review. 🙏

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -940,3 +940,83 @@ eval range from 0 to 4m step 1m sort_desc(test_metric)
...
+
+# absent() tests
+load 4m
+  series{case="-"} _x11
+  series{case="0a"} _x11
+  series{case="0b"} _x11
+  series{case="0"} 0x11
+  series{case="a"} _ 1 _ 1 _ 1 _ 1 _ 1 _ 
+  series{case="b"} _ _ 2 _ 2 _ 2 _ 2 _ 2 
+  series{case="c"} _ _ _ 3 _ _ _ 3 _ _ _
```
charleskorn, line 954: Could you please add a test for instant and range queries with `absent(series{case=~"(a|b)"})`? This will flush out the changes required in the operator to correctly handle multiple input series.

pkg/streamingpromql/operators/functions/absent.go
charleskorn, line 1: This won't handle the case where there are multiple inner series correctly.

I think we need to do something like this:

In `SeriesMetadata`:

* call `Inner.SeriesMetadata()` to get the list of inner series
* create a slice to keep track of whether we've seen a point at each time step
* for each inner series:
  * call `Inner.NextSeries()` to get the data for that series
  * update the presence slice based on the series' data
* if there is a point present at every time step, return no series
* otherwise, if there are any points absent:
  * store the presence slice for use in `NextSeries()` later
  * return a single series (as it does currently)

In `NextSeries`:

* if there is a point present at every time step, or if `NextSeries()` has been called before, return `EOS`
* otherwise: 
  * construct the slice of `FPoint`s based on the presence slice created in `SeriesMetadata()` 
  * return the `FPoint`s

pkg/streamingpromql/functions_test.go
```diff
@@ -80,6 +80,7 @@ func TestFunctionDeduplicateAndMerge(t *testing.T) {
 	expressions := map[string]string{
 		// Please keep this list sorted alphabetically.
 		"abs":                `abs({__name__=~"float.*"})`,
+		"absent":             `absent({__name__=~"float.*"})`,
```
charleskorn, line 83: It's probably better to skip this for `absent` - `absent` doesn't need deduplication and merging:

```suggestion
		"absent":             `<skip>`,
```

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -940,3 +940,83 @@ eval range from 0 to 4m step 1m sort_desc(test_metric)
...
+eval range from 0s to 40m step 4m absent(series{case=~"(0a|0b)"})
+	{} 1x10
+
+eval range from 0s to 40m step 4m absent(series > Inf)
+	{} 1x10
+
+eval range from 0s to 40m step 4m absent(non_existent_metric)
+	{} 1x10
+
+# test look back period
```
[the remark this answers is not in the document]
lamida: Removed in eb544c1484bc931d63b4ca05fad331d1e640f0af

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -940,3 +940,83 @@ eval range from 0 to 4m step 1m sort_desc(test_metric)
   test_metric{case="histogram with NaN"}  {{count:0 sum:NaN}} {{count:0 sum:NaN}} {{count:0 sum:NaN}} {{count:0 sum:NaN}} {{count:0 sum:NaN}}
   test_metric{case="histogram with +Inf"} {{count:0 sum:Inf}} {{count:0 sum:Inf}} {{count:0 sum:Inf}} {{count:0 sum:Inf}} {{count:0 sum:Inf}}
   test_metric{case="histogram with -Inf"} {{count:0 sum:-Inf}} {{count:0 sum:-Inf}} {{count:0 sum:-Inf}} {{count:0 sum:-Inf}} {{count:0 sum:-Inf}}
+
+clear
+
+# absent() tests
+load 4m
```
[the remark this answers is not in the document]
lamida: Removed in 600248eb24990d2390f9393ce00727f8e34d2b40

pkg/streamingpromql/testdata/ours/functions.test
lamida: Added in 3765fcf6d007f560ed79ce8c2d19f9d9522e74f7

pkg/streamingpromql/testdata/ours/functions.test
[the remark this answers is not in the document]
charleskorn: 600248e is only part of what I'd recommend: I'd also recommend changing the step on the range query test cases to 6m as well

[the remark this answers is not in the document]
charleskorn: (same applies to the test added in 3765fcf)

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -940,3 +940,86 @@ eval range from 0 to 4m step 1m sort_desc(test_metric)
...
+
+eval range from 0s to 40m step 4m absent(series{case="-"})
+	{case="-"} 1x10
+
+eval range from 0s to 40m step 4m absent(series{case="0"})
+  # expected no result
+
+eval range from 0s to 40m step 4m series{case=~"(a|b)"}
+  series{case="a"} _ _ 1 _ _ 1 _ _ 1 
+  series{case="b"} _ _ _ 2 2 _ 2 2 _ 2 2
```
charleskorn, line 988: This should be this, no?

```suggestion
  {} 1 1 _ _ _ _ _ _ _ _ _
```

lamida: This is just the series range evaluation `series{case=~"(a|b)"}`. The absent evaluation itself `absent(series{case=~"(a|b)"})` is in the next line. It is easier to see the expected output of absent after seeing the multiple series range evaluation above.

pkg/streamingpromql/testdata/ours/functions.test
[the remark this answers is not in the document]
lamida: I thought you want us to test unaligned range query with the sample. Nevertheless set the evaulation step to 6m too in ac1f1c96f26623280f7c7d392bfefdbe17865ca3

pkg/streamingpromql/testdata/ours/functions.test
lamida: To make this less confusing, let's just remove that range query evaluation c9de184e2b07e39a9014e3962b3789f122dfd048

jhesketh (commented):
Sorry if this has already been discussed/considered (if so, please disregard this comment).

However, I wonder if given the uniqueness of this operator if it makes sense to handle it as a special case instead of as an `InstantVectorFunctionOperatorFactory`?

Eg in `convertFunctionCallToInstantVectorOperator` in `query.go` we check if it's the `absent` function and then create the operator and pass it the full inner expression instead of looking it up in `instantVectorFunctionOperatorFactories`.

This way the `absent` operator can reason about the inner series. For example, if I'm understand the function correctly, there is nothing to do if the inner expression is something other than a selector (eg there's no point evaluating a `sum`).

Sorry if this has already been discussed or derails things too much!

charleskorn:
> This way the `absent` operator can reason about the inner series. For example, if I'm understand the function correctly, there is nothing to do if the inner expression is something other than a selector (eg there's no point evaluating a `sum`).

`absent` can be run over any expression that produces an instant vector. The only difference in behaviour between the case where `absent` is run over a instant vector selector and some other expression is the labels it will return:

* if the argument is an instant vector selector, `absent` should try to infer the output labels
* if the argument isn't an instant vector selector, `absent` will always produce a series with no labels (ie. `{}`)

jhesketh:
> > This way the `absent` operator can reason about the inner series. For example, if I'm understand the function correctly, there is nothing to do if the inner expression is something other than a selector (eg there's no point evaluating a `sum`).
> 
> `absent` can be run over any expression that produces an instant vector. The only difference in behaviour between the case where `absent` is run over a instant vector selector and some other expression is the labels it will return:
> 
>     * if the argument is an instant vector selector, `absent` will infer the output labels from the selector
> 
>     * if the argument isn't an instant vector selector, `absent` will always produce a series with no labels (ie. `{}`)

Right, I see. However it may still be neater to have the operator called specifically in query.go to avoid passing both an evaluated and non-evaluated expression to all the functions (which feels odd).

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (a *Absent) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
+	innerMetadata, err := a.inner.SeriesMetadata(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer types.PutSeriesMetadataSlice(innerMetadata)
+
+	if a.presence == nil {
```
jhesketh, line 51: I'm not sure this check is needed. Won't this always be nil?

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+		}
+
+	}
+
+	return metadata, nil
+}
+
+func (a *Absent) NextSeries(_ context.Context) (types.InstantVectorSeriesData, error) {
+	output := types.InstantVectorSeriesData{}
```
jhesketh, line 95: Do we want to have an EOS error if this has been called already?

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,154 @@
...
+
+	var err error
+	output.Floats, err = types.FPointSlicePool.Get(a.timeRange.StepCount, a.memoryConsumptionTracker)
```
charleskorn, line 98: [nit] We could defer allocating this until we know we're returning at least one point in the `if` below

charleskorn:
> However it may still be neater to have the operator called specifically in query.go to avoid passing both an evaluated and non-evaluated expression to all the functions (which feels odd).

Are you referring to passing the expression to the function factory? If so, I don't mind this so much, but I don't hold this opinion strongly. Making `absent` a special case also feels a bit odd to me.

jhesketh:
> > However it may still be neater to have the operator called specifically in query.go to avoid passing both an evaluated and non-evaluated expression to all the functions (which feels odd).
> 
> Are you referring to passing the expression to the function factory? If so, I don't mind this so much, but I don't hold this opinion strongly. Making `absent` a special case also feels a bit odd to me.

Yes, that's what I was referring to.
It feels odd to me to pass both the parsed arguments, and unparsed expressions. Especially since the expressions are only used by absent.
I don't hold this strongly either, but find of the two oddities, handling absent in a special case is my personal preference.

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,154 @@
...
+	if a.presence == nil {
+		a.presence, err = types.BoolSlicePool.Get(a.timeRange.StepCount, a.memoryConsumptionTracker)
+		if err != nil {
+			return nil, err
+		}
+
+		// Initialize presence slice
+		for range a.timeRange.StepCount {
+			a.presence = append(a.presence, false)
+		}
```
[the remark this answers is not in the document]
lamida: Addressed in 658b1c448963b4c71b782f4978a20af1051d3132

pkg/streamingpromql/operators/functions/absent.go
lamida: Added in 12ee13c41834e59c5b71ff6a1e7a0c36c2f154d9

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,154 @@
...
+	metadata := types.GetSeriesMetadataSlice(1)
+	metadata = append(metadata, types.SeriesMetadata{
+		Labels: createLabelsForAbsentFunction(a.argExpressions),
+	})
+
+	for range innerMetadata {
+		series, err := a.inner.NextSeries(ctx)
+		if err != nil && errors.Is(err, types.EOS) {
+			return metadata, err
+		}
```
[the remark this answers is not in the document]
lamida: Addressed in 4f59a5db646857eabc94eb701f9dd5067907576b.

lamida:
> It feels odd to me to pass both the parsed arguments, and unparsed expressions. Especially since the expressions are only used by absent.

@jhesketh I will consider to refactor this once I am making progress with `absent_over_time`. Probably with that `absent_over_time`, it will be more obvious, whether we should pull up passing the unparsed arguments, or instead we should just keep this way. But for now, I prefer just to keep this and have the PR merged once all other concerns are addressed.

But you have some valid points.

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -95,14 +82,18 @@ func (a *Absent) NextSeries(_ context.Context) (types.InstantVectorSeriesData, e
...
-	if err != nil {
-		return output, err
-	}
-
+	idx := 0
 	for step := range a.timeRange.StepCount {
 		t := a.timeRange.IndexTime(int64(step))
 		if !a.presence[step] {
+			// We only allocate the slice once we know we have point to add.
+			if idx == 0 {
```
charleskorn, line 90: This can be simplified further: we can check if `output.Floats` is `nil`, and if it is, then that's our signal to get the slice from the pool. Then we don't need `idx` at all.

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,145 @@
...
+	metadata = append(metadata, types.SeriesMetadata{
+		Labels: createLabelsForAbsentFunction(a.argExpressions),
+	})
+
+	for range innerMetadata {
+		series, err := a.inner.NextSeries(ctx)
+		if err != nil {
+			return metadata, err
+		}
+		defer types.PutInstantVectorSeriesData(series, a.memoryConsumptionTracker)
```
charleskorn, line 68: Using `defer` in a loop like this will cause an allocation for every iteration of the loop, and also won't call the deferred function until `SeriesMetadata()` returns, which means that the slice can't be reused by `a.inner`.

Instead, what if we move the `types.PutInstantVectorSeriesData(series, a.memoryConsumptionTracker)` to the last line of this loop body and drop the `defer`?

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (a *Absent) NextSeries(_ context.Context) (types.InstantVectorSeriesData, error) {
+	output := types.InstantVectorSeriesData{}
+
+	var err error
+	idx := 0
+	for step := range a.timeRange.StepCount {
+		t := a.timeRange.IndexTime(int64(step))
+		if !a.presence[step] {
```
charleskorn, line 88: [nit] This could be rewritten to something like this:

```go
if a.presence[step] { 
    continue
}

// ...append point to output.Floats
```

pkg/streamingpromql/operators/functions/absent.go
lamida: Fixed in aa5df1c42dc1daaf8ff1d2245149e64800c322b6

pkg/streamingpromql/operators/functions/absent.go
lamida: Fixed in 8cb866def5805a9949ed1a502f1f209dde7975d6

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,144 @@
...
+
+	metadata := types.GetSeriesMetadataSlice(1)
+	metadata = append(metadata, types.SeriesMetadata{
+		Labels: createLabelsForAbsentFunction(a.argExpressions),
+	})
+
+	for range innerMetadata {
+		series, err := a.inner.NextSeries(ctx)
+		if err != nil {
+			return metadata, err
```
charleskorn, line 66: For safety, if we're returning an error, we shouldn't return the metadata slice:

```suggestion
			return nil, err
```

pkg/streamingpromql/operators/functions/absent.go
charleskorn: Boolean flag sounds good to me. Like you say, this method _should_ only be called once, but it'd be a good thing to enforce.

pkg/streamingpromql/testdata/ours/functions.test
charleskorn, line 1: Could you please make the indentation in the tests below consistent? Some use two spaces, some use tabs.

pkg/streamingpromql/operators/functions/absent.go
lamida: Added here 8a6c28840ce5c91a1cef6f490914b1395639806b

pkg/streamingpromql/testdata/ours/functions.test
lamida: Fixed for consistency here: cfe75d219b2860c1c4a6f472d3df7772eaa62300

charleskorn (approved):
(meant to approve with the comment above 🤦)

lamida:
> Could you please add a benchmark for this and post a comparison with Prometheus' engine? Assuming the results look OK, I'm happy for this to be merged, but if there's any significant regression with MQE, then let's discuss before merging.

Sure I will do the benchmarking shortly. Thanks for checking.

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -0,0 +1,150 @@
...
+	return metadata, nil
+}
+
+func (a *Absent) NextSeries(_ context.Context) (types.InstantVectorSeriesData, error) {
+	output := types.InstantVectorSeriesData{}
+	defer func() { a.called = true }()
+
+	if a.called {
+		return output, types.EOS
+	}
```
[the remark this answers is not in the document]
lamida: Renamed to exhausted and refactor as suggested by removing defer. Except that output initiliasation still need on the top of the line before the first return.

c1536e7080af5831531b2b9ecfcda11eef397f5c

PR #10526, opened 2025-01-28
Modify mergeQueryable to allow reuse

zenador:
#### What this PR does

Refactor mergeQueryable so it can be reused outside of Mimir.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

Not a user-facing change.

### DIFF
--- a/pkg/querier/tenantfederation/merge_queryable.go
+++ b/pkg/querier/tenantfederation/merge_queryable.go
@@ -51,7 +51,7 @@ func NewQueryable(upstream storage.Queryable, bypassWithSingleID bool, maxConcur
 			}, nil
 		},
 	}
-	return NewMergeQueryable(defaultTenantLabel, callbacks, tenant.NewMultiResolver(), bypassWithSingleID, maxConcurrency, reg, logger)
+	return NewMergeQueryable(defaultTenantLabel, callbacks, tenant.NewMultiResolver(), bypassWithSingleID, maxConcurrency, reg, logger, nil)
 }
 
 // MergeQueryableCallbacks contains callbacks to NewMergeQueryable, for customizing its behaviour.
@@ -103,17 +103,11 @@ func (q *tenantQuerier) Close() error {
 // If the label `idLabelName` already exists, its value is overwritten and
 // the previous value is exposed through a new label prefixed with "original_".
 // This behaviour is not implemented recursively.
-func NewMergeQueryable(idLabelName string, callbacks MergeQueryableCallbacks, resolver tenant.Resolver, bypassWithSingleID bool, maxConcurrency int, reg prometheus.Registerer, logger log.Logger) storage.Queryable {
-	// Note that we allow tenant.Resolver to be injected instead of using the
-	// tenant.TenantIDs() method because GEM needs to inject different behavior
-	// here for the cluster federation feature.
-	return &mergeQueryable{
-		logger:             logger,
-		idLabelName:        idLabelName,
-		callbacks:          callbacks,
-		resolver:           resolver,
-		bypassWithSingleID: bypassWithSingleID,
-		maxConcurrency:     maxConcurrency,
+func NewMergeQueryable(idLabelName string, callbacks MergeQueryableCallbacks, resolver tenant.Resolver, bypassWithSingleID bool, maxConcurrency int, reg prometheus.Registerer, logger log.Logger, innerSelectFunc InnerSelectFunc) storage.Queryable {
+	if innerSelectFunc == nil {
+		innerSelectFunc = defaultInnerSelectFunc
+	}
+	metrics := mergeQueryableMetrics{
 		tenantsQueried: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
 			Name:    "cortex_querier_federation_tenants_queried",
 			Help:    "Number of tenants queried for a single standard query.",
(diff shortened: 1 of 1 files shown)


charleskorn (approved):

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -103,17 +103,11 @@ func (q *tenantQuerier) Close() error {
...
-	// tenant.TenantIDs() method because GEM needs to inject different behavior
-	// here for the cluster federation feature.
-	return &mergeQueryable{
-		logger:             logger,
-		idLabelName:        idLabelName,
-		callbacks:          callbacks,
-		resolver:           resolver,
-		bypassWithSingleID: bypassWithSingleID,
-		maxConcurrency:     maxConcurrency,
+func NewMergeQueryable(idLabelName string, callbacks MergeQueryableCallbacks, resolver tenant.Resolver, bypassWithSingleID bool, maxConcurrency int, reg prometheus.Registerer, logger log.Logger, innerSelectFunc InnerSelectFunc) storage.Queryable {
```
dimitarvdimitrov, line 106: can you document the new parameter in the godoc?

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -352,29 +367,36 @@ func (m *mergeQuerier) Select(ctx context.Context, sortSeries bool, hints *stora
...
+type InnerSelectFunc func(ctx context.Context, jobs []string, start time.Time, upstreamQueryWaitDuration prometheus.Histogram, upstream MergeQuerierUpstream, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, filteredMatchers ...*labels.Matcher) storage.SeriesSet
+
+func defaultInnerSelectFunc(ctx context.Context, jobs []string, start time.Time, upstreamQueryWaitDuration prometheus.Histogram, upstream MergeQuerierUpstream, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, filteredMatchers ...*labels.Matcher) storage.SeriesSet {
+	seriesSets := make([]storage.SeriesSet, len(jobs))
+
 	// We don't use the context passed to this function, since the context has to live longer
 	// than the call to ForEachJob (i.e. as long as seriesSets)
 	run := func(_ context.Context, idx int) error {
-		m.upstreamQueryWaitDuration.Observe(time.Since(start).Seconds())
+		upstreamQueryWaitDuration.Observe(time.Since(start).Seconds())
```
dimitarvdimitrov, line 385: we can get rid of `upstreamQueryWaitDuration` and `start` by wrapping the `upstream` with our own implementation which tracks these. That way other implementations don't need to remember to record the metric.

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -352,29 +367,36 @@ func (m *mergeQuerier) Select(ctx context.Context, sortSeries bool, hints *stora
...
 	jobs := make([]string, 0, len(matchedIDs))
-	seriesSets := make([]storage.SeriesSet, len(matchedIDs))
 	for id := range matchedIDs {
 		jobs = append(jobs, id)
 	}
 
+	return m.innerSelectFunc(ctx, jobs, start, m.upstreamQueryWaitDuration, m.upstream, m.idLabelName, m.maxConcurrency, sortSeries, hints, filteredMatchers...)
+}
+
+type InnerSelectFunc func(ctx context.Context, jobs []string, start time.Time, upstreamQueryWaitDuration prometheus.Histogram, upstream MergeQuerierUpstream, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, filteredMatchers ...*labels.Matcher) storage.SeriesSet
```
dimitarvdimitrov, line 377: let's come up with a better name. Currently inner refers to the fact that it happens within another select func, but in reality there are a dozen nested select funcs. How about `MultiTenantSelectFunc` (the package is called tenantfederation, so this func is supposed to run the select for multiple tenants)?


In a godoc can you also explain what this function is supposed to be used for and what some of the parameters mean? like `jobs` is very unclear, how is `idLabelName` intended to be used and that `sortSeries, hints, filteredMatchers` are there only to be passed to the `upstream`

Also can we call `filteredMatchers` `matchers`?

dimitarvdimitrov (commented):
i know we spoke that we might not keep this for long, but we don't actually know. So i'd prefer if we leave a slightly smaller mess in case it stays there for years 😄

I think it should be fast to address these comments. Feel free to ping me directly on slack if you've got doubts

pkg/querier/tenantfederation/merge_queryable.go
```diff
@@ -371,21 +375,30 @@ func (m *mergeQuerier) Select(ctx context.Context, sortSeries bool, hints *stora
...
+// MultiTenantSelectFunc is an overrideable function that queries multiple tenants in parallel.
+// By default, jobs would be the list of tenant IDs, idLabelName is used to add a label to the series
+// to identify the tenant it belongs to, maxConcurrency is the maximum number of concurrent queries allowed,
+// and ctx, sortSeries, hints, and matchers are the same as the Select method.
+type MultiTenantSelectFunc func(ctx context.Context, jobs []string, wrMergeQuerier WaitRecordingMergeQuerier, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet
 
-func defaultInnerSelectFunc(ctx context.Context, jobs []string, start time.Time, upstreamQueryWaitDuration prometheus.Histogram, upstream MergeQuerierUpstream, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, filteredMatchers ...*labels.Matcher) storage.SeriesSet {
+func defaultMultiTenantSelectFunc(ctx context.Context, jobs []string, wrMergeQuerier WaitRecordingMergeQuerier, idLabelName string, maxConcurrency int, sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
```
dimitarvdimitrov, line 393: shouldn't we still pass `MergeQuerierUpstream`? `WaitRecordingMergeQuerier` should be just an implementation of that interface

PR #10533, opened 2025-01-29
MQE: fix functions that could return multiple series with the same labels

charleskorn:
#### What this PR does

This PR fixes an issue with the `clamp`, `clamp_min`, `clamp_max`, `round` and `histogram_fraction` functions in MQE: they could return multiple series with the same labels if multiple series with the same labels but different metric names were passed to the function.

This PR also adds a test to ensure we catch issues like these when new functions are added in the future.

#10067

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/functions_test.go
+++ b/pkg/streamingpromql/functions_test.go
@@ -3,10 +3,18 @@
 package streamingpromql
 
 import (
+	"context"
+	"fmt"
+	"strings"
 	"testing"
+	"time"
 
+	"github.com/go-kit/log"
+	"github.com/prometheus/prometheus/model/timestamp"
+	"github.com/prometheus/prometheus/promql/promqltest"
 	"github.com/stretchr/testify/require"
 
+	"github.com/grafana/mimir/pkg/querier/stats"
 	"github.com/grafana/mimir/pkg/streamingpromql/operators/functions"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/functions_test.go
```diff
@@ -50,3 +58,94 @@ func TestRegisterScalarFunctionOperatorFactory(t *testing.T) {
 	// Cleanup changes to instantVectorFunctionOperatorFactories
 	delete(scalarFunctionOperatorFactories, "new_function")
 }
+
+// This test ensures that all functions correctly merge series after dropping the metric name.
+func TestFunctionDeduplicateAndMerge(t *testing.T) {
+	data := `
+		load 30s
```
jhesketh, line 65: We should use spaces here as tabs don't ensure alignment for all

image

pkg/streamingpromql/functions_test.go
```diff
@@ -50,3 +58,94 @@ func TestRegisterScalarFunctionOperatorFactory(t *testing.T) {
...
+		"histogram_quantile": "0.1",
+		"histogram_fraction": "0, 0.1",
+	}
+
+	postSelectorFunctionArgs := map[string]string{
+		"clamp":         "-Inf, Inf",
+		"clamp_min":     "-Inf",
+		"clamp_max":     "Inf",
+		"label_replace": `"__name__", "$1", "env", "(.*)"`,
+	}
```
jhesketh, line 92: (nit) I feel like it would be simpler to just define a query string for each function, and then assert that the number of functions checked is equal to `len(instantVectorFunctionOperatorFactories)`

pkg/streamingpromql/functions_test.go
```diff
@@ -50,3 +58,94 @@ func TestRegisterScalarFunctionOperatorFactory(t *testing.T) {
...
+				postSelectorArgs = ", " + postSelectorArgs
+			}
+
+			expr := fmt.Sprintf("%s(%s%s%s)", name, preSelectorArgs, selector, postSelectorArgs)
+
+			q, err := engine.NewRangeQuery(ctx, storage, nil, expr, start, end, step)
+			require.NoError(t, err)
+			defer q.Close()
+
+			mimirResult := q.Exec(ctx)
```
jhesketh, line 130: I think we should consider passing this into Prometheus too so that we can verify we merge in the same way (as I think we've found a case previous where we don't).

Additionally, I think we should consider seeing if we can make this part of the gauntlet. ie make sure we have cases represented in there along with the selectors that would cause this to happen.

jhesketh (approved):
I'm going to merge this so we get it into next weeks release. However I think it's worth coming back to these nits.

pkg/streamingpromql/functions_test.go
charleskorn: 🤦 Fixed in https://github.com/grafana/mimir/pull/10553

pkg/streamingpromql/functions_test.go
charleskorn: I've done something like this in https://github.com/grafana/mimir/pull/10553

pkg/streamingpromql/functions_test.go
charleskorn: I originally had this, but unfortunately a bunch of these cases fail with Prometheus' engine due to https://github.com/prometheus/prometheus/issues/14695.

> Additionally, I think we should consider seeing if we can make this part of the gauntlet. ie make sure we have cases represented in there along with the selectors that would cause this to happen.

I think adding these cases to the gauntlet would be tricky, and I think having a dedicated test for this specific problem helps make the issue (and the solution) clear for any future functions we add that are missing the deduplication step.

PR #10536, opened 2025-01-29
ruler: add "user" and "reason" labels to ruler's queries metrics

narqo:
#### What this PR does

In this PR I'm adding the `user` and `reason` labels to the `cortex_ruler_queries_failed_total` and `cortex_ruler_write_requests_failed_total` metrics. The idea behind the change is to allow tenants segregate and track separately the failures, that happened due to issues on the server vs. those happened due to issues on the client-side (e.g. bad rule).

The exact list of "reasons" is up for a discussion. I've noticed that Loki uses `error` and `upstream_error` ([code](https://github.com/grafana/loki/blob/4df8c60b6ba9f89ee3c8965e78dfae376e0ed929/pkg/ruler/evaluator_remote.go#L260-L273)). I'm not sure if those are well translated into what we want. Opinions are welcome.

For now, I've chosen to have only two reasons: 

- `error` for everything 
- `4xx` for the case when there is *a client-side error in the remote-querier* — I think, this will create the most value to start with. I think, we can expand/change the groups in the future. What do you think?

Note, for consistency, I also added the `user` labels to `cortex_ruler_write_requests_total` and `cortex_ruler_queries_total`.

---

**TODO**

- [ ] Decide the set of "reason" values
- [x] Update the existing alerts, that currently expect only server-side failures

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
+++ b/operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
@@ -453,7 +453,7 @@ spec:
             expr: |
               100 * (
               # Here it matches on empty "reason" for backwards compatibility, with when the metric didn't have this label.
-              sum by (cluster, namespace, pod) (rate(cortex_ruler_write_requests_failed_total{reason=~"(error|$^)"}[1m]))
+              sum by (cluster, namespace, pod) (rate(cortex_ruler_write_requests_failed_total{reason=~"(error|^$)"}[1m]))
                 /
               sum by (cluster, namespace, pod) (rate(cortex_ruler_write_requests_total[1m]))
               ) > 1
@@ -468,7 +468,7 @@ spec:
             expr: |
               100 * (
               # Here it matches on empty "reason" for backwards compatibility, with when the metric didn't have this label.
-              sum by (cluster, namespace, pod) (rate(cortex_ruler_queries_failed_total{reason=~"(error|$^)"}[1m]))
+              sum by (cluster, namespace, pod) (rate(cortex_ruler_queries_failed_total{reason=~"(error|^$)"}[1m]))
                 /
               sum by (cluster, namespace, pod) (rate(cortex_ruler_queries_total[1m]))
               ) > 1
--- a/operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
+++ b/operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
@@ -452,7 +452,7 @@ spec:
               runbook_url: https://grafana.com/docs/mimir/latest/operators-guide/mimir-runbooks/#mimirrulertoomanyfailedpushes
             expr: |
               100 * (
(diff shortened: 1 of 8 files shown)


pkg/ruler/compat.go
```diff
@@ -233,17 +236,17 @@ func MetricsQueryFunc(qf rules.QueryFunc, queries, failedQueries prometheus.Coun
...
 			// Return unwrapped error.
 			return result, origErr
 		} else if remoteQuerier {
-			// When remote querier enabled, consider anything an error except those with 4xx status code.
-			st, ok := grpcutil.ErrorToStatus(err)
-			if !(ok && st.Code()/100 == 4) {
-				failedQueries.Inc()
+			// When remote querier enabled, consider anything an "error" except those with 4xx status code.
+			if mimirpb.IsClientError(err) {
+				failureReason = "4xx"
```
dimitarvdimitrov, line 247: can we call them `client_error` and `server_error` instead, so they follow the same pattern?

operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
```diff
@@ -452,7 +452,8 @@ spec:
               runbook_url: https://grafana.com/docs/mimir/latest/operators-guide/mimir-runbooks/#mimirrulertoomanyfailedpushes
             expr: |
               100 * (
-              sum by (cluster, namespace, pod) (rate(cortex_ruler_write_requests_failed_total[1m]))
+              # Here it matches on empty "reason" for backwards compatibility, with when the metric didn't have this label.
+              sum by (cluster, namespace, pod) (rate(cortex_ruler_write_requests_failed_total{reason=~"(error|$^)"}[1m]))
```
dimitarvdimitrov, line 456: the regex matches `EndStart` instead of `StartEnd` is this what you intended?

pkg/ruler/compat.go
```diff
@@ -329,23 +332,23 @@ func DefaultTenantManagerFactory(
...
 	reg prometheus.Registerer,
 ) ManagerFactory {
-	totalWrites := promauto.With(reg).NewCounter(prometheus.CounterOpts{
+	totalWrites := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "cortex_ruler_write_requests_total",
 		Help: "Number of write requests to ingesters.",
-	})
-	failedWrites := promauto.With(reg).NewCounter(prometheus.CounterOpts{
+	}, []string{"user"})
+	failedWrites := promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
```
dimitarvdimitrov, line 339: don't forget to make this change in GEM's codebase too

dimitarvdimitrov (commented):
sorry for the scattered review comments

operations/helm/tests/metamonitoring-values-generated/mimir-distributed/templates/metamonitoring/mixin-alerts.yaml
narqo: Ooops, thank you.

pkg/ruler/compat.go
narqo: Oh. I didn't know there is an override. I will open a PR with the update there right after merging this one.

dimitarvdimitrov (approved):
thanks! LGTM

PR #10537, opened 2025-01-29
Fix instance where context cancellation error was cached

56quarters:
#### What this PR does

This change fixes the error mapping done in the query sharding middleware to correctly map all possible errors from the PromQL engine to an appropriate APIError type. This fixes an issue where `context.Canceled` was treated as an execution error and cached for subsequent requests.

As part of this, instrumentation for the error caching middleware was improved and Jaeger tracing in various Docker compose setups was fixed by pinning the image used.

Internal escalation.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
 * [BUGFIX] Query-frontend and querier: show warning/info annotations in some cases where they were missing (if a lazy querier was used). #10277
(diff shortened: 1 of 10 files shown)


56quarters:
For some context, this is the code in the PromQL engine that can return a `context.Canceled` error instead of one of the expected types: https://github.com/grafana/mimir/blob/17121779f12f1b5f8dc5ba1a2c6a5c22981fd692/vendor/github.com/prometheus/prometheus/promql/engine.go#L1039-L1049

dimitarvdimitrov (approved):
SGTM, but i suppose something changed in the engine :( 

thanks for fixing it

dimitarvdimitrov:
> something changed in the engine

I now saw your comment. Does this mean this bug has been here ever since?

56quarters:
> > 
> 
> I now saw your comment. Does this mean this bug has been here for a while? That code was added 5 years ago

This behavior (query sharding treating a `context.Canceled` as `apierror.TypeExecution`) would have been here since the beginning but presumably wouldn't have affected anything since when the context is canceled the user has already hung up and doesn't care about the response.

PR #10538, opened 2025-01-29
block-builder: Add cortex_blockbuilder_blocks_produced_total metric

codesome:
#### What this PR does

Adds a metric to determine how many blocks were produced for previous, current, and next block period w.r.t. the cycle end time.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/blockbuilder.go
```diff
@@ -74,10 +75,6 @@ func newWithSchedulerClient(
 	limits *validation.Overrides,
 	schedulerClient schedulerpb.SchedulerClient,
 ) (*BlockBuilder, error) {
-	if cfg.NoPartiallyConsumedRegion {
-		// We should not have a large buffer if we are putting all the records into a block.
-		cfg.ConsumeIntervalBuffer = 5 * time.Minute
-	}
```
codesome, line 80: By removing this override, we can now try different buffer times when we have set NoPartiallyConsumedRegion to true

pkg/blockbuilder/blockbuilder.go
```diff
@@ -643,6 +645,22 @@ consumerLoop:
 	return newState, nil
 }
 
+func getBlockCategoryCount(sectionEndTime time.Time, blockMetas []tsdb.BlockMeta) (prev, curr, next int) {
+	// Doing -30m will take care
```
narqo, line 649: (nit) What does this comment say?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -610,11 +607,16 @@ consumerLoop:
...
-	numBlocks, err = builder.CompactAndUpload(ctx, b.uploadBlocks)
+	blockMetas, err = builder.CompactAndUpload(ctx, b.uploadBlocks)
 	if err != nil {
 		return state, err
 	}
 
+	prev, curr, next := getBlockCategoryCount(sectionEndTime, blockMetas)
+	b.blockBuilderMetrics.blockCounts.WithLabelValues("previous").Add(float64(prev))
+	b.blockBuilderMetrics.blockCounts.WithLabelValues("current").Add(float64(curr))
+	b.blockBuilderMetrics.blockCounts.WithLabelValues("next").Add(float64(next))
```
narqo, line 618: Have you considered putting these metrics inside `TSDBBuilder.CompactAndUpload` (ie. make them a part of `tsdbMetrics` collection). It feels this would require fewer changes in the overall code, and make "magic knowledge", like 2h block ranges, scoped closer together.

codesome: That was my first try, but giving the information about current block end to TSDB did not look right. I am not too worried about the number of lines of change here because it's mostly replacing old stuff and only a minor addition of this function.

pkg/blockbuilder/blockbuilder.go
codesome: Whoops looks like I got distracted midway. Will fix it.

narqo (approved):
🔥 Works for me

PR #10545, opened 2025-01-30
influx: rename extra label to match existing series

alexgreenbank:
#### What this PR does

Minor adjustment to experimental feature.

Just tweaking the extra label to match that of the influx2cortex proxy. If we don't match the labels on ingestion it means series will be different when/if customers migrate from the influx2cortex proxy to this Influx endpoint in Mimir.

Following on from https://github.com/grafana/mimir/pull/10153

#### Checklist

- [x] Tests updated.

No need to modify changelog, documentation or versioning info.


aknuds1 (approved):
LGTM

PR #10547, opened 2025-01-30
Add a flag to adjust queries to some max future time

chencs:
#### What this PR does
This PR introduces an experimental flag, `query-frontend.adjust-to-max-future-query-window`. When set with a duration, any range queries querying farther than `now + maxFutureWindow` into the future will be adjusted to query only up to `now + maxFutureWindow`. 

Fixes #

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_evictions_total`
 * [ENHANCEMENT] Compactor: Shuffle users' order in `BlocksCleaner`. Prevents bucket indexes from going an extended period without cleanup during compactor restarts. #10513
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
-* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into future to a max duration with `-query-frontend.adjust-to-max-future-window`. #10547
+* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
 * [BUGFIX] Query-frontend and querier: show warning/info annotations in some cases where they were missing (if a lazy querier was used). #10277
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_evictions_total`
 * [ENHANCEMENT] Compactor: Shuffle users' order in `BlocksCleaner`. Prevents bucket indexes from going an extended period without cleanup during compactor restarts. #10513
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
+* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into future to a max duration with `-query-frontend.adjust-to-max-future-window`. #10547
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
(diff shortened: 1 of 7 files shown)


cmd/mimir/config-descriptor.json
```diff
@@ -4369,6 +4369,17 @@
...
           "fieldCategory": "experimental"
         },
+        {
+          "kind": "field",
+          "name": "max_future_query_window",
+          "required": false,
+          "desc": "Mutate incoming queries that look far into the future to only look into the future by the set duration. 0 to disable.",
+          "fieldValue": null,
+          "fieldDefaultValue": 0,
+          "fieldFlag": "query-frontend.adjust-to-max-future-query-window",
```
narqo, line 4379: (nit) Can we align the names of the flag and the name of the config? E.g. `-query-frontend.max-future-query-window`. This will make it simpler to reason about the values in the overrides.

narqo (commented):
I think, in the current form, the new limit won't touch queries that were modified by the `max-lookback` limit (because the latter happens before the new one). Is this what we want for this PR?

chencs:
>  I think, in the current form, the new limit won't touch queries that were modified by the max-lookback limit — because the latter adjusts the query and moves on, before the new one takes place. Is this what we want for this PR?

I'm not sure what you mean by this. IIUC queries which fully query outside of the allowed `max-lookback` will not  be executed, but requests which are only _modified_ by `max-lookback` will continue on to be checked against other limits: https://github.com/grafana/mimir/blob/bfde5dbf82363e9d3ed2adb9b9f837cc065cc5ae/pkg/frontend/querymiddleware/limits.go#L180

cmd/mimir/config-descriptor.json
chencs: Done!

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
   * `cortex_ingester_tsdb_block_postings_for_matchers_cache_evictions_total`
 * [ENHANCEMENT] Compactor: Shuffle users' order in `BlocksCleaner`. Prevents bucket indexes from going an extended period without cleanup during compactor restarts. #10513
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
+* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into future to a max duration with `-query-frontend.adjust-to-max-future-window`. #10547
```
tacole02, line 44:
```suggestion
* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with `-query-frontend.adjust-to-max-future-window`. #10547
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3625,6 +3625,11 @@ The `limits` block configures default and per-tenant limits imposed by component
 # optimize their performance.
 [instant_queries_with_subquery_spin_off: <list of strings> | default = ]
 
+# (experimental) Mutate incoming queries that look far into the future to only
```
tacole02, line 3628: Could we use a word other than `mutate`? `Update` maybe?

tacole02 (approved):
Thank you, @chencs !

CHANGELOG.md
narqo, line 44:
```suggestion
* [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into future to a max duration with experimental `-query-frontend.max-future-query-window` flag. #10547
```

narqo (approved):
> queries which fully query outside of the allowed max-lookback will not be executed, but requests which are only modified by max-lookback will continue on to be checked against other limits

You are right (I read the surrounding code wrong).

🔥 The changes looks good to me (except the comments about the changelog)

docs/sources/mimir/configure/configuration-parameters/index.md
chencs: I was following the pattern of the `query-frontend.align-queries-with-step` description. Is "mutate" an ineffective word choice? This does _change_ the actual query being executed.

tacole02: If `mutate` is the technically accurate word choice, we can leave it. It just sounds somewhat awkward/jarring to me.

chencs:
> We just removed basically the same option in 2.14

I didn't realize! Thank you for pointing that out.

> The assumption was that these queries are very cheap to execute.

This is true, but if a tenant tries to execute a future-looking query over many days, they can also quickly hit `query-scheduler.max-outstanding-requests-per-tenant`, and it can take a while for those "cheap" queries to clear in an environment with many tenants. I thought about checking the query for a `predict_linear` and not applying the limit if found; do you think this would be a reasonable workaround, or would it make this flag too confusing?

PR #10548, opened 2025-01-31
Expose 'ingestion_artificial_delay_msec' via overrides-exporter

pracucci:
#### What this PR does

I would like to expose 'ingestion_artificial_delay_msec' via overrides-exporter. I haven't added to the list of default exported config options, so that it needs to be explicitly enabled to get exposed.

Example from local dev env (I've configured the override for `anonymous` tenant at 10:23):

Screenshot 2025-01-31 at 10 25 47

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


gotjosh (approved):
LGTM

56quarters:
Why are we using `msec` here instead of the base unit of seconds? We use seconds elsewhere, even for gauges.

pracucci:
> Why are we using `msec` here instead of the base unit of seconds? We use seconds elsewhere, even for gauges.

Address here: https://github.com/grafana/mimir/pull/10549

PR #10558, opened 2025-02-03
MQE: fix issue where `group_left` / `group_right` could return series with incorrect labels

charleskorn:
#### What this PR does

This PR fixes a bug where MQE's implementation of `group_left` / `group_right` could sometimes return incorrect labels for a series.

In particular, this could happen if the number of additional labels in the `group_left(...)` or `group_right(...)` modifier was not a power of two. `VectorMatching.Include` was effectively being mutated in `manySideGroupKeyFunc` to include `__name__` and drop the last label (rather than being copied to a new slice with `__name__` appended), and so the incorrect set of additional label names was used in `outputSeriesLabelsFunc`. 

If the list of additional labels had a length that was a power of two, then the `append` in `manySideGroupKeyFunc` caused a new underlying array to be created and so the bug was not triggered.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
+++ b/pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
@@ -416,7 +416,9 @@ func (g *GroupedVectorVectorBinaryOperation) manySideGroupKeyFunc() func(manySid
 	labelsToRemove := g.VectorMatching.Include
 
 	if g.shouldRemoveMetricNameFromManySide() {
+		labelsToRemove = make([]string, 0, len(g.VectorMatching.Include)+1)
 		labelsToRemove = append(labelsToRemove, labels.MetricName)
+		labelsToRemove = append(labelsToRemove, g.VectorMatching.Include...)
 		slices.Sort(labelsToRemove)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
```diff
@@ -416,7 +416,9 @@ func (g *GroupedVectorVectorBinaryOperation) manySideGroupKeyFunc() func(manySid
 	labelsToRemove := g.VectorMatching.Include
 
 	if g.shouldRemoveMetricNameFromManySide() {
+		labelsToRemove = make([]string, 0, len(g.VectorMatching.Include)+1)
```
jhesketh, line 419: Not for this change, but should we consider using a string pool?

jhesketh (approved):
Nice solve.

pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
charleskorn: Would be interesting to try, but I suspect the impact will be small given we tend to only use string slices for things that are created once per query (eg. this list of label names).

Issue #9916, opened 2024-11-15
Flaky `TestConcurrentFetchers`


flxbk:
This test seems to be flaky ([CI run](https://github.com/grafana/mimir/actions/runs/11853375063/job/33033407798?pr=9915))

```
--- FAIL: TestConcurrentFetchers (2.11s)
    --- FAIL: TestConcurrentFetchers/update_concurrency_with_continuous_production (10.00s)
        logging.go:33: level info component kafka_client msg immediate metadata update triggered why from user ForceMetadataRefresh
        logging.go:33: level info component kafka_client msg immediate metadata update triggered why forced load because we are producing to a topic for the first time
        logging.go:33: level info component kafka_client msg producing to a new topic for the first time, fetching metadata to learn its partitions topic test-topic
        logging.go:33: level info component kafka_client msg done waiting for metadata for new topic topic test-topic
        logging.go:33: level info component kafka_client msg initializing producer id
        logging.go:33: level info component kafka_client msg producer id initialization success id 8363768319828142695 epoch 0
        reader_test.go:2332: 
            	Error Trace:	/__w/mimir/mimir/pkg/storage/ingest/reader_test.go:2332
            	            				/__w/mimir/mimir/pkg/storage/ingest/fetcher_test.go:650
            	            				/usr/local/go/src/runtime/asm_amd64.s:1700
            	Error:      	Received unexpected error:
            	            	context canceled
            	Test:       	TestConcurrentFetchers/update_concurrency_with_continuous_production
        fetcher_test.go:692: 
            	Error Trace:	/__w/mimir/mimir/pkg/storage/ingest/fetcher_test.go:692
            	Error:      	Not equal: 
            	            	expected: 646
            	            	actual  : 647
            	Test:       	TestConcurrentFetchers/update_concurrency_with_continuous_production
            	Messages:   	Should not fetch more records than produced
        fetcher_test.go:706: Total produced: 647, Total fetched: 646
        fetcher_test.go:707: Fetched with initial concurrency: 199
        fetcher_test.go:708: Fetched with high concurrency: 300
        fetcher_test.go:709: Fetched with low concurrency: 147
FAIL
FAIL	github.com/grafana/mimir/pkg/storage/ingest	117.983s
```

zenador:
Looks like it's still happening:
https://github.com/grafana/mimir/actions/runs/12400788936/job/34618765095?pr=10277
```
[...]

PR #10567, opened 2025-02-03
Fix flaky `TestConcurrentFetchers_fetchSingle`

dimitarvdimitrov:
#### Why it's falky

The test was failing because the control function was used up by the periodic fetch loop of the started `concurrentFetchers`. This meant that the test sometimes ran when we don't have a control function registered and didn't inject an error.

At the same time the test cases didn't retain their control functions so that they don't interfere with each other.

#### What this PR does

This change makes the test cases independent, so we can use `KeepControl`.

Fixes https://github.com/grafana/mimir/issues/9916

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/fetcher_test.go
+++ b/pkg/storage/ingest/fetcher_test.go
@@ -945,7 +945,6 @@ func TestConcurrentFetchers_fetchSingle(t *testing.T) {
 			fetchers             = createConcurrentFetchers(ctx, t, client, topic, partitionID, 0, 1, 0)
 		)
 		t.Cleanup(cluster.Close)
-		t.Cleanup(fetchers.Stop)
 
 		// Produce some records.
 		produceRecord(ctx, t, client, topic, partitionID, []byte("record-1"))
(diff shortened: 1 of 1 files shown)


pkg/storage/ingest/fetcher_test.go
```diff
@@ -936,19 +937,25 @@ func TestConcurrentFetchers_fetchSingle(t *testing.T) {
...
-	)
+	setup := func(t *testing.T) (*concurrentFetchers, *kfake.Cluster, context.Context) {
+		var (
+			ctx                  = context.Background()
+			cluster, clusterAddr = testkafka.CreateCluster(t, partitionID+1, topic)
+			client               = newKafkaProduceClient(t, clusterAddr)
+			fetchers             = createConcurrentFetchers(ctx, t, client, topic, partitionID, 0, 1, 0)
+		)
+		t.Cleanup(cluster.Close)
+		t.Cleanup(fetchers.Stop)
```
dimitarvdimitrov, line 948: this was already done in `createConcurrentFetchers`

```suggestion
```

PR #10569, opened 2025-02-03
Expose index_last_successful_update before starting block cleanup service

tinitiuset:
#### What this PR does

This PR exposes `cortex_bucket_index_last_successful_update_timestamp_seconds` for all tenants assigned to the compactor before starting the block cleanup job for any tenant.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -154,6 +154,8 @@ func NewBlocksCleaner(cfg BlocksCleanerConfig, bucketClient objstore.Bucket, own
 		}),
 	}
 
+	c.instrumentBeforeStartBlocksCleaner(context.Background())
+
 	c.Service = services.NewTimerService(cfg.CleanupInterval, c.starting, c.ticker, c.stopping)
 
 	return c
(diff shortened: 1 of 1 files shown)


pkg/compactor/blocks_cleaner.go
```diff
@@ -206,6 +208,22 @@ func (c *BlocksCleaner) runCleanup(ctx context.Context, async bool) {
 	}
 }
 
+func (c *BlocksCleaner) instrumentBeforeStartBlocksCleaner(ctx context.Context) {
```
dimitarvdimitrov, line 211: instrumentBeforeStartBlocksCleaner -> instrumentBucketIndexUpdate

dimitarvdimitrov, line 211: also talked about calling this from `starting` instead of from the constructor

dimitarvdimitrov, line 211: would be nice if we can test this. See `ErrorInjectedBucketClient`

tinitiuset: Addressed in `26641d7`. Thank you!

tinitiuset: Addressed in `26641d7`. Thank you!

tinitiuset: Should be addressed in `835891b`. Let me know what you think!

pkg/compactor/blocks_cleaner.go
```diff
@@ -167,6 +165,7 @@ func (c *BlocksCleaner) stopping(error) error {
 }
 
 func (c *BlocksCleaner) starting(ctx context.Context) error {
+	c.instrumentBucketIndexUpdate(context.Background())
```
dimitarvdimitrov, line 168: can we use the contexts that's here?

pkg/compactor/compactor_test.go
```diff
@@ -305,7 +305,7 @@ func TestMultitenantCompactor_ShouldRetryCompactionOnFailureWhileDiscoveringUser
 	require.NoError(t, services.StopAndAwaitTerminated(context.Background(), c))
 
 	// Ensure the bucket iteration has been retried the configured number of times.
-	bucketClient.AssertNumberOfCalls(t, "Iter", 1+3)
+	bucketClient.AssertNumberOfCalls(t, "Iter", 1+1+3)
```
dimitarvdimitrov, line 308: can you add a comment explaining the `1+1+3`? or at the very least explain one of the `+1`s

tacole02 (approved):

pkg/compactor/blocks_cleaner.go
tinitiuset: Thanks for noticing. Definitely!

pkg/compactor/compactor_test.go
tinitiuset: Sure! `6a46bda`

PR #10574, opened 2025-02-04
Add reactive limiters for ingesters

jhalterman:
#### What this PR does

This PR introduces reactive concurrency limiters into Mimir and integrates separate limiters into ingesters for push and read operations. These are meant to protect ingesters from overload by adjusting concurrency limits as needed. When limiters are full, some amount of requests may queue. The inflight limit and the amount of queueing that are allowed are regularly adjusted using an algorithm that considers recent latency and throughput. When latency increases sharply or throughput appears to be deteriorating the limit is decreased, otherwise it may be increased up to some point.

#### Detecting overload from latency

Reactive limiters are concurrency limiters that adjust their limit based on overload. Overload is detected primarily by using a [TCP Vegas](https://en.wikipedia.org/wiki/TCP_Vegas) style algorithm where recent latency changes are converted to estimated request queue sizes and are thresholded to either increase, decrease, or hold the limit.

The queue size is calculated by comparing "short-term RTT" (latency) and "long-term RTT" measurements. Short RTT represents recent sample measurements which have been aggregated over a short window (ex: 1s) using a TDigest. These are additionally run through a median filter and smoothed to avoid limit oscillations that can happen with highly variable workloads such as queries. The long RTT is an EWMA covering a longer time window, such as 1 minute, representing the baseline load on the system, which gradually adjusts based on short RTTs.

#### Detecting overload from throughput

A challenge with Vegas implementations is that we want the baseline to periodically or gradually adjust, and in some cases it may adjust upwards which makes room for further upward adjustment. Under moderate load, this can cause the limit and latency to increase uncontrollably.

To solve this problem, we introduce an additional way of detecting overload: throughput [correlation](https://en.wikipedia.org/wiki/Correlation_coefficient). This indicates how strongly increases in throughput correlate with increases in inflight requests. When this correlation flatlines or goes negative, we know that the system is overloaded, and we should decrease the limit regardless of what the queue size indicates. We also track RTT correlation to confirm situations where throughput flatlines but RTT [variation](https://en.wikipedia.org/wiki/Coefficient_of_variation) is high.

#### Limit behavior

When a system is not overloaded, the limit will be driven up to some multiple (max_limit_factor) of the current inflight requests. This is the common behavior - to increase the limit unless the queue size is very large. Driving the limit up is important to give headroom to incoming work, but also for detecting if/when overload occurs. If the system is overloaded while the limit is being driven up, this will be detected as an increase in latency or decrease in throughput, and the limit will adjust accordingly. 

When a system is overloaded, the limit will gradually decrease to wherever the system's capacity happens to be, which is where latency and throughput stop suffering, then will oscillate around that point as long as the overload continues. This is the desired behavior, as it allows the limiter to adapt to and discover any system's capacity.

#### Blocking

When not overloaded, the concurrency limit will be set to some multiple of the inflight requests, and when overloaded, the limit will discover and be set based on the system's capacity. For example: if the system has 8 cores and is performing CPU bound work, the limit might naturally settle around 8. Since an inflight limit that low could be somewhat strict when it comes to handling short bursts of work, the reactive limiter allows blocking (queueing) of requests before they are accepted for processing, up to some limit. This allows us to keep the number of actual inflight requests small to better guard against overload, without being overly aggressive at rejecting requests. The amount of blocking that is allowed is based on the initial-rejection-factor and max-rejection-factor. Requests only block/queue when the limiter is full.

#### Prioritization

When an ingester is overloaded, we prefer to reject reads before pushes. To accomplish this, a Prioritizer component periodically inspects recent in/out queue stats for each of the limiters and determines if the system is overloaded and should reject requests. If rejection is needed, the Prioritizer determines which % of requests should be rejected, and based on the distribution of recent request priorities, it determines what priority below which requests should be rejected. This allows, for example, all reads to be rejected but not pushes, or some portion of pushes to be rejected. In this PR, pushes are prioritized as High, and reads as Low. The actual priorities don't matter when it comes to rejection, only relative priorities.

#### Separate push and query limiters

Separate adaptlive limiters are used for push and read requests. This allows those two workloads to naturally set different limits, based on their unique resource usages. For example, pushes could become constrained by disk whereas queries could become constrained by memory. If one type of operation is constrained by some resource, there's no need to punish other operations that are not. Having different limiters for the different workloads allows for this.

#### gRPC call checking

The ingester's reactive limiter uses a two-step check to permit gRPC calls:
- `CanAcquirePermit` is called via a gRPC tap handle, which allows requests to be quickly rejected before too many resources are allocated, when the limiter and queue are at capacity.
- `AcquirePermit` is called via a gRPC interceptor to actually acquire a permit. The reason we use an interceptor rather than using a tap handle for this is because `AcquirePermit` is a blocking call. If the limiter is full, this may block, as described above, up to some limit.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/util/adaptivelimiter/adaptivelimiter.go
```diff
@@ -0,0 +1,400 @@
...
+	alphaFunc, betaFunc        func(int) int
+	increaseFunc, decreaseFunc func(int) int
+}
+
+func (cfg *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, prefix+"enabled", false, "Enable adaptive limiting when making requests to ingesters")
+	f.DurationVar(&cfg.ShortWindowMinDuration, prefix+"short-window-min-duration", time.Second, "Min duration of the window that is used to determine recent, short-term load on the system")
+	f.DurationVar(&cfg.ShortWindowMaxDuration, prefix+"short-window-max-duration", 30*time.Second, "Max duration of the window that is used to determine recent, short-term load on the system")
+	f.UintVar(&cfg.ShortWindowMinSamples, prefix+"short-window-min-samples", 1, "Min number of samples that must be recorded in the window")
+	f.UintVar(&cfg.LongWindow, prefix+"long-window", 60, "Short-term window measurements that will be stored in an exponentially weighted moving average window, representing the long-term baseline inflight time")
```
56quarters, line 118: Is this description wrong?

56quarters (commented):
Overall, this looks pretty good to me. Nice job!

I didn't go very deep into the TCP vegas-like algorithm. At a high-level, the only thing that worries me is that it's potentially active at the same time as circuit breakers and utilization limits which seems like it makes the behavior of ingesters very difficult to predict.

pkg/util/math/median.go
```diff
@@ -0,0 +1,47 @@
...
+	f.values[f.index] = value
+	f.index = (f.index + 1) % len(f.values)
+
+	if f.size < len(f.values)-1 {
+		f.size++
+		return value
+	}
+
+	copy(f.sorted, f.values)
+	sort.Slice(f.sorted, func(i, j int) bool {
```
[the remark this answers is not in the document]
56quarters: Nevermind, I see that this is called at most once per-second.

jhalterman:
> At a high-level, the only thing that worries me is that it's potentially active at the same time as circuit breakers and utilization limits which seems like it makes the behavior of ingesters very difficult to predict.

Agree. Each limiter has an opportunity to reject a request. A circuit breaker should only care about timeouts, so if an adaptive limiter rejects a request, the circuit breaker should ignore it.

pkg/util/adaptivelimiter/adaptivelimiter.go
```diff
@@ -0,0 +1,400 @@
...
+		longRTT:               math2.NewEwma(config.LongWindow, warmupSamples),
+		nextUpdateTime:        time.Now(),
+		rttCorrelation:        math2.NewCorrelationWindow(config.CorrelationWindow, warmupSamples),
+		throughputCorrelation: math2.NewCorrelationWindow(config.CorrelationWindow, warmupSamples),
+		medianFilter:          math2.NewMedianFilter(smoothedSamples),
+		smoothedShortRTT:      math2.NewEwma(smoothedSamples, warmupSamples),
+	}
+}
+
+func NewPrioritizedLimiter(config *Config, prioritizer Prioritizer, logger log.Logger) PriorityLimiter {
```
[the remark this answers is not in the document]
jhalterman: Good suggestion.

pkg/util/sync/semaphore.go
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) HashiCorp, Inc.
```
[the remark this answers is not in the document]
jhalterman: I left the SPDX tag on there from where the file came from. This was based on a dynamic semaphore from Consul, originally here: https://github.com/hashicorp/consul/blob/b64674623e22e0873e88811830ad9c218f146d98/lib/semaphore/semaphore.go

pkg/util/adaptivelimiter/adaptivelimiter.go
```diff
@@ -0,0 +1,400 @@
...
+// longer-term execution times. Limit increases are additionally controlled to ensure they don't increase execution times. Any
+// executions in excess of the limit will be rejected with ErrExceeded.
+//
+// By default, an AdaptiveLimiter will converge on a concurrency limit that represents the capacity of the machine it's
+// running on, and avoids having executions block. Since enforcing a limit without allowing for blocking is too strict in
+// some cases and may cause unexpected rejections, optional blocking of executions when the limiter is full can be
+// enabled via WithBlocking, and blocking with prioritized rejection can be enabled via WithPrioritizedRejection.
+//
+// This type is concurrency safe.
+type AdaptiveLimiter interface {
```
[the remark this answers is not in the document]
jhalterman: Yea since we're just using PriorityLimiter in Mimir, AdaptiveLimiter isn't needed. I'll remove it from here, and re-add if/when this makes it upstream.

pkg/util/adaptivelimiter/adaptivelimiter.go
```diff
@@ -0,0 +1,400 @@
...
+func (td *tDigestSample) Reset() {
+	td.TDigest.Reset()
+	td.MinRTT = 0
+	td.MaxInflight = 0
+	td.Size = 0
+}
+
+type adaptiveLimiter struct {
+	logger log.Logger
+	*Config
```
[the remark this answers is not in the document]
jhalterman: Made non-embedded

pkg/ingester/ingester.go
```diff
@@ -1066,6 +1098,11 @@ func (i *Ingester) startPushRequest(ctx context.Context, reqSize int64) (context
 	}
 	ctx = context.WithValue(ctx, pushReqCtxKey, st)
 
+	if i.adaptiveLimiter != nil && i.adaptiveLimiter.push != nil && !i.adaptiveLimiter.push.CanAcquirePermit() {
```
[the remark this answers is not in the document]
jhalterman: I agree, it is too many limiters. So yea, the intent is to replace circuit breakers once these have been in use for a while and look good. The utilization limiters could possibly stay though.

pkg/util/math/median.go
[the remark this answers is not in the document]
jhalterman: Yea, this is ~once a second, and the median window size is hardcoded at 5.

pkg/ingester/ingester.go
```diff
@@ -1685,7 +1727,24 @@ func (i *Ingester) StartReadRequest(ctx context.Context) (resultCtx context.Cont
 	}), nil
 }
 
-func (i *Ingester) PrepareReadRequest(_ context.Context) (finishFn func(error), err error) {
+func (i *Ingester) PrepareReadRequest(ctx context.Context) (finishFn func(error), err error) {
+	defer func() { err = i.mapErrorToErrorWithStatus(err) }()
+
+	if i.adaptiveLimiter != nil && i.adaptiveLimiter.read != nil {
+		// Acquire a permit, blocking if needed
```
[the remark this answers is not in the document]
jhalterman: Yes blocking will be pretty rare, only when the limiter is full, up to the configured max amounts. But it's not really prevented by the earlier `CanAcquirePermit` check. That check just ensures the incoming request's priority is not below the current rejection threshold. Whether we end up blocking or not happens here, based on whether the limiter is full.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
   # and its timeouts aren't reported as errors.
   # CLI flag: -ingester.read-circuit-breaker.request-timeout
   [request_timeout: <duration> | default = 30s]
+
+rejection_prioritizer:
+  # (experimental) Enable adaptive limiting when making requests to ingesters
+  # CLI flag: -ingester.rejection-prioritizer.enabled
+  [enabled: <boolean> | default = false]
+
+  # (experimental) The interval at which the rejection threshold should be
```
tacole02, line 1359:
```suggestion
  # (experimental) The interval at which the rejection threshold is
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # calibrated
+  # CLI flag: -ingester.rejection-prioritizer.calibration-interval
+  [calibration_interval: <duration> | default = 1s]
+
+push_adaptive_limiter:
+  # (experimental) Enable adaptive limiting when making requests to ingesters
+  # CLI flag: -ingester.push-adaptive-limiter.enabled
+  [enabled: <boolean> | default = false]
+
+  # (experimental) Min duration of the window that is used to determine recent,
```
tacole02, line 1369:
```suggestion
  # (experimental) Minimum duration of the window that is used to determine the recent,
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # short-term load on the system
+  # CLI flag: -ingester.push-adaptive-limiter.short-window-min-duration
+  [short_window_min_duration: <duration> | default = 1s]
+
+  # (experimental) Max duration of the window that is used to determine recent,
```
tacole02, line 1374:
```suggestion
  # (experimental) Maximum duration of the window that is used to determine the recent,
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # short-term load on the system
+  # CLI flag: -ingester.push-adaptive-limiter.short-window-max-duration
+  [short_window_max_duration: <duration> | default = 30s]
+
+  # (experimental) Min number of samples that must be recorded in the window
```
tacole02, line 1379:
```suggestion
  # (experimental) Minimum number of samples that must be recorded in the window
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.push-adaptive-limiter.short-window-min-samples
+  [short_window_min_samples: <int> | default = 1]
+
+  # (experimental) Short-term window measurements that will be stored in an
```
tacole02, line 1383:
```suggestion
  # (experimental) Short-term window measurements that are stored in an
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # baseline inflight time
+  # CLI flag: -ingester.push-adaptive-limiter.long-window
+  [long_window: <int> | default = 60]
+
+  # (experimental) The quantile of recorded response times to consider when
+  # adjusting the concurrency limit
+  # CLI flag: -ingester.push-adaptive-limiter.sample-quantile
+  [sample_quantile: <float> | default = 0.9]
+
+  # (experimental) Min inflight requests limit
```
tacole02, line 1394:
```suggestion
  # (experimental) Minimum inflight requests limit
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.push-adaptive-limiter.min-inflight-limit
+  [min_inflight_limit: <int> | default = 2]
+
+  # (experimental) Max inflight requests limit
```
tacole02, line 1398:
```suggestion
  # (experimental) Maximum inflight requests limit
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.push-adaptive-limiter.max-inflight-limit
+  [max_inflight_limit: <int> | default = 200]
+
+  # (experimental) Initial inflight requests limit
+  # CLI flag: -ingester.push-adaptive-limiter.initial-inflight-limit
+  [initial_inflight_limit: <int> | default = 20]
+
+  # (experimental) The max limit as a multiple of current inflight requests
```
tacole02, line 1406:
```suggestion
  # (experimental) The maximum limit as a multiple of current inflight requests
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  [max_limit_factor: <float> | default = 5]
+
+  # (experimental) How many recent limit and inflight time measurements are
+  # stored to detect whether increases in limits correlate with increases in
+  # inflight times
+  # CLI flag: -ingester.push-adaptive-limiter.correlation-window
+  [correlation_window: <int> | default = 50]
+
+  # (experimental) The number of allowed queued requests, as a multiple of
+  # current inflight requests, after which rejections will start
```
tacole02, line 1417:
```suggestion
  # current inflight requests, after which rejections start
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.push-adaptive-limiter.initial-rejection-factor
+  [initial_rejection_factor: <float> | default = 2]
+
+  # (experimental) The number of allowed queued requests, as a multiple of
+  # current inflight requests, after which all requests will be rejected
```
tacole02, line 1422:
```suggestion
  # current inflight requests, after which all requests are rejected
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.push-adaptive-limiter.max-rejection-factor
+  [max_rejection_factor: <float> | default = 3]
+
+read_adaptive_limiter:
+  # (experimental) Enable adaptive limiting when making requests to ingesters
+  # CLI flag: -ingester.read-adaptive-limiter.enabled
+  [enabled: <boolean> | default = false]
+
+  # (experimental) Min duration of the window that is used to determine recent,
```
tacole02, line 1431:
```suggestion
  # (experimental) Minimum duration of the window that is used to determine recent,
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # short-term load on the system
+  # CLI flag: -ingester.read-adaptive-limiter.short-window-min-duration
+  [short_window_min_duration: <duration> | default = 1s]
+
+  # (experimental) Max duration of the window that is used to determine recent,
```
tacole02, line 1436:
```suggestion
  # (experimental) Maximum duration of the window that is used to determine recent,
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # short-term load on the system
+  # CLI flag: -ingester.read-adaptive-limiter.short-window-max-duration
+  [short_window_max_duration: <duration> | default = 30s]
+
+  # (experimental) Min number of samples that must be recorded in the window
```
tacole02, line 1441:
```suggestion
  # (experimental) Minimum number of samples that must be recorded in the window
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.read-adaptive-limiter.short-window-min-samples
+  [short_window_min_samples: <int> | default = 1]
+
+  # (experimental) Short-term window measurements that will be stored in an
```
tacole02, line 1445:
```suggestion
  # (experimental) Short-term window measurements that are stored in an
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # baseline inflight time
+  # CLI flag: -ingester.read-adaptive-limiter.long-window
+  [long_window: <int> | default = 60]
+
+  # (experimental) The quantile of recorded response times to consider when
+  # adjusting the concurrency limit
+  # CLI flag: -ingester.read-adaptive-limiter.sample-quantile
+  [sample_quantile: <float> | default = 0.9]
+
+  # (experimental) Min inflight requests limit
```
tacole02, line 1456:
```suggestion
  # (experimental) Minimum inflight requests limit
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.read-adaptive-limiter.min-inflight-limit
+  [min_inflight_limit: <int> | default = 2]
+
+  # (experimental) Max inflight requests limit
```
tacole02, line 1460:
```suggestion
  # (experimental) Maximum inflight requests limit
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.read-adaptive-limiter.max-inflight-limit
+  [max_inflight_limit: <int> | default = 200]
+
+  # (experimental) Initial inflight requests limit
+  # CLI flag: -ingester.read-adaptive-limiter.initial-inflight-limit
+  [initial_inflight_limit: <int> | default = 20]
+
+  # (experimental) The max limit as a multiple of current inflight requests
```
tacole02, line 1468:
```suggestion
  # (experimental) The maximum limit as a multiple of current inflight requests
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  [max_limit_factor: <float> | default = 5]
+
+  # (experimental) How many recent limit and inflight time measurements are
+  # stored to detect whether increases in limits correlate with increases in
+  # inflight times
+  # CLI flag: -ingester.read-adaptive-limiter.correlation-window
+  [correlation_window: <int> | default = 50]
+
+  # (experimental) The number of allowed queued requests, as a multiple of
+  # current inflight requests, after which rejections will start
```
tacole02, line 1479:
```suggestion
  # current inflight requests, after which rejections start
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1350,6 +1350,140 @@ read_circuit_breaker:
...
+  # CLI flag: -ingester.read-adaptive-limiter.initial-rejection-factor
+  [initial_rejection_factor: <float> | default = 2]
+
+  # (experimental) The number of allowed queued requests, as a multiple of
+  # current inflight requests, after which all requests will be rejected
```
tacole02, line 1484:
```suggestion
  # current inflight requests, after which all requests are rejected
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2163,6 +2163,21 @@ How to **fix** it:
 - Check the write requests latency through the `Mimir / Writes` dashboard and come back to investigate the root cause of high latency (the higher the latency, the higher the number of in-flight write requests).
 - Consider scaling out the ingesters.
 
+### err-mimir-ingester-max-inflight-read-requests
+
+This error occurs when an ingester rejects a read request because the maximum in-flight requests limit has been reached.
+
+How it **works**:
+
+- The ingester has a per-instance limit on the number of in-flight read requests.
```
tacole02, line 2172: Here, you hyphenate `in-flight`, whereas above you have it as one word, i.e. `inflight`. We should standardize this spelling everywhere. My preference is `in-flight`.

tacole02 (approved):
Thank you so much for updating this! I left a few suggestions.

docs/sources/mimir/manage/mimir-runbooks/_index.md
jhalterman: Good point. There's precedent for using `inflight` in some other settings and metrics, like `max_inflight_push_requests`, so perhaps we should just go with that?

pkg/util/math/windows.go
```diff
@@ -0,0 +1,132 @@
...
+
+// CalculateCV calculates the coefficient of variation (relative variance), mean, and variance for the sum. Returns NaN
+// values if there are < 2 samples, the variance is < 0, or the mean is 0.
+func (r *RollingSum) CalculateCV() (cv, mean, variance float64) {
+	if r.size < 2 {
+		return math.NaN(), math.NaN(), math.NaN()
+	}
+
+	mean = r.sumY / float64(r.size)
+	variance = (r.sumSquares / float64(r.size)) - (mean * mean)
```
duricanikolic, line 64: Where does this formula come from?
I thought the variance was calculated like
$\frac{\sum_0^n{(a_i-\overline{a})^2}}{n}$, where $\overline{a} = \frac{\sum_0^n{a_i}}{n}$

pkg/util/adaptivelimiter/prioritizer.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func computeRejectionRate(queueSize, rejectionThreshold, maxQueueSize int) float64 {
+	if queueSize <= rejectionThreshold {
+		return 0
+	}
+	if queueSize >= maxQueueSize {
+		return 1
+	}
+	return float64(queueSize-rejectionThreshold) / float64(maxQueueSize-rejectionThreshold)
```
duricanikolic, line 118: And what about this formula? Is it a known fact or your preference?

pkg/util/adaptivelimiter/adaptivelimiter.go
```diff
@@ -0,0 +1,359 @@
...
+	f.UintVar(&cfg.MaxInflightLimit, prefix+"max-inflight-limit", 200, "Maximum inflight requests limit")
+	f.UintVar(&cfg.InitialInflightLimit, prefix+"initial-inflight-limit", 20, "Initial inflight requests limit")
+	f.Float64Var(&cfg.MaxLimitFactor, prefix+"max-limit-factor", 5, "The maximum limit as a multiple of current inflight requests")
+	f.UintVar(&cfg.CorrelationWindow, prefix+"correlation-window", 50, "How many recent limit and inflight time measurements are stored to detect whether increases in limits correlate with increases in inflight times")
+	f.Float64Var(&cfg.InitialRejectionFactor, prefix+"initial-rejection-factor", 2, "The number of allowed queued requests, as a multiple of current inflight requests, after which rejections start")
+	f.Float64Var(&cfg.MaxRejectionFactor, prefix+"max-rejection-factor", 3, "The number of allowed queued requests, as a multiple of current inflight requests, after which all requests are rejected")
+}
+
+func newLimiter(config *Config, logger log.Logger) *adaptiveLimiter {
+	config.alphaFunc = math2.Log10Func(3)
```
duricanikolic, line 96: [nit]
What if we replace `math2` with `mimirmath` or `mimirutil` or something similar?

pkg/util/math/windows.go
jhalterman: Yes, the implementation uses a variation of that formula which should be equivalent, and is meant to be easier to use when working with a rolling window, basically: (Σx²/n) - x̄ ²

pkg/util/adaptivelimiter/prioritizer.go
jhalterman: I'm not sure if there's a known formula for this one. The intent is that the rejection rate ranges between the rejectionThreshold and maxQueueSize. When the queueSize hits the rejectionThreshold, then rejections begin, and increase up to the maxQueueSize. I'll add a test for this as well.

pkg/util/adaptivelimiter/adaptivelimiter.go
jhalterman: Good suggestion - done

pkg/util/adaptivelimiter/prioritylimiter.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (l *priorityLimiter) AcquirePermit(ctx context.Context, priority Priority) (Permit, error) {
+	// Generate a granular priority for the request and compare it to the prioritizer threshold
+	granularPriority := randomGranularPriority(priority)
+	if granularPriority < l.prioritizer.RejectionThreshold() {
+		return nil, ErrExceeded
+	}
+
+	// Maintain queue stats
+	l.inCount.Add(1)
```
duricanikolic, line 79: When I read the explanations of `inCount` and `outCount` above 
```
	inCount  atomic.Uint32 // Requests received in current calibration period
	outCount atomic.Uint32 // Requests permitted in current calibration period
```
I got an impression that `inCount` will be increased independently of the `AcquirePermit()` outcome, while `outCount` will be increased only if `AcquirePermit()` will be successful. 
This implementation doesn't respect that
```
	// Maintain queue stats
	l.inCount.Add(1)
	defer l.outCount.Add(a)
```
Or maybe I don't see the usecase when `inCount` gets increased, and `outCount` not.

jhalterman: I was previously using the difference between inCount and outCount to compute the rejection rate each time the prioritizer calibrates, but now I'm just using the current queueSize which is simple enough and works fine. So inCount and outCount actually aren't needed anymore - removing.

pkg/ingester/ingester.go
```diff
@@ -3906,6 +3968,18 @@ func (i *Ingester) Push(ctx context.Context, req *mimirpb.WriteRequest) (*mimirp
 	return &mimirpb.WriteResponse{}, err
 }
 
+func (i *Ingester) mapErrorToErrorWithStatus(err error) error {
```
duricanikolic, line 3971: I think that this method is not needed. A year ago we spent several months to re-organize error handling and to get rid of redundant things and errors, so I would prefer not to introduce redundancy where it is not needed. One of the main results of that code re-organization was that all the errors produced by distributors and ingesters are instances of `distributor.Error` and `ingester.ingesterError` respectively. These errors are mapped to the corresponding HTTP or gRPC error at the topmost level (for example, in `ingester.Push()` we wrap `ingester.ingesterError` to a gRPC error that can be recognized by `grpc.Status` and `gogo.Status`). We can discuss this further if needed.

My recommendation would be the following:

In `mimir.RPCCallProcessing()` I would use the same approach that was adopted in `mimir.RPCCallStarting()`, even because there is no need that these 2 methods, implementing the same interface, handle errors differently. `mimir.RPCCallStarting()` gets an error of type `ingester.ingesterError` or `distributor.Error`, depending on whether it came from `ingeter.StartPushRequest()` or `distributor.StartPushRequest`. Both `ingester.StartPushRequest()` and `distributor.StartPushRequest()` currently return only errors with cause `mimirpb.INSTANCE_LIMIT`. We simply create an gRPC error out of it, by calling `status.Error()`. The code assigned to these error is `codes.Unavailable`. You have modified `ingester.StartPushRequest()` by allowing it to return an error with cause `mimirpb.REQUEST_RATE_LIMITED`, which is also mapped to `codes.Unavailable`. But if at a point we decide to map different possible error causes to different gRPC codes, that could easily be done in `mimir.RPCCallStarting()`.
You have recently introduced `mimir.RPCCallProcessing()` that calls `ingester.PreparePushRequest()` or `ingester.PrepareReadRequest`, and the former just propagates the error returned by the latter, while the latter does the error mapping by calling the newly introduced `ingester.mapErrorToErrorWithStatus()`. This method (`ingester.mapErrorToErrorWithStatus()`) checks if the error passed as parameter corresponds to all possible error causes, and according to the result mapps it to a different gRPC code. But this is actually not needed: both `ingester.PreparePushRequest()` and `ingester.PrepareReadRequest()` just check the reactive limiter, and the only error that it can return is of type `ingester.reactiveLimiterExceededError`, which is an `ingester.ingesterError` with error cause `mimirpb.REQUEST_RATE_LIMITED`. I would recommend that you get rid of `ingester.mapErrorToErrorWithStatus()`, and let `mimir.RPCCallProcessing()` just use the `ingester.ingesterError` obtained from `ingester.PreparePushRequest()` and `ingester.PrepareReadRequest()`.

```
func (g *grpcInflightMethodLimiter) RPCCallProcessing(ctx context.Context, _ string) (func(error), error) {
	if rpcCall, ok := ctx.Value(rpcCallCtxKey).(int); ok {
		switch rpcCall {
		case rpcCallIngesterPush:
			if ing := g.getIngester(); ing != nil {
				fn, err := ing.PreparePushRequest(ctx)
				if err != nil {
					return nil, status.Error(codes.Unavailable, err)
				}
				return fn, nil
			}
		case rpcCallIngesterRead:
			if ing := g.getIngester(); ing != nil {
				fn, err := ing.PrepareReadRequest(ctx)
				if err != nil {
					return nil, status.Error(codes.Unavailable, err)
				}
				return fn, nil
			}
		}
	}
	return nil, nil
}
```

jhalterman: Thanks Yuri - this is a good suggestion. I just pushed a commit with this.

pkg/util/reactivelimiter/reactivelimiter.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, err
+	}
+	return &recordingPermit{
+		limiter:         l,
+		startTime:       time.Now(),
+		currentInflight: l.semaphore.Used(),
+	}, nil
+}
+
+func (l *reactiveLimiter) TryAcquirePermit() (Permit, bool) {
```
duricanikolic, line 187: Is this actually needed? It is used in tests only

pkg/util/sync/semaphore.go
```diff
@@ -0,0 +1,141 @@
...
+		return err
+
+	case <-ready:
+		return nil
+	}
+}
+
+// TryAcquire attempts to acquire one "slot" in the semaphore without blocking.
+// Returns true if successful, else false if no slots are immediately available.
+func (s *DynamicSemaphore) TryAcquire() bool {
```
duricanikolic, line 89: I am a bit concerned about this method.
It is true that it is basically not needed (it is used only in `reactivelimiter.TryAcquire()`, which is not used, but if we assume it is going to be used, once it is called, `s.used` gets increased. And it basically means that we occupy a slot in a semaphore, that will be released only if we call `semaphore.Release()`.
I would expect `semaphore.Release()` to just check if `s.used + 1 < s.size`, but not to do `s.used++`.

pkg/util/reactivelimiter/reactivelimiter.go
```diff
@@ -0,0 +1,359 @@
...
+	minLimit, maxLimit float64
+
+	// Mutable state
+	semaphore *mimirsync.DynamicSemaphore
+	mu        sync.Mutex
+
+	// Guarded by mu
+	limit            float64        // The current concurrency limit
+	shortRTT         *tDigestSample // Short term execution times
+	medianFilter     *mimirmath.MedianFilter
```
duricanikolic, line 163: Is this actually needed or it is just a left-over from some previous versions?
I see that it is used only in a test.

pkg/util/math/log.go
```diff
@@ -0,0 +1,26 @@
...
+
+package math
+
+import (
+	"math"
+)
+
+var log10Values []int
+
+func init() {
```
duricanikolic, line 11: Do we really need to calculate 1000 values? The values we store in the array are only 1 and 2.

```
func init() {
  for i := 0; i < 100; i++ {
    log10Values = append(log10Values, 1)
  }
  for i := 100; i < 1000; i++ {
    log10Values = append(log10Values, 2)
  }
}

Comment (jhalterman) [reply pkg/util/math/log.go:11]: We don't have to, but I'm trying to avoid unnecessary math.Log10 computations that would yield the same results.

File: CHANGELOG.md [hunk 38]
```diff
@@ -43,6 +43,7 @@
 * [ENHANCEMENT] Compactor: Shuffle users' order in `BlocksCleaner`. Prevents bucket indexes from going an extended period without cleanup during compactor restarts. #10513
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
 * [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
+* [ENHANCEMENT] Ingester: Add adaptive concurrency limiters to protect push and read operations from overload. #10574
```
duricanikolic, line 46: Don't we call them reactive limiters?
Could you also please explain what the overload is (in this PR we are protecting the system from hitting push and read inflight requests)?

jhalterman: They were originally called adaptive. I had to rename to reactive and missed this. Fixed.

pkg/util/reactivelimiter/reactivelimiter.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func computeChange(queueSize, alpha, beta int, overloaded bool, throughputCorr, throughputCV, rttCorr float64) (change limitChange, reason string) {
+	if queueSize > beta {
+		// This condition handles severe overload where recent RTT significantly exceeds the baseline
+		return decrease, "queue"
+	} else if overloaded && throughputCorr < 0 {
+		// This condition prevents runaway limit increases during moderate overload where inflight is increasing but throughput is decreasing
+		return decrease, "thrptCorr"
+	} else if overloaded && throughputCorr < .3 && rttCorr > .5 {
```
duricanikolic, line 298: I really tried hard to understand this (`if overloaded && throughputCorr < .3 && rttCorr > .5`) and the next condition (`if overloaded && throughputCV < .2 && rttCorr > .5`), but I couldn't.

duricanikolic:
This is a very complicated PR, that requires deep understanding of all concepts introduced by the PR.
The PR description is clear, but too technical. 
I would recommend to write an abstract that says what concretely this PR brings in terms of ingester, and then to leave the rest of the details in a separate section.
For example, in order to understand what this PR actually does, I had to go through the changes done in the ingester, and understand how different technical notions introduced by this PR are connected with already known ingester notions.
For example, I would say (this is just an inaccurate example):
```
This PR introduces adaptive limiters and apply them to ingester infligh push and read requests. Differently from traditional limiters, that reject additional requests as soon as the configured limit is reached, these limiters allow a certain number of requests to be queued before they are actually rejected. The queued requests are rejected only when their size reaches a certain limit. This limit is not fixed, and is calculated by a special algorithm introduced by adaptive limiters, which correlates the current number of inflight requests with allowed queue size. Moreover, adaptive limiters dynamically change their limit, i.e., the number of allowed inflight push and write requests. New limit values depends on whether the number of queued requests grows slower or faster wrt. the requests that are successfully processed...
```

pkg/util/reactivelimiter/reactivelimiter_test.go
```diff
@@ -0,0 +1,138 @@
...
+		_, _ = limiter.AcquirePermit(context.Background())
+	}()
+	assert.Eventually(t, func() bool {
+		return limiter.Blocked() == 1
+	}, 100*time.Millisecond, 10*time.Millisecond)
+	permit.Record()
+	require.Equal(t, 0, limiter.Blocked())
+}
+
+func TestReactiveLimiter_record(t *testing.T) {
```
duricanikolic, line 74: Would it be possible to extend this test with the usecases when the new queue size is between `alpha` and `beta`, but when the underlying semaphore is overloaded. In order to test the 2 special cases from `computeChange()` that I commented above?

duricanikolic:
WDYT if before merging this PR we let it run for some time in a dev cell, set the inflight request limits to a low value and see how it would behave? I would also like to see what the metrics I recommended you to add would show.

pkg/util/reactivelimiter/reactivelimiter.go
jhalterman: Sure, I can explain. In a healthy system, if inflight requests are increasing, throughput should increase as well. If throughput peaks or starts to decrease relative to inflight requests, then it's an indication of overload. So that's what these 3 conditions are checking for.

The first condition, `if overloaded && throughputCorr < 0` checks if the correlation between inflight requests and throughput is negative, which would indicate severe overload (throughput is declining while inflight requests are increasing).

The second condition, `overloaded && throughputCorr < .3 && rttCorr > .5`, is a bit more generous than the first condition. It's looking where throughput [correlation coefficient](https://en.wikipedia.org/wiki/Correlation_coefficient) is weak, which again indicates overload, though not as strongly as a negative throughputCorr. So we use an additional check of RTT correlation which should be moderate when overloaded (increasing inflight requests correlate with increasing RTTs).

The third condition, `overloaded && throughputCV < .2 && rttCorr > .5`, is looking at the throughput's [coefficient of variation](https://en.wikipedia.org/wiki/Coefficient_of_variation) as an indication of overload. When inflight requests are at the limit (`overloaded`) and the CV is low, this also indicates we've hit peak throughput, and again, we control by checking for a moderate rttCorr.

jhalterman:
> I would recommend to write an abstract that says what concretely this PR brings in terms of ingester, and then to leave the rest of the details in a separate section.

I added a bit of an abstract to the PR description as suggested and left the rest as details and implementation notes.

jhalterman:
> I would recommend to write an abstract that says what concretely this PR brings in terms of ingester, and then to leave the rest of the details in a separate section.

I added a bit of an abstract to the PR description as suggested and left the rest as details and implementation notes.

pkg/util/reactivelimiter/reactivelimiter_test.go
jhalterman: Yea we can feed in some values which should cause the correlation window and queueSize to hit one of the decrease limit conditions, but there's no good way to verify which decrease limit condition we hit. We could test the `computeChange` function in isolation instead?

pkg/util/sync/semaphore.go
duricanikolic: > I would expect semaphore.Release() to just check if s.used + 1 < s.size, but not to do s.used++.
@jhalterman it was a typo, I was referring to `semaphore.TryAcquire()`.

jhalterman: `TryAcquire` is a real acquisition attempt, so it should increment the used count if possible. It just doesn't block when full like `Acquire` does.

pkg/util/reactivelimiter/reactivelimiter.go
duricanikolic: Only now I realized that the only difference between the 3 `decrease` cases is the reason.

duricanikolic (approved):
Great work.
Thank you

pkg/util/sync/semaphore.go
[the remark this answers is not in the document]
56quarters: > And is tagged similar to other Hashicorp bits Mimir uses, ex:

That's a vendored file, it gets copied in verbatim. Files/Code that we move into the Mimir tree need SPDX tags. You can find examples of the format for lots of stuff that originated in Cortex or Thanos. [Here's an example from Thanos](https://github.com/grafana/mimir/blob/bcd6c8718c28c0b3633c3547ab3d4f366de8446a/pkg/storegateway/bucket.go#L1-L4).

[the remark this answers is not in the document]
jhalterman: I ended up just writing and adding a new DynamicSemaphore impl, which is a little cleaner anyways.

56quarters (commented):
a1ab89f90672ede55556794215d288677a8c376c LGTM, just a few comments on the testing.

pkg/util/sync/semaphore_test.go
```diff
@@ -1,184 +1,48 @@
...
+		s := NewDynamicSemaphore(1)
 		require.NoError(t, s.Acquire(context.Background()))
 
-		// Should timeout while acquiring permit
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
-		defer cancel()
-		require.Error(t, s.Acquire(ctx))
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		require.ErrorIs(t, s.Acquire(ctx), context.Canceled)
```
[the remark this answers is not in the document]
jhalterman: For this one, used is still 1 since the permit was not released, it's just that we canceled an attempt to acquire a new permit, so Waiters is 0. Adding assertions for these.

pkg/util/sync/semaphore_test.go
```diff
@@ -208,15 +72,56 @@ func TestDynamicSemaphore_TryAcquire(t *testing.T) {
...
+		}()
+		go func() {
+			_ = s.Acquire(context.Background())
+		}()
+
+		waitForWaiters(t, s, 2)
+
+		// Increase size which should release waiters
+		s.SetSize(3)
+		assert.Equal(t, 0, s.Waiters())
```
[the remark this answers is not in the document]
jhalterman: Good suggestion. Added.

pkg/util/sync/semaphore.go
```diff
@@ -96,16 +94,20 @@ func (s *DynamicSemaphore) SetSize(size int) {
...
-			close(next.Value.(chan struct{}))
+			s.wakeAndAcquire(s.waiters.Front())
 		}
 	}
 
 	s.size = size
 }
 
+// Wake a blocked waiter and acquires a permit.
+func (s *DynamicSemaphore) wakeAndAcquire(waiterElem *list.Element) {
```
56quarters, line 105: nit: Maybe note that this requires the mutex to be held?

56quarters (approved):
LGTM, thanks!

PR #10580, opened 2025-02-04
Build both Mimir docs sources in deploy preview

jdbaldry:

#### What this PR does

This change to the workflow builds both the Mimir docs and the Mimir Helm chart docs in a single deploy preview and posts the links to both back on the PR.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


robbymilo (approved):
Sweet!

tacole02 (approved):

jdbaldry:
Thanks for the reviews, a mimir maintainer can merge this whenever they want the feature

.github/workflows/deploy-pr-preview.yml
```diff
@@ -11,15 +11,28 @@ on:
...
-      index_file: true
+      sha: ${{ github.event.pull_request.head.sha }}
+      sources: |
+        [
+          {
+            "index_file": null,
+            "relative_prefix": "/docs/mimir/v2.15.x/",
+            "repo": "mimir",
+            "source_directory": "docs/sources/mimir",
+            "website_directory": "content/docs/mimir/v2.15.x"
```
narqo, line 28: Does Mimir release shepherds need to keep these values in-sync with current active releases? If yes, can we add a note about that in the [RELEASE.md](https://github.com/grafana/mimir/blob/13ff857dcc011f2b4c65ae3438ddd85600c8e206/RELEASE.md#how-to-cut-an-individual-release).

jdbaldry: Yeah, I think so because I can't add steps in to the workflow because its using a reusable workflow rather than calling an action.

I went to update the RELEASE file but I wasn't sure where to put things.

I don't see anything about updating the variables in https://github.com/grafana/mimir/blob/jdb%2F2025-02-build-both-mimir-docs/docs/sources/helm-charts/mimir-distributed/_index.md?plain=1#L10-L12 which is an associated release task. Do you have any ideas where I should put it?

narqo: Helm chart has a separate release process, and is described in its own [release guide](https://github.com/grafana/mimir/blob/2f822fb83e30f18a86053f4132cf294972fe69f6/operations/helm/charts/mimir-distributed/RELEASE.md) (the reason is that we must release a version Mimir, then release a version of GEM, and only then, a version of the Helm chart, that uses new Mimir/GEM).

Related to the changes in this PR, I think adding a note to Mimir's [RELEASE.md][1] is the right place. Let's one to the check-list in the "Release issue template". Somewhere in the end — before "_Announce the release on socials_" — is where I'd do that.

[1]: https://github.com/grafana/mimir/blob/13ff857dcc011f2b4c65ae3438ddd85600c8e206/RELEASE.md#how-to-cut-an-individual-release

jdbaldry: Sorry, left this hanging and then went on holiday.

My understanding is that the Mimir version in the deploy preview should match the Mimir version in the Helm chart so I think it makes sense to put this step alongside the Mimir and GEM version parameters step in the Helm release: https://github.com/grafana/mimir/blob/2f822fb83e30f18a86053f4132cf294972fe69f6/operations/helm/charts/mimir-distributed/RELEASE.md?plain=1#L107-L109

jdbaldry: Added in [`5a3aeee` (#10580)](https://github.com/grafana/mimir/pull/10580/commits/5a3aeee9a5c1d224234a714198b565aa27b65395)

narqo (approved):
Works for me 🔥

Issue #7664, opened 2024-03-19
Compactor: Reduce memory consumption from large meta.json files


aknuds1:
#### Is your feature request related to a problem? Please describe.

Mimir's block meta.json files can get really large as the `sources` list grows due to intermediate blocks resulting from split groups and compactor merge shards, recompactions due to OOO blocks. One large (748 KB) meta.json file encountered had 22947(!) source ULIDs. These large block metas can again cause the Mimir compactor to use large amounts of memory.

#### Describe the solution you'd like

We should attempt to reduce the amount of memory required for holding block metadata. The compactor needs the `compaction.sources` list of each meta object to figure out which blocks have already been compacted. Possible solutions:

* Keep only a merged unique list of block sources (dropping the `sources` field from each decoded `Meta`?)
* Cache compactor job plans in memcached (cache key being a hash of sorted block IDs, the plan should always be the same for a certain input)

dmwilson-grafana:
Addressed by https://github.com/grafana/mimir/pull/10585.

PR #10585, opened 2025-02-04
Add Compactor.MaxLookback Option for Limiting Blocks Loaded in Compaction Cycles

dmwilson-grafana:
#### What this PR does

As described in https://github.com/grafana/mimir/issues/7664, a Mimir block's `meta.json` file can become large if its sources list grows due to intermediate blocks from split groups, compactor merge shards, or re-compactions due to OOO blocks. When a bucket contains many blocks with many sources each, this may cause unexpectedly high resource usage on the compactor.

This PR adds an experimental option, `-compactor.max-lookback` which can reduce the CPU time and memory spent reading blocks' metadata files by instructing the compactor to read metadata for only the blocks that have been uploaded within the `max-lookback` period.  If set, this value should be set well above the maximum of `-compactor.block-ranges` (e.g. `168h`) to not interfere w. the regular compaction process. 

This PR also adds a metric, `cortex_compactor_meta_blocks_synced` which describes the number of blocks discovered in object storage and labels them with their `state`, i.e. were they loaded,  marked for deletion,  skipped because they were outside the lookback period, etc.

- https://github.com/grafana/mimir/issues/7664

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


seizethedave (commented):
I'm thinking about this new config setting, ops-wise. The primary concern we've talked about with this new setting is around block upload and/or backfills: If we set `max-lookback` (a server-level option) to X days, then any tenant-level block upload with a block ID over X days old will go unnoticed. (My cursory scan of the block upload code says that it will honor whatever block ID is [given on import](https://github.com/grafana/mimir/blob/b54ab806621ad9f5c39861198539a7e9e33710e5/pkg/compactor/block_upload.go#L188), rather than creating new `$now` ULIDs.)

Two specific bad things I'm thinking about are:
1. footguns around forgetting to adjust `max-lookback` in preparation of uploads
2. operating a large multitenant Mimir cluster and one tenant needs to upload data from 1 year ago. This will cause us to set `max-lookback` to 1y for _all tenants_ and probably blow up compactor resources.

What do you think of a slightly different impl: using `IterWithAttributes` to learn each block's `mtime` and instead using that as a filtering device. Then (at least, from my 30' view) a tenant can perform a block upload and the server doesn't have to be specially reconfigured to match the data.

seizethedave:
RE my 2nd point:
 
> operating a large multitenant Mimir cluster and one tenant needs to upload data from 1 year ago. This will cause us to set max-lookback to 1y for all tenants and probably blow up compactor resources.

If the attributes thing doesn't work out, I think we should consider doing something like one of these:
1. disable max-lookback for a tenant if they have [this](https://github.com/grafana/mimir/blob/741644ef016b91a8786e9db7255ada4476c3692a/pkg/util/validation/limits.go#L945) option enabled. (Then just the minority of tenants with uploading enabled would contribute to waste in meta loading.)
2. or expose another setting for a per-tenant max-lookback override, which would probably have to be changed alongside changes to `compactor.block_upload_enabled`.

dmwilson-grafana:
> If we set max-lookback (a server-level option) to X days, then any tenant-level block upload with a block ID over X days old will go unnoticed. 

Agreed. It doesn't make sense to have this as a server-level option. I did some research and seems it could be configured per-tenant if the option was defined in [limits.go](https://github.com/grafana/mimir/blob/main/pkg/util/validation/limits.go) with an override (ref: [docs](https://grafana.com/docs/mimir/latest/configure/configuration-parameters/#limits)). As you mentioned in the second comment, probably need some additional work on limits.

dmwilson-grafana:
> What do you think of a slightly different impl: using IterWithAttributes to learn each block's mtime and instead using that as a filtering device. 

I believe both `Iter` and `IterWithObjectAttributes` are non-recursive ([link](https://github.com/grafana/mimir/blob/6e1e4bd68d28400f76969905955dfe25484257d2/pkg/storage/bucket/prefixed_bucket_client.go#L56-L75)), which would mean an extra call to check `mtime` for each blocks meta file at `/$tenant/$block/meta.json`. Also a bit hesitant to use mtime since not all `IterOptions` are supported by all object storage providers.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
 * [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
 * [ENHANCEMENT] Ruler: Adds support for filtering results from rule status endpoint by `file[]`, `rule_group[]` and `rule_name[]`. #10589
+* [ENHANCEMENT] Compactor: Add experimental `-compactor.max-lookback` option to limit blocks considered in each compaction cycle. Blocks uploaded prior to the lookback period won't be processed. This option may help reduce CPU utilization in tenants with large block metadata files that are processed before each compaction. #10585
```
tacole02, line 47:
```suggestion
* [ENHANCEMENT] Compactor: Add experimental `-compactor.max-lookback` option to limit blocks considered in each compaction cycle. Blocks uploaded prior to the lookback period aren't processed. This option helps reduce CPU utilization in tenants with large block metadata files that are processed before each compaction. #10585
```

docs/sources/mimir/configure/about-versioning.md
```diff
@@ -69,6 +69,8 @@ The following features are currently experimental:
     - `-compactor.no-blocks-file-cleanup-enabled`
   - In-memory cache for parsed meta.json files:
     - `-compactor.in-memory-tenant-meta-cache-size`
+  - Limit blocks processed in each compaction cycle. Blocks uploaded prior to the maximum lookback won't be processed.
```
tacole02, line 72:
```suggestion
  - Limit blocks processed in each compaction cycle. Blocks uploaded prior to the maximum lookback aren't processed.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -4816,6 +4816,13 @@ sharding_ring:
 # smallest-range-oldest-blocks-first, newest-blocks-first.
 # CLI flag: -compactor.compaction-jobs-order
 [compaction_jobs_order: <string> | default = "smallest-range-oldest-blocks-first"]
+
+# (experimental) Blocks uploaded before the lookback will not be considered in
```
tacole02, line 4820:
```suggestion
# (experimental) Blocks uploaded before the lookback aren't considered in
```

tacole02: We avoid using future tense in docs.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -4816,6 +4816,13 @@ sharding_ring:
...
+# compactor cycles. If set, this value should be larger than all values in
+# `-blocks-storage.tsdb.block-ranges-period`. The value 0s means all blocks will
+# be considered regardless of their upload time.
```
tacole02, line 4823:
```suggestion
# `-blocks-storage.tsdb.block-ranges-period`. A value of 0s means that all blocks
# are considered regardless of their upload time.
```

tacole02 (approved):

pkg/compactor/syncer_metrics.go
```diff
@@ -83,6 +89,15 @@ func (m *aggregatedSyncerMetrics) gatherThanosSyncerMetrics(reg *prometheus.Regi
 	m.metaSyncFailures.Add(mfm.SumCounters("blocks_meta_sync_failures_total"))
 	m.metaSyncDuration.Add(mfm.SumHistograms("blocks_meta_sync_duration_seconds"))
 
+	for _, met := range mfm["blocks_meta_synced"].GetMetric() {
```
seizethedave, line 92: Is the purpose of this passage to increment the gauge? More evidence that this should be a counter. :)

pkg/storage/tsdb/block/fetcher_test.go
```diff
@@ -385,6 +392,153 @@ func TestMetaFetcher_ShouldNotParseMetaJsonFilesAgain(t *testing.T) {
...
+		f, err := NewMetaFetcher(logger, 10, objstore.WrapWithMetrics(bkt, reg, "test"), fetcherDir, reg, nil, nil, maxLookbackThreshold)
+		require.NoError(t, err)
+
+		actualMetas, _, actualErr := f.Fetch(ctx)
+		require.NoError(t, actualErr)
+		require.Len(t, actualMetas, 3)
+		require.Contains(t, actualMetas, olderULID)
+	})
+
+	t.Run("should return no block metas when fetcher lookback is set short", func(t *testing.T) {
```
seizethedave, line 526: Since ulid uses a funky and specific timestamp format, how about a test case that verifies that we're playing nice with that? Something like three blocks at -3G, -2G, -1G where G is near the timestamp granularity and verify that maxlookback of -1.5G does the right thing.

pkg/storage/tsdb/block/fetcher.go
```diff
@@ -370,6 +391,12 @@ func (f *MetaFetcher) fetchMetadata(ctx context.Context, excludeMarkedForDeletio
 				return nil
 			}
 
+			// skip any blocks older than the fetcher's max lookback.
+			if id.Compare(minAllowedBlockID) == -1 {
```
seizethedave, line 395: This would not work correctly if, for some reason, there was a default-constructed `ulid` among the input. Seems like slightly less chance of surprising behavior if the test is skipped altogether when lookback is disabled. (More thinking of accidentally changing default behavior if this experimental flag isn't enabled.)

```suggestion
			if f.maxLookback > 0 && id.Compare(minAllowedBlockID) == -1 {
```

pkg/storage/tsdb/block/fetcher.go
```diff
@@ -307,7 +317,18 @@ func (f *MetaFetcher) fetchMetadata(ctx context.Context, excludeMarkedForDeletio
 		ch  = make(chan ulid.ULID, f.concurrency)
 		mtx sync.Mutex
 	)
-	level.Debug(f.logger).Log("msg", "fetching meta data", "concurrency", f.concurrency)
+
+	level.Debug(f.logger).Log("msg", "fetching meta data", "concurrency", f.concurrency, "max-lookback", f.maxLookback)
+	var minAllowedBlockID ulid.ULID
+	if f.maxLookback.Milliseconds() > 0 {
+		var err error
+		minAllowedBlockID, err = ulid.New(
```
seizethedave, line 325: The way we are constructing a ULID that represents a point in time - and then comparing that to items in the list in order to filter by age is deserving of a handrail "why" comment.

pkg/compactor/syncer_metrics_test.go
```diff
@@ -133,6 +139,11 @@ func newTestSyncerMetrics(reg prometheus.Registerer) *testSyncerMetrics {
 		Buckets: []float64{0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120, 240, 360, 720},
 	})
 
+	m.metaBlocksSynced = promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
dmwilson-grafana: This struct, `syncerMetrics`, just exposes metrics from `tsdb/block/fetcher.go` ([here](https://github.com/grafana/mimir/blob/bcd6c8718c28c0b3633c3547ab3d4f366de8446a/pkg/storage/tsdb/block/fetcher.go#L40-L46)) where `m.Synced` is a gauge rather than a counter (and we use it like a gauge, i.e. we `Set(n)`, not just `Add(n)`). We do this because we have a special wrapper around a gauge to update values atomically (see: [tx_gauge.go](https://github.com/grafana/mimir/blob/main/pkg/util/extprom/tx_gauge.go#L23)) in the fetcher. I can write this metric out as a counter, but it I thought changing it would make this more confusing.

pkg/storage/tsdb/block/fetcher_test.go
dmwilson-grafana: ulid is precise down to 1ms (see: [docs](https://pkg.go.dev/github.com/oklog/ulid#section-readme)), so it's possible that we will fail to filter out blocks that are less than 1ms older than the threshold (e.g. `168h0m0.0009s`). Small illustrative example below. 

```go
func main() {
	ns := time.Nanosecond
	tdel, olddel, newdel := -1000000 * ns, -1500000 * ns, -500000 * ns
	now := time.Now()
	minAllowed, _ := ulid.New(ulid.Timestamp(now.Add(time.Duration(tdel))), nil)  // threshold: 1ms
	justOlder, _ := ulid.New(ulid.Timestamp(now.Add(time.Duration(olddel))), nil) // 1.5ms old
	justNewer, _ := ulid.New(ulid.Timestamp(now.Add(time.Duration(newdel))), nil) // 0.5ms old
	// can be (-1, 0) OR (0, 1); fetcher skips if `id.Compare(threshold)` == -1 
	fmt.Println(justOlder.Compare(minAllowed), justNewer.Compare(minAllowed))
}
``` 

I can't see an easy way to test this in `fetcher_test.go` without getting very flaky results. I just added a test that checks that we will skip a block that's 1s beyond the threshold.

seizethedave: Sounds good! Thanks.

seizethedave (approved):
LGTM!

PR #10588, opened 2025-02-05
distributor: report OTLP parse errors back to client

narqo:
#### What this PR does

This PR comes with mostly structural changes. The idea here is that, currently, `OTLPHandler` silently drops series, it couldn't parse. This makes it complicated for the users, who see the increase in the dropped OTLP series due to a parse error, but who don't have access to the internal logs (e.g. when Mimir is deployed in Grafana Cloud).

Here, the handler holds the errors received during the OTLP conversion, and, if there wasn't any explicit issues from the downstream push services, it returns the status 400 (_it's up for a debate if this new behaviour is more correct; refer to the internal discussion https://github.com/grafana/mimir-squad/issues/2716_).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"fmt"
 	"io"
 	"net/http"
 	"time"
@@ -251,16 +252,16 @@ func OTLPHandler(
 		}
 		req := newRequest(supplier)
 
-		err := push(ctx, req)
-		if err == nil {
-			// Push was successful, but OTLP converter left out some of the incoming samples. We let the user know about it by returning 4xx (and an insight log).
-			if err := otlpConverter.Err(); err != nil {
-				err = httpgrpc.Error(http.StatusBadRequest, err.Error())
+		pushErr := push(ctx, req)
+		if pushErr == nil {
+			// Push was successful, but OTLP converter left out some samples. We let the client know about it by replying with 4xx (and an insight log).
+			if otlpErr := otlpConverter.Err(); otlpErr != nil {
+				pushErr = httpgrpc.Error(http.StatusBadRequest, otlpErr.Error())
 			}
 		}
-		if err != nil {
-			if errors.Is(err, context.Canceled) {
-				level.Warn(logger).Log("msg", "push request canceled", "err", err)
+		if pushErr != nil {
+			if errors.Is(pushErr, context.Canceled) {
+				level.Warn(logger).Log("msg", "push request canceled", "err", pushErr)
 				writeErrorToHTTPResponseBody(r.Context(), w, statusClientClosedRequest, codes.Canceled, "push request context canceled", logger)
 				return
 			}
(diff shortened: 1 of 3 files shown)


narqo:
@aknuds1 I think you may be the best one for the review here (it is not urgent, though)

pkg/distributor/otel.go
```diff
@@ -71,169 +72,165 @@ func OTLPHandler(
...
-		switch contentEncoding {
-		case "gzip":
-			compression = util.Gzip
-		case "lz4":
-			compression = util.Lz4
-		case "":
-			compression = util.NoCompression
-		default:
-			return httpgrpc.Errorf(http.StatusUnsupportedMediaType, "unsupported compression: %s. Only \"gzip\", \"lz4\", or no compression supported", contentEncoding)
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
narqo, line 75: A note for the reviewer: this chunk looks "weird" in the github's diff. All it does is
1. inlines the `otlpHandler` wrapper function;
2. introduces the `otlpConverter`, that holds the paring errors.

The rest of the logic (in this chunk) is the same as it was before.

aknuds1:
Taking a look!

pkg/distributor/otel.go
```diff
@@ -71,169 +72,165 @@ func OTLPHandler(
...
-			"promoted_resource_attributes", promoteResourceAttributes,
-		)
+			pushMetrics.IncOTLPRequest(tenantID)
+			pushMetrics.ObserveUncompressedBodySize(tenantID, float64(uncompressedBodySize))
 
-		req.Timeseries = metrics
-		req.Metadata = otelMetricsToMetadata(addSuffixes, otlpReq.Metrics())
+			metrics, metricDropped, err := otelMetricsToTimeseries(ctx, otlpConverter, addSuffixes, enableCTZeroIngestion, enableStartTimeQuietZero, promoteResourceAttributes, keepIdentifyingResourceAttributes, otlpReq.Metrics(), spanLogger)
+			if metricDropped > 0 {
+				discardedDueToOtelParseError.WithLabelValues(tenantID, "").Add(float64(metricDropped)) // "group" label is empty here as metrics couldn't be parsed
```
aknuds1, line 201: Plural metrics are dropped:
```suggestion
			metrics, metricsDropped, err := otelMetricsToTimeseries(ctx, otlpConverter, addSuffixes, enableCTZeroIngestion, enableStartTimeQuietZero, promoteResourceAttributes, keepIdentifyingResourceAttributes, otlpReq.Metrics(), spanLogger)
			if metricsDropped > 0 {
				discardedDueToOtelParseError.WithLabelValues(tenantID, "").Add(float64(metricsDropped)) // "group" label is empty here as metrics couldn't be parsed
```

pkg/distributor/otel.go
```diff
@@ -71,169 +72,165 @@ func OTLPHandler(
...
+			for _, m := range metrics {
+				sampleCount += len(m.Samples)
+				histogramCount += len(m.Histograms)
+				exemplarCount += len(m.Exemplars)
 			}
+
+			level.Debug(spanLogger).Log(
+				"msg", "OTLP to Prometheus conversion complete",
+				"metric_count", metricCount,
+				"metric_dropped", metricDropped,
```
aknuds1, line 221:
```suggestion
				"metrics_dropped", metricsDropped,
```

pkg/distributor/otel.go
aknuds1: The diff is understandable in itself, but why do you need to inline `otlpHandler`? I don't like that change, because `OTLPHandler` becomes a mega function. Usually you will go the other way around, and break down such mega functions so they're more readable.

pkg/distributor/otel.go
```diff
@@ -414,33 +419,73 @@ func otelMetricsToMetadata(addSuffixes bool, md pmetric.Metrics) []*mimirpb.Metr
...
+		return nil
+	}
+
+	_, c.err = c.converter.FromMetrics(ctx, md, settings, utillog.SlogFromGoKit(logger))
+
+	mimirTS := c.converter.TimeSeries()
+	if len(mimirTS) == 0 {
+		return nil
+	}
+	return mimirTS
```
aknuds1, line 470: I don't understand what there's to gain from converting empty `mimirTS` to nil. They should be functionally equivalent.
```suggestion
	return mimirTS
```

aknuds1 (commented):
Please see comments. Mainly, I think you should avoid the inlining since I think it's a step in the wrong direction (towards less readable code). Also if I'm not mistaken, the inlining is just a (significant) stylistic change so it shouldn't be done in this PR, which is meant to fix a bug.

pkg/distributor/otel.go
```diff
@@ -203,6 +291,7 @@ func OTLPHandler(
 		level.Debug(spanLogger).Log(
 			"msg", "OTLP to Prometheus conversion complete",
 			"metric_count", metricCount,
+			"metric_dropped", metricsDropped,
```
aknuds1, line 294: Plural metrics are dropped:
```suggestion
			"metrics_dropped", metricsDropped,
```

pkg/distributor/otel.go
```diff
@@ -414,33 +430,70 @@ func otelMetricsToMetadata(addSuffixes bool, md pmetric.Metrics) []*mimirpb.Metr
...
-	return mimirTS, nil
+func (c *otlpMimirConverter) ToTimeseries(ctx context.Context, md pmetric.Metrics, settings otlp.Settings, logger log.Logger) []mimirpb.PreallocTimeseries {
+	if c.err != nil {
+		return nil
+	}
+
+	_, c.err = c.converter.FromMetrics(ctx, md, settings, utillog.SlogFromGoKit(logger))
+
+	mimirTS := c.converter.TimeSeries()
+	return mimirTS
```
aknuds1, line 478: [Nit] Easier to read:
```suggestion
	return c.converter.TimeSeries()
```

aknuds1 (approved):
LGTM modulo nits. Thanks!

PR #10590, opened 2025-02-06
Add dashboard panels for ingester reactive limiters

jhalterman:
#### What this PR does

This PR adds dashboard panels for ingester reactive limiters, which are being introduced via #10574.

The Mimir / Writes dashboard had some panels added to and below the existing Instance Limits row:

Screenshot 2025-02-05 at 5 02 20 PM

And the Mimir / Reads dashboard had a new Instance Limits row added:

Screenshot 2025-02-05 at 5 02 13 PM

These can be enabled via `show_reactive_limiter_panels: true`.

Depends on #10574

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):
LGTM

operations/mimir-mixin/dashboards/reads.libsonnet
```diff
@@ -441,5 +441,42 @@ local filename = 'mimir-reads.json';
...
     .addRows(
       $.getObjectStoreRows('Blocks object store (querier accesses)', 'querier')
+    )
+    .addRowIf(
+      $._config.show_reactive_limiter_panels,
+      $.row('Instance Limits')
+      .addPanel(
+        $.timeseriesPanel('Ingester per %s blocked requests' % $._config.per_instance_label) +
+        $.hiddenLegendQueryPanel(
+          'sum by (%s) (cortex_ingester_reactive_limiter_blocked_requests{%s, request_type="read"})'
```
56quarters, line 451: These metric names don't seem right, are they already rates? We should be emitting counters and computing a rate at query time.

56quarters (changes requested):
These panels seem fine but it doesn't seem like the metrics we're emitting are correct.

operations/mimir-mixin/dashboards/reads.libsonnet
56quarters: In that case, the current names are fine. I was just confused because they seemed like they were counters.

56quarters (approved):

PR #10592, opened 2025-02-06
Add script to flush all Memcached pods in a StatefulSet

charleskorn:
#### What this PR does

This PR adds a script that gradually flushes all Memcached pods in a StatefulSet.

This is useful when a cache has been poisoned and all items need to be purged, but restarting all Memcached pods at once with `kubectl rollout restart` would cause too much cache churn at once. For example, purging all Memcached pods in the query results cache at once would cause too much load on ingesters and store-gateways.

I haven't added a changelog entry given this is primarily a debugging tool.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/flush-memcached.sh
+++ b/tools/flush-memcached.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-only
+
+# This script flushes all cache items in all Memcached pods in a Kubernetes StatefulSet, waiting between each pod.
+
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


jhesketh (approved):
Neat :-).

nit: a default for the delay would be nice, but not worth blocking on here.

charleskorn:
> nit: a default for the delay would be nice, but not worth blocking on here.

Yep, I agree, but until we use this more in some real-world clusters I'm not sure what a sensible default value would be.

PR #10597, opened 2025-02-06
Replace markblocks with rewritten mark-blocks

andyasp:
#### What this PR does

Replaces `markblocks` with a rewritten version uncreatively named `mark-blocks`.

The motivation for this is that I'm in a situation (don't ask) that requires marking and then later unmarking a large number of blocks.

`markblocks` as written was limiting for this because:
-  it only supports adding marks, not removing them
-  it only supports blocks specified as an argument, which doesn't work for large lists
-  marking blocks can fail halfway (on the global marker) and then can't be retried without manual intervention
-  when a failure is encountered on a large number of blocks it would be nice to easily resume in a way that does not replicate work or require modifying the input

The rewrite addresses these points.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


docs/internal/tools/mark-blocks.md
```diff
@@ -0,0 +1,80 @@
+# Mark blocks tool
```
tacole02, line 1:
```suggestion
# Mark-blocks tool
```

docs/internal/tools/mark-blocks.md
```diff
@@ -0,0 +1,80 @@
...
+
+`mark-blocks` is a tool that assists in uploading or removing block markers to a specified backend.
+Marks are uploaded to both the block folder and the global marks folder for the provided tenant.
+
+Two types of marks can be created, depending on the `-mark-type` flag provided:
+
+## `deletion` mark
+
+When `-mark-type deletion` is provided, `mark-blocks` uploads a `DeletionMark` which tells the compactor that the block provided should be deleted.
+This is a **destructive operation** (although it won't happen immediately, the default delation delay is 12h, see `-compactor.deletion-delay` value), proceed with caution.
```
tacole02, line 11:
```suggestion
This is a **destructive operation** (although it doesn't happen immediately, the default delation delay is 12h, see `-compactor.deletion-delay` value), proceed with caution.
```

tools/mark-blocks/README.md
```diff
@@ -0,0 +1,63 @@
+# Mark Blocks
```
tacole02, line 1:
```suggestion
# Mark-blocks
```

tools/mark-blocks/README.md
```diff
@@ -0,0 +1,63 @@
...
+- `--resume-index` (optional) The index of the block to resume from. This index is logged to assist in recovering from partial failures
+- `--concurrency` (optional) How many markers to upload or remove concurrently. Defaults to `16`
+- `--details` (optional) Details to include in an added mark
+- `--dry-run` (optional) Log changes that would be made instead of actually making them
+
+Each supported object storage service also has an additional set of flags (see examples in [Running](##Running)).
+
+## Running
+
+Running `go build .` in this directory builds the program. Then use an example below as a guide.
```
tacole02, line 22:
```suggestion
Run `go build .` in this directory to build the program. Then, use an example below as a guide.
```

tacole02 (approved):

docs/internal/tools/mark-blocks.md
andyasp: Thank you for this and the other suggestions! I committed most of them and percolated them out the changes to other tools READMEs for consistency. These titles seem tricky though. "Mark-blocks" strikes me as off because it's like a mixture of the actual tool name and a title but neither. I changed it to mark-blocks here and then `mark-blocks` in the READMEs. Does that seem okay?

narqo (approved):
🔥 Code-wise, looks good to me.

I'm not sure about removing the old version, though. Would it be a better for the users to leave a note that the old version is deprecated? I'd used the old `markblock` in the past, so I'd be stroked with a mild panic if I've learned the tool's gone right when I needed it next 🥲

andyasp:
> Code-wise, looks good to me.

Appreciate the review, thanks!

> I'm not sure about removing the old version, though.
> Would it be a better for the users to leave a note that the old version is deprecated?

These changes started as me wanting to improve this tool in-place, but by the time I was done I realized it was more of a rewrite. I renamed it to strongly signal "this has changed!" and because I think the dashed name format is more general.

As far as whether we should deprecate rather than remove, I'd much rather remove it because it's a potential footgun. When the old version fails halfway it leaves an invalid state, and rerunning results in seeing an "already marked" message rather than fixing the situation. To me, that's yet another sign that https://github.com/grafana/mimir-squad/issues/2026 should be looked into.

I added a signpost note in case a user does experience confusion. I suppose we can remove that after a release or two.

PR #5738, opened 2023-08-14
added metrics for block upload stats

aldernero:
This PR adds some basic counters around completed block uploads. We can see failed upload information in logs, but currently successful block uploads don't produce any logs unless using the debug log level.

The idea is to be able to track
1. blocks successfully uploaded (count) - `cortex_block_upload_api_blocks_total`
2. total bytes of uploaded blocks - `cortex_block_upload_api_bytes_total`
3. total number of files in uploaded blocks - `cortex_block_upload_api_files_total`

These can be helpful when correlating changes seen in the compactor or object storage.

#### Checklist

- [X] Tests updated
- [N/A] Documentation added
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #10601, opened 2025-02-07
Compactor: Turn block upload metrics from gauges to counters

santihernandezc:
We only ever increase these metrics. They should be counters, not gauges.


bboreham (approved):
Thanks!  The [original PR](https://github.com/grafana/mimir/pull/5738) asserts that they are counters, so must have just been brain-fade.

santihernandezc:
@bboreham yep, the error was introduced when [making these metrics per-tenant](https://github.com/grafana/mimir/pull/5738/commits/d78d8c4d20f7eedfa69619e7198a4e6d22b1793e).

PR #10612, opened 2025-02-10
Query-tee: Allow skipping comparison when preferred backend fails

julienduchesne:
My use case is that my secondary backend allows more requests through than the preferred backend (where they fail)

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


tools/querytee/proxy.go
```diff
@@ -305,7 +307,10 @@ func (p *Proxy) Start() error {
 		if p.cfg.CompareResponses {
 			comparator = route.ResponseComparator
 		}
-		router.Path(route.Path).Methods(route.Methods...).Handler(NewProxyEndpoint(p.backends, route, p.metrics, p.logger, comparator, p.cfg.LogSlowQueryResponseThreshold, p.cfg.SecondaryBackendsRequestProportion))
+		router.Path(route.Path).Methods(route.Methods...).Handler(
+			NewProxyEndpoint(p.backends, route, p.metrics, p.logger, comparator, p.cfg.LogSlowQueryResponseThreshold, p.cfg.SecondaryBackendsRequestProportion).
+				WithSkipPreferredBackendFailures(p.cfg.SkipPreferredBackendFailures),
```
charleskorn, line 312: Given the pattern so far has been to pass config as an argument to `NewProxyEndpoint`, is there a reason to introduce the `With...` pattern here?

julienduchesne: It was getting long 🤷. I don't mind either way

charleskorn: I think I'd prefer to either keep it consistent or change it for all of the options, with a slight preference for keeping things consistent with how they were to reduce the size of this PR

julienduchesne: 👍 Done!

CHANGELOG.md
```diff
@@ -101,6 +101,8 @@
 
 ### Query-tee
 
+* [ENHANCEMENT] Allow setting a "skip" result on comparisons when preferred backend fails #10612
```
charleskorn, line 104: [nit]

```suggestion
* [ENHANCEMENT] Allow skipping comparisons when preferred backend fails. Disabled by default, enable with `-proxy.compare-skip-preferred-backend-failures=true`. #10612
```

charleskorn (approved):

dimitarvdimitrov (approved):

PR #10620, opened 2025-02-10
querier,ingester: support for limit parameter in the finding series endpoint

narqo:
#### What this PR does

This PR updates the `querier` and the `ingester` services, with a basic support for the `limit` parameter in the `/series` API endpoint.

Here the code only passes the limit to the downstream services, as-is. This is not ideal, because the limit of MM series in the distributor can overwhelm the ingesters. I want to focus on that in the follow-up PR to make the review simpler.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/ingester/ingester.go
```diff
@@ -1910,6 +1910,7 @@ func (i *Ingester) MetricsForLabelMatchers(ctx context.Context, req *client.Metr
 		hints := &storage.SelectHints{
 			Start: mint,
 			End:   maxt,
+			Limit: hists.Limit,
```
dimitarvdimitrov, line 1913: hints or hists?

dimitarvdimitrov (approved):
easy, LGTM with two small suggestions

pkg/distributor/distributor.go
```diff
@@ -2771,8 +2771,15 @@ func (d *Distributor) MetricsForLabelMatchers(ctx context.Context, from, through
...
 	queryLimiter := mimir_limiter.QueryLimiterFromContextWithFallback(ctx)
 
-	result := make([]labels.Labels, 0, len(metrics))
+	resultCap := len(metrics)
+	if hints != nil && hints.Limit > 0 {
+		resultCap = min(resultCap, hints.Limit)
+	}
+	result := make([]labels.Labels, 0, resultCap)
 	for _, m := range metrics {
+		if len(result) >= resultCap {
```
pracucci, line 2780: Why not limit few lines above when we populate `metrics` to deduplicate series? We could just stop once `len(metrics)` reached the limit.

Similarly, I'm wondering why we don't apply the limiter in the same `for` loop above when we populate `metrics`. Do you see a good reason to not do it there?

pkg/distributor/distributor_test.go
```diff
@@ -2518,14 +2560,20 @@ func TestDistributor_MetricsForLabelMatchers(t *testing.T) {
...
 						return
 					}
-
 					require.NoError(t, err)
-					requireLabelAdaptersMatchLabels(t, testData.expectedResult, metrics)
+
+					if testData.hints == nil || testData.hints.Limit == 0 {
+						requireLabelAdaptersMatchLabels(t, testData.expectedResult, metrics)
+					} else {
+						// The order of ingester responses isn't guarantied. Thus, we can only test that the distributor's response is in the subset of expectedResult.
```
pracucci, line 2573: [nit]
```suggestion
						// The order of ingester responses isn't guaranteed. Thus, we can only test that the distributor's response is in the subset of expectedResult.
```

pracucci (approved):
Nice work! I left a comment about where to apply limit in a function.

pkg/distributor/distributor.go
narqo: > Do you see a good reason to not do it there?

I think, the only one is that, in the worst case, this will check every duplicate against the series limiter, calculating the hash of the series twice. I cannot tell if that better or worse, from applying this limit later.

PR #10621, opened 2025-02-10
Subquery Spin-Off: Prioritize query sharding

julienduchesne:
In cases where we aggregate the results of a subquery, sharding is faster in some cases, slower in some. Overall, the impact of the feature seems to be near-zero.

Since:
- we're not gaining anything
- subquery spin-off is the newer feature, and it would be preferrable not to introduce a regression on existing queries
- subquery spin-off has an additional overhead of running a second query

I think we can give query sharding the priority

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
 * [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
 * [ENHANCEMENT] Ruler: Adds support for filtering results from rule status endpoint by `file[]`, `rule_group[]` and `rule_name[]`. #10589
+* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries. This is enabled by setting the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regex expressions matching the queries to enable. #10460 #10603 #10621
 * [BUGFIX] Distributor: Use a boolean to track changes while merging the ReplicaDesc components, rather than comparing the objects directly. #10185
 * [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #10154
 * [BUGFIX] Query-frontend and querier: show warning/info annotations in some cases where they were missing (if a lazy querier was used). #10277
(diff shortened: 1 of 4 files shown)


colega (commented):
I think we should have a test that checks this. Also changelog should be updated (adding this PR to the line where feature was implemented)

julienduchesne:
> I think we should have a test that checks this
Done!

> Also changelog should be updated

🤦 I was sure I added a CHANGELOG for this, but seems I forgot. Added it now, with all three PRs listed

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
 * [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
 * [ENHANCEMENT] Ruler: Adds support for filtering results from rule status endpoint by `file[]`, `rule_group[]` and `rule_name[]`. #10589
+* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries. This is enabled by setting the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regex expressions matching the queries to enable. #10460 #10603 #10621
```
tacole02, line 47:
```suggestion
* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries. To enable this, set the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621
```

tacole02 (approved):

CHANGELOG.md
julienduchesne: Thank you 🙌

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Distributor, querier, ingester and store-gateway: Add support for `limit` parameter for label names and values requests. #10410
 * [ENHANCEMENT] Query-frontend: Allow adjustment of queries looking into the future to a maximum duration with experimental `-query-frontend.max-future-query-window` flag. #10547
 * [ENHANCEMENT] Ruler: Adds support for filtering results from rule status endpoint by `file[]`, `rule_group[]` and `rule_name[]`. #10589
+* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries. To enable this, set the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621
```
charleskorn, line 47: [nit] Might be good to add some more detail about why someone would want to do this:

```suggestion
* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries, so that they benefit from query acceleration techniques such as sharding, splitting and caching. To enable this, set the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621
```

charleskorn (approved):

PR #10642, opened 2025-02-13
[Docs] Remove deprecated config: align_queries_with_step

leehack:
#### What this PR does

Remove deprecated config that was removed since 2.14. 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/sources/mimir/configure/configuration-parameters/index.md
+++ b/docs/sources/mimir/configure/configuration-parameters/index.md
@@ -3602,11 +3602,6 @@ The `limits` block configures default and per-tenant limits imposed by component
 # (experimental) List of http requests to block.
 [blocked_requests: <blocked_requests_config...> | default = ]
 
-# Mutate incoming queries to align their start and end with their step to
-# improve result caching.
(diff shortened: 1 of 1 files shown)


56quarters (changes requested):
I don't think this PR is valid. The setting you've mentioned, YAML `limits.align_queries_with_step` and CLI `-query-frontend.align-queries-with-step` are both still supported and not deprecated. The YAML configuration `frontend.align_queries_with_step` _was_ removed in Mimir 2.14 and was removed from the docs. The documentation you are referencing is not that.

leehack:
> I don't think this PR is valid. The setting you've mentioned, YAML `limits.align_queries_with_step` and CLI `-query-frontend.align-queries-with-step` are both still supported and not deprecated. The YAML configuration `frontend.align_queries_with_step` _was_ removed in Mimir 2.14 and was removed from the docs. The documentation you are referencing is not that.

Ah. You are right. Sorry for the confusion. Closing the PR.

Issue #1811, opened 2022-05-03
distributor: add HA elected replica metrics


ortuman:
Add extra `cortex_ha_tracker_elected_replica_status` gauge metric to keep track of elected cluster/replica pair when HA is enabled. 

This new metric should include `user`,`cluster` and `replica` labels, and be described as follows:
```
The elected replica status where 1 means elected and 0 means not elected.
```

Extra logic should be added get rid of metrics for old replicas (no longer sending samples).

pracucci:
FYI, if you're looking for the currently elected replica, it's already exposed by the HA tracker web page. The metric could help investigating the history, but I'm bit concerned about the high cardinality it would have. Maybe logging each replica switch could be a better approach.

PR #10644, opened 2025-02-13
Distributor: Optionally record elected HA replica in metric

julienduchesne:
Closes https://github.com/grafana/mimir/issues/1811

Optionally expose the current HA replica for each tenant in the `cortex_ha_tracker_elected_replica_status` metric. This is enabled with the `-distributor.ha-tracker.enable-elected-replica-metric=true` flag. The flag is there to prevent adding an unexpectedly high cardinality metric.

- [x] Tests updated.
- [x] Documentation added.


aldernero (approved):

56quarters (approved):
Change LGTM but I'm not sure how much of a worry this cardinality really is. There should generally only be two replicas with HA. We're already including cluster and user here so it doesn't seem like much of a burden to include replica, I'm not sure we need the config.

julienduchesne:
> Change LGTM but I'm not sure how much of a worry this cardinality really is. There should generally only be two replicas with HA. We're already including cluster and user here so it doesn't seem like much of a burden to include replica, I'm not sure we need the config.

Some of our tenants do use IPs or UUIDs as replica identifiers 🤷

PR #10604, opened 2025-02-07
Wrap level

bboreham:
#### What this PR does

Logging: expose log level through log wrappers

So that Debug() calls are fast when debug logging is disabled.

#### Which issue(s) this PR fixes or relates to

https://github.com/prometheus/prometheus/pull/15993

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10646, opened 2025-02-14
Logging: detect log level through layers of wrappers

bboreham:
#### What this PR does

Expose log level through log wrappers, so that `Debug()` calls are fast when debug logging is disabled.

It works by making a `Log()` call so it reaches the bottom layer.

However, if `SlogFromGoKit` is used with a logger outside of this package, you will get a spurious log line.

Inspired by https://github.com/prometheus/prometheus/pull/15993

This is a competitor to #10604.  Credit @dimitarvdimitrov for the idea.

Before:
```
BenchmarkSlogFromGoKit/log-28             981555              1176 ns/op            1008 B/op         16 allocs/op
BenchmarkSlogFromGoKit/debuglog-28        949378              1175 ns/op            1008 B/op         16 allocs/op
BenchmarkDebugLog-28                    26430567                44.55 ns/op           96 B/op          2 allocs/op
```

After:
```
BenchmarkSlogFromGoKit/log-28             994729              1165 ns/op            1008 B/op         16 allocs/op
BenchmarkSlogFromGoKit/debuglog-28      358293070                3.342 ns/op           0 B/op          0 allocs/op
BenchmarkDebugLog-28                    24998392                45.67 ns/op           96 B/op          2 allocs/op
```

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/log/slogadapter.go
+++ b/pkg/util/log/slogadapter.go
@@ -12,20 +12,20 @@ import (
 // SlogFromGoKit returns slog adapter for logger.
 func SlogFromGoKit(logger log.Logger) *slog.Logger {
 	var sl slog.Level
-	x, ok := logger.(leveledLogger)
-	if !ok {
+	x := privateLevelDetector{
+		string:   "This struct is expected to be used with levelLogger only",
+		logLevel: debugLevel,
+	}
+	logger.Log("test", &x)
+	switch x.logLevel {
+	case infoLevel:
+		sl = slog.LevelInfo
+	case warnLevel:
+		sl = slog.LevelWarn
+	case errorLevel:
+		sl = slog.LevelError
+	default:
 		sl = slog.LevelDebug
-	} else {
-		switch x.level() {
-		case infoLevel:
-			sl = slog.LevelInfo
-		case warnLevel:
-			sl = slog.LevelWarn
-		case errorLevel:
-			sl = slog.LevelError
-		default:
-			sl = slog.LevelDebug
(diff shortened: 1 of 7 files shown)


dimitarvdimitrov (approved):
LGTM, it's how I imagined it. But would be nice to get a second/third opinion

dimitarvdimitrov:
for the record: I prefer this approach because the other one requires committers be aware of and keep the conventions in the package (of using `wrap` in #10604)

bboreham:
Rebased against current main.

pkg/util/log/slogadapter.go
```diff
@@ -12,20 +12,20 @@ import (
 // SlogFromGoKit returns slog adapter for logger.
 func SlogFromGoKit(logger log.Logger) *slog.Logger {
 	var sl slog.Level
-	x, ok := logger.(leveledLogger)
-	if !ok {
+	x := privateLevelDetector{
+		string:   "This struct is expected to be used with levelLogger only",
+		logLevel: debugLevel,
+	}
+	logger.Log("test", &x)
```
narqo, line 19: Could you add a one-liner code comment with what this "test" does? From the initial pass over the code, I thought it was a left-over, and it took me a while to process it.

bboreham: Done; also changed the string to "probe".  Let me know what you think.

narqo (approved):
Works for me 🔥

PR #10652, opened 2025-02-14
querier: adjust requested limit parameter for downstream requests to series endpoint

narqo:
#### What this PR does

This is a follow-up to https://github.com/grafana/mimir/pull/10620

Here I'm updating the `MetricsForLabelMatchers` to adjust the `limit` parameter passed to the downstream ingesters, base on the replication sets. The idea, suggested by @pracucci is to approximate the limit as:

```
L / (<shard-size> / <replication-factor>)
```

Note that for ingest storage, the above formula is equivalent to `L / <partition-shard-size>`.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -2757,7 +2757,14 @@ func (d *Distributor) MetricsForLabelMatchers(ctx context.Context, from, through
 	resultLimit := math.MaxInt
 	if hints != nil && hints.Limit > 0 {
 		resultLimit = hints.Limit
-		req.Limit = int64(adjustRequestLimitForReplicationSets(ctx, d, replicationSets, resultLimit))
+
+		userID, err := tenant.TenantID(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		// Adjust the limit passed to the downstream ingester with respect to how series are sharded inside the ring.
+		req.Limit = int64(d.adjustQueryRequestLimit(ctx, userID, resultLimit))
 	}
 
 	resps, err := forReplicationSets(ctx, d, replicationSets, func(ctx context.Context, client ingester_client.IngesterClient) (*ingester_client.MetricsForLabelMatchersResponse, error) {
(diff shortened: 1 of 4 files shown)


pkg/distributor/distributor.go
```diff
@@ -2789,6 +2791,41 @@ respsLoop:
...
+func adjustRequestLimitForReplicationSets(ctx context.Context, d *Distributor, replicationSets []ring.ReplicationSet, limit int) int {
+	if limit == 0 {
+		return limit
+	}
+
+	var shardSize int
+	if d.cfg.IngestStorageConfig.Enabled {
+		// When ingest storage is enabled, each partition, represented by one replication set is owned by only one ingester.
+		// So we use the number of replication sets to count the number of shards.
+		shardSize = len(replicationSets)
```
pracucci, line 2805: `len(replicationSets)` includes both ACTIVE and INACTIVE partitions. To be more accurate, I think we should only consider ACTIVE partitions (the shard size on the write path, where series are written, is made only of ACTIVE partitions). It may not be a big deal: if we decide to not address it, then I would at least leave a comment here about it.

pkg/distributor/distributor.go
```diff
@@ -2789,6 +2791,41 @@ respsLoop:
...
+	} else if len(replicationSets) == 1 {
+		// We expect to always have exactly 1 replication set when ingest storage is disabled.
+		// In classic Mimir the total number of shards (ingestion-tenant-shard-size) is the number of ingesters in the shard across all zones.
+		shardSize = len(replicationSets[0].Instances) / d.ingestersRing.ReplicationFactor()
```
pracucci, line 2809: Similar comment about read-only instances here. Another case where the number of instances is higher than the real shard size is when the "lookback" threshold triggers (e.g. for 12h after a scale up). The number of instances we lookup is higher than the real shard size. I think what we really want here is the shard size as computed on the write path, and not the read path. It would be more accurate computing the actual shard size by looking at the min between "configured tenant shard size" and "writeable instances/partitions in the ring".

narqo: I've updated the PR. The idea makes sense to me, although I'm not fully sure whether I've picked the right methods in the ring to calculate what you'd suggested. PTAL

pkg/distributor/distributor.go
```diff
@@ -2789,6 +2798,48 @@ respsLoop:
...
+		// number of shards, or more shards than the number of active partitions in the ring.
+		shardSize = d.partitionsRing.PartitionRing().ShuffleShardSize(d.limits.IngestionPartitionsTenantShardSize(userID))
+	} else {
+		tenantShardSize := d.limits.IngestionTenantShardSize(userID)
+
+		// The ShuffleShard filters out read-only instances, leaving us with the number of active ingesters.
+		subring := d.ingestersRing.ShuffleShard(userID, tenantShardSize)
+		ingestionShardSize := subring.InstancesWithTokensCount()
+		if tenantShardSize > 0 {
+			ingestionShardSize = min(tenantShardSize, ingestionShardSize)
```
pracucci, line 2820: We already pass the `tenantShardSize` to `ShuffleShard()` above. Why do we need this logic here as well? Isn't `ingestionShardSize` already the actual shard size?

pkg/distributor/distributor.go
```diff
@@ -2789,6 +2798,48 @@ respsLoop:
...
+	var shardSize int
+	if d.cfg.IngestStorageConfig.Enabled {
+		// Get the number of active partitions in the ring. Here the ShuffleShardSize handles cases when a tenant has 0 or negative
+		// number of shards, or more shards than the number of active partitions in the ring.
+		shardSize = d.partitionsRing.PartitionRing().ShuffleShardSize(d.limits.IngestionPartitionsTenantShardSize(userID))
+	} else {
+		tenantShardSize := d.limits.IngestionTenantShardSize(userID)
+
+		// The ShuffleShard filters out read-only instances, leaving us with the number of active ingesters.
+		subring := d.ingestersRing.ShuffleShard(userID, tenantShardSize)
```
pracucci, line 2817: This is an expensive operation. We support optional caching. The caching is enabled by default, unless you explictly disable it (because the caching has an impact on memory utilization). I don't see it disabled for querier, so I think we're good using this function.

pkg/distributor/distributor_test.go
```diff
@@ -2593,6 +2593,98 @@ func TestDistributor_MetricsForLabelMatchers(t *testing.T) {
...
+			hints: &storage.SelectHints{
+				Limit: 50,
+			},
+			expectedLimit: 25, // the limit is divided between two shards (ingest_shard_size=6; RF=3)
+		},
+		"should adjust push-down limit if ingest-storage is enabled": {
+			setupFunc: func(testConfig *prepConfig) {
+				testConfig.ingestStorageEnabled = true
+				testConfig.limits = prepareDefaultLimits()
+				testConfig.limits.IngestionPartitionsTenantShardSize = 3
```
pracucci, line 2635: I suggest a test with ingest storage enabled and IngestionPartitionsTenantShardSize=0 (disabled).

pracucci (approved):
Nice work! LGTM. I just a couple of nits.

pkg/distributor/distributor.go
narqo: Good call. Looking at it now, I think I was overly protective here for now reason. I've simplified the chunk.

PR #10659, opened 2025-02-14
[feature] helm: add service.extraPorts to all components

TheRealNoob:
#### What this PR does

In the Helm chart, add `service.extraPorts` to all components that allow defining `extraContainers`.  The intention being that people may expose additional ports on the Pod, and that should be exposed from the Service as well.

I haven't bumped the chart version, could you tell me what it should be bumped to?  I'm not familiar with the format being used.

Didn't open one

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


TheRealNoob:
@narqo Thank you for having a look.  The use-case that I'm working on is exposing prometheus and alertmanager to the internet behind an oauth2 sidecar.  I'm setting them up under new ingress objects so as to not interfere with the main ingress which I have using basicAuth authentication.

I saw in the chart that there are already comments in place on the `alertmanager.extraContainers` spec for exactly this kind of sidecar, in which case this PR is just an expansion of that intention.

Let me know if you need me to go into any more detail.  In short, if someone adds an extraContainer that exposes a new port, there is a very high likelihood that they'll need to expose it on the SVC as well.

operations/helm/charts/mimir-distributed/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 5.7.0-weekly.330
+version: 5.8.0-weekly.330
```
narqo, line 2: Please, remove the changes around the version's bump. We typically don't tag a version on every change; instead, there is a dedicated CI job, that periodically releases a weekly version.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * [CHANGE] Ruler: Add `user` and `reason` labels to `cortex_ruler_write_requests_failed_total` and `cortex_ruler_queries_failed_total`; add `user` to
     `cortex_ruler_write_requests_total` and `cortex_ruler_queries_total` metrics. #10536
 * [FEATURE] Distributor: Add experimental Influx handler. #10153
+* [FEATURE] Helm: Add support for `.service.extraPorts` on all components. #10659
```
narqo, line 19: The Helm cart has its own changelog in the [`operations/helm/charts/mimir-distributed/CHANGELOG.md`](https://github.com/grafana/mimir/blob/03e9612ac35d709c1e8678423e9b3e105379dca3/operations/helm/charts/mimir-distributed/CHANGELOG.md). Please, move this one there.

Also, it would be great if the changelog could add some context about the change, e.g.

```suggestion
[ENHANCEMENT] All componentes can expose additional ports with their respective services via the *.service.extraPorts values. This allows exposing the containers defined in a component's *.extraContainers. #10659
```

TheRealNoob: Updated with your syntax.  Nice suggestion, thanks.

operations/helm/charts/mimir-distributed/Chart.yaml
TheRealNoob: Done.  Appreciate the pointer.

TheRealNoob:
I think the lint check is a false positive caused by the single quote in the word `component's`.  Any preference on how you want to solve?  I see other entries in the changelog with the same single-use single-quote so I'm guessing this is an ignored warning?

narqo (approved):
Sorry for the delayed review. The changes look good to me, I believe the markdown-lint's failure is indeed a false-positive. I can take care of it.

narqo:
@TheRealNoob the SLA signing job says

> Mike Timmerman seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

It seems that this comes after one of your commits that uses this author's name and the email. Could you either add the email to your github account, or update the changeset to use the git author, that signs the SLA. Sorry for extra bureaucracy.

TheRealNoob:
well that was fun.  I was previously changing some settings on my gitconfig to make it easier to work with two accounts from one machine (work and personal) which led to mis-signed commits.  And then the marvelous joys of learning rebase (again).  Never fun, but hopefully one day I'll be able to do something without having to look it up :).  Apologies for the troubles.

PR #10669, opened 2025-02-17
Update `mimirtool grafana analyze` to support more panel types

LasseHels:
#### What this PR does

This pull request updates `mimirtool grafana analyze` to support the following panel types:
* Custom.
* Bar chart.
* Pie chart.
* State timeline.
* Status history.
* Histogram.
* Candlestick.
* Canvas.
* Flame graph.
* Geomap.
* Node Graph.
* Trend.
* XY chart.

At Maersk, our adaptive metrics solution uses `mimirtool grafana analyze` to determine whether a metric is referenced in a dashboard. We discovered that the command skips more than half of the panel types that can reference Prometheus metrics. This is problematic, as we don't want to drop or aggregate metrics that are used in a dashboard.

The panel types in `pkg/mimirtool/minisdk` were unnecessarily complex with most of their fields being unused. Now, there is a single `minisdk.Panel` type that has only the fields that are used.

Previously, the targets of a custom panel would always be skipped. Now, `mimirtool` will include metrics from custom panels if the custom panel follows the same structure as normal panels of having a `target` field in the root panel JSON.

The standalone `TestParseMetricsInBoardWithTimeseriesPanel` test has been merged with the `TestParseMetricsInBoard` test which now tests all panel types in a single test.

#### Checklist

- [x] Tests updated.
- [ ] ~Documentation added.~


CHANGELOG.md
```diff
@@ -113,6 +113,9 @@
 
 * [BUGFIX] Fix issue where `MIMIR_HTTP_PREFIX` environment variable was ignored and the value from `MIMIR_MIMIR_HTTP_PREFIX` was used instead. #10207
 * [ENHANCEMENT] Unify mimirtool authentication options and add extra-headers support for commands that depend on MimirClient. #10178
+* [ENHANCEMENT] `mimirtool grafana analyze` now supports custom panels. #10669
+* [ENHANCEMENT] `mimirtool grafana analyze` now supports bar chart, pie chart, state timeline, status history,
+  histogram, candlestick, canvas, flame graph, geomap, node graph, trend and XY chart panels. #10669
```
tacole02, line 118:
```suggestion
  histogram, candlestick, canvas, flame graph, geomap, node graph, trend, and XY chart panels. #10669
```

LasseHels: Fixed!

PR #10682, opened 2025-02-18
fix: update azure blob storage documentation

andreppires:
#### What this PR does
Backend type definition for storage was missing for non-common scenarios.
This PR updated the example for Azure Blob Storage

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/sources/mimir/configure/configure-object-storage-backend.md
+++ b/docs/sources/mimir/configure/configure-object-storage-backend.md
@@ -133,14 +133,17 @@ common:
       endpoint_suffix: "blob.core.windows.net"
 
 blocks_storage:
+  backend: azure
   azure:
     container_name: mimir-blocks
 
 alertmanager_storage:
+  backend: azure
   azure:
     container_name: mimir-alertmanager
 
 ruler_storage:
+  backend: azure
   azure:
     container_name: mimir-ruler
 ```


colega:
Hi, you're supposed to not need to specify the `backend` (or any other config) on each storage if it's already specified in `common:`, didn't this work for you?

There's [a test for this, for reference](https://github.com/grafana/mimir/blob/99118d9561d0390fa0fbb6699e8087598cc0bc44/pkg/mimir/mimir_config_test.go#L17-L55).

tacole02 (approved):
Thanks for updating!

andreppires:
Hey @colega 
That was what I was expecting, following the documentation. 
But turns out, adding the backend field was what unblocked my day. 

> Hi, you're supposed to not need to specify the `backend` (or any other config) on each storage if it's already specified in `common:`, didn't this work for you?
> 
> There's [a test for this, for reference](https://github.com/grafana/mimir/blob/99118d9561d0390fa0fbb6699e8087598cc0bc44/pkg/mimir/mimir_config_test.go#L17-L55).

dimitarvdimitrov:
if you're using this in the helm chart, then you need per-block backend 😞 that's because the helm chart explicitly sets the backend 

https://github.com/grafana/mimir/blob/3b560f01bac5e8b382afed62f61902597445fe96/operations/helm/charts/mimir-distributed/values.yaml#L189

colega:
@dimitarvdimitrov maybe we should fix the Helm chart instead of documenting the bug?

I see all the rest of the sections in Helm only set `backend: s3` under the `{{- if .Values.minio.enabled }}` condition, except for the `blocks_storage`, and `block_storage` only configures `s3` if that condition is true, so it just doesn't work out of the box.

dimitarvdimitrov:
ok, so i merged this PR but then @colega convinced me if it is only the Helm chart, then we should fix that instead of the docs. So I reverted this PR.

Is my assumption correct, @andreppires, that you are using this in the helm chart, so that will be the main problem?

andreppires:
Sorry for the late reply, @dimitarvdimitrov 
Yes, I'm using helm chart to deploy Mimir. Good to know that this was indeed an issue related with only one of the deployment mechanisms.

PR #10684, opened 2025-02-18
Compactor: Upload Sparse Index Headers to Object Storage

### LINKED ISSUE #8166: Compactor: store sparse index headers in object store
### Background 

https://github.com/grafana/mimir/issues/5046

### Problem

This manifests when scaling out store-gateways: new store-gateways download index headers to disk whenever they start and before becoming ready. However, the **sparse** index headers are only constructed when the block is lazily-loaded and are local to a store-gateway replica. This means that a new store-gateway replica doesn't have any sparse index headers right after starting up and has to load the index header and create the sparse header from the data on disk before being able to serve requests for a given block. This leads to severe increases in latency after scaling out store-gateways.

### Proposal 

1. In store-gateways download sparse index headers when adding a new block (either periodic syncs or at startup). This solves the problem of scaling out 
2. If the sparse header doesn't exist in object store, then construct the sparse header and upload it.

#### Compactor or store-gateway

Part 2. can be done by either the compactor or the store-gateway. Doing this in the store-gateway would require less code restructuring and can be a good first iteration. 

The possible complication is that the store-gateway now has to load the index header upon downloading a new block which no other store-gateway has synced before. While the store-gateway should unload it immediately after constructing the sparse index header, this can increase the time to do the periodic syncs. The situation shouldn't be worse than today where this time penalty is paid at query time. Having to load the index header would also impact the compactor compaction cycle latency.

dmwilson-grafana:
#### What this PR does

This PR enables uploading sparse index headers to object storage from the compactor. Previously, managing sparse index headers has been the sole responsibility of each store-gateway instance (see: #8166).

This PR moves the contents of `/pkg/storegateway/indexheader` to `/pkg/util/indexheader` and uses a `StreamBinaryReader` in the compactor to take the just-compacted block index file and create sparse-index-header file. 

Fixes #8166 - New store-gateway replicas don't have any sparse index headers right after starting up and must load the index header and create the sparse header from the data on disk before being able to serve requests for a given block. This leads to severe increases in latency after scaling out store-gateways. 

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


56quarters:
Before this is marked as ready for review, `pkg/storage` or somewhere under there is the directory that the index header code should live in instead of `pkg/util`

pkg/compactor/bucket_compactor.go
```diff
@@ -431,6 +448,32 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
+		if uploadSparseIndexHeaders {
+			// Calling NewStreamBinaryReader reads a block's index and writes a sparse-index-header to disk. Because we
+			// don't use the writer, we pass a default indexheader.Config and don't register metrics.
+			if _, err := indexheader.NewStreamBinaryReader(
+				ctx,
+				jobLogger,
+				fsbkt,
+				subDir,
+				blockToUpload.ulid,
+				mimir_tsdb.DefaultPostingOffsetInMemorySampling,
```
dmwilson-grafana, line 460: @dimitarvdimitrov  - I wanted your get your opinion on using `DefaultPostingOffsetInMemorySampling` here. Could a situation arise where:

1. `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` is set to a non-default value
2.  store gateway reads sparse headers uploaded by the compactor with the default sampling value 
3.  Because the store gateway has sparse headers with an unexpected sampling rate - query performance is worse than if the store gateway built the header itself.

dimitarvdimitrov: oh, nice catch!

This is actually something that we have overlooked when building sparse headers. The same can happen when a store-gateway is restarted with a different value for the sampling flag.

I can think of three ways to solve this

1. Persist the sampling inside the sparse header protobuf. If the sampling doesn't match what the store-gateway is currently configured with, then either
    1a. the store-gateway rebuilds the sparse header with its configured sampling rate
    1b. the store-gateway uses the sampling found in the sparse header when calling `NewPostingOffsetTableFromSparseHeader`
3. remove the flag. We haven't used it in the last 3.5 years since I've been here (it was added 4.5y ago). The 32 is maybe a good tradeoff and a very reasonable default.

I'd go with either 1a. or 2., leaning more towards 2. to reduce config surface and simplify the code. @56quarters what's your take?

dmwilson-grafana: I don't know how much of a difference this would make in terms of query performance. I think we need to do some testing to figure that out, but in the event it's a meaningful difference, I'm also leaning towards 1a. 

On the most recent commit I began implementing 1a with a minor change. When the two rates aren't equal, we can downsample (rather than rebuild) if `-blocks-storage.bucket-store.posting-offsets-in-mem-sampling` < `DefaultPostingOffsetInMemorySampling` (e.g. 1/64 < 1/32).

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -184,6 +208,18 @@ func newFileStreamBinaryReader(binPath string, id ulid.ULID, sparseHeadersPath s
...
 
+func (r *StreamBinaryReader) downsampleSparseIndexHeader(curSampling, tgtSampling int) error {
+	tbl, ok := r.postingsOffsetTable.(*streamindex.PostingOffsetTableV2)
+	if !ok {
+		return fmt.Errorf("postings offset table has incompatible version v%d", r.indexVersion)
+	}
+	if err := tbl.DownsamplePostings(curSampling, tgtSampling); err != nil {
+		return err
+	}
+	r.postingsOffsetTable = tbl
```
dimitarvdimitrov, line 219: why is this reassignment necessary?

pkg/storage/indexheader/header_test.go
```diff
@@ -129,6 +129,68 @@ func TestReadersComparedToIndexHeader(t *testing.T) {
...
+			require.NoError(t, err)
+
+			// label names are equal between original and downsampled sparse index-headers
+			require.ElementsMatch(t, downsampleLabelNames, origLabelNames)
+
+			// downsampled postings are a subset of the original sparse index-header postings
+			origIdxpbTbl := br1.postingsOffsetTable.NewSparsePostingOffsetTable()
+			downsampleIdxpbTbl := br2.postingsOffsetTable.NewSparsePostingOffsetTable()
+
+			rr := tt.inMemSamplingRate / tt.protoRate
```
dimitarvdimitrov, line 184: can you rename `rr` to something else? I'm not sure what it's meant to mean

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -156,18 +184,14 @@ func newFileStreamBinaryReader(binPath string, id ulid.ULID, sparseHeadersPath s
...
-				return nil, fmt.Errorf("cannot load sparse index-header from disk: %w", err)
-			}
 		}
 	} else {
 		if err = r.loadFromIndexHeader(logger, id, cfg, indexLastPostingListEndBound, postingOffsetsInMemSampling); err != nil {
 			return nil, fmt.Errorf("cannot load sparse index-header: %w", err)
 		}
 	}
 
+	level.Debug(logger).Log("msg", "built sparse index-header file", "id", id, "path", sparseHeadersPath)
```
dimitarvdimitrov, line 194: we keep repeating the `id` and `path` args. Maybe we can just swap out the logger with one which has them attached. Something like `logger = log.With(logger, "id", id, "path", path)` at the top of the function

dimitarvdimitrov (commented):
thanks for working on this. The logic looks sound, but i think we can move around where it sits

pkg/storage/indexheader/index/postings.go
```diff
@@ -608,10 +641,18 @@ func (t *PostingOffsetTableV2) LabelNames() ([]string, error) {
 	return labelNames, nil
 }
 
+func (t *PostingOffsetTableV2) PostingOffsetInMemSampling() int {
+	if t != nil {
```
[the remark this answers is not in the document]
dmwilson-grafana: As things are written right now, this method is never called when `t` is nil. However, if you create a `StreamBinaryReader` and call `sbr.postingsOffsetTable.PostingOffsetInMemSampling` before `sbr.loadFromSparseIndexHeader`, this could panic unless we handle for `t` being nil. To prevent that, I just took the approach that most proto structs take for their `GetSomeField()` method.

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -141,13 +141,41 @@ func newFileStreamBinaryReader(binPath string, id ulid.ULID, sparseHeadersPath s
...
 	if r.indexVersion == index.FormatV2 {
+		var reconstruct bool
 		sparseData, err := os.ReadFile(sparseHeadersPath)
 		if err != nil && !os.IsNotExist(err) {
 			level.Warn(logger).Log("msg", "failed to read sparse index-headers from disk; recreating", "id", id, "err", err)
 		}
 
+		// read persisted sparseHeaders from disk to memory.
+		if err == nil {
+			if err = r.loadFromSparseIndexHeader(logger, id, sparseHeadersPath, sparseData, postingOffsetsInMemSampling); err != nil {
```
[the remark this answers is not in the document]
dmwilson-grafana: This is a good suggestion. I  think I can rework this block (L144-L165) so it's is essentially unchanged from the current main branch.

pkg/compactor/bucket_compactor.go
```diff
@@ -393,13 +395,28 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
+		// error affecting the sparse-index-header upload.
+		fsbkt, err = filesystem.NewBucket(subDir)
+		if err != nil {
+			level.Warn(jobLogger).Log("msg", "failed to create filesystem bucket, skipping sparse header upload", "err", err)
+			uploadSparseIndexHeaders = false
+		}
+	}
+
 	blocksToUpload := convertCompactionResultToForEachJobs(compIDs, job.UseSplitting(), jobLogger)
 	err = concurrency.ForEachJob(ctx, len(blocksToUpload), c.blockSyncConcurrency, func(ctx context.Context, idx int) error {
```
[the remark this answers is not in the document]
dmwilson-grafana: I agree w. this suggestion. Most recent commit includes this as part of `block.Upload`. Made it best effort, the upload will proceed even if the header upload fails. Including it in `block.Upload` is also nice b.c. it ensures that a block directory will have no files uploaded after `meta.json`.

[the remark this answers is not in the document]
dmwilson-grafana: I wanted to keep it to one set of concurrent workers to make sure we called `indexheader.NewStreamBinaryReader` after we call `block.VerifyBlock` and before we call `block.Upload`.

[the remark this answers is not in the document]
dmwilson-grafana: `indexheader.NewStreamBinaryReader` is doing all the work to create the sparse headers. Since we're passing constants into `indexheader.NewStreamBinaryReader`, I can wrap it with a function that has a smaller signature.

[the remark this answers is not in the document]
dimitarvdimitrov: i still think we should break this out into stages and not do everything in a funciton (verify block, create index header, create sparse header, upload block, upload sparse header). Can we break out block verification into its own concurrency job? or move block verification after block creation in the compaction concurrency job?

pkg/storage/indexheader/header_test.go
```diff
@@ -177,15 +179,31 @@ func Test_DownsampleSparseIndexHeader(t *testing.T) {
...
+				// downsampled postings are a subset of the original sparse index-header postings
+				if (tt.inMemSamplingRate > tt.protoRate) && (tt.inMemSamplingRate%tt.protoRate == 0) {
+					require.Equal(t, (len(vals.Offsets)+step-1)/step, len(downsampledOffsets))
+					require.Subset(t, vals.Offsets, downsampledOffsets)
+				}
+
+				// cannot "upsample" a sparse index header. When the sampling value found in sparse header is larger
+				// than the configured rate, should rebuild sparse headers. The rebuilt headers should be at least
+				// as large as the original sparse index header postings.
+				if tt.inMemSamplingRate < tt.protoRate {
```
dimitarvdimitrov, line 198: since you moved downsampling to the postingsOffseTable, do you want to make the test simpler and just test this in index/postings_test.go too or prefer to keep it here?

pkg/storage/indexheader/index/postings_test.go
```diff
@@ -101,32 +100,3 @@ func TestPostingValueOffsets(t *testing.T) {
 		})
 	}
 }
-
-func Test_PostingValueOffsets_Downsample(t *testing.T) {
```
dimitarvdimitrov: can you rework it so we still have a unit test for downsampling next to the code?

pkg/storage/indexheader/index/postings.go
```diff
@@ -234,18 +238,34 @@ func NewPostingOffsetTableFromSparseHeader(factory *streamencoding.DecbufFactory
...
+		step, ok = stepSize(pbSampling, postingOffsetsInMemSampling)
+		if !ok {
+			return nil, fmt.Errorf("sparse index-header sampling rate not compatible with in-mem-sampling rate")
 		}
+	} else if (pbSampling > 0) && (pbSampling > postingOffsetsInMemSampling) {
+		// if the sparse index-header sampling rate is set higher must reconstruct from index-header
+		return nil, fmt.Errorf("sparse index-header sampling rate exceeds in-mem-sampling rate")
+	}
 
+	for sName, sOffsets := range postingsOffsetTable.Postings {
```
dimitarvdimitrov, line 256: what happens if the proto file didn't contain offset sampling? don't we get a panic? I'm thinking about the case when a new store-gateway with this change starts against a sparse header written by an old store-gateway without this change 

i think we need a unit test for this too

pkg/storage/indexheader/index/postings.go
```diff
@@ -234,18 +238,34 @@ func NewPostingOffsetTableFromSparseHeader(factory *streamencoding.DecbufFactory
 		postingOffsetsInMemSampling: postingOffsetsInMemSampling,
 	}
 
-	for sName, sOffsets := range postingsOffsetTable.Postings {
-		t.postings[sName] = &postingValueOffsets{
-			offsets: make([]postingOffset, len(sOffsets.Offsets)),
+	var step int
+	var ok bool
+	pbSampling := int(postingsOffsetTable.GetPostingOffsetInMemorySampling())
+	if (pbSampling > 0) && (pbSampling <= postingOffsetsInMemSampling) {
```
dimitarvdimitrov, line 244: if `pbSampling == postingOffsetsInMemSampling` do we need to run downsampling at all?

pkg/storage/tsdb/block/block.go
```diff
@@ -121,6 +121,14 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blockDi
 		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload index"))
 	}
 
+	if err := objstore.UploadFile(ctx, logger, bkt,
+		filepath.Join(blockDir, SparseIndexHeaderFilename),
+		filepath.Join(id.String(), SparseIndexHeaderFilename),
+	); err != nil {
+		// Don't call cleanUp. Uploading sparse index headers is best effort.
+		level.Warn(logger).Log("msg", "failed to upload sparse index headers", "block", id.String(), "err", err)
+	}
```
dimitarvdimitrov, line 130: this would log a warning even if the config option to build sparser headers is disabled. Can we not attempt to upload the file if it's not on the local fs?

pkg/storage/indexheader/index/postings.go
dmwilson-grafana: if `pbSampling == postingOffsetsInMemSampling` we'd use a step size of 1 which would be the same as constructing the posting offset table normally. Agree with your comment above re: testing, I'll add tests for this so it's more clear.

pkg/storage/indexheader/index/postings_test.go
dmwilson-grafana: Wrote a new test, `Test_NewPostingOffsetTableFromSparseHeader` that tests going from `indexheaderpb.PostingOffsetTable` to `index.PostingOffsetTableV2` and simplified `Test_DownsampleSparseIndexHeader`.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269 #10702
 * [FEATURE] Ruler: Added `/ruler/tenants` endpoints to list the discovered tenants with rule groups. #10738
+* [FEATURE] Distributor: Add experimental Influx handler. #10153
```
dimitarvdimitrov, line 9:
```suggestion
```

CHANGELOG.md
```diff
@@ -16,8 +16,8 @@
...
 * [CHANGE] Ruler: Add `user` and `reason` labels to `cortex_ruler_write_requests_failed_total` and `cortex_ruler_queries_failed_total`; add `user` to
     `cortex_ruler_write_requests_total` and `cortex_ruler_queries_total` metrics. #10536
-* [CHANGE] Querier / Query-frontend: Remove experimental `-querier.promql-experimental-functions-enabled` and `-query-frontend.block-promql-experimental-functions` CLI flags and respective YAML configuration options to enable experimental PromQL functions. Instead access to experimental PromQL functions is always blocked. You can enable them using the per-tenant setting `enabled_promql_experimental_functions`. #10660 #10712
-* [FEATURE] Distributor: Add experimental Influx handler. #10153
+* [CHANGE] Querier / Query-frontend: Remove experimental `-querier.promql-experimental-functions-enabled` and `-query-frontend.block-promql-experimental-functions` CLI flags and respective YAML configuration options to enable experimental PromQL functions. Instead access to experimental PromQL functions is always blocked. You can enable them using the per-tenant setting `enabled_promql_experimental_functions`. #10660
```
dimitarvdimitrov, line 19:
```suggestion
* [CHANGE] Querier / Query-frontend: Remove experimental `-querier.promql-experimental-functions-enabled` and `-query-frontend.block-promql-experimental-functions` CLI flags and respective YAML configuration options to enable experimental PromQL functions. Instead access to experimental PromQL functions is always blocked. You can enable them using the per-tenant setting `enabled_promql_experimental_functions`. #10660 #10712
```

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@
...
     `cortex_ruler_write_requests_total` and `cortex_ruler_queries_total` metrics. #10536
 * [CHANGE] Querier / Query-frontend: Remove experimental `-querier.promql-experimental-functions-enabled` and `-query-frontend.block-promql-experimental-functions` CLI flags and respective YAML configuration options to enable experimental PromQL functions. Instead access to experimental PromQL functions is always blocked. You can enable them using the per-tenant setting `enabled_promql_experimental_functions`. #10660 #10712
-* [FEATURE] Distributor: Add experimental Influx handler. #10153
+* [CHANGE] Store-gateway: Include posting sampling rate in sparse index headers. When the sampling rate isn't set in a sparse index header, store gateway will rebuild the sparse header with the configured `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` value. If the sparse header's sampling rate is set, but doesn't match the configured rate, store gateway will either rebuild the sparse header or downsample to the configured sampling rate. #10684
```
dimitarvdimitrov, line 20:
```suggestion
* [FEATURE] Distributor: Add experimental Influx handler. #10153
* [CHANGE] Store-gateway: Include posting sampling rate in sparse index headers. When the sampling rate isn't set in a sparse index header, store gateway will rebuild the sparse header with the configured `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` value. If the sparse header's sampling rate is set, but doesn't match the configured rate, store gateway will either rebuild the sparse header or downsample to the configured sampling rate. #10684
```

pkg/compactor/bucket_compactor.go
```diff
@@ -457,10 +489,20 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
 
+func prepareSparseIndexHeader(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string, id ulid.ULID, sampling int, cfg indexheader.Config) error {
+	// Calling NewStreamBinaryReader reads a block's index and writes a sparse-index-header to disk.
+	mets := indexheader.NewStreamBinaryReaderMetrics(nil)
+	br, err := indexheader.NewStreamBinaryReader(ctx, logger, bkt, dir, id, sampling, mets, cfg)
+	if err != nil {
+		return err
+	}
+	br.Close()
+	return nil
```
dimitarvdimitrov, line 503:
```suggestion
	return br.Close()
```

pkg/storage/indexheader/index/postings_test.go
```diff
@@ -100,3 +104,57 @@ func TestPostingValueOffsets(t *testing.T) {
...
+		"downsample_noop_proto_has_equal_sampling_rate":           {100, 32, 32, 100, false},
+		"downsample_short_offsets":                                {2, 32, 16, 1, false},
+		"downsample_noop_short_offsets":                           {1, 32, 16, 1, false},
+		"downsample_proto_has_divisible_sampling_rate":            {100, 32, 16, 50, false},
+		"cannot_downsample_proto_has_no_sampling_rate":            {100, 32, 0, 0, true},
+		"cannot_upsample_proto_has_less_frequent_sampling_rate":   {100, 32, 64, 0, true},
+		"cannot_downsample_proto_has_non_divisible_sampling_rate": {100, 32, 10, 0, true},
+		"downsample_sampling_rates_ratio_does_not_divide_offsets": {33, 32, 16, 17, false},
+		"downsample_sampling_rates_ratio_exceeds_offset_len":      {10, 1024, 8, 1, false},
+		"downsample_sampling_rates_ratio_equals_offset_len":       {100, 100, 1, 1, false},
```
dimitarvdimitrov, line 134: can you put the fields on separate lines and add the field names? it's easier to read the test cases that way without having to look back at the test case definition. 

I normally do it for structs that have more than 2 fields

pkg/compactor/bucket_compactor.go
dimitarvdimitrov: this made me realize we don't ever log these errors. Can we do that before we decide to reconstruct the sparse header?

pkg/storage/indexheader/stream_binary_reader_test.go
```diff
@@ -91,6 +91,7 @@ func TestStreamBinaryReader_CheckSparseHeadersCorrectnessExtensive(t *testing.T)
 
 				// Check correctness of sparse index headers.
 				compareIndexToHeader(t, b, r2)
+				compareIndexToHeaderPostings(t, b, r2)
```
dimitarvdimitrov, line 94: can you embed this into the `comapreIndexToHeader`? I see `TestReadersComparedToIndexHeader` is also suffering from the same

pkg/storage/indexheader/header_test.go
```diff
@@ -129,6 +129,166 @@ func TestReadersComparedToIndexHeader(t *testing.T) {
...
+	require.NoError(t, err)
+	defer func() {
+		_ = ir.Close()
+	}()
+
+	toc, err := index.NewTOCFromByteSlice(indexByteSlice)
+	require.NoError(t, err)
+
+	tblOffsetBounds := make(map[string][2]int64)
+	err = index.ReadPostingsOffsetTable(indexByteSlice, toc.PostingsTable, func(label []byte, _ []byte, _ uint64, offset int) error {
```
dimitarvdimitrov, line 268: here can't we use `ir.PostingsRanges()` instead of gathering the ranges ourselves?

pkg/storage/indexheader/header_test.go
```diff
@@ -129,6 +129,166 @@ func TestReadersComparedToIndexHeader(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+
+	tbl := sbr.postingsOffsetTable.NewSparsePostingOffsetTable()
+
+	expLabelNames, err := ir.LabelNames(context.Background())
+	require.NoError(t, err)
+	for _, lname := range expLabelNames {
+		offsets := tbl.Postings[lname].Offsets
+		assert.Equal(t, offsets[0].TableOff, tblOffsetBounds[lname][0])
```
dimitarvdimitrov, line 287: also think we should test both the start and end of each posting offset, not just the start of the first and the end of the last

pkg/storage/indexheader/header_test.go
```diff
@@ -129,6 +129,166 @@ func TestReadersComparedToIndexHeader(t *testing.T) {
...
+		return nil
+	})
+	require.NoError(t, err)
+
+	tbl := sbr.postingsOffsetTable.NewSparsePostingOffsetTable()
+
+	expLabelNames, err := ir.LabelNames(context.Background())
+	require.NoError(t, err)
+	for _, lname := range expLabelNames {
+		offsets := tbl.Postings[lname].Offsets
```
dimitarvdimitrov, line 286: didn't we talk that we need to test the reverse of `compareIndexToHeader`? So we would need to test all the elements in what the prometheus `indexReader.PostingRanges()` returns (`expRanges` in the test below) is also contained in what `headerReader.LabelValuesOffsets` and `headerReader.PostingsOffset` return? 

right now this tests the first and last posting list offsets elements, not all of them.

dmwilson-grafana: I think I misunderstood you before, but I just added a few lines to this test to explicitly check that in addition to "all ranges in the header are in the index" we also test "all ranges in the index are in the header".

pkg/storage/indexheader/header_test.go
dmwilson-grafana: It's not clear to me how we'd do that without re-implementing `index.newV2PostingOffsetTable`. I was under the impression that the only (labelname, labelvalue) pairs that must appear in `Postings.Offsets` are the first and last for a label.

pkg/storage/indexheader/header_test.go
dmwilson-grafana: `ir.PostingsRanges()` returns ranges (`index.Range`). But we want to check that the sparse header contains specific entries from the index. So I used `index.ReadPostingsOffsetTable()` and this map to get the first and last offset for each label in the index and then I check for those offsets in the sparse index header.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -4976,6 +4976,13 @@ sharding_ring:
 # blocks are considered regardless of their upload time.
 # CLI flag: -compactor.max-lookback
 [max_lookback: <duration> | default = 0s]
+
+# (experimental) If enabled, the compactor will construct and upload sparse
```
tacole02, line 4980:
```suggestion
# (experimental) If enabled, the compactor constructs and uploads sparse
```

tacole02: We try to avoid using future tense (i.e. "will") in the docs.

tacole02 (approved):
Thanks, @dmwilson-grafana !

PR #10690, opened 2025-02-19
mixin: add GEM-flavoured compiled mixin

dimitarvdimitrov:
#### What this PR does

This is intended to be used by GEM installations. Soon it will include GEM-specific dashboards and alerts.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


56quarters (approved):
Nicely done.

operations/mimir-mixin-gem.zip
charleskorn, line 1: Should this be ignored in `operations/.gitignore` like the other compiled zip files?

operations/mimir-mixin/mixin-compiled-gem.libsonnet
```diff
@@ -0,0 +1,26 @@
+(import 'mixin.libsonnet') + {
+  // This config enables features that are relevant to most GEM deployments.
+  _config+:: {
+    gem_enabled: true,
+
+    gateway_enabled: true,
```
charleskorn, line 6: Should we make `gateway_enabled` default to `true` if `gem_enabled` is `true`?

dimitarvdimitrov: yep; will do in the next PR because gem_enabled isn't actually defined

PR #10691, opened 2025-02-19
Alertmanager: Initialize skipped Grafana Alertmanagers receiving requests

santihernandezc:
### Description

If the `grafana_alertmanager_conditionally_skip_tenant_suffix` option is configured and a Grafana Alertmanager tenant doesn't have a promoted, non-default, non-empty configuration, we skip initializing it.

The problem is that clients making requests against an uninitialized Alertmanager get `406`s. Unless a Grafana Alertmanager has a "usable" configuration, users won't be able to test templates, test and get receivers, create silences, etc.

This PR makes the multi-tenant Alertmanager start per-tenant Grafana Alertmanagers on incoming requests. This way, requests can be handled even if tenants were initially skipped.

It also adds a grace period for idle Alertmanagers. Whenever a skipped Alertmanager gets a request, we start the Alertmanager and keep a timestamp indicating when this request was received. After the grace period elapses, we shut down the Alertmanager.

### Testing

I tested this PR by spinning up two Alertmanagers ([read-write mode](https://github.com/grafana/mimir/tree/main/development/mimir-read-write-mode)) with:
- `multitenancy_enabled: true`
- `grafana_alertmanager_compatibility_enabled: true`
- `grafana_alertmanager_conditionally_skip_tenant_suffix: -grafana`

I then created 200 Alertmanager tenants with empty configurations:
- 100 of them not matching the configured suffix and using an empty configuration
- 99 of them matching the suffix and using an empty configurations
- 1 of them matching the suffix and using a promoted, non-default, non-empty config

The 99 tenants with empty configuration and matching the suffix were initially skipped. I then sent test alerts for each tenant matching the suffix. Alertmanagers for each of them were started.

After the default grace period passed, all Alertmanagers for tenants matching the configured suffix were stopped, except the one using a promoted, non-default, non-empty configuration.

<details><summary>Query for tenants skipped per instance</summary>

Screenshot 2025-02-20 at 5 20 16 PM

</details>

<details><summary>Query for active Alertmanagers by type (Mimir/Grafana)</summary>

Screenshot 2025-02-20 at 5 23 46 PM

</details>


pkg/alertmanager/alertmanager.go
```diff
@@ -104,7 +104,6 @@ type Config struct {
 	PersisterConfig   PersisterConfig
 
 	GrafanaAlertmanagerCompatibility bool
-	GrafanaAlertmanagerTenantSuffix  string
```
santihernandezc, line 107: Unrelated fix, this was not being used here.

tacole02 (approved):
Docs look great! Thank you!

pkg/alertmanager/multitenant.go
```diff
@@ -1025,6 +1084,33 @@ func (am *MultitenantAlertmanager) serveRequest(w http.ResponseWriter, req *http
 	http.Error(w, "the Alertmanager is not configured", http.StatusPreconditionFailed)
 }
 
+// startAlertmanager will start the Alertmanager for a tenant, using the fallback configuration if no config is found.
+func (am *MultitenantAlertmanager) startAlertmanager(ctx context.Context, userID string) (*Alertmanager, error) {
```
JacobsonMT, line 1088: I think this method can be combined with `alertmanagerFromFallbackConfig`, they are effectively doing the same thing. Probably keep the `startAlertmanager` name and make sure the correct errors are returned from the combined method.

pkg/alertmanager/multitenant.go
```diff
@@ -1002,6 +1040,27 @@ func (am *MultitenantAlertmanager) serveRequest(w http.ResponseWriter, req *http
...
+		if _, ok := am.receivingRequests.Load(userID); ok {
+			level.Debug(am.logger).Log("msg", "updating last request reception time", "user", userID)
+			am.receivingRequests.Store(userID, time.Now().Unix())
+		}
+		return
+	}
+
+	// If the Alertmanager initialization was skipped, start the Alertmanager.
+	if _, ok := am.receivingRequests.Load(userID); ok {
+		userAM, err = am.startAlertmanager(req.Context(), userID)
```
JacobsonMT, line 1054: This seems like a more interesting metric to collect: When a skipped config was initialized because of a request. This will give us a feel for flappiness and how effectively our idle timeout value is.

pkg/alertmanager/multitenant.go
```diff
@@ -727,12 +743,35 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
...
+			return cfg, true, nil
+		}
+
+		// If the tenant ID matches the configured Grafana suffix, only run the Alertmanager if it's receiving requests.
+		createdAt, ok := am.receivingRequests.Load(cfgs.Mimir.User)
+		if !ok {
+			// Store a zero-value timestamp to indicate that we've skipped the tenant.
+			am.receivingRequests.Store(cfgs.Mimir.User, time.Time{}.Unix())
+			return cfg, false, nil
+		}
```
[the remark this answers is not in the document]
santihernandezc: Good catch! I don't know why I thought it was not possible to use it here...

pkg/alertmanager/multitenant.go
```diff
@@ -727,12 +743,35 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 		AlertConfigDesc: cfgs.Mimir,
```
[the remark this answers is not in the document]
santihernandezc: That bit of code gets executed after we've checked for usable (promoted, non-default, non-empty) Grafana configuration. If we reach this far, the tenant has a Grafana configuration we can use to start their Alertmanager.

It does indeed not match our current approach. It's part of the original one, where Grafana and non-Grafana tenants were the same, and they couldn't be distinguished by a suffix.

It's not strictly necessary to add `computeConfig()` to the `startAlertmanager()` function as it will only be called for skipped Grafana Alertmanagers, but it could be a good idea. Otherwise, configurations will change on the next sync loop (we'll use the Grafana configuration instead of the default Mimir one).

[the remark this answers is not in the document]
santihernandezc: That bit of code gets executed after we've checked for usable (promoted, non-default, non-empty) Grafana configuration. If we reach this far, the tenant has a Grafana configuration we can use to start their Alertmanager.

It does indeed not match our current approach. It's part of the original one, where Grafana and non-Grafana tenants were the same, and they couldn't be distinguished by a suffix.

It's not necessary to add `computeConfig()` to the `startAlertmanager()` function as it will only be called for skipped Grafana Alertmanagers. If a Grafana Alertmanager was skipped, it had no usable configuration, so we can just use a default config.

pkg/alertmanager/multitenant.go
santihernandezc: A per-tenant metric would be kind of high in cardinality, and we can assess the flappiness of a given tenant by using something like `group by (pod) (cortex_alertmanager_config_hash{user="<user>"}`

pkg/alertmanager/multitenant.go
santihernandezc: I didn't want to change the behavior of Mimir Alertmanager tenants, at least not yet. I was planning on doing that on a future PR, when applying strict initialization to all tenants.

pkg/alertmanager/multitenant.go
```diff
@@ -83,8 +83,9 @@ type MultitenantAlertmanagerConfig struct {
 
 	EnableAPI bool `yaml:"enable_api" category:"advanced"`
 
-	GrafanaAlertmanagerCompatibilityEnabled bool   `yaml:"grafana_alertmanager_compatibility_enabled" category:"experimental"`
-	GrafanaAlertmanagerTenantSuffix         string `yaml:"grafana_alertmanager_conditionally_skip_tenant_suffix" category:"experimental"`
+	GrafanaAlertmanagerCompatibilityEnabled bool          `yaml:"grafana_alertmanager_compatibility_enabled" category:"experimental"`
+	GrafanaAlertmanagerTenantSuffix         string        `yaml:"grafana_alertmanager_conditionally_skip_tenant_suffix" category:"experimental"`
+	GrafanaAlertmanagerIdleGracePeriod      time.Duration `yaml:"grafana_alertmanager_idle_grace_period" category:"experimental"`
```
santihernandezc, line 88: This is the only new argument.

PR #10696, opened 2025-02-19
frontend: More versatile configuration for subquery spin-off

julienduchesne:
Expands on https://github.com/grafana/mimir/pull/10460

Testing has shown nice results for the feature. In our main Mimir cell, the overall runtime of all instant queries containing the "subquery pattern" (`[<range>:(<step>)?]`) has gone down by half, and the feature allows for queries further into the past by sharding by time

Currently, this is only configurable with `-query-frontend.spin-off-instant-subqueries-to-url` + a required per-tenant `instant_queries_with_subquery_spin_off` opt-in (list of regexp)

This PR makes the configs more consistent and allows for a global enable. Here are the new configs:
- `-query-frontend.instant-subquery-spin-off.enabled`. False by default, set to true to enable for all tenants. This is overridden with `instant_subquery_spin_off_enabled` on tenants. There is an extra list of regexp that can be given per-tenant as well: `instant_subquery_spin_off_enabled_regexp`. This can be useful to roll out the feature one query at a time
- `-query-frontend.instant-subquery-spin-off.min-range-duration`. Defaults to 12h. This is overridden with `instant_subquery_spin_off_min_range_duration` on tenants
- `-query-frontend.instant-subquery-spin-off.url`. The frontend config part of the feature. Required

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -4719,7 +4719,18 @@
         },
         {
           "kind": "field",
-          "name": "instant_queries_with_subquery_spin_off",
+          "name": "max_future_query_window",
+          "required": false,
+          "desc": "Mutate incoming queries that look far into the future to only look into the future by the set duration. 0 to disable.",
+          "fieldValue": null,
+          "fieldDefaultValue": 0,
+          "fieldFlag": "query-frontend.max-future-query-window",
+          "fieldType": "duration",
+          "fieldCategory": "experimental"
+        },
+        {
+          "kind": "field",
+          "name": "instant_subquery_spin_off_enabled_regexp",
           "required": false,
           "desc": "List of regular expression patterns matching instant queries. Subqueries within those instant queries will be spun off as range queries to optimize their performance.",
           "fieldValue": null,
@@ -4729,12 +4740,12 @@
         },
         {
           "kind": "field",
-          "name": "max_future_query_window",
+          "name": "instant_subquery_spin_off_min_range_duration",
           "required": false,
-          "desc": "Mutate incoming queries that look far into the future to only look into the future by the set duration. 0 to disable.",
+          "desc": "Minimum range duration of subqueries for subquery spin-off to be enabled.",
           "fieldValue": null,
-          "fieldDefaultValue": 0,
-          "fieldFlag": "query-frontend.max-future-query-window",
+          "fieldDefaultValue": 43200000000000,
+          "fieldFlag": "query-frontend.instant-subquery-spin-off.min-range-duration",
           "fieldType": "duration",
           "fieldCategory": "experimental"
         },
@@ -6957,12 +6968,23 @@
         },
         {
           "kind": "field",
-          "name": "spin_off_instant_subqueries_to_url",
+          "name": "instant_subquery_spin_off_enabled_by_default",
           "required": false,
-          "desc": "If set, subqueries in instant queries are spun off as range queries and sent to the given URL. This parameter also requires you to set `instant_queries_with_subquery_spin_off` for the tenant.",
+          "desc": "If set, subqueries in instant queries are spun off as range queries by default. This parameter also requires you to set `-query-frontend.instant-subquery-spin-off.url=\u003curl\u003e`.",
+          "fieldValue": null,
+          "fieldDefaultValue": false,
+          "fieldFlag": "query-frontend.instant-subquery-spin-off.enable-by-default",
+          "fieldType": "boolean",
+          "fieldCategory": "experimental"
+        },
+        {
+          "kind": "field",
+          "name": "instant_subquery_spin_off_url",
+          "required": false,
+          "desc": "If set, subqueries in instant queries are spun off as range queries and sent to the given URL. This parameter also requires you to enable the feature with `-query-frontend.instant-subquery-spin-off.enable-by-default` or by setting `instant_subquery_spin_off_enabled_regexp` for the tenant.",
           "fieldValue": null,
           "fieldDefaultValue": "",
-          "fieldFlag": "query-frontend.spin-off-instant-subqueries-to-url",
+          "fieldFlag": "query-frontend.instant-subquery-spin-off.url",
           "fieldType": "string",
           "fieldCategory": "experimental"
         },
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -4730,9 +4730,20 @@
         },
         {
           "kind": "field",
+          "name": "instant_subquery_spin_off_enabled",
+          "required": false,
+          "desc": "Enable subquery spin-off for instant queries. Subqueries within those instant queries will be spun off as range queries to optimize their performance.",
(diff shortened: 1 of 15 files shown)


cmd/mimir/config-descriptor.json
```diff
@@ -6957,12 +6968,23 @@
         },
         {
           "kind": "field",
-          "name": "spin_off_instant_subqueries_to_url",
+          "name": "instant_subquery_spin_off_enabled_by_default",
```
dimitarvdimitrov, line 6971: the way we have this for other configs is that if the option is not overridden for the tenant in the runtime config, then we fall back to the global setting. But we don't include `default` in the name, it's kind of implied that the CLI option is the default and then each tenant can have a different config. See how `alertmanager_max_alerts_count` is used for example https://github.com/grafana/mimir/blob/741644ef016b91a8786e9db7255ada4476c3692a/pkg/util/validation/limits.go#L260

so what I'm suggesting is to call this `instant_subquery_spin_off_enabled` so that it can be set as `instant_subquery_spin_off_enabled: true` in the overrides yaml and as `-query-frontend.instant-subquery-spin-off.enabled`[^1] which controls the fallback for all tenants


[^1]: I also just noticed there's a mismatching spelling in the CLI arg and the YAML tag. Enable vs enabled

cmd/mimir/config-descriptor.json
```diff
@@ -6957,12 +6968,23 @@
...
+          "desc": "If set, subqueries in instant queries are spun off as range queries by default. This parameter also requires you to set `-query-frontend.instant-subquery-spin-off.url=\u003curl\u003e`.",
+          "fieldValue": null,
+          "fieldDefaultValue": false,
+          "fieldFlag": "query-frontend.instant-subquery-spin-off.enable-by-default",
+          "fieldType": "boolean",
+          "fieldCategory": "experimental"
+        },
+        {
+          "kind": "field",
+          "name": "instant_subquery_spin_off_url",
```
dimitarvdimitrov, line 6982: i haven't been involved so far, but IIRC this extra flag is just for the current testing, right? not planned to be kept long-term?

pkg/util/validation/limits.go
```diff
@@ -193,9 +193,12 @@ type Limits struct {
...
 	Prom2RangeCompat                       bool                   `yaml:"prom2_range_compat" json:"prom2_range_compat" category:"experimental"`
-	InstantQueriesWithSubquerySpinOff      []string               `yaml:"instant_queries_with_subquery_spin_off" json:"instant_queries_with_subquery_spin_off" doc:"nocli|description=List of regular expression patterns matching instant queries. Subqueries within those instant queries will be spun off as range queries to optimize their performance." category:"experimental"`
 	MaxFutureQueryWindow                   model.Duration         `yaml:"max_future_query_window" json:"max_future_query_window" category:"experimental"`
 
+	// Subquery Spin-off
+	InstantSubquerySpinOffEnabledRegexp    []string       `yaml:"instant_subquery_spin_off_enabled_regexp" json:"instant_subquery_spin_off_enabled_regexp" doc:"nocli|description=List of regular expression patterns matching instant queries. Subqueries within those instant queries will be spun off as range queries to optimize their performance." category:"experimental"`
```
dimitarvdimitrov, line 199: would you need this as a global setting as well?

dimitarvdimitrov (commented):
the changes look ok, but it's not super clear if you intend to keep these configs long-term or they are for some validaiton testing now. Maybe you can include this in the PR description.

I don't think they are of great use from an operator's perspective and I'm assuming you're adding them so you can progressively roll this out.

cmd/mimir/config-descriptor.json
julienduchesne: Done!

pkg/util/validation/limits.go
julienduchesne: The fact that it's a slice doesn't lend itself well to a CLI arg. I think the current model of being able to enable/disable per-tenant + this extra per-tenant override that isn't available as a flag is enough configurability

cmd/mimir/config-descriptor.json
julienduchesne: It's needed. The feature is about spinning off subqueries as actual range queries. We do that by sending an entirely new request to a given URL. Mimir doesn't really know the URL of the frontend service, as Mimir can run in various set-ups

julienduchesne:
> I don't think they are of great use from an operator's perspective and I'm assuming you're adding them so you can progressively roll this out.

Yes, these new flags will allow me to roll this out at large. The feature is still being tested. When things are settled, I will create a configure page, like this one for example: https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/configure-blocked-queries.md

docs/sources/mimir/configure/about-versioning.md
```diff
@@ -215,7 +215,7 @@ The following features are currently experimental:
   - Server-side write timeout for responses to active series requests (`-query-frontend.active-series-write-timeout`)
   - Caching of non-transient error responses (`-query-frontend.cache-errors`, `-query-frontend.results-cache-ttl-for-errors`)
   - Blocking HTTP requests on a per-tenant basis (configured with the `blocked_requests` limit)
-  - Spinning off (as actual range queries) subqueries from instant queries (`-query-frontend.spin-off-instant-subqueries-to-url` and the `instant_queries_with_subquery_spin_off` per-tenant limit)
+  - Spinning off (as actual range queries) subqueries from instant queries (`-query-frontend.instant-subquery-spin-off.url` and the `instant_subquery_spin_off_enabled_regexp` per-tenant limit)
```
tacole02, line 218:
```suggestion
  - Spinning off subqueries from instant queries as actual range queries (`-query-frontend.instant-subquery-spin-off.url` and the `instant_subquery_spin_off_enabled_regexp` per-tenant limit)
```

tacole02: I tried to eliminate the double parenthesis for clarity. Let me know if this works!

tacole02 (approved):
Looks good! One minor question!

cmd/mimir/config-descriptor.json
dimitarvdimitrov: oh i didn't know we go over the HTTP boundary again. Can't we just send it to `localhost`? What is this pointed to otherwise?

dimitarvdimitrov:
> > I don't think they are of great use from an operator's perspective and I'm assuming you're adding them so you can progressively roll this out.
> 
> Yes, these new flags will allow me to roll this out at large. The feature is still being tested. When things are settled, I will create a configure page, like this one for example: https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/configure-blocked-queries.md

I was more concerned about all the slices and min-durations and especially the URL. Would be nice if this feature just works and doesn't need any fiddling with flags

julienduchesne:
I will test out using localhost in our cells. I've given it more thought and I think what you're saying makes sense. We could get away with only the regexp (one flag/user limit)

PR #10699, opened 2025-02-19
StoreGateway: Parallelize Object Storage Requests While Building Index Headers

dmwilson-grafana:
#### What this PR does

- Improves the performance of `WriteBinary` by refactoring two sequential calls to object storage to run in parallel. The Store Gateway calls `syncBlocks` on startup and periodically while running (see: `blocks-storage.bucket-store.sync-interval`)  to refresh the bucket index and look for new blocks shipped by ingesters or blocks deleted by retention/ compaction. Part of this sync involves calling `WriteBinary` which updates a local index-header file using pieces of the index in object storage. The requests for these pieces of the index file can be sent in parallel if we write them to the local index-header in the correct order.

- Adds a new benchmark, `WriteBinary_DelayedBucket`, which uses a mock bucket with a built-in delay to simulate calls to remote providers. 

```text
pkg: github.com/grafana/mimir/pkg/storegateway/indexheader
                                            │   old.txt    │               new.txt               │
                                            │    sec/op    │   sec/op     vs base                │
WriteBinary-8                                 9.156m ±  1%   9.994m ± 5%   +9.15% (p=0.000 n=10)
WriteBinary_DelayedBucket/max_delay_10_ms-8   66.72m ±  4%   58.33m ± 4%  -12.57% (p=0.000 n=10)
WriteBinary_DelayedBucket/max_delay_20_ms-8   120.8m ±  8%   105.4m ± 4%  -12.72% (p=0.000 n=10)
WriteBinary_DelayedBucket/max_delay_50_ms-8   285.8m ± 11%   224.3m ± 9%  -21.53% (p=0.000 n=10)
```

#8166

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/indexheader/binary_reader.go
+++ b/pkg/storegateway/indexheader/binary_reader.go
@@ -9,7 +9,6 @@ import (
 	"bufio"
 	"context"
 	"encoding/binary"
-	"golang.org/x/sync/errgroup"
 	"hash"
 	"hash/crc32"
 	"io"
@@ -25,6 +24,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb/fileutil"
 	"github.com/prometheus/prometheus/tsdb/index"
 	"github.com/thanos-io/objstore"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/grafana/mimir/pkg/storage/tsdb/block"
 )
--- a/pkg/storegateway/indexheader/binary_reader.go
+++ b/pkg/storegateway/indexheader/binary_reader.go
@@ -100,19 +100,19 @@ func WriteBinary(ctx context.Context, bkt objstore.BucketReader, id ulid.ULID, f
 	var g errgroup.Group
 	var sym, tbl io.ReadCloser
 	g.Go(func() (err error) {
-		sym, err = ir.bkt.GetRange(ir.ctx, ir.path, int64(ir.toc.Symbols), int64(ir.toc.Series-ir.toc.Symbols))
+		sym, err = ir.GetSymbolsReader()
 		if err != nil {
-			return errors.Wrapf(err, "get symbols from object storage of %s", ir.path)
+			return err
 		}
-		return
+		return nil
 	})
 
 	g.Go(func() (err error) {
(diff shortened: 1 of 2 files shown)


pkg/storegateway/indexheader/binary_reader.go
```diff
@@ -93,16 +95,52 @@ func WriteBinary(ctx context.Context, bkt objstore.BucketReader, id ulid.ULID, f
...
+
+	if _, err := io.CopyBuffer(bw.SymbolsWriter(), sym, buf); err != nil {
+		return errors.Wrap(err, "copy posting offsets")
 	}
 
 	if err := bw.f.Flush(); err != nil {
 		return errors.Wrap(err, "flush")
 	}
 
-	if err := ir.CopyPostingsOffsets(bw.PostingOffsetsWriter(), buf); err != nil {
```
56quarters, line 104: `ir.CopyPostingOffsets()` and `ir.CopySymbols()` are unused now, right? They should be modified to accept a reader and then used here. That would allow this new code to follow the same pattern as the original code but keep the requests to object storage in parallel.

56quarters (commented):
Thanks for finding this improvement!

pkg/storegateway/indexheader/binary_reader.go
dmwilson-grafana: Newest commit updates `ir.CopyPostingOffsets()` and `ir.CopySymbols()` to use `io.ReadWriter`, we write into a buffer for each, and then into the `binaryWriter`. Is this what you were thinking?

dmwilson-grafana:
> I'm not 100% convinced this will give us meaningful speedups for larger index headers. The GetRange readers we get buffer only about 64K of data. For for larger index headers where delay actually matters (where downloading the index header is in the seconds/minutes) I think this will perform effectively the same.

You're right. With this change we only get an improvement that's roughly the round-trip time to object storage. The original benchmarks used a 6MB index, when I tested w. a 2.2GB block (see below) we see about the same improvement in ms and almost no improvement in %'age terms. 


```
                                            │ regular.txt  │          parallel.txt          │
                                            │    sec/op    │    sec/op     vs base          │
WriteBinary_DelayedBucket/max_delay_20_ms-8   483.8m ± 53%   470.7m ± 37%  ~ (p=0.853 n=10)
```

PR #10702, opened 2025-02-20
Ingester: Ensure cost attribution tracker is updated when configuration changes

ying-jeanne:

This bug was discovered during deployment. The issue occurs because while we compare configuration differences, we never update the tracker in the ingester. As a result, the tracker remains in conflict mode, continuously exporting `cortex_ingester_active_series_loading`. This fix ensures proper config consolidation and resolves the issue. Additionally, the unit test included in this PR would fail without the fixed code, confirming the correctness of the fix.

The test showed on dev with fix, similar to usage group update, after ttl, the active series sum is correct
Screenshot 2025-02-20 at 16 16 38

without fix, the active series are just gone for good.
Screenshot 2025-02-20 at 17 03 52

#### Checklist

- [x] Tests updated.
- na Documentation added.
- na [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Grafana Mimir
 
-* [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269
+* [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269 #10702
 * [CHANGE] Querier: pass context to queryable `IsApplicable` hook. #10451
 * [CHANGE] Distributor: OTLP and push handler replace all non-UTF8 characters with the unicode replacement character `\uFFFD` in error messages before propagating them. #10236
 * [CHANGE] Querier: pass query matchers to queryable `IsApplicable` hook. #10256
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
   * `go_cpu_classes_gc_total_cpu_seconds_total`
   * `go_cpu_classes_total_cpu_seconds_total`
   * `go_cpu_classes_idle_cpu_seconds_total`
+* [BUGFIX] Ingester: Ensure cost attribution tracker is updated when configuration changes. #10702
```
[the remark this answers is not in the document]
ying-jeanne: addressed in b6ded04c20a240ad403ae12a23c5b3d3d4bfdd5b, my appologies for the force push, it was a mistake since I was rebasing on r329 for testing

pkg/ingester/activeseries/active_series.go
```diff
@@ -114,14 +114,14 @@ func (c *ActiveSeries) ConfigDiffers(ctCfg asmodel.CustomTrackersConfig, caCfg *
...
+func (c *ActiveSeries) ReloadMatchersAndTrackers(asm *asmodel.Matchers, cat *costattribution.ActiveSeriesTracker, now time.Time) {
 	c.configMutex.Lock()
 	defer c.configMutex.Unlock()
-
 	for i := 0; i < numStripes; i++ {
-		c.stripes[i].reinitialize(asm, &c.deleted, c.cat)
+		c.stripes[i].reinitialize(asm, &c.deleted, cat)
 	}
 	c.matchers = asm
+	c.cat = cat
```
56quarters, line 124: nit: Can you add a note on the `cat` field of this struct to indicate it is protected by the `configMutex`?

56quarters (approved):
LGTM with one small suggestion

pkg/ingester/activeseries/active_series.go
ying-jeanne, line 124: Here is the bug, it occurred because c.cat was compared with the current cost attribution tracker when the config changed, but it was never updated beforehand. As a result, active_series got stuck during the active series loading process.

colega: I'd say that instead of the comment, it should be moved one line to be covered by the _mutex hat_.

I agree that with current arrangement it's not clear whether it's covered by the `configMutex`

colega (approved):

pkg/ingester/activeseries/active_series.go
ying-jeanne: added comments + move cat cerca de mutex

PR #10703, opened 2025-02-20
frontend: Simplify configuration for subquery spin-off

julienduchesne:
## Testing of the feature so far

Testing has shown nice results for the feature. In our main Mimir cell, the overall runtime of all instant queries containing the "subquery pattern" ([<range>:(<step>)?]) has gone down by half, and the feature allows for queries further into the past by sharding by time

From tests I've run, I also have not seen a difference between spinning off queries to localhost or the whole frontend fleet. Since we're spinning off one query at a time, it makes sense

## This PR

Given the results, I think we can start considering enabling this feature at large:

- Add the `-query-frontend.instant-queries-with-subquery-spin-off` flag, linked to the user limit. That way we, can set a global cell config for this
- Remove the URL flag. Let's make things simpler and always spin-off range queries to localhost. **I had to pass the HTTP listen port from the server config, into the middlewares, for that**
- Added an integration test to make sure the http listen port is passed correctly

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


CHANGELOG.md
```diff
@@ -46,7 +46,7 @@
...
 * [ENHANCEMENT] Ruler: Adds support for filtering results from rule status endpoint by `file[]`, `rule_group[]` and `rule_name[]`. #10589
-* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries, so that they benefit from query acceleration techniques such as sharding, splitting and caching. To enable this, set the `-query-frontend.spin-off-instant-subqueries-to-url=<url>` option on the frontend and the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621
+* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries, so that they benefit from query acceleration techniques such as sharding, splitting and caching. To enable this, set the `-query-frontend.instant-queries-with-subquery-spin-off=<comma separated list>` option on the frontend or the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621 #10703
```
tacole02, line 49:
```suggestion
* [ENHANCEMENT] Query-frontend: Add option to "spin off" subqueries as actual range queries, so that they benefit from query acceleration techniques such as sharding, splitting, and caching. To enable this feature, set the `-query-frontend.instant-queries-with-subquery-spin-off=<comma separated list>` option on the frontend or the `instant_queries_with_subquery_spin_off` per-tenant override with regular expressions matching the queries to enable. #10460 #10603 #10621 #10703
```

tacole02 (approved):

integration/query_frontend_test.go
```diff
@@ -417,8 +433,14 @@ func runQueryFrontendTest(t *testing.T, cfg queryFrontendTestConfig) {
 		expectedQueriesCount++
 		expectedIngesterQueriesCount++
 	}
+	expectedFrontendQueriesCount := expectedQueriesCount
+	if cfg.testSubquerySpinOff {
+		expectedFrontendQueriesCount += 2 // One for the instant query and one for the spun-off subquery.
```
NickAnge, line 438:
```suggestion
		expectedFrontendQueriesCount += 2 // One for the instant query and one for the spin-off subquery.
```

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -379,23 +379,15 @@ func newQueryMiddlewares(
...
-			queryInstantMiddleware = append(
-				queryInstantMiddleware,
-				newInstrumentMiddleware("spin_off_subqueries", metrics),
-				newSpinOffSubqueriesMiddleware(limits, log, engine, spinOffQueryHandler, registerer, defaultStepFunc),
-			)
-		}
-	}
+	// Loopback subqueries to localhost. Only those in instant queries are supported.
+	// Add the retry middleware to the spin-off query handler.
+	// Spun-off queries are terminated in that handler (they don't call "next" so the retry middleware has to be added here).
```
NickAnge, line 384: Spin-off, I think.

integration/query_frontend_test.go
julienduchesne: I meant spun as the past of spin. I'm french so who knows if that's wrong

NickAnge: No its correct 😅 , I thought it was a typo

pkg/frontend/querymiddleware/roundtrip.go
NickAnge: Same as above

julienduchesne:
@dimitarvdimitrov This is what I had tried previously to avoid doing a network call: https://github.com/grafana/mimir/commit/ac1546c463e4fb578ac7e037f3b4004cf70b7541

Passing the handler by context is the only way I found because:
- the tripperware/middlewares doesn't have access to the full handler. The final roundtripper is injected from the top-level
- many parts of the request processing (like query stats) happens before going into the middlewares

So, it works but I thought that having a specified URL was cleaner, more predictable. Essentially, the options I saw were:
- Passing http handlers around through the request context. A bit messy and hard to follow but no network call. Works in all cases
- Using localhost under the covers (like the PR is currently): No extra arg, but with an additional hidden gotcha where it uses HTTP and will fail in some cases (HTTPs or auth for example)
- Explicit passing of a URL as an argument: Easy to reason about but there's an extra arg. One benefit is that the middleware doesn't leak its concerns. It's 100% contained and easy to take out (both the other options need something else outside of the middleware)


Given these choices and after some thought, I think that the easiest way to go is to keep the URL arg. Also, I think I can close this PR and put this on ice. I will enable and test the feature more widely and if it keeps being unproblematic, I can make the `...-url` arg the main one that also enables the feature on the whole cell. What do you think?

dimitarvdimitrov:
in case anyone is following this PR: I think I found a way to reuse the current middleware architecture without passing the roundtrippers through the context. I sketched it in https://github.com/grafana/mimir/commit/d8ebcddff4749dea88fc43262beca62241b1090a

julienduchesne:
Superseded by https://github.com/grafana/mimir/pull/10742

Issue #10688, opened 2025-02-19
Flaky `TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp`


charleskorn:
[Logs](https://github.com/grafana/mimir/actions/runs/13404725165/job/37443477785#step:8:82):

```
--- FAIL: TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp (0.30s)
    ha_tracker_test.go:1429: 
        	Error Trace:	/__w/mimir/mimir/pkg/distributor/ha_tracker_test.go:1429
        	Error:      	An error is expected but got nil.
        	Test:       	TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp
2025/02/19 03:56:29 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/02/19 03:56:29 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/02/19 03:56:29 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/02/19 03:56:29 label __name__ is overwritten. Check if Prometheus reserved labels are used.
2025/02/19 03:56:29 label __name__ is overwritten. Check if Prometheus reserved labels are used.
level=info msg="server listening on addresses" http=127.0.0.1:43431 grpc=127.0.0.1:46853
[...]

PR #10707, opened 2025-02-20
test: verify that the cache has the right value before proceeding to …

NickAnge:
#### What this PR does

his test is not related to `memberlist`. After running it with a -1000 value, it was failing quite often. In this PR, we add a check with a delay of up to 2 seconds to allow the cache of ha_tracker t2 to be populated before proceeding to the next step.

I ran  to verify ⬇️ 
```
go test -v -failfast ./pkg/distributor/ -run TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp -count=1000
```

Should fix https://github.com/grafana/mimir/issues/10688

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/ha_tracker_test.go
+++ b/pkg/distributor/ha_tracker_test.go
@@ -1419,6 +1419,8 @@ func TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp(t *testing.T) {
 		require.Zero(t, info.electedLastSeenTimestamp)
 	}
 
+	// We add a polling check for t2, up to 2 seconds, to allow the cache to be populated with the correct value (secondReplica) before proceeding
+	// with the checkReplica for firstReplica.
 	checkReplicaTimestamp(t, 2*time.Second, t2, userID, cluster, secondReplica, secondReplicaReceivedAtT2, secondReplicaReceivedAtT2)
 
 	// Continuing the test, say both t1 and t2 receive request from "first" replica now.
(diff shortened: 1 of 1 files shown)


pkg/distributor/ha_tracker_test.go
```diff
@@ -1419,6 +1419,8 @@ func TestHATrackerChangeInElectedReplicaClearsLastSeenTimestamp(t *testing.T) {
 		require.Zero(t, info.electedLastSeenTimestamp)
 	}
 
+	checkReplicaTimestamp(t, 2*time.Second, t2, userID, cluster, secondReplica, secondReplicaReceivedAtT2, secondReplicaReceivedAtT2)
```
narqo, line 1422: (nit) I would add a comment outlining what is happening here and why. Without previous context, it's hard to reason if this is only an assertion, or it also guards any synchronisation.

narqo (approved):
🔥 Left a nitpicking comment, but looks good overall

pkg/distributor/ha_tracker_test.go
NickAnge: Great, suggestion. Thank you. Actually I always debate about comments (with myself 😅 ). But in this case makes sense . I will add it

NickAnge: Hey @narqo. I added it [d225d9c](https://github.com/grafana/mimir/pull/10707/commits/d225d9c495a333bbc9e7df60728a20398e3ae61b). Could you please check again, and merge, cause I do not have permissions 😅 . Whenever you have time

NickAnge:
Actually this might not be needed if we proceed with the  following change https://github.com/grafana/mimir/pull/10714. Lets not merge this yet. I am switching to draft

NickAnge:
This will not be merged, we are gonna go for the sync pr

PR #10714, opened 2025-02-21
feat: synchronize cache update with KV store modification.

NickAnge:
#### What this PR does

With this change we make the `updateCache` synchronous with the kvStore change. Since the KVstore is the source of truth , regarding the correct ha_tracker entry  it makes sense to update the cache immediately.

##### Would it have any performance impact ?

This change will add a few map accesses for Replicas that have an actual change. **Not for every Replica**.  This operation was already happening from [processKVStoreEntry](https://github.com/grafana/mimir/blob/c86963a119123161ca9f1deae538f8ac281b3a8a/pkg/distributor/ha_tracker.go#L380-L379), so we just do the change earlier.   So no impact is expected.

Thanks @dimitarvdimitrov  for pointing this out.

Fix
- https://github.com/grafana/mimir/issues/10350#issuecomment-2672294698
- https://github.com/grafana/mimir/issues/10688#issuecomment-2669235048

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 * [BUGFIX] Distributor: Fix panics in `DurationWithJitter` util functions when computed variance is zero. #10507
 * [BUGFIX] Ingester: Fixed a race condition in the `PostingsForMatchers` cache that may have infrequently returned expired cached postings. #10500
 * [BUGFIX] Distributor: Report partially converted OTLP requests with status 400 Bad Request. #10588
+* [ENHANCEMENT] Distributor: synchronize cache update with KV store modification for ha_tracker. #10714
 
 ### Mixin
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,6 @@
 * [BUGFIX] Distributor: Fix panics in `DurationWithJitter` util functions when computed variance is zero. #10507
 * [BUGFIX] Ingester: Fixed a race condition in the `PostingsForMatchers` cache that may have infrequently returned expired cached postings. #10500
 * [BUGFIX] Distributor: Report partially converted OTLP requests with status 400 Bad Request. #10588
-* [ENHANCEMENT] Distributor: synchronize cache update with KV store modification for ha_tracker. #10714
 
 ### Mixin
 
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -668,9 +668,7 @@ func (h *defaultHaTracker) updateKVStore(ctx context.Context, userID, cluster, r
 	// If cache is currently empty, add the data we either stored or received from KVStore
 	if err == nil && desc != nil {
 		h.electedLock.Lock()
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 * [BUGFIX] Distributor: Fix panics in `DurationWithJitter` util functions when computed variance is zero. #10507
 * [BUGFIX] Ingester: Fixed a race condition in the `PostingsForMatchers` cache that may have infrequently returned expired cached postings. #10500
 * [BUGFIX] Distributor: Report partially converted OTLP requests with status 400 Bad Request. #10588
+* [ENHANCEMENT] Distributor: synchronize cache update with KV store modification for ha_tracker. #10714
```
NickAnge, line 78: I think its `ENHANCEMENT`, as it  updates the cache earlier with the agreed value. But let me know what you think

pkg/distributor/ha_tracker_test.go
```diff
@@ -318,9 +318,6 @@ func TestHaTrackerWithMemberList(t *testing.T) {
 	// Update KVStore - this should elect replica 2.
 	tracker.updateKVStoreAll(context.Background(), now)
 
-	// Evaluate up to 10 seconds to verify whether the tracker’s cache replica has been updated to r2.
-	checkReplicaTimestamp(t, 10*time.Second, tracker, "user", cluster, replica2, now, now)
```
dimitarvdimitrov, line 322: this used to do other assertions too, not just the replica. Maybe we can have a synchronous version of it or just pass a very small deadline (although that comes with other race conditions)

pkg/distributor/ha_tracker.go
```diff
@@ -668,9 +668,7 @@ func (h *defaultHaTracker) updateKVStore(ctx context.Context, userID, cluster, r
 	// If cache is currently empty, add the data we either stored or received from KVStore
```
dimitarvdimitrov, line 668: this comment is out of date now

CHANGELOG.md
dimitarvdimitrov: would users/operators notice a difference with this change?

NickAnge: They shouldn't notice a change, no. Should we remove the changelog?

NickAnge: I removed it, let me know what you think

pkg/distributor/ha_tracker_test.go
NickAnge: You are correct. I re-added it with a very short deadline. I do not think it should cause any issues, as the updateKVStoreAll should update all checked values beforehand synchronously.

CHANGELOG.md
dimitarvdimitrov: yeah, the changelog is for things that will have some impact on the operators or users of Mimir. In this case we're mostly doing it for non-flaky tests, so a changelog entry isn't necessary

dimitarvdimitrov (approved):
LGTM, thanks!

PR #10715, opened 2025-02-21
store-gateway: reduce garbage from index header

bboreham:
#### What this PR does

This one call to io.ReadAll was generating 6% of all garbage in some production store-gateways recently.

~Since gzip puts the size at the end of the compressed data, we can use that size to pre-allocate the buffer and avoid resizing.
Code is adapted from some that used to be inside gRPC, from https://github.com/grpc/grpc-go/pull/3048~

We can use `util.ParseProtoReader` instead, which goes via `bytes.Buffer.ReadFrom` which is much more efficient.  And save a few lines due to the reuse.
    
~I left the bucket path still reading into memory before calling `ParseProtoReader`. Maybe that's unnecessary.~

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

---

> [!NOTE]
> Stream sparse index-header loading from disk/bucket and decode via util.ParseProtoReader with context, reducing allocations and updating call sites/tests.
> 
> - **Index header readers**:
>   - Stream sparse header IO instead of buffering: replace `tryReadBucketSparseHeader` with `tryOpenBucketSparseHeader` (returns `io.ReadCloser`), use `os.Open` and pass readers through.
>   - Replace manual gzip + `io.ReadAll` with `util.ParseProtoReader` for decoding (context-aware, size-bounded); add cancellation checks after bucket reads.
>   - Write sparse header using `atomicfs.CreateFile` with a reader; ensure proper `Close` via `runutil.CloseWithLogOnErr`.
>   - Change `loadFromSparseIndexHeader` to accept `context.Context` and `io.Reader`; update all call sites (disk, bucket, lazy download path).
> - **Tests**:
>   - Update to open files and pass readers; adjust decoding to use reader + context; verify bucket fetch and local persistence.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b24e4ab3adc4ce81c36e01773a2d40806c75455. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dimitarvdimitrov (commented):
this looks good in general. Can't we change `util.ParseProtoReader` to do this estimation? It currently takes `expectedSize` I guess we can take `-1` and let it try to estimate the size based on the method here. Do you want to try that or should I?

jhalterman:
Having a separate function for this seems fine, but could it live in the util package for better reuse?

bboreham:
Sorry I think I was too cryptic; I looked at possible utility functions but they don't do exactly the same operation.

In order to use `ParseProtoReader` instead of the current code we would have to come up with a maximum size for `sparseData`.  Do you have any suggestions how I would go about that?

Is it ok that buffers would be pooled?

dimitarvdimitrov:
> In order to use `ParseProtoReader` instead of the current code we would have to come up with a maximum size for `sparseData`. Do you have any suggestions how I would go about that?

can we just use `math.MaxInt` effectively removing the limit?



> Is it ok that buffers would be pooled?

I think so. We use stock protobuf unmarshalling, so we don't retain pointers to the decompressed buffer.

bboreham:
Updated to use `ParseProtoReader`, choosing 2GB as the max size.  

I left out reading the size from the gzipped data, as that would have to go inside `ParseProtoReader`, and I'm uncertain what degree of benefit we'd get.

Since I first made this PR, an alternative path has been added reading from a bucket.  I changed that to read via `bytes.buffer.ReadFrom` and then wrap with a `bytes.Reader`, but this could simply pass on the `Reader`, if it's ok to catch failures mid-decompression.

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -239,18 +242,18 @@ func tryReadBucketSparseHeader(ctx context.Context, logger log.Logger, bkt objst
 
 	level.Info(logger).Log("msg", "downloaded sparse index-header from bucket")
 
-	return data, nil
+	return buf.Bytes(), nil
```
dimitarvdimitrov: i think it's safe to return an io.ReadCloser here. Buffering the whole result here might make a difference only in terms of what is logged if reading from the bucket fails mid-way, but I don't think that's a big deal

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -268,25 +271,16 @@ func (r *StreamBinaryReader) loadFromSparseIndexHeader(logger log.Logger, sparse
...
-
-	gzipped := bytes.NewReader(sparseData)
-	gzipReader, err := gzip.NewReader(gzipped)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create sparse index-header gzip reader: %w", err)
-	}
-	defer runutil.CloseWithLogOnErr(logger, gzipReader, "close sparse index-header gzip reader")
-
-	sparseData, err = io.ReadAll(gzipReader)
+func decodeSparseData(ctx context.Context, _ log.Logger, sparseData io.Reader) (*indexheaderpb.Sparse, error) {
```
dimitarvdimitrov: why not remove the logger parameter?

pkg/storage/indexheader/stream_binary_reader.go
```diff
@@ -268,25 +271,16 @@ func (r *StreamBinaryReader) loadFromSparseIndexHeader(logger log.Logger, sparse
...
+	sparseHeaders := indexheaderpb.Sparse{}
+	// ParseProtoReader requires a max size; we supply a very broad estimate.
+	// We could read the size from last 4 bytes of the file (per gzip spec), but it's probably not worth it.
+	const maxSize = 1 << 31 // 2GB
```
dimitarvdimitrov:
```suggestion
	const maxSize = 1 << 31 // 2GiB
```

dimitarvdimitrov (approved):
nice one, only a few nitpicks

PR #10716, opened 2025-02-21
chore: bump golang versions (#10708)

aldernero:
* chore: bump golang versions

* Update build image version to pr10708-6841577474

backport #10708 to `r330`


narqo (approved):
🔥 Works for me. Although, do we really have to backport this to this weekly release? 

GEM uses its own toolchain, and its weeklies have their own set of Go-version constraints. So this change shouldn't make any difference there (_I could be wrong about that_).

I'd _maybe_ backported it to `release-2.15`, if we're aiming at shipping 2.15.1. But _maybe_ this can be just a part of the future 2.16?

PR #10738, opened 2025-02-25
ruler: add /ruler/tenants service API

narqo:
#### What this PR does

This PR adds a new `/ruler/tenants` endpoint to list tenants, the `ruler` service discovered in its storage. The idea here is to provide an external operator-component a way to observe which tenants have rules, without asking Mimir to retrieve the actual rules.

The API endpoint is complementary to the existing `/store-gateway/tenants`, `/compactor/tenants`, and `/ingesters/tenants`.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -7,10 +7,12 @@ package ruler
 
 import (
 	"context"
+	_ "embed" // Used to embed html template
 	"encoding/base64"
 	"flag"
 	"fmt"
 	"hash/fnv"
+	"html/template"
 	"net/http"
 	"net/url"
 	"path/filepath"
@@ -21,6 +23,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/gorilla/mux"
 	"github.com/grafana/dskit/concurrency"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/grpcclient"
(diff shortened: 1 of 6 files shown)


pkg/ruler/ruler.go
```diff
@@ -1339,6 +1342,81 @@ func (r *Ruler) ListAllRules(w http.ResponseWriter, req *http.Request) {
...
+		level.Error(logger).Log("msg", "failed to list user rule groups", "user", userID, "err", err)
+		http.Error(w, fmt.Sprintf("failed to list user rule groups: %s", err.Error()), http.StatusInternalServerError)
+		return
+	}
+
+	groups := make([]ruleGroupData, 0, len(rg))
+	for _, g := range rg {
+		groups = append(groups, ruleGroupData{
+			Name:      g.GetName(),
+			Namespace: g.GetNamespace(),
```
colega, line 1409: How about we include `RuleCount: len(g.Rules)` here as well?

colega (commented):
Can you add a couple of tests? Especially for requests sending JSON.

It's not clear from the code that we're creating a contract on the contents of the JSON returned.

pkg/ruler/ruler.go
narqo: I actually wanted to do the same initially, but `len(g.Rules)` doesn't have any rules until we ask the `Ruler` to go to storage, and to populate them for every namespace and group — via the `RuleStore.LoadRuleGroups` — this is confusing, yes. I'd skip doing that (at least for now).

narqo:
> Can you add a couple of tests? Especially for requests sending JSON.

Sure, good point. We seem to not see these service APIs as something we guarantee (at I haven't noticed any similar tests in other places), but I don't see why the behaviour around how limits work with this API shouldn't be covered.

pkg/ruler/ruler.go
colega: I see, okay!

colega (approved):

PR #10742, opened 2025-02-25
query frontend: Subquery spin-off without HTTP call

julienduchesne:
#### What this PR does

TODO

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


integration/query_frontend_test.go
```diff
@@ -43,6 +43,7 @@ type queryFrontendTestConfig struct {
 	setup                       func(t *testing.T, s *e2e.Scenario) (configFile string, flags map[string]string)
 	withHistograms              bool
 	shardActiveSeriesQueries    bool
+	testSubquerySpinOff         bool
```
dimitarvdimitrov, line 46: why can't we always test this?

pkg/util/validation/limits.go
```diff
@@ -398,6 +398,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.EnabledPromQLExperimentalFunctions, "query-frontend.enabled-promql-experimental-functions", "Enable certain experimental PromQL functions, which are subject to being changed or removed at any time, on a per-tenant basis. Defaults to empty which means all experimental functions are disabled. Set to 'all' to enable all experimental functions.")
 	f.BoolVar(&l.Prom2RangeCompat, "query-frontend.prom2-range-compat", false, "Rewrite queries using the same range selector and resolution [X:X] which don't work in Prometheus 3.0 to a nearly identical form that works with Prometheus 3.0 semantics")
 	f.Var(&l.MaxFutureQueryWindow, maxFutureQueryWindowFlag, "Mutate incoming queries that look far into the future to only look into the future by the set duration. 0 to disable.")
+	f.Var(&l.InstantQueriesWithSubquerySpinOff, "query-frontend.instant-queries-with-subquery-spin-off", "List of regular expression patterns matching instant queries. Subqueries within those instant queries will be spun off as range queries to optimize their performance.")
```
dimitarvdimitrov, line 401: i think this should be a `f.StringVar` now

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -419,11 +399,15 @@ func newQueryMiddlewares(
 			resultsCacheEnabledByOption,
 			log,
 			registerer,
-		))
+		)
+	}
+
+	// Inject the extra middlewares provided by the user before the query pruning and query sharding middleware.
+	if len(cfg.ExtraRangeQueryMiddlewares) > 0 {
```
dimitarvdimitrov, line 406: why did this move?

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -432,14 +416,19 @@ func newQueryMiddlewares(
 		prom2CompatMiddleware,
 	)
 
+	rangeQuerySpinOffMiddlewares := append([]MetricsQueryMiddleware{
+		splitAndCacheMiddleware,
+	}, cfg.ExtraRangeQueryMiddlewares...)
```
dimitarvdimitrov, line 421: my thinking was that the split and cache will only be a detour into the split-and-cache if we have to do it. After the split-and-cache the request will still go through the rest of the middlewares all instant queries go through (cardinality, sharding, pruning retry, experimental features)

why do we need the extra middlewares here again?

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -95,19 +91,27 @@ func newSpinOffSubqueriesMiddleware(
 	limits Limits,
 	logger log.Logger,
 	engine *promql.Engine,
-	rangeHandler MetricsQueryHandler,
 	registerer prometheus.Registerer,
+	rangeMiddlewares []MetricsQueryMiddleware,
```
dimitarvdimitrov, line 95: let's take a single `MetricsQueryMiddleware` and use `middleware.Merge` (but again, i'm not sure we even need to merge anything if we want to add only the split-and-cache middleware)

dimitarvdimitrov (commented):
nice one, thanks for cleaning up my POC

pkg/util/validation/limits.go
julienduchesne: hmm, all of the other `flagext.StringSliceCSV` are `Var`. StringVar can only be assigned to a string

integration/query_frontend_test.go
julienduchesne: We can. Done

pkg/util/validation/limits.go
dimitarvdimitrov: i was mislead by the type of the field in the config descriptor

dimitarvdimitrov (approved):
nice one, LGTM

julienduchesne:
Everything docs related is exactly the same as in: https://github.com/grafana/mimir/pull/10703, so merging without Taylor's review

PR #10744, opened 2025-02-26
MQE: add support for `count_values`

charleskorn:
#### What this PR does

This PR adds support for `count_values` in MQE.

Benchmark results are mixed. MQE's implementation always consumes less memory than Prometheus' engine in our benchmarks, but in some cases runs slower, and in others runs faster. MQE is up to 36% faster for the common case where many series have only a handful of values, but up to 20% slower for the uncommon case where every sample has a different value.

The `count_values('value', h_X)` cases exercise the scenario where every sample has a different value, resulting in `X × 100` output series. The `count_values('value', h_X * 0)` cases exercise the scenario where every sample has the same value, resulting in 1 output series.

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                                                                       │ Prometheus  │               Mimir                │
                                                                       │   sec/op    │   sec/op     vs base               │
Query/count_values('value',_h_1),_range_query_with_100_steps-10          833.8µ ± 1%   733.8µ ± 3%  -11.99% (p=0.002 n=6)
Query/count_values('value',_h_100),_range_query_with_100_steps-10        70.70m ± 1%   84.77m ± 4%  +19.91% (p=0.002 n=6)
Query/count_values('value',_h_2000),_range_query_with_100_steps-10        2.300 ± 2%    2.574 ± 2%  +11.92% (p=0.002 n=6)
Query/count_values('value',_h_1_*_0),_range_query_with_100_steps-10      384.5µ ± 2%   266.9µ ± 1%  -30.59% (p=0.002 n=6)
Query/count_values('value',_h_100_*_0),_range_query_with_100_steps-10    19.00m ± 5%   13.54m ± 3%  -28.71% (p=0.002 n=6)
Query/count_values('value',_h_2000_*_0),_range_query_with_100_steps-10   413.2m ± 1%   264.3m ± 1%  -36.04% (p=0.002 n=6)
geomean                                                                  27.25m        23.13m       -15.12%

                                                                       │  Prometheus   │                Mimir                 │
                                                                       │       B       │       B        vs base               │
Query/count_values('value',_h_1),_range_query_with_100_steps-10           64.55Mi ± 1%   65.10Mi ±  1%        ~ (p=0.074 n=6)
Query/count_values('value',_h_100),_range_query_with_100_steps-10         298.2Mi ± 2%   271.9Mi ± 11%   -8.81% (p=0.002 n=6)
Query/count_values('value',_h_2000),_range_query_with_100_steps-10        2.579Gi ± 1%   1.834Gi ±  1%  -28.88% (p=0.002 n=6)
Query/count_values('value',_h_1_*_0),_range_query_with_100_steps-10       63.30Mi ± 1%   63.24Mi ±  1%        ~ (p=0.729 n=6)
Query/count_values('value',_h_100_*_0),_range_query_with_100_steps-10     67.12Mi ± 1%   61.92Mi ±  2%   -7.74% (p=0.002 n=6)
Query/count_values('value',_h_2000_*_0),_range_query_with_100_steps-10   149.60Mi ± 3%   66.44Mi ±  1%  -55.59% (p=0.002 n=6)
geomean                                                                   178.5Mi        143.3Mi        -19.72%
```

The difference in performance for the uncommon case is because Prometheus' engine uses a hash of the  series labels to identify output series, which runs the risk of hash collisions, while MQE uses a string representation of the series labels to identify output series. In the case where there are many output series, this means MQE must allocate many strings, which causes the performance impact we see here.

Given `count_values` is a relatively uncommonly used PromQL feature, the performance of the common case is significantly better than Prometheus' engine, and MQE's implementation avoids possible hash collisions, I'm OK with this performance regression.

#10067

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/aggregations/count_values.go
+++ b/pkg/streamingpromql/operators/aggregations/count_values.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"slices"
 	"strconv"
+	"sync"
 
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
@@ -66,6 +67,12 @@ type countValuesSeries struct {
 	labels           labels.Labels
 	outputPointCount int
 	count            []int // One entry per timestamp.
+}
+
+var countValuesSeriesPool = sync.Pool{
+	New: func() interface{} {
+		return &countValuesSeries{}
+	},
 }
 
 func (c *CountValues) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
@@ -126,6 +133,8 @@ func (c *CountValues) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
 		c.series = append(c.series, points)
 
 		types.IntSlicePool.Put(s.count, c.MemoryConsumptionTracker)
+		s.count = nil
+		countValuesSeriesPool.Put(s)
 	}
 
 	return outputMetadata, nil
@@ -160,9 +169,9 @@ func (c *CountValues) incrementCount(seriesLabels labels.Labels, t int64, value 
 	series, exists := accumulator[string(c.labelsBytesBuffer)] // Important: don't extract the string(...) call here - passing it directly allows us to avoid allocating it.
 
 	if !exists {
-		series = &countValuesSeries{
-			labels: l,
-		}
+		series = countValuesSeriesPool.Get().(*countValuesSeries)
+		series.labels = l
+		series.outputPointCount = 0
 
 		var err error
 		series.count, err = types.IntSlicePool.Get(c.TimeRange.StepCount, c.MemoryConsumptionTracker)
--- a/pkg/streamingpromql/operators/aggregations/count_values.go
+++ b/pkg/streamingpromql/operators/aggregations/count_values.go
@@ -1,0 +1,227 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/prometheus/prometheus/blob/main/promql/engine.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Prometheus Authors
+
+package aggregations
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strconv"
+
+	"github.com/prometheus/common/model"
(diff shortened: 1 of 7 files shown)


pkg/streamingpromql/operators/aggregations/count_values.go
```diff
@@ -0,0 +1,247 @@
...
+	} else {
+		c.labelsBuilder.Keep(c.Grouping...)
+	}
+
+	c.labelsBuilder.Set(c.resolvedLabelName, value)
+
+	return c.labelsBuilder.Labels()
+}
+
+func (s *countValuesSeries) ToPoints(memoryConsumptionTracker *limiting.MemoryConsumptionTracker, timeRange types.QueryTimeRange) ([]promql.FPoint, error) {
```
lamida, line 211: Is there any reason to make this method public?

pkg/streamingpromql/operators/aggregations/count_values.go
```diff
@@ -0,0 +1,247 @@
...
+	accumulator := map[string]*countValuesSeries{}
+
+	for _, s := range innerMetadata {
+		data, err := c.Inner.NextSeries(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, p := range data.Floats {
+			if err := c.incrementCount(s.Labels, p.T, c.formatFloatValue(p.F), accumulator); err != nil {
```
lamida, line 108: Would caching `c.formatFloatValue(p.F)` going to be beneficial? I will imagine if we have many same values across the result, this will be formatted multiple times for that same values.

charleskorn: The benchmarks show this implementation is already significantly faster than Prometheus' engine for the case where there are many series with the same value, and caching like this would make the worst case (many series with different values) slower, so I'm tempted to leave things as-is.

pkg/streamingpromql/operators/aggregations/count_values.go
charleskorn: Good catch, fixed in 13fbbe8

lamida (approved):
LGTM

jhesketh (approved):
lgtm :rocket:

PR #10748, opened 2025-02-26
querier: allow to enforce series query request limit

narqo:
#### What this PR does

Following #10620 #10652

This PR adds a new config and its related per-tenant override (`max_series_query_limit`), that allows to enforce the limit of a `/series` API endpoint.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/querier/querier.go
```diff
@@ -382,6 +387,16 @@ func (mq multiQuerier) Select(ctx context.Context, _ bool, sp *storage.SelectHin
 		return storage.ErrSeriesSet(NewMaxQueryLengthError(endTime.Sub(startTime), maxQueryLength))
 	}
 
+	// If the original request didn't have a limit but the limit was enforced, annotate the resulting series set with a warning.
+	if sp.Limit == 0 && sp.Limit != limit {
+		defer func() {
+			var warning annotations.Annotations
+			warning.Add(errors.New("results may be truncated due to limit"))
```
dimitarvdimitrov, line 394: can we add more details - what is the limit and what was the requested limit

dimitarvdimitrov (approved):

pkg/querier/querier.go
narqo: This only adds a warning when the request didn't send a limit. My thinking was that when I pass a `?limit=` explicitly, Prometheus adds its own warning to the response (ref [code](https://github.com/grafana/mimir/blob/9dd7b61a2bf7f2755cda179bf6ad2251d98ade77/vendor/github.com/prometheus/prometheus/web/api/v1/api.go#L982)).

dimitarvdimitrov: oh, i misread `&&` as `||`, so I assumed this will add a warning when we clamp the limit. Do you want to do that too?

narqo: I'm not sure. We don't do anything like that when we clamp other request parameters. I'm not sure if we should do that for the `limit`. So I thought to only warn the user when they didn't pass the limit, but the limit was enforced, as this is the most unexpected scenario for them.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3635,6 +3635,13 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -store.max-labels-query-length
 [max_labels_query_length: <duration> | default = 0s]
 
+# Maximum number of items, series queries. This limit is enforced in the
+# querier. If the requested limit is outside the allowed value, the request will
+# not fail but will be manipulated to only query data up to allowed limit. 0 to
+# disable.
```
tacole02, line 3641:
```suggestion
# querier. If the requested limit is outside of the allowed value, the request doesn't
# fail, but is manipulated to only query data up to the allowed limit. Set to `0` to
# disable.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3635,6 +3635,13 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -store.max-labels-query-length
 [max_labels_query_length: <duration> | default = 0s]
 
+# Maximum number of items, series queries. This limit is enforced in the
```
tacole02, line 3638: I'm a little confused by the phrasing here. Is this the maximum number of series queries permitted? If so, we can remove the `items` and just say `Maximum number of series queries`.

tacole02 (approved):

pkg/querier/querier.go
narqo: Could you take another a look. In the https://github.com/grafana/mimir/pull/10748/commits/50eaaf85eddc34852d0c11ea7888d517893794c0 I've added a more detailed warning to the response.

dimitarvdimitrov (approved):
LGTM, thanks

PR #10749, opened 2025-02-26
Reduce estimated duration for strong read consistency rule evaluations

pracucci:
#### What this PR does

We recently discussed that the 3s latency for rule evaluation with strong read consistency (which includes both enforcing the consistency + writing to the ingest storage) may be too pessimistic. In prod we actually see about 1s for consistency + 500ms for writing to ingest storage, on average. I propose to lower it to 1.5s.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/rule-group-evals/main.go
+++ b/tools/rule-group-evals/main.go
@@ -16,15 +16,16 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/grafana/mimir/pkg/ruler"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/rules"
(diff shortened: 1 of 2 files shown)


alexweav (approved):

alexweav:
I think we may need to rebase the analyzer branch first; the check seems to fail because the old code is violating newer lint rules.

pracucci:
> I think we may need to rebase the analyzer branch first; the check seems to fail because the old code is violating newer lint rules.

Yep, done.

pracucci:
> I think we may need to rebase the analyzer branch first; the check seems to fail because the old code is violating newer lint rules.

Yep, done.

PR #10754, opened 2025-02-26
distributor: Run ingestion limiter after custom push wrappers

julienduchesne:
I was chasing an issue where a tenant was limited on a 10k/s limit, with a ~200/s received rate.

image

Turns out that the ingestion limit was applied and then custom push wrappers were removing timeseries and samples. The number of series recorded in the metrics was the one after custom push wrappers

Ingestion rate limit should apply after all push wrappers, since its purpose is to limit what we send to ingesters

With this bug, the `Mimir / Tenants` dashboard is also wrong because the limit is shown against the number of received samples (see screenshot for example of that) _and_ we have no way to alert when we're approaching limits since we would typically use `cortex_distributor_received_samples_total` against the limit for that

#### Checklist

- [x] Tests updated.
- [N\A] Documentation added.
- [N\A] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.


leizor (approved):
LGTM!

bboreham:
I don’t agree with the premise. The point of the limit is to protect distributors.

bboreham:
* Possibly related: #10833

julienduchesne:
> I don’t agree with the premise. The point of the limit is to protect distributors.

Should we publish a new metric or record `cortex_distributor_received_samples_total` before the rate limiter? Because as it is, we're a bit blind on whether or not we'll be rate limited soon

tacole02 (approved):

56quarters:

PR #10755, opened 2025-02-27
MQE: Add quantile aggregation

jhesketh:

https://github.com/grafana/mimir/issues/10067

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/operators/functions/quantile.go
```diff
@@ -476,14 +476,14 @@ func ensureMonotonicAndIgnoreSmallDeltas(buckets buckets, tolerance float64) (bo
...
-// quantile calculates the given quantile of a vector of samples.
+// Quantile calculates the given Quantile of a vector of samples.
 //
 // values will be sorted in place.
 // If values has zero elements, NaN is returned.
 // If q==NaN, NaN is returned.
 // If q<0, -Inf is returned.
 // If q>1, +Inf is returned.
-func quantile(q float64, values []float64) float64 {
+func Quantile(q float64, values []float64) float64 {
```
jhesketh, line 486: Note to reviewers: this is exported for now for simplicity. I will come back with a new PR to sync this file with upstream and move it someplace more common.

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -113,6 +118,15 @@ func (a *Aggregation) ExpressionPosition() posrange.PositionRange {
 }
 
 func (a *Aggregation) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
+	// Pre-process any Scalar params
+	var err error
```
charleskorn, line 122: [nit] We could move this declaration of `err` inside the `if a.Param != nil` block

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -263,7 +277,17 @@ func (a *Aggregation) emitAnnotation(generator types.AnnotationGenerator) {
 	a.Annotations.Add(generator(metricName, a.Inner.ExpressionPosition()))
 }
 
+func (a *Aggregation) emitAnnotationParam(param float64, generator types.AnnotationParamGenerator) {
```
charleskorn, line 280: [nit] `emitParamAnnotation` reads nicer to me

(similar feedback applies for `ParamAnnotationGenerator` rather than `AnnotationParamGenerator`)

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -263,7 +277,17 @@ func (a *Aggregation) emitAnnotation(generator types.AnnotationGenerator) {
...
+	a.Annotations.Add(generator(param, a.Param.ExpressionPosition()))
+}
+
 func (a *Aggregation) Close() {
+	if len(a.paramData.Samples) > 0 {
```
charleskorn, line 285: [nit] This can be simplified to:

```suggestion
	if a.paramData.Samples != nil {
```

pkg/streamingpromql/types/functions.go
```diff
@@ -11,3 +11,9 @@ type EmitAnnotationFunc func(generator AnnotationGenerator)
 
 // AnnotationGenerator is a function that returns an annotation for the given metric name and expression position.
 type AnnotationGenerator func(metricName string, expressionPosition posrange.PositionRange) error
+
+// EmitAnnotationParamFunc is a function that emits the annotation for a scalar param created by generator.
+type EmitAnnotationParamFunc func(param float64, generator AnnotationParamGenerator)
+
+// AnnotationParamGenerator is a function that returns an annotation for the given scalar and expression position.
+type AnnotationParamGenerator func(param float64, expressionPosition posrange.PositionRange) error
```
charleskorn, line 19: Given these are currently only used for aggregations, I think we should keep them in the `aggregations` package until we have a need for them elsewhere.

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -203,6 +252,9 @@ eval range from 0 to 5m step 1m  sum(single_histogram)
 eval range from 0 to 5m step 1m  avg(single_histogram)
 	{} 0 0.5 0.5 2.5 4 {{sum:2 count:4 buckets:[1 2 1]}}
 
+eval_info range from 0 to 5m step 1m  quantile(0.5, single_histogram)
```
charleskorn, line 255:
```suggestion
eval_info range from 0 to 5m step 1m quantile(0.5, single_histogram)
```

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -188,6 +233,10 @@ eval range from 0 to 4m step 1m  avg by (label) (single_histogram)
 	{label="value"}  0 1   1   1   2
 	{label="value2"} 0 2.5 2.5 2.5 5
 
+eval range from 0 to 4m step 1m  quantile by (label) (0.5, single_histogram)
```
charleskorn, line 236:
```suggestion
eval range from 0 to 4m step 1m quantile by (label) (0.5, single_histogram)
```

pkg/streamingpromql/testdata/ours/aggregators.test
charleskorn, line 1: Could you please add a test for the case where the parameter varies over the time range, similar to https://github.com/grafana/mimir/blob/9d6eb2312b070892152aab1c36566407bcb411a3/pkg/streamingpromql/testdata/ours-only/aggregators.test? It'll have to go in `ours-only` due to https://github.com/prometheus/prometheus/issues/15971.

charleskorn, line 1: I'd also like to see a test for the case where the input contains NaN and Inf / -Inf.

pkg/streamingpromql/operators/functions/quantile.go
```diff
@@ -476,14 +476,14 @@ func ensureMonotonicAndIgnoreSmallDeltas(buckets buckets, tolerance float64) (bo
 	return forcedMonotonic, fixedPrecision
 }
 
-// quantile calculates the given quantile of a vector of samples.
+// Quantile calculates the given Quantile of a vector of samples.
```
charleskorn, line 479:
```suggestion
// Quantile calculates the given quantile of a vector of samples.
```

pkg/streamingpromql/operators/aggregations/common.go
```diff
@@ -16,22 +16,23 @@ import (
 // AggregationGroup accumulates series that have been grouped together and computes the output series data.
 type AggregationGroup interface {
 	// AccumulateSeries takes in a series as part of the group
-	AccumulateSeries(data types.InstantVectorSeriesData, timeRange types.QueryTimeRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, emitAnnotationFunc types.EmitAnnotationFunc) error
+	AccumulateSeries(data types.InstantVectorSeriesData, timeRange types.QueryTimeRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, emitAnnotationFunc types.EmitAnnotationFunc, remainingSeriesToGroup uint) error
```
charleskorn, line 19: [nit] Might be good to document that `remainingSeriesToGroup` includes the series being passed to `AccumulateSeries` (eg. if this is the last series, then `remainingSeriesToGroup` is 1)

charleskorn, line 19: [nit] Perhaps `remainingSeriesInGroup` would be a slightly clearer name?

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -218,7 +232,7 @@ func (a *Aggregation) NextSeries(ctx context.Context) (types.InstantVectorSeries
 	}
 
 	// Construct the group and return it
-	seriesData, hasMixedData, err := thisGroup.aggregation.ComputeOutputSeries(a.TimeRange, a.MemoryConsumptionTracker)
+	seriesData, hasMixedData, err := thisGroup.aggregation.ComputeOutputSeries(a.paramData, a.TimeRange, a.MemoryConsumptionTracker, a.emitAnnotationParam)
```
charleskorn, line 235: I suspect this will cause an allocation for `a.emitAnnotationParam` on every call - we may need to do the same thing we do for `a.emitAnnotationFunc` / `a.emitAnnotation` in `NewAggregation`.

Another option would be to separate out the validation of the parameter and do that just once, rather than in every call to `ComputeOutputSeries` - this would save us doing the same work over and over again, and remove the need to pass the function here as well. And then we can likely just accept the single allocation and pass `a.emitAnnotationParam` directly.

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+		q.heaps[idx].points = append(q.heaps[idx].points, p.F)
+	}
+
+	return nil
+}
+
+func (q *QuantileAggregationGroup) ComputeOutputSeries(param types.ScalarData, timeRange types.QueryTimeRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, emitAnnotationParamFunc types.EmitAnnotationParamFunc) (types.InstantVectorSeriesData, bool, error) {
+	hasMixedData := false
```
charleskorn, line 80: Given this is always `false`, do we need a variable for this?

pkg/streamingpromql/testdata/ours/aggregators.test
charleskorn, line 1: Could you please add a test for the case where we're running a range query and one timestamp has no input points?

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const maxExpectedQuantileGroups = 64 // There isn't much science to this
+
+var heapPool = types.NewLimitingBucketedPool(
+	pool.NewBucketedPool(maxExpectedQuantileGroups, func(size int) []heap {
+		return make([]heap, 0, size)
+	}),
+	uint64(unsafe.Sizeof(heap{})),
+	true,
```
charleskorn, line 37: This isn't necessary - there's no need to clear returned slices given we always set `points` to nil on line 99 (assuming we make the suggested fix there).

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		p := param.Samples[i].F
+		if math.IsNaN(p) || p < 0 || p > 1 {
+			emitAnnotationParamFunc(p, annotations.NewInvalidQuantileWarning)
+		}
+		t := timeRange.StartT + int64(i)*timeRange.IntervalMilliseconds
+		f := functions.Quantile(p, heap.points)
+		quantilePoints = append(quantilePoints, promql.FPoint{T: t, F: f})
+		types.Float64SlicePool.Put(heap.points, memoryConsumptionTracker)
+		heap.points = nil
```
charleskorn, line 99: This will be ineffective because `heap` is not a pointer type, and so won't change the value in `q.heaps[i]`. 
```suggestion
		q.heaps[i].points = nil
```

charleskorn (commented):
Overall LGTM.

I'd like to see a benchmark comparison between this and Prometheus' engine.

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/grafana/mimir/pkg/streamingpromql/operators/functions"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+	"github.com/grafana/mimir/pkg/util/pool"
+)
+
+type QuantileAggregationGroup struct {
+	heaps []heap // A heap per point in time
+}
+
+type heap struct {
```
charleskorn, line 26: [nit] Is there another name we could use for this? This isn't implementing a heap, it's just a list of values.

Alternatively: should this be a heap? `functions.Quantile()` sorts the entire list, but that's overkill, we just need a way to get the first X values.

jhesketh: This is the terminology used by prometheus, but I agree it's not strictly a heap. I'll rename to `qGroup` since `group` is taken by the aggregation.

pkg/streamingpromql/testdata/ours/aggregators.test
jhesketh: This is covered in the upstream test cases

pkg/streamingpromql/operators/aggregations/aggregation.go
jhesketh: > I suspect this will cause an allocation for `a.emitAnnotationParam` on every call - we may need to do the same thing we do for `a.emitAnnotationFunc` / `a.emitAnnotation` in `NewAggregation`.

Good catch, have fixed that.

> Another option would be to separate out the validation of the parameter and do that just once, rather than in every call to `ComputeOutputSeries` - this would save us doing the same work over and over again, and remove the need to pass the function here as well. And then we can likely just accept the single allocation and pass `a.emitAnnotationParam` directly.

The parameter is processed during the aggregations `SeriesMetadata`, so it should only be once. The values (`a.paramData` on this quoted line) are then passed into the quantile `ComputeOutputSeries` to be looked up.

jhesketh:
Benchmark results:

<details>

```

                                                                   │  Prometheus  │               Mimir                │
                                                                   │    sec/op    │   sec/op     vs base               │
Query/quantile(0.9,_a_1),_instant_query-8                             674.2µ ± 1%   639.4µ ± 1%   -5.16% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_100_steps-8                742.9µ ± 1%   689.5µ ± 1%   -7.18% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_1000_steps-8               1.172m ± 0%   1.040m ± 1%  -11.28% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_instant_query-8                           2.285m ± 1%   2.202m ± 1%   -3.64% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_100_steps-8              5.167m ± 1%   3.774m ± 1%  -26.97% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_1000_steps-8             31.71m ± 2%   18.94m ± 1%  -40.29% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_instant_query-8                          26.65m ± 1%   25.56m ± 1%   -4.11% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_100_steps-8             85.56m ± 1%   60.01m ± 1%  -29.86% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_1000_steps-8            614.7m ± 1%   377.9m ± 1%  -38.52% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_instant_query-8                     686.7µ ± 1%   649.9µ ± 1%   -5.36% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_100_steps-8        748.0µ ± 1%   699.5µ ± 0%   -6.49% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_1000_steps-8       1.189m ± 1%   1.048m ± 1%  -11.79% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_instant_query-8                   2.389m ± 1%   2.403m ± 1%        ~ (p=0.132 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_100_steps-8      5.340m ± 2%   5.320m ± 1%        ~ (p=0.093 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_1000_steps-8     35.43m ± 6%   31.84m ± 1%  -10.14% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_instant_query-8                  28.52m ± 1%   28.93m ± 1%   +1.42% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_100_steps-8    106.38m ± 1%   89.76m ± 1%  -15.62% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_1000_steps-8   1032.4m ± 2%   613.6m ± 2%  -40.56% (p=0.002 n=6)
geomean                                                               9.291m        7.854m       -15.47%

                                                                   │   Prometheus   │                Mimir                │
                                                                   │      B/op      │     B/op      vs base               │
Query/quantile(0.9,_a_1),_instant_query-8                              23.53Ki ± 0%   20.89Ki ± 0%  -11.24% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_100_steps-8                 33.46Ki ± 0%   24.93Ki ± 0%  -25.51% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_1000_steps-8               116.33Ki ± 0%   55.23Ki ± 1%  -52.53% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_instant_query-8                            190.1Ki ± 0%   153.1Ki ± 0%  -19.46% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_100_steps-8              1794.3Ki ± 0%   213.6Ki ± 0%  -88.10% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_1000_steps-8            16159.6Ki ± 0%   685.4Ki ± 1%  -95.76% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_instant_query-8                           3.328Mi ± 0%   2.631Mi ± 0%  -20.95% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_100_steps-8             35.076Mi ± 0%   3.744Mi ± 0%  -89.33% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_1000_steps-8            317.39Mi ± 0%   11.90Mi ± 1%  -96.25% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_instant_query-8                      23.70Ki ± 0%   22.26Ki ± 0%   -6.08% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_100_steps-8         33.59Ki ± 0%   26.29Ki ± 0%  -21.72% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_1000_steps-8       116.58Ki ± 0%   56.56Ki ± 1%  -51.49% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_instant_query-8                    224.7Ki ± 0%   174.0Ki ± 0%  -22.58% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_100_steps-8       988.7Ki ± 0%   552.7Ki ± 0%  -44.10% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_1000_steps-8      7.599Mi ± 0%   3.346Mi ± 0%  -55.97% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_instant_query-8                   4.124Mi ± 0%   3.265Mi ± 1%  -20.83% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_100_steps-8      18.92Mi ± 0%   10.64Mi ± 1%  -43.76% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_1000_steps-8    214.18Mi ± 0%   69.27Mi ± 1%  -67.66% (p=0.002 n=6)
geomean                                                                1.223Mi        486.0Ki       -61.21%

                                                                   │  Prometheus   │               Mimir                │
                                                                   │   allocs/op   │  allocs/op   vs base               │
Query/quantile(0.9,_a_1),_instant_query-8                               420.0 ± 0%    356.0 ± 0%  -15.24% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_100_steps-8                  630.0 ± 0%    365.0 ± 0%  -42.06% (p=0.002 n=6)
Query/quantile(0.9,_a_1),_range_query_with_1000_steps-8                2457.0 ± 0%    392.0 ± 0%  -84.05% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_instant_query-8                            2.341k ± 0%   2.264k ± 0%   -3.29% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_100_steps-8               3.669k ± 0%   2.676k ± 0%  -27.06% (p=0.002 n=6)
Query/quantile(0.9,_a_100),_range_query_with_1000_steps-8             14.480k ± 0%   5.304k ± 0%  -63.37% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_instant_query-8                           38.95k ± 0%   38.87k ± 0%   -0.22% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_100_steps-8              50.45k ± 0%   46.90k ± 0%   -7.02% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_1000_steps-8            115.27k ± 0%   99.03k ± 0%  -14.09% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_instant_query-8                       428.0 ± 0%    363.0 ± 0%  -15.19% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_100_steps-8          638.0 ± 0%    372.0 ± 0%  -41.69% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_1000_steps-8        2465.0 ± 0%    399.0 ± 0%  -83.81% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_instant_query-8                    2.664k ± 0%   2.579k ± 0%   -3.19% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_100_steps-8      23.086k ± 0%   3.107k ± 0%  -86.54% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_1000_steps-8    205.724k ± 0%   5.785k ± 0%  -97.19% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_instant_query-8                   45.07k ± 0%   44.95k ± 0%   -0.27% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_100_steps-8     453.27k ± 0%   55.25k ± 0%  -87.81% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_1000_steps-8    4115.0k ± 0%   109.2k ± 0%  -97.35% (p=0.002 n=6)
geomean                                                                11.54k        4.245k       -63.21%

                                                                   │  Prometheus   │                Mimir                │
                                                                   │       B       │      B        vs base               │
Query/quantile(0.9,_a_1),_instant_query-8                             69.27Mi ± 1%   68.80Mi ± 2%        ~ (p=0.310 n=6)
Query/quantile(0.9,_a_1),_range_query_with_100_steps-8                69.24Mi ± 1%   69.59Mi ± 1%        ~ (p=0.394 n=6)
Query/quantile(0.9,_a_1),_range_query_with_1000_steps-8               68.30Mi ± 1%   67.81Mi ± 1%        ~ (p=0.310 n=6)
Query/quantile(0.9,_a_100),_instant_query-8                           66.78Mi ± 1%   66.62Mi ± 1%        ~ (p=0.699 n=6)
Query/quantile(0.9,_a_100),_range_query_with_100_steps-8              67.27Mi ± 1%   67.25Mi ± 1%        ~ (p=1.000 n=6)
Query/quantile(0.9,_a_100),_range_query_with_1000_steps-8             70.37Mi ± 1%   68.24Mi ± 1%   -3.03% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_instant_query-8                          68.12Mi ± 1%   67.51Mi ± 1%        ~ (p=0.180 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_100_steps-8             75.13Mi ± 1%   69.97Mi ± 2%   -6.87% (p=0.002 n=6)
Query/quantile(0.9,_a_2000),_range_query_with_1000_steps-8           131.66Mi ± 1%   95.29Mi ± 1%  -27.63% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_instant_query-8                     69.64Mi ± 1%   69.04Mi ± 2%        ~ (p=0.457 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_100_steps-8        69.09Mi ± 2%   69.29Mi ± 1%        ~ (p=0.589 n=6)
Query/quantile_by_(l)_(0.1,_b_1),_range_query_with_1000_steps-8       67.82Mi ± 1%   67.50Mi ± 1%        ~ (p=0.485 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_instant_query-8                   66.91Mi ± 1%   66.76Mi ± 2%        ~ (p=0.485 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_100_steps-8      67.41Mi ± 2%   67.63Mi ± 1%        ~ (p=0.699 n=6)
Query/quantile_by_(l)_(0.1,_b_100),_range_query_with_1000_steps-8     73.80Mi ± 1%   73.80Mi ± 1%        ~ (p=0.937 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_instant_query-8                  68.50Mi ± 2%   68.52Mi ± 2%        ~ (p=0.853 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_100_steps-8     83.99Mi ± 1%   87.72Mi ± 1%   +4.44% (p=0.002 n=6)
Query/quantile_by_(l)_(0.1,_b_2000),_range_query_with_1000_steps-8    188.4Mi ± 1%   196.7Mi ± 7%        ~ (p=0.394 n=6)
geomean                                                               76.59Mi        75.03Mi        -2.03%


```

</details>

pkg/streamingpromql/testdata/ours/aggregators.test
charleskorn: I think I'm missing something sorry - where is the upstream test case for this? The only upstream test cases I can see are [here](https://github.com/grafana/mimir/blob/68139de4c46d13977819c7f7f4583cf67dc9ba32/pkg/streamingpromql/testdata/upstream/aggregators.test#L441-L482), but none of the `data` series contain NaN or Inf.

pkg/streamingpromql/benchmarks/benchmarks.go
```diff
@@ -275,6 +275,12 @@ func TestCases(metricSizes []int) []BenchCase {
 		{
 			Expr: "topk by (le) (5, h_X)",
 		},
+		{
+			Expr: "quantile(0.9, a_X)",
+		},
+		{
+			Expr: "quantile by (l) (0.1, b_X)",
```
charleskorn, line 282: `l` is the index of each series (eg. `b_100` will have `b_100{l="0"}`, `b_100{l="1"}` etc.), so this will produce X series, one for each input series. 

Perhaps a more interesting benchmark would be grouping to a smaller number of output groups? I'm imagining something like `quantile by (le) (0.1, h_X)`.

pkg/streamingpromql/operators/aggregations/aggregation.go
charleskorn: > The parameter is processed during the aggregations `SeriesMetadata`, so it should only be once. The values (`a.paramData` on this quoted line) are then passed into the quantile `ComputeOutputSeries` to be looked up.

Sorry, should have been clearer: I was referring to the work done to validate the parameter values and emit an annotation if it is not between 0 and 1 [here](https://github.com/grafana/mimir/blob/68139de4c46d13977819c7f7f4583cf67dc9ba32/pkg/streamingpromql/operators/aggregations/quantile.go#L91-L93) - this will be repeated for every output group, and is the only reason we need to pass a `emitParamAnnotationFunc` to `ComputeOutputSeries`. 

If we instead move that validation out of `ComputeOutputSeries`, then we'll only do it once regardless of the number of groups, and we don't need to pass `emitParamAnnotationFunc` to `ComputeOutputSeries` either.

jhesketh: Ah, I see. Good point.

I've refactored things to avoid this. Now Quantile is it's own operator that is a thing wrapper around Aggregation. This is to avoid repeating all of the grouping logic. The downside is that Aggregation still needs to keep ParamData, which is a bit odd, but I think a decent trade off.

As a side effect: Previously if no series were output, no annotation would be emitted for a bad quantile value. However since we process them independently we now always output an annotation. This is more consistent with Prometheus anyway.

pkg/streamingpromql/testdata/ours/aggregators.test
jhesketh: Sorry I thought you meant for the quantile param. Also I see it doesn't have Infinities, so not sure where I imagined that. I'll add them.

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,180 @@
...
+		Aggregation:              a,
+		Param:                    param,
+		MemoryConsumptionTracker: memoryConsumptionTracker,
+		Annotations:              annotations,
+	}
+
+	return q, nil
+}
+
+func (q *QuantileAggregation) ExpressionPosition() posrange.PositionRange {
```
charleskorn, line 68: [tiny nit] I'd move this down so that `SeriesMetadata` and `NextSeries` appear first

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return q.Aggregation.SeriesMetadata(ctx)
+}
+
+func (q *QuantileAggregation) NextSeries(ctx context.Context) (types.InstantVectorSeriesData, error) {
+	return q.Aggregation.NextSeries(ctx)
+}
+
+func (q *QuantileAggregation) Close() {
+	if q.Aggregation.ParamData.Samples != nil {
```
charleskorn, line 93: It feels weird that we're reaching into `q.Aggregation` here - I think it'd be OK to move this into `Aggregation.Close()`.

pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -0,0 +1,180 @@
...
+	uint64(unsafe.Sizeof(qGroup{})),
+	false,
+	nil,
+)
+
+func (q *QuantileAggregationGroup) AccumulateSeries(data types.InstantVectorSeriesData, timeRange types.QueryTimeRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, emitAnnotationFunc types.EmitAnnotationFunc, remainingSeriesInGroup uint) error {
+	defer types.PutInstantVectorSeriesData(data, memoryConsumptionTracker)
+
+	if len(data.Histograms) > 0 {
+		emitAnnotationFunc(func(_ string, expressionPosition posrange.PositionRange) error {
```
charleskorn, line 125: [nit] I realise this was what it was called previously, but I think `emitAnnotation` would be a better name for this parameter, with `EmitAnnotationFunc` remaining as the type.

(could be something for a follow-up PR)

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -171,15 +216,27 @@ load 1m
...
 
 # What both engines return
+eval range from 0 to 4m step 1m sum by (label) (single_histogram)
+  {label="value"}  0 1 1 1 2
+  {label="value2"} 0 5 5 5 10
+
+eval range from 0 to 4m step 1m avg by (label) (single_histogram)
+  {label="value"}  0 1   1   1   2
+  {label="value2"} 0 2.5 2.5 2.5 5
+
```
charleskorn, line 235: Looks like these tests have been duplicated (see tests after `quantile` test below)

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -197,12 +254,21 @@ load 1m
...
+  {} 0 1 1 5 8 {{sum:4 count:8 buckets:[2 4 2]}}
+
+eval range from 0 to 5m step 1m avg(single_histogram)
+  {} 0 0.5 0.5 2.5 4 {{sum:2 count:4 buckets:[1 2 1]}}
+
 eval range from 0 to 5m step 1m  sum(single_histogram)
   {} 0 1 1 5 8 {{sum:4 count:8 buckets:[2 4 2]}}
 
 eval range from 0 to 5m step 1m  avg(single_histogram)
   {} 0 0.5 0.5 2.5 4 {{sum:2 count:4 buckets:[1 2 1]}}
```
charleskorn, line 268: Same as above - these tests are here twice now

pkg/streamingpromql/testdata/ours/aggregators.test
```diff
@@ -625,3 +700,20 @@ eval range from 0 to 12m step 6m count_values by (idx) ("idx", series)
...
+  series{idx="3"}   7 8 Inf NaN -Inf
+
+# Quantile value warning is emitted even when no series are returned
+eval_warn range from 0 to 12m step 6m quantile(20, noseries)
+
+eval_warn range from 0 to 12m step 6m quantile(Inf, noseries)
+
+eval_warn range from 0 to 12m step 6m quantile(-Inf, noseries)
+
+eval range from 0 to 30m step 6m quantile(0.9, series)
```
charleskorn, line 718: Could you add a test like this with `quantile(0, series)` and `quantile(1, series)` to confirm whether Inf and NaN values sort to the start or end?

charleskorn (commented):
Overall LGTM. Noticed something weird in the test cases though.

pkg/streamingpromql/operators/aggregations/quantile.go
jhesketh: Will leave for followup

charleskorn (approved):

pkg/streamingpromql/operators/aggregations/quantile.go
jhesketh: Done in https://github.com/grafana/mimir/pull/10885

PR #10758, opened 2025-02-27
mixin: make federation-frontend remote_cluster variable multi-select

dimitarvdimitrov:
This works for more grafana versions. Some versions don't default to `All`


narqo (approved):
🔥

PR #10760, opened 2025-02-27
distributor: add temporary metric to see if we're dropping nh samples

krajorama:
#### What this PR does

Measure how many native histograms samples we're dropping silently. One native histograms are stable and enabled by default, this metric will no longer be needed.

N/A

#### Checklist

- [x] Tests updated.
- [N/A] Documentation added.


ying-jeanne (commented):
The PR looks good and straight forward, nit: the changelog entry/name says this metric tracks all dropped native histograms, but from the code, it seems to only count those dropped due to native histograms not being enabled by config. Could you clarify this in changelog or comments to avoid confusion?

pkg/distributor/distributor.go
```diff
@@ -467,6 +470,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 			Help: "Total number of label values with newlines seen at ingestion time.",
 		}, []string{"user"}),
 
+		droppedNativeHistograms: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_distributor_dropped_native_histograms_total",
+			Help: "The total number of native histograms that were silently dropped.",
```
dimitarvdimitrov, line 475: could that be only because they are disabled for the user? I'd add that to the description

pkg/distributor/distributor.go
```diff
@@ -467,6 +470,11 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 			Help: "Total number of label values with newlines seen at ingestion time.",
 		}, []string{"user"}),
 
+		droppedNativeHistograms: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_distributor_dropped_native_histograms_total",
```
dimitarvdimitrov, line 474: there's also `cortex_discarded_samples_total{user=, reason=}` if you want to use that. It's already displayed in the tenants dashboard

krajorama: I don't really want to use that as all of those return 4xx so this would be a major inconsistency I think.

pkg/distributor/distributor.go
krajorama: updated

dimitarvdimitrov (approved):
SGTM but tests are failing

krajorama:
> SGTM but tests are failing

yep, help text is actually in unit tests. updated.

PR #10765, opened 2025-02-28
Add a linter to enforce sorted lines using lint:sorted directive.

colega:
#### What this PR does

_As someone once commented somewhere: we are not savages._

The lack of sorting in the modules.go modules constant definitions was bothering me big time, especially since CostAttribution was added after All.

This sorts them and adds a linter that is triggered by adding a // lint:sorted comment which would fail if lines aren't sorted.

The linter checks lines from the line containing the directive until the next white space or a line that only contains closing bracket.

Fixes my OCD

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimir/modules.go
+++ b/pkg/mimir/modules.go
@@ -72,48 +72,50 @@ import (
 
 // The various modules that make up Mimir.
 const (
-	ActivityTracker                  string = "activity-tracker"
+	//lint:sorted
 	API                              string = "api"
-	SanityCheck                      string = "sanity-check"
-	IngesterRing                     string = "ingester-ring"
-	IngesterPartitionRing            string = "ingester-partitions-ring"
-	RuntimeConfig                    string = "runtime-config"
-	Overrides                        string = "overrides"
-	OverridesExporter                string = "overrides-exporter"
-	Server                           string = "server"
 	ActiveGroupsCleanupService       string = "active-groups-cleanup-service"
-	Distributor                      string = "distributor"
-	DistributorService               string = "distributor-service"
-	Ingester                         string = "ingester"
-	IngesterService                  string = "ingester-service"
-	Flusher                          string = "flusher"
-	Querier                          string = "querier"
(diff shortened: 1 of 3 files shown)


tools/lint-sorted/main.go
```diff
@@ -0,0 +1,115 @@
...
+
+}
+
+// checkSorted verifies whether lines after the directive are sorted lexicographically
+func checkSorted(lines []string, filePath string, startIdx int) bool {
+	var block []string
+	currentIdx := startIdx
+	for currentIdx < len(lines) {
+		trimmedLine := strings.TrimSpace(lines[currentIdx])
+		if trimmedLine == "" || trimmedLine == "}" {
```
56quarters, line 100: Do we need to check for comments here or in general? It doesn't seem out-of-place for someone to not expect comments to be sorted.

56quarters (approved):
LGTM. Do you have any sense of how fast this is since we're going through every line of every go file in the repo (~200K lines last I looked).

tools/lint-sorted/main.go
colega: I expect this code to be evolved with our needs in the future.

colega:
> Do you have any sense of how fast this is since we're going through every line of every go file in the repo
The slowest part is to build the app (which is quite fast), scanning files is sub-second:
```
$ time go run ./tools/lint-sorted/ --include '*.go' --path ./pkg
go run ./tools/lint-sorted/ --include '*.go' --path ./pkg  0.20s user 0.45s system 443% cpu 0.148 total
```

pkg/mimir/modules.go
```diff
@@ -72,48 +72,50 @@ import (
 
 // The various modules that make up Mimir.
 const (
-	ActivityTracker                  string = "activity-tracker"
+	//lint:sorted
```
dimitarvdimitrov, line 75: amazing

PR #10772, opened 2025-02-28
Add additional check on reactive limiter queue sizes

jhalterman:
#### What this PR does

This change improves on #10574 by adding an additional check for reactive limiter queue sizes that limits the amount of queueing in each limiter based on its individual settings. At present, rejections are based on the prioritizer's threshold, which is computed periodically and looks at queueing across limiters along with request priorities, but this doesn't prevent a large burst of requests from enqueueing before the next calibration takes place. This PR adds an additional guard against this.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


56quarters (approved):
LGTM, thank you!

PR #10774, opened 2025-03-03
MQE: enforce consistent formatting for test files

charleskorn:
#### What this PR does

This PR adds a script to enforce consistent formatting of PromQL test files.

I've chosen two spaces as the default indentation given that seems to be what most of our existing test files already use.

I've introduced each of the changes incrementally in their own commit.

If we find this useful, I'll also propose adding it upstream in Prometheus.

#10067

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/tools/format-promql-test.sh
+++ b/tools/format-promql-test.sh
@@ -1,4 +1,5 @@
 #! /usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-only
 
 set -euo pipefail
 
--- a/tools/format-promql-test.sh
+++ b/tools/format-promql-test.sh
@@ -3,10 +3,10 @@
 set -euo pipefail
 
 # Ensure all lines except load, eval and clear commands and comments are indented
-sed -E -i '' '/^(load|eval|clear|.*#)/! s/^[[:space:]]*/  /g' "$1"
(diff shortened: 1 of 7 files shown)


tools/format-promql-test.sh
```diff
@@ -0,0 +1,22 @@
...
+sed -E "${EDIT_IN_PLACE[@]}" '/^(load|eval|clear|.*#)/! s/^[[:space:]]*/  /g' "$@"
+
+# Ensure load, eval and clear commands are not indented
+sed -E "${EDIT_IN_PLACE[@]}" 's/^[[:space:]]+(load|eval|clear)/\1/g' "$@"
+
+# Convert leading whitespace to two spaces
+sed -E "${EDIT_IN_PLACE[@]}" 's/^[[:space:]]+/  /g' "$@"
+
+# Strip trailing whitespace
+sed -E "${EDIT_IN_PLACE[@]}" 's/[[:space:]]+$//g' "$@"
```
jhesketh, line 22: Is it worth adding a check that there is never more than one blank line?

jhesketh (approved):

tools/format-promql-test.sh
charleskorn: Added in edb7395.

PR #10781, opened 2025-03-03
Stop Nginx socket listening on IPv6 when IPv6 is disabled

dsgnr:
#### What this PR does
Fixes an issue where the Nginx deployment would still bing to IPv6 even if we set `gateway.nginx.config.enableIPv6` to false.
This causes the deployment to crash if IPv6 is disabled on the host.

~~~
✗ k logs deploy/mimir-nginx
/docker-entrypoint.sh: No files found in /docker-entrypoint.d/, skipping configuration
2025/03/03 10:06:22 [emerg] 1#1: socket() [::]:8080 failed (97: Address family not supported by protocol)
nginx: [emerg] socket() [::]:8080 failed (97: Address family not supported by protocol)
~~~

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -41,6 +41,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [ENHANCEMENT] Set resources for smoke-test job. #10608
 * [BUGFIX] Create proper in-cluster remote URLs when gateway and nginx are disabled. #10625
 * [BUGFIX] Fix calculation of `mimir.siToBytes` and use floating point arithmetics. #10044
+* [BUGFIX] Fix Nginx listening sockets when IPv6 is disabled on in Helm values.
 
 ## 5.6.0
 
--- a/operations/helm/charts/mimir-distributed/values.yaml
+++ b/operations/helm/charts/mimir-distributed/values.yaml
@@ -3233,7 +3233,9 @@ nginx:
         proxy_read_timeout 300;
         server {
           listen 8080;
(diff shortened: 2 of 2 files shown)


dsgnr:
@dimitarvdimitrov, the Helm chart currently fails with the default `values.yaml` if the host has IPv6 disabled. Would you prefer a `nginx.nginxConfig.disableIPv6` boolean instead?

dimitarvdimitrov:
> Would you prefer a nginx.nginxConfig.disableIPv6 boolean instead?

Is it significantly different to migrate to the `gateway` nginx vs the root `nginx` deployment?

I would prefer not to reduce the friction around a deprecated component. That would force installs to migrate to the recommended deployment rather than fixing a problem early in the install and then having to do a migration in production systems.

operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [ENHANCEMENT] Set resources for smoke-test job. #10608
 * [BUGFIX] Create proper in-cluster remote URLs when gateway and nginx are disabled. #10625
 * [BUGFIX] Fix calculation of `mimir.siToBytes` and use floating point arithmetics. #10044
+* [BUGFIX] Fix Nginx listening sockets when IPv6 is disabled on in Helm values.
```
tacole02, line 44:
```suggestion
* [BUGFIX] Fix Nginx listening sockets when IPv6 is disabled in Helm values.
```

56quarters:

PR #10796, opened 2025-03-04
frontend: Add back split and cache to range middlewares

julienduchesne:
It was removed here: https://github.com/grafana/mimir/pull/10742 by mistake

Also, change `TestMiddlewaresConsistency` so that we catch those

#### Checklist

- [x] Tests updated.
- [N/A] Documentation added.
- [N/A] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.


narqo (approved):
Thank you for a quick fix 🔥

PR #10808, opened 2025-03-05
Update push-mimir-build-image.yml to use app token

armandgrillet:
Second commit is just to test that the workflow is running properly.


armandgrillet:
We can see that the new workflow works from the last comment and the commit [3994c08](https://github.com/grafana/mimir/pull/10808/commits/3994c082b1551fdc560519ee238dbd60328091a8) which I didn't create myself. I will now delete the change in the Dockerfile that I had added to test that the updated workflow was working.

PR #10809, opened 2025-03-05
Update grafanabot_reviewer.yml to use app token

armandgrillet:
Due to https://github.com/grafana/deployment_tools/tree/master/docs/platform/continuous-integration#avoid-pull_request_target the workflow had to be changed from `on: pull_request_target` to `on: pull_request`. However that is fine as Grot PRs are created in the upstream repository thus that only makes the workflow safer!

### DIFF
--- a/.github/workflows/grafanabot_reviewer.yml
+++ b/.github/workflows/grafanabot_reviewer.yml
@@ -10,8 +10,6 @@ jobs:
   dependabot-reviewer:
     runs-on: ubuntu-latest
 
-    if: ${{ github.event.pull_request.user.login == 'grafanabot' }}
-
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):

armandgrillet:
We can see from that message and https://github.com/grafana/mimir/actions/runs/13674321077/job/38231333682 that the workflow still works.

narqo (approved):
I think, it may not work in practice — it doesn't work in the `main` branch — due to the author of the PR and reviewer cannot be the same (providing the `mimir-github-bot:app_id` belongs to `grafanabot`).

But this is not the problem of the change, let's finish getting rig of PAT's first, and get to fixing what's been already broken after 🔥

**Update** I think, the GitHub app will use the `mimir-github-bot[bot]` user. So we should be good. I.e. the app should be able to approve the grafanabot's PRs now

PR #10815, opened 2025-03-05
Return context.Cause(ctx) as error so we can use custom timeout errors

zenador:
#### What this PR does

To use https://pkg.go.dev/context#WithTimeoutCause , we need to return the cause as error, which is the normal behaviour, except that when we use https://github.com/golang/sync/blob/master/semaphore/semaphore.go#L49 in this code, it does not use the cause as error. To fix that, we check the error returned from there and replace it with the cause if relevant. Without this change, using `WithTimeoutCause` has no effect when calling `Do` on `limitedParallelismRoundTripper`.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

Minor change which by itself should not result in any user-facing changes, but this makes other changes possible.

### DIFF
--- a/pkg/frontend/querymiddleware/limits.go
+++ b/pkg/frontend/querymiddleware/limits.go
@@ -7,6 +7,7 @@ package querymiddleware
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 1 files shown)


dimitarvdimitrov (approved):
nice find

pkg/frontend/querymiddleware/limits.go
```diff
@@ -250,6 +251,9 @@ func (rt limitedParallelismRoundTripper) RoundTrip(r *http.Request) (*http.Respo
 	response, err := rt.middleware.Wrap(
 		HandlerFunc(func(ctx context.Context, r MetricsQueryRequest) (Response, error) {
 			if err := sem.Acquire(ctx, 1); err != nil {
+				if errors.Is(err, ctx.Err()) {
```
charleskorn, line 254: [nit] Should we add a comment here explaining why we're doing this?

charleskorn (approved):

PR #10816, opened 2025-03-05
ruler: Make alertmanager client and all related config per-tenant configurable

alexweav:
#### What this PR does

This PR adds a new per-tenant configuration block that allows the alertmanager URL and all related client options to be changed on a per-tenant basis. This allows different ruler tenants to send alerts to different alertmanagers.

There are a few deployment topologies where this might be desirable:
- Moving a tenant from one Mimir cell to another. If the new cell is pointed at a different alertmanager (or has its own alertmanager deployed internally), you no longer have to migrate the tenant's alertmanager configuration in lock-step.
- If you have multiple alertmanager environments with different auth parameters or URLs, you don't need to run separate mimir cells to map each tenant to each alertmanager.

The new config is a sub-block of the tenant overrides. An example might look like:
```
user1:
  ruler_alertmanager_client_config:
    alertmanager_url: http://custom-url-for-this-tenant:8080
```

If a tenant does not supply any tenant-specific config, it falls back to the global ruler-level configurations given by the existing parameters.

**Note:** This PR is part 1 of the change that proposes and applies the configuration. It does not yet implement hot-reload for this config block; you need to restart the ruler for changes to this config to take effect. I will support hot-reload, I just decided to post it in a separate PR to keep the diff reviewable.

n/a

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


CHANGELOG.md
```diff
@@ -64,6 +64,7 @@
 * [ENHANCEMENT] All: Add `cortex_client_request_invalid_cluster_validation_labels_total` metrics, that is used by Mimir's gRPC clients to track invalid cluster validations. #10767
 * [ENHANCEMENT] Ingester client: Add support to configure cluster validation for ingester clients. Failed cluster validations are tracked by `cortex_client_request_invalid_cluster_validation_labels_total` with label `client=ingester`. #10767
 * [ENHANCEMENT] Add experimental metric `cortex_distributor_dropped_native_histograms_total` to measure native histograms silently dropped when native histograms are disabled for a tenant. #10760
+* [ENHANCEMENT] Add tenant configuration block `ruler_alertmanager_client_config` which allows the Ruler's Alertmanager client options to be specified on a per-tenant basis. #10816
```
tacole02, line 67:
```suggestion
* [ENHANCEMENT] Add tenant configuration block `ruler_alertmanager_client_config`, which allows you to specify the ruler's Alertmanager client options on a per-tenant basis. #10816
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3881,6 +3881,84 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -ruler.max-independent-rule-evaluation-concurrency-per-tenant
 [ruler_max_independent_rule_evaluation_concurrency_per_tenant: <int> | default = 4]
 
+# Per-tenant alertmanager client configuration. If not supplied, the tenant's
+# notifications will be sent to the ruler-wide default.
```
tacole02, line 3885:
```suggestion
# notifications are sent to the ruler-wide default.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3881,6 +3881,84 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -ruler.max-independent-rule-evaluation-concurrency-per-tenant
 [ruler_max_independent_rule_evaluation_concurrency_per_tenant: <int> | default = 4]
 
+# Per-tenant alertmanager client configuration. If not supplied, the tenant's
```
tacole02, line 3884:
```suggestion
# Per-tenant Alertmanager client configuration. If not supplied, the tenant's
```

tacole02 (approved):
Thank you!

pkg/util/validation/limits.go
```diff
@@ -208,17 +209,18 @@ type Limits struct {
...
+	RulerMaxRuleGroupsPerTenantByNamespace                flagext.LimitsMap[int]            `yaml:"ruler_max_rule_groups_per_tenant_by_namespace" json:"ruler_max_rule_groups_per_tenant_by_namespace" category:"experimental"`
+	RulerProtectedNamespaces                              flagext.StringSliceCSV            `yaml:"ruler_protected_namespaces" json:"ruler_protected_namespaces" category:"experimental"`
+	RulerMaxIndependentRuleEvaluationConcurrencyPerTenant int64                             `yaml:"ruler_max_independent_rule_evaluation_concurrency_per_tenant" json:"ruler_max_independent_rule_evaluation_concurrency_per_tenant" category:"experimental"`
+	RulerAlertmanagerClientConfig                         notifier.AlertmanagerClientConfig `yaml:"ruler_alertmanager_client_config" json:"ruler_alertmanager_client_config" category:"experimental" doc:"nocli|description=Per-tenant Alertmanager client configuration. If not supplied, the tenant's notifications are sent to the ruler-wide default."`
```
francoposa, line 223: From checking existing help docs, it looks like we mostly use the phrase "If unset"

pkg/ruler/notifier/notifier_config.go
```diff
@@ -0,0 +1,99 @@
...
+var DefaultAlertmanagerClientConfig = AlertmanagerClientConfig{
+	NotifierConfig: Config{
+		OAuth2: OAuth2Config{
+			EndpointParams: flagext.NewLimitsMap[string](nil),
+		},
+	},
+}
+
+type AlertmanagerClientConfig struct {
+	AlertmanagerURL string `yaml:"alertmanager_url" doc:"nocli"`
```
francoposa, line 26: I believe this also should have the CLI params and be visible in the `limits` help block, same as the other pattern discussed

56quarters:
I agree with @francoposa that this should follow the usual pattern for per-tenant settings where they are defined in `util/validation/limits`, especially because this is going to involve hot-reloading which is already supported with limits in runtime config. 

Since this changes configuration, it's very difficult and disruptive to change it later. I'd much rather be consistent with all the existing per-tenant configuration unless there's an overwhelming reason not do that here. The `CustomTrackersConfig` might be a useful example for how this could be structured or laid out in code.

56quarters:

alexweav:
I've pivoted this PR to be compliant with the config migration process. The core idea/code is largely intact, but there's some key internal changes:
- The config flags are moved entirely to tenant limits. They're still backwards compatible from user perspective, but they set their values in the default limits rather than the ruler config struct.
- The old ruler config fields are marked deprecated. If they're supplied (only possible via YAML), they're copied to the new spot in tenant limits during initialization.
- The ruler `manager` code is now much simpler. It just directly reads the fields out the tenant limits, straight up, without needing to worry about the rulerCfg fields. It no longer builds any notifiers on startup at all.
- I've migrated all the tests to use the new limits-based path and not reference the deprecated yaml fields.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1962,32 +1957,23 @@ The `ruler` block configures the ruler.
 [notification_timeout: <duration> | default = 10s]
 
 alertmanager_client:
-  # (advanced) Enable TLS for gRPC client connecting to alertmanager.
-  # CLI flag: -ruler.alertmanager-client.tls-enabled
-  [tls_enabled: <boolean> | default = true]
+  # (advanced)
+  [tls_enabled: <boolean> | default = ]
```
alexweav, line 1961: Since we unbound the flags for the old ruler config fields, we also lose the docstrings. It's just moved down below under limits. That seems to be desired/expected by the process though. I chose not to publicly mark these as deprecated just yet.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3822,6 +3786,120 @@ The `limits` block configures default and per-tenant limits imposed by component
...
 [ruler_max_independent_rule_evaluation_concurrency_per_tenant: <int> | default = 4]
 
+# Per-tenant Alertmanager client configuration. If not supplied, the tenant's
+# notifications are sent to the ruler-wide default.
+ruler_alertmanager_client_config:
+  # Comma-separated list of URL(s) of the Alertmanager(s) to send notifications
+  # to. Each URL is treated as a separate group. Multiple Alertmanagers in HA
+  # per group can be supported by using DNS service discovery format,
+  # comprehensive of the scheme. Basic auth is supported as part of the URL.
+  # CLI flag: -ruler.alertmanager-url
```
alexweav, line 3796: Here's the result of following the process. It's still a sub-block under limits in yaml, but each option also has its own flag, and it's exactly the set of flags that already existed. The process made this much cleaner IMO, good work @francoposa!

operations/compare-helm-with-jsonnet/components/config/irrelevant-config.yaml
```diff
@@ -81,3 +81,6 @@ config:
   common:
     storage:
       backend:
+  limits:
+    ruler_alertmanager_client_config:
+      alertmanager_url:
```
alexweav, line 86: This is not really new. It's already marked further up in the file as irrelevant for the helm/jsonnet differ: https://github.com/grafana/mimir/blob/9d620ef5cf9a2c595eb48b2a0ecc47033e1a3492/operations/compare-helm-with-jsonnet/components/config/irrelevant-config.yaml#L39

Since the config is in a new spot, we have to suppress it in the new spot too.

pkg/mimir/modules.go
```diff
@@ -238,11 +238,14 @@ func (t *Mimir) initVault() (services.Service, error) {
...
 	t.Cfg.QueryScheduler.GRPCClientConfig.TLS.Reader = t.Vault
 	t.Cfg.Ruler.ClientTLSConfig.TLS.Reader = t.Vault
-	t.Cfg.Ruler.Notifier.TLS.Reader = t.Vault
+	t.Cfg.Ruler.DeprecatedNotifier.TLS.Reader = t.Vault
 	t.Cfg.Ruler.QueryFrontend.GRPCClientConfig.TLS.Reader = t.Vault
 	t.Cfg.Worker.QueryFrontendGRPCClientConfig.TLS.Reader = t.Vault
 	t.Cfg.Worker.QuerySchedulerGRPCClientConfig.TLS.Reader = t.Vault
 
+	// Update Configs - LimitsConfigs
+	t.Cfg.LimitsConfig.RulerAlertmanagerClientConfig.NotifierConfig.TLS.Reader = t.Vault
```
alexweav, line 247: Making sure we propagate the reader so vault-based TLS secrets are not broken

pkg/ruler/notifier.go
```diff
@@ -34,45 +31,6 @@ var (
 	errRulerSimultaneousBasicAuthAndOAuth = errors.New("cannot use both Basic Auth and OAuth2 simultaneously")
 )
 
-type NotifierConfig struct {
```
alexweav, line 37: I moved this to a sub-package to avoid a circular dependency. (`validation` referencing `ruler` and `ruler` references `validation` many places in tests for creating mocks).

Such issues could also be avoided by putting the `validation` pkg test utils in a subpackage. `ruler` doesn't really depend on `validation` at a code level, only tests. What I did here is a less heavyweight change.

pkg/util/validation/limits.go
alexweav: I've made quite a few changes to better follow the process now: https://github.com/grafana/mimir/pull/10816#issuecomment-2828920344

56quarters (approved):
LGTM but I'd like @francoposa to take a look as well.

pkg/util/validation/limits.go
```diff
@@ -210,17 +211,18 @@ type Limits struct {
...
+	RulerMaxRuleGroupsPerTenantByNamespace                flagext.LimitsMap[int]            `yaml:"ruler_max_rule_groups_per_tenant_by_namespace" json:"ruler_max_rule_groups_per_tenant_by_namespace" category:"experimental"`
+	RulerProtectedNamespaces                              flagext.StringSliceCSV            `yaml:"ruler_protected_namespaces" json:"ruler_protected_namespaces" category:"experimental"`
+	RulerMaxIndependentRuleEvaluationConcurrencyPerTenant int64                             `yaml:"ruler_max_independent_rule_evaluation_concurrency_per_tenant" json:"ruler_max_independent_rule_evaluation_concurrency_per_tenant" category:"experimental"`
+	RulerAlertmanagerClientConfig                         notifier.AlertmanagerClientConfig `yaml:"ruler_alertmanager_client_config" json:"ruler_alertmanager_client_config" category:"experimental" doc:"description=Per-tenant Alertmanager client configuration. If not supplied, the tenant's notifications are sent to the ruler-wide default."`
```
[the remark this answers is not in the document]
alexweav: The other blocks/`omitempty` fields are all lists of things, or aliases for lists like `LimitedQueriesConfig` being a `[]*LimitedQuery`, where this one is just a plain set of sub-fields.

Our yaml library doesn't support go 1.24's `omitzero` yet and panics when you set it :cry: 

It can be made omitempty if it's changed to a `*notifier.AlertmanagerClientConfig`, but this also brings in extra risk for nil dereferences. Happy to go that route if you have a strong opinion here, but it has a drawback.

docs/sources/mimir/configure/configuration-parameters/index.md
francoposa: Personally I do not see a reason not to mark as deprecated. It's another place where people can be warned that it will be deprecated, and if it's going to get moved I would want as much visibility as possible

pkg/mimir/runtime_config.go
```diff
@@ -231,6 +231,20 @@ func NewRuntimeManager(cfg *Config, name string, reg prometheus.Registerer, logg
 	loader := runtimeConfigLoader{validate: cfg.ValidateLimits}
 	cfg.RuntimeConfig.Loader = loader.load
 
+	// AlertmanagerURL and Notifier sub-options are moving from a global config to a per-tenant config.
+	// We need to preserve the option in the ruler yaml for two releases.
+	// If the ruler config is configured by the user, map it to its new place in the default limits.
```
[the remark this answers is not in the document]
alexweav: Updated! (With 3.0 coming up, I think that's *3.1* right?)

pkg/util/validation/limits.go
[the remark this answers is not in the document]
francoposa: Been squinting at this for a bit and I think it is better as-is. It seems more complicated than not to use `omitempty` for a struct member like this, whereas it makes more sense for lists in Go since they are already pointer types and have semantics where an empty list is almost always treated the same as a nil list

francoposa (approved):
LGTM with small caveat that I would prefer it get marked deprecated in the original spot & have that reflected in docs/config-descriptor.

Not something I would block it on though

pkg/mimir/runtime_config.go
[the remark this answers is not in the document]
56quarters: Nope, it'd be 3.0 (or 2.19 which won't exist)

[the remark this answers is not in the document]
alexweav: Cool, fixed the comment.

PR #10820, opened 2025-03-06
Use the app token in backport.yml

armandgrillet:
This is the nearly the same as https://github.com/grafana/backend-enterprise/blob/master/.github/workflows/backport.yml as this workflow works: https://github.com/grafana/backend-enterprise/pull/8440


narqo (approved):
🔥

PR #10824, opened 2025-03-07
Upgrade mimir-prometheus

pr00se:
Pulls in https://github.com/grafana/mimir-prometheus/pull/838, which pulls in https://github.com/prometheus/prometheus/pull/16060


charleskorn (approved):
Does this need a changelog entry?

pr00se:
> Does this need a changelog entry?

This is a prometheus bugfix -- not sure if we list those in our CHANGELOG as well?

charleskorn:
> > Does this need a changelog entry?
> 
> This is a prometheus bugfix -- not sure if we list those in our CHANGELOG as well?

We're pretty inconsistent about this... given this is something a Mimir user could observe, I would lean towards adding one. I'm not sure a user would care about the distinction between a Mimir bugfix and a Prometheus bugfix, from their perspective, Mimir isn't doing what it should.

PR #10826, opened 2025-03-07
Upgrade mimir-prometheus

charleskorn:
#### What this PR does

This PR upgrades mimir-prometheus to include the changes from https://github.com/grafana/mimir-prometheus/pull/837.

I've added a changelog entry for one of the fixes in that set of changes, however the remainder don't seem worthy of a changelog entry (eg. because they're not relevant for Mimir, or are for experimental native histograms).

I've tried to build up this PR incrementally. I would suggest reviewing each commit separately.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/engine_test.go
```diff
@@ -2527,6 +2527,124 @@ func TestDeltaAnnotations(t *testing.T) {
...
+		},
+		"idelta() over metric with second point not being a gauge native histogram": {
+			data: ideltaData,
+			expr: `idelta(metric{series="nh-second-not-gauge"}[1m1s])`,
+			expectedWarningAnnotations: []string{
+				`PromQL warning: this native histogram metric is not a gauge: "metric" (1:8)`,
+			},
+		},
+	}
+
```
jhesketh, line 2646: What about a test case for when the second last point and last point change types, and don't change types (in both orders)?

jhesketh (commented):
lgtm, thanks for working on this! Just one new change that I don't think is adequately tested.

pkg/streamingpromql/engine_test.go
charleskorn: This is quite difficult to test completely due to https://github.com/prometheus/prometheus/issues/14172. 

[The test cases added upstream](https://github.com/prometheus/prometheus/pull/15853/files#diff-f277c3549a17549dbc047c4f13cdeebd450a0a69b2a568f7dde50935583b3540) and included in this PR should cover most of the scenarios: float then float, float then histogram and histogram then histogram. Histogram then float is blocked by https://github.com/prometheus/prometheus/issues/14172.

Are there other scenarios you had in mind?

jhesketh: Right, makes sense. However, we can test "histogram then float" in ours-only. Moving this to a nit though, so lgtm

jhesketh: I missed those, thanks. lgtm

jhesketh (approved):

PR #10841, opened 2025-03-10
MQE: Implement absent_over_time

lamida:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/timestamp"
 	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/promql/parser/posrange"
 	"github.com/prometheus/prometheus/promql/promqltest"
 	"github.com/prometheus/prometheus/storage"
@@ -47,9 +48,19 @@ func init() {
 func TestUnsupportedPromQLFeatures(t *testing.T) {
 	features := EnableAllFeatures
 
+	// Aliasing holt_winters call
+	promql.FunctionCalls["holt_winters"] = promql.FunctionCalls["double_exponential_smoothing"]
+	parser.Functions["holt_winters"] = parser.Functions["double_exponential_smoothing"]
+	parser.Functions["holt_winters"].Experimental = false
+
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
 	// different cases and make sure we produce a reasonable error message when these cases are encountered.
-	unsupportedExpressions := map[string]string{}
+	unsupportedExpressions := map[string]string{
+		// expected error for holt_winters is the same as double_exponential_smoothing because mimir just aliasing it as above
+		"holt_winters(metric{}[1h], 1, 1)": "'double_exponential_smoothing' function",
+
+		"double_exponential_smoothing(metric{}[1h], 1, 1)": "'double_exponential_smoothing' function",
+	}
 
 	for expression, expectedError := range unsupportedExpressions {
 		t.Run(expression, func(t *testing.T) {
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -51,6 +51,7 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
 	// Aliasing holt_winters call
 	promql.FunctionCalls["holt_winters"] = promql.FunctionCalls["double_exponential_smoothing"]
 	parser.Functions["holt_winters"] = parser.Functions["double_exponential_smoothing"]
+	// Disable experimental so that parser will parse it without the updating experiemental flag
 	parser.Functions["holt_winters"].Experimental = false
 
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -49,9 +49,7 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
 
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
 	// different cases and make sure we produce a reasonable error message when these cases are encountered.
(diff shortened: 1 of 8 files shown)


pkg/streamingpromql/engine_test.go
```diff
@@ -49,9 +49,7 @@ func TestUnsupportedPromQLFeatures(t *testing.T) {
 
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
 	// different cases and make sure we produce a reasonable error message when these cases are encountered.
-	unsupportedExpressions := map[string]string{
-		"absent_over_time(nonexistent{}[1h])": "'absent_over_time' function",
-	}
+	unsupportedExpressions := map[string]string{}
```
charleskorn, line 52: Should we add `holt_winters` here until we add support for that? Or perhaps an experimental function?

pkg/streamingpromql/functions.go
```diff
@@ -96,6 +96,23 @@ func AbsentFunctionOperatorFactory(args []types.Operator, memoryConsumptionTrack
 	return o, nil
 }
 
+func AbsentOverTimeFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, argExpressions parser.Expressions) (types.InstantVectorOperator, error) {
+	functionName := "absent_over_time"
+	if len(args) != 1 && len(argExpressions) != 1 {
```
charleskorn, line 101: Should we fail if either does not contain exactly one element?

```suggestion
	if len(args) != 1 || len(argExpressions) != 1 {
```

pkg/streamingpromql/operators/functions/absent_over_time.go
```diff
@@ -0,0 +1,124 @@
...
+	for range innerMetadata {
+		err := a.inner.NextSeries(ctx)
+		if err != nil {
+			return nil, err
+		}
+		for {
+			step, err := a.inner.NextStepSamples()
+			// nolint:errorlint // errors.Is introduces a performance overhead, and NextStepSamples is guaranteed to return exactly EOS, never a wrapped error.
+			if err == types.EOS {
+				break
```
charleskorn, line 77: We could simplify this slightly by changing the `for` to something like `for stepIdx := range a.timeRange.StepCount`. We can then drop the special check for `err == types.EOS` (we should never receive an `EOS` for any of the steps, and if we do, we should return an error) and line 82 then doesn't need to call `PointIndex`.

pkg/streamingpromql/operators/functions/absent_over_time.go
```diff
@@ -0,0 +1,124 @@
...
+	output := types.InstantVectorSeriesData{}
+	if a.exhausted {
+		return output, types.EOS
+	}
+
+	a.exhausted = true
+
+	var err error
+	for step := range a.timeRange.StepCount {
+		t := a.timeRange.IndexTime(int64(step))
```
charleskorn, line 100: [nit] We only need to call `IndexTime` if we're going to emit a sample for this step, so this could move down beneath the `if`

pkg/streamingpromql/operators/functions/absent_over_time.go
```diff
@@ -0,0 +1,124 @@
...
+
+// AbsentOverTime performs a rate calculation over a range vector.
+type AbsentOverTime struct {
+	timeRange                types.QueryTimeRange
+	argExpressions           parser.Expr
+	inner                    types.RangeVectorOperator
+	expressionPosition       posrange.PositionRange
+	memoryConsumptionTracker *limiting.MemoryConsumptionTracker
+	presence                 []bool
+	exhausted                bool
```
charleskorn, line 27: [nit] We usually export some of these - could you please make this consistent with the other operators? The usual rule of thumb is that internal implementation details aren't exported.

charleskorn:
Could you please also add this to #10067?

pkg/streamingpromql/operators/functions/absent_over_time.go
lamida: Moved in b99205543f1e374dca07a0c3818e67f072037c9d

pkg/streamingpromql/operators/functions/absent_over_time.go
lamida: Updated in 4726cddd0f329867b1d4b3b1acc803be52e8fb17

pkg/streamingpromql/operators/functions/absent_over_time.go
lamida: Updated in dcd5239fdf2cf631c87b0f8dcbc8d807cc257e4f

pkg/streamingpromql/engine_test.go
lamida: Updated in d646f61089ed0868dd997ed45790c881f679cfb8. Let me know if something is incorrect there.

lamida:
@charleskorn all of your PR feedbacks has already been addressed. Please take another look. I am also still checking the failure in test with `out of order sample` error. 🤔

pkg/streamingpromql/engine_test.go
```diff
@@ -47,10 +48,19 @@ func init() {
...
+	parser.Functions["holt_winters"].Experimental = false
+
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
 	// different cases and make sure we produce a reasonable error message when these cases are encountered.
 	unsupportedExpressions := map[string]string{
-		"absent_over_time(nonexistent{}[1h])": "'absent_over_time' function",
+		// expected error for holt_winters is the same as double_exponential_smoothing because mimir just aliasing it as above
+		"holt_winters(metric{}[1h], 1, 1)": "'double_exponential_smoothing' function",
+
+		"double_exponential_smoothing(metric{}[1h], 1, 1)": "'double_exponential_smoothing' function",
```
charleskorn, line 63: I think adding just `double_exponential_smoothing` here is sufficient - no need to add `holt_winters` (and the aliasing above) as well.

lamida: Updated in 54f36f4243488cb8e9ce656d9d76d5e75602a256

charleskorn (approved):

Issue #10847, opened 2025-03-10
Bug: OTLP responses should match content-type of requests


fionaliao:
### What is the bug?

From the OTLP spec:

> The server MUST set “Content-Type: application/x-protobuf” header if the response body is binary-encoded Protobuf payload. The server MUST set “Content-Type: application/json” if the response is JSON-encoded Protobuf payload. The server MUST use the same “Content-Type” in the response as it received in the request.

Currently the the response format does not change depending on request format.
No content-type and a blank response for success is returned for success, and `Content-Type: application/octet-stream` + binary-encoded proto for errors:
https://github.com/grafana/mimir/blob/ba300f5950b2d78bc0cf37f06822bcda91865339/pkg/distributor/otel.go#L342

JSON requests should have `Content-Type: application/json` set in responses, with the responses following the [JSON protobuf encoding](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding), while protobuf requests should have `Content-Type: application/x-protobuf` set in responses (for the protobuf error case, the response body is fine but the Content-Type should be set correctly)

### How to reproduce it?

A request like

```
curl -i -XPOST -H 'Content-Type: application/json' <mimir endpoint> -d '{
  "resourceMetrics": [
    {
      "scopeMetrics": [
        {
          "metrics": [
            {
              "name": "test_metric",
              "unit": "s",
              "description": "",
              "gauge": {
                "dataPoints": [
                  {
                    "asInt": 1,
                    "timeUnixNano": 1741602284008000000,
                    "attributes": [
                      {
                        "key": "bar_label",
                        "value": {
                          "stringValue": "abc"
                        }
                      }
                    ]
                  }
                ]
              }
            }
          ]
        }
      ]
    }
  ]
}' 
```

returns a 200 success without a content-type. For a valid json response `{}` should be returned. Similar issues apply to the other response combinations (success/error + request format)

### What did you think would happen?

A response with `Content-Type: application/json` and body `{}`

### What was your environment?

Issue can be seen in the current code (main currently at commit https://github.com/grafana/mimir/commit/20548c3829fde9e7736d1f65ca428f3ea76d3530)

aknuds1:
Thanks for reporting this :)

aknuds1:
@fionaliao have agreed that we need to seek spec clarification on whether errors should be JSON encoded if the request has this Content-Type. The spec is currently ambiguous on this point.

santileira:
[...]

PR #10852, opened 2025-03-10
OTLP: Respond according to spec

aknuds1:
#### What this PR does

Respond to OTLP writes according to [spec](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response).

##### Success case with lossless conversion

Respond with either protobuf or JSON encoded `ExportMetricsServiceResponse` (depending on request's `Content-Type` header), and with correct `Content-Type` header.

##### Failure case

Respond with either protobuf or JSON encoded `rpc.Status` (depending on request's `Content-Type` header), and with correct `Content-Type` header.

Solves part of https://github.com/grafana/mimir/issues/10847.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -87,6 +87,7 @@
 * [BUGFIX] Ruler: fix issue where rule evaluations could be missed while shutting down a ruler instance if that instance owns many rule groups. prometheus/prometheus#15804 #10762
 * [BUGFIX] Ingester: Add additional check on reactive limiter queue sizes. #10722
 * [BUGFIX] TSDB: fix unknown series errors and possible lost data during WAL replay when series are removed from the head due to inactivity and reappear before the next WAL checkpoint. https://github.com/prometheus/prometheus/pull/16060 #10824
+* [BUGFIX] OTLP: Fix response body and Content-Type header, to align with spec. #10852
```
tacole02, line 90:
```suggestion
* [BUGFIX] OTLP: Fix response body and Content-Type header to align with spec. #10852
```

pkg/distributor/otel.go
```diff
@@ -109,43 +111,60 @@ func OTLPHandler(
...
+					w.Header().Set("Content-Type", contentType)
+					data, _ := json.Marshal(&expResp)
+					_, _ = w.Write([]byte(data))
+				} else if contentType == pbContentType {
+					w.Header().Set("Content-Type", contentType)
+					data, _ := proto.Marshal(&expResp)
+					_, _ = w.Write(data)
+				} else {
+					level.Warn(logger).Log("msg", "unrecognized Content-Type request header", "content_type", contentType)
 				}
```
fionaliao, line 132: I think we could remove the if statement for the protobuf case, and just assume if it's not json then the content-type is protobuf, similar to what happens for the error case. The request content type is checked in the function returned for `newOTLPParser` so we can be sure it's only json or protobuf by the time we're returning a response

```suggestion
					_, _ = w.Write([]byte(data))
				} else {
					w.Header().Set("Content-Type", contentType)
					data, _ := proto.Marshal(&expResp)
					_, _ = w.Write(data)
				}
```

If we do end up keeping the unrecogised content-type case though, we should make that return an error.

fionaliao (commented):
One comment about handling the unknown content-type case, otherwise looks good

pkg/distributor/otel.go
aknuds1: Thanks, good point.

fionaliao (approved):
LTGM

pkg/distributor/otel.go
```diff
@@ -335,27 +352,44 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+		}
+	} else {
+		respBytes, err = st.Marshal()
+		if err != nil {
+			level.Error(logger).Log("msg", "otlp response marshal failed", "err", err)
+			writeResponseFailedBody, _ := proto.Marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
+			_, _ = w.Write(writeResponseFailedBody)
+			return
+		}
 	}
```
charleskorn, line 390: Could we remove some of this duplication by introducing the concept of a codec or something like that? Then we could have one codec for JSON and another for Protobuf, resolve the codec in one place, and then we wouldn't need so many conditions based on the content type here and above.

pkg/distributor/otel.go
```diff
@@ -109,43 +111,58 @@ func OTLPHandler(
...
-				if httpCode/100 == 4 {
-					msgs = append(msgs, "insight", true)
+				// Respond as per spec:
+				// https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response.
+				var expResp colmetricpb.ExportMetricsServiceResponse
+				contentType := r.Header.Get("Content-Type")
+				if contentType == jsonContentType {
+					w.Header().Set("Content-Type", contentType)
+					data, _ := json.Marshal(&expResp)
+					_, _ = w.Write([]byte(data))
```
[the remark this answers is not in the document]
aknuds1: I'm surprised the linter let this pass.

pkg/distributor/otel.go
```diff
@@ -109,43 +111,58 @@ func OTLPHandler(
...
+				} else {
+					w.Header().Set("Content-Type", contentType)
```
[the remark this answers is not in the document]
aknuds1: Yes, good point.

pkg/distributor/otel.go
aknuds1: Please check my refactoring.

charleskorn: I think we could take this one step further: can we pass the response payload (eg. `st` here) to `writeOTLPResponse`, and then it can call `json.Marshal` or `proto.Marshal`? Then we shouldn't need to pass a function to `writeOTLPResponse`.

aknuds1: I did some more refactoring, PTAL.

pkg/distributor/otel.go
```diff
@@ -335,28 +342,61 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+		}
+	}
 	if err != nil {
-		level.Error(logger).Log("msg", "write error to otlp response failed", "err", err)
+		level.Error(logger).Log("msg", "OTLP response marshal failed, responding without payload", "err", err, "content_type", contentType)
+		contentType = "text/plain"
+		body = nil
+	}
+
+	w.Header().Set("Content-Type", contentType)
```
charleskorn, line 393: Do we also need to set `Content-Length`?

tacole02 (approved):

pkg/distributor/otel.go
```diff
@@ -335,28 +342,61 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+			level.Error(logger).Log("msg", "failed to marshal payload to JSON", "err", err, "payload", payload)
+			body, err = json.Marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
+			err = errors.Wrapf(err, "marshalling %T to JSON", payload)
+		}
+	} else {
+		body, err = proto.Marshal(payload)
+		if err != nil {
+			httpCode = http.StatusInternalServerError
+			level.Error(logger).Log("msg", "failed to marshal payload to protobuf", "err", err, "payload", payload)
+			body, err = proto.Marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
```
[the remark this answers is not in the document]
aknuds1: I don't think you're right on this one. `err` will only be non-nil after this branch if line 383 returns an error. Same for the other branch.

pkg/distributor/otel.go
```diff
@@ -335,28 +342,61 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+		if err != nil {
+			httpCode = http.StatusInternalServerError
+			level.Error(logger).Log("msg", "failed to marshal payload to JSON", "err", err, "payload", payload)
+			body, err = json.Marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
+			err = errors.Wrapf(err, "marshalling %T to JSON", payload)
+		}
+	} else {
+		body, err = proto.Marshal(payload)
+		if err != nil {
+			httpCode = http.StatusInternalServerError
```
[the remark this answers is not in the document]
aknuds1: No can do, since the `if` condition on line 387 is only true if `err` is non-nil. Normally, line 383 should assign nil to `err`.

pkg/distributor/otel.go
```diff
@@ -335,28 +342,61 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+			level.Error(logger).Log("msg", "failed to marshal payload to protobuf", "err", err, "payload", payload)
```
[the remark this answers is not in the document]
aknuds1: Line 388 only executes (logs) if fallback marshaling (lines 375/383) _also_ fails, so we can't rely on the logging on line 388.

pkg/distributor/otel.go
aknuds1: The OTLP spec doesn't say anything about Content-Length. As such, let's keep it out of the scope of this PR?

tacole02 (approved):

pkg/distributor/otel.go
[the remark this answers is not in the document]
charleskorn: Ah, I missed that, thanks.

pkg/distributor/otel.go
```diff
@@ -335,28 +342,61 @@ func httpRetryableToOTLPRetryable(httpStatusCode int) int {
...
+			body, err = proto.Marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
+			err = errors.Wrapf(err, "marshalling %T to protobuf", payload)
+		}
+	}
```
charleskorn, line 386: I think there's scope to simplify this further. If we define a function that marshals to JSON or Protobuf like this...

```golang
func marshal(payload proto.Message, contentType string) ([]byte, error) {
	if contentType == jsonContentType {
		return json.Marshal(payload)
	}

	return proto.Marshal(payload)
}
```

...then this could become something like this:

```suggestion
	body, err := marshal(payload, contentType)

	if err != nil {
		httpCode = http.StatusInternalServerError
			level.Error(logger).Log("msg", "failed to marshal payload", "err", err, "payload", payload, "content_type", contentType)
			body, err = marshal(status.New(codes.Internal, "failed to marshal OTLP response").Proto())
			err = errors.Wrapf(err, "marshalling %T as %v", payload, contentType)
	}
```

charleskorn (approved):

pkg/distributor/otel.go
charleskorn: It's good practice to set it where possible, it allows clients to size buffers appropriately etc.

pkg/distributor/otel.go
aknuds1: Good idea, I've made a practically equivalent revision.

pkg/distributor/otel.go
aknuds1: I'll do it in a follow-up, to separate changes.

PR #10863, opened 2025-03-12
Catch up faster when an idle partition becomes active after some time

codesome:
#### What this PR does

Currently when a partition was active and goes idle, its kafka commit point stays in the past. Once that partition gets active, block builder will try to consume every 1h interval since the last commit.

For example if the last commit was on 2025-02-15 00:00 before it went idle, and then later it became active on 2025-03-05, then it would try consuming every hours since Feb 15 to Mar 5, which takes forever to catch up. 

This PR shortcuts this and once it finds that the first record is in the 'future', it shortcuts to the first consumable record or the latest consumption cycle - whichever is earlier.

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -254,7 +254,7 @@ func (b *BlockBuilder) runningStandaloneMode(ctx context.Context) error {
 	// Examples for interval=1h, buffer=15m:
 	//   (1) current time is 14:12, cycle end is 13:15
 	//   (2) current time is 14:17, cycle end is 14:15
-	cycleEndTime := cycleEndAtStartup(time.Now(), b.cfg.ConsumeInterval, b.cfg.ConsumeIntervalBuffer)
+	cycleEndTime := cycleEndBefore(time.Now(), b.cfg.ConsumeInterval, b.cfg.ConsumeIntervalBuffer)
 	var waitDur time.Duration
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -450,15 +459,39 @@ func (b *BlockBuilder) consumePartition(ctx context.Context, partition int32, st
...
 		logger := log.With(logger, "section_end", sectionEndTime, "offset", state.Commit.At)
 		state, err = b.consumePartitionSection(ctx, logger, builder, partition, state, sectionEndTime, cycleEndOffset)
-		if err != nil {
+		if err == nil {
+			sectionEndTime = sectionEndTime.Add(b.cfg.ConsumeInterval)
+			continue
+		}
+
+		var recInFutureErr *errFirstRecordInFuture
+		if !errors.As(err, &recInFutureErr) {
```
seizethedave, line 468: You may have tried this, I think this passage would read easier if it were something like:

```
if err != nil {
   if it's an InFuture error {
      err = nil
      ...
   }
   return PartitionState{}, fmt.Errorf("consume partition %d: %w", partition, err)
} else {
   sectionEndTime = sectionEndTime.Add(b.cfg.ConsumeInterval)
}
```

pkg/blockbuilder/blockbuilder.go
```diff
@@ -489,7 +522,8 @@ func (b *BlockBuilder) consumePartitionSection(
 
 		dur := time.Since(t)
 
-		if retErr != nil {
+		var temp *errFirstRecordInFuture
+		if retErr != nil && !errors.As(retErr, &temp) {
```
seizethedave, line 526: {nit}
`if !errors.Is()` if you don't need the var.

seizethedave (approved):
LGTM%c!

pkg/blockbuilder/blockbuilder.go
codesome: Thanks! Did this with a small change - I kept the nested `if` as `if it's *not* an InFuture error {` in the spirit of keeping the `if` small, and also to avoid `else` since there it no return if its an InFuture error.

PR #10869, opened 2025-03-12
Allow cluster validation for gRPC clients

duricanikolic:
#### What this PR does
This PR upgrades dskit to its latest commit, [4320c75](https://github.com/grafana/dskit/commit/4320c75000d96ae6636f6c1ec48447854458f8c9), where `grpcclient.Config` has been extended with a possibility to configure the cluster validation label. Assuming that a gRPC client configuration is reached via `-<grpc-client-config-path>`, the cluster validation label can be configured via `-<grpc-client-config-path>.cluster-validation.label`. For example, in case of ingester clients, the cluster validation labels are configured via `-ingester.client.cluster-validation.label`. 

When a gRPC client is configured with a cluster validation label, a call to `grpcclient.Config.DialOption()` implicitly injects the cluster unary client interceptor among returned `grpc.DialOption`s, meaning that the corresponding client starts sending the configured cluster validation label through its requests.

Additionally, the signature of dskit's `grpcclient.DialOption()` changed from
```
func (cfg *Config) DialOption([]grpc.UnaryClientInterceptor, []grpc.StreamClientInterceptor) ([]grpc.DialOption, error)
```
to
```
func (cfg *Config) DialOption([]grpc.UnaryClientInterceptor, []grpc.StreamClientInterceptor, middleware.InvalidClusterValidationReporter) ([]grpc.DialOption, error)
```

This made all the current usages of `grpcclient.DialOption()` invalid. This PR fixes that problem.

Last but not least, this PR removes `util.RegisteredFlags` and all its usages, because it is now replaced by dskit's `flagext.RegisteredFlags`.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/alertmanager/alertmanager_client.go
```diff
@@ -104,7 +105,8 @@ func (f *alertmanagerClientsPool) GetClientFor(addr string) (Client, error) {
 // dialAlertmanagerClient establishes a GRPC connection to an alertmanager that is aware of the the health of the server
 // and collects observations of request durations.
 func dialAlertmanagerClient(cfg grpcclient.Config, inst ring.InstanceDesc, requestDuration *prometheus.HistogramVec) (*alertmanagerClient, error) {
-	opts, err := cfg.DialOption(grpcclient.Instrument(requestDuration))
+	unary, stream := grpcclient.Instrument(requestDuration)
+	opts, err := cfg.DialOption(unary, stream, middleware.NoOpInvalidClusterValidationReporter)
```
colega, line 109: Why are we using noop here?

colega: To avoid commenting this on each one of the noops, can you explain each one? I would expect to see no noop usages.

colega:
I would rename the PR's title. That's what people will see in the commit list, and they will just see "Upgrade to latest dskit".

I would call the PR something like "Add cluster validation to grpc clients", and not mention dskit in the PR title because the that is just _how_ you do it, it's a _minor implementation detail_ of the change.

pkg/alertmanager/alertmanager_client.go
duricanikolic: @colega this is simply because I wanted to do it in 2 steps:
- the first step (to allow further dskit upgrades) is where we allow the cluster validation
- the second step is where we create all client-related metrics and allow tracking of invalid cluster validations

The second step would be merged anyway between today and Monday, depending on the review.
WDYT?

colega: I find it harder to review changes on already merged changes, because then I lose the context of the entire change. What is new? What is being cleaned up from previous one?

Plus, merging this as is actually leaves it in a non-fully-working state.

If you really want to split into multiple PRs, then I would suggest to send a PR to be merged into this PR, review that, merge it here, and then merge this full PR.

pkg/storage/bucket/client.go
```diff
@@ -72,7 +73,7 @@ type StorageBackendConfig struct {
 	ExtraBackends []string `yaml:"-"`
 
 	// Used to keep track of the flag names registered in this config, to be able to overwrite them later properly.
-	RegisteredFlags util.RegisteredFlags `yaml:"-"`
+	TrackedRegisteredFlags flagext.RegisteredFlags `yaml:"-"`
```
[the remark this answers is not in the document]
duricanikolic: I have addressed this in f5fad4acee61ad186276438927f6b2f1c46dd1d1.

pkg/util/metrics.go
```diff
@@ -26,3 +29,10 @@ func NewRequestInvalidClusterValidationLabelsTotalCounter(reg prometheus.Registe
 		},
 	}, []string{methodLabel})
 }
+
+func NewInvalidClusterValidationReporter(cluster string, invalidClusterValidations *prometheus.CounterVec, logger log.Logger) middleware.InvalidClusterValidationReporter {
+	return func(msg string, method string) {
+		level.Warn(logger).Log("msg", msg, "method", method, "clusterValidationLabel", cluster)
```
aknuds1, line 35: Our convention is snake case in for log message keys:
```suggestion
		level.Warn(logger).Log("msg", msg, "method", method, "cluster_validation_label", cluster)
```

CHANGELOG.md
```diff
@@ -59,14 +59,15 @@
   * `go_cpu_classes_gc_total_cpu_seconds_total`
   * `go_cpu_classes_total_cpu_seconds_total`
   * `go_cpu_classes_idle_cpu_seconds_total`
-* [ENHANCEMENT] All: Add support for cluster validation in gRCP calls. When it is enabled, gRPC server verifies if a request coming from a gRPC client comes from an expected cluster. This validation can be configured by the following experimental configuration options: #10767
+* [ENHANCEMENT] All: Add experimental support for cluster validation in gRCP calls. When it is enabled, gRPC server verifies if a request coming from a gRPC client comes from an expected cluster. This validation can be configured by the following experimental configuration options: #10767
```
aknuds1, line 62: Typo:
```suggestion
* [ENHANCEMENT] All: Add experimental support for cluster validation in gRPC calls. When it is enabled, gRPC server verifies if a request coming from a gRPC client comes from an expected cluster. This validation can be configured by the following experimental configuration options: #10767
```

aknuds1 (commented):
Generally LPGTM, please see suggestions.

colega (approved):
LGTM 👌 

Left one nit about docs structure.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1583,6 +1584,11 @@ store_gateway_client:
   # CLI flag: -querier.store-gateway-client.tls-min-version
   [tls_min_version: <string> | default = ""]
 
+  cluster_validation:
+    # (experimental) Optionally define the cluster validation label.
+    # CLI flag: -querier.store-gateway-client.cluster-validation.label
+    [label: <string> | default = ""]
```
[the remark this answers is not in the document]
colega: Nevermind, I see it's not a new issue. I the entire `grpc_client` should've been extracted to a separate block a while ago, and it was, but it wasn't removed from the specific locations.

aknuds1 (approved):
LGTM

PR #10873, opened 2025-03-13
Add double_exponential_smoothing MQE function

lamida:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1183,3 +1183,19 @@ eval range from 0s to 66m step 6m absent_over_time(histo{case=~"(0|a)"}[10m])
...
 
 clear
+
+# Tests for double_exponential_smoothing to trigger circular buffer boundary wrap around
+
+load 10s
+  http_requests{job="api-server", instance="0", group="production"}	8000-10x1000
+  http_requests{job="api-server", instance="1", group="production"}	0-20x1000
+  http_requests{job="api-server", instance="0", group="canary"}		0+30x1000 300-80x1000
+  http_requests{job="api-server", instance="1", group="canary"}		0-40x1000 0+40x1000
```
charleskorn, line 1193: Could you please add some test cases where the input contains NaN or Inf values?

pkg/streamingpromql/functions.go
```diff
@@ -709,3 +710,38 @@ func unaryNegationOfInstantVectorOperatorFactory(inner types.InstantVectorOperat
 	o := functions.NewFunctionOverInstantVector(inner, nil, memoryConsumptionTracker, f, expressionPosition, timeRange)
 	return operators.NewDeduplicateAndMerge(o, memoryConsumptionTracker)
 }
+
+func DoubleExponentialSmoothingFunctionOperatorFactory(args []types.Operator, memoryConsumptionTracker *limiting.MemoryConsumptionTracker, annotations *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange, argExpressions parser.Expressions) (types.InstantVectorOperator, error) {
+	f := functions.DoubleExponentialSmoothing
+
+	functionName := "double_exponential_smoothing"
+	if len(args) != 3 || len(argExpressions) != 3 {
```
charleskorn, line 718: Do we care about `argExpressions` here?

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,95 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+func doubleExponentialSmoothing(step *types.RangeVectorStepData, _ float64, args []types.ScalarData, timeRange types.QueryTimeRange, emitAnnotation types.EmitAnnotationFunc, _ *limiting.MemoryConsumptionTracker) (float64, bool, *histogram.FloatHistogram, error) {
+	smoothingFactor := args[0].Samples[timeRange.PointIndex(step.StepT)].F
+	if smoothingFactor <= 0 || smoothingFactor >= 1 {
+		return 0, false, nil, fmt.Errorf("invalid smoothing factor. Expected: 0 < sf < 1, got: %f", smoothingFactor)
+	}
+
+	trendFactor := args[1].Samples[timeRange.PointIndex(step.StepT)].F
+	if trendFactor <= 0 || trendFactor >= 1 {
+		return 0, false, nil, fmt.Errorf("invalid trend factor. Expected: 0 < tf < 1, got: %f", trendFactor)
+	}
```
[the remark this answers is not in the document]
lamida: In prometheus it is panic

```
panic(fmt.Errorf("invalid smoothing factor. Expected: 0 < sf < 1, got: %f", sf))
```

But in the engine the panic is caught in recover which I believe turn them back into err. Definitely they don't use annotation here.

pkg/streamingpromql/engine_test.go
```diff
@@ -304,6 +299,9 @@ func TestNewInstantQuery_Strings(t *testing.T) {
 // Test cases that are not supported by the streaming engine are commented out (or, if the entire file is not supported, .disabled is appended to the file name).
 // Once the streaming engine supports all PromQL features exercised by Prometheus' test cases, we can remove these files and instead call promql.RunBuiltinTests here instead.
 func TestUpstreamTestCases(t *testing.T) {
+	// Enable experimental functions testing
+	parser.Functions["double_exponential_smoothing"].Experimental = false
```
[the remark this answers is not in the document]
lamida: Good point. Moved in d1563bd9f7932b23fa81b4a821b086deee2bf11b

pkg/streamingpromql/testdata/ours/functions.test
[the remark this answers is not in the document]
lamida: 0d4efe324419490f29d005a9cf342236ab5cbcef

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,95 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+
+	fHead, fTail := step.Floats.UnsafePoints()
+
+	if step.Floats.Any() && step.Histograms.Any() {
+		emitAnnotation(annotations.NewHistogramIgnoredInMixedRangeInfo)
```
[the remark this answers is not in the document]
lamida: ade72b2b480f84a487f0dc63c2a9adcf3908d66f

pkg/streamingpromql/operators/functions/range_vectors.go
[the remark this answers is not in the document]
lamida: 0113d89a29707df8e0b6b53617fb739b0a15e10c

pkg/streamingpromql/functions.go
lamida: 0113d89a29707df8e0b6b53617fb739b0a15e10c

pkg/streamingpromql/testdata/ours/functions.test
lamida: 6bdfe23894ddbdf5c7bda875a3288fa8299d35d5

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,95 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+
+	accumulate := func(samples []promql.FPoint, head bool) {
+		var x, y float64
+		var startLoop int
+
+		// We start looping from index 1 for head, but for tail,
+		// we are resuming the looping, hence we start just from 0
+		if head {
+			startLoop = 1
+		}
```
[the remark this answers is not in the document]
lamida: Hopefully nothing breaks, but this is the refactored code as the suggestion ecabe075c68c97d2e62c107d262bde52828d021f

pkg/streamingpromql/engine_test.go
```diff
@@ -39,23 +38,21 @@ import (
...
 
-	// Disable experimental so that parser will parse it without the updating experiemental flag
-	parser.Functions["double_exponential_smoothing"].Experimental = false
-
 	// The goal of this is not to list every conceivable expression that is unsupported, but to cover all the
 	// different cases and make sure we produce a reasonable error message when these cases are encountered.
-	unsupportedExpressions := map[string]string{
-		"double_exponential_smoothing(metric{}[1h], 1, 1)": "'double_exponential_smoothing' function",
-	}
+	unsupportedExpressions := map[string]string{}
```
jhesketh, line 55: Should we place one of the other unsupported experimental functions in here?

pkg/streamingpromql/functions.go
```diff
@@ -570,75 +570,76 @@ func SortOperatorFactory(descending bool) InstantVectorFunctionOperatorFactory {
...
+	"cosh":                         InstantVectorTransformationFunctionOperatorFactory("cosh", functions.Cosh),
+	"count_over_time":              FunctionOverRangeVectorOperatorFactory("count_over_time", functions.CountOverTime),
+	"day_of_month":                 TimeTransformationFunctionOperatorFactory("day_of_month", functions.DayOfMonth),
+	"day_of_week":                  TimeTransformationFunctionOperatorFactory("day_of_week", functions.DayOfWeek),
+	"day_of_year":                  TimeTransformationFunctionOperatorFactory("day_of_year", functions.DayOfYear),
+	"days_in_month":                TimeTransformationFunctionOperatorFactory("days_in_month", functions.DaysInMonth),
+	"deg":                          InstantVectorTransformationFunctionOperatorFactory("deg", functions.Deg),
+	"delta":                        FunctionOverRangeVectorOperatorFactory("delta", functions.Delta),
+	"deriv":                        FunctionOverRangeVectorOperatorFactory("deriv", functions.Deriv),
+	"double_exponential_smoothing": DoubleExponentialSmoothingFunctionOperatorFactory,
```
jhesketh, line 598: (nit) perhaps a comment that the parser translates `holt_winters` into `double_exponential_smoothing` so there is no need to register it separately.

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,155 @@
...
+				StartT:               0,
+				EndT:                 4000,
+				IntervalMilliseconds: 1000,
+				StepCount:            5,
+			}
+
+			gotValue, gotOk, gotHist, err := doubleExponentialSmoothing(step, 0, args, timeRange, nil, nil)
+
+			// Check error
+			if (err != nil) != tt.wantErr {
```
jhesketh, line 124: (nit) Generally easier to use `require.True` or `require.Equal` etc. for these comparisions

pkg/streamingpromql/engine_test.go
```diff
@@ -39,23 +38,21 @@ import (
 	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
 	"github.com/grafana/mimir/pkg/streamingpromql/types"
 	"github.com/grafana/mimir/pkg/util/globalerror"
+	"github.com/grafana/mimir/pkg/util/promqlext"
 )
 
 func init() {
 	types.EnableManglingReturnedSlices = true
+	// To test double_exponential_smoothing and holt_winters
+	promqlext.ExtendPromQL()
```
charleskorn, line 47: This could go in an `init` function for the whole `streamingpromql` package, not just the tests.

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"math"
+	"testing"
+
+	"github.com/prometheus/prometheus/promql"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func TestDoubleExponentialSmoothing_Validation(t *testing.T) {
```
charleskorn, line 15: There's no need to test this with a Go test: we can make assertions about errors in `.test` files with `expected_fail_msg`.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,85 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+
+	accumulate := func(samples []promql.FPoint, head bool) {
+		var x, y float64
+
+		for i, sample := range samples {
+			// Scale the raw value against the smoothing factor.
+			x = smoothingFactor * sample.F
+
+			// Scale the last smoothed value for all samples except the first one in head.
+			if head && i != 0 || !head {
```
charleskorn, line 877: If we're calling `accumulate` with `fHead[1:]`, do we need this condition?

pkg/streamingpromql/operators/functions/range_vectors.go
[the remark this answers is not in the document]
jhesketh: Can you also please add a test for when there are only native histograms?

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh, line 877: This seems odd to me. What if `len(fHead)` is `1` so the initial `estimatedTrend` used the tail? Is this still correct?

Looking at the upstream code, the `calcTrendValue` is applied to each sample after the initial values are set.

I think we should always do `calcTrendValue` and pass in the correct slices to `accumulate`.

For example, I think `accumulate(fTail, false)` below is incorrect for when we use `fTail[0].F` in the initial `estimatedTrend`.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,85 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+				estimatedTrend = calcTrendValue(trendFactor, smooth0, smooth1, estimatedTrend)
+			}
+			y = (1 - smoothingFactor) * (smooth1 + estimatedTrend)
+
+			smooth0, smooth1 = smooth1, x+y
+		}
+	}
+	accumulate(fHead[1:], true)
+	accumulate(fTail, false)
```
jhesketh, line 886: perhaps something like?
```suggestion
	accumulate(fHead[1:])
	accumulate(fTail[fTailPos:0])
```
where fTailPos is set to either `0` or `1` at the initial value part.

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1183,3 +1183,31 @@ eval range from 0s to 66m step 6m absent_over_time(histo{case=~"(0|a)"}[10m])
...
+  {job="api-server", instance="1", group="canary"} _ -8000 -16000 -24000 -32000 -40000 -6689.664952482304 10592.070195990285 23960 31960 39960
+  {job="api-server", instance="2", group="canary"} _ -8000 -16000 -24000 -32000 -40000 NaN NaN 23800 31800 39800
+
+eval range from 0 to 20000s step 2000s holt_winters(http_requests[5000s], 0.01, 0.1)
+  {job="api-server", instance="0", group="production"} _ 6000 4000 2000 0 -2000 -2000 -2000
+  {job="api-server", instance="1", group="production"} _ -4000 -8000 -12000 -16000 -20000 -20000 -20000
+  {job="api-server", instance="2", group="production"} _ -4000 -8000 -12000 -16000 NaN NaN NaN
+  {job="api-server", instance="0", group="canary"} _ 6000 12000 18000 24000 30000 -4774.4112298142 -27650.83746192864 -47620 -63620 -79620
+  {job="api-server", instance="1", group="canary"} _ -8000 -16000 -24000 -32000 -40000 -6689.664952482304 10592.070195990285 23960 31960 39960
+  {job="api-server", instance="2", group="canary"} _ -8000 -16000 -24000 -32000 -40000 NaN NaN 23800 31800 39800
```
jhesketh, line 1211: What about some tests where the scalars change over time and have a mix of valid and invalid values?

pkg/streamingpromql/engine_test.go
```diff
@@ -3217,6 +3219,7 @@ func TestCompareVariousMixedMetricsVectorSelectors(t *testing.T) {
 
 		expressions = append(expressions, fmt.Sprintf(`predict_linear(series{label=~"(%s)"}[1m], 30)`, labelRegex))
 		expressions = append(expressions, fmt.Sprintf(`quantile_over_time(scalar(series{label="i"}), series{label=~"(%s)"}[1m])`, labelRegex))
+		expressions = append(expressions, fmt.Sprintf(`double_exponential_smoothing(series{label=~"(%s)"}[1m], 0.01, 0.1)`, labelRegex))
```
jhesketh, line 3222: Could add another test with `scalar(series{label="f"})` and  `scalar(series{label="i"})`  for the values

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Yes we need. estimatedTrend is only calculated after the first sample.

Failed test if we remove this.

```
=== RUN   TestOurTestCases/ours/functions.test/Mimir's_engine/line_1197/double_exponential_smoothing(http_requests[5000s],_0.01,_0.1)
    /Users/lamida/github/grafana/mimir/pkg/streamingpromql/test.go:1141:
        	Error Trace:	/Users/lamida/github/grafana/mimir/vendor/github.com/prometheus/prometheus/promql/promqltest/test.go:1141
        	Error:      	Received unexpected error:
        	            	error in eval double_exponential_smoothing(http_requests[5000s], 0.01, 0.1) (line 1197): expected float value at index 0 (t=2000000) for {group="canary", instance="0", job="api-server"} to be 6000, but got 6000.78975612423 (result has 10 float points [6000.78975612423 @[2000000] 12000.578004600346 @[4000000] 18068.180964039137 @[6000000] 24204.542892117348 @[8000000] 30340.904820195698 @[10000000] -4297.144481540203 @[12000000] -27037.208785576273 @[14000000] -47793.17964865944 @[16000000] -64156.811456868294 @[18000000] -80520.44326507693 @[20000000]] and 0 histogram points [])
        	Test:       	TestOurTestCases/ours/functions.test/Mimir's_engine/line_1197/double_exponential_smoothing(http_requests[5000s],_0.01,_0.1)
=== RUN   TestOurTestCases/ours/functions.test/Mimir's_engine/line_1205/holt_winters(http_requests[5000s],_0.01,_0.1)
    /Users/lamida/github/grafana/mimir/pkg/streamingpromql/test.go:1141:
        	Error Trace:	/Users/lamida/github/grafana/mimir/vendor/github.com/prometheus/prometheus/promql/promqltest/test.go:1141
        	Error:      	Received unexpected error:
        	            	error in eval holt_winters(http_requests[5000s], 0.01, 0.1) (line 1205): expected float value at index 0 (t=2000000) for {group="canary", instance="0", job="api-server"} to be 6000, but got 6000.78975612423 (result has 10 float points [6000.78975612423 @[2000000] 12000.578004600346 @[4000000] 18068.180964039137 @[6000000] 24204.542892117348 @[8000000] 30340.904820195698 @[10000000] -4297.144481540203 @[12000000] -27037.208785576273 @[14000000] -47793.17964865944 @[16000000] -64156.811456868294 @[18000000] -80520.44326507693 @[20000000]] and 0 histogram points [])
        	Test:       	TestOurTestCases/ours/functions.test/Mimir's_engine/line_1205/holt_winters(http_requests[5000s],_0.01,_0.1)
```

pkg/streamingpromql/engine_test.go
lamida: Updated in e7514ed403cf7987e3c4b42669317cf340430223 and dc1a3f2af27d5927646c08ff2b6c55c44d49a3b7

pkg/streamingpromql/operators/functions/range_vectors_test.go
lamida: Removed the go test and add expted_fail_message promql.test fe571cbc0242191f5b4db79dfee7af3c69f8a1be

pkg/streamingpromql/engine_test.go
lamida: Should like this be good enough? 1e70c324bd48135a3c2b8dc8373bcd41861614f0

pkg/streamingpromql/functions.go
lamida: 2de690cdb2ebc607c43e651eb5a77c7b80b905bc

pkg/streamingpromql/operators/functions/range_vectors_test.go
lamida: This test is removed in fe571cbc0242191f5b4db79dfee7af3c69f8a1be

pkg/streamingpromql/operators/functions/range_vectors.go
charleskorn: Won't this skip the second sample in `fHead` though? `accumulate` is called with `fHead[1:]`, so the first iteration through this loop will have `i == 0` and `sample` will be `fHead[1]`.

pkg/streamingpromql/functions.go
```diff
@@ -570,31 +570,32 @@ func SortOperatorFactory(descending bool) InstantVectorFunctionOperatorFactory {
...
+	"cosh":             InstantVectorTransformationFunctionOperatorFactory("cosh", functions.Cosh),
+	"count_over_time":  FunctionOverRangeVectorOperatorFactory("count_over_time", functions.CountOverTime),
+	"day_of_month":     TimeTransformationFunctionOperatorFactory("day_of_month", functions.DayOfMonth),
+	"day_of_week":      TimeTransformationFunctionOperatorFactory("day_of_week", functions.DayOfWeek),
+	"day_of_year":      TimeTransformationFunctionOperatorFactory("day_of_year", functions.DayOfYear),
+	"days_in_month":    TimeTransformationFunctionOperatorFactory("days_in_month", functions.DaysInMonth),
+	"deg":              InstantVectorTransformationFunctionOperatorFactory("deg", functions.Deg),
+	"delta":            FunctionOverRangeVectorOperatorFactory("delta", functions.Delta),
+	"deriv":            FunctionOverRangeVectorOperatorFactory("deriv", functions.Deriv),
+	// We are calling promqlext.ExtendPromQL() in astmapper package, therefore we don't need to register holt_winters separately here.
```
charleskorn, line 598:
```suggestion
	// We are calling promqlext.ExtendPromQL() in streamingpromql package's init(), therefore we don't need to register holt_winters separately here.
```

pkg/streamingpromql/engine_test.go
lamida: sf and tf must be > 0 and < 0. Setting to labels i and labels f will fail that and throw the error. Can you suggest in more details?

pkg/streamingpromql/testdata/ours/functions.test
lamida: Do you want to varies `sf` and `tf` scalar arguments? Would you mind give more example? The sf and tf range area between 0 and 1 (open interval). And the validation already tested in 

```
eval_fail instant at 0s double_exponential_smoothing(http_requests[5000s], 0.01, 0)
  expected_fail_message invalid trend factor. Expected: 0 < tf < 1, got: 0.000000
```

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: > This seems odd to me. What if len(fHead) is 1 so the initial estimatedTrend used the tail? Is this still correct?

Initial estimation is handled in line aboves
```
	// Initial estimated trend is the difference between the first and second samples.
	// First sample will always be in the head.
	// If we have only one point in head, the second sample is the first index in tail.
	if len(fHead) == 1 {
		estimatedTrend = fTail[0].F - fHead[0].F
	} else {
		estimatedTrend = fHead[1].F - fHead[0].F
	}
```

> Looking at the upstream code, the calcTrendValue is applied to each sample after the initial values are set.

In prometheus calcTrendValue will be really computed on the third sample onward.

```
// this is how they call it
b = calcTrendValue(i-1, tf, s0, s1, b)

// this is the upstream function
func calcTrendValue(i int, tf, s0, s1, b float64) float64 {
	if i == 0 { // when i in the caller is on the second sample (i = 1), we just return b directly 
		return b
	}
 
        // We only calculate the trend from sample 3 (i =2) onward
	x := tf * (s1 - s0)
	y := (1 - tf) * b

	return x + y
}
```

> I think we should always do calcTrendValue and pass in the correct slices to accumulate.

With the explanation above, we really can do calcTrendValue on sample 3 onward

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: We need flag to pass to accumulate that tell the passed slice is a head. Because we need the head flag to tell when we can compute bt start from t >= 3.

image

Somehow in prometheus actually it computes b1 as b1 = 0 as I mentioned in another comment too.

```
// this is how they call it
b = calcTrendValue(i-1, tf, s0, s1, b)

// this is the upstream function
func calcTrendValue(i int, tf, s0, s1, b float64) float64 {
	if i == 0 { // when i in the caller is on the second sample (i = 1 or b1), we just return b directly 
		return b // so b1 = b0
	}
 
        // We only calculate the trend from sample 3 (i=2) onward
	x := tf * (s1 - s0)
	y := (1 - tf) * b

	return x + y
}
```

> `accumulate(fTail[fTailPos:0])` where fTailPos is set to either 0 or 1 at the initial value part.

I don't get this part. Can you elaborate more?

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: I replied to @jhesketh in another comment. But if I checked the prometheus code, it is indeed prometheus only compute trend bt only from 3rd sample onward.

b0 = x1 - x0
b1 = b0
image

The if branch condition above is to follow that setup to ensure we get the same result with Prometheus.

lamida:
@charleskorn @jhesketh thanks for the review. The implementation is following this [wikipedia formula](https://en.wikipedia.org/wiki/Exponential_smoothing) but also I try to follow the Prometheus upstream implementation.

image

Several things to note are the following points: 

1. When we initialize b0 which depends on where are samples between head or tail of the buffer instead of just doing b0=x1-x0. But this is pretty straightforward.
2. When we start calculate bt. In prometheus as I mentioned in some comments above, bt is calculated only from 3rd sample onward.

Seeing the test are passing for both upstream and "ours" test, I am pretty sure the implementation is already correct. Please suggest some test data that can prove this wrong.

lamida:
@jhesketh replying https://github.com/grafana/mimir/pull/10873#discussion_r1994648380 (not sure why I can't reply inline there.

> Can you also please add a test for when there are only native histograms?
There won't be any annotation emitted if samples are only native histogram.

The operator will just return.
```
	if (len(fHead) + len(fTail)) < 2 {
		return 0, false, nil, nil
	}
```

So I don't think there will be any annotation we can test there. We can test for no annotation emitted but I believe it will be pointless.

Please let me know if I understand you incorrectly.

pkg/streamingpromql/operators/functions/range_vectors.go
charleskorn: Makes sense, thanks for the explanation.

charleskorn:
> So I don't think there will be any annotation we can test there. We can test for no annotation emitted but I believe it will be pointless.

I think it'd be worthwhile adding a test for this: this codifies that we expect that no annotation is returned when only native histograms are selected.

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -874,8 +874,8 @@ func doubleExponentialSmoothing(step *types.RangeVectorStepData, _ float64, args
 			x = smoothingFactor * sample.F
 
 			// Scale the last smoothed value for all samples except the first one in head.
-			if head && i != 0 || !head {
-				estimatedTrend = calcTrendValue(trendFactor, smooth0, smooth1, estimatedTrend)
+			if head && i > 0 || !head {
```
charleskorn, line 877: [nit] Might be able to simplify this to make this unambiguous:

```suggestion
			if !head || i > 0 {
```

pkg/streamingpromql/testdata/ours/functions.test
jhesketh: It's to check we emit the same annotations and results as Prometheus when those arguments are things like `NaN`, `stale,` `+/-Inf`, etc. Also to check as they change over time.

pkg/streamingpromql/engine_test.go
jhesketh: It's to check we emit the same annotations and results as Prometheus when those arguments are things like `NaN`, `stale,` `+/-Inf`, etc. Also to check as they change over time.

```
		expressions = append(expressions, fmt.Sprintf(`double_exponential_smoothing(series{label=~"(%s)"}[1m], scalar(series{label="f"}),  scalar(series{label="i"}))`, labelRegex))
```

jhesketh:
> @jhesketh replying [#10873 (comment)](https://github.com/grafana/mimir/pull/10873#discussion_r1994648380) (not sure why I can't reply inline there.
> 
> > Can you also please add a test for when there are only native histograms?
> > There won't be any annotation emitted if samples are only native histogram.
> 
> The operator will just return.
> 
> ```
> 	if (len(fHead) + len(fTail)) < 2 {
> 		return 0, false, nil, nil
> 	}
> ```
> 
> So I don't think there will be any annotation we can test there. We can test for no annotation emitted but I believe it will be pointless.
> 
> Please let me know if I understand you incorrectly.

Yes, that was my intention. That we check we have the same annotations, or lack-thereof, as Prometheus. This can be one as a .test file.

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: Ignore my suggestion, I was trying to suggest a simplification to avoid the head check in accumulate, but did not take into account that `y` is still recalculated.

However, my concern here is the case where `len(fHead)` is equal to `1`. In that case, I think `if head && i > 0 || !head` check is incorrect as it'll be the second point, and not the third since `head = false`. Does that make sense?

I think something is missing around the edge case of the ring buffer here. A test would be good if possible.

lamida:
@jhesketh added test 8826a7fe841217ec82793aaed10b8d523c7b5448

pkg/streamingpromql/testdata/ours/functions.test
lamida: Added this 62995ad62bd5f01b46b3e31c3776b0a1d8caba40

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: @jhesketh thanks for this point I think the concern is valid. I will double check this again and the necessary test.

In the meantime do you have any suggestion how to setup engine_test in such a way ring buffer will have exactly 1 point in the head?

pkg/streamingpromql/engine_test.go
lamida: As I tried to add this, I found a bit strange comparison result. That can be reproduced with this test.

```
func TestDoubleExponentialRange(t *testing.T) {
	engine := promqltest.NewTestEngineWithOpts(t, promql.EngineOpts{
		EnableAtModifier:         true,
		MaxSamples:               10000,
		Timeout:                  10 * time.Hour,
		EnableDelayedNameRemoval: false,
	})
	mimirEngine, _ := NewEngine(NewTestEngineOpts(), NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), log.NewNopLogger())
	fmt.Println(engine)
	fmt.Println(mimirEngine)

	input := `
# Don't return anything when there's something there.
load 1m
		series{label="e", group="b"} {{schema:4 sum:12 count:8 buckets:[2 3 3]}} {{schema:4 sum:14 count:9 buckets:[3 3 3]}} _ stale Inf NaN {{schema:4 sum:18 count:11 buckets:[4 4 3]}}
		series{label="f", group="b"} 1 NaN _ Inf 5 NaN _
		series{label="i", group="c"} 1 {{schema:5 sum:15 count:10 buckets:[3 2 5]}} _ NaN {{schema:2 sum:30 count:25 buckets:[6 9 10 9 1]}} stale Inf

eval range from 0 to 7m step 1m scalar(series{label="f"})
	{} 1 NaN NaN Inf 5 NaN NaN NaN

eval instant at 4m series{label="f"}
	series{group="b", label="f"} 5

#eval instant at 4m double_exponential_smoothing(series{label=~"(e)"}[1m], scalar(series{label="f"}),  scalar(series{label="i"}))
eval range from 0 to 7m step 1m double_exponential_smoothing(series{label=~"(e)"}[1m], scalar(series{label="f"}),  scalar(series{label="i"}))

clear
`
	promqltest.RunTest(t, input, engine)
	promqltest.RunTest(t, input, mimirEngine)

}
```

Prometheus and Mimir gave two different errors.

Prometheus: `error evaluating query "double_exponential_smoothing(series{label=~\"(e)\"}[1m], scalar(series{label=\"f\"}),  scalar(series{label=\"i\"}))" (line 14): invalid smoothing factor. Expected: 0 < sf < 1, got: 5.000000
`
MQE: `error evaluating query "double_exponential_smoothing(series{label=~\"(e)\"}[1m], scalar(series{label=\"f\"}),  scalar(series{label=\"i\"}))" (line 14): invalid smoothing factor. Expected: 0 < sf < 1, got: 1.000000
`
I will try to continue checking this.

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: My approach would be to just test `doubleExponentialSmoothing`. Set up all the parameters for the function, you should be able to define a `types.RangeVectorStepData` with `FPointRingBuffer` where you set the slice very specifically

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1183,3 +1183,60 @@ eval range from 0s to 66m step 6m absent_over_time(histo{case=~"(0|a)"}[10m])
...
+
+# Test with variation of sf and tf
+load 2000s
+  sfUp 0.1+0.1x9
+  tfUp 0.1+0.1x9
+  sfDown 0.9-0.1x9
+  tfDown 0.9-0.1x9
+
+eval range from 0 to 20000s step 2000s double_exponential_smoothing(http_requests[5000s], scalar(sfUp), scalar(tfUp))
```
jhesketh, line 1220: (nit) The parameters change at the same rate / have the same value. It would be better to have them use different values to make sure we aren't somehow re-using a parameter.

pkg/streamingpromql/engine_test.go
jhesketh: The gauntlet strikes again!

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Thanks for this suggestion, I think uncover the bug in the original implementation by adding the suggested test!

Added the test and also the fix here: 6c09841c5cdcd101aa4a053decee93ff13c68576

pkg/streamingpromql/engine_test.go
```diff
@@ -2381,6 +2384,17 @@ func TestAnnotations(t *testing.T) {
 				`PromQL warning: quantile value should be between 0 and 1, got 1.5 (1:10)`,
 			},
 		},
+		"double_exponential_smoothing() with float and native histogram at same step": {
+			data:                    `some_metric 10 {{schema:0 sum:1 count:1 buckets:[1]}}`,
+			expr:                    "double_exponential_smoothing(some_metric[1m1s], 0.5, 0.5)",
+			expectedInfoAnnotations: []string{`PromQL info: ignored histograms in a range containing both floats and histograms for metric name "some_metric" (1:30)`},
+		},
+		"double_exponential_smoothing() with only native histogram at same step will result with no annotations": {
```
charleskorn, line 2392: [nit] We haven't included the expected behaviour in the test names elsewhere in this method:

```suggestion
		"double_exponential_smoothing() with only native histogram at same step": {
```

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,82 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+	fHead, fTail := step.Floats.UnsafePoints()
+
+	if step.Floats.Any() && step.Histograms.Any() {
+		emitAnnotation(annotations.NewHistogramIgnoredInMixedRangeInfo)
+	}
+
+	return runDoubleExponentialSmoothing(fHead, fTail, smoothingFactor, trendFactor)
+}
+
+func runDoubleExponentialSmoothing(fHead []promql.FPoint, fTail []promql.FPoint, smoothingFactor float64, trendFactor float64) (float64, bool, *histogram.FloatHistogram, error) {
```
charleskorn, line 854: [nit] Perhaps `calculateDoubleExponentialSmoothing` would be a clearer name?

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("runDoubleExponentialSmoothing() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if gotFloat != tt.wantFloat {
+				t.Errorf("runDoubleExponentialSmoothing() gotFloat = %v, wantFloat %v", gotFloat, tt.wantFloat)
+			}
+			if gotHasFloat != tt.wantHasFloat {
+				t.Errorf("runDoubleExponentialSmoothing() gotHasFloat = %v, wantHasFloat %v", gotHasFloat, tt.wantHasFloat)
+			}
```
charleskorn, line 105: It'd be clearer to use `require.EqualError` and `require.Equal` assertions here - any reason not to use them?

pkg/streamingpromql/testutils/utils.go
```diff
@@ -20,6 +20,10 @@ import (
 // It's possible that floating point values are slightly different due to imprecision, but require.Equal doesn't allow us to set an allowable difference.
 func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Result, skipAnnotationComparison bool) {
 	require.Equal(t, expected.Err, actual.Err)
+
+	if expected.Value == nil || actual.Value == nil {
+		return
+	}
```
charleskorn, line 26: Why is this change necessary?

pkg/streamingpromql/operators/functions/range_vectors.go
```diff
@@ -815,3 +816,82 @@ func quantileOverTime(step *types.RangeVectorStepData, _ float64, args []types.S
...
+		for _, sample := range samples {
+			// Scale the raw value against the smoothing factor.
+			x = smoothingFactor * sample.F
+
+			// Scale the last smoothed value for all samples except the first two points (point at actualPointIndex 0 and 1),
+			// regardless whether they are in head or tail.
+			if actualPointIndex > 1 {
+				estimatedTrend = trendFactor*(smooth1-smooth0) + (1-trendFactor)*estimatedTrend
+			} else {
+				actualPointIndex += 1
```
charleskorn, line 886: [nit]

```suggestion
				actualPointIndex++
```

pkg/streamingpromql/testutils/utils.go
lamida: In my attempt to address https://github.com/grafana/mimir/pull/10873#discussion_r1994656984, we will have a condition where test might not giving `Value` when validation is failing. In that situation we only want to compare the error.

pkg/streamingpromql/testdata/ours/functions.test
lamida: Changed the parameters rate 0608a2ef6006fca825ed2b73167bd31ed969b6de. The result is still the same though but we are sure there shouldn't be possibility of parameter reuse now.

pkg/streamingpromql/operators/functions/range_vectors.go
jhesketh: Why not use a bool for actualPointIndex? Maybe for clarity it could be `firstPointAfterInit bool`

pkg/streamingpromql/engine_test.go
lamida: Putting not on this failure. The test failure was due to the inconsistent ordering of prometheus processing matchers that result in multiples series, e.g `series{label=~"(c|e)"}`. Prometheus might process series c first or e first which will trigger different validation errors for second and third parameter of `double_exponential_smoothing`.

See the following example in more detail.

We want to to test
```
eval range from 0 to 4m step 1m double_exponential_smoothing(series{label=~"(c|e)"}[1m], scalar(series{label="f"}),  scalar(series{label="i"}))
```

With the following data
```
series{label="c", group="a"} 1 {{schema:3 sum:5 count:3 buckets:[1 1 1]}} -Inf {{schema:3 sum:10 count:6 buckets:[2 2 2]}} NaN {{schema:3 sum:12 count:7 buckets:[2 2 3]}} 5
series{label="e", group="b"} {{schema:4 sum:12 count:8 buckets:[2 3 3]}} {{schema:4 sum:14 count:9 buckets:[3 3 3]}} _ stale Inf NaN {{schema:4 sum:18 count:11 buckets:[4 4 3]}}
series{label="f", group="b"} 1 NaN _ Inf 5 NaN _
```

Ignore the second parameter `scalar(series{label=“i”})` and just focus on `scalar(series{label=“f”})`.

If prometheus process c first, it will fail with 1.0000, because it fail on the first step (sample 1) of the query range.
but
If prometheus process e first, it will fail with 5.0000, because it fail on the fifth step (sample Inf) of the query range. (edited) 

Prometheus will skip processing series and validate the double_exponential_smoothing for non float sample.

Series c fail on first step because it has sample float on step 1. Series e didn’tfail on step 1 because it has histogram sample which it ignores. It only fail on step 5th on Inf sample.

lamida: After discussing with @jhesketh we found the issue is because of prometheus behaviour I noted in the previous comment. Therefore we decided to drop the test on dynamic value for tf and sf parameter from series.

pkg/streamingpromql/engine_test.go
lamida: Addressed in 4e76aca6303b7d2fb9727105298032bbea8d57c6

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: I think with the latest code state, the concern should already be addressed. I will resolve this.

pkg/streamingpromql/testutils/utils.go
lamida: Removed this in d0de323fe7486b9aa25cb51ebe12d767cbe15225 because @jhesketh [suggested test](https://github.com/grafana/mimir/pull/10873#discussion_r1994656984) won't work with the inconsistent prometheus behaviour.

pkg/streamingpromql/operators/functions/range_vectors.go
lamida: Initially I was thinking to track the first two steps, but apparently we only need to skip the first iteration, hence boolean works. Updated in 19ad21cc120d87578d98598efdfc5bda7e2aa5a3

pkg/streamingpromql/testutils/utils.go
lamida: Sorry need to reenable again because we eventually add the error test although only with one label. See aaebaa28df8302ecae7669f0ee65be795718831e

pkg/streamingpromql/engine_test.go
```diff
@@ -3094,13 +3094,20 @@ func TestCompareVariousMixedMetricsVectorSelectors(t *testing.T) {
 
 	labelsToUse, pointsPerSeries, seriesData := getMixedMetricsForTests(true)
 
+	expressions := []string{}
+
 	// Test each label individually to catch edge cases in with single series
 	labelCombinations := testutils.Combinations(labelsToUse, 1)
+
+	// We tried to have this test with 2 labels, but it was failing due to the inconsistent ordering of prometheus processing matchers that result in multiples series, e.g series{label=~"(c|e)"}.
+	// Prometheus might process series c first or e first which will trigger different validation errors for second and third parameter of double_exponential_smoothing.
```
jhesketh, line 3103: I think it's helpful to explain why it gives a different validation error as you explained to me earlier.
It's not a different argument to the function, it's a different value in the range vector.
That is because the range vector of the series being computed against, values are skipped for the native histograms until it gets to a value where it has a float. That aligns with a different scalar value for the argument and thus gives a different error.

pkg/streamingpromql/testutils/utils.go
jhesketh: I think it would be more useful to have a flag to `RequireEqualResults` for `skipValueComparisonOnError`, so we only skip it if we've also gotten an error and then pass that through from the relevant test (if that makes sense?)

pkg/streamingpromql/engine_test.go
lamida: Effectively it is the different argument that is passed. But there is a part of how native histogram being skipped too. I added your point here da3a0a3fa567884904f918ab20e58b56ca3b8995.

lamida:
@jhesketh 
> I think it would be more useful to have a flag to RequireEqualResults for skipValueComparisonOnError, so we only skip it if we've also gotten an error and then pass that through from the relevant test (if that makes sense?)

A bit large changes, but this should address your feedback: 611e1b1af7ecdac4451b6fb164c56d2dfbd29aa4

pkg/streamingpromql/testutils/utils.go
lamida: Reverted this again with 611e1b1af7ecdac4451b6fb164c56d2dfbd29aa4

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,105 @@
...
+		fTail           []promql.FPoint
+		smoothingFactor float64
+		trendFactor     float64
+	}
+	tests := []struct {
+		name         string
+		args         args
+		wantFloat    float64
+		wantHasFloat bool
+		wantErr      bool
```
charleskorn, line 27: `wantErr` is never used - should we just remove it?

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"testing"
+
+	"github.com/prometheus/prometheus/promql"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_runDoubleExponentialSmoothing(t *testing.T) {
+	type args struct {
```
charleskorn, line 16: [nit] Why not just inline these fields into the struct below?

pkg/streamingpromql/engine_test.go
charleskorn: It's not clear to me why, if we might get different errors, we skip comparing the values - shouldn't we skip comparing the errors?

lamida: When we get validation error, `Value` will be nil and we got panic. We still want to compare the error to ensure the validation is correct. More detail discussion is in this comment: https://github.com/grafana/mimir/pull/10873/files/b1eb614926072c1a01437ddd2e6f53b176459960..aaebaa28df8302ecae7669f0ee65be795718831e#r2015549294

pkg/streamingpromql/operators/functions/range_vectors_test.go
lamida: Make it little clearer which one is arguments and which one are `want`s. I think I will just keep it this way.

pkg/streamingpromql/operators/functions/range_vectors_test.go
lamida: Removed 18188b0bd8009f1cb4185d62e35ccd967c6a0b13

pkg/streamingpromql/operators/functions/range_vectors_test.go
```diff
@@ -0,0 +1,100 @@
...
+				smoothingFactor: 0.1,
+				trendFactor:     0.1,
+			},
+			wantFloat:    1,
+			wantHasFloat: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotFloat, gotHasFloat, _, _ := calculateDoubleExponentialSmoothing(tt.args.fHead, tt.args.fTail, tt.args.smoothingFactor, tt.args.trendFactor)
```
charleskorn, line 95: We should check that we got no error here.

pkg/streamingpromql/engine_test.go
charleskorn: If `result.Err` is not `nil`, then shouldn't `result.Value` always be `nil`? 

If so, then I think we can drop `skipValueComparison` and modify `RequireEqualResults` to ensure that both results have the same error (or both don't have an error), and if both have the same error, then both `Value`s should be `nil`.

lamida: I am still thinking the original check to just return and skip Value comparison if they are nil is the simplest.

But let me attempt to still address this suggestion.

>  to ensure that both results have the same error (or both don't have an error)
We can achieve  just using the existing first check
```
	require.Equal(t, expected.Err, actual.Err)
```	
> and if both have the same error, then both Values should be nil.
```
	if expected.Err != nil && actual.Err != nil && expected.Err.Error() == actual.Err.Error() {
		require.Nil(t, expected.Value)
		require.Nil(t, actual.Value)
		return
	}
```

The proposed change is already applied in the last commit.

lamida: Note: I can't use `errors.Is(expected.Err, actual.Err)`. It won't check the error as the same although the error string is same.

pkg/streamingpromql/testutils/utils.go
```diff
@@ -20,6 +20,13 @@ import (
 // It's possible that floating point values are slightly different due to imprecision, but require.Equal doesn't allow us to set an allowable difference.
 func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Result, skipAnnotationComparison bool) {
 	require.Equal(t, expected.Err, actual.Err)
+
+	if expected.Err != nil && actual.Err != nil && expected.Err.Error() == actual.Err.Error() {
```
charleskorn, line 24: This should be sufficient:

```suggestion
	if expected.Err != nil {
```

We've already checked that `expected.Err` and `actual.Err` are equal above, so we only need to check if there's an error.

charleskorn (approved):
LGTM modulo suggestion below

PR #10877, opened 2025-03-13
mimir-mixin-compiled-gem: change product to GEM

dimitarvdimitrov:
#### What this PR does

This changes the product from `Mimir` to `GEM`, but keeps the alert prefixes as `Mimir` so that the runbook links still work.

To do that I needed to add a new config option: `_config.alert_product`

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


56quarters (approved):
LGTM

PR #10892, opened 2025-03-14
Ingester: Export native histogram and buckets gauge metrics for cost attribution

ying-jeanne:
Native histogram is now counted as one single active series, which is wrong. In order to do the cost attribution with precision, we need to export the native histogram series number and its buckets number.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/costattribution/active_tracker.go
```diff
@@ -110,24 +147,47 @@ func (at *ActiveSeriesTracker) Increment(lbls labels.Labels, now time.Time) {
...
+	nhBucketNum := atomic.NewInt64(0)
+	if nativeHistogramBucketNum >= 0 {
+		nhCount.Inc()
+		nhBucketNum.Add(int64(nativeHistogramBucketNum))
+	}
+	at.observed[string(buf.Bytes())] = asMetrics{
+		asCount:     atomic.NewInt64(1),
+		nhCount:     nhCount,
+		nhBucketNum: nhBucketNum,
+	}
```
krajorama, line 187: nit, this looks a bit faster:
```suggestion

	nhCount := int64(0)
	nhBucketNum := int64(0)
	if nativeHistogramBucketNum >= 0 {
		nhCount++
		nhBucketNum+=int64(nativeHistogramBucketNum)
	}
	at.observed[string(buf.Bytes())] = asMetrics{
		asCount:     atomic.NewInt64(1),
		nhCount:     atomic.NewInt64(nhCount),
		nhBucketNum: atomic.NewInt64(nhBucketNum),
	}
```

pkg/costattribution/active_tracker.go
```diff
@@ -169,19 +237,29 @@ func (at *ActiveSeriesTracker) Collect(out chan<- prometheus.Metric) {
...
 	}
 	// We don't know the performance of out receiver, so we don't want to hold the lock for too long
 	var prometheusMetrics []prometheus.Metric
 	for key, as := range at.observed {
 		keys := strings.Split(key, string(sep))
 		keys = append(keys, at.userID)
-		prometheusMetrics = append(prometheusMetrics, prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.Load()), keys...))
+		prometheusMetrics = append(prometheusMetrics,
+			prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.asCount.Load()), keys...),
+			prometheus.MustNewConstMetric(at.activeNativeHistogramSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.nhCount.Load()), keys...),
```
krajorama, line 260: Always adding the native histogram series and buckets triples the number of series even for tenant that don't use native histograms. I think we should only append them is series count > 0. Of course this would need a new test case where we go from no native series to a couple then back 0 and check that the two additional series come and go.

I assume it that same for the basic series attribution as well, if the attributed series go away, we remove the corresponding `cortex_ingester_attributed_active_series` series !?

krajorama (commented):
looking good, needs changlog entry for sure. Also possibly modify docs?

pkg/costattribution/active_tracker.go
ying-jeanne: At first, I wasn't sure how recording rules would behave when evaluating active_series{attribution='a'} - histogram{attribution='a'} + 0.5 * histogram_buckets{attribution='a'} if histogram and histogram_buckets were missing. If the result treated missing metrics as 0, then it would work fine, but otherwise, it might lead to an empty result. Taking a second thought, we could work around this by using active_series{attribution='a'} - (histogram{attribution='a'} or 0) + 0.5 * (histogram_buckets{attribution='a'} or 0) in the worst case. let me remove it for now.

ying-jeanne: > if the attributed series go away, we remove the corresponding cortex_ingester_attributed_active_series series !?
correct, that attribution would be purged from map (cache), so yes metrics would be gone.

ying-jeanne: that should work, i removed those 2 native histogram metrics when it is zero accordingly.

tacole02 (approved):

pkg/costattribution/active_tracker.go
```diff
@@ -16,10 +16,18 @@ import (
 	"go.uber.org/atomic"
 )
 
+type asCounters struct {
+	asCount     *atomic.Int64
+	nhCount     *atomic.Int64
+	nhBucketNum *atomic.Int64
+}
```
colega, line 23: Is there any reason why you chose to have struct of pointers instead of a pointer to a struct with just atomic.Int64 values?

colega, line 23: Opinionated, but given it's `asCounters`, adding `Count` suffix to the field is redundant and doesn't help readability. 

OTOH, do we need `as` prefix in the struct? Isn't that implicit from this entire feature?

IMO this reads better:
```suggestion
type counters struct {
	activeSeries           *atomic.Int64
	nativeHistograms       *atomic.Int64
	nativeHistogramBuckets *atomic.Int64
}
```

pkg/costattribution/active_tracker.go
```diff
@@ -138,15 +198,19 @@ func (at *ActiveSeriesTracker) Decrement(lbls labels.Labels) {
...
+			as.nhCount.Dec()
+			as.nhBucketNum.Sub(int64(nativeHistogramBucketNum))
+		}
 		if nv > 0 {
 			at.observedMtx.RUnlock()
 			return
 		}
 		at.observedMtx.RUnlock()
 		at.observedMtx.Lock()
 		as, ok := at.observed[string(buf.Bytes())]
```
colega, line 212: I think this `as` variable everywhere is now `counters` (or maybe just `c`? but not `as`)

pkg/ingester/activeseries/active_series.go
```diff
@@ -472,7 +472,7 @@ func (s *seriesStripe) purge(keepUntil time.Time, idx tsdb.IndexReader) {
 				if err := idx.Series(ref, &buf, nil); err != nil {
 					s.activeSeriesAttributionFailureCounter.Add(1)
 				} else {
-					s.cat.Decrement(buf.Labels())
+					s.cat.Decrement(buf.Labels(), entry.numNativeHistogramBuckets)
```
colega, line 475: I see that in `findAndUpdateOrCreateEntryForSeries` we can modify `entry.numNativeHistogramBuckets`, which means that this number isn't necessarily the same we called on `s.cat.Increment(series, time.Unix(0, nowNanos), numNativeHistogramBuckets)` above, which means we could go negative values (or have leftovers).

Please add a test that tracks a series with given labels, then that series turns into a native histogram, and then it's removed, and check that cost attribution tracker is at zero.

colega (changes requested):
Requesting changes to handle native histogram bucket number changes.

pkg/ingester/activeseries/active_series.go
ying-jeanne: in commit fe7acada914ff9ac74f5992afeb86f857a54ba83, i added a test to decrement more on bucket than incremented, the bucket counter should stay at 0, also when number of native histogram series become 0, the total bucket should be 0

56quarters:

pkg/ingester/activeseries/active_series.go
ying-jeanne: addressed accordingly in commit bb0a5dcb405890615156fecaaa44b4c003e4d7a1

pkg/costattribution/active_tracker.go
krajorama: that commit did not change this line, still called `as`

pkg/costattribution/active_tracker.go
```diff
@@ -136,17 +187,21 @@ func (at *ActiveSeriesTracker) Decrement(lbls labels.Labels) {
 	defer bufferPool.Put(buf)
 	at.fillKeyFromLabels(lbls, buf)
 	at.observedMtx.RLock()
-	as, ok := at.observed[string(buf.Bytes())]
+	c, ok := at.observed[string(buf.Bytes())]
 	if ok {
-		nv := as.Dec()
+		nv := c.activeSeries.Dec()
+		if nativeHistogramBucketNum >= 0 && c.nativeHistograms.Load() > 0 {
```
krajorama, line 193: we don't check for consistency of c.activeSeries so I think we should do this:
```suggestion
		if nativeHistogramBucketNum >= 0  {
```

pkg/costattribution/active_tracker.go
```diff
@@ -159,7 +214,11 @@ func (at *ActiveSeriesTracker) Decrement(lbls labels.Labels) {
 	defer at.observedMtx.RUnlock()
 
 	if !at.overflowSince.IsZero() {
-		at.overflowCounter.Dec()
+		at.overflowCounter.activeSeries.Dec()
+		if nativeHistogramBucketNum >= 0 && at.overflowCounter.nativeHistograms.Load() > 0 {
```
krajorama, line 218: same as above
```suggestion
		if nativeHistogramBucketNum >= 0 {
```

pkg/costattribution/active_tracker.go
```diff
@@ -169,19 +228,42 @@ func (at *ActiveSeriesTracker) Collect(out chan<- prometheus.Metric) {
...
 	// We don't know the performance of out receiver, so we don't want to hold the lock for too long
 	var prometheusMetrics []prometheus.Metric
 	for key, as := range at.observed {
 		keys := strings.Split(key, string(sep))
 		keys = append(keys, at.userID)
-		prometheusMetrics = append(prometheusMetrics, prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.Load()), keys...))
+		prometheusMetrics = append(prometheusMetrics,
+			prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.activeSeries.Load()), keys...),
+		)
+		if as.nativeHistograms.Load() > 0 {
```
krajorama, line 257: I think the way you do it for overflow is quicker (no need to load twice) and cleaner. So load into local variable nhcounter .

pkg/costattribution/active_tracker.go
```diff
@@ -169,19 +228,42 @@ func (at *ActiveSeriesTracker) Collect(out chan<- prometheus.Metric) {
...
+			prometheusMetrics = append(prometheusMetrics,
+				prometheus.MustNewConstMetric(at.activeNativeHistogramSeriesPerUserAttribution, prometheus.GaugeValue, float64(as.nativeHistograms.Load()), keys...),
+			)
+			if as.nativeHistogramBuckets.Load() > 0 {
```
krajorama, line 261: same as above, load into local variable bcounter

pkg/costattribution/manager_test.go
```diff
@@ -65,11 +67,72 @@ func TestManager_CreateDeleteTracker(t *testing.T) {
...
+			strings.NewReader(expectedMetrics),
+			"cortex_discarded_attributed_samples_total",
+			"cortex_distributor_received_attributed_samples_total",
+			"cortex_ingester_attributed_active_series",
+			"cortex_ingester_attributed_active_native_histogram_series",
+			"cortex_ingester_attributed_active_native_histogram_buckets",
+		))
+
+		// decrement native histogram series, the metrics should be removed, so now the native histogram counter is 0, and the active series counter is 1
+		manager.ActiveSeriesTracker("user1").Decrement(labels.FromStrings("team", "bar"), 60)
```
krajorama, line 90: is this a valid use case? I think we made sure to reduce with the correct amount, so this would be either -1, 50 or 2, but not 60?

pkg/costattribution/active_tracker.go
```diff
@@ -169,19 +228,42 @@ func (at *ActiveSeriesTracker) Collect(out chan<- prometheus.Metric) {
...
+			nativeHistogram += as.nativeHistograms.Load()
+			nhBucketNum += as.nativeHistogramBuckets.Load()
 		}
 		at.observedMtx.RUnlock()
-		out <- prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(activeSeries+at.overflowCounter.Load()), at.overflowLabels[:len(at.overflowLabels)-1]...)
+		out <- prometheus.MustNewConstMetric(at.activeSeriesPerUserAttribution, prometheus.GaugeValue, float64(activeSeries+at.overflowCounter.activeSeries.Load()), at.overflowLabels[:len(at.overflowLabels)-1]...)
+
+		if nhcounter := float64(nativeHistogram + at.overflowCounter.nativeHistograms.Load()); nhcounter > 0 {
+			out <- prometheus.MustNewConstMetric(at.activeNativeHistogramSeriesPerUserAttribution, prometheus.GaugeValue, nhcounter, at.overflowLabels[:len(at.overflowLabels)-1]...)
+			if bcounter := float64(nhBucketNum + at.overflowCounter.nativeHistogramBuckets.Load()); bcounter > 0 {
```
krajorama, line 243: Interestingly in the normal tracker+custom tracker, the two gauges are independently checked https://github.com/grafana/mimir/blob/bb31bb80cd62a5ef5525eebdd500c4cafa7b65f6/pkg/ingester/ingester.go#L828
Just for consistency I think we should do the same. It's easier to reason about if something goes wrong.

pkg/ingester/activeseries/active_series.go
```diff
@@ -369,6 +371,8 @@ func (s *seriesStripe) findAndUpdateOrCreateEntryForSeries(ref storage.SeriesRef
 					s.activeMatchingNativeHistograms[match]++
 					s.activeMatchingNativeHistogramBuckets[match] += uint32(numNativeHistogramBuckets)
 				}
+				s.cat.Decrement(series, -1)
```
colega, line 374:
```suggestion
				s.cat.Decrement(series, entry.numNativeHistogramBuckets)
```

colega: This would make this block equal in all three cases, so why we don't just move it down where we do `entry.numNativeHistogramBuckets = numNativeHistogramBuckets`?

pkg/costattribution/active_tracker.go
ying-jeanne: addressed in 54875a7a10319987325308faf27fbe58129527f2

pkg/costattribution/active_tracker.go
ying-jeanne: addressed in 54875a7a10319987325308faf27fbe58129527f2

pkg/costattribution/active_tracker.go
ying-jeanne: addressed in 47deaae69ffe17c051c46b4ecd9b9ddf67ab4c6c

pkg/costattribution/manager_test.go
ying-jeanne: changed to 50

colega (approved):

PR #10897, opened 2025-03-16
Expose query stats in `Server-Timing` header when `X-Mimir-Response-Query-Stats` header is present in the request.

### LINKED ISSUE #10192: Idea: expose query stats in Grafana

The query-frontend already keeps track of aggregated query stats. Grafana has a "Stats" tab in the "Query Inspector" view. The graphite datasource already uses that tab to feed some stats back to the user. We should expose (most of) the existing query stats in that pane.

These stats helps users fine-tune queries or spot underscaled Mimir clusters by looking at shardability, volume of data touched, time spent in the queue, and time spent encoding the result.

<details><summary>Existing query stats</summary>
<p>

```
caller=handler.go:383
component=query-frontend
encode_time_seconds=4.0957e-05
estimated_series_count=31950
fetched_chunk_bytes=23791904
fetched_chunks_count=329964
fetched_index_bytes=28647863
fetched_series_count=29906
header_cache_control=
header_x_forwarded_for="10.201.114.35, 10.63.49.131"
length=24h5m0s
level=info
method=POST
msg="query stats"
param_query="REDACTED"
param_time=2024-12-03T08:45:11Z
path=/prometheus/api/v1/query
query_wall_time_seconds=6.259888235
queue_time_seconds=1.9816e-05
response_size_bytes=118
response_time=6.267380946s
results_cache_hit_bytes=0
results_cache_miss_bytes=0
route_name=prometheus_api_v1_query
sharded_queries=0
split_queries=0
status=success
status_code=200
time_since_max_time=1m0.763386437s
time_since_min_time=24h6m0.763386437s
ts=2024-12-03T08:45:18.030817629Z
user=10428
user_agent=Opencost/1.100.0 
```

</p>
</details>

### Format 

The query-frontend results each query stat in the [`Server-Timing` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing). The same header is already used for some statistics. We can publish all relevant query stats to the same header. For example

```
Server-Timing: encode;dur=0.041, series_count;c=31950, chunk_bytes;c=23791904, chunks_count;c=329964, index_bytes;c=28647863, series_fetched;c=29906, wall_time;dur=6259.888, queue;dur=0.020, response_size;c=118, response_time;dur=6267.381, cache_hit;c=0, cache_miss;c=0, sharded;c=0, split;c=0
```

### Considerations

The header above is 293 characters. For a lot of queries the header will comprise most of the HTTP response. We don't want to unnecessarily amplify data transfer at the query-frontend. For that I propose to include an optional request header. The query-frontend only includes these stats when requested to.

```
X-Mimir-Response-Query-Stats: true
```

### Stability

This is intended to be ultimately interpreted by humans. As such we can change the set of stats we expose or rename them without notice.

Use the "Mimir / Slow queries" dashboard to find query stats.

* Grafana also needs to implement support for this. It probably makes sense to do this after we have a released version of it in Mimir.
* This is a feature request. See linked internal issue.

Small (<= 1 month dev)

Update the HTTP reference to mention the additional header and the potential stats and the (lack of) stability guarantee.

### Proposer?

_No response_

santileira:
#### What this PR does

Expose query stats in `Server-Timing` header when `X-Mimir-Response-Query-Stats` header is present in the request.

Fixes #10192 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * [CHANGE] Store-gateway: Include posting sampling rate in sparse index headers. When the sampling rate isn't set in a sparse index header, store gateway rebuilds the sparse header with the configured `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` value. If the sparse header's sampling rate is set but doesn't match the configured rate, store gateway either rebuilds the sparse header or downsamples to the configured sampling rate. #10684 #10878
 * [CHANGE] Distributor: Return specific error message when burst size limit is exceeded. #10835
 * [CHANGE] Ingester: enable native histograms ingestion by default, meaning`ingester.native-histograms-ingestion-enabled` defaults to true. #10867
+* [FEATURE] Query Frontend: Expose query stats in `Server-Timing` header when `X-Mimir-Response-Query-Stats` header is present in the request. #10192
 * [FEATURE] Ingester/Distributor: Add support for exporting cost attribution metrics (`cortex_ingester_attributed_active_series`, `cortex_distributor_received_attributed_samples_total`, and `cortex_discarded_attributed_samples_total`) with labels specified by customers to a custom Prometheus registry. This feature enables more flexible billing data tracking. #10269 #10702
 * [FEATURE] Ruler: Added `/ruler/tenants` endpoints to list the discovered tenants with rule groups. #10738
 * [FEATURE] Distributor: Add experimental Influx handler. #10153
--- a/pkg/frontend/transport/handler.go
+++ b/pkg/frontend/transport/handler.go
@@ -39,10 +39,11 @@ import (
 
 const (
 	// StatusClientClosedRequest is the status code for when a client request cancellation of an http request
-	StatusClientClosedRequest = 499
-	ServiceTimingHeaderName   = "Server-Timing"
-	cacheControlHeader        = "Cache-Control"
-	cacheControlLogField      = "header_cache_control"
+	StatusClientClosedRequest    = 499
+	ServiceTimingHeaderName      = "Server-Timing"
+	cacheControlHeader           = "Cache-Control"
+	cacheControlLogField         = "header_cache_control"
+	responseQueryStatsHeaderName = "X-Mimir-Response-Query-Stats"
 )
 
 var (
@@ -177,7 +178,8 @@ func (f *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Initialise the queryDetails in the context and make sure it's propagated
 	// down the request chain.
-	if f.cfg.QueryStatsEnabled {
+	queryStatsHeaderNameOk, queryStatsHeaderNameOkErr := strconv.ParseBool(r.Header.Get(responseQueryStatsHeaderName))
+	if f.cfg.QueryStatsEnabled || (queryStatsHeaderNameOk && queryStatsHeaderNameOkErr == nil) {
 		var ctx context.Context
 		queryDetails, ctx = querymiddleware.ContextWithEmptyDetails(r.Context())
 		r = r.WithContext(ctx)
@@ -244,10 +246,6 @@ func (f *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		hs[h] = vs
 	}
 
-	if f.cfg.QueryStatsEnabled {
-		writeServiceTimingHeader(queryResponseTime, hs, queryDetails.QuerierStats)
-	}
-
 	w.WriteHeader(resp.StatusCode)
 	// we don't check for copy error as there is no much we can do at this point
 	queryResponseSize, _ := io.Copy(w, resp.Body)
@@ -257,6 +255,17 @@ func (f *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	if f.cfg.QueryStatsEnabled {
 		f.reportQueryStats(r, params, startTime, queryResponseTime, queryResponseSize, queryDetails, resp.StatusCode, nil)
+	}
+
+	var parts []string
+	if f.cfg.QueryStatsEnabled {
+		parts = getQueryStats(queryResponseTime, queryDetails.QuerierStats)
+	}
+	if queryStatsHeaderNameOk && queryStatsHeaderNameOkErr == nil {
+		parts = append(parts, getResponseQueryStats(queryResponseTime, queryResponseSize, queryDetails)...)
+	}
+	if len(parts) > 0 {
+		hs.Set(ServiceTimingHeaderName, strings.Join(parts, ", "))
 	}
 }
 
@@ -481,26 +490,55 @@ func writeError(w http.ResponseWriter, err error) int {
 	return statusCode
 }
 
-func writeServiceTimingHeader(queryResponseTime time.Duration, headers http.Header, stats *querier_stats.Stats) {
-	if stats != nil {
-		parts := make([]string, 0)
-		parts = append(parts, statsValue("querier_wall_time", stats.LoadWallTime()))
-		parts = append(parts, statsValue("response_time", queryResponseTime))
-		parts = append(parts, statsValue("bytes_processed", stats.LoadFetchedChunkBytes()+stats.LoadFetchedIndexBytes()))
-		parts = append(parts, statsValue("samples_processed", stats.GetSamplesProcessed()))
-		headers.Set(ServiceTimingHeaderName, strings.Join(parts, ", "))
+func getQueryStats(queryResponseTime time.Duration, stats *querier_stats.Stats) []string {
+	if stats == nil {
+		return nil
 	}
+	parts := make([]string, 0)
+	parts = append(parts, statsValue("querier_wall_time", stats.LoadWallTime()))
+	parts = append(parts, statsValue("response_time", queryResponseTime))
+	parts = append(parts, statsValue("bytes_processed", stats.LoadFetchedChunkBytes()+stats.LoadFetchedIndexBytes()))
+	parts = append(parts, statsValue("samples_processed", stats.GetSamplesProcessed()))
+	return parts
+}
+
+// getResponseQueryStats returns the response query stats in the format of Server-Timing header.
+func getResponseQueryStats(queryResponseTime time.Duration, queryResponseSizeBytes int64, details *querymiddleware.QueryDetails) []string {
+	if details == nil {
+		return nil
+	}
+	stats := details.QuerierStats
+	parts := make([]string, 0)
+	parts = append(parts, statsValue("encode", stats.LoadEncodeTime()))
+	parts = append(parts, statsValueWithKey("series_count", "c", stats.LoadEstimatedSeriesCount()))
+	parts = append(parts, statsValueWithKey("chunk_bytes", "c", stats.LoadFetchedChunkBytes()))
+	parts = append(parts, statsValueWithKey("chunks_count", "c", stats.LoadFetchedChunks()))
+	parts = append(parts, statsValueWithKey("index_bytes", "c", stats.LoadFetchedIndexBytes()))
+	parts = append(parts, statsValueWithKey("series_fetched", "c", stats.LoadFetchedSeries()))
+	parts = append(parts, statsValueWithKey("wall_time", "c", stats.LoadWallTime()))
+	parts = append(parts, statsValue("queue", stats.LoadQueueTime()))
+	parts = append(parts, statsValueWithKey("response_size", "c", queryResponseSizeBytes))
+	parts = append(parts, statsValue("response_time", queryResponseTime))
+	parts = append(parts, statsValueWithKey("cache_hit", "c", details.ResultsCacheHitBytes))
+	parts = append(parts, statsValueWithKey("cache_miss", "c", details.ResultsCacheMissBytes))
+	parts = append(parts, statsValueWithKey("sharded", "c", stats.LoadShardedQueries()))
+	parts = append(parts, statsValueWithKey("split", "c", stats.LoadSplitQueries()))
+	return parts
 }
 
 func statsValue(name string, val interface{}) string {
+	return statsValueWithKey(name, "val", val)
+}
+
+func statsValueWithKey(name string, key string, val interface{}) string {
 	switch v := val.(type) {
 	case time.Duration:
 		durationInMs := strconv.FormatFloat(float64(v)/float64(time.Millisecond), 'f', -1, 64)
 		return name + ";dur=" + durationInMs
 	case uint64:
-		return name + ";val=" + strconv.FormatUint(v, 10)
+		return name + ";" + key + "=" + strconv.FormatUint(v, 10)
 	default:
-		return name + ";val=" + fmt.Sprintf("%v", v)
+		return name + ";" + key + "=" + fmt.Sprintf("%v", v)
 	}
 }
 
--- a/pkg/frontend/transport/handler_test.go
+++ b/pkg/frontend/transport/handler_test.go
@@ -14,6 +14,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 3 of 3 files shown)


pkg/frontend/transport/handler_test.go
```diff
@@ -359,6 +436,7 @@ func TestHandler_ServeHTTP(t *testing.T) {
 			require.NoError(t, err)
 			require.Empty(t, activities)
 
+			tt.assertHeaders(t, resp.Header())
```
santileira: We want to check headers if the function is defined, it doesn't matter if QueryStatsEnabled is true or not.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [CHANGE] Store-gateway: Include posting sampling rate in sparse index headers. When the sampling rate isn't set in a sparse index header, store gateway rebuilds the sparse header with the configured `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` value. If the sparse header's sampling rate is set but doesn't match the configured rate, store gateway either rebuilds the sparse header or downsamples to the configured sampling rate. #10684 #10878
 * [CHANGE] Distributor: Return specific error message when burst size limit is exceeded. #10835
 * [CHANGE] Ingester: enable native histograms ingestion by default, meaning`ingester.native-histograms-ingestion-enabled` defaults to true. #10867
+* [FEATURE] Query Frontend: Expose query stats in `Server-Timing` header when `X-Mimir-Response-Query-Stats` header is present in the request. #10192
```
santileira, line 21: @dimitarvdimitrov what documentation do I have tu update? The issue says this: 
```
What are the documentation dependencies?
Update the HTTP reference to mention the additional header and the potential stats and the (lack of) stability guarantee.
```

dimitarvdimitrov: i had in mind this doc https://github.com/grafana/mimir/blob/248ca33ba181fb1c0326ddbc79962edde3098640/docs/sources/mimir/references/http-api/index.md


i'd expect to see a mention of the request and response headers. What they are, why are they might be useful, what are the tradeoffs. There is a mention of `Sharding-Control` in one place in that doc already. maybe you can use the same format ("Headers" and "Response headers" for example).

dimitarvdimitrov (commented):
thanks for working on this! great job on the tests. This looks very well. I think the only missing part is the docs you asked about.

I'm having second thoughts about how to name the response fields though.

pkg/frontend/transport/handler.go
```diff
@@ -177,7 +178,8 @@ func (f *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Initialise the queryDetails in the context and make sure it's propagated
 	// down the request chain.
-	if f.cfg.QueryStatsEnabled {
+	queryStatsHeaderNameOk, queryStatsHeaderNameOkErr := strconv.ParseBool(r.Header.Get(responseQueryStatsHeaderName))
```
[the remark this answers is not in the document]
santileira: You are right. I will change it.

pkg/frontend/transport/handler.go
```diff
@@ -481,26 +490,56 @@ func writeError(w http.ResponseWriter, err error) int {
...
 	case time.Duration:
 		durationInMs := strconv.FormatFloat(float64(v)/float64(time.Millisecond), 'f', -1, 64)
 		return name + ";dur=" + durationInMs
 	case uint64:
-		return name + ";val=" + strconv.FormatUint(v, 10)
+		return name + ";" + key + "=" + strconv.FormatUint(v, 10)
 	default:
-		return name + ";val=" + fmt.Sprintf("%v", v)
+		return name + ";" + key + "=" + fmt.Sprintf("%v", v)
 	}
```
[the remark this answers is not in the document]
santileira: Yes. I will change it.

pkg/frontend/transport/handler.go
```diff
@@ -481,26 +490,56 @@ func writeError(w http.ResponseWriter, err error) int {
...
+	parts = append(parts, statsValueWithKey("series_fetched", "c", stats.LoadFetchedSeries()))
+	parts = append(parts, statsValueWithKey("wall_time", "c", stats.LoadWallTime()))
+	parts = append(parts, statsValue("queue", stats.LoadQueueTime()))
+	parts = append(parts, statsValueWithKey("response_size", "c", queryResponseSizeBytes))
+	parts = append(parts, statsValue("response_time", queryResponseTime))
+	parts = append(parts, statsValueWithKey("cache_hit", "c", details.ResultsCacheHitBytes))
+	parts = append(parts, statsValueWithKey("cache_miss", "c", details.ResultsCacheMissBytes))
+	parts = append(parts, statsValueWithKey("sharded", "c", stats.LoadShardedQueries()))
+	parts = append(parts, statsValueWithKey("split", "c", stats.LoadSplitQueries()))
+	return parts
```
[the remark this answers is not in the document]
santileira: Yes. I will change it.

pkg/frontend/transport/handler.go
```diff
@@ -481,26 +490,56 @@ func writeError(w http.ResponseWriter, err error) int {
...
+}
+
+// getResponseQueryStats returns the response query stats in the format of Server-Timing header.
+func getResponseQueryStats(queryResponseTime time.Duration, queryResponseSizeBytes int64, details *querymiddleware.QueryDetails) []string {
+	if details == nil {
+		return nil
+	}
+	stats := details.QuerierStats
+	parts := make([]string, 0)
+	parts = append(parts, statsValue("encode", stats.LoadEncodeTime()))
```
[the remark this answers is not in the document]
santileira: I will wait the answer from the guys before doing any change.

CHANGELOG.md
santileira: Makes sense. I will update the doc.

tacole02, line 21:
```suggestion
* [FEATURE] Query Frontend: Expose query stats in the `Server-Timing` header when the `X-Mimir-Response-Query-Stats` header is present in the request. #10192
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
 #### Headers
 
 - `Sharding-Control` - _optional_ - Integer value specifying how many shards to use for request execution.
+- `X-Mimir-Response-Query-Stats` - _optional_ - Boolean value specifying if the response must have query stats on the `Server-Timing` header.
```
tacole02, line 595:
```suggestion
- `X-Mimir-Response-Query-Stats` - _optional_ - Boolean value specifying if the response must have query stats in the `Server-Timing` header.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+
+#### Response Headers
```
tacole02, line 597:
```suggestion
#### Response headers
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+
+- `Server-Timing` - _optional_ - Contains query execution stats if the `X-Mimir-Response-Query-Stats` header is set to `true`.
```
tacole02, line 599: Could we remove the word `execution` here and just say `Contains query stats`?

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+  The content of the header is a comma-separated list of key-value pairs, where the key is the name of the metric and the value is the metric value.
+  The contents of this header may change without notice. Example: `Server-Timing: encode;dur=0.041, series_count;c=31950, chunk_bytes;c=23791904, chunks_count;c=329964, index_bytes;c=28647863, series_fetched;c=29906, wall_time;dur=6259.888, queue;dur=0.020, response_size;c=118, response_time;dur=6267.381, cache_hit;c=0, cache_miss;c=0, sharded;c=0, split;c=0`
```
tacole02, line 601:
```suggestion
  The contents of this header might change without notice. For example: `Server-Timing: encode;dur=0.041, series_count;c=31950, chunk_bytes;c=23791904, chunks_count;c=329964, index_bytes;c=28647863, series_fetched;c=29906, wall_time;dur=6259.888, queue;dur=0.020, response_size;c=118, response_time;dur=6267.381, cache_hit;c=0, cache_miss;c=0, sharded;c=0, split;c=0`
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+  - Metrics:
+    - `encode`: The time spent in milliseconds at the frontend encoding the query's final results. Does not include time spent serializing results at the querier.
```
tacole02, line 603:
```suggestion
    - `encode`: The time, in milliseconds, spent at the frontend encoding the query's final results. This value does not include time spent serializing results at the querier.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `series_count`: The estimated number of series to be fetched for the query.
+    - `chunk_bytes`: The number of bytes of the chunks fetched for the query, after any deduplication.
+    - `chunks_count`: The number of chunks fetched for the query, after any deduplication.
+    - `index_bytes`: The number of index bytes fetched on the store-gateway for the query.
+    - `series_fetched`: The number of series fetched for the query.
+    - `wall_time`: The sum of all wall time spent in milliseconds in the querier to execute the query.
```
tacole02, line 609:
```suggestion
    - `wall_time`: The sum of all wall time, in milliseconds, spent in the querier to run the query.
```

tacole02: We try to avoid using the word `execute` in docs, when possible.

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `queue`: The sum of durations in milliseconds that the query spent in the queue, before it was handled by querier.
```
tacole02, line 610:
```suggestion
    - `queue`: The sum of durations, in milliseconds, that the query spent in the queue, before being handled by querier.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `response_size`: The number of bytes of the response in bytes.
```
tacole02, line 611:
```suggestion
    - `response_size`: The number of bytes of the response.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `response_time`: The time spent in milliseconds to process the query.
```
tacole02, line 612:
```suggestion
    - `response_time`: The time, in milliseconds, spent processing the query.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `cache_hit`: The number of results cache hits.
+    - `cache_miss`: The number of results cache misses.
+    - `sharded`: The number of sharded queries executed. 0 if sharding is disabled or the query can't be sharded.
```
tacole02, line 615:
```suggestion
    - `sharded`: The number of sharded queries run. Set to `0` if sharding is disabled, or if the query can't be sharded.
```

docs/sources/mimir/references/http-api/index.md
```diff
@@ -592,6 +592,28 @@ Requires [authentication](#authentication).
...
+    - `split`: The number of split partial queries executed. 0 if splitting is disabled or the query can't be split.
```
tacole02, line 616:
```suggestion
    - `split`: The number of split partial queries run. Set to `0` if splitting is disabled, or if the query can't be split.
```

tacole02 (approved):
Thanks for updating these docs! I left a few suggestions.

docs/sources/mimir/references/http-api/index.md
santileira: Done

docs/sources/mimir/references/http-api/index.md
tacole02: It just doesn't have a nice connotation. And, generally, using `run` works just as well 😄

56quarters:

santileira:
@dimitarvdimitrov could you review the PR again? Thanks

pkg/frontend/transport/handler.go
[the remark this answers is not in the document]
narqo: > It will probably be less surpsising if we use exactly the same names as in the query stats logs. I'm not sure if the difference here is worth the saving of ~100 bytes

I don't strong opinion on that one, but having a consistent naming is always welcome; for sure 👍

dimitarvdimitrov:
Screenshot 2025-03-19 at 10 41 15


@santileira this diff looks weird, can you check if the rebases went ok?

pkg/frontend/transport/handler.go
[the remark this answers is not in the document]
56quarters: I'd prefer the full names to match the fields we use elsewhere. I think the extra space that will take is fine since this is gated behind a new header and we can remove fields if we find they're not useful to expose.

pkg/frontend/transport/handler.go
```diff
@@ -258,6 +256,17 @@ func (f *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	var parts []string
+	if f.cfg.QueryStatsEnabled {
+		parts = getQueryStats(queryResponseTime, queryDetails)
+	}
+	if queryStatsHeaderNameOk {
+		parts = append(parts, getResponseQueryStats(queryResponseTime, queryResponseSize, queryDetails)...)
+	}
+	if len(parts) > 0 {
+		hs.Set(ServiceTimingHeaderName, strings.Join(parts, ", "))
```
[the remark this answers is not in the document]
santileira: @narqo I don't follow you. I think that the `trailer` is needed to send headers after the body but in this case we are just adding the headers after the body but not sending it after the body. Is this right?

pkg/frontend/transport/handler.go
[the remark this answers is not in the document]
santileira: @dimitarvdimitrov done. Could you check it again?

santileira:
> Screenshot 2025-03-19 at 10 41 15
> @santileira this diff looks weird, can you check if the rebases went ok?

Done! Could you check it again?

pkg/frontend/transport/handler.go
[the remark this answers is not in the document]
narqo: After one calls `w.WriteHeader(resp.StatusCode)` that ends the response headers section. Everything after, will be sent out to the wire as part of the response body (or trailers).

_Recall that an HTTP response is basically a stream of text. Once something was written to the wire, e.g. after the call of `io.Copy(resp, body)`, one cannot go back to the headers section, to amend them._

Our existing tests were inaccurate, it's turned. So the changes in the PR didn't catch that. But one can observe this behaviour by fixing the tests in the following way,

```diff
if tt.assertHeaders != nil {
-				tt.assertHeaders(t, resp.Header())
+				tt.assertHeaders(t, resp.Result().Header)
}
```

Refer to the docs of [`httptest.ResponseRecorder.Result()`][1]

[1]: https://pkg.go.dev/net/http/httptest#ResponseRecorder.Result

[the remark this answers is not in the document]
santileira: But I ended up in a deadlock situation where I need to get the queryResponseSize (`queryResponseSize, _ := io.Copy(w, resp.Body)`) to write the `Server-Timing` header but I can't write the header after I called the `io.Copy` so maybe we don't want to add the queryResponseSize as stats, right? What do you think?

pkg/frontend/transport/handler_test.go
```diff
@@ -359,6 +436,9 @@ func TestHandler_ServeHTTP(t *testing.T) {
 			require.NoError(t, err)
 			require.Empty(t, activities)
 
+			if tt.assertHeaders != nil {
+				tt.assertHeaders(t, resp.Header())
```
narqo, line 440: Related to the [discussion above](https://github.com/grafana/mimir/pull/10897/files#r2002657907), let's fix the existing test. Otherwise, we aren't testing what we wanted to.

```suggestion
				tt.assertHeaders(t, resp.Result().Header())
```

pkg/frontend/transport/handler.go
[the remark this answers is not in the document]
narqo: Generally, `Server-Timing` is a [valid use-case for an HTTP Trailer][1]. Thus, at the top I asked if that was the part of the plan.

@dimitarvdimitrov what do you think: do we have a use-case, where `Server-Timing` needs to be a header, or will something break if we move them to the trailer?

* Technically, we could buffer the response, before sending it to the wire — that doesn't feel fun. 
* Alternatively, we could try using `Content-Length` instead of the `queryResponseSize` (providing the downstream API sets one).
* Or, we could skip emitting the stats we can't calculate before.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing#server-timing_as_http_trailer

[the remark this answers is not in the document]
santileira: Good call. I will wait to have @dimitarvdimitrov 's thoughts.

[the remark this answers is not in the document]
dimitarvdimitrov: trailers would need different handling on the client side. I'm more inclined to try and use content-length like @narqo suggested. If it's not there, then we can drop it from the server-timing header.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 * [CHANGE] Store-gateway: Include posting sampling rate in sparse index headers. When the sampling rate isn't set in a sparse index header, store gateway rebuilds the sparse header with the configured `blocks-storage.bucket-store.posting-offsets-in-mem-sampling` value. If the sparse header's sampling rate is set but doesn't match the configured rate, store gateway either rebuilds the sparse header or downsamples to the configured sampling rate. #10684 #10878
 * [CHANGE] Distributor: Return specific error message when burst size limit is exceeded. #10835
 * [CHANGE] Ingester: enable native histograms ingestion by default, meaning`ingester.native-histograms-ingestion-enabled` defaults to true. #10867
+* [FEATURE] Query Frontend: Expose query stats in the `Server-Timing` header when the `X-Mimir-Response-Query-Stats` header is present in the request. #10192
```
dimitarvdimitrov, line 41:
```suggestion
* [FEATURE] Query Frontend: Expose query stats in the `Server-Timing` header when the `X-Mimir-Response-Query-Stats: true` header is present in the request. #10192
```

dimitarvdimitrov:
there is a failing test and it looks related to this change. Can you take a look? 


```
--- FAIL: TestHandler_ServeHTTP (0.24s)
    --- FAIL: TestHandler_ServeHTTP/handler_with_QueryStatsEnabled_and_X-Mimir-Response-Query-Stats_in_true,_check_ServiceTimingHeader (0.01s)
        handler_test.go:344: 
            	Error Trace:	/__w/mimir/mimir/pkg/frontend/transport/handler_test.go:344
            	            				/__w/mimir/mimir/pkg/frontend/transport/handler_test.go:440
            	Error:      	"querier_wall_time;dur=0, response_time;dur=7.212478, bytes_processed;val=0, samples_processed;val=0, encode_time_seconds;dur=0, estimated_series_count;val=0, fetched_chunk_bytes;val=0, fetched_chunks_count;val=0, fetched_index_bytes;val=0, fetched_series_count;val=0, query_wall_time_seconds;dur=0, queue_time_seconds;dur=0, response_size_bytes;val=0, response_time;dur=7.212478, results_cache_hit_bytes;val=0, results_cache_miss_bytes;val=0, sharded_queries;val=0, split_queries;val=0" does not contain "response_size_bytes;val=2"
            	Test:       	TestHandler_ServeHTTP/handler_with_QueryStatsEnabled_and_X-Mimir-Response-Query-Stats_in_true,_check_ServiceTimingHeader
FAIL

```

dimitarvdimitrov (approved):
LGTM, solid work 🙌 thank you for your time on this!

santileira:
@narqo could you review it again? Thanks in advance.

dimitarvdimitrov:
I think we can merge this. We can address comments in a follow-up if necessary

PR #10901, opened 2025-03-17
Increase delve version for mimir-ingest-storage development

jhalterman:
#### What this PR does

Increases the delve version for mimir-ingest-storage development since the current version is not compatible with go 1.24.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


charleskorn (commented):
Should we make the same change in `development/mimir-read-write-mode/dev.dockerfile`?

jhalterman:
Good call - I think read-write mode is a noop since the dockerfile doesn't actually use delve, but I updated it for microservices-mode, which does.

charleskorn (approved):

PR #10925, opened 2025-03-18
OTLP: Set Content-Length header on responses

aknuds1:
#### What this PR does

Modify OTLP distributor endpoint to set the `Content-Length` header on responses. Follow-up to #10852, where @charleskorn suggested we set also `Content-Length`.

Also improve OTLP tests to check for `Content-Type` and `Content-Length` headers, for both protobuf and JSON requests.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


aknuds1:
I see I need to fix a test.

narqo (approved):
Works for me 🔥

PR #10933, opened 2025-03-18
block-builder: Dedicated consumePartitionSectionPullMode for the scheduler

codesome:
#### What this PR does

Adds a dedicated function that consumes between given start and end offset with no partially consumed region. This is only for the use of scheduler.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -670,7 +670,7 @@ consumerLoop:
 }
 
 // consumePartitionSectionPullMode is for the use of scheduler based architecture.
-// Both startOffset and endOffset are inclusive, and must be valid offsets and not something in the future.
+// startOffset is inclusive, endOffset is exclusive, and must be valid offsets and not something in the future (endOffset can be technically 1 offset in the future).
 // All the records and samples between these offsets will be consumed and put into a block.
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -673,6 +669,122 @@ consumerLoop:
...
+consumerLoop:
+	for recOffset := int64(-1); recOffset < endOffset-1; {
+		if err := context.Cause(ctx); err != nil {
+			return 0, err
+		}
+
+		// PollFetches can return a non-failed fetch with zero records. In such a case, with only the fetches at hands,
+		// we cannot tell if the consumer has already reached the latest end of the partition, i.e. no more records to consume,
+		// or there is more data in the backlog, and we must retry the poll. That's why the consumer loop above has to guard
+		// the iterations against the endOffset, so it retried the polling up until the expected end of the partition is reached.
```
seizethedave, line 732:
```suggestion
		// the iterations against the endOffset, so it retries the polling up until the expected end of the partition is reached.
```
(I know it's copy-pasted.)

seizethedave (approved):
Great! 1 tiny comment.

PR #10828, opened 2025-03-07
Safeguard backport workflow

armandgrillet:
This check if the PR needs to be backported and if it's coming from a fork before trying the actual backport.

PR #10819, opened 2025-03-06
Use the app token in helm-weekly-release-pr.yml

armandgrillet:
This follows the documentation in https://github.com/grafana/shared-workflows/tree/main/actions/get-vault-secrets.

PR #10254, opened 2024-12-16
[CI] Add auto approval/merge for helm weekly release

ying-jeanne:
#### What this PR does
Add auto approval/merge for helm weekly release, the condition for auto-approval/merge is
- the PR is raised by grafana-bot
- the branch name contains `helm-chart-weekly-`

Since we already switched to renovate and no dependabot enabled anymore on Mimir, remove the out dated automerge logic on dependabot PR and repurpose it to helm chart weekly auto approval/merge

Example PR that we would like to auto-approval/merge https://github.com/grafana/mimir/pull/10253

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [na] Tests updated.
- [na] Documentation added.
- [na] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [na ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #10943, opened 2025-03-19
Simplify github workflows to use auto-generater CI token

narqo:
#### What this PR does

This is the follow-up to #10819, #10820, etc.

The two workflows (backport and weekly-helm-release) don't need the private token to do the job. That is, these workflows only open a new PR, and, thus, don't act as a "maintainer". Switching the workflows to the temporal auto-generated `GITHUB_TOKEN` (ref [GitHub docs][1]) makes two things simpler:

1. We can use private Mimir token to auto-review the weekly PRs (this has been [broken since forever](https://github.com/grafana/mimir/pull/10254#issuecomment-2560989258)).
2. Backporting of the PRs opened by the contributors will just work, because the contributors won't try to access the private Vault — it's [broken now](https://github.com/grafana/mimir/actions/runs/13930549296)).

Closes #10828 (mine fixes the same problem but in a simpler way).

Note that Tempo is currently doing the same (ref [grafana/tempo][2]).

[1]: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication
[2]: https://github.com/grafana/tempo/blob/9bc9043e66d8af65fa22f9b61a35b55ffcaa1195/.github/workflows/backport.yml

### DIFF
--- a/.github/workflows/scripts/helm-weekly-release.sh
+++ b/.github/workflows/scripts/helm-weekly-release.sh
@@ -104,4 +104,4 @@ update_yaml_node $chart_file .version $new_chart_version
 
 make TTY='' doc
 
-echo "::set-output name=new_chart_version::$new_chart_version"
+echo "new_chart_version=$new_chart_version" >> $GITHUB_OUTPUT
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -6,31 +6,11 @@ on:
       - closed
       - labeled
 
-permissions:
-  contents: read
-  id-token: write
-
 jobs:
   main:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Retrieve GitHub App credentials from Vault
-        id: get-secrets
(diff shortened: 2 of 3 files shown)


narqo:
I've tested the helm-weekly workflow by triggering from the branch. Here's how the created PR looked like https://github.com/grafana/mimir/pull/10944

.github/workflows/scripts/helm-weekly-release.sh
```diff
@@ -104,4 +104,4 @@ update_yaml_node $chart_file .version $new_chart_version
 
 make TTY='' doc
 
-echo "::set-output name=new_chart_version::$new_chart_version"
+echo "new_chart_version=$new_chart_version" >> $GITHUB_OUTPUT
```
narqo, line 107: `set-output` is [deprecated since 2022](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

armandgrillet (approved):
LGTM

56quarters (approved):
Nicely done!

PR #10266, opened 2024-12-17
KEDA autoscaling additional requirement

bjorns163:
#### What this PR does

I've been testing the KEDA autoscaling option on my environment and was missing some fields to get this working.
The missing fields, authentication and label filters that were needed I've added. 

#### Which issue(s) this PR fixes or relates to

Fixes [#10265](https://github.com/grafana/mimir/issues/10265)

PR #10945, opened 2025-03-19
Add support for all prometheus query operators (=, !=, =~ and !~) in KEDA Autoscaling

Ranexia04:
#### What this PR does
Add support for all prometheus query operators (=, !=, =~ and !~) in KEDA Autoscaling in Mimir Helm Chart
Added an optional variable toPromQLLabelOperator that determines the operator when adding filters to the default queries in KEDA Autoscaling. Key labels in toPromQLLabelSelector match key labels in toPromQLLabelOperator to update the operator (default is still '=').

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov:
can you share some details around the use case for this change?

Ranexia04:
> can you share some details around the use case for this change?

My personal use case is when Mimir and Loki are deployed in the same OpenShift project (should be the same in vanilla Kubernetes). With the default queries, pods from both Mimir and Loki get monitored. Now, if all pods had a label indicating the software, this would be overcome. However, some queries look at metrics computed by OpenShift and not the Pods themselves; therefore, these labels do not include a "cluster" metric. The only way I found of separating Mimir and Loki would be through the "pod" label but we would need to apply a regex operation for this, hence the PR.

Ranexia04:
> Thanks for opening the PR. In retrospect I think we made a mistake with the current format of the matchers (a map). What do you think about changing the current implementation to receive a string? That way you can provide as many matchers and repeat them and use whichever operator you want. For example
> 
> ```yaml
>   toPromQLLabelSelectors: 'cluster!="eu-west", cluster!="eu-east"'
> ```
> 
> This is a breaking change. I think it's to ok to make it because 1) the selector change (#10266) is yet to be included in a stable release and 2) autoscaling is experimental, so we don't have very strong guarantees on YAML API.
> 
> @bjorns163 what do you think about this idea as the one who worked on #10266

This idea would fix this specific problem, but perhaps a list would be more intuitive and less error-prone since users would not need to think about whether they need prior commas and such. Here is my idea:

```
  toPromQLLabelSelectors:
    - cluster!="eu-west"
    - cluster!="eu-east"'
```

bjorns163:
I'd stay adding for more query operators sounds like a nice addition. I'd personally be for a list set up like @Ranexia04 mentioned.

Ranexia04:
> I'd stay adding for more query operators sounds like a nice addition. I'd personally be for a list set up like @Ranexia04 mentioned.

Updated the PR with the list changes

dimitarvdimitrov (approved):
awesome! Can you include a changelog entry in `operations/helm/charts/mimir-distributed/CHANGELOG.md` so that operators know what has changed?

After that, I think I can merge on green

Ranexia04:
Like so?

operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## 5.7.0-rc.0
 
+* [CHANGE] KEDAAutoscaling: Changed toPromQLLabelSelector from object to list of strings, adding support for all PromQL operators. #10945
```
tacole02, line 34:
```suggestion
* [CHANGE] KEDA Autoscaling: Changed toPromQLLabelSelector from object to list of strings, adding support for all PromQL operators. #10945
```

tacole02 (approved):

bjorns163:
Did someone test this change? I see 5.7 is now release with this change in it, but it's not working. The output in the scaled object files is incorret.  I've but in the vales files multiple different ways of butting the values for the `toPromQLLabelSelector` but all give wrong output:

Vaules file:
```
  toPromQLLabelSelector:
  - 'cluster_name="test"'
  - 'cluster_name2="test2"'
  - 'cluster_name3="test3"'
  ```
  
  Output:
  ```
  %!s(int=0)="cluster_name="test""
  ```

**Comment by Ranexia04** [comment]
I did test it locally before the merge.

values-test.yaml:
```
kedaAutoscaling:
  toPromQLLabelSelector:
  - 'cluster_name="test"'
  - 'cluster_name2="test2"'
  - 'cluster_name3="test3"'

distributor:
  enabled: true
  kedaAutoscaling:
    enabled: true
```

helm template -f values-test.yaml my-mimir-distributed grafana/mimir-distributed --version 5.8.0-weekly.337 | grep query

```
...
      query: max_over_time(sum(sum by (pod) (rate(container_cpu_usage_seconds_total{container="distributor",namespace="test",cluster_name="test",cluster_name2="test2",cluster_name3="test3"}[5m])) and max by (pod) (up{container="distributor",namespace="test",cluster_name="test",cluster_name2="test2",cluster_name3="test3"}) > 0)[15m:]) * 1000
...
```

PR #10947, opened 2025-03-19
Compactor: Upload Block Data and Index Concurrently

dmwilson-grafana:
#### What this PR does

This PR enables concurrent uploads of block's segment files and index file during compaction runs. We occasionally see large blocks take >10 minutes to upload from the compactor to object storage. This PR addresses these slow upload times with two changes:

1. Upload Index and Chunks Concurrently &mdash; When we call `block.Upload`, we upload a block's chunks and index sequentially (see: [tsdb.block](https://github.com/grafana/mimir/blob/1ff367ef58987cd1941de03a8d6923fde82dfdd3/pkg/storage/tsdb/block/block.go#L116-L122)).

2. Set Higher Concurrency on Chunks Upload &mdash; When we call `objstore.UploadDir` to upload blocks' chunks, we currently fallback to the default of `1` (see: [thanos/objstore](https://github.com/thanos-io/objstore/blob/3f8838fdd5a2c99bc3d771ef78816201e7d0b69e/objstore.go#L247)). This PR sets concurrency>=1 as a function of block size, up to a limit of `maxBlockUploadConcurrency` (8).

For a hypothetical block with 12GB of data and a 1GB index, the current upload options would lead to a 3m upload (13GB at  70MB/s). With the changes from this PR, this blocks' 24 segment files would be split across 8 goroutines each uploading multiple 512MB chunks. Provided the network bandwidth is available, the upload completes after the index uploader and the last block data uploader complete (est. 30s).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/compactor/bucket_compactor.go
```diff
@@ -461,7 +462,14 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
 		blockToUpload := blocksToUpload[idx]
 		bdir := filepath.Join(subDir, blockToUpload.ulid.String())
 		begin := time.Now()
-		if err := block.Upload(ctx, jobLogger, c.bkt, bdir, nil); err != nil {
+
+		chunksDir := filepath.Join(bdir, block.ChunksDirname)
+		opts, err := getUploadOptions(chunksDir)
+		if err != nil {
+			level.Warn(jobLogger).Log("msg", "using default block upload options", "block", blockToUpload.ulid.String(), "shard", blockToUpload.shardIndex, "err", err)
```
56quarters, line 469: No need to call `.String()` on the ULID

pkg/compactor/bucket_compactor.go
```diff
@@ -501,6 +509,22 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
 	return true, compIDs, nil
 }
 
+func getUploadOptions(chunksDir string) ([]objstore.UploadOption, error) {
+	var con int
+	if err := filepath.WalkDir(chunksDir, func(_ string, _ fs.DirEntry, _ error) error {
```
56quarters, line 514: The upload concurrency option sets an upper limit. I don't think it's really necessary to figure out exactly what it should be. The default maximum of 8 seems fine to me.

56quarters (approved):
LGTM, just one suggestion.

pkg/compactor/bucket_compactor.go
dmwilson-grafana: Agreed, I originally planned on this scaling by a configurable amount with TSDB segment files (e.g. 8:1, 4:1, 2:1), but since it's just 1:1, I think it's redundant. Removing this.

pkg/compactor/block_upload.go
```diff
@@ -43,6 +43,7 @@ const (
 	validationHeartbeatInterval = 1 * time.Minute       // Duration of time between heartbeats of an in-progress block upload validation
 	validationHeartbeatTimeout  = 5 * time.Minute       // Maximum duration of time to wait until a validation is able to be restarted
 	maximumMetaSizeBytes        = 1 * 1024 * 1024       // 1 MiB, maximum allowed size of an uploaded block's meta.json file
+	maxBlockUploadConcurrency   = 8                     // Max concurrency allowed while uploading a single block's TSDB segment files
```
seizethedave, line 46: This would make a good config knob. Selfishly, so that we can dial it back if it ends up causing drama. What do you think?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 * [ENHANCEMENT] Dashboards: Include absolute number of notifications attempted to alertmanager in 'Mimir / Ruler'. #10918
 * [ENHANCEMENT] Alerts: Make `MimirRolloutStuck` a critical alert if it has been firing for 6h. #10890
+* [ENHANCEMENT] Compactor: Upload block index file and multiple segment files concurrently. Concurrency scales linearly with block size up to `maxBlockUploadConcurrency`. #10947
```
seizethedave, line 11: I think the scaling linearly bit is no longer true.

seizethedave (commented):
LGTM%question

CHANGELOG.md
dmwilson-grafana: I can reword it to be more clear, but this is still technically true. In thanos/objstore ([link](https://github.com/thanos-io/objstore/blob/a0136a6f898d43ec0e654a98ac322554f2ae259b/objstore.go#L346-L386)) it starts a goroutine uploading a file to the maxConcurrency set. For any number of segment files < maxConcurrency; you'd get that number of goroutines, and then constant at maxConcurrency beyond that.

seizethedave: Ah, nvm then.

pkg/compactor/block_upload.go
dmwilson-grafana: I was on the fence about this. We have several concurrency related options that affect the compactor already, i.e `-blocks-storage.bucket-store.tenant-sync-concurrency`, `-blocks-storage.bucket-store.block-sync-concurrency`, and `-blocks-storage.bucket-store.meta-sync-concurrency`. Is there another way to set it back to 1 if needed?

dmwilson-grafana: I'll just mark it an `advanced` option and make it clear it's within-block concurrency to differentiate from the others.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3962,6 +3962,11 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -compactor.max-lookback
 [compactor_max_lookback: <duration> | default = 0s]
 
+# (advanced) Maximum number of TSDB segment files that can be uploaded
```
tacole02, line 3965:
```suggestion
# (advanced) Maximum number of TSDB segment files that you can upload
```

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -1692,6 +1692,7 @@ type mockConfigProvider struct {
 	verifyChunks                 map[string]bool
 	perTenantInMemoryCache       map[string]int
 	maxLookback                  map[string]time.Duration
+	maxPerBlockUploadConurrency  map[string]int
```
seizethedave, line 1695:
```suggestion
	maxPerBlockUploadConcurrency  map[string]int
```

pkg/compactor/blocks_cleaner_test.go
```diff
@@ -1707,6 +1708,7 @@ func newMockConfigProvider() *mockConfigProvider {
 		verifyChunks:                 make(map[string]bool),
 		perTenantInMemoryCache:       make(map[string]int),
 		maxLookback:                  make(map[string]time.Duration),
+		maxPerBlockUploadConurrency:  make(map[string]int),
```
seizethedave, line 1711: here, too.

seizethedave (approved):
LGTM modulo typo. 😸

PR #10958, opened 2025-03-21
Feat/dynamic step

Dhouib-Mohamed:
# What this PR does
This PR introduces a middleware for the Mimir frontend to dynamically optimize query step sizes for complex queries.

### Problem Statement
Many queries from Grafana dashboards, especially those involving high-cardinality data, can take more than 10 seconds to execute. This is often due to:
- Queries with a low step size (step) and a long time range, which significantly increase the amount of data fetched.
- Users importing large dashboards without adjusting the step size to match their dataset's growth over time. Initially, when data volume was lower, these settings may have worked fine, but as data complexity increases, queries become slower.

### Proposed Solution
This middleware detects "complex" queries—specifically, high-cardinality queries with a low step size—and dynamically adjusts the step parameter to improve query performance. By increasing the step size where appropriate, we can:
- Reduce query execution time.
- Improve dashboard responsiveness.
- Allow users to retrieve more granular data on demand rather than by default.

### Notes
The mathematical function and step-size limits used in this implementation are preliminary estimates and have not yet been validated in a high-load production environment. Further tuning and real-world testing will be required.

#### Checklist

- [*] Tests updated.
- [*] Documentation added.

### DIFF
--- a/docs/sources/mimir/configure/configuration-parameters/index.md
+++ b/docs/sources/mimir/configure/configuration-parameters/index.md
@@ -1876,6 +1876,14 @@ results_cache:
 # CLI flag: -query-frontend.shard-active-series-queries
 [shard_active_series_queries: <boolean> | default = false]
 
+# (experimental) True to enable dynamic step calculation for range series queries.
+# CLI flag: -query-frontend.dynamic_step_enabled
+[dynamic_step_enabled: <boolean> | default = false]
+
+# (experimental) The maximum complexity level to allow in a query when dynamic
+# step calculation is enabled.
+[dynamic_step_complexity_threshold: <float> | default = 1]
+
 # (experimental) Set to true to use the zero-allocation response decoder for
 # active series queries.
 # CLI flag: -query-frontend.use-active-series-decoder
--- a/pkg/frontend/querymiddleware/dynamic_step.go
+++ b/pkg/frontend/querymiddleware/dynamic_step.go
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package querymiddleware
+
+import (
+	"context"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"math"
+	"strconv"
+)
+
+const (
+	complexityThreshold = 1
+)
+
+// dynamicStep is a MetricsQueryHandler that change the step of a query dynamically
+// based on the complexity of the query.
+type dynamicStep struct {
+	next   MetricsQueryHandler
+	logger log.Logger
+
+	dynamicStepUsageCount prometheus.CounterVec
+	complexityThreshold   float64
+}
+
+func newDynamicStepMiddleware(complexityThreshold float64, logger log.Logger, registerer prometheus.Registerer) MetricsQueryMiddleware {
+	dynamicStepUsageCount := promauto.With(registerer).NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "mimir_query_frontend_dynamic_step_usage_count",
+			Help: "Number of queries whose step has been adjusted dynamically.",
+		},
+		[]string{"cardinality", "step", "new_step"},
+	)
+	return MetricsQueryMiddlewareFunc(func(next MetricsQueryHandler) MetricsQueryHandler {
+		return &dynamicStep{
+			next:   next,
+			logger: logger,
+
+			dynamicStepUsageCount: *dynamicStepUsageCount,
+			complexityThreshold:   complexityThreshold,
+		}
+	})
+}
+
+// Do change the step of a query dynamically based on the complexity of the query in the request.
+func (d *dynamicStep) Do(ctx context.Context, request MetricsQueryRequest) (Response, error) {
+
+	// Get the cardinality of the query
+
+	hints := request.GetHints()
+
+	if hints == nil || hints.GetCardinalityEstimate() == nil {
+		level.Debug(d.logger).Log("msg", "no cardinality estimation required for dynamic step. Skipping it")
+		return d.next.Do(ctx, request)
+	}
+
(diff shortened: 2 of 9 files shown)


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1876,6 +1876,14 @@ results_cache:
 # CLI flag: -query-frontend.shard-active-series-queries
 [shard_active_series_queries: <boolean> | default = false]
 
+# (experimental) True to enable dynamic step calculation for range series queries.
```
tacole02, line 1879:
```suggestion
# (experimental) Set to true to enable dynamic step calculation for range series queries.
```

tacole02 (approved):

56quarters (changes requested):
Thank you for this PR but changing the step of a range query contrary to the step included in the request is:
* Surprising behavior that changes the response returned to users from what they would normally expect
* Isn't allowed for Prometheus compatible APIs as far as I can tell. 

Because of this, we're unlikely to merge this PR.

tacole02 (approved):

dimitarvdimitrov:
> * Reduce query execution time.

@Dhouib-Mohamed do you have any data on this? A lot of compute is spent regardless of the step. Some comparison between smaller and larger steps would help inform this decision. 

@56quarters do you think we can keep this as an opt-in feature for operators to decide if they want it or do you think it doesn't have a place in Mimir?

pkg/frontend/querymiddleware/dynamic_step.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// getNewStep returns the new step based on the complexity of the query and the current step.
+func (d *dynamicStep) getNewStep(cardinality uint64, step int64) int64 {
+
+	complexity := float64(cardinality) / float64(step)
+
+	if complexity > d.complexityThreshold {
+		// Calculate the new step based on the complexity
+		step = 100 * int64(10*math.Pow(complexity/d.complexityThreshold, 1.2))
```
dimitarvdimitrov, line 88: how did you come up with this formula?

Dhouib-Mohamed: This was mainly created by try and refine , I started by reading the cardinality formula, trying some ideas , testing it with some examples with mimirir continious test ( tweaked it a little to generate some queries with different parameters ) and some manual tests but there is no direct logic to it

Dhouib-Mohamed:
About the data , I have none for now but i was planning to set up a test cluster with production metrics to test this functionality 
But I need to see how likely is it to be considered since this is relatively time consuming

56quarters:
> @56quarters do you think we can keep this as an opt-in feature for operators to decide if they want it or do you think it doesn't have a place in Mimir?

I'm leaning towards "no" because I think it makes the behavior of Mimir harder to understand.

It also has the potential to make performance worse. For example, Grafana dashboards adjust start and end times for range queries to be multiples of the step and we cache results based on that. If the step is being adjusted after Grafana has already picked something to help with caching, we run the risk of reducing our ability to cache queries.

I think the correct place for adjusting the step based on a heuristic is in Grafana, if it doesn't already do something like this.

56quarters:
> I think the correct place for adjusting the step based on a heuristic is in Grafana, if it doesn't already do something like this.

Though I should point out, if the goal of this is to improve query performance I'd expect to see some numbers from a proof-of-concept that proves this is a valid approach to improving query performance before a PR to merge something to main.

Dhouib-Mohamed:
Sorry for the delay — I had to take care of a few other things. I've now completed the benchmarking and documented the results to assess the impact of this PR. 
Let me know if you have any feedback!
[Dynamic Step Middleware Performance Report (1).pdf](https://github.com/user-attachments/files/19778180/Dynamic.Step.Middleware.Performance.Report.1.pdf)

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1886,6 +1886,16 @@ results_cache:
 # CLI flag: -query-frontend.query-result-response-format
 [query_result_response_format: <string> | default = "protobuf"]
 
+# (experimental) True to enable dynamic step calculation for range queries based
```
tacole02, line 1889:
```suggestion
# (experimental) Enables dynamic step calculation for range queries based
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1886,6 +1886,16 @@ results_cache:
...
+# on the query's complexity.
+# CLI flag: -query-frontend.dynamic-step-enabled
+[dynamic_step_enabled: <boolean> | default = false]
+
+# (advanced) Complexity threshold for dynamic step adjustment ( Only taken into
+# consideration if dynamic step is enabled ).
```
tacole02, line 1895:
```suggestion
# (advanced) Complexity threshold for dynamic step adjustment. This setting is only taken into
# consideration if you enable dynamic step.
```

tacole02 (approved):

dimitarvdimitrov:
@Dhouib-Mohamed, thank you for the work so far including the experiment you ran, the results you posted, and the collaboration on the PR so far. However, at this point we don't think this is a change that will fit in Mimir. There are several reasons for this:

* Making this change in the Prometheus API behaviour could introduce unexpected behavior for end users. Think of a large organization where metrics are controlled centrally.
* It is out of scope for Mimir. If the tradeoff is to choose less complete query results in favour of speed, then Grafana or a proxy in front of Mimir, which users can choose to route to or not, is a better component to do this. Mimir's goal is to provide fast query results for what was actually requested, not approximate the request to something which is fast to serve. We encourage you to explore implementing this functionality in Grafana or a proxy if it aligns with your needs.
* The [Mimir query engine (MQE)](https://grafana.com/docs/mimir/latest/references/architecture/mimir-query-engine/) is gaining maturity and will soon be promoted to a stable feature. The MQE deals much more efficiently with high-cardinality series through more efficient query processing, so this change might have limited impact in those cases.

Apologies for taking so long to come to this conclusion and for closing this PR.

PR #10959, opened 2025-03-21
Alertmanager: Broadcast oversized Grafana state

santihernandezc:
When the Grafana configuration is promoted, Mimir fetches the Grafana state from object storage and merges it with its own, incorporating any silences and nflog entries. After merging the Grafana state and broadcasting it to other pods, it deletes it from object storage. The resulting state is persisted as normal Mimir Alertmanager state.

As we use upstream code to perform the state merge, [oversized messages are dropped](https://github.com/grafana/mimir/blob/d6c7e94e46d51e7851b15dfa9522c7abfca9f13c/vendor/github.com/prometheus/alertmanager/silence/silence.go#L942-L967). As these oversized silences are not broadcasted, other pods that could not reach the Grafana state before it was deleted won't have those silences.

<details><summary>Grafana Alertmanager tenants with silences larger than Prometheus' message size limit will only have their silences in the Alertmanager pod that merged them before they were deleted.</summary>
<br/>
Only one out of three Alertmanagers picked up the silence.

image

</details>

<details><summary>To fix this, this PR checks for silences larger than Prometheus' message size limit and broadcasts them.</summary>
<br/>

The silence was gossiped to all Alertmanager instances.
image

</details>

### DIFF
--- a/pkg/alertmanager/state_replication.go
+++ b/pkg/alertmanager/state_replication.go
@@ -262,6 +262,23 @@ func (s *state) Ready() bool {
 	return s.Service.State() == services.Running
 }
 
+func (s *state) MergeGrafanaState(fs []*clusterpb.FullState) error {
+	if err := s.MergeFullStates(fs); err != nil {
+		return err
+	}
+
+	for _, fs := range fs {
+		for _, p := range fs.Parts {
+			if cluster.OversizedMessage(p.Data) {
+				// When merging state, upstream Alertmanager code drops oversized messages.
+				// Manually broadcast oversized Grafana states to avoid missing silences/nflog entries.
+				s.broadcast(p.Key, p.Data)
+			}
+		}
+	}
+	return nil
+}
+
 // MergeFullStates attempts to merge all full states received from peers during settling.
 func (s *state) MergeFullStates(fs []*clusterpb.FullState) error {
 	s.mtx.Lock()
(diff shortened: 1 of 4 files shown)


pkg/alertmanager/state_replication.go
```diff
@@ -262,6 +262,23 @@ func (s *state) Ready() bool {
 	return s.Service.State() == services.Running
 }
 
+func (s *state) MergeGrafanaState(fs []*clusterpb.FullState) error {
+	if err := s.MergeFullStates(fs); err != nil {
+		return err
+	}
+
+	for _, fs := range fs {
```
JacobsonMT, line 270: Feels like we should have a lock around this whole process but `MergeFullStates` already locks internally.

Why not add the broadcasting directly to `MergeFullStates`? This is technically a bug in Cloud AM to, even though there aren't really many (any?) current avenues where new states are created directly in storage without calling  `setSilence`.

JacobsonMT (approved):
Approving as it's likely fine, locking is less important for the broadcast than it is for the merge. We can consider adding the fix to `MergeFullStates` later.

pkg/alertmanager/state_replication.go
santihernandezc: I don't think there are any ways we could be getting a silence from the state without it being created via `setSilence()` before. This is only possible for Grafana AM silences, so I'd avoid changing how merging full state works for all tenants.

stevesg (approved):
👍

Issue #10963, opened 2025-03-21
Idea: Allow runtime config managed externally


fyang13:
### What is the problem you are trying to solve?

In some deployments, user might want to manage runtime config dynamically outside of Helm-chart(e.g.: max_global_series_per_user for an environment with lots of tenants), it would be nice if user can optionally disable runtime config update through value file.

### Which solution do you envision (roughly)?

Update runtime-config template to observe the apply_runtime.enabled value, default to true.

### How long do you think this would take to be developed?

Small (<= 1 month dev)

narqo:
> In some deployments, user might want to manage runtime config dynamically outside of Helm-chart

Mimir's [`-runtime-config.file`][1] config option (and its YAML variant) takes a list of files. Rather than expanding the chart with a new on/off switch, have you considered passing a second file, that your deployment manages externally?

This is what we do in Grafana cloud:

```
spec:
  containers:
  - args:
    ···
    -runtime-config.file=/etc/<··>/overrides.yaml,/etc/<··>/overrides.yaml,/etc/<··>/overrides.json.gz
```

[1]: https://grafana.com/docs/mimir/latest/configure/about-runtime-configuration/

fyang13:
@narqo I like that approach! This way user can optionally use helm-chart to manage the initial set of the config and handle the dynamic part externally. And I see your example has .gz format, so that will also help the situation when config map size grows beyond 1MB.

56quarters:
Since multiple runtime config files can be used it seems like that solves this use case. I'm going to close this issue but please re-open if needed.

fyang13:
@56quarters This issue is still needed for the deployment that's being managed via helm-chart? I don't see current helm-chart in this repo support the runtime-config.file option and it seems it's using configmap only? I have a draft PR [here](https://github.com/grafana/mimir/pull/10964) and probably will just go with the suggestion Dimitar has. What do you think?

56quarters:
> [@56quarters](https://github.com/56quarters) This issue is still needed for the deployment that's being managed via helm-chart? I don't see current helm-chart in this repo support the runtime-config.file option and it seems it's using configmap only? I have a draft PR [here](https://github.com/grafana/mimir/pull/10964) and probably will just go with the suggestion Dimitar has. What do you think?

You can specify a list of runtime configuration files from within the main configuration file:

```
runtime_config:
  # (advanced) How often to check runtime config files.
  # CLI flag: -runtime-config.reload-period
[...]

PR #10964, opened 2025-03-21
Optional runtime config

fyang13:
#### What this PR does
Allow runtime config be managed outside of helm-chart optionally, default behavior remain the same

Fixes [ #<10963>](https://github.com/grafana/mimir/issues/10963)

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/templates/runtime-configmap.yaml
+++ b/operations/helm/charts/mimir-distributed/templates/runtime-configmap.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.federation_frontend.disableOtherComponents }}
+{{- if and (not .Values.federation_frontend.disableOtherComponents) (.Values.apply_runtime.enabled) }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov:
Thanks for opening this! I think the use case makes sense and we would accept a change in this direction.

we already have a similar setup for the regular config. It's a Secret or ConfigMap which the operator provisions manually and is mounted to Mimir/GEM pods.

https://github.com/grafana/mimir/blob/3f01cb6b64e39d20de062e07b02e714394a7ad04/operations/helm/charts/mimir-distributed/values.yaml#L88-L100


I'm inclined to go that route to have consistency even though it's much bigger change than what you have here. What do you think, would be up to doing it?


```yaml
useExternalRuntimeConfig: false
runtimeConfigStorageType: ConfigMap
externalRuntimeConfigSecretName: '{{ include "mimir.resourceName" (dict "ctx" . "component" "runtime") }}'
# no need for externalConfigVersion since the runtime config is hot-reloaded. The version was used to trigger Pod restarts
```

dimitarvdimitrov:
closing as https://github.com/grafana/mimir/issues/10963 was also closed

PR #10981, opened 2025-03-24
Expose per-user metrics for unknown series references during WAL/WBL replay

pr00se:
Expose per-user versions of prometheus metrics added in https://github.com/prometheus/prometheus/pull/16166

Metrics are only exported when non-zero, so overall increase in metric count should be minimal.


56quarters (approved):
LGTM

seizethedave (approved):

tacole02 (approved):

PR #10983, opened 2025-03-25
MQE: set correct lookback delta on statement

charleskorn:
#### What this PR does

This PR fixes the issue where the `parser.Statement` returned by a MQE `Query` does not have the correct lookback delta set if the engine default or global default is being used.

As far as I can see, nothing uses this information, so there was no impact to this, but wanted to tidy it up before it did cause a problem.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -3392,3 +3392,57 @@ func TestQueryStats(t *testing.T) {
 		})
 	}
 }
+
+func TestQueryStatementLookbackDelta(t *testing.T) {
+	limitsProvider := NewStaticQueryLimitsProvider(0)
+	stats := stats.NewQueryMetrics(nil)
+	logger := log.NewNopLogger()
+
+	runTest := func(t *testing.T, engine promql.QueryEngine, queryOpts promql.QueryOpts, expectedLookbackDelta time.Duration) {
+		q, err := engine.NewInstantQuery(context.Background(), nil, queryOpts, "1", time.Now())
+		require.NoError(t, err)
+
+		require.Equal(t, expectedLookbackDelta, q.Statement().(*parser.EvalStmt).LookbackDelta)
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/engine_test.go
```diff
@@ -3392,3 +3392,57 @@ func TestQueryStats(t *testing.T) {
...
+		})
+
+		t.Run("no query options provided", func(t *testing.T) {
+			runTest(t, engine, nil, defaultLookbackDelta)
+		})
+
+		t.Run("lookback delta set in query options", func(t *testing.T) {
+			queryOpts := promql.NewPrometheusQueryOpts(false, 14*time.Minute)
+			runTest(t, engine, queryOpts, 14*time.Minute)
+		})
```
jhesketh, line 3425: We should test where a query option sets the lookback delta to 0

pkg/streamingpromql/engine_test.go
```diff
@@ -3392,3 +3392,57 @@ func TestQueryStats(t *testing.T) {
...
+
+		t.Run("no query options provided", func(t *testing.T) {
+			runTest(t, engine, nil, 12*time.Minute)
+		})
+
+		t.Run("lookback delta set in query options", func(t *testing.T) {
+			queryOpts := promql.NewPrometheusQueryOpts(false, 14*time.Minute)
+			runTest(t, engine, queryOpts, 14*time.Minute)
+		})
```
jhesketh, line 3446: ditto here

jhesketh (commented):
Nice catch! Just one suggestion

pkg/streamingpromql/engine_test.go
charleskorn: 0 means "not set", and is tested in the first test case above

jhesketh: not sure how I missed that first case, but yep, lgtm, thanks

jhesketh (approved):

PR #10986, opened 2025-03-25
github: fix helm weekly auto-review

narqo:
#### What this PR does

This is another follow-up to #10926 

As I discovered in https://github.com/grafana/mimir/pull/10984 the PRs created by the `github-actions[bot]` don't trigger the CI jobs. This, in turn, blocks the auto-review.

This PR tries to fix it but changing (reversing) the flow:
- now `mimir-github-bot` opens a weekly PR (as it used to be)
- the `github-actions` bot approves and auto-merges the PR.

To be fair, I'm not yet sure if this will work with how Mimir codeowners are currently defined. But I'll test it after merging this one.


.github/workflows/helm-weekly-release-reviewer.yml
```diff
@@ -48,13 +32,12 @@ jobs:
           --approve -b "**I'm approving** this pull request, since it is a helm release."
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
-          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
dimitarvdimitrov, line 35: i thought we decomissioned using GITHUB_TOKEN PAT. Or is this the ephemeral token that gets issued for every action run?

dimitarvdimitrov (approved):
makes sense, but honestly i can never spot faults with gh actions, so if you try it and it works 🤷 ✅

PR #10988, opened 2025-03-25
Add new metric to help debug HA problems

zenador:
#### What this PR does

Adds a new metric to track problems with HA. According to [this note](https://grafana.com/docs/mimir/latest/configure/configure-high-availability-deduplication/#distributor-high-availability-ha-tracker), if the first series in the request happens to be incorrectly configured, HA will not work properly for all the other series in the request. If I'm understanding how this works correctly, this means that for users with a lot of series, if only a minority of them are configured incorrectly, it can be very hard to track down the problem. This PR tries to fix that by using a new metric to show the metric names of the series that are misconfigured, and the reason why the HA fails e.g. is it the cluster or replica label that's missing? It also adds a log line with more details.

This could potentially create a lot of new series and log lines, so it's behind a per-tenant config flag `track_ha_failures` (off by default) so we can turn it on only temporarily for users as needed.

Will update changelog after getting feedback on this approach.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -1010,8 +1010,8 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
 
 		haReplicaLabel := d.limits.HAReplicaLabel(userID)
 		cluster, replica, metricName := findHALabels(haReplicaLabel, d.limits.HAClusterLabel(userID), req.Timeseries[0].Labels)
-		// Make a copy of these, since they may be retained as labels on our metrics, e.g. dedupedSamples.
-		cluster, replica = strings.Clone(cluster), strings.Clone(replica)
+		// Make a copy of these, since they may be retained as labels on our metrics, e.g. dedupedSamples, failedHARequests.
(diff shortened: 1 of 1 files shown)


tacole02 (approved):
Thank you!

pkg/distributor/distributor.go
```diff
@@ -1036,6 +1051,10 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
 		} else {
 			// If there wasn't an error but removeReplica is false that means we didn't find both HA labels.
 			d.nonHASamples.WithLabelValues(userID).Add(float64(numSamples))
+			if d.limits.TrackHAFailures(userID) {
+				d.failedHARequests.WithLabelValues(userID, metricName, keepReplicaReason).Inc()
```
bboreham, line 1055: I think you need to copy `metricName`, for this reason:
```
// Strings in pushReq may be pointers into the gRPC buffer which will be reused, so must be copied if retained.
```

zenador: Good catch, thanks!

tacole02 (approved):
Docs look good!

PR #11001, opened 2025-03-26
MQE: avoid using statement in `Query` where possible

charleskorn:
#### What this PR does

This PR refactors the `Query` type to avoid using the `EvalStmt` where possible.

This will make it easier to introduce queries based on query plans, as query plans may not have an equivalent PromQL statement.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/query.go
+++ b/pkg/streamingpromql/query.go
@@ -885,7 +885,3 @@ func (q *Query) Cancel() {
 func (q *Query) String() string {
 	return q.qs
 }
-
-func timeMilliseconds(t time.Time) int64 {
-	return t.UnixNano() / int64(time.Millisecond/time.Nanosecond)
-}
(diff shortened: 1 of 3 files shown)


pkg/streamingpromql/query.go
```diff
@@ -37,6 +37,7 @@ import (
 var errQueryCancelled = cancellation.NewErrorf("query execution cancelled")
 var errQueryClosed = cancellation.NewErrorf("Query.Close() called")
 var errQueryFinished = cancellation.NewErrorf("query execution finished")
+var errStringForRangeQuery = errors.New("query expression produces a string, but expression for range queries must produce an instant vector or scalar")
```
jhesketh, line 40: (nit) could make this generic to use in newQuery too
```suggestion
var strStringForRangeQuery = "query expression produces a %s but expression for range queries must produce an instant vector or scalar"
```

jhesketh (approved):
lgtm, just a nit

pkg/streamingpromql/query.go
charleskorn: I've tackled this in a slightly different way in f178f97: I've changed `Exec` to not check if it's an instant query if we get a `StringOperator`. This makes the behaviour consistent with how we handle `RangeVectorOperator`s, which are also not supported as the root operator for instant queries.

jhesketh (approved):

PR #11004, opened 2025-03-26
MQE: initial query planner implementation

charleskorn:
#### What this PR does

This PR introduces a query planner for MQE. At present, it takes the original PromQL expression, translates it to a query plan, then converts that to operators for evaluation. No significant optimisations (eg. common subexpression elimination) are applied to the query plan as part of this PR.

The main parts of this PR are:

* the `QueryPlanner` type in `pkg/streamingpromql/planning.go`: the main entrypoint to the query planning functionality, responsible for creating query plans from PromQL expressions
* the `QueryPlan` type in `pkg/streamingpromql/planning/plan.go`: represents a query plan
* the `Node` type in `pkg/streamingpromql/planning/plan.go`: represents a node in a query plan, and provides methods to examine, manipulate and convert the node to an operator
* the `EncodedQueryPlan` and `EncodedNode` types in `pkg/streamingpromql/planning/plan.proto`: represent a query plan transformed into a form suitable for serialization / deserialization
* two simple AST optimisation passes in `pkg/streamingpromql/optimize/ast`

I've run our existing benchmarks with query planning enabled and disabled. For non-trivial cases, the cost of running planning is noise compared to the cost of evaluating the query itself.

There are a number of improvements I'd like to introduce as follow-up PRs:

* metrics for planning latency
* investigate using code generation for methods like `EquivalentTo()`, `Details()`, `Children()`, `SetChildren()` and `ChildrenLabels()`
* moving the responsibility for converting a node to an operator elsewhere: the current design will make passing configuration (eg. cache connection parameters) difficult

I'll also introduce the first optimisation that builds on this as a follow-up PR.

I'll add a changelog entry when this PR is nearly ready to merge.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/planning.go
```diff
@@ -0,0 +1,663 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
jhesketh, line 1: (tests for this file would be good, but fine for followup)

pkg/streamingpromql/optimize/ast/collapse_constants_test.go
```diff
@@ -0,0 +1,64 @@
...
+		"bar * foo":  "bar * foo",
+		"2 > bool 3": "2 > bool 3",
+
+		"metric":        "metric",
+		"(metric)":      "(metric)",
+		"-metric":       "-metric",
+		"metric[1m]":    "metric[1m]",
+		"metric[1m:1m]": "metric[1m:1m]",
+		`"abc"`:         `"abc"`,
+	}
```
jhesketh, line 44: (nit) would be good to have some cases where constants are a small part of a larger query with lots of irrelevant parts.

pkg/mimir/modules.go
```diff
@@ -856,6 +859,23 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
 	}), nil
 }
 
+func (t *Mimir) initQueryPlanner() (services.Service, error) {
+	if t.Cfg.Querier.EngineConfig.MimirQueryEngine.UseQueryPlanning {
+		_, mqeOpts := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, nil, nil, nil)
+		t.QueryPlanner = streamingpromql.NewQueryPlanner(mqeOpts)
+
+		t.QueryPlanner.RegisterASTOptimizationPass(&ast.SortLabelsAndMatchers{}) // This is a prerequisite for other optimization passes such as common subexpression elimination.
```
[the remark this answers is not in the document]
jhesketh: Additionally I wonder if SortLabelsAndMatchers needs to be handled as a special case and always applied through some other mechanism rather than registering it with the planner.

Unless we create some kind of DAG for optimisations, it is a prerequisite that many optimisations will need so we should make sure it is done or implement a dependency system.

pkg/streamingpromql/planning/core/function_call.go
```diff
@@ -0,0 +1,130 @@
...
+	if ok {
+		o, err := scalarFactory(children, params.MemoryConsumptionTracker, params.Annotations, f.ExpressionPosition.ToPrometheusType(), timeRange)
+		if err != nil {
+			return nil, err
+		}
+
+		return planning.NewSingleUseOperatorFactory(o), nil
+	}
+
+	switch f.FunctionName {
```
jhesketh, line 100: (nit) a comment why these are handled specially

pkg/streamingpromql/engine_test.go
```diff
@@ -79,118 +79,149 @@ func TestUnsupportedPromQLFeaturesWithFeatureToggles(t *testing.T) {
...
 
-	_, err = engine.NewInstantQuery(context.Background(), nil, nil, expression, time.Now())
-	require.NoError(t, err)
+	testWithAndWithoutQueryPlanner(t, opts, func(t *testing.T, engine *Engine) {
+		_, err := engine.NewInstantQuery(context.Background(), nil, nil, expression, time.Now())
+		require.NoError(t, err)
+	})
+}
+
+func testWithAndWithoutQueryPlanner(t *testing.T, opts EngineOpts, test func(t *testing.T, engine *Engine)) {
```
jhesketh, line 146: (this is awesome for coverage :-))

pkg/streamingpromql/engine_test.go
```diff
@@ -2921,7 +2987,7 @@ func runMixedMetricsTests(t *testing.T, expressions []string, pointsPerSeries in
 	// - Look backs
 
 	opts := NewTestEngineOpts()
-	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), log.NewNopLogger())
+	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), nil, log.NewNopLogger())
```
jhesketh, line 2990: I'd be curious to run the gauntlet through w/wout the planner. Maybe not permanently, but just as an early validation?

jhesketh (approved):
This is very impressive, awesome work!

It's a lot to review at once, so I'll avoid too many nit-picks. Most of my comments can be ignored or followed up with later.

In general on reviewing the approach it matches the design doc and is, in my opinion, a good approach. As a whole this serves as a great place to start and I'm good to merge it as-is and work from there :-).  (Up to you if you want to address the comments here or later etc)

> investigate using code generation for methods like EquivalentTo(), Details(), Children(), SetChildren() and ChildrenLabels()

I wonder if generics would be an option for this duplication? The overhead may be higher, but is likely negligible in the scheme of executing a query and would reduce code maintainership burden

charleskorn:
Given the approval, I'm going to merge this so we can start testing it and come back and address the comments separately.

pkg/streamingpromql/planning.go
charleskorn: Could you elaborate? Most of the functionality here should be covered by the tests in `planning_test.go` and the existing tests for the engine.

pkg/streamingpromql/optimize/ast/collapse_constants_test.go
charleskorn: Could you give an example of the kind of thing you had in mind? There are already a couple of cases like this earlier in this list (eg. `sum(vector(2 + 3))`)

pkg/streamingpromql/planning/core/core.proto
```diff
@@ -0,0 +1,153 @@
...
+message VectorSelectorDetails {
+  repeated LabelMatcher matchers = 1;
+  Timestamp timestamp = 2; // FIXME: the timestamp needs to be nullable, but optional isn't supported in gogoproto and gogoproto.nullable = true does nothing on int64 fields, so an inner message seems like the least-worst option
+  google.protobuf.Duration offset = 3 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
+  PositionRange expressionPosition = 4 [(gogoproto.nullable) = false];
+}
+
+message MatrixSelectorDetails {
+  repeated LabelMatcher matchers = 1;
+  Timestamp timestamp = 2; // FIXME: the timestamp needs to be nullable, but optional isn't supported in gogoproto and gogoproto.nullable = true does nothing on int64 fields, so an inner message seems like the least-worst option
```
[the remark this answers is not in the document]
charleskorn: It doesn't make a huge difference, but it makes sense to use that type given we're using the duration type below: https://github.com/grafana/mimir/pull/11124/commits/d32265a84ff1a7af50893122662697ed93ed1cfa.

pkg/streamingpromql/engine_test.go
charleskorn: Done: https://github.com/grafana/mimir/pull/11124/commits/7e81b8f1b96ff89ed0b094659a30589e478403d9

pkg/streamingpromql/planning/core/function_call.go
charleskorn: Done: https://github.com/grafana/mimir/pull/11124/commits/b25909f22de0a52207ba21bb56f9a7cfa20966eb

pkg/mimir/modules.go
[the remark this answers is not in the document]
jhesketh: > > I don't know about registering optimisations here. I think it should go elsewhere, but this is likely something we can return to later.
> 
> Could you elaborate on this?

I'm not sure of the advantage of registering each optimisation as part of the modules.go. In my mind we should keep modules.go as lean as possible in terms of just registering the modules and not doing anything specific to the module itself.

For example, why not have `t.QueryPlanner.InitOptimizations()` or similar that sets up all of the defined optimizations in there?

The Register hooks still exist so GEM and other modules can register more optimizations if needed, but for the ones that are all applicable to Mimir/MQE, and what I expect to be the vast majority, we list them as part of the planner packages.

Hope that makes sense.

pkg/streamingpromql/planning.go
jhesketh: sorry that comment should have been deleted. It was an oversight in reviewing the files in order.

pkg/streamingpromql/optimize/ast/collapse_constants_test.go
jhesketh: Just some needlessly complex query like `label_join(...) + (sum(vector(2+3)) / (sum(count) + sum(min(series))) * 600 + 3000`, to check the `2+3` gets collapsed and nothing else is unexpectedly changed.

Also, not sure if it's already covered, but that example could be interesting where we want to test the order of operations. ie we don't want the last part to become `* 3600`.

charleskorn: Done: https://github.com/grafana/mimir/pull/11124/commits/ce15a445676542222d8482f6525f4f06a83a6a3a

pkg/mimir/modules.go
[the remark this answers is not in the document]
charleskorn: My thought was that all of the "generic" (not specific to Mimir) optimization passes would live in the `streamingpromql` package, while optimization passes that are specific to Mimir would live in their corresponding component.

This is particularly important for query sharding, splitting etc. that I imagine will live somewhere in `pkg/frontend/querymiddleware` so that they can reuse the existing logic there without creating a circular dependency. (If `pkg/streamingpromql` depends on the frontend package, then we'll create a circular dependency.)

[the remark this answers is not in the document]
jhesketh: Okay, I can see that rationale.

However, ignoring the circular dependencies (eg we move or copy what is needed from `querymiddleware`), would it make more sense to have all optimisiations inside the planning package?

I'm not opposed to it, and just thinking out loud.

We also don't have to solve this now nor be beholden to it, so we can return to it later.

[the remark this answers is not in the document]
charleskorn: >would it make more sense to have all optimisiations inside the planning package?

Not sure, let's see how things play out as we implement them.

PR #11006, opened 2025-03-26
Improve MimirIngesterFailsToProcessRecordsFromKafka to not fire during forced TSDB head compaction

pracucci:
#### What this PR does

The `MimirIngesterFailsToProcessRecordsFromKafka` alert currently fires each time there a forced TSDB head compaction in progress, because samples older than the new "head min time" will fail to be ingested (will be retried over and over, until the head compaction completes).

I would like to have the alert not firing during forced head compaction, given it's not actionable, but we miss a metric to know when there's a forced head compaction in progress. In this PR I'm adding a metric and using it in the alert. The alert query has been changed so that it keep working even if the new metric has not been deployed it.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir-mixin/alerts/ingest-storage.libsonnet
+++ b/operations/mimir-mixin/alerts/ingest-storage.libsonnet
@@ -123,12 +123,18 @@
           alert: $.alertName('IngesterFailsToProcessRecordsFromKafka'),
           'for': '5m',
           expr: |||
-            sum by (%(alert_aggregation_labels)s, %(per_instance_label)s) (
-                # This is the old metric name. We're keeping support for backward compatibility.
-              rate(cortex_ingest_storage_reader_records_failed_total{cause="server"}[1m])
-              or
-              rate(cortex_ingest_storage_reader_requests_failed_total{cause="server"}[1m])
-            ) > 0
+            (
+              sum by (%(alert_aggregation_labels)s, %(per_instance_label)s) (
+                  # This is the old metric name. We're keeping support for backward compatibility.
+                rate(cortex_ingest_storage_reader_records_failed_total{cause="server"}[1m])
+                or
+                rate(cortex_ingest_storage_reader_requests_failed_total{cause="server"}[1m])
+              ) > 0
+            )
+
+            # Tolerate failures during the forced TSDB head compaction, because samples older than the
+            # new "head min time" will fail to be appended while the forced compaction is running.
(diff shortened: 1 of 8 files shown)


operations/mimir-mixin/alerts/ingest-storage.libsonnet
```diff
@@ -123,12 +123,18 @@
...
+                  # This is the old metric name. We're keeping support for backward compatibility.
+                rate(cortex_ingest_storage_reader_records_failed_total{cause="server"}[1m])
+                or
+                rate(cortex_ingest_storage_reader_requests_failed_total{cause="server"}[1m])
+              ) > 0
+            )
+
+            # Tolerate failures during the forced TSDB head compaction, because samples older than the
+            # new "head min time" will fail to be appended while the forced compaction is running.
+            unless (max by (%(alert_aggregation_labels)s, %(per_instance_label)s) (cortex_ingester_tsdb_forced_compactions_in_progress) > 0)
```
colega, line 137:
```suggestion
            unless (max by (%(alert_aggregation_labels)s, %(per_instance_label)s) (max_over_time(cortex_ingester_tsdb_forced_compactions_in_progress[1m]) > 0)
```

Let's say the scrape interval is 15s, then:
- 1st scrape at 00s: `cortex_ingester_tsdb_forced_compactions_in_progress=1`
- 2nd scrape at 15s: `cortex_ingest_storage_reader_records_failed_total++`
- 3rd scrape at 45s: `cortex_ingester_tsdb_forced_compactions_in_progress=0`
- Somewhere between 45s and 1m14s999ms we evaluate the rule, and lookback says that `cortex_ingester_tsdb_forced_compactions_in_progress=0` so it _fires_ 

(Okay, it shouldn't because it should be `for:5m`, but it fires today, I guess it's because it force-compacts different tenants, anyway, this would be more correct IMO).

colega (approved):
LGTM but please take a look at my comment about using `max_over_time`.

operations/mimir-mixin/alerts/ingest-storage.libsonnet
pracucci: The `for: 5m` should just avoid the scenario you're describing 🤔

pracucci: But also doing as you suggest doesn't hurt

pracucci: Addressed in https://github.com/grafana/mimir/pull/11006/commits/42b7430b0494b6eabf6a37d421a14c8b599dcf4f

Issue #10987, opened 2025-03-25
Flaky TestMergeGrafanaState in alertmanager


narqo:
I've bumped into, what looked like a deadlock, somewhere in the alertmanager's `TestMergeGrafanaState` (introduced in https://github.com/grafana/mimir/pull/10959).

That is, [my PR](https://github.com/grafana/mimir/pull/10986), which doesn't touch any Go code, timed out on the CI ([full logs][1]):

<details>
<summary>Goroutine stacktrace</summary>

```
panic: test timed out after 30m0s
	running tests:
		TestMergeGrafanaState (29m5s)

goroutine 46640 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:2373 +0x265
created by time.goFunc
	/usr/local/go/src/time/sleep.go:215 +0x45

goroutine 1 [chan receive, 29 minutes]:
testing.(*T).Run(0xc00056cd00, {0x26f416b, 0x15}, 0x2890b38)
	/usr/local/go/src/testing/testing.go:1751 +0x851
testing.runTests.func1(0xc00056cd00)
	/usr/local/go/src/testing/testing.go:2168 +0x86
testing.tRunner(0xc00056cd00, 0xc00080fae0)
	/usr/local/go/src/testing/testing.go:1690 +0x227
testing.runTests(0xc0008f22b8, {0x3b98200, 0x41, 0x41}, {0x7f5c8f834108?, 0x40?, 0x3baa4a0?})
	/usr/local/go/src/testing/testing.go:2166 +0x8bf
testing.(*M).Run(0xc0008ea8c0)
	/usr/local/go/src/testing/testing.go:2034 +0xf18
main.main()
	_testmain.go:173 +0x165

goroutine 5 [select]:
github.com/golang/glog.(*fileSink).flushDaemon(0x3baa538)
	/__w/mimir/mimir/vendor/github.com/golang/glog/glog_file.go:380 +0x17c
created by github.com/golang/glog.init.1 in goroutine 1
	/__w/mimir/mimir/vendor/github.com/golang/glog/glog_file.go:188 +0x2c7

goroutine 6 [select, 30 minutes]:
github.com/dgraph-io/ristretto.(*defaultPolicy).processItems(0xc0005c9580)
	/__w/mimir/mimir/vendor/github.com/dgraph-io/ristretto/policy.go:102 +0x106
created by github.com/dgraph-io/ristretto.newDefaultPolicy in goroutine 1
[...]

PR #11009, opened 2025-03-26
Alertmanager: Fix flaky TestMergeGrafanaState

santihernandezc:
If the context cancellation is not caught by `state.running()`, the test can time out waiting to receive a message on `msgc`. This PR changes the test to check for state replication metrics instead of waiting for a message.

Fixes https://github.com/grafana/mimir/issues/10987


stevesg (approved):
Makes sense to me

PR #11010, opened 2025-03-26
Add support for HTTP cluster validation

duricanikolic:
#### What this PR does
This PR adds support for HTTP cluster validation in the query-frontend.
On client side, cluster validation label propagation through HTTP request headers can be enabled by setting `-query-frontend.client-cluster-validation.label` configuration option.

On server side, cluster validation can be enabled through the following configuration options (inherited from https://github.com/grafana/dskit/pull/671):
- `-server.cluster-validation.label`, that defines the server's cluster validation label. When the validation is enabled, this value will be compared with the cluster validation label received through the requests.
- `-server.cluster-validation.http.enabled`, used for enabling the cluster validation.
- `-server.cluster-validation.http.soft-validation`, that can be enabled only together with the cluster validation, and that implies a soft validation.
- `-server.cluster-validation.http.exclude-paths`, a comma-separated list of url paths that are excluded from the cluster validation check.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/frontend/frontend_test.go
```diff
@@ -99,7 +101,139 @@ func TestFrontend_RequestHostHeaderWhenDownstreamURLIsConfigured(t *testing.T) {
...
+			softValidation:     true,
+			expectedStatusCode: 200,
+		},
+		"if client has no cluster validation label and cluster validation is enabled an error is returned": {
+			clientCluster:      "",
+			serverCluster:      "server-cluster",
+			enabled:            true,
+			softValidation:     false,
+			expectedStatusCode: 511,
+			expectedError:      "rejected request with empty cluster validation label - it should be \\\"server-cluster\\\"",
```
aknuds1, line 170: [Nit] Could you write this as a raw string literal instead, so we can avoid so many backslashes for readability?

pkg/frontend/frontend_test.go
```diff
@@ -99,7 +101,139 @@ func TestFrontend_RequestHostHeaderWhenDownstreamURLIsConfigured(t *testing.T) {
...
+				handler = middleware.ClusterValidationMiddleware(testCase.serverCluster, []string{}, testCase.softValidation, logger).Wrap(handler)
+			}
+			downstreamServer := http.Server{
+				Handler:      handler,
+				ReadTimeout:  10 * time.Second,
+				WriteTimeout: 10 * time.Second,
+			}
+
+			defer downstreamServer.Shutdown(context.Background()) //nolint:errcheck
+			go downstreamServer.Serve(downstreamListen)           //nolint:errcheck
```
aknuds1, line 195: To make sure that goroutines don't hang around after tests finish (never fun to debug, if it should become a problem), please use a `WaitGroup`:
```suggestion
			var wg sync.WaitGroup
			wg.Add(1)
			go func() {
				defer wg.Done()
				downstreamServer.Serve(downstreamListen)           //nolint:errcheck
			}()
			t.Cleanup(wg.Wait)
```

aknuds1 (commented):
Did a first pass at reviewing. Continuing tomorrow.

pkg/frontend/frontend_test.go
duricanikolic: I have fixed this

colega (approved):

PR #11012, opened 2025-03-26
Add -ingest-storage.kafka.fetch-max-wait configuration option

pracucci:
#### What this PR does

I would like to experiment with an higher value of Fetch "max wait". In this PR I propose to add `-ingest-storage.kafka.fetch-max-wait` configuration option to configure the maximum amount of time a Kafka broker waits for some records before a Fetch response is returned.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/config.go
```diff
@@ -220,6 +223,10 @@ func (cfg *KafkaConfig) Validate() error {
 		return ErrInvalidMaxConsumerLagAtStartup
 	}
 
+	if cfg.FetchMaxWait < 5*time.Second || cfg.FetchMaxWait > 30*time.Second {
+		return ErrInvalidFetchMaxWait
+	}
```
narqo, line 228: For a non-Warpstream broker — i.e. real Kafka, that most users will use — does it still make sense to clamp the wait time to 5s?

narqo (approved):
🔥 Works for me.

The small question about clamping the value to 5s isn't blocking, really.

pkg/storage/ingest/config.go
pracucci: I think in practice lower values don't really matter, because all our Fetch requests have min bytes set to 1. Kafka will reply nearly immediately anyway when there are some records to fetch.

PR #11025, opened 2025-03-27
Remove maximum value of `limit` parameter in `/cardinality/label_names` and `/cardinality/label_values` HTTP endpoints

### LINKED ISSUE #10328: Idea: Make it possible to analyse all values of a label with the `/api/v1/cardinality/label_values` endpoint

We are currently working to reduce the amount of active metric series of the observability platform at Maersk. As part of this effort, we need a way to retrieve all metric names and, ideally, rank them by series count.

Knowing the series count of each metric is important as it allows us to do things like:
1. Focusing on the highest cardinality metrics first.
2. Having a lower limit of cardinality below which we ignore the metric (i.e., don't take action on metrics with less than 5000 series).
3. Building intelligence around how many series we're saving with various measures; answering questions like "how many series would we save by dropping this metric?" or "how many series have we saved in total with the measures implemented thus far?".

The [`GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_values`](https://grafana.com/docs/mimir/v2.14.x/references/http-api/#label-values-cardinality) endpoint is a prime candidate for our use-case. The issue is that the endpoint has a maximum limit of `500`. In other words, with this endpoint, we can only get the top 500 highest cardinality metrics of the platform.

This is an example of the request we're currently making:
```sh
curl --location --globoff 'localhost:8080/prometheus/api/v1/cardinality/label_values?label_names[]=__name__&count_method=active&limit=500' --header 'X-Scope-OrgID: fake'
```

Initially, we suspected that the `limit` parameter was in place to protect server-side resources. However, we inspected the code of the endpoint, and it is our understanding that the endpoint always fetches _all_ label values, and that the only thing the `limit` parameter does is control how many are _returned_ ([source](https://github.com/grafana/mimir/blob/cb24dfec7a56636e8bb99c2801e480c01170cc53/pkg/querier/cardinality_analysis_handler.go#L276)).

We'd like for the `limit` parameter to have no maximum value:
* Behaviour for all requests that don't specify the `limit` parameter is the same: `limit` will default to `20`.
* Behaviour for all requests that specify a `limit` smaller than `500` is the same.
* Behaviour for all requests that specify a `limit` greater than `500` changes: they would previously fail with `'limit' param cannot be greater than '500'` but will succeed after the change.
* Clients who would like to analyse a larger set of label values can set `limit` to a greater value like `100000`.

We've considered two alternative solutions:
1. The [label values endpoint](https://prometheus.io/docs/prometheus/3.0/querying/api/#querying-label-values) can return all metric names: `GET /prometheus/api/v1/label/__name__/values`. The issue with this endpoint is that it doesn't return series count.
2. The [instant query endpoint](https://prometheus.io/docs/prometheus/3.0/querying/api/#instant-queries) is theoretically capable of returning all metric names and count of series for each name with this query: `count by (__name__) ({ __name__=~".+"})`. This query is prohibitively expensive to run, and fails almost immediately with a `the query exceeded the maximum number of chunks` error.

We might consider whether to make the same change for the `limit` parameter of the [`GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_names`](https://grafana.com/docs/mimir/v2.14.x/references/http-api/#label-names-cardinality) endpoint.

My soft opinion might be that the two endpoints should have the same implementation of the `limit` parameter, but I am not privy to the details of this endpoint.

Small (<= 1 month dev)

We'd need to update the documentation for the label cardinality endpoint: https://grafana.com/docs/mimir/v2.14.x/references/http-api/#label-values-cardinality

### Proposer?

Lasse Hels

LasseHels:
#### What this PR does

This pull request removes the maximum value of the `limit` parameter in these endpoints:
* `GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_names`.
* `GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_values`.

Previously, the `limit` parameter had a maximum value of `500`. 

With these changes, the `limit` parameter no longer has an explicit maximum value, making the de facto maximum of the parameter the upper bound of the range imposed by `strconv.Atoi()`.

Although the `limit` parameter _technically_ still has a maximum value, I've chosen not to include this in the documentation. There are two reasons for this:
1. Almost all clients are not going to care about a maximum `limit` value of several billion.
2. If a client _does_ execute a request like `?limit=9999999999999999999999999999`, then they will get back a `400 Bad Request` response with a clear error message:
```
strconv.Atoi: parsing "9999999999999999999999999999": value out of range
```

Other minor changes:
* I've added some test cases for the `label_values` endpoint that seemed to be missing.
* I've added `t.Parallel()` to nearby tests where it seemed safe.

Fixes #10328

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


flxbk:
Thanks! This is looking good to me overall, however I'm wondering about your motivation for leaving the maximum limit in place for the `label_names` endpoint. Removing it on both would make things more consistent and I can see no reason that would justify keeping it on one but not the other. Is there's something I've missed?

LasseHels:
@flxbk Good question. I should have elaborated on this decision in the pull request description.

The reason is that the two endpoints handle response allocations differently.

`GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_values` does not use `limit` to allocate slices; it always allocates based on the count of returned items, and _then_ it slices the allocation based on `limit`:
```go
func limitLabelValuesCardinality(labelValuesCardinality []api.LabelValuesCardinality, limit int) []api.LabelValuesCardinality {
	if len(labelValuesCardinality) <= limit {
		return labelValuesCardinality
	}
	return labelValuesCardinality[:limit]
}
```

In other words, the `limit` parameter will never impact the amount of memory allocation that the `label_values` endpoint performs.

`GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_names` works differently: it allocates memory based on the minimum of `len(labelsWithValues)` and `limit`:
```go
items := make([]*api.LabelNamesCardinalityItem, min(len(labelsWithValues), limit))
```

Regardless of the length of `labelsWithValues`, the length of the `items` slice has a hard limit of `500`. If we remove the maximum value of `limit`, then this is no longer the case.

Consider an example where a client makes a request that causes `len(labelsWithValues)` to be one million:
* **Currently:** `items` will allocate with at most a length of `500`.
* **Removing maximum value of `limit`:** `items` can now allocate a length of up to one million if we set `limit=1000000`.

The way `label_names` allocates is more economical than `label_values`, but the fact that `label_values` always greedily allocates means that we are less concerned with removing the maximum value of `limit`.

What do you think? Am I being overly cautious?

flxbk:
Thanks for explaining your reasoning!

> Regardless of the length of `labelsWithValues`, the length of the `items` slice has a hard limit of `500`. If we remove the maximum value of `limit`, then this is no longer the case.

Given that allocations for `labelsWithValues` are not limited by the `limit` parameter, limiting the output slice by that value seems like a band-aid that's not worth introducing more inconsistency between the endpoints for in my opinion. If response size proves to be problematic for either of these endpoints, I think we'd be better off actually bounding the size of the response, in which case allocations for the output slice would also be bounded respectively. What do you think?

LasseHels:
@flxbk Roger that. I will update the `GET,POST <prometheus-http-prefix>/api/v1/cardinality/label_names` endpoint to also not have a maximum value for `limit`.

LasseHels:
@flxbk Pull request has been updated. One of the `ci / integration` checks is failing. I took a look, and I think it's because of a flaky test? I don't have permissions to re-try the workflow.

This is the error message:
```
    integration_memberlist_single_binary_test.go:240: 
        	Error Trace:	/home/runner/work/mimir/mimir/integration/integration_memberlist_single_binary_test.go:240
        	Error:      	Condition never satisfied
        	Test:       	TestSingleBinaryWithMemberlistScaling
        	Messages:   	expected all instances to have 15.000000 ring members and 68.000000 tombstones
```
([source](https://github.com/grafana/mimir/actions/runs/14196945433/job/39774492789?pr=11025#step:15:867))

pracucci (commented):
Jumping on this a bit late, sorry. I don't think it's wise to completely remove the max limit. I think we should make it configurable on a per-tenant basis (via our limits in runtime config).

LasseHels:
@narqo @pracucci Thanks for chiming in. Let me close this pull request and author a new solution proposal on [the issue](https://github.com/grafana/mimir/issues/10328). This way, we can maintain the record in a single place rather than burying it in this pull request.

zhuoyuan-liu:
Hi @narqo @pracucci , I noticed that the limit is only used to format the HTTP  [response](https://github.com/grafana/mimir/blob/0acc18f440cca04b2d6049984a4f2c6abaed7502/pkg/querier/cardinality_analysis_handler.go#L81). Its primary role appears to be controlling the size of the output, not the computation itself.

The label value pairs are already computed earlier via the [LabelNamesAndValues(ctx, cardinalityRequest.Matchers, cardinalityRequest.CountMethod)](https://github.com/grafana/mimir/blob/0acc18f440cca04b2d6049984a4f2c6abaed7502/pkg/querier/cardinality_analysis_handler.go#L44)` or `[link](https://github.com/grafana/mimir/blob/0acc18f440cca04b2d6049984a4f2c6abaed7502/pkg/querier/cardinality_analysis_handler.go#L75) function. Additionally, there are already two per-tenant  [config](https://github.com/grafana/mimir/blob/0acc18f440cca04b2d6049984a4f2c6abaed7502/pkg/util/validation/limits.go#L205) in place that govern the actual computation limits.

Since this particular limit only affects the structure of the HTTP response — and not the underlying label pair computation — applying a per-tenant configuration here doesn't seem necessary or meaningful. 

What do you think — does that align with your understanding, or is there another use case I might have missed?

PR #11026, opened 2025-03-27
fix: make grpc max send and receive size configurable

### LINKED ISSUE #10295: Bug: Hardcoded GRPC limits in store gateway client prevents data from quering
### What is the bug?

When trying to query data with a large time range, one can face the GRPC error in the mimir-quirier. 
"rpc error: code = ResourceExhausted desc = grpc: received message larger than max (133997632 vs. 104857600)"

Increasing grpc_server_max_recv_msg_size and grpc_server_max_send_msg_size does nothing because it is hardcoded in the 
https://github.com/grafana/mimir/blob/main/pkg/querier/store_gateway_client.go#L80

### How to reproduce it?

It fails only on a limited set of queries, and it is not easy to reproduce without having the same metrics data stored in storage. 

The data size reached hardcoded limits. 

Kubernetes, mimir-distributed helm chart, GCS bucket. 

Image

joobisb:
#### What this PR does

GRPC limits were hard coded in store_gateway_client. This PR makes it configurable by taking the values from the server config and keeping the current hard-code values as default.

Fixes #10295

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimir/modules.go
+++ b/pkg/mimir/modules.go
@@ -647,7 +647,7 @@ func (t *Mimir) initQuerier() (serv services.Service, err error) {
 
 func (t *Mimir) initStoreQueryable() (services.Service, error) {
 	q, err := querier.NewBlocksStoreQueryableFromConfig(
-		t.Cfg.Querier, t.Cfg.StoreGateway, t.Cfg.BlocksStorage, t.Overrides, util_log.Logger, t.Registerer,
+		t.Cfg.Querier, t.Cfg.StoreGateway, t.Cfg.BlocksStorage, t.Cfg.Server, t.Overrides, util_log.Logger, t.Registerer,
 	)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


joobisb:
I did not see any unit tests related to config in `store_gateway_client_test.go`. Please do let me what's the best way to add a unit test here.

joobisb:
@narqo could you please have a look at this PR?

dimitarvdimitrov:
was this covered by #11074 by any chance?

joobisb:
> #11074

I think this might be. @charleskorn can you please confirm?

dimitarvdimitrov:
Charles is away for the week. I think it was covered in his PR. I'll close this PR and we can reopen if we missed something here

charleskorn:
> > #11074
> 
> I think this might be. @charleskorn can you please confirm?

Yep, this was covered by #11074.

PR #11041, opened 2025-03-28
fix: Use dnssrvnoa to connect to memcached instances

AndreZiviani:
#### What this PR does
Fix the default memcached address to use `dnssrvnoa` instead of `dns` to avoid conflicts/rebalancing on the hashing algorithm when a memcached pod is restarted

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

This is the default in [Loki](https://github.com/grafana/loki/blob/b442d27fa6a0f9819d679c960ed5f396cbec199e/production/helm/loki/values.yaml#L210)

Thanks Gergely Madarász for raising this up on [Slack](https://grafana.slack.com/archives/C039863E8P7/p1742550271473749)

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -29,6 +29,8 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## main / unreleased
 
+* BUGFIX Memcached: Use `dnssrvnoa+` address prefix instead of `dns+` which results in DNS `SRV` record lookups instead of `A` or `AAAA`. This results in fewer cache evictions when the members of the Memcached cluster change. #11041
+
 ## 5.7.0-rc.0
 
 * [CHANGE] KEDA Autoscaling: Changed toPromQLLabelSelector from object to list of strings, adding support for all PromQL operators. #10945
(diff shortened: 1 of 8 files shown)


z0rc:
When using `dnssrvnoa` discovery, `addresses` value must not include numbered port part `:11211`. Named port number discovery is part of the process now.

AndreZiviani:
> When using `dnssrvnoa` discovery, `addresses` value must not include numbered port part `:11211`. Named port number discovery is part of the process now.

thanks, fixed

56quarters (changes requested):
Please don't change the hostname and port as part of this change, the `dnssrvnoa+` prefix change will work without it.

AndreZiviani:
@56quarters do you mean use like this?
```
dnssrvnoa+{{ template "mimir.fullname" . }}-index-cache.{{ .Release.Namespace }}.svc.{{ .Values.global.clusterDomain }}
```

56quarters:
> @56quarters do you mean use like this?
> 
> ```
> dnssrvnoa+{{ template "mimir.fullname" . }}-index-cache.{{ .Release.Namespace }}.svc.{{ .Values.global.clusterDomain }}
> ```

Roughly, I mean only change `dns+` to `dnssrvnoa+` instead of removing the port number and adding `_memcached-client._tcp`

AndreZiviani:
@56quarters done

Issue #10686, opened 2025-02-19
Race condition in TestPartitionReader_ConsumeAtStartup


charleskorn:
Failed in [this CI run](https://github.com/grafana/mimir/actions/runs/13404725165/job/37442517090#step:8:92):

```
==================
WARNING: DATA RACE
Read at 0x00c004f1f3c3 by goroutine 199722:
  testing.(*common).logDepth()
      /usr/local/go/src/testing/testing.go:1024 +0x504
  testing.(*common).log()
      /usr/local/go/src/testing/testing.go:1011 +0x7d
  testing.(*common).Log()
      /usr/local/go/src/testing/testing.go:1052 +0x55
  testing.(*T).Log()
      <autogenerated>:1 +0x4f
  github.com/grafana/mimir/pkg/util/test.(*TestingLogger).Log()
      /__w/mimir/mimir/pkg/util/test/logger.go:38 +0x177
  github.com/go-kit/log.(*context).Log()
      /__w/mimir/mimir/vendor/github.com/go-kit/log/log.go:168 +0x4ba
  github.com/grafana/mimir/pkg/storage/ingest.(*KafkaLogger).Log()
      /__w/mimir/mimir/pkg/storage/ingest/logger.go:34 +0x6a9
  github.com/twmb/franz-go/pkg/kgo.(*wrappedLogger).Log()
      /__w/mimir/mimir/vendor/github.com/twmb/franz-go/pkg/kgo/logger.go:123 +0xc6
  github.com/twmb/franz-go/pkg/kgo.(*consumer).assignPartitions()
      /__w/mimir/mimir/vendor/github.com/twmb/franz-go/pkg/kgo/consumer.go:976 +0x2ec
  github.com/twmb/franz-go/pkg/kgo.(*consumer).doOnMetadataUpdate.func1()
      /__w/mimir/mimir/vendor/github.com/twmb/franz-go/pkg/kgo/consumer.go:1208 +0x173
  github.com/twmb/franz-go/pkg/kgo.(*consumer).doOnMetadataUpdate.func2()
      /__w/mimir/mimir/vendor/github.com/twmb/franz-go/pkg/kgo/consumer.go:1220 +0x3a

Previous write at 0x00c004f1f3c3 by goroutine 184514:
  testing.tRunner.func1()
      /usr/local/go/src/testing/testing.go:1677 +0x8fa
  runtime.deferreturn()
      /usr/local/go/src/runtime/panic.go:605 +0x5d
[...]

PR #11065, opened 2025-03-31
Maybe fix TestPartitionReader_ConsumeAtStartup race condition

pracucci:
#### What this PR does

This PR is a follow up of #9428. Assuming the direction in that PR is the right one, then we need to use the synchronized and shared logger in `newKafkaProduceClient()` (which is used by `TestPartitionReader_ConsumeAtStartup`) and other tests. This change may fix the race condition described in https://github.com/grafana/mimir/issues/10686 (I haven't been able to reproduce it, so I'm not sure it's enough to fix it).

Maybe #10686

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥 Works for me.

The original stack-trace from #10686 feels like a bug in franz-go to me (similar to what you described in the comments there). But I'm happy if these changes resolves it for mimir.

Issue #10917, opened 2025-03-17
Release Mimir 2.16


56quarters:
### Publish the release candidate

- [x] Begin drafting the [release notes](https://github.com/grafana/mimir/blob/main/RELEASE.md#write-release-notes-document)
  - Create the release notes PR targeting the main branch
  - This step shouldn't block from publishing release candidate
  - After the release notes PR is merged (which usually happen after RC is published), cherry-pick them into the release branch
  - [PR](https://github.com/grafana/mimir/pull/10921)
- [x] Wait for any open PR we want to get merged before cutting the release candidate
  - We shouldn't wait for the open PRs beyond the scheduled release date
- [x] Update `CHANGELOG.md` against `main`
  - [x] Run `./tools/release/check-changelog.sh LAST-RELEASE-TAG...main` and add missing PRs to CHANGELOG
  - [x] Ensure CHANGELOG entries are [sorted by type](https://github.com/grafana/mimir/blob/main/docs/internal/contributing/README.md#changelog)
  - [x] Add a new section for the new release so that `## main / unreleased` is blank and at the top. The new section should say `## x.y.0-rc.0`.
  - [x] [PR](https://github.com/grafana/mimir/pull/10928)
- [x] Run `./tools/release/notify-changelog-cut.sh CHANGELOG.md`
- [x] Run `make mixin-screenshots` https://github.com/grafana/mimir/pull/10930
  - Before opening the PR, review all updated screenshots and ensure no sensitive data is disclosed
- [x] Create new release branch https://github.com/grafana/mimir/pull/10931
  - [x] Create the branch
    ```bash
    git checkout r<xxx> # xxx is the latest weekly release
    git checkout -b release-<version>
    git push -u origin release-<version>
    ```
  - [x] Remove "main / unreleased" section from the CHANGELOG
  - [x] Backport screenshot updates to release branch
  - [x] Backport any fixes applied to the selected weekly release branch to the release branch https://github.com/grafana/mimir/pull/10932
  - [x] If a new minor or major version is being released, adjust the settings in the `renovate.json5` configuration on the `main` branch by adding the new version.
        This way we ensure that dependency updates maintain the new version, as well as the latest two minor versions.
        For instance, if versions 3.0 and 2.10 are configured in `renovate.json`, and version 3.1 is being released,
        during the release process `renovate.json5` should keep updated the following branches: `main`, `release-3.1`, `release-3.0` and `release-2.10`. https://github.com/grafana/mimir/pull/10934
[...]

PR #11071, opened 2025-03-31
chore: Remove release-2.14 from automated updates

56quarters:
Mimir 2.16 is releases so 2.14 no longer gets automatic updates.

Part of https://github.com/grafana/mimir/issues/10917


narqo (approved):
🔥

PR #11083, opened 2025-04-02
Don't log unavailable statistics in ruler query stats messages when using remote query-frontend

charleskorn:
#### What this PR does

When the ruler is using a remote query-frontend, query stats such as the number of series fetched will not be available.

Prior to this PR, this would cause the `query stats` log line emitted by rulers to include these statistics with zeroes, for example:

```
query_wall_time_seconds=0 fetched_series_count=0 fetched_chunk_bytes=0 fetched_chunks_count=0 sharded_queries=0
```

This PR modifies the ruler to not log these fields at all when a remote query-frontend is in use.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/ruler/compat.go
+++ b/pkg/ruler/compat.go
@@ -255,7 +255,7 @@ func MetricsQueryFunc(qf rules.QueryFunc, userID string, queries, failedQueries 
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Thanks!

PR #11084, opened 2025-04-02
Update PromQL test syntax to use new style

charleskorn:
#### What this PR does

https://github.com/prometheus/prometheus/pull/15995 introduced a new syntax for annotation assertions in PromQL test files.

This PR updates our test cases to use the new syntax, and updates the formatting script to enforce that the old syntax is not used.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


jhesketh (approved):
we should add the messages, but that's laboursome and we're otherwise comparing them

charleskorn:
> we should add the messages, but that's laboursome and we're otherwise comparing them

Agree - this is no worse than what we had before.

PR #11086, opened 2025-04-02
Fix -query-frontend.query-sharding-max-sharded-queries enforcement for instant queries with binary operators

pracucci:
#### What this PR does

Instant queries don't set `Hints.TotalQueries` because they're not split by time (only range queries do). This means that we never enforce `-query-frontend.query-sharding-max-sharded-queries` when the query has 1+ binary operators. This PR fixes it, assuming that if `TotalQueries` is unknown then it's just 1.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥

PR #11089, opened 2025-04-02
[chore]: `validation.NewOverides` doesn't return an error

colega:
#### What this PR does

This method always returns a nil error, so all the error handling is unnecessary.

I've been looking for a moment to do this for a while now, I guess it's now time, because its usage just becomes wider and wider.

None.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


flxbk (approved):
LGTM

PR #11090, opened 2025-04-02
Cost attribution: remove `max_cost_attribution_labels_per_user` limit

colega:
#### What this PR does

There were two limits for cost attribution labels:
- One configurable per tenant which we added because at some point tenants would self-serve this configuration (I insisted), but it doesn't really make sense to have it _here_, because it's too late in the pipeline. If there's a component rendering the runtime self-served configurations, it should enforce the labels amount there. Checking it here just makes the overrides loading fail.
- The other one, hard limit of 4 labels, was enforced by some technical detail we had in the first implementation (somewhere we had an array and we had to have an upper bound on it?), but it doesn't exist anymore, so we can remove it.

Follow up on https://github.com/grafana/mimir/pull/10269

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


narqo (approved):
Works for me 🔥

pracucci (approved):

PR #11096, opened 2025-04-02
block-builder: Fetch records and process them concurrently

codesome:
#### What this PR does

TODO: still testing this out

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sync"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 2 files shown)


codesome:
I tried to write a benchmark, but since it runs locally (meaning fetching are basically instant), and that I had to add a timeout when partition ends in this concurrent case, we cannot compare before and after times properly. So I will test it in dev and report any improvements.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -542,6 +543,39 @@ func (b *BlockBuilder) consumePartitionSection(
...
+			select {
+			case <-done:
+				return
+			case <-ctx.Done():
+				return
+			default:
+				// Put a timeout on PollFetches to avoid indefinite blocking when we reach the
+				// end of partitions.
+				pollCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+				fetches := b.kafkaClient.PollFetches(pollCtx)
```
dimitarvdimitrov, line 565: `PollFetches` doesn't actually do any work. It waits until the client receives records from kafka. And the client only has one inflight Fetch request per partition. The client waits for that Fetch request to finish before firing off the next one. It does them sequentially because it uses the offset of the last record in the first response as the offset in the second request. That's the reason we implemented our own concurrentFetchers

codesome:
Superseded by https://github.com/grafana/mimir/pull/11185

PR #11097, opened 2025-04-03
Add query limiter middleware for instant queries

chencs:
#### What this PR does

This PR introduces a middleware (modeled after the query blocker and error caching middlewares) which checks if a query should not be run more than a configured `AllowedFrequency`. If the last time the query was allowed less than `AllowedFrequency` time ago, it is blocked with a `query-limited` error, and returns a 429 status. It works with the range query, instant query, and remote read paths, though this PR only adds it to the instant query middlewares, since frequent, expensive instant queries can easily overwhelm the system due to the lack of results caching on that path.

Limitations:
* only matches exact queries
* does not check query time (e.g., range start/end, or instant time), since the intent is to block excessive unruly/expensive instant queries that are automatically executed at "random" times.

~I'm not 100% convinced that 429 is the appropriate status code to return, but it _is_ a retryable error.~

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


dimitarvdimitrov:
This runs once in each query-frontend. So in effect the rate limit will scale as the query-frontend deployment scales. Any thoughts on doing this in the query-scheduler where we have fewer replicas? Or perhaps using memcached for this?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 * [ENHANCEMENT] Ingester: Remove cost-attribution experimental `max_cost_attribution_labels_per_user` limit. #11090
 * [ENHANCEMENT] Update Go to 1.24.2. #11114
 * [ENHANCEMENT] Ingester/Distributor: Add `cortex_cost_attribution_*` metrics to observe the state of the cost-attribution trackers. #11112
+* [ENHANCEMENT] Add configurable per-tenant `limited_queries`, which may only be executed at or less than a certain frequency. #11097
```
tacole02, line 33:
```suggestion
* [ENHANCEMENT] Add configurable per-tenant `limited_queries`, which you can only run at or less than a certain frequency. #11097
```

chencs:
@56quarters no need to apologize! These are good suggestions, thanks for the feedback. I'll work on a change next week.

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+	return false
+}
+
+func (ql *queryLimiterMiddleware) enoughTimeSinceLastQuery(ctx context.Context, cacheValue, hashedKey string, limitedQuery *validation.LimitedQuery) bool {
+	// If the query fails to be added to the cache with ErrNotStored, that means the last time the query was called
+	// is within AllowedFrequency, so we should block it. Otherwise, the query has now been added to the cache and
+	// will be blocked next time if the entry still exists in the cache.
+	err := ql.cache.Add(ctx, hashedKey, []byte(cacheValue), limitedQuery.AllowedFrequency)
```
dimitarvdimitrov, line 97: memcached can't enforce TTLs of less than a second https://github.com/grafana/mimir/blob/ec3000eacf862d9b773117dc47a41caae98e4101/vendor/github.com/grafana/dskit/cache/memcached_client.go#L378

I think we should validate the user config somehow and error out that somehow.


This way we're also not allowing for any bursts. For example 1 qps means up to 1 request every second. A more relaxed definition is up to 60 requests in a minute. The `Inc` approach would allow that.

I think the current approach works better for when the allowed rate is much lower than 1 QPS. Something like 1 query per minute or one per 5 minutes.

On the other hand, with the `Inc` approach we can't go less frequent than once per minute (or whatever time period we allow bursts in).

Do you have a use case in mind? How frequent are the queries we're trying to prevent?

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+	if errors.Is(err, cache.ErrNotStored) {
+		return false
+	} else {
+		ql.logger.Log("msg", "error while adding to query limiter cache", "err", err)
```
dimitarvdimitrov, line 101: `else` is also when there is no error. We probably don't want to log then

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -392,6 +398,7 @@ func newQueryMiddlewares(
 		newLimitsMiddleware(limits, log),
 		newSplitInstantQueryByIntervalMiddleware(limits, log, engine, registerer),
 		queryBlockerMiddleware,
+		queryLimiterMiddleware,
```
dimitarvdimitrov, line 401: why only on instant queries? isn't it a trivial change to support it on remote read and range queries?

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2655,7 +2655,20 @@ How it **works**:
 
 How to **fix** it:
 
-This error only occurs when an administrator has explicitly define a blocked list for a given tenant. After assessing whether or not the reason for blocking one or multiple queries you can update the tenant's limits and remove the pattern.
+This error only occurs when an administrator has explicitly defined a blocked list for a given tenant. After assessing the reason for blocking one or multiple queries you can update the tenant's limits and remove the pattern.
+
+### err-mimir-query-limited
+
+This error occurs when a query-frontend blocks a read request because the query matches at least one of the rules defined in the limits.
```
dimitarvdimitrov, line 2662: "...and the query is being ran too frequently."

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2655,7 +2655,20 @@ How it **works**:
...
+
+How it **works**:
+
+- The query-frontend implements a middleware responsible for assessing whether the query should be limited, and whether it has been run within the last allowed frequency.
+- To configure the limit, set the block `limited_queries` in the `limits`.
+
+How to **fix** it:
+
+This error only occurs when an administrator has explicitly defined a limited queries list for a given tenant. After assessing the reason for limiting one or multiple queries, you can update the tenant's limits and remove the pattern.
```
dimitarvdimitrov, line 2671: i'd also mention to just consider running the query less frequently

dimitarvdimitrov (commented):
> I'm not 100% convinced that 429 is the appropriate status code to return, but it is a retryable error.

429 sounds reasonable to me

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (ql *queryLimiterMiddleware) Do(ctx context.Context, req MetricsQueryRequest) (Response, error) {
+	spanLog := spanlogger.FromContext(ctx, ql.logger)
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return ql.next.Do(ctx, req)
+	}
+
+	key := ql.keyGen.QueryRequestLimiter(ctx, tenant.JoinTenantIDs(tenantIDs), req)
+	hashedKey := cacheHashKey(key)
```
56quarters, line 60: The hash being done here is not a cryptographic hash so collisions are possible. We work around that elsewhere in the query-frontend by also storing the entire key as part of the cache payload. Since we're not doing that here, I think we need to use a cryptographic hash for this key. There are a few examples of using the `blake2b` hash in Mimir:

* [CachingBucket](https://github.com/grafana/mimir/blob/1eef2a6e0e67a5f767630b5eec08dea173f14dc1/pkg/storage/tsdb/bucketcache/caching_bucket.go#L799-L809)
* [IndexCache](https://github.com/grafana/mimir/blob/1eef2a6e0e67a5f767630b5eec08dea173f14dc1/pkg/storegateway/indexcache/remote.go#L141-L187)

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+
+	for _, tenantID := range tenantIDs {
```
56quarters, line 62: The cache key is built including all tenant IDs joined together but we're checking the rate limit against the cache for each tenant.

Would it make sense to:
* Collect all rules for all tenants
* See if any of them matches the query for this request
* Of those, pick the smallest `AllowedFrequency`
* Try to do a single `.Add()` or `.Incr()` against the cache with the key that includes all tenant IDs and the query

pkg/frontend/querymiddleware/query_limiter.go
chencs: The queries we're trying to prevent _should_ be running a little less than once every 5 minutes. The highest frequency I've seen in the wild is up to 3x/min.

pkg/frontend/querymiddleware/roundtrip.go
chencs: It is trivial, although we might need to add some information to the model, e.g., what if someone only wants to limit a query if it's coming through as an instant query, but not as a range query? I prioritized instant queries here because they seemed most prone to system abuse, since results are not cached.

pkg/frontend/querymiddleware/query_limiter.go
chencs: We could also build the key using only one tenant ID, right? e.g., move key generation into the `isBlocked` block.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 * [ENHANCEMENT] Update Go to 1.24.2. #11114
 * [ENHANCEMENT] Ingester/Distributor: Add `cortex_cost_attribution_*` metrics to observe the state of the cost-attribution trackers. #11112
 * [ENHANCEMENT] gRPC/HTTP servers: Add `cortex_server_invalid_cluster_validation_label_requests_total` metric, that is increased for every request with an invalid cluster validation label. #11241
+* [ENHANCEMENT] Add configurable per-tenant `limited_queries`, which you can only run at or less than an allowed frequency. #11097
```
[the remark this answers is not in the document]
chencs: We don't currently document this for any of the per-tenant config blocks under [`limits`](https://grafana.com/docs/mimir/latest/configure/configuration-parameters/#limits), so I think we'd need more input from docs (i.e., @tacole02) about how to represent this, or how much support makes sense given these are all experimental.

Here's an example `runtime.yaml` when running the local microservices deployment (using the example query in the tests):
```
overrides:
  "anonymous":
    limited_queries:
      - query: "rate(metric_counter[5m])"
        allowed_frequency: "1m"
```

pkg/frontend/querymiddleware/query_limiter.go
chencs: Oh, never mind, I'm understand what you mean and my idea doesn't work.

CHANGELOG.md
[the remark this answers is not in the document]
tacole02: I agree, I think we should add them both to https://grafana.com/docs/mimir/latest/configure/configuration-parameters/#limits and mark that they are experimental. If either of you want to draft up descriptions for these, I'm happy to take a look.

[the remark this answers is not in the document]
chencs: I can add an example block for limited queries in this PR, but I'd like to scope it to just that and I'm happy to add example blocks for blocked requests and queries in a separate PR.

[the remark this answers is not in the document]
chencs: > I can add an example block for limited queries in this PR

This is in.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3676,6 +3676,15 @@ The `limits` block configures default and per-tenant limits imposed by component
 # (experimental) List of queries to block.
 [blocked_queries: <blocked_queries_config...> | default = ]
 
+# (experimental) List of queries to limit and duration to limit them for.
+# Example:
+#   The following configuration limits the query "rate(metric_counter[5m])" to
+#   being run, at most, every 1m.
```
tacole02, line 3682:
```suggestion
#  running, at most, every minute.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2655,7 +2655,20 @@ How it **works**:
 
 How to **fix** it:
 
-This error only occurs when an administrator has explicitly define a blocked list for a given tenant. After assessing whether or not the reason for blocking one or multiple queries you can update the tenant's limits and remove the pattern.
+This error only occurs when an administrator has explicitly defined a blocked list for a given tenant. After assessing the reason for blocking one or multiple queries you can update the tenant's limits and remove the pattern.
```
tacole02, line 2658:
```suggestion
This error only occurs when an administrator has explicitly defined a blocked list for a given tenant. After assessing the reason for blocking one or multiple queries, you can update the tenant's limits and remove the pattern.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2655,7 +2655,20 @@ How it **works**:
...
+
+### err-mimir-query-limited
+
+This error occurs when a query-frontend blocks a read request because the query matches at least one of the rules defined in the limits, and the query is being run too frequently.
```
tacole02, line 2662:
```suggestion
This error occurs when a query-frontend blocks a read request because the query matches at least one of the rules defined in the limits and the query is being run too frequently.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2655,7 +2655,20 @@ How it **works**:
...
+
+How it **works**:
+
+- The query-frontend implements a middleware responsible for assessing whether the query should be limited, and whether it has been run within the last allowed frequency.
```
tacole02, line 2666:
```suggestion
- The query-frontend implements a middleware responsible for assessing whether the query should be limited and whether it has been run within the last allowed frequency.
```

tacole02 (approved):
Looks really good. Thanks, @chencs !

pkg/frontend/querymiddleware/roundtrip.go
dimitarvdimitrov: > edge cases, e.g., what if someone only wants to limit queries if they're coming through as an instant query, but not as a range query?

isn't the other case way more common? If an instant query is so expensive we don't allow running it more than once a minute, why would a the same query as a range query be ok to run? Conceptually a range query just computes the instant query over many steps along, so it's much more expensive. I'd be happy to treat them as the same cache key.

The existing query blocking middleware doesn't differentiate between instant and range queries.

Today users can _potentially_ circumvent the limit by running a range query with a single step. "Potentially" is doing a lot of the work in that sentence, because it's usually tools running those queries, and tools are harder to change.

IMO we should limit at least range query as well, without differentiating between the two. You can also do that in a follow-up PR if you think it'll add too much to the scope of this one.

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (ql *queryLimiterMiddleware) Do(ctx context.Context, req MetricsQueryRequest) (Response, error) {
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return ql.next.Do(ctx, req)
+	}
+
+	key := ql.keyGen.QueryRequestLimiter(ctx, tenant.JoinTenantIDs(tenantIDs), req)
+	hashedKey := maybeHashCacheKey(key)
+	// start at max duration value
```
56quarters, line 61: This seems to contradict the comment below?

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,106 @@
...
+	cacheValue := validation.LimitedQuery{
```
56quarters, line 62: nit: calling this `cacheValue` made me think we were still storing something in the cache. You also don't need to specify the fields here since `""` and `0` are the default values.

pkg/frontend/querymiddleware/query_limiter_test.go
```diff
@@ -0,0 +1,376 @@
...
+			name: "non-empty limit should block",
+			limits: mockLimits{limitedQueries: []*validation.LimitedQuery{
+				{Query: query, AllowedFrequency: time.Minute},
+			}},
+			expectSecondQueryBlocked: true,
+		},
+		{
+			name: "short max frequency should not block",
+			limits: mockLimits{limitedQueries: []*validation.LimitedQuery{
+				{Query: query, AllowedFrequency: time.Nanosecond},
```
56quarters, line 48: We can't actually use nanosecond TTLs in practice. Since we can advance "now" for the mock cache by any amount, might as well use a TTL that we could use in practice.

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -194,6 +194,8 @@ type CacheKeyGenerator interface {
 	// QueryRequestError should generate a cache key based on errors for the tenant ID and MetricsQueryRequest.
 	QueryRequestError(ctx context.Context, tenantID string, r MetricsQueryRequest) string
```
56quarters, line 196: Can you add docs for this?

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,106 @@
...
+		Query:            "",
+		AllowedFrequency: 0, // start with min duration
+	}
+	query := strings.TrimSpace(req.GetQuery())
+	var tenantMinAllowedFrequency string
+
+	for _, tenantID := range tenantIDs {
+		logger := log.With(ql.logger, "user", tenantID)
```
56quarters, line 70: If you're including the `tenant` field below, no need to include `user` here.

pkg/frontend/querymiddleware/query_limiter.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	// If we found any matching limited query, we should try to cache it
+	if cacheValue.Query != "" {
+		if err := ql.cache.Add(ctx, hashedKey, []byte(key), cacheValue.AllowedFrequency); err != nil {
+			// If we receive ErrNotStored, the entry is still in the cache and the query should be blocked
+			if errors.Is(err, cache.ErrNotStored) {
+				ql.blockedQueriesCounter.WithLabelValues(tenantMinAllowedFrequency, "limited").Inc()
+				return nil, newQueryLimitedError(cacheValue.AllowedFrequency, tenantMinAllowedFrequency)
+			}
+			ql.logger.Log("msg", "error while adding to query limiter cache", "err", err)
```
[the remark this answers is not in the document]
chencs: Ah, because I didn't realize that the `logger` passed in could in all likelihood be a `nopLogger`, and I didn't have any tests hitting this line. Good catch.

pkg/frontend/querymiddleware/query_limiter.go
chencs: Missed this comment, thank you!

pkg/frontend/querymiddleware/roundtrip.go
chencs: > If an instant query is so expensive we don't allow running it more than once a minute, why would a the same query as a range query be ok to run?

If a user is running step-aligned range queries with caching, I imagine that query could end up being less expensive to run frequently than the same instant query, no?

> Today users can potentially circumvent the limit by running a range query with a single step. "Potentially" is doing a lot of the work in that sentence, because it's usually tools running those queries, and tools are harder to change.

++


I'm not strongly opinionated against this being consistent across range and instant read paths, I was just trying to limit the blast radius by default. I can add it to range queries in this PR.

dimitarvdimitrov (approved):
THANK YOU! LGTM

PR #11100, opened 2025-04-03
Add RW2 support

krajorama:
#### What this PR does

Optimized version of #10432 .
Instead of unmarshal into RW2 then convert to RW1, unmarshal directly into RW1.

#### Which issue(s) this PR fixes or relates to

Related to #9072

#### Leftovers for next PRs

- [ ] check if `-timeseries-unmarshal-caching-optimization-enabled` is possible for RW2
- [ ] NHCB should be allowed #11143 
- [ ] drop empty labels during unmarshal, see test case in integration/distributor.go "drop empty labels"
- [ ] implement created timestamp support (RW2)
- [ ] implement created timestamp support (OTEL)

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11101, opened 2025-04-03
feat(rw2): prepare for RW2 support

krajorama:
Add the Remote Write 2.0 proto definition to mimirpb. The implementation of Remote Write 2.0  will follow but it needs to patch the generated files, so we commit the proto now to have actually readable diff in the PR that has the functionality.

See  #11100 diff.


aknuds1 (approved):
LGTM, sans Copilot comment.

PR #11103, opened 2025-04-03
block-builder: implement Is on errFirstRecordInFuture

dimitarvdimitrov:
without this we were getting misleading error logs here

https://github.com/grafana/mimir/blob/961ad26da672ba89af27cd0d38190be9f807e73f/pkg/blockbuilder/blockbuilder.go#L520-L523

### DIFF
--- a/pkg/blockbuilder/blockbuilder_test.go
+++ b/pkg/blockbuilder/blockbuilder_test.go
@@ -39,6 +39,9 @@ func TestErrFirstRecordInFuture(t *testing.T) {
 	var err error
 	err = &errFirstRecordInFuture{recordTs: time.Now()}
 	assert.True(t, errors.Is(err, &errFirstRecordInFuture{}))
+
+	firstRecErr := &errFirstRecordInFuture{}
+	assert.True(t, errors.As(err, &firstRecErr))
 }
 
 func TestBlockBuilder_WipeOutDataDirOnStart(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder_test.go
```diff
@@ -35,6 +35,15 @@ import (
 	"github.com/grafana/mimir/pkg/util/validation"
 )
 
+func TestErrFirstRecordInFuture(t *testing.T) {
+	var err error
+	err = &errFirstRecordInFuture{recordTs: time.Now()}
```
dimitarvdimitrov, line 40:
```suggestion
	var err error = &errFirstRecordInFuture{recordTs: time.Now()}
```

seizethedave:
@dimitarvdimitrov  Huh, I would expect errors.Is() to already handle such a simple check. Can you give me the toddler-friendly explanation of why it doesn't?

dimitarvdimitrov:
> @dimitarvdimitrov Huh, I would expect errors.Is() to already handle such a simple check. Can you give me the toddler-friendly explanation of why it doesn't?

from the docs of `errors.Is`:
```go
// An error is considered to match a target if it is equal to that target or if
// it implements a method Is(error) bool such that Is(target) returns true.
```

basically if the error doesn't implement neither `Unwrap` nor `Is` then we're back to direct `==` comparison. With `==` we need exactly the same instance of the error, same type doesn't cut it.

https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/errors/wrap.go;l=55-57
```go
		if targetComparable && err == target {
			return true
		}
		// ...
		switch err.(type) {
		// ...
		default:
			return false
		}
```

seizethedave (approved):
Sweet!

PR #11112, opened 2025-04-04
Expose cost attribution operational insights

colega:
#### What this PR does

Adds metrics that observe the cardinality and the overflow state of each one of the cost-attribution trackers.

These metrics are exposed per user and tracker. The "overflown" metric is only exposed when tracker is actually overflown.

Fixes an internal need.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/costattribution/manager.go
+++ b/pkg/costattribution/manager.go
@@ -4,6 +4,7 @@ package costattribution
 
 import (
 	"context"
+	"fmt"
 	"maps"
 	"slices"
 	"sync"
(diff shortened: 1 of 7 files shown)


pkg/costattribution/manager.go
```diff
@@ -138,20 +172,79 @@ func (m *Manager) ActiveSeriesTracker(userID string) *ActiveSeriesTracker {
...
+			m.sampleTrackerCardinalityDesc,
+			prometheus.GaugeValue,
+			float64(cardinality),
+			tracker.userID,
+		)
+		if overflown {
+			out <- prometheus.MustNewConstMetric(
+				m.sampleTrackerOverflowDesc,
+				prometheus.GaugeValue,
+				1,
```
56quarters, line 193: Would be easier to understand this metric if it was 1 or 0 instead of 1 or missing?

56quarters (approved):
Nicely done.

pkg/costattribution/manager.go
colega: It is, but we'd also need to pay for that extra cardinality (one metric per user per ingester) that is never needed (we don't expect tenants to be in oveflow), so I thought we'd rather just pay the price at query time doing something like `max by (user) (overflown or 0*caradinality)`

PR #11120, opened 2025-04-06
querymiddleware: cache instant query errors

### LINKED ISSUE #11029: Idea: cache errors from instant queries

We currently cache non-transient errors per #9028. However this is only applied to range queries. This becomes an issue with badly written recording rules (e.g. rules that have huge range selectors): they'll be run every minute and hit limit errors every time, creating a huge amount of load for no reason.

We should extend the error caching middleware to support instant queries. This would prevent repeated evaluations from creating unnecessary work to some extent (we cache errors for 5m by default).

_No response_

_No response_

Small (<= 1 month dev)

_No response_

### Proposer?

_No response_

narqo:
#### What this PR does

The PR adds errors caching middleware for instant query. This is basically the amendment to the original #9028 ~~(I haven't tested it yet, but I think it should just work)~~.

**Update** I ran some manual testing in a dev environment, and it works as I'd expected:

<details>
<summary>Details</summary>

Non-cachable errors aren't cached:
Screenshot 2025-04-11 at 14 30 04

Cachable errors are stored in the cache on first occurrence:
Screenshot 2025-04-11 at 14 58 00

...and are retrived from the cache on subsequent requests
Screenshot 2025-04-11 at 14 59 04

</details>

Fixes #11029

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

56quarters (commented):
Thanks for adding tests but I don't think this will work as written. The way the cache key for errors is built includes `req.GetStart()` and `req.GetEnd()` in the key. For an instant query this is the timestamp the query is executed at which will be different for every instant query.

Additionally, the screenshots of testing you did indicate another bug. A context deadline exceeded error is a timeout and shouldn't be cached.

56quarters:
> Thanks for adding tests but I don't think this will work as written. The way the cache key for errors is built includes `req.GetStart()` and `req.GetEnd()` in the key. For an instant query this is the timestamp the query is executed at which will be different for every instant query.
> 
> Additionally, the screenshots of testing you did indicate another bug. A context deadline exceeded error is a timeout and shouldn't be cached.

The first issue can be fix by changes to `DefaultCacheKeyGenerator` which can detect instant queries by comparing `GetStart()` and `GetEnd()` and build the key without start/end/step parameters.

56quarters:
> Additionally, the screenshots of testing you did indicate another bug. A context deadline exceeded error is a timeout and shouldn't be cached.

Opened https://github.com/grafana/mimir/pull/11198 to fix this

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -234,7 +234,15 @@ func (g DefaultCacheKeyGenerator) QueryRequest(_ context.Context, tenantID strin
 }
 
 func (g DefaultCacheKeyGenerator) QueryRequestError(_ context.Context, tenantID string, r MetricsQueryRequest) string {
-	return fmt.Sprintf("EC:%s:%s:%d:%d:%d", tenantID, r.GetQuery(), r.GetStart(), r.GetEnd(), r.GetStep())
+	start := r.GetStart()
+	end := r.GetEnd()
+	if start == end && r.GetStep() == 0 {
```
56quarters, line 239: I know that we document that `start == end` for instant queries. Do we specify anywhere that `step == 0` for instant queries?

pkg/frontend/querymiddleware/error_caching_test.go
```diff
@@ -22,200 +22,263 @@ import (
...
+				inner.On("Do", mock.Anything, mock.Anything).Return(innerRes, nil)
+
+				ctx := user.InjectOrgID(context.Background(), "1234")
+				req := tc.newQueryRequest(Options{})
+
+				key := keyGen.QueryRequestError(ctx, "1234", req)
+				bytes := []byte{0x0, 0x0, 0x0, 0x0}
+
+				// NOTE: We rely on this mock cache being synchronous
+				c.SetAsync(cacheHashKey(key), bytes, time.Minute)
```
56quarters, line 162: The `Cache` interface has grown a synchronous `.Set()` method since this test was written.

pkg/frontend/querymiddleware/error_caching_test.go
```diff
@@ -22,200 +22,263 @@ import (
...
+	res, err := testHandler(ctx, inner, c, req1)
+	require.Error(t, err)
+	require.Nil(t, res)
+	require.Equal(t, 1, c.CountFetchCalls())
+	require.Equal(t, 1, c.CountStoreCalls())
+
+	// Second query with the updated timestamp.
+	req2 := newInstantQueryRequest(t, 200)
+	res, err = testHandler(ctx, inner, c, req2)
+	require.Error(t, err)
```
56quarters, line 280: Maybe assert the particular type of error here as well?

56quarters (approved):
LGTM with a few comments. Thanks!

PR #11132, opened 2025-04-07
MQE: introduce `ResultType` method for query planner `Node`

charleskorn:
#### What this PR does

This PR introduces a `ResultType` method for the query planner `Node` interface.

This will allow optimization passes to correctly handle nodes that produce different kinds of values. 

I'm raising this as a separate PR in the interests of keeping the upcoming common subexpression elimination PR smaller.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/planning/core/function_call.go
+++ b/pkg/streamingpromql/planning/core/function_call.go
@@ -131,6 +131,8 @@ func (f *FunctionCall) OperatorFactory(children []types.Operator, timeRange type
 }
 
 func (f *FunctionCall) ResultType() (parser.ValueType, error) {
+	// absent and absent_over_time are special cases that aren't present in InstantVectorFunctionOperatorFactories,
+	// so we have to handle them specifically.
 	if _, ok := functions.InstantVectorFunctionOperatorFactories[f.FunctionName]; ok || f.FunctionName == "absent" || f.FunctionName == "absent_over_time" {
 		return parser.ValueTypeVector, nil
 	}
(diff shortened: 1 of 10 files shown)


pkg/streamingpromql/planning/core/function_call.go
```diff
@@ -128,3 +129,15 @@ func (f *FunctionCall) OperatorFactory(children []types.Operator, timeRange type
 		return nil, compat.NewNotSupportedError(fmt.Sprintf("'%v' function", f.FunctionName))
 	}
 }
+
+func (f *FunctionCall) ResultType() (parser.ValueType, error) {
+	if _, ok := functions.InstantVectorFunctionOperatorFactories[f.FunctionName]; ok || f.FunctionName == "absent" || f.FunctionName == "absent_over_time" {
```
jhesketh, line 134: Bit of a nit, but I dislike we have to keep this special case in multiple places.

We should add a comment at least.

But I wonder if we can make this better. Maybe some kind of dummy entry in `functions.InstantVectorFunctionOperatorFactories` ?

charleskorn: I worry that a dummy entry in `functions.InstantVectorFunctionOperatorFactories` might cause other problems or confusion.

I'm not expecting to need further instances of this special case anywhere else, if that helps allay your concerns.

jhesketh: I think if it panics when used incorrectly it could work, but I'm nit picking so I'm okay with it as is, but would love a comment on the line please.

PR #11136, opened 2025-04-07
Upgrade google.golang.org/grpc to v1.71.1

aknuds1:
#### What this PR does

Upgrade google.golang.org/grpc to v1.71.1, and drop `replace` directive in go.mod (which was preventing upgrade from v1.65).

The upgrade beyond gRPC v1.65 is particularly difficult since v1.66 introduces re-use (through memory pooling) of gRPC's unmarshalling buffers, and we depend in many cases on re-use of these buffers for performance reasons.

To make our re-use of gRPC unmarshalling buffers work also post gRPC v1.66, we introduce a type `BufferHolder` that is embedded by protobuf message types that retain unmarshalling buffer references. A custom gRPC codec (`pkg/mimirpb.codecV2`) increases the reference count of the unmarshalling buffer for every `BufferHolder` object it unmarshals. The users of such messages are responsible for calling their `FreeBuffer` method, so buffers can be released back to gRPC.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/grpc.go
```diff
@@ -30,10 +32,16 @@ func CloseAndExhaust[T any](stream Stream[T]) error {
...
 	go func() {
 		for {
-			if _, err := stream.Recv(); err != nil {
+			msg, err := stream.Recv()
+			if err != nil {
 				close(done)
 				return
 			}
+
+			// If the message has a buffer reference, free it.
```
56quarters, line 41: nit: Can you add a comment here somewhere that explains that there's nothing to free if the stream returned an error? It took me a second to understand what was going on here.

56quarters (approved):
LGTM with a few suggestions

pkg/distributor/query.go
```diff
@@ -195,7 +201,18 @@ func mergeExemplarQueryResponses(results []*ingester_client.ExemplarQueryRespons
 
 	result := make([]mimirpb.TimeSeries, len(exemplarResults))
 	for i, k := range keys {
-		result[i] = exemplarResults[k]
+		ts := exemplarResults[k]
+		for i, l := range ts.Labels {
+			ts.Labels[i].Name = strings.Clone(l.Name)
```
charleskorn, line 206: Not for this PR, but for all the places where we do cloning of strings like this for `LabelAdapter` instances and then later convert to `labels.Labels`, we should consider just converting to `labels.Labels` instances here. This would save us cloning the strings and then making another copy when we convert to `labels.Labels`.

charleskorn (approved):
Querying changes look good to me modulo Nick's suggestions.

pkg/util/grpc.go
charleskorn: Does what I've added in d2f3a6c make this clearer?

pkg/distributor/query.go
```diff
@@ -239,92 +256,37 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
 		// If we iteratively build a single slice, we'll spend a lot of time copying elements as the slice grows beyond its capacity.
 		// So instead, we build the slice in one go once we know how many series we have.
 		var streamingSeriesBatches [][]labels.Labels
 		streamingSeriesCount := 0
 
 		for {
-			resp, err := stream.Recv()
+			var err error
+			var isEOS bool
+			streamingSeriesCount, streamingSeriesBatches, isEOS, err = receiveResponse(stream, streamingSeriesCount, streamingSeriesBatches, queryLimiter, &result)
```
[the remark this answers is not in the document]
charleskorn: What if we moved this method to be a method on `ingesterQueryResult`? That would at least resolve the "modifying an out parameter" thing.

We could also modify `receiveResponse` to return `labelsBatch` rather than append it to `streamingSeriesBatches`, and move the responsibility for updating `streamingSeriesCount` and `streamingSeriesBatches` to the caller of `receiveResponse`.

What do you think @aknuds1? I believe you made this change originally.

go.mod
```diff
@@ -115,6 +115,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.29.1 // indirect
 	github.com/bboreham/go-loser v0.0.0-20230920113527-fcc2c21820a3 // indirect
 	github.com/ebitengine/purego v0.8.1 // indirect
+	github.com/envoyproxy/go-control-plane/envoy v1.32.3 // indirect
```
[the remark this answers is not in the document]
aknuds1: I don't quite understand how the `exclude` of `github.com/envoyproxy/go-control-plane/envoy` is linked to gRPC. Could you explain?

I can see OTOH, that the `replace cloud.google.com/go/auth => cloud.google.com/go/auth v0.15.0` directive is connected to the gRPC pinning. Should I keep that and upgrade in another PR?

pkg/distributor/query.go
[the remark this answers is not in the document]
aknuds1: Yep, it was done by me. Considering your suggestion @charleskorn.

>  I'm guessing the new method is so that we can defer resp.FreeBuffer() ?

Mainly, yes.

pkg/distributor/query.go
```diff
@@ -195,7 +201,18 @@ func mergeExemplarQueryResponses(results []*ingester_client.ExemplarQueryRespons
...
 	for i, k := range keys {
-		result[i] = exemplarResults[k]
+		ts := exemplarResults[k]
+		for i, l := range ts.Labels {
+			ts.Labels[i].Name = strings.Clone(l.Name)
+			ts.Labels[i].Value = strings.Clone(l.Value)
+		}
+		for i, e := range ts.Exemplars {
+			for j, l := range e.Labels {
+				ts.Exemplars[i].Labels[j].Name = strings.Clone(l.Name)
```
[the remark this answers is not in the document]
aknuds1: @56quarters I opted for the alternate solution of adding a method `CloneUnsafe` to respectively `ingester/client.TimeSeriesChunk`, `storegateway/storepb.Series` and `mimirpb.TimeSeries`, as I found this approach more natural in that the knowledge of unsafe fields are moved to a type method. It has the additional benefit of making the code more readable I think. I included a test for each method.

By doing the above, I might have actually found an omission in `pkg/querier/blocks_store_queryable.go`: I only copied `Series` labels and not chunk data slices. I rectified the latter in `Series.CloneUnsafe`. Does that look right to you @56quarters @charleskorn?

pkg/distributor/query.go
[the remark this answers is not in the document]
aknuds1: @charleskorn PTAL at my revision of `receiveResponse`.

aknuds1:
@56quarters @charleskorn I applied your feedback and left some questions. I also realized I could probably optimize conversion from `LabelAdapter`s to `Label`s by re-using the `labels.ScratchBuilder` between iterations. I introduced a new function `mimirpb.FromLabelAdaptersOverwriteLabelsSafe` for the latter, PTAL :)

pkg/ingester/client/custom.go
```diff
@@ -65,3 +68,14 @@ func ChunkFromMeta(meta chunks.Meta) (Chunk, error) {
 func DefaultMetricsMetadataRequest() *MetricsMetadataRequest {
 	return &MetricsMetadataRequest{Limit: -1, LimitPerMetric: -1, Metric: ""}
 }
+
+// CloneUnsafe converts all of c's unsafe references to safe copies.
```
56quarters, line 72: Having "unsafe" in the name makes me think this is doing something with the `unsafe` module. Logic looks good though :+1:

pkg/ingester/client/custom.go
```diff
@@ -6,6 +6,9 @@
 package client
 
 import (
+	"slices"
+	strings "strings"
```
56quarters, line 10: ?

pkg/querier/blocks_store_queryable.go
```diff
@@ -958,12 +958,7 @@ func (q *blocksStoreQuerier) receiveMessage(c BlocksStoreClient, stream storegat
 
 	// Response may either contain series, streaming series, warning or hints.
 	if s := resp.GetSeries(); s != nil {
-		// Take a safe copy of every label.
-		for i, l := range s.Labels {
-			s.Labels[i].Name = strings.Clone(l.Name)
-			s.Labels[i].Value = strings.Clone(l.Value)
-		}
-		mySeries = append(mySeries, s)
+		mySeries = append(mySeries, s.CloneUnsafe())
```
56quarters, line 961: This doesn't seem correct, `CloneUnsafe()` doesn't return anything, right?

pkg/distributor/query.go
```diff
@@ -433,8 +433,12 @@ func (r *ingesterQueryResult) receiveResponse(stream ingester_client.Ingester_Qu
 		r.chunkseriesBatches = append(r.chunkseriesBatches, resp.Chunkseries)
 	} else if len(resp.StreamingSeries) > 0 {
 		labelsBatch := make([]labels.Labels, 0, len(resp.StreamingSeries))
+		var (
+			labelsBuilder labels.ScratchBuilder
+			l             labels.Labels
+		)
 		for _, s := range resp.StreamingSeries {
-			l := mimirpb.FromLabelAdaptersToLabelsWithCopy(s.Labels)
+			mimirpb.FromLabelAdaptersOverwriteLabelsSafe(&labelsBuilder, s.Labels, &l)
```
56quarters, line 441: It seems like this is going to override `l labels.Labels` on every iteration leading to corrupted results?

56quarters: I see that a copy is made to fix an issue in a later commit. I'm struggling to understand if this is safe or not. Maybe it's easier to return the newly created `label.Labels` if we're making copies anyway?

56quarters (approved):
The label handling code in `pkg/mimirpb` and the safety of it is worrying me but overall, the rest of this looks good to me.

pkg/ingester/client/custom.go
aknuds1: I suppose I should rename it to something less ominous sounding in that case. Would `CloneRefs` work?

pkg/ingester/client/custom.go
aknuds1: Oops, must be an automatic thing.

pkg/distributor/query.go
aknuds1: The variable `l labels.Labels` is copied when appending to `labelsBatch`:

```go
labelsBatch = append(labelsBatch, l)
```

> I see that a copy is made to fix an issue in a later commit

This fix is specific to `stringlabels` build mode, where I realized I had to copy the `ScratchBuilder`'s internal buffer.

pkg/querier/blocks_store_queryable.go
aknuds1: This `CloneUnsafe` method does return the `Series`. I can make the other `CloneUnsafe` implementations do the same, for consistency.

pkg/ingester/client/custom.go
aknuds1: Pushed a revision renaming to `CloneRefs`, let me know what you think.

pkg/querier/blocks_store_queryable.go
aknuds1: In the latest revision they all return the method receiver.

pkg/ingester/client/custom_test.go
```diff
@@ -0,0 +1,52 @@
...
+	c := TimeSeriesChunk{
+		Labels: []mimirpb.LabelAdapter{
+			{
+				Name:  yoloString(labelNameBytes),
+				Value: yoloString(labelValueBytes),
+			},
+		},
+		Chunks: []Chunk{
+			{
+				Data: mimirpb.UnsafeByteSlice(labelValueBytes),
```
charleskorn, line 30: This is a bit confusing to me - is there a reason to reuse `labelValueBytes` here? Should we use a different byte slice for the chunks?

pkg/storegateway/storepb/custom_test.go
```diff
@@ -0,0 +1,54 @@
...
+		Labels: []mimirpb.LabelAdapter{
+			{
+				Name:  yoloString(labelNameBytes),
+				Value: yoloString(labelValueBytes),
+			},
+		},
+		Chunks: []AggrChunk{
+			{
+				Raw: Chunk{
+					Data: mimirpb.UnsafeByteSlice(labelValueBytes),
```
charleskorn, line 31: Same as above, reusing the same byte slice here seems confusing to me.

pkg/distributor/query.go
```diff
@@ -428,6 +389,79 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
+		}
+		r.chunkseriesBatches = append(r.chunkseriesBatches, resp.Chunkseries)
+	} else if len(resp.StreamingSeries) > 0 {
+		labelsBatch := make([]labels.Labels, 0, len(resp.StreamingSeries))
+		var (
+			labelsBuilder labels.ScratchBuilder
+			l             labels.Labels
+		)
+		for _, s := range resp.StreamingSeries {
+			l = mimirpb.FromLabelAdaptersOverwriteLabelsSafe(&labelsBuilder, s.Labels, &l)
```
charleskorn, line 441: The reuse and overwriting here makes me nervous and is a little difficult to reason about. Given we were previously using `mimirpb.FromLabelAdaptersToLabels(s.Labels)` here (which does a copy of the label names and values with `stringlabels` anyway, and so is equivalent to `FromLabelAdaptersToLabelsWithCopy`), what if we just use `FromLabelAdaptersToLabelsWithCopy` here?

pkg/querier/blocks_store_queryable.go
```diff
@@ -986,7 +938,92 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
...
+
+	if ss := resp.GetStreamingSeries(); ss != nil {
+		myStreamingSeriesLabels = slices.Grow(myStreamingSeriesLabels, len(ss.Series))
+
+		var (
+			labelsBuilder labels.ScratchBuilder
+			ls            labels.Labels
+		)
+		for _, s := range ss.Series {
+			ls = mimirpb.FromLabelAdaptersOverwriteLabelsSafe(&labelsBuilder, s.Labels, &ls)
```
charleskorn, line 1011: Same as above - I think we can use `FromLabelAdaptersToLabelsWithCopy` here which should have the same performance characteristics as `FromLabelAdaptersToLabels` with `stringlabels`.

This removes the need for `FromLabelAdaptersOverwriteLabelsSafe` entirely.

pkg/mimirpb/compat_slice.go
```diff
@@ -28,9 +29,16 @@ func FromLabelAdaptersOverwriteLabels(_ *labels.ScratchBuilder, ls []LabelAdapte
 	*dest = FromLabelAdaptersToLabels(ls)
 }
 
+// FromLabelAdaptersOverwriteLabelsSafe casts []LabelAdapter to labels.Labels, but with cloned label names/values (to avoid unsafe references) and returns the result.
```
charleskorn, line 32: See below - I don't think we need this method.

pkg/storegateway/series_refs.go
```diff
@@ -1268,6 +1268,7 @@ func fetchCachedSeriesForPostings(ctx context.Context, userID string, indexCache
 		logSeriesForPostingsCacheEvent(ctx, logger, userID, blockID, shard, itemID, "msg", "can't decode series cache", "err", err)
 		return seriesChunkRefsSet{}, false
 	}
+	defer entry.FreeBuffer()
```
charleskorn, line 1271: I don't think we need this: `entry` is a Protobuf message that's come from a cache, not a gRPC response.

This also means the labels-related changes below are unnecessary.

pkg/storegateway/storepb/custom.go
```diff
@@ -6,6 +6,9 @@
 package storepb
 
 import (
+	"slices"
+	strings "strings"
```
charleskorn, line 10:
```suggestion
	"strings"
```

pkg/ingester/client/custom_test.go
aknuds1: I re-used because I didn't see any strict need to have a separate reference for the chunk, but I've pushed a revision with a dedicated one.

pkg/storegateway/storepb/custom_test.go
aknuds1: I re-used because I didn't see any strict need to have a separate reference for the chunk, but I've pushed a revision with a dedicated one.

pkg/storegateway/storepb/custom.go
aknuds1: Strange that the automation would make a prefix automatically :/ Thanks.

pkg/storegateway/series_refs.go
aknuds1: Thanks for noticing, removed it :)

pkg/distributor/query.go
aknuds1: As discussed offline, I can extract this potential optimization and raise it separately from this PR.

pkg/querier/blocks_store_queryable.go
aknuds1: Done.

pkg/distributor/query.go
aknuds1: Done.

charleskorn (approved):

bboreham (commented):
A brave piece of work.  I fear this is all getting a bit too complicated.

I have a few superficial comments.

pkg/distributor/distributor.go
```diff
@@ -2811,6 +2817,8 @@ respsLoop:
 		if err := queryLimiter.AddSeries(m); err != nil {
 			return nil, err
 		}
+		// Make safe copies of labels.
+		m.InternStrings(strings.Clone)
```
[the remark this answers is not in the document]
aknuds1: I was actually wondering. Can we assume that Mimir is always built with `-tags stringlabels`?

pkg/distributor/query.go
```diff
@@ -428,6 +389,75 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
 	return resp, nil
 }
 
+func (r *ingesterQueryResult) receiveResponse(stream ingester_client.Ingester_QueryStreamClient, queryLimiter *limiter.QueryLimiter) ([]labels.Labels, bool, error) {
```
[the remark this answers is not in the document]
aknuds1: I can add it.

pkg/mimirpb/custom.go
```diff
@@ -7,9 +7,166 @@ import (
...
+	protobufproto "google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/protoadapt"
 )
 
+func init() {
+	c := encoding.GetCodecV2(proto.Name)
+	encoding.RegisterCodecV2(&codecV2{codec: c})
+}
+
+// codecV2 customizes gRPC unmarshalling.
```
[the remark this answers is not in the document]
aknuds1: I'll elaborate.

pkg/querier/distributor_queryable.go
```diff
@@ -82,6 +83,9 @@ type distributorQuerier struct {
 	mint, maxt   int64
 	cfgProvider  distributorQueryableConfigProvider
 	queryMetrics *stats.QueryMetrics
+
+	streamReadersMtx sync.Mutex
```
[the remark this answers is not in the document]
aknuds1: @charleskorn would have to answer this one.

pkg/storegateway/storepb/custom.go
```diff
@@ -146,3 +149,15 @@ func (c AggrChunk) GetChunkEncoding() (chunk.Encoding, bool) {
 		return 0, false
 	}
 }
+
+// CloneRefs converts all of s' unsafe references to safe copies.
+func (s *Series) CloneRefs() *Series {
```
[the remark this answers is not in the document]
aknuds1: Returning the receiver is purely for convenience. Is it preferable if I name the method e.g. `TakeOwnership`?

pkg/ingester/client/custom.go
```diff
@@ -65,3 +68,16 @@ func ChunkFromMeta(meta chunks.Meta) (Chunk, error) {
 func DefaultMetricsMetadataRequest() *MetricsMetadataRequest {
 	return &MetricsMetadataRequest{Limit: -1, LimitPerMetric: -1, Metric: ""}
 }
+
+// CloneRefs converts all of c's unsafe references to safe copies.
+func (c *TimeSeriesChunk) CloneRefs() *TimeSeriesChunk {
```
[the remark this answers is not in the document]
aknuds1: I changed so it doesn't return anything.

pkg/mimirpb/timeseries.go
```diff
@@ -802,3 +802,19 @@ func preallocSliceIfNeeded[T any](size int) []T {
 	}
 	return nil
 }
+
+// CloneRefs converts all of ts' unsafe references to safe copies.
+func (ts *TimeSeries) CloneRefs() *TimeSeries {
```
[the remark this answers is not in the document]
aknuds1: I changed so it doesn't return anything.

pkg/distributor/distributor.go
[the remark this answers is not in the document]
bboreham: Well, _we_ always do, but there's nothing to stop someone else building it differently.

Perhaps: change `FromMetricsForLabelMatchersResponse` to call `FromLabelAdaptersToLabelsWithCopy`, and remove this `InternStrings` line.  `FromMetricsForLabelMatchersResponse` only seems to be used here.

Incidentally this whole loop seems misplaced, since the point is to count how many unique Labels are in the map we can just call `len`.  I think it makes more sense to do the limiting when `metrics` is built.  And then we could notice that we're computing a hash twice.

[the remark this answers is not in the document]
aknuds1: Thanks for the insight @bboreham - I changed `FromMetricsForLabelMatchersResponse` to call `FromLabelAdaptersToLabelsWithCopy` instead, and dropped `m. InternStrings(strings.Clone)`.

pkg/storegateway/storepb/custom.go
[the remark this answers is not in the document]
bboreham: We name several functions that get into this state as "Unsafe", so maybe this should be something like `MakeReferencesSafe` ?

`MakeReferencesSafeToRetain` ?

seizethedave (approved):
The ingestion-side changes look good to me. The proof will be in resource usage, I suppose.

`FreeBuffer()` is sprinkled so many times and in different formats that it's hard to spot-check, and a likely thing for future people to mess up. So I'm curious if you have thoughts on somehow automating/verifying that we're freeing things appropriately.

I see response resources being freed, and I see request resources being freed. Is that now our responsibility? I think with this direction we'll either need a readme on how to effectively interact with shared buffers from grpc-go, or a followup project where we measure the impact of removing the buffer sharing. Maybe you've already done that second one?

pkg/storegateway/storepb/custom.go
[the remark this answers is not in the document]
aknuds1: Renamed to `MakeReferencesSafeToRetain`.

pkg/mimirpb/custom.go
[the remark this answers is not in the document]
aknuds1: Done.

pkg/distributor/query.go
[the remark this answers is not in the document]
aknuds1: Done.

pkg/mimirpb/custom.go
```diff
@@ -7,9 +7,166 @@ import (
...
+// BufferHolder is a base type for protobuf messages that keep unsafe references to the unmarshalling buffer.
+type BufferHolder struct {
+	buffer mem.Buffer
+}
+
+func (m *BufferHolder) SetBuffer(buf mem.Buffer) {
+	m.buffer = buf
+}
+
+func (m *BufferHolder) FreeBuffer() {
```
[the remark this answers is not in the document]
aknuds1: > What happens if we miss one? They just get GC'd regularly later, I presume?

I think that should be what happens in most cases. I believe OTOH that there were cases where memory usage would blow up in queriers at least if buffers weren't freed. There should still be querier paths where we don't properly release buffers at the end, but memory usage looks OK. I think the pragmatic approach for now is to go with what we've got, and fix it if we were to hit the problem again of uncontrolled memory usage in the future.

> Is there a way to assert in tests or otherwise that we've correctly freed all buffers?

It is a real problem you are raising, and I'm not sure how to best solve it. Developing a linter for it wouldn't be feasible, right? I guess it would have to be solved through tests. I'd really prefer not to delay the merging of the PR any longer at this point though, as we risk bad merge conflicts the longer we delay it. Plus, we run the risk of missing security updates to gRPC the longer we hold off.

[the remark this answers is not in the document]
seizethedave: 🚢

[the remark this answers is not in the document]
charleskorn: >Is there a way to assert in tests or otherwise that we've correctly freed all buffers?

What if we wrapped the default pool with a custom one that tracks everything that has been taken and returned to the pool in tests? Then we'd be able to assert that everything was returned, and only returned once.

pkg/querier/distributor_queryable.go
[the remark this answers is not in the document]
charleskorn: This is a good catch, I had this confused with the `Queryable` that could be used from multiple places at once. I'll put together a PR to remove this.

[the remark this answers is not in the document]
charleskorn: #11376

pkg/mimirpb/custom.go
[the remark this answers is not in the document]
aknuds1: That sounds like a viable idea @charleskorn, nice :) Is that something you'd like to work on?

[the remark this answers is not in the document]
charleskorn: I don't have capacity for this at the moment unfortunately. If you'd like to give it a go, I'd be happy to review it.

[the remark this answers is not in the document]
aknuds1: Not sure if I have capacity either, but I will make a note of it at least.

pkg/mimirpb/custom.go
```diff
@@ -7,9 +7,170 @@ import (
...
+
+	return data, nil
+}
+
+// mem.DefaultBufferPool() only has five sizes: 256 B, 4 KB, 16 KB, 32 KB and 1 MB.
+// This means that for messages between 32 KB and 1 MB, we may over-allocate by up to 992 KB,
+// or ~97%. If we have a lot of messages in this range, we can waste a lot of memory.
+// So instead, we create our own buffer pool with more sizes to reduce this wasted space, and
+// also include pools for larger sizes up to 256 MB.
+var unmarshalSlicePool = mem.NewTieredBufferPool(unmarshalSlicePoolSizes()...)
```
bboreham, line 104: Why do we only do this on unmarshalling?  Doesn't the same argument apply to marshalling?

charleskorn: I don't remember why - it may be the impact of this wasn't large enough to see in profiles for marshalling. I agree that this could apply to marshalling as well.

PR #11139, opened 2025-04-07
Add conditional artificial delay

pracucci:
#### What this PR does

This PR is a follow up of https://github.com/grafana/mimir/pull/10107. I have more uses cases where I need to conditionally enable the artificial delay (for testing purposes) on a large number of tenants that may change over time. The two specific use cases are:

- Add latency to tenants with a numeric ID greater than X
- Add latency to tenants with configured max global series < Y

In this PR I'm adding more config options to support such use cases (hidden from the doc, given they're for testing purposes).

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/validation/limits.go
+++ b/pkg/util/validation/limits.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


NickAnge (approved):
LGtm

pkg/util/validation/limits.go
```diff
@@ -129,7 +130,13 @@ type Limits struct {
...
+	IngestionArtificialDelayConditionForTenantsWithLessThanMaxSeries int            `yaml:"ingestion_artificial_delay_condition_for_tenants_with_less_than_max_series" json:"ingestion_artificial_delay_condition_for_tenants_with_less_than_max_series" category:"experimental" doc:"hidden"`
+	IngestionArtificialDelayDurationForTenantsWithLessThanMaxSeries  model.Duration `yaml:"ingestion_artificial_delay_duration_for_tenants_with_less_than_max_series" json:"ingestion_artificial_delay_duration_for_tenants_with_less_than_max_series" category:"experimental" doc:"hidden"`
+	IngestionArtificialDelayConditionForTenantsWithIDGreaterThan     int            `yaml:"ingestion_artificial_delay_condition_for_tenants_with_id_greater_than" json:"ingestion_artificial_delay_condition_for_tenants_with_id_greater_than" category:"experimental" doc:"hidden"`
+	IngestionArtificialDelayDurationForTenantsWithIDGreaterThan      model.Duration `yaml:"ingestion_artificial_delay_duration_for_tenants_with_id_greater_than" json:"ingestion_artificial_delay_duration_for_tenants_with_id_greater_than" category:"experimental" doc:"hidden"`
```
gotjosh, line 138: I think this option is potentially _very_ dangerous as it becomes extremely easy to be greedy with the selection - I was going to suggest logging for this condition so that we can audit the ID selection but this has the potential to be extremely expensive on such a hot path.

I'll leave the judgement up to you.

gotjosh (approved):
LGTM

pkg/util/validation/limits.go
pracucci: I don't see it more risky than the global config option that applies the same latency to **all** tenants. This option, at most, will apply it to all tenants. What risk do you foresee? Any practical example?

gotjosh: When you set it globally, it's because you _really_ want to set it for all tenants. 

Maybe, I don't understand the practical use case well enough - but I'm thinking about a scenario where I want to select IDs `1`, `2` and `3` so I set it to `0` and then either `4` gets created in gcom or I don't realise there was a `4` to begin with so that also ends up with the latency when I originally didn't want to select it. 

It might as well be safe to select an ID that was newly created or wasn't intended, but the point stand that you select something you weren't expecting.

As I said, I don't have a problem with it and feel free to merge/continue it was mostly an observation.

pracucci: Ok let's discuss it offline.

pracucci: Discussed offline. Josh is fine.

PR #11146, opened 2025-04-07
Alertmanager: Use Notifer wrapper function for Grafana configurations.

stevesg:
The wrapper was only being used for the vanilla integrations.

Pulls in https://github.com/grafana/alerting/pull/307


titolins (approved):
:shipit:

PR #11153, opened 2025-04-08
frontend: Enable subquery spin-off with boolean instead of regexp

julienduchesne:
We have enabled the feature in all of our cells internally, and we don't use any regexp other than `.*`. It's unnecessary processing to run a regexp for each request

#### Checklist

- [x] Tests updated.
- [N/A] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/spin_off_subqueries.go
+++ b/pkg/frontend/querymiddleware/spin_off_subqueries.go
@@ -129,7 +129,7 @@ func (s *spinOffSubqueriesMiddleware) Do(ctx context.Context, req MetricsQueryRe
 	}
 
 	if !validation.AllTrueBooleansPerTenant(tenantIDs, s.limits.SubquerySpinOffEnabled) {
-		spanLog.DebugLog("msg", "subquery spin-off is disabled for this tenant")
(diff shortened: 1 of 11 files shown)


pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -128,30 +127,11 @@ func (s *spinOffSubqueriesMiddleware) Do(ctx context.Context, req MetricsQueryRe
...
-			matcher, err := labels.NewFastRegexMatcher(pattern)
-			if err != nil {
-				return nil, apierror.New(apierror.TypeBadData, err.Error())
-			}
-
-			if matcher.MatchString(req.GetQuery()) {
-				matched = true
-				break
-			}
+		if !s.limits.SubquerySpinOffEnabled(tenantID) {
```
56quarters, line 131: This loop can be replaced by something like:

```go
if !validation.AllTrueBooleansPerTenant(tenantIDs, s.limits.SubquerySpinOffEnabled)
    return s.next.Do(ctx, req)
}
```

56quarters (approved):
LGTM with one suggestion

Issue #5713, opened 2023-08-11
`Close` never called on PromQL `Query` used in query splitting or sharding


charleskorn:
We call `Exec` on a PromQL `Query` [here](https://github.com/grafana/mimir/blob/23e30d0a63d319fa19d33e3b796512e3179bc5ac/pkg/frontend/querymiddleware/split_by_instant_interval.go#L183) and [here](https://github.com/grafana/mimir/blob/23e30d0a63d319fa19d33e3b796512e3179bc5ac/pkg/frontend/querymiddleware/querysharding.go#L156) for query splitting and sharding, respectively, but never call `Close` on it.

This means that the PromQL engine can never release resources associated with the query. (At present, it returns a bunch of data structures to a pool, but this could change in the future.)

Fixing this isn't straightforward, as we currently use unsafe conversion to retain a reference to the query result until we encode it and send it to the caller. We'd likely need to add a cleanup mechanism to `PrometheusResponse`.

PR #11157, opened 2025-04-09
query-frontend: Ensure queries are properly closed

jhesketh:
Make sure for each execution of a query we properly call the Close() interface. This is necessary where an engine (such as MQE) pools result slices.

Fixes: https://github.com/grafana/mimir/issues/5713

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -89,9 +89,15 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
 		if err != nil {
 			return storage.ErrSeriesSet(err)
 		}
-		promRes, ok := resp.(*PrometheusResponse)
+
+		// FIXME: newSeriesSetFromEmbeddedQueriesResults copies samples, but does not do a deep copy of histogram spans.
+		// We either need to do a deep copy of the histogram spans or somehow wrap the SeriesSets with a closer.
+		// Labels also need closer inspection as they appear to be pointers too.
```
jhesketh, line 95: I'm inclined to suggest we do a deep copy in these cases. I would have to investigate the impact, but the complexity of changing the Queryable interface or doing something else seems quite high.

That coupled with the possibility that any change here may be short lived after the introduction of the planner.

CC @charleskorn @julienduchesne

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -401,6 +409,8 @@ func mergeCacheExtentsForRequest(ctx context.Context, r MetricsQueryRequest, mer
 		}
 	}
 
+	// The response is marsheld in mergeCacheExtentsWithAccumulator, so we can safely close it afterwards
```
charleskorn, line 412: [nit]
```suggestion
	// The response is marshalled in mergeCacheExtentsWithAccumulator, so we can safely close it afterwards
```

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -365,6 +371,8 @@ func mergeCacheExtentsForRequest(ctx context.Context, r MetricsQueryRequest, mer
 
 	for i := 1; i < len(extents); i++ {
 		if accumulator.End+r.GetStep() < extents[i].Start {
+			// The response is marsheld in mergeCacheExtentsWithAccumulator, so we can safely close it afterwards
+			defer accumulator.Close()
```
charleskorn, line 375: Should we call `accumulator.Close` after the error check below? Otherwise this deferred `Close` won't happen until `mergeCacheExtentsForRequest` returns.

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -858,6 +879,7 @@ func TestSplitAndCacheMiddleware_ResultsCache_ShouldNotCacheRequestEarlierThanMa
 }
 
 func TestSplitAndCacheMiddleware_ResultsCacheFuzzy(t *testing.T) {
+	t.Skip()
```
charleskorn, line 882: Why is this test skipped?

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -1014,6 +1036,7 @@ func TestSplitAndCacheMiddleware_ResultsCacheFuzzy(t *testing.T) {
 }
 
 func TestSplitAndCacheMiddleware_ResultsCache_ExtentsEdgeCases(t *testing.T) {
+	t.Skip()
```
charleskorn, line 1039: Why is this test skipped?

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
charleskorn: >That coupled with the possibility that any change here may be short lived after the introduction of the planner.

I think this is the deciding factor for me, and the fact that histograms aren't yet widely used - so I'm OK with the idea of doing a deep copy of histograms.

>Labels also need closer inspection as they appear to be pointers too.

Can you elaborate on this?

pkg/frontend/querymiddleware/results_cache.go
jhesketh: Actually this was the wrong place to close. The response is still needed after mergeCacheExtentsForRequest and the merged responses are closed further downstream already so this is removed now.

pkg/frontend/querymiddleware/split_and_cache_test.go
jhesketh: This PR is incomplete ;-).

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
jhesketh: > Can you elaborate on this?

Labels are not copied, they are also pointers:
https://github.com/grafana/mimir/blob/67a7b0e15af580a93ba1c644a75d5df1d3a3caa2/pkg/frontend/querymiddleware/sharded_queryable.go#L322
https://github.com/grafana/mimir/blob/67a7b0e15af580a93ba1c644a75d5df1d3a3caa2/pkg/mimirpb/compat_slice.go#L22

Although likely not an issue for MQE at the moment, if we start pooling label resources it may become one. Thus we likely need to consider deep copying labels as well at this point.

charleskorn: `labels.Labels` instances are not pointers:

* with `stringlabels` (current default), they are [a string](https://github.com/prometheus/prometheus/blob/bce72b93d91186785a675fcd45b5bca2669d30de/model/labels/labels_stringlabels.go#L29)
* without `stringlabels` (old default), they are [an immutable slice of labels](https://github.com/prometheus/prometheus/blob/bce72b93d91186785a675fcd45b5bca2669d30de/model/labels/labels.go#L29)

https://github.com/grafana/mimir/blob/67a7b0e15af580a93ba1c644a75d5df1d3a3caa2/pkg/mimirpb/compat_slice.go#L22 is only used if `stringlabels` is not being used, and `stringlabels` has been the default for some time (might be worth proposing removing it to avoid confusion). https://github.com/grafana/mimir/blob/67a7b0e15af580a93ba1c644a75d5df1d3a3caa2/pkg/mimirpb/compat_stringlabels.go#L13 is the equivalent for `stringlabels`.

jhesketh: > labels.Labels instances are not pointers

No, but if MQE starts to pool them and re-use them we need to be aware of when we `Close` the query. My point was more that we might need to put some protection or be careful here, not that they are currently an issue.

charleskorn: We can't pool `labels.Labels` instances, as they're immutable.

pkg/frontend/querymiddleware/codec_json_test.go
```diff
@@ -231,6 +232,26 @@ func TestPrometheusCodec_JSONResponse_Metrics(t *testing.T) {
...
 
+// requireEqualHttpResponse checks the responses are the same with special handling for the Body.
+func requireEqualHttpResponse(t *testing.T, expected, actual *http.Response) {
+	// Compare all HTTP response fields except the Body
+	require.Equal(t, expected.StatusCode, actual.StatusCode)
+	require.Equal(t, expected.Header, actual.Header)
+	require.Equal(t, expected.ContentLength, actual.ContentLength)
+
+	// Verify that body types match
+	require.Equal(t, reflect.TypeOf(expected.Body), reflect.TypeOf(actual.Body))
```
charleskorn, line 243: Why do we need to compare the body types? Can't we just confirm they contain the same or equivalent content?

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -146,7 +146,10 @@ type PrometheusResponseExtractor struct{}
 
 // Extract extracts response for specific a range from a response.
 func (PrometheusResponseExtractor) Extract(start, end int64, from Response) Response {
```
charleskorn, line 148: Should we expand the comment here to clarify whose responsibility it is to close `resp`, and how this relates to the returned `Response`? (eg. does closing `resp` also close the returned value?)

(same comment applies to `ResponseWithoutHeaders`)

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -1698,6 +1729,9 @@ func (q roundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	// TODO Check if it's safe to close the response here. I suspect EncodeMetricsQueryResponse does not do a deep copy and thus it is not yet safe.
```
charleskorn, line 1732: Is this TODO still relevant?

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -179,9 +185,14 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
 			if err != nil {
 				return storage.ErrSeriesSet(fmt.Errorf("error running subquery: %w", err))
 			}
-			promRes, ok := resp.(*PrometheusResponse)
+			// FIXME: newSeriesSetFromEmbeddedQueriesResults copies samples, but does not do a deep copy of histogram spans.
```
charleskorn, line 188: Does this FIXME still apply?

pkg/frontend/querymiddleware/codec_json_test.go
jhesketh: Because we want the tests to be sure that the response has a closer or not.

pkg/frontend/querymiddleware/spin_off_subqueries_queryable.go
```diff
@@ -90,9 +90,8 @@ func (q *spinOffSubqueriesQuerier) Select(ctx context.Context, _ bool, hints *st
 			return storage.ErrSeriesSet(err)
 		}
 
-		// FIXME: newSeriesSetFromEmbeddedQueriesResults copies samples, but does not do a deep copy of histogram spans.
-		// We either need to do a deep copy of the histogram spans or somehow wrap the SeriesSets with a closer.
-		// Labels also need closer inspection as they appear to be pointers too.
+		// newSeriesSetFromEmbeddedQueriesResults copies the values, so it is safe to close the query after it is done.
+		// (It does not copy labels, but MQE does not reuse labels so it is still safe).
```
charleskorn, line 94: This comment (and the similar comment on line 188 below) is a bit misleading: while this is true, the main thing we're relying on is the fact that a `labels.Labels` instance is immutable.

charleskorn (approved):
LGTM modulo suggestion for clarifying comment:

PR #11159, opened 2025-04-09
WIP: Distributor: Log deduplicated samples with conflicting timestamps

aknuds1:
#### What this PR does

Enhance distributor to log deduplicated samples with conflicting timestamps.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 * [ENHANCEMENT] Query-frontend: Add `cortex_query_samples_processed_cache_adjusted_total` metric. #11164
 * [ENHANCEMENT] Ingester/Distributor: Add `cortex_cost_attribution_*` metrics to observe the state of the cost-attribution trackers. #11112
 * [ENHANCEMENT] gRPC/HTTP servers: Add `cortex_server_invalid_cluster_validation_label_requests_total` metric, that is increased for every request with an invalid cluster validation label. #11241
+* [ENHANCEMENT] Distributor: Trace when deduplicating a metric's samples/histograms due to conflicting timestamps. #11159
```
tacole02, line 41:
```suggestion
* [ENHANCEMENT] Distributor: Trace when deduplicating a metric's samples or histograms due to conflicting timestamps. #11159
```

tacole02 (approved):

pkg/distributor/distributor.go
```diff
@@ -1208,6 +1208,9 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
 		var firstPartialErr error
 		var removeIndexes []int
 		totalSamples, totalExemplars := 0, 0
+		// Cap how many metrics with dedupled samples to emit trace events for.
+		const maxTrace = 10
```
bboreham, line 1212:
```suggestion
		const maxMetricsWithDeduplicatedSamplesToTrace = 10
```

Then you don't need a comment.

pkg/distributor/distributor.go
```diff
@@ -1208,6 +1208,9 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
+		dedupedPerMetric := make(map[string]int, maxTrace)
```
bboreham, line 1213: Can we avoid making this allocation in the happy path?

pkg/distributor/distributor.go
```diff
@@ -1246,6 +1264,18 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
 			validatedExemplars += len(ts.Exemplars)
 		}
 
+		// Emit tracing span events for metrics with deduped samples.
+		spanLogger, ctx := spanlogger.NewWithLogger(ctx, d.log, "Distributor.prePushValidationMiddleware")
```
bboreham, line 1268: Can we avoid this work in the happy path?

pkg/distributor/distributor.go
```diff
@@ -1223,12 +1226,27 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
 
+			dedupedSamplesAndHistograms := (rawSamples - len(ts.Samples)) + (rawHistograms - len(ts.Histograms))
+			if dedupedSamplesAndHistograms > 0 {
+				for _, l := range ts.Labels {
+					if l.Name != labels.MetricName {
+						continue
+					}
+					if _, exists := dedupedPerMetric[l.Value]; exists || len(dedupedPerMetric) < maxTrace {
+						dedupedPerMetric[l.Value] += dedupedSamplesAndHistograms
+						break
```
narqo, line 1245: I don't follow this condition. It breaks the loop when a series's name exists in the set, or the length of the set exceeds the max. Where does the code insert something into the set?

bboreham:
Context: the ability to see such dropped samples was lost in https://github.com/grafana/mimir/pull/10430#discussion_r1914434267

pkg/distributor/distributor.go
```diff
@@ -1246,6 +1264,18 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
 
+		// Emit tracing span events for metrics with deduped samples.
+		spanLogger, ctx := spanlogger.NewWithLogger(ctx, d.log, "Distributor.prePushValidationMiddleware")
+		defer spanLogger.Finish()
+		for m, c := range dedupedPerMetric {
+			spanLogger.DebugLog(
+				"msg", "deduplicated samples/histograms with conflicting timestamps from write request",
+				"metric", m,
+				"count", c,
+				"tenant", userID,
```
colega, line 1275: I don't think you need the tenant here, that's already inferred from the span.

pkg/distributor/distributor.go
bboreham: Line 1244.

pkg/distributor/distributor.go
```diff
@@ -1223,12 +1226,27 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
 				droppedNativeHistograms += len(ts.Histograms)
 			}
 
+			dedupedSamplesAndHistograms := (rawSamples - len(ts.Samples)) + (rawHistograms - len(ts.Histograms))
+			if dedupedSamplesAndHistograms > 0 {
+				for _, l := range ts.Labels {
+					if l.Name != labels.MetricName {
+						continue
+					}
+					if _, exists := dedupedPerMetric[l.Value]; exists || len(dedupedPerMetric) < maxTrace {
```
bboreham, line 1243: The `len` check is cheaper than the `exists` check so should come first.
Even better: it could go on the `if` outside the loop, and then the loop can be replaced with `UnsafeMetricNameFromLabelAdapters`.

pkg/distributor/distributor.go
```diff
@@ -1223,12 +1226,27 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
+						dedupedPerMetric[l.Value] += dedupedSamplesAndHistograms
```
bboreham, line 1244: The strings here point into a large gRPC buffer, so care needs to be taken not to retain them.
I don't think they get retained here, but some comment to this effect is probably a good idea.

pkg/distributor/distributor.go
bboreham: Line 1244

pkg/distributor/distributor.go
aknuds1: Done, thx.

pkg/distributor/distributor.go
aknuds1: Done.

pkg/distributor/distributor.go
aknuds1: Can you please check my revision in this regard.

pkg/distributor/distributor.go
aknuds1: Also, `SpanLogger.DebugLog` logs the arguments if in debug mode. Isn't there a risk that the logger retains the unsafe strings until flushed?

tacole02 (approved):

bboreham:
Another nit: I found it confusing that the title talks about "conflicting timestamps".  
I thought the situation addressed is that both the timestamp and the sample value are identical, within the same request.  This isn't a "conflict", though it is something the sender might like to know more about because it is likely unintended.
Or am I wrong?

pkg/distributor/distributor.go
```diff
@@ -1246,6 +1269,19 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
 			validatedExemplars += len(ts.Exemplars)
 		}
 
+		if len(dedupedPerMetric) > 0 {
+			// Emit tracing span events for metrics with deduped samples.
+			spanLogger, _ := spanlogger.NewWithLogger(ctx, d.log, "Distributor.prePushValidationMiddleware")
```
bboreham, line 1274: I'm a little hesitant about adding a whole new span just to report some debug information, but I guess we can go with it and see if the cost is noticeable.

pkg/distributor/distributor.go
```diff
@@ -1223,12 +1226,32 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
...
 			if countDroppedNativeHistograms {
 				droppedNativeHistograms += len(ts.Histograms)
 			}
 
+			dedupedSamplesAndHistograms := (rawSamples - len(ts.Samples)) + (rawHistograms - len(ts.Histograms))
+			if dedupedSamplesAndHistograms > 0 {
+				if dedupedPerMetric == nil {
+					dedupedPerMetric = make(map[string]int, maxMetricsWithDeduplicatedSamplesToTrace)
+				}
+				name, err := extract.UnsafeMetricNameFromLabelAdapters(ts.Labels)
```
bboreham, line 1242: We could have a fallback of `"unnamed"`, something like that.

bboreham, line 1242: This whole bit should come after the check on `validationErr`; there's no point in logging deduplication of series that were removed due to an error.

pkg/distributor/distributor.go
```diff
@@ -1246,6 +1269,19 @@ func (d *Distributor) prePushValidationMiddleware(next PushFunc) PushFunc {
 			validatedExemplars += len(ts.Exemplars)
 		}
 
+		if len(dedupedPerMetric) > 0 {
+			// Emit tracing span events for metrics with deduped samples.
+			spanLogger, _ := spanlogger.NewWithLogger(ctx, d.log, "Distributor.prePushValidationMiddleware")
+			for m, c := range dedupedPerMetric {
+				spanLogger.DebugLog(
+					"msg", "deduplicated samples/histograms with conflicting timestamps from write request",
```
bboreham, line 1277:
```suggestion
					"msg", "deduplicated samples/histograms with identical timestamp and value from write request",
```

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 * [ENHANCEMENT] Ingest-Storage: Add `ingest-storage.kafka.producer-record-version` to allow control Kafka record versioning. #11244
 * [ENHANCEMENT] Ruler: Update `<prometheus-http-prefix>/api/v1/rules` and `<prometheus-http-prefix>/api/v1/alerts` to reply with HTTP error 422 if rule evaluation is completely disabled for the tenant. If only recording rule- or alerting rule evaluation is disabled for the tenant, the response now includes a corresponding warning. #11321
 * [ENHANCEMENT] Add tenant configuration block `ruler_alertmanager_client_config` which allows the Ruler's Alertmanager client options to be specified on a per-tenant basis. #10816
+* [ENHANCEMENT] Distributor: Trace when deduplicating a metric's samples/histograms due to conflicting timestamps. #11159
```
bboreham, line 47:
```suggestion
* [ENHANCEMENT] Distributor: Trace when deduplicating a metric's samples/histograms. #11159
```

bboreham (commented):
One important point about ordering; some other thoughts about wording.

tacole02 (approved):

pkg/distributor/distributor.go
aknuds1: Done.

aknuds1: Done.

pkg/distributor/distributor.go
aknuds1: The validation code only checks for duplicated timestamps though. Isn't it more correct to reflect this detail?

aknuds1:
@bboreham I pushed a revision taking a span from the context instead, to avoid having to create a new one. Checking Mimir traces, there should be a parent span `HTTP POST - api_v1_push` if I'm not mistaken.

tacole02 (approved):

pkg/distributor/distributor_test.go
```diff
@@ -1524,7 +1606,7 @@ func TestDistributor_Push_CountDroppedNativeHistograms(t *testing.T) {
 	}
 }
 
-func TestDistributor_SampleDuplicateTimestamp(t *testing.T) {
+func TestDistributor_ValidateSeries(t *testing.T) {
```
aknuds1, line 1606: After fiddling with this test, and realizing I need to make my change in `TestDistributor_Push` instead due to the former only testing `Distributor.validateSeries`, I think the test's name should reflect what it tests. I.e., it's really a method unit test.

It's not a big deal, so I would be OK with reverting this change :) I just think it will help in future maintenance.

pkg/distributor/distributor_test.go
```diff
@@ -235,8 +241,41 @@ func TestDistributor_Push(t *testing.T) {
...
+			`,
+		},
 	} {
 		t.Run(name, func(t *testing.T) {
+			spanRecorder := tracetest.NewSpanRecorder()
+			origTracerProvider := otel.GetTracerProvider()
+			t.Cleanup(func() {
+				otel.SetTracerProvider(origTracerProvider)
+			})
+			otel.SetTracerProvider(trace.NewTracerProvider(trace.WithSpanProcessor(spanRecorder)))
```
colega: I'm not sure you need to do this. Can you just call `trace.NewTracerProvider(trace.WithSpanProcessor(spanRecorder)).Tracer("test")` below?

If you still want to replace the global tracer provider, I would rather just set a global once in `init()`. OTel doesn't play very well with replacing an existing tracer provider, because the [global tracer](https://github.com/grafana/mimir/blob/2658b3c477772b54c2149103bfa03f10d1d6dba4/pkg/distributor/distributor.go#L67-L67) isn't going to change after the first setting.

pkg/distributor/distributor_test.go
```diff
@@ -277,6 +321,45 @@ func TestDistributor_Push(t *testing.T) {
...
+				for _, ev := range sp.Events() {
+					if ev.Name != eventName {
+						continue
+					}
+					require.Emptyf(t, event.Name, "Expected only one event with name %q", eventName)
+
+					event = ev
+				}
+			}
+			if tc.expectDedupTraceEvents {
```
colega: Can you extract this into a helper? Something like `findEventWithName(spanName, eventName)`

pkg/distributor/distributor_test.go
```diff
@@ -277,6 +321,45 @@ func TestDistributor_Push(t *testing.T) {
...
+						count = attr.Value.AsInt64()
+					default:
+						require.Failf(t, "Unrecognized event attribute %q", string(attr.Key))
+					}
+				}
+				assert.Equal(t, "test_metric", metric)
+				assert.Equal(t, int64(1), count)
+			} else {
+				require.Emptyf(t, event.Name, "Expected no sample deduplication trace events")
+			}
```
colega: Same here, it would be easier to read the test if we had a helper saying `requireContainsAttributes(t, expectedAttributes, event.Attributes)`.

pkg/distributor/distributor_test.go
aknuds1: Done, thanks!

pkg/distributor/distributor_test.go
aknuds1: PTAL.

pkg/distributor/distributor_test.go
aknuds1: PTAL.

pkg/distributor/distributor_test.go
```diff
@@ -277,10 +316,50 @@ func TestDistributor_Push(t *testing.T) {
...
+				})
+			}
 		})
 	}
 }
 
+func verifySpanEvent(t *testing.T, event trace.Event, name string, attributes []attribute.KeyValue) {
+	t.Helper()
+
+	assert.Equal(t, name, event.Name)
```
colega, line 338: Do you need to check the name? You already found the event with this name.

pkg/distributor/distributor_test.go
```diff
@@ -277,10 +316,50 @@ func TestDistributor_Push(t *testing.T) {
...
+	assert.Equal(t, attributes, event.Attributes)
+}
+
+func findSpanEvent(t *testing.T, spanRecorder *tracetest.SpanRecorder, span oteltrace.Span, spanName, eventName string) trace.Event {
+	t.Helper()
+
+	span.End()
```
colega: Why are we passing the `span` to this method? Why is this method ending the span? Seems to be totally unrelated.

pkg/distributor/distributor_test.go
```diff
@@ -277,10 +316,50 @@ func TestDistributor_Push(t *testing.T) {
...
 }
 
+func verifySpanEvent(t *testing.T, event trace.Event, name string, attributes []attribute.KeyValue) {
+	t.Helper()
+
+	assert.Equal(t, name, event.Name)
+	assert.Equal(t, attributes, event.Attributes)
+}
+
+func findSpanEvent(t *testing.T, spanRecorder *tracetest.SpanRecorder, span oteltrace.Span, spanName, eventName string) trace.Event {
```
colega: Nit, but idiomatic go would be to return `trace.Event, bool` indicating whether it was found.

pkg/distributor/distributor_test.go
aknuds1: Good point, I will fix it.

pkg/distributor/distributor_test.go
aknuds1: Done. PTAL.

pkg/distributor/distributor_test.go
aknuds1: Done. PTAL.

pkg/distributor/distributor_test.go
aknuds1: Only saw this now. Maybe it's best for completeness, since it's a separate function for verifying a span event?

colega (approved):

PR #11164, opened 2025-04-09
Query Frontend: add cortex_query_samples_processed_cache_adjusted_total metric

Konstantinov-Innokentii:
#### What this PR does
The metric tracks the number of samples processed to execute a query if no query results cache was enabled.

**How does it work:**
  The extent now has a SamplesProcessedPerStep field. splitAndCacheMiddleware tracks partial stats of each downstream response it made and when downstreamRequest is converted to an extent perStepStats are approximated from totalSamplesProcessed for this request, based on request range and step. It's approximated since MQE does not expose per-step sample statistics.  
  When extents are used to respond to a query in the partitionCacheExtents, the corresponding perStepStats are extracted and merged. The same happens when extents are merged or truncated when filtering recent cache extents.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +649,16 @@ func cacheHashKey(key string) string {
 	// Hex because memcache keys must be non-whitespace non-control ASCII
 	return hex.EncodeToString(hasher.Sum(nil))
 }
+
+// extentSamplesInSubrange counts the appropriate number of samples for the subrange within the extent.
+// subrangeStart and subrangeEnd must be within the extent.
+func extentSamplesInSubrange(extent Extent, subrangeStart int64, subrangeEnd int64) uint64 {
+	extentTimeRange := extent.End - extent.Start
+	subrange := subrangeEnd - subrangeStart
+	if extentTimeRange > 0 {
```
pracucci, line 658: Shouldn't happen, but maybe we should also add a protection on subrange too?

```suggestion
	if extentTimeRange > 0 && subrange > 0 {
```

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +649,16 @@ func cacheHashKey(key string) string {
...
+		ratio := float64(subrange) / float64(extentTimeRange)
+		return uint64(float64(extent.SamplesProcessed) * ratio)
+	} else {
+		return extent.SamplesProcessed
```
pracucci, line 662: Ot should we return 0 instead? Again, this is an edge case that shouldn't happen in practice, but maybe 0 is more accurate in this case because it means we're looking up for a 0-length range.

pracucci (approved):
Very nice work! I just left a couple of nits. 👏

pkg/frontend/transport/handler.go
```diff
@@ -148,6 +149,10 @@ func NewHandler(cfg HandlerConfig, roundTripper http.RoundTripper, log log.Logge
 			Name: "cortex_query_samples_processed_total",
 			Help: "Number of samples processed to execute a query.",
 		}, []string{"user"})
+		h.querySamplesProcessedCacheAdjusted = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_query_samples_processed_cache_adjusted_total",
+			Help: "Number of samples processed to execute a query if no query results cache was enabled. Note, that this number is approximate.",
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: Thanks, I spent quite a time with chatgpt for naming :)

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: So, there are actually 3 edge cases I can think about:
1. **subrangeStart >= subrangeEnd, which means subrange is <=0**. That's the case you mention in [previous thread](https://github.com/grafana/mimir/pull/11164#discussion_r2036870591). Should never happen but we can add a check you proposed above, makes sense for me to return 0 in this case.
2. **subrangeStart and subrangeEnd are not withing extent at all.** It should never happen too, but I can add a check and return 0 if subrange is not within extent.
3. **Extent is 0 length.** I'm not sure about that. I return all SamplesProcessed in this case, because I saw this subtest in the `TestPartitionCacheExtents` and assumed that a case when the extent has a zero range, but some samples is valid.
```
{
			name: "Test when hit has a large step and only a single sample extent.",
			// If there is a only a single sample in the split interval, start and end will be the same.
			input: &PrometheusRangeQueryRequest{
				start: 100,
				end:   100,
				step:  10,
			},
			prevCachedResponse: []Extent{
				mkExtent(100, 100),
			},
			expectedCachedResponse: []Response{
				mkAPIResponse(100, 105, 10),
			},
},
```
My logic is based on the `mergeCacheExtentsForRequest`.  If there is a base extent with range 100-200 and there is 0-range extent in range 205-205 it's still withing a step range with base extent, so it will be accumulated -> we need to merge samples too. 

Does it makes sense?

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: I'll answer in the thread below, to keep conversation about edge cases in one place.

pkg/frontend/querymiddleware/model.proto
```diff
@@ -67,6 +67,7 @@ message Extent  {
   // If the response is combination of multiple queries over time, all of which had timestamp set, this is the timestamp of oldest query.
   // When merging extents and some of them have 0 query timestamp, we keep non-zero timestamp, if possible.
   int64 query_timestamp_ms = 6;
+  uint64 samples_processed = 7;
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: done

pkg/frontend/querymiddleware/model.proto
```diff
@@ -67,6 +67,13 @@ message Extent  {
   // If the response is combination of multiple queries over time, all of which had timestamp set, this is the timestamp of oldest query.
   // When merging extents and some of them have 0 query timestamp, we keep non-zero timestamp, if possible.
   int64 query_timestamp_ms = 6;
+  // Number of samples processed per step to create this Extent.
```
pracucci, line 70: Can we sort it by timestamp by contract? If so, I would document it here, given it's a design decision.

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +660,64 @@ func cacheHashKey(key string) string {
 	// Hex because memcache keys must be non-whitespace non-control ASCII
 	return hex.EncodeToString(hasher.Sum(nil))
 }
+
+// approximateSamplesProcessedPerStep approximates the samples processed per step from the total samples processed.
+// We had to use appoximation, becase MQE doesn't support per step stats yet.
```
pracucci, line 665: [nit]
```suggestion
// We had to use approximation, because MQE doesn't support per step stats yet.
```

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +660,64 @@ func cacheHashKey(key string) string {
...
+func approximateSamplesProcessedPerStep(start int64, end int64, step int64, samplesProcessed uint64) []StepStat {
+	numSteps := (end-start)/step + 1
+	// approxPerStep is not float becase:
```
pracucci, line 668: [nit]
```suggestion
	// approxPerStep is not float because:
```

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +660,64 @@ func cacheHashKey(key string) string {
...
+	// 1. Real per step stats always have integer values. Having int in approximation will simplify migration to per step stats from engine
+	// 2. approxPerStep < 0 doesn't make sense
```
pracucci, line 670: [nit] I would remove this, because you can store negatives in ints too :)
```suggestion
```

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -639,3 +660,64 @@ func cacheHashKey(key string) string {
...
+	for _, step := range extent.SamplesProcessedPerStep {
+		if start <= step.Timestamp && step.Timestamp <= end {
+			result = append(result, step)
+		}
+	}
+
+	return result
+}
+
+func mergeSamplesProcessedPerStep(steps ...[]StepStat) []StepStat {
```
pracucci, line 700: If we simply define that `StepStat` must be ordered by timestamp, then you don't need to use a map here. You can just go building the merged slice in a single pass. Keep it simple, just take 2 inputs. Don't need to be variadic.

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -519,6 +529,15 @@ func partitionCacheExtents(req MetricsQueryRequest, extents []Extent, minCacheEx
...
 	}
 
+	// Track samples processed from cache
+	var cachedSamplesProcessed uint64 = 0
+	for _, stepStat := range cachedPerStepStat {
+		cachedSamplesProcessed += uint64(stepStat.Value)
+	}
+	if details := QueryDetailsFromContext(ctx); details != nil {
+		details.SamplesProcessedFromCache = cachedSamplesProcessed
+	}
```
pracucci, line 539: Design-wise, I don't think it's `partitionCacheExtents()` responsibility taking the context in input and updating it.

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -519,6 +529,15 @@ func partitionCacheExtents(req MetricsQueryRequest, extents []Extent, minCacheEx
 		requests = append(requests, req)
 	}
 
+	// Track samples processed from cache
+	var cachedSamplesProcessed uint64 = 0
+	for _, stepStat := range cachedPerStepStat {
+		cachedSamplesProcessed += uint64(stepStat.Value)
+	}
+	if details := QueryDetailsFromContext(ctx); details != nil {
+		details.SamplesProcessedFromCache = cachedSamplesProcessed
```
pracucci, line 538: I'm a bit confused. `partitionCacheExtents()` can be called multiple times for a single input query. Here we overwrite `SamplesProcessedFromCache` each time 🤔

Konstantinov-Innokentii: Yep, my bad; thanks for catching it.  Pushed a fix

tacole02 (approved):

PR #11171, opened 2025-04-10
MQE: ensure aggregations release any intermediate state if they are closed before all series are read

charleskorn:
#### What this PR does

This PR fixes an issue where aggregations could not return some slices to their respective pools if the operator is closed before all series have been read. This reduces the effectiveness of memory pooling, but does not cause any other negative consequences.

This could happen if a binary operation is consuming the output of the aggregation and not all series are needed for that binary operation, for example.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/aggregations/count.go
+++ b/pkg/streamingpromql/operators/aggregations/count.go
@@ -87,7 +87,9 @@ func (g *CountGroupAggregationGroup) ComputeOutputSeries(_ types.ScalarData, tim
 		}
 	}
 
-	types.Float64SlicePool.Put(g.values, memoryConsumptionTracker)
-
 	return types.InstantVectorSeriesData{Floats: floatPoints}, false, nil
+}
+
+func (g *CountGroupAggregationGroup) Close(memoryConsumptionTracker *limiting.MemoryConsumptionTracker) {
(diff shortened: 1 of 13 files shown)


pkg/streamingpromql/operators/aggregations/count.go
```diff
@@ -87,7 +87,9 @@ func (g *CountGroupAggregationGroup) ComputeOutputSeries(_ types.ScalarData, tim
 		}
 	}
 
-	types.Float64SlicePool.Put(g.values, memoryConsumptionTracker)
-
 	return types.InstantVectorSeriesData{Floats: floatPoints}, false, nil
 }
+
+func (g *CountGroupAggregationGroup) Close(memoryConsumptionTracker *limiting.MemoryConsumptionTracker) {
+	types.Float64SlicePool.Put(g.values, memoryConsumptionTracker)
```
jhesketh, line 94: This doesn't look safe to close multiple times. Granted this isn't an operator so the same expectation may not apply, but I wonder if we should be consistent?
(same for other changes in this pr)

charleskorn: Done: a333aab

jhesketh (approved):

Issue #9072, opened 2024-08-22
Add remote write protocol 2.0 support


dimitarvdimitrov:
### What should we do?

Prometheus added remote write protocol 2.0 **experimental** support in v2.54.0 (released on 2024-08-09). We should add the support in Mimir too.

### How will we do it (roughly)?

- Add remote write 2.0 support in Mimir distributors
- Backport all (applicable) optimizations we did to remote write 1.0 (un)marshalling
- Compare performance between remote write 1.0 and 2.0
- Allow to enable 2.0 support on a per-tenant basis

Out of the scope of this work:

- Change data format between distributors and ingesters / Kafka (keep using protocol 1.0 format). This should be a follow up deliverable.

### Size?

Between Medium  (= ~1 month) and Large (= ~3 month).

### What will we deliver?

- Add remote write 2.0 experimental support in Mimir, fully merged in Mimir but disabled by default
- Test remote write 2.0 in dev

### What are the documentation dependencies?

- No documentation changes to docs until enabled by default

### Urgency?

Not urgent yet, but we can't lag too much behind Prometheus

jmichalek132:
FYI otel collector contrib has lfx mentorship project starting in September to add support in the remote write exporter for remote write 2.0. Tracking issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33661.

krajorama:
This is becoming a bit more urgent now due to two projects depending on it: native histograms with custom buckets ([NHCB](https://github.com/prometheus/prometheus/issues/13485)) and proper handling of OTLP created timestamp (current [solution](https://github.com/grafana/mimir-prometheus/pull/791) we don't want to upstream because it's a bit of a workaround).

krajorama:
Based on #10432 

Task list:

- [x] Remove/refactor debug statements in PreallocWriteRequest.unmarshalRW2
- [x] Add unit test for PreallocWriteRequest.unmarshalRW2
- [x] Add benchmark test for PreallocWriteRequest.unmarshalRW2
- [x] DECIDED AGAINST this: Add configuration option to enable/disable RW2.0 support. If disabled it should return 415 as in #10423 
- [x] Add metric to measure which protocol is in use
- [x] Add more testcases to TestDistributorRemoteWrite2 integration test
- [x] Patch mimir.pb.go to ignore RW2 only fields when unmarshal RW1
- [x] Add Changelog entry
- [ ] Run Prometheus tool that checks compatibility against out endpoint if possible
[...]

PR #11192, opened 2025-04-11
feat(rw2): add version label to cortex_distributor_requests_in_total

krajorama:
#### What this PR does

Currently we're only interested in adoption so only changing one metric and not complicating things with validating/saving the version provided by the client. We simply add `version="1.0"` or `version="2.0"` label to `cortex_distributor_requests_in_total`. This also reduces churn caused by the labels change.

Related to: #9072

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/distributor/distributor.go
```diff
@@ -421,7 +421,7 @@ func New(cfg Config, clientConfig ingester_client.Config, limits *validation.Ove
 		incomingRequests: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "cortex_distributor_requests_in_total",
 			Help: "The total number of requests that have come in to the distributor, including rejected or deduped requests.",
-		}, []string{"user"}),
+		}, []string{"user", "version"}),
```
krajorama, line 424: I wonder whether we should have two metrics instead with constant labels. Maybe slightly more efficient? Although we already have a dynamic attribute in "user" . Also we might have more detailed version later?

npazosmendez (approved):

pkg/distributor/distributor.go
aknuds1: I don't know personally.

pkg/mimirpb/mimir.pb.go
```diff
@@ -7584,7 +7584,6 @@ func (m *WriteRequest) Unmarshal(dAtA []byte) error {
 			m.Metadata = append(m.Metadata, metadata)
 		}
 		m.rw2symbols.releasePages()
-		m.unmarshalFromRW2 = false
```
aknuds1, line 7587: What does this change signify?

aknuds1 (commented):
Please see question.

pkg/mimirpb/mimir.pb.go
krajorama: Before this PR I cleaned up almost everything that was used to decode the data buffer as remote-write 2, including this marker. However at the place where I need the information about the protocol `(d *Distributor) metricsMiddleware(next PushFunc) PushFunc`  , I only get `WriteRequest`, not `PreallocWriteRequest` so I don't have the info anymore. Easiest was to not clear this marker - doesn't hurt, cause it's not allocation.

aknuds1 (approved):

PR #11197, opened 2025-04-11
Alertmanager: fix mixin queries for ignored zero value metrics

titolins:
## Context
- NPS panels for on-prem customers might display gaps since we've added `WithSkipZeroValueMetrics` for some metrics (e.g. [alertmanager_notifications_total](https://github.com/grafana/mimir/blob/main/pkg/alertmanager/alertmanager_metrics.go#L343-L343))


56quarters (approved):
LGTM

PR #11203, opened 2025-04-13
MQE improve memory consumption estimate with Label information

lamida:
#### What this PR does

Add memory consumption estimate with Label information. The estimation is based on the following:
- The size of `labels.Label` struct (32 bytes).
- Average length of string name of value of the label (8 bytes each).
- Average number of labels per series (guesstimating to 10).

I will try to figure out to improve the guesstimate of the two averages above by looking at the ops data.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -1519,7 +1519,7 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			// At peak, we'll hold all the output samples plus one series, which has one sample.
 			// The output contains five samples, which will be rounded up to 8 (the nearest power of two).
 			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.LabelPairEstimatedSize),
-			instantQueryLimit:        4944,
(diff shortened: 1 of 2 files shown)


charleskorn (commented):
This isn't quite what I had in mind: I think we need to accurately account for the size of labels given they can vary wildly, and we need to track this for anything that contains labels (`SeriesMetadata` slices being the main missing source not yet accounted for).

lamida:
@charleskorn 
> I think we need to accurately account for the size of labels given they can vary wildly
Yes I am thinking about this too, but I started only with this estimate, because to track the size labels accurately, it means we need to move the call to `types.VectorPool.Get(len(series), q.memoryConsumptionTracker)` from its current place in `populateVectorFromInstantVectorOperator`. Now we just get the pool based on the count of series. But to account the exact size of labels, we need to move this or having additional place to calculate it at the point we load the series. Is this inline with what you have in mind for handling labels size variation?

> SeriesMetadata slices being the main missing source not yet accounted for
To be honest, I didn't think about this. Will add this. Do you think there is additional place where labels are used apart of `SeriesMetadata` and the time when we load sample?

lamida:
Closing this first in favor of https://github.com/grafana/mimir/pull/11219

PR #11204, opened 2025-04-14
MQE: make `nil` checks consistent in operator `Close` methods

charleskorn:
#### What this PR does

It is safe to pass a `nil` slice to the `Put` method of a `BucketedPool` or `LimitingBucketedPool`, however some parts of MQE unnecessarily check if a slice is `nil` before calling `Put`. 

This PR makes all places consistent, removing the unnecessary checks.

Addresses https://github.com/grafana/mimir/pull/11162#discussion_r2035802201

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/aggregations/topkbottomk/instant_query.go
+++ b/pkg/streamingpromql/operators/aggregations/topkbottomk/instant_query.go
@@ -286,10 +286,8 @@ func (t *InstantQuery) Close() {
 	t.Inner.Close()
 	t.Param.Close()
 
-	if t.values != nil {
-		types.Float64SlicePool.Put(t.values, t.MemoryConsumptionTracker)
-		t.values = nil
-	}
+	types.Float64SlicePool.Put(t.values, t.MemoryConsumptionTracker)
+	t.values = nil
 }
 
 type instantQueryGroup struct {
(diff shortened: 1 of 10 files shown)


lamida (approved):
The change makes sense

jhesketh (commented):
Are there any cases where we would do a `Put` without then setting the slice to `nil`?

Since the nil is required I wonder if we can do something to make it harder to forget/miss. Eg linting, or returning the slice value (ie nil) so you always have to handle it?

charleskorn:
> Are there any cases where we would do a `Put` without then setting the slice to `nil`?
> 
> Since the nil is required I wonder if we can do something to make it harder to forget/miss. Eg linting, or returning the slice value (ie nil) so you always have to handle it?

I like the idea of trying to make it harder to forget to do this. 

There are some cases where it would be reasonable to not set the slice to `nil`. The most common is a function-scoped variable that is returned to the pool before the function returns (eg. with a `defer`), for example:

```go
func foo() {
    someSeries := getSomeSeriesMetadata()
    defer types.PutSeriesMetadataSlice(someSeries)

    // ... do something with someSeries
}
```

I had a quick look for existing Go linters that would be helpful here (either checking for return values that aren't used or for requiring that a code pattern is followed by another one) and I wasn't able to find any - did you have something in mind?

PR #11219, opened 2025-04-14
MQE: Add SeriesMetadata limiting_pool memory tracker

lamida:
#### What this PR does

- Move SeriesMetadata into limiting_pool as `SeriesMetadataSlicePool` so that it will be part of memory limit and estimate tracker in MQE
- Updated `TestOperator` to update MemoryConsumptionTracker in SeriesMetadata method call so that all tests behave correctly.
- Update the use of old SeriesMetadata pool in all operators into the new `SeriesMetadataSlicePool`
- Update and fix all tests.
- Notable tests to check are `TestMemoryConsumptionLimit_SingleQueries` and `TestMemoryConsumptionLimit_MultipleQueries`.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
+++ b/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
@@ -675,17 +675,17 @@ func TestOrBinaryOperation_ReleasesIntermediateStateIfClosedEarly(t *testing.T) 
 	for name, testCase := range testCases {
 		t.Run(name, func(t *testing.T) {
 			timeRange := types.NewInstantQueryTimeRange(time.Now())
-			left := &operators.TestOperator{Series: testCase.leftSeries, Data: make([]types.InstantVectorSeriesData, len(testCase.leftSeries))}
-			right := &operators.TestOperator{Series: testCase.rightSeries, Data: make([]types.InstantVectorSeriesData, len(testCase.rightSeries))}
-			vectorMatching := parser.VectorMatching{On: true, MatchingLabels: []string{"group"}}
 			memoryConsumptionTracker := limiting.NewMemoryConsumptionTracker(0, nil)
+			left := &operators.TestOperator{Series: testCase.leftSeries, Data: make([]types.InstantVectorSeriesData, len(testCase.leftSeries)), MemoryConsumptionTracker: memoryConsumptionTracker}
+			right := &operators.TestOperator{Series: testCase.rightSeries, Data: make([]types.InstantVectorSeriesData, len(testCase.rightSeries)), MemoryConsumptionTracker: memoryConsumptionTracker}
+			vectorMatching := parser.VectorMatching{On: true, MatchingLabels: []string{"group"}}
 			o := NewOrBinaryOperation(left, right, vectorMatching, memoryConsumptionTracker, timeRange, posrange.PositionRange{})
 
 			ctx := context.Background()
 			outputSeries, err := o.SeriesMetadata(ctx)
 			require.NoError(t, err)
 			require.Equal(t, testutils.LabelsToSeriesMetadata(testCase.expectedOutputSeries), outputSeries)
-
+			types.SeriesMetadataSlicePool.Put(outputSeries, memoryConsumptionTracker)
 			// Read the output series to trigger the loading of some intermediate state for at least one of the output groups.
 			for range testCase.closeAfterReadingIndex + 1 {
 				_, err := o.NextSeries(ctx)
--- a/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
+++ b/pkg/streamingpromql/operators/binops/or_binary_operation_test.go
@@ -545,6 +545,7 @@ func TestOrBinaryOperation_ClosesInnerOperatorsAsSoonAsPossible(t *testing.T) {
 			} else {
 				require.Equal(t, testutils.LabelsToSeriesMetadata(testCase.expectedOutputSeries), outputSeries)
 			}
+			types.SeriesMetadataSlicePool.Put(outputSeries, memoryConsumptionTracker)
 
 			if testCase.expectLeftSideClosedAfterOutputSeriesIndex == -1 {
 				require.True(t, left.Closed, "left side should be closed after SeriesMetadata, but it is not")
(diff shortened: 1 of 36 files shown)


pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -200,7 +219,14 @@ func (p *LimitingBucketedPool[S, E]) Put(s S, tracker *limiting.MemoryConsumptio
 		}
 	}
 
-	tracker.DecreaseMemoryConsumption(uint64(cap(s)) * p.elementSize)
+	estimatedBytes := uint64(cap(s)) * p.elementSize
+	if sm, isSeriesMetadata := any(s).(SeriesMetadata); isSeriesMetadata {
+		for _, l := range sm.Labels {
+			estimatedBytes += uint64(len(l.Name)+len(l.Value)) * StringSize
+		}
+	}
```
charleskorn, line 227: I think we should tackle incorporating the label memory consumption in a separate PR - this will require a number of changes in a number of places, not just in this pools like this. 

It may also be worthwhile raising a Prometheus PR to add something like a `ByteSize()` method to `labels.Labels`, so we can cheaply get the size of a `labels.Labels` without iterating over all the labels like this.

pkg/streamingpromql/operators/scalars/scalar_to_instant_vector.go
```diff
@@ -8,28 +8,35 @@ import (
...
 
 // ScalarToInstantVector is an operator that implements the vector() function.
 type ScalarToInstantVector struct {
 	Scalar types.ScalarOperator
 
-	expressionPosition posrange.PositionRange
-	consumed           bool
+	expressionPosition       posrange.PositionRange
+	consumed                 bool
+	MemoryConsumptionTracker *limiting.MemoryConsumptionTracker
```
56quarters, line 21: nit: Can you group the exported members together?

pkg/streamingpromql/operators/scalars/scalar_to_instant_vector.go
```diff
@@ -8,28 +8,35 @@ import (
...
 var _ types.InstantVectorOperator = &ScalarToInstantVector{}
 
-func NewScalarToInstantVector(scalar types.ScalarOperator, expressionPosition posrange.PositionRange) *ScalarToInstantVector {
+func NewScalarToInstantVector(scalar types.ScalarOperator, expressionPosition posrange.PositionRange, memoryConsumptionTracker *limiting.MemoryConsumptionTracker) *ScalarToInstantVector {
 	return &ScalarToInstantVector{
-		Scalar:             scalar,
-		expressionPosition: expressionPosition,
+		Scalar:                   scalar,
+		expressionPosition:       expressionPosition,
+		MemoryConsumptionTracker: memoryConsumptionTracker,
```
56quarters, line 30: nit: Can you group exported members together?

pkg/streamingpromql/operators/selectors/selector_test.go
```diff
@@ -16,11 +16,12 @@ import (
 	"github.com/prometheus/prometheus/util/annotations"
 	"github.com/stretchr/testify/require"
 
+	"github.com/grafana/mimir/pkg/streamingpromql/limiting"
 	"github.com/grafana/mimir/pkg/streamingpromql/types"
 )
 
 func TestSeriesList_BasicListOperations(t *testing.T) {
-	list := newSeriesList()
+	list := newSeriesList(&limiting.MemoryConsumptionTracker{})
```
56quarters, line 24: Can this be replaced with the constructor?

pkg/streamingpromql/operators/selectors/selector_test.go
```diff
@@ -56,7 +57,7 @@ func TestSeriesList_OperationsNearBatchBoundaries(t *testing.T) {
 
 	for _, seriesCount := range cases {
 		t.Run(fmt.Sprintf("N=%v", seriesCount), func(t *testing.T) {
-			list := newSeriesList()
+			list := newSeriesList(&limiting.MemoryConsumptionTracker{})
```
56quarters, line 60: Can this be replaced with the constructor?

56quarters (approved):
This LGTM, just a few small suggestions.

pkg/streamingpromql/operators/binops/or_binary_operation_test.go
```diff
@@ -579,6 +580,7 @@ func TestOrBinaryOperation_ClosesInnerOperatorsAsSoonAsPossible(t *testing.T) {
 
 			o.Close()
 			// Make sure we've returned everything to their pools.
+			types.SeriesMetadataSlicePool.Put(outputSeries, memoryConsumptionTracker)
```
charleskorn, line 583: Same as above 🙂

pkg/streamingpromql/operators/test_operator.go
```diff
@@ -26,7 +28,15 @@ func (t *TestOperator) ExpressionPosition() posrange.PositionRange {
...
 func (t *TestOperator) SeriesMetadata(_ context.Context) ([]types.SeriesMetadata, error) {
-	return testutils.LabelsToSeriesMetadata(t.Series), nil
+	series := testutils.LabelsToSeriesMetadata(t.Series)
+	seriesPool, err := types.SeriesMetadataSlicePool.Get(len(series), t.MemoryConsumptionTracker)
+	if err != nil {
+		return nil, err
+	}
+
+	seriesPool = seriesPool[:len(series)]
+	copy(seriesPool, series)
```
charleskorn, line 38: The use of `copy` here seems a bit awkward - perhaps we should reproduce the behaviour of `LabelsToSeriesMetadata` in this method so we don't need the copy?

pkg/streamingpromql/types/limiting_pool.go
lamida: @charleskorn I will add this in separate PR, but can you ellaborate more how adding such method in Prometheus can make it cheap? Won't be at the end still have to iterate the slice anyway?

charleskorn: > Won't be at the end still have to iterate the slice anyway?

No: we can just return the length of the underlying byte slice when `stringlabels` is in use.

pkg/streamingpromql/operators/test_operator.go
lamida: Updated in afcae368b0fcd074d6b95ab57a37aca95718d06f

pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -34,7 +35,8 @@ type Selector struct {
 	querier storage.Querier
 	series  *seriesList
 
-	seriesIdx int
+	seriesIdx                int
+	MemoryConsumptionTracker *limiting.MemoryConsumptionTracker
```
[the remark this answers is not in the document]
lamida: This and all above are updated in 9d3d9399da84e78a0c9d307af1722fa80914e2be

charleskorn (approved):
LGTM, thanks Jon 🚀

PR #11222, opened 2025-04-14
move readTsdb to parquet storage sub-package; rename method & vars for clarity & testability

francoposa:
We want to be able to isolate & unit test this stuff and reading a TSDB block to parquet rows should be general enough that it does not need to be specific to the compactor.

Also updated `pkg/storage/parquet` files to be the package name `pkg/storage/parquet` rather than just `pkg/storage`, to match other storage subpackages.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/compactor/parquet_compactor.go
+++ b/pkg/compactor/parquet_compactor.go
@@ -6,17 +6,13 @@
 package compactor
 
 import (
-	"bytes"
 	"context"
 	"fmt"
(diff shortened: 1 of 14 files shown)


jesusvazquez (approved):
LGTM thanks for making the clear separation Franco!

PR #11229, opened 2025-04-15
#11228 -gem secrets generator

### LINKED ISSUE #11228: Idea: Provide tooling to create secrets that are required for GEM install

Often new users to Kubernetes and Mimir struggle with creating secrets/proper base64 of data, etc. Create a tool that users can provide admin, bucket, license data and the tool generates the secrets for them. 

CLI to create secrets from config.json or command line input. 

manual creation of secrets. 

_No response_

Small (<= 1 month dev)

few, aside from a readme

### Proposer?

pbmoses

pbmoses:
#### What this PR does
addresses #11228 

#11228

Fixes #11228 

#### Checklist
external tool, CHANGELOG NA.

### DIFF
--- a/tools/gem-secrets-generator/READEME.md
+++ b/tools/gem-secrets-generator/READEME.md
@@ -1,0 +1,43 @@
+# GEM secrets generator is provided to assist in creation of the secrets required to install Grafana Enterprise Metrics
+
+## DO NOT STORE GENERATED SECRETS IN GIT!!!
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov (commented):
doesn't kubectl already do this? for example


```bash
kubectl create secret generic --dry-run=client --from-literal=a=b my-secret -oyaml
```
```yaml
apiVersion: v1
data:
  a: Yg==
kind: Secret
metadata:
  creationTimestamp: null
  name: my-secret
```

PR #11233, opened 2025-04-15
ruler: Refactor notifier builder to not depend on ruler config struct

alexweav:
#### What this PR does

This is a refactor in support of https://github.com/grafana/mimir/pull/10816 which makes the Alertmanager client config fields (AM url, etc) per-tenant-configurable.

The change above means:
- We need the ability to build notifiers from something that's not necessarily the ruler service config struct (tenant limits, in the future)
- Per the [tenant limit migration process](https://github.com/grafana/mimir/blob/main/docs/internal/contributing/how-to-convert-config-to-per-tenant-limit.md), the fields in the ruler config struct need to be renamed as deprecated in the code. There's a huge number of references to these fields in tests, that are only used to carry fields in and out and aren't being tested directly.

This PR makes `buildNotifierConfig` and friends take their requested fields directly instead of the entire ruler config blob. That means it's portable with new sources of config, and doesn't reference fields that are about to be marked deprecated.

The linked PR already contains this refactor. I've split this PR out of that one, as these changes bring lots of clutter to its diff.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/manager.go
+++ b/pkg/ruler/manager.go
@@ -61,7 +61,7 @@ type DefaultMultiTenantManager struct {
 
 func NewDefaultMultiTenantManager(cfg Config, managerFactory ManagerFactory, reg prometheus.Registerer, logger log.Logger, dnsResolver AddressProvider) (*DefaultMultiTenantManager, error) {
 	refreshMetrics := discovery.NewRefreshMetrics(reg)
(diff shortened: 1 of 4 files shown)


56quarters (approved):
Nicely done. Thanks!

PR #11242, opened 2025-04-16
Grafana Alertmanager: Preserve order while removing duplicate receivers

JacobsonMT:
#### What this PR does

Previously, while removing duplicate receivers and intermediate map was used and original order was lost. Furthermore, the output order was undefined causing change detection on config syncs to continuously detect a spurious change.

#### Checklist

- [x] Tests updated.


yuri-tceretian (approved):
LGTM!

titolins (approved):
:shipit:

PR #11244, opened 2025-04-16
ingest-storage: Introduce a concept of versioning to Kafka records

alexweav:
#### What this PR does

This PR adds an experimental new custom header to Kafka records, called `X-Record-Version`, that carries versioning information about the record payload. This can be used in the future to make breaking changes to the record format in a backward-compatible way.

It's controlled via two toggles:
-`ingest-storage.kafka.producer-supported-record-version` which controls the max version that distributors can send
-`ingest-storage.kafka.consumer-supported-record-version` which controls the max version that ingesters/blockbuilders can understand.

It's expected that consumers support backwards compatibility for all versions older than the current supported one, so that rollouts and upgrades work correctly. During a rollout it can be expected to receive a mixture of versions.

This PR defines two versions to start with:
- Version 0 (or unversioned) is what is sent today, a protobuf-encoded remote write request with no identifying info.
- Version 1 has the new versioning header, but has content identical to version 0.

The default consumer supported version is 1, implicitly, because the record content remains the same.

Rel: https://github.com/grafana/mimir-squad/issues/2253

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/storage/ingest/config.go
```diff
@@ -108,6 +109,9 @@ type KafkaConfig struct {
 
 	WaitStrongReadConsistencyTimeout time.Duration `yaml:"wait_strong_read_consistency_timeout"`
 
+	ProducerSupportedRecordVersion int `yaml:"producer_supported_record_version" category:"experimental"`
+	ConsumerSupportedRecordVersion int `yaml:"consumer_supported_record_version" category:"experimental"`
```
seizethedave, line 113: Curious about the idea behind these flags. In things I've seen it's often the reader/writer code that knows which version it supports/does not support, so these facts end up baked into the code itself.

alexweav: The consumer flag is essentially pointless in the long run :smile: I debated on leaving it in or not, but decided to keep for the sake of this change. We're still in dev with this idea and it lets me toggle things like a feature gate and test quicker. It also more clearly documents that this is experimental. But, you're absolutely right, it probably doesn't need to exist in the long run, happy to even drop it here if we feel strongly.

The producer flag I would at least prefer to keep a little longer. It gives us a quick way to revert or toggle the version being written without needing to revert the entire binary. (For example, when we build a V2, let's say we write a bug or something where a few requests fail to convert, that only affects a few tenants). I also imagine this will rollout a bit slower than the usual wave cycle, this makes it easier to control.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -4142,6 +4142,17 @@ kafka:
   # CLI flag: -ingest-storage.kafka.wait-strong-read-consistency-timeout
   [wait_strong_read_consistency_timeout: <duration> | default = 20s]
 
+  # (experimental) The record version preferred by this producer.
```
tacole02, line 4145:
```suggestion
  # (experimental) The record version that this producer prefers.
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -4142,6 +4142,17 @@ kafka:
...
+  # CLI flag: -ingest-storage.kafka.producer-supported-record-version
+  [producer_supported_record_version: <int> | default = 0]
+
+  # (experimental) The highest supported record version understood by this
+  # consumer. Versions lower than this value are also assumed to be supported.
```
tacole02, line 4150:
```suggestion
  # (experimental) The highest supported record version that this
  # consumer understands. Versions lower than this value are assumed to also be supported.
```

tacole02 (approved):

pkg/storage/ingest/config.go
narqo: > It gives us a quick way to revert or toggle the version being written without needing to revert the entire binary

I relate to the idea. But just to be frank, in practice, updating a Mimir config, and updating k8s deployments to roll back the binary is the same amount of work. If things went bad because of a bug in a hypothetical V2 pusher, rolling back the whole version is what I would do.

alexweav: I'll remove the consumer flag. No reason to introduce a toggle that isn't really needed. On the producer flag, rollouts are still a bit of a concern to me.

Let's say we add a V2. If a user rolls a new binary to both distributors and ingesters, you can end up with an old ingester receiving a record from a new distributor. If that rollout also started sending V2 messages at the same time, an ingester can receive a message that it doesn't understand yet.

Depending on how we handle such a record (probably we want to switch to blocking consumers instead of skipping the record?) you can end up with either missing samples, or a temporary ingestion hang during the rollout.

A way around that could be to merge a V2 consumer support well ahead of V2 producer support. But, a user might upgrade skipping multiple versions anyway, so eventually someone will run into this problem. The producer flag provides a workaround, as it leaves it up to the administrator. You can bump the version directly at any time, after you are certain that consumer support has rolled out fully.

alexweav:
Yes, I intend to move on with a V2 directly after this, based on the remote write 2.0 format with some slight modifications. 
Some internal analysis is here: https://github.com/grafana/mimir-squad/issues/2253

The validation is admittedly a bit ahead of its time. Remote Write 2.0 has a peculiar property where it successfully marshals into a Remote Write 1.0 struct with no error, but the result of doing this gives an empty request. Without version validation, that means Remote Write 2.0 samples can be silently dropped if there is a mismatch in the version supported by ingesters. This felt bad to me, it seems like we'd want to fail in a more observable and easier to recover way, the version validation provides a way to better see when this problem happens. It's not strictly needed for v0 and v1, but will become relevant soon. I can drop it here, but will need to reintroduce it (or something similar) soon.

---

I imagine in the long run we'd rather block ingestion, rather than skip the sample as this PR does. To start with, I decided to simply match the behavior when the message is malformed for simplicity, since there's nothing in mimir that can currently produce an unexpected version.

I expect the most common situation where a version mismatch can happen is during a rollout, i.e. a distributor rolls out and enables a V2 before an ingester. By switching to blocking ingestion, an ingester can eventually catch back up and reprocess the samples.

docs/sources/mimir/configure/configuration-parameters/index.md
alexweav: I dropped this config option entirely, so this sentence doesn't exist anymore :+1:

pkg/storage/ingest/config.go
```diff
@@ -178,6 +181,8 @@ func (cfg *KafkaConfig) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet)
 
 	f.DurationVar(&cfg.WaitStrongReadConsistencyTimeout, prefix+"wait-strong-read-consistency-timeout", 20*time.Second, "The maximum allowed for a read requests processed by an ingester to wait until strong read consistency is enforced. 0 to disable the timeout.")
 
+	f.IntVar(&cfg.ProducerSupportedRecordVersion, prefix+"producer-supported-record-version", 0, "The record version that this producer prefers.")
```
pracucci, line 184: [nit] The description says "The record version that this producer prefers" but is it really a prefer? I think it's a "must use", no?

pkg/storage/ingest/version.go
```diff
@@ -0,0 +1,86 @@
...
+	"encoding/binary"
+	"fmt"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+
+	"github.com/grafana/mimir/pkg/mimirpb"
+)
+
+const (
+	RecordVersionHeaderKey = "X-Record-Version"
```
pracucci, line 15: [nit] Have you considered a shorter string, given this is going to be encoded in each record?

pracucci (approved):
Nice job, LGTM!

pkg/storage/ingest/config.go
pracucci: > Consumer toggle is gone

Agree on having removed it. I also understand why you need the producer's one.

pkg/storage/ingest/version.go
alexweav: Fair enough. I'll move to just `Version` then. No need to use HTTP semantics and it's obviously a record :smile:

pkg/storage/ingest/config.go
alexweav: Yes, I wrote these toggles before the code and kept the wording a bit intentionally ambiguous. It needs refining.

alexweav: - Changed to `...this producer sends`
- Removed `supported` from the flag name. Now it's just `producer-record-version`

narqo (approved):
🔥 

The new `producer-record-version` flag's name makes more sense to me. Thank you

Issue #2708, opened 2022-08-12
Migrate Mimir tracing to opentelemetry


pracucci:
#### Is your feature request related to a problem? Please describe.

Mimir tracing is based on jaeger client and opentracing, which are deprecated.

#### Describe the solution you'd like

- Migrate to opentelemetry.
- Being consistent between Mimir, Loki and Tempo (hopefully sharing the same code)
  - Tempo introduced opentelemetry tracer support (optional) in http://github.com/grafana/tempo/pull/842
  - Loki hasn't introduced opentelemetry yet

#### Describe alternatives you've considered

_Nothing else considered._

bboreham:
* https://github.com/grafana/dskit/pull/385

colega:
I am currently working on this issue.

[Tempo has been fully migrated to opentelemetry libraries](https://github.com/grafana/tempo/pull/3646), making a breaking change in the exporter config, switching directly to OTel.

Loki (and obviously, Mimir) still use Opentracing. Pyroscope uses OpenTracing, I think, I didn't check, it doesn't matter (since Loki still does).

Both Loki and Mimir depend on `dskit`, and @bboreham pointed out in the previous comment that there were efforts to migrate `dskit` to OTel, however I don't have enough bandwidth to migrate both Mimir and Loki (and Pyroscope?) to OTel, so I'm going to go for a solution that would allow us a seamless transition.

### Step 1: Decouple SpanLogger usages from OpenTracing.
One of the biggest offenders in the transition is `spanlogger` package, which is tightly tied to OpenTracing, embedding an `opentracing.Span` which is accessed by both Mimir and Loki (and some more private repos I found out later). I'm solving that by [not embedding opentracing.Span in spanlogger.SpanLogger](https://github.com/grafana/dskit/pull/679), but by exposing the methods used instead. I haven't put much effort into making the nicest API out there, honestly, I think we can do that once we fully transition Grafana products to OTel.


- [x] : https://github.com/grafana/dskit/pull/679

### Step 2: Make `dskit` compatible with both OpenTracing and OTel

Next step would be to add OTel support like in https://github.com/grafana/dskit/pull/385 but without removing the OpenTracing support.

The approach would be the following:
- Wherever we take a span from the context, we're going to try to find an OTel one and an OpenTracing one, instantiate a wrapper and propagate the span with the library of the parent span.
[...]

PR #11249, opened 2025-04-17
Use OTel library from tracing while keeping Jaeger exporter config

colega:
#### What this PR does

Migrates Mimir to OTel tracing. This is Step 4 from https://github.com/grafana/mimir/issues/2708#issuecomment-2812332898

This should be a noop from a user perspective, as we still configure tracing using Jaeger's env vars and we export traces in Jaeger format.

Ref: https://github.com/grafana/mimir/issues/2708

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * [ENHANCEMENT] Added `-ingest-storage.kafka.fetch-max-wait` configuration option to configure the maximum amount of time a Kafka broker waits for some records before a Fetch response is returned. #11012
 * [ENHANCEMENT] Ingester: Add `cortex_ingester_tsdb_forced_compactions_in_progress` metric reporting a value of 1 when there's a forced TSDB head compaction in progress. #11006
 * [ENHANCEMENT] Ingester: Add `cortex_ingest_storage_reader_records_batch_fetch_max_bytes` metric reporting the distribution of `MaxBytes` specified in the Fetch requests sent to Kafka. #11014
-* [ENHANCEMENT] All: Add experimental support for cluster validation in HTTP calls. When it is enabled, HTTP server verifies if a request coming from an HTTP client comes from an expected cluster. This validation can be configured by the following experimental configuration options: #11010
+* [ENHANCEMENT] All: Add experimental support for cluster validation in HTTP calls. When it is enabled, HTTP server verifies if a request coming from an HTTP client comes from an expected cluster. This validation can be configured by the following experimental configuration options: #11010 #11549
   * `-server.cluster-validation.label`
   * `-server.cluster-validation.http.enabled`
   * `-server.cluster-validation.http.soft-validation`
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/golang/snappy v1.0.0
 	github.com/google/gopacket v1.1.19
 	github.com/gorilla/mux v1.8.1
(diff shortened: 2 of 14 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -35,6 +35,8 @@ import (
 	"github.com/grafana/mimir/pkg/util/validation"
 )
 
+var tracer = otel.Tracer("pkg/blockbuilder")
```
bboreham, line 38: Is it guaranteed that this initialization runs after Otel gets parameters from config?
Would it be worth having some kind of `sync.Once` thing called when it is first needed?

bboreham (commented):
Great work!

pkg/blockbuilder/blockbuilder.go
colega: I was wondering same question, because this definitely runs at package initialization time and we configure tracer for the first time from `main()`.

So I dug the code, and found out that the default implementation of the tracer is a special one which is `sync.Once`-replaced to the first one you manually configure. 

That's actually an inconvenience for tests, but apparently having per-package global tracers is the way to go so I won't fight it.

colega:
> Great work!

@bboreham does that count as an approval? :D

docs/sources/mimir/configure/configure-tracing.md
```diff
@@ -11,36 +11,29 @@ weight: 100
...
+During this transition, Mimir uses OTel libraries, but performs the configuration using Jaeger environment variables.
 
 ## Dependencies
 
-Set up Jaeger deployment to collect and store traces from Grafana Mimir. A
-deployment includes either the Jaeger all-in-one binary, or a distributed
-system of agents, collectors, and queriers. If you run Grafana Mimir on Kubernetes, refer to [Jaeger
-Kubernetes](https://github.com/jaegertracing/jaeger-kubernetes).
+Set up a Jaeger deployment to collect and store traces from Grafana Mimir.
+A deployment includes either the Jaeger all-in-one binary, or a distributed system of agents, collectors, and queriers.
```
aknuds1, line 22: [Nit] That comma looks out of place to me.
```suggestion
A deployment includes either the Jaeger all-in-one binary or a distributed system of agents, collectors, and queriers.
```

pkg/util/spanlogger/spanlogger.go
```diff
@@ -23,15 +22,11 @@ var defaultTenantResolver = tenant.NewMultiResolver()
...
-// to it and can be retrieved with FromContext.
-func New(ctx context.Context, method string, kvps ...interface{}) (*SpanLogger, context.Context) {
-	return spanlogger.New(ctx, util_log.Logger, method, defaultTenantResolver, kvps...)
-}
-
-// NewWithLogger is like New but allows to pass a logger.
-func NewWithLogger(ctx context.Context, logger log.Logger, method string, kvps ...interface{}) (*SpanLogger, context.Context) {
-	return spanlogger.New(ctx, logger, method, defaultTenantResolver, kvps...)
+// New makes a new SpanLogger with a log.Logger to send logs to.
+// It starts a new OTEL span with the method name provided using the tracer provided.
```
aknuds1, line 26: [Nit] OTel isn't an acronym:
```suggestion
// It starts a new OTel span with the method name provided using the tracer provided.
```

aknuds1 (approved):
That's... a lot of changes! LGTM modulo nits.

PR #11256, opened 2025-04-18
undelete-blocks: avoid copies by removing S3 delete markers

andyasp:
Currently `undelete-blocks` is written in a generic way that relies on copying a noncurrent version's data to restore an object. 

#### What this PR does

When using S3-compatible object storage, copying data can be avoided by removing S3 delete markers instead (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManagingDelMarkers.html#RemDelMarker). This PR adds support for this  in `undelete-blocks`. This is a significant optimization when large amounts of data are involved.

It is important to note that deleting versions does introduce risk since it is irreversible. I made a choice to restrict the version deletion to only delete markers (which have no data). If an object is overwritten with a non-delete marker then data is copied like before. This felt like a good compromise to make since differing overwrites for an object within a block is unexpected in Mimir.

Related to https://github.com/grafana/mimir-squad/issues/1805 (this is a private issue, but in summary it's listing potential follow-up improvements related to this tool).

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/tools/undelete-blocks/main.go
+++ b/tools/undelete-blocks/main.go
@@ -20,17 +20,19 @@ import (
 	"github.com/oklog/ulid/v2"
 	"github.com/pkg/errors"
 
+	"github.com/grafana/mimir/pkg/storage/bucket"
 	"github.com/grafana/mimir/pkg/storage/tsdb/block"
 	"github.com/grafana/mimir/pkg/util/objtools"
 )
 
 type config struct {
-	bucketConfig   objtools.BucketConfig
-	blocksFrom     string
-	inputFile      string
-	includeTenants flagext.StringSliceCSV
-	excludeTenants flagext.StringSliceCSV
-	dryRun         bool
+	bucketConfig       objtools.BucketConfig
+	blocksFrom         string
+	inputFile          string
+	includeTenants     flagext.StringSliceCSV
+	excludeTenants     flagext.StringSliceCSV
+	allowVersionDelete bool
+	dryRun             bool
 }
 
 func (c *config) registerFlags(f *flag.FlagSet) {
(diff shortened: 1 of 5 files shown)


tools/undelete-blocks/main.go
```diff
@@ -493,6 +500,75 @@ func versionToRestore(versions []version, targetSize *int64) (v *version, ok boo
...
+
+		if err := bkt.RestoreVersion(ctx, target.objectName, target.info); err != nil {
+			return errors.Wrapf(err, "failed to restore object %s with version %s", target.objectName, target.info.VersionID)
+		}
+
+		logger.Info("restored an object version", "object", target.objectName, "version", target.info.VersionID)
+		return nil
+	}
+
+	for _, deleteMarker := range restoreInfo.shadowingDeleteMarkers {
```
56quarters, line 554: Since deletion markers are being deleted in a loop, is it safe to run this operation (`undelete-blocks`) multiple times in case one of the deletes fails for some reason? Could we note that somewhere for users?

56quarters (approved):
LGTM

tools/undelete-blocks/main.go
andyasp: > is it safe to run this operation (undelete-blocks) multiple times in case one of the deletes fails for some reason?

If I understand the question you mean `undelete-blocks` as a whole. Yes, you can rerun `undelete-blocks` and it will do the right thing after a partial failure at any point. The same is true for if a block was half-deleted by a compactor. A block is recoverable as long as the required versions are present in the bucket. If nothing needs to be recovered (or if the block can't be recovered because some data is missing) the tool doesn't modify anything.

If it helps to understand it think of each object as a stack of versions. `undelete-blocks` starts by digging for the most recent (non-delete marker) `meta.json` version of a block, then references the files in its JSON to determine what other objects to look for. For each object (including the `meta.json` itself) it detects if the version we want is the visible one (on top of that object's version stack). If a failure was encountered when removing an S3 delete marker it would recognize on a re-run that the version we want isn't visible and do the right thing from there.

andyasp: Added a note in the README at the top to make this clear.

56quarters (approved):

PR #11259, opened 2025-04-18
chore: Remove use of metric namespaces

56quarters:
#### What this PR does

Instead of using the `Namespace` configuration for metrics emitted by Mimir, include the namespace directly in the `Name`. This makes metrics easier to grep for in the codebase.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


seizethedave (approved):
Is this a case of it being pointless if both the namespace and name are literals?

56quarters:
> Is this a case of it being pointless if both the namespace and name are literals?

Yeah, there's no value in specifying both when they're both literals and putting the entire name in `Name` means it's easier to find out where an exact metric is defined.

PR #11178, opened 2025-04-10
Report `artificialDelay` on `ServerTiming` header for push requests

tinitiuset:
#### What this PR does

This PR reports added `artificialDelay` as part of the `ServerTiming` header on push requests.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11265, opened 2025-04-21
Report `artificialDelay` on `ServerTiming` header for push requests

tinitiuset:
#### What this PR does
This PR is an update of https://github.com/grafana/mimir/pull/11178 that better reports added `artificialDelay` as part of the `ServerTiming` header on push requests.

Check out for reviewer's notes on the code:

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -1366,12 +1366,17 @@ func (d *Distributor) outerMaybeDelayMiddleware(next PushFunc) PushFunc {
 
 		userID, userErr := tenant.TenantID(ctx) // Log tenant ID if available.
 		if userErr == nil {
-			// Target delay - time spent processing the middleware chain including the push.
-			// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a negative value.
-			if delay := d.limits.DistributorIngestionArtificialDelay(userID) - d.now().Sub(start); delay > 0 {
-				delay = util.DurationWithJitter(delay, 0.10)
-				pushReq.artificialDelay = delay
-				d.sleep(delay)
+			if delay := d.limits.DistributorIngestionArtificialDelay(userID); delay > 0 {
+				// Target delay - time spent processing the middleware chain including the push.
+				// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a negative value.
+				if delay = delay - d.now().Sub(start); delay > 0 {
+					// Delay is configured but request is taking less than the delay
+					pushReq.artificialDelay = util.DurationWithJitter(delay, 0.10)
+					d.sleep(pushReq.artificialDelay)
+				} else {
+					// Delay is configured but request is already taking longer than the delay
+					pushReq.artificialDelay = 0
+				}
 			}
 			return err
 		}
--- a/pkg/distributor/push.go
+++ b/pkg/distributor/push.go
@@ -373,7 +373,7 @@ func toHTTPStatus(ctx context.Context, pushErr error, limits *validation.Overrid
 }
 
 func addSuccessHeaders(w http.ResponseWriter, delay time.Duration) {
-	if delay > 0 {
+	if delay >= 0 {
 		w.Header().Add("Server-Timing", fmt.Sprintf("artificial_delay;dur=%d", delay))
 	}
 }
--- a/pkg/distributor/push.go
+++ b/pkg/distributor/push.go
@@ -246,8 +246,10 @@ func handler(
 				}
 				level.Error(logger).Log(msgs...)
 			}
-			addHeaders(w, err, r, code, retryCfg, req.artificialDelay)
+			addErrorHeaders(w, err, r, code, retryCfg)
 			http.Error(w, validUTF8Message(msg), code)
+		} else {
+			addSuccessHeaders(w, req.artificialDelay)
 		}
 	})
 }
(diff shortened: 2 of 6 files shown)


pkg/distributor/distributor.go
```diff
@@ -1366,12 +1366,17 @@ func (d *Distributor) outerMaybeDelayMiddleware(next PushFunc) PushFunc {
...
+			if delay := d.limits.DistributorIngestionArtificialDelay(userID); delay > 0 {
+				// Target delay - time spent processing the middleware chain including the push.
+				// If the request took longer than the target delay, we don't delay at all as sleep will return immediately for a negative value.
+				if delay = delay - d.now().Sub(start); delay > 0 {
+					// Delay is configured but request is taking less than the delay
+					pushReq.artificialDelay = util.DurationWithJitter(delay, 0.10)
+					d.sleep(pushReq.artificialDelay)
+				} else {
+					// Delay is configured but request is already taking longer than the delay
+					pushReq.artificialDelay = 0
```
tinitiuset, line 1378: Reporting artificial delay for these cases is very important. 0 artificial delay means request is taking longer than configured artificial delay, worst case scenario.

pkg/distributor/request.go
```diff
@@ -21,14 +21,18 @@ type Request struct {
...
 	getRequest supplierFunc
 
-	request         *mimirpb.WriteRequest
-	err             error
+	request *mimirpb.WriteRequest
+	err     error
+
+	// artificialDelay is the artificial delay for the request.
+	// Negative values are treated as "not set".
 	artificialDelay time.Duration
```
tinitiuset, line 29: `artificialDelay` will be initialized to `-1`, because we want to use `0` as a value, so `-1` would be the equivalent to "not set".

pkg/distributor/request.go
```diff
@@ -21,14 +21,18 @@ type Request struct {
 
 	getRequest supplierFunc
 
-	request         *mimirpb.WriteRequest
-	err             error
+	request *mimirpb.WriteRequest
+	err     error
+
+	// artificialDelay is the artificial delay for the request.
+	// Negative values are treated as "not set".
```
tinitiuset, line 28: `artificialDelay` will be initialized to `-1`, because we want to use `0` as a value, so `-1` would be the equivalent to "not set".

pkg/distributor/request.go
```diff
@@ -21,14 +21,18 @@ type Request struct {
...
 	artificialDelay time.Duration
 }
 
 func newRequest(p supplierFunc) *Request {
 	r := &Request{
-		getRequest: p,
+		getRequest:      p,
+		artificialDelay: -1,
```
tinitiuset, line 35: `artificialDelay` will be initialized to `-1`, because we want to use `0` as a value, so `-1` would be the equivalent to "not set".

pkg/distributor/push.go
```diff
@@ -246,8 +246,10 @@ func handler(
 				}
 				level.Error(logger).Log(msgs...)
 			}
-			addHeaders(w, err, r, code, retryCfg, req.artificialDelay)
+			addErrorHeaders(w, err, r, code, retryCfg)
 			http.Error(w, validUTF8Message(msg), code)
+		} else {
+			addSuccessHeaders(w, req.artificialDelay)
```
dimitarvdimitrov, line 252: nit to invert he `if` and return early if there is no error. That way both branches of the `if` are closer to the condition

pkg/distributor/push.go
```diff
@@ -370,16 +372,18 @@ func toHTTPStatus(ctx context.Context, pushErr error, limits *validation.Overrid
 	return http.StatusInternalServerError
 }
 
-func addHeaders(w http.ResponseWriter, err error, r *http.Request, responseCode int, retryCfg RetryConfig, delay time.Duration) {
+func addSuccessHeaders(w http.ResponseWriter, delay time.Duration) {
+	if delay >= 0 {
+		w.Header().Add("Server-Timing", fmt.Sprintf("artificial_delay;dur=%d", delay))
```
dimitarvdimitrov, line 377: wouldn't this output something like `artificial_delay;dur=500000000` for a 500ms delay? Here's how we do it in the query-frontend https://github.com/grafana/mimir/blob/ee15923e347229c9b2b0d9e95ea9cd4dc2e6883b/pkg/frontend/transport/handler.go#L569-L570

tinitiuset: I'll adjust this for consistency, but do you think it's worth converting to milliseconds? When the header is read it will have to be reconverted to `time.Duration` most probably.

pkg/distributor/push.go
tinitiuset: Agree. Thank you

pkg/distributor/push.go
dimitarvdimitrov: nanosecond looks like a bit too much detail for us to care - it's much more granular than other operations, so granularity beyond, say, 0.5ms might just be gibberish. And it's not easy for humans to understand.

dimitarvdimitrov:
this made me realize that this PR doesn't document the header anywhere. Can you add it to the API reference and add a changelog entry too?

dimitarvdimitrov:
> this made me realize that this PR doesn't document the header anywhere. Can you add it to the API reference and add a changelog entry too?

Martin pointed me to a [previous discussion](https://github.com/grafana/mimir/pull/11178#discussion_r2038924934) where Marco asked to not document this feature (yet?). We'll do the same now too

dimitarvdimitrov (approved):
thanks

PR #11271, opened 2025-04-21
general cleanup: move all identical prom sample impls to test util

francoposa:
#### What this PR does

Lots of things in tests need a type that satisfies prometheus' `chunks.Sample` interface, and prometheus does not provide any public implementation.
We ended up with about 10 identical implementations, each private to a separate test file.

This makes them all one in our test utils package

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/test/tsdb.go
+++ b/pkg/util/test/tsdb.go
@@ -14,6 +14,10 @@ type Series struct {
 	samples []Sample
 }
 
+func NewSeries(lbls labels.Labels, samples []Sample) Series {
+	return Series{lbls, samples}
+}
+
 type Sample struct {
 	t  int64
 	v  float64
@@ -21,7 +25,7 @@ type Sample struct {
 	fh *histogram.FloatHistogram
 }
 
-func NewSample(t int64, v float64, h *histogram.Histogram, fh *histogram.FloatHistogram) chunks.Sample {
+func NewSample(t int64, v float64, h *histogram.Histogram, fh *histogram.FloatHistogram) Sample {
 	return Sample{t, v, h, fh}
 }
 
--- a/pkg/util/test/tsdb.go
+++ b/pkg/util/test/tsdb.go
@@ -14,6 +14,10 @@ type Series struct {
 	Samples []Sample
 }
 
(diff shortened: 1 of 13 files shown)


pkg/util/test/tsdb.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type Sample struct {
+	TS        int64
+	Val       float64
+	Hist      *histogram.Histogram
+	FloatHist *histogram.FloatHistogram
+}
+
+func (s Sample) T() int64                      { return s.TS }
```
56quarters, line 24: Can we add a few comments here or otherwise document the semantics? Are float values, histograms, and float histograms mutually exclusive in a sample? Could we add constructors to enforce this?

56quarters (approved):
Nice cleanup. Since we're enshrining this code as the correct way to create samples in tests, maybe we could clean it up a bit. Otherwise, this looks good to me.

pkg/util/test/tsdb.go
francoposa: Each sample can only have one of each value type.
I think I would lean towards comments - I started this with a constructor, but it took up extra space in tests providing nils everywhere like `util_test.NewSample(0, 1.0, nil, nil)` vs. `util_test.Sample{TS: 0, Val: 1.0}`.

francoposa: The other option is something like this - fine with a panic here as it's a test util. Having the option to return an error would make all the tests a lot worse to write:
```golang
func NewSample(ts int64, val any) Sample {
	switch v := val.(type) {
	case float64:
		return Sample{TS: ts, Val: v}
	case *histogram.Histogram:
		return Sample{TS: ts, Hist: v}
	case *histogram.FloatHistogram:
		return Sample{TS: ts, FloatHist: v}
	default:
		panic("unsupported sample value type")
	}
}
```

Part of my problem with this is that type switches are not super cheap, and is an artificial thing to introduce into test paths that include benchmarking.

56quarters (approved):

PR #10540, opened 2025-01-30
Jsonnet: allow to not configure topologySpreadConstraints at all

pracucci:
#### What this PR does

Allow to not configure `topologySpreadConstraints` by setting the following configuration options to a negative value:
  * `distributor_topology_spread_max_skew`
  * `query_frontend_topology_spread_max_skew`
  * `querier_topology_spread_max_skew`
  * `ruler_topology_spread_max_skew`
  * `ruler_querier_topology_spread_max_skew`

```diff
diff /Users/marco/workspace/src/github.com/grafana/mimir/operations/mimir-tests/test-all-components-generated.yaml /Users/marco/workspace/src/github.com/grafana/mimir/operations/mimir-tests/test-all-components-without-pod-topology-constraints-generated.yaml

563,569d562
<       topologySpreadConstraints:
<       - labelSelector:
<           matchLabels:
<             name: distributor
<         maxSkew: 1
<         topologyKey: kubernetes.io/hostname
<         whenUnsatisfiable: ScheduleAnyway
660,666d652
<       topologySpreadConstraints:
<       - labelSelector:
<           matchLabels:
<             name: querier
<         maxSkew: 1
<         topologyKey: kubernetes.io/hostname
<         whenUnsatisfiable: ScheduleAnyway
739,745d724
<       topologySpreadConstraints:
<       - labelSelector:
<           matchLabels:
<             name: query-frontend
<         maxSkew: 1
<         topologyKey: kubernetes.io/hostname
<         whenUnsatisfiable: ScheduleAnyway
905,911d883
<       topologySpreadConstraints:
<       - labelSelector:
<           matchLabels:
<             name: ruler
<         maxSkew: 1
<         topologyKey: kubernetes.io/hostname
<         whenUnsatisfiable: ScheduleAnyway
```

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11274, opened 2025-04-21
[helm] enable topologySpreadConstrains on all components by default

TheRealNoob:
#### What this PR does

Enables `topologySpreadConstraints` on all components by default.  It is being enabled on the following
1. overrides_exporter
2. chunks-cache
3. index-cache
4. metadata-cache
5. results-cache
6. admin-cache

None

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo:
With regards the `topologySpreadConstraints`, mimir Jsonnet bundle, historically, set them by default. Not so long ago we'd discovered that this led to problems in large clusters, that run under AWS Karpenter (related to specifics of how it treats the best effort constraints `ScheduleAnyway`). To mitigate that in a backwards compatible way, the Jsonnet bundle now has a special semantics for the `maxSkew` config (https://github.com/grafana/mimir/pull/10540).

Long story short, I don't think we should do something by default, until there is a compelling explanation for why doing so on the side of the user, who control the underlying infra, isn't a better option.

tacole02 (approved):

narqo (changes requested):
Just so it was clear: I'm not yet convinced the Helm chart should accept these changes. We need more details about the problem, the PR's solving.

TheRealNoob:
I opened the PR because I thought having it enabled by default was a sensible default - harmless if `replicas=1`, beneficial if `>1`.  However if there is evidence that's harmful in select cases then I'm fine closing this.  Thank you for clarifying @narqo.

PR #11283, opened 2025-04-22
MQE: Make memory consumption tracker concurrency safe

56quarters:
#### What this PR does

This change makes it safe to use `MemoryConsumptionTracker` from multiple goroutines. This is a prerequisite to tracking memory used by chunks loaded from store-gateways and distributors which is done in multiple goroutines in parallel.

N/A

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/limiting/memory_consumption.go
+++ b/pkg/streamingpromql/limiting/memory_consumption.go
@@ -3,6 +3,8 @@
 package limiting
 
 import (
+	"sync"
+
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/grafana/mimir/pkg/util/limiter"
@@ -11,20 +13,23 @@ import (
 // MemoryConsumptionTracker tracks the current memory utilisation of a single query, and applies any max in-memory bytes limit.
 //
 // It also tracks the peak number of in-memory bytes for use in query statistics.
-//
-// It is not safe to use this type from multiple goroutines simultaneously.
 type MemoryConsumptionTracker struct {
-	MaxEstimatedMemoryConsumptionBytes     uint64
-	CurrentEstimatedMemoryConsumptionBytes uint64
-	PeakEstimatedMemoryConsumptionBytes    uint64
+	maxEstimatedMemoryConsumptionBytes     uint64
+	currentEstimatedMemoryConsumptionBytes uint64
+	peakEstimatedMemoryConsumptionBytes    uint64
 
 	rejectionCount        prometheus.Counter
 	haveRecordedRejection bool
+
+	// mtx protects all mutable state of the memory consumption tracker. We use a mutex
+	// rather than atomics because we only want to adjust the memory used after checking
+	// that it would not exceed the limit.
+	mtx sync.Mutex
 }
 
 func NewMemoryConsumptionTracker(maxEstimatedMemoryConsumptionBytes uint64, rejectionCount prometheus.Counter) *MemoryConsumptionTracker {
 	return &MemoryConsumptionTracker{
-		MaxEstimatedMemoryConsumptionBytes: maxEstimatedMemoryConsumptionBytes,
+		maxEstimatedMemoryConsumptionBytes: maxEstimatedMemoryConsumptionBytes,
 
 		rejectionCount: rejectionCount,
 	}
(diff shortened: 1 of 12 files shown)


jhalterman (approved):

charleskorn (approved):
LGTM, one non-blocking suggestion below.

I'd be interested to see the performance impact of this change, but I'm not sure we have much choice apart from trying atomics if this turns out to be expensive.

pkg/streamingpromql/limiting/memory_consumption.go
```diff
@@ -11,20 +13,23 @@ import (
...
+	maxEstimatedMemoryConsumptionBytes     uint64
+	currentEstimatedMemoryConsumptionBytes uint64
+	peakEstimatedMemoryConsumptionBytes    uint64
 
 	rejectionCount        prometheus.Counter
 	haveRecordedRejection bool
+
+	// mtx protects all mutable state of the memory consumption tracker. We use a mutex
+	// rather than atomics because we only want to adjust the memory used after checking
+	// that it would not exceed the limit.
```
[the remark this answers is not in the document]
56quarters: I'll work on benchmarking no concurrency vs mutex vs atomics today and seeing if it's possible to do this with atomics at all.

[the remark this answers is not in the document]
56quarters: I actually can't realistically compare these. Using no locking (obviously) runs into race conditions and the logic we use won't work with atomics. `IncreaseMemoryConsumption()` reads the currently used memory, compares it to the max, and then either returns an error or increases the currently used memory. It's possible for the currently used memory amount to change in either direction before it's increased. FWIW the benchmarks with locking don't seem too bad compared to not using locking at all (~40ns for locking increment/decrement vs ~5 for no locking).

Locking:
```
$ go test -run=XXX -benchmem -bench=BenchmarkMemoryConsumptionTracker ./pkg/streamingpromql/limiting/
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/streamingpromql/limiting
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkMemoryConsumptionTracker/no_limits_single_threaded-16          36335068                35.70 ns/op            0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/with_limits_single_threaded-16        33024090                35.12 ns/op            0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/no_limits_multiple_threads-16         15300205                85.11 ns/op            0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/with_limits_multiple_threads-16       14355288                89.51 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/grafana/mimir/pkg/streamingpromql/limiting   5.060s

```

vs an atomics implementation that I was not able to get fully functional but I think is close enough to demonstrate the performance (single atomic operation for most increment/decrements):

```
$ go test -run=XXX -benchmem -bench=BenchmarkMemoryConsumptionTracker ./pkg/streamingpromql/limiting/
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/streamingpromql/limiting
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkMemoryConsumptionTracker/no_limits_single_threaded-16          84930708                15.61 ns/op            0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/with_limits_single_threaded-16        84038815                14.80 ns/op            0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/no_limits_multiple_threads-16          9757600               122.6 ns/op             0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/with_limits_multiple_threads-16        8812524               123.1 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/grafana/mimir/pkg/streamingpromql/limiting   4.868s
```

vs no locking:

```
$ go test -run=XXX -benchmem -bench=BenchmarkMemoryConsumptionTracker ./pkg/streamingpromql/limiting/
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/streamingpromql/limiting
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkMemoryConsumptionTracker/no_limits_single_threaded-16          311093976                3.876 ns/op           0 B/op          0 allocs/op
BenchmarkMemoryConsumptionTracker/with_limits_single_threaded-16        273341175                4.589 ns/op           0 B/op          0 allocs/op
panic: Estimated memory consumption of this query is negative. This indicates something has been returned to a pool more than once, which is a bug.

goroutine 54 [running]:
github.com/grafana/mimir/pkg/streamingpromql/limiting.(*MemoryConsumptionTracker).DecreaseMemoryConsumption(...)
        /home/pillin/programming/git_repos/mimir/pkg/streamingpromql/limiting/memory_consumption.go:60
github.com/grafana/mimir/pkg/streamingpromql/limiting.BenchmarkMemoryConsumptionTracker.func3.1()
        /home/pillin/programming/git_repos/mimir/pkg/streamingpromql/limiting/memory_consumption_test.go:157 +0xd4
created by github.com/grafana/mimir/pkg/streamingpromql/limiting.BenchmarkMemoryConsumptionTracker.func3 in goroutine 47
        /home/pillin/programming/git_repos/mimir/pkg/streamingpromql/limiting/memory_consumption_test.go:151 +0xbf
exit status 2
FAIL    github.com/grafana/mimir/pkg/streamingpromql/limiting   2.481s
FAIL

```

Based on this, I think mutexes are the right call.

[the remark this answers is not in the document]
charleskorn: Sounds good to me, thanks for adding the benchmarks.

pkg/streamingpromql/limiting/memory_consumption_test.go
```diff
@@ -117,3 +119,81 @@ func createRejectedMetric() (*prometheus.Registry, prometheus.Counter) {
 
 	return reg, metric
 }
+
+func BenchmarkMemoryConsumptionTracker(b *testing.B) {
+	b.Run("no limits single threaded", func(b *testing.B) {
+		l := NewMemoryConsumptionTracker(0, nil)
+		for i := 0; i < b.N; i++ {
```
charleskorn, line 126: [nit] Could use the new `b.Loop()` method here and below

pkg/streamingpromql/limiting/memory_consumption_test.go
```diff
@@ -117,3 +119,81 @@ func createRejectedMetric() (*prometheus.Registry, prometheus.Counter) {
...
+
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			for run.Load() {
+				_ = l.IncreaseMemoryConsumption(1024)
+				l.DecreaseMemoryConsumption(1024)
+			}
+		}()
```
charleskorn, line 158: If we don't use `b.Loop()` below, we should probably call `b.ResetTimer()` after all this setup work is done. Same goes for the test case below.

pkg/streamingpromql/limiting/memory_consumption_test.go
56quarters: We still declare `go 1.23.7` in `go.mod` so the linter warns about using `b.Loop()`. I've added timer resets to the loops.

pkg/streamingpromql/limiting/memory_consumption_test.go
56quarters: Added calls to `b.ResetTimer()`

PR #11285, opened 2025-04-22
MQE: Include chunks loaded from store-gateway in memory estimate

56quarters:
#### What this PR does

This increases the tracked memory consumption when chunks streamed
from store-gateways are loaded into memory as part of series iterators.
Memory consumption is decreased when the iterators are exhausted.
This decrease is optimistic since there is no guarantee that iterators
are completely consumed during a query (but they _usually_ are unless
the query is canceled or aborted).

Related #11283

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/block_streaming.go
+++ b/pkg/querier/block_streaming.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/gogo/status"
 	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/model/histogram"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
@@ -31,13 +32,14 @@ import (
 
 // Implementation of storage.SeriesSet, based on individual responses from store client.
 type blockStreamingQuerierSeriesSet struct {
-	series       []labels.Labels
-	streamReader chunkStreamReader
+	series        []labels.Labels
+	streamReader  chunkStreamReader
+	memoryTracker memoryConsumptionTracker
 
 	// next response to process
 	nextSeriesIndex int
 
-	currSeries storage.Series
+	currSeries *blockStreamingQuerierSeries
 
 	// For debug logging.
 	chunkInfo     *chunkinfologger.ChunkInfoLogger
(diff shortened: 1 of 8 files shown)


lamida (commented):
The approach seems reasonable and I can't think another other alternative ways. Let's see what @charleskorn and @jhesketh think.

56quarters:
> The approach seems reasonable and I can't think another other alternative ways. Let's see what @charleskorn and @jhesketh think.

This approach won't actually work as written due to the order in which `SeriesSet` and `Series` methods are called. I'm going to try a few things. It may be the case that we need to add some lifecycle methods to Prometheus interfaces. I imagine that'd be hugely disruptive to other Prometheus consumers and unlikely to be accepted. This can be a last resort

charleskorn:
> > The approach seems reasonable and I can't think another other alternative ways. Let's see what @charleskorn and @jhesketh think.
> 
> This approach won't actually work as written due to the order in which `SeriesSet` and `Series` methods are called. I'm going to try a few things. ~It may be the case that we need to add some lifecycle methods to Prometheus interfaces. I imagine that'd be hugely disruptive to other Prometheus consumers and unlikely to be accepted. That can be a last resort~. I'm not actually sure that would help.

We may be able to piggy-back on the changes in #11136: we have a similar problem there where we need to return the underlying gRPC message buffers to a pool when we're done with the labels or chunks, so we could increase / decrease the current memory consumption at the same point where we acquire / release the buffers. This would also be close to the point at which the buffers (and therefore the chunks within them) are eligible to be garbage collected, which would ensure that the limit roughly tracks the amount of heap that the query is keeping alive.

56quarters:
> We may be able to piggy-back on the changes in #11136: we have a similar problem there where we need to return the underlying gRPC message buffers to a pool when we're done with the labels or chunks, so we could increase / decrease the current memory consumption at the same point where we acquire / release the buffers. This would also be close to the point at which the buffers (and therefore the chunks within them) are eligible to be garbage collected, which would ensure that the limit roughly tracks the amount of heap that the query is keeping alive.

That definitely looks helpful for this work. I'll work on finalizing #11283 for now and come back to this after taking a more in-depth look at #11136.

56quarters:
> > We may be able to piggy-back on the changes in #11136: we have a similar problem there where we need to return the underlying gRPC message buffers to a pool when we're done with the labels or chunks, so we could increase / decrease the current memory consumption at the same point where we acquire / release the buffers. This would also be close to the point at which the buffers (and therefore the chunks within them) are eligible to be garbage collected, which would ensure that the limit roughly tracks the amount of heap that the query is keeping alive.
> 
> That definitely looks helpful for this work. I'll work on finalizing #11283 for now and come back to this after taking a more in-depth look at #11136.

I think that will work: use the lifecycle of `lastMessage` in the `storeGatewayStreamReader` to increment and decrement memory used. I'll have to try it out a bit to ensure we don't leak anything after either #11283 or #11136 is merged to main.

56quarters:
Superceded by https://github.com/grafana/mimir/pull/11453

PR #11286, opened 2025-04-23
ingest: cheaper hash function for parallelStorageShards

bboreham:
#### What this PR does

Save the work of converting LabelAdapters to Labels just to compute a hash. This was about 6% of all garbage in one case.

There is no particular benefit to match the hash function used by TSDB; its main stripes are split by unique ID not label hash.
(The main ingester 'push' function will lock first by hash then by unique ID for each series, and there is only one set of stripe locks, so they will tend to bounce around anyway.)

NA, just general performance.

#### Checklist

- NA Tests updated.
- NA Documentation added.
- NA `CHANGELOG.md` - not user visible.


pracucci (commented):
I agree that using the same hashing function of TSDB doesn't yield the expected behaviour. I left a question.

dimitarvdimitrov (approved):

PR #11287, opened 2025-04-23
Mixin: fix cluster_namespace_deployment:actual_replicas:count recording rule when there's a mix on single-zone and multi-zone deployments

pracucci:
#### What this PR does

Today we were migrating a Mimir cluster from single-AZ to multi-AZ deployment. During the migration the recording rule `cluster_namespace_deployment:actual_replicas:count` failed with the error `vector cannot contain metrics with the same labelset`. The problem is that if there's a Deployment with the name `xxx` and other Deployments with the name `xxx-zone-[abc]`, the query will fail after the `label_replace()` because of clashing series.

This PR fixes it.

#### Manual tests

I manually tested it.

Before:

```
sum by (cluster, namespace, deployment) (
  label_replace(
    kube_deployment_spec_replicas,
    # The question mark in "(.*?)" is used to make it non-greedy, otherwise it
    # always matches everything and the (optional) zone is not removed.
    "deployment", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
  )
)
or
sum by (cluster, namespace, deployment) (
  label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*?)(?:-zone-[a-z])?")
)
```

After:

```
sum by (cluster, namespace, deployment) (
  label_replace(
    sum by (cluster, namespace, deployment_without_zone) (
      label_replace(
        kube_deployment_spec_replicas,
        # The question mark in "(.*?)" is used to make it non-greedy, otherwise it
        # always matches everything and the (optional) zone is not removed.
        "deployment_without_zone", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
      )
    )
    or
    sum by (cluster, namespace, deployment_without_zone) (
      label_replace(kube_statefulset_replicas, "deployment_without_zone", "$1", "statefulset", "(.*?)(?:-zone-[a-z])?")
    ),
    "deployment", "$1", "deployment_without_zone", "(.*)"
  )
)
```

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@
 * [ENHANCEMENT] Alerts: Improve `MimirIngesterFailsToProcessRecordsFromKafka` to not fire during forced TSDB head compaction. #11006
 * [ENHANCEMENT] Alerts: Add alerts for invalid cluster validation labels. #11255 #11282
 * [BUGFIX] Dashboards: fix "Mimir / Tenants" legends for non-Kubernetes deployments. #10891
+* [BUGFIX] Recording rules: fix `cluster_namespace_deployment:actual_replicas:count` recording rule when there's a mix on single-zone and multi-zone deployments. #11287
 
 ### Jsonnet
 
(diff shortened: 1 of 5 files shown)


56quarters (approved):
LGTM

PR #11288, opened 2025-04-23
Alerts: Enhance MimirRolloutStuck so it ignores kube-state-metrics instances

aknuds1:
#### What this PR does

Enhance the `MimirRolloutStuck` alert, so it checks whether rollout groups as a whole are changing (i.e. not stuck). Previously, it would look at rollout groups per kube-state-metrics instance. The latter might cause the alert to fire if the same rollout group has multiple kube-state-metrics instances, so that instances covering replicas that are done updating appear unchanging.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@
 * [ENHANCEMENT] Alerts: Improve `MimirIngesterFailsToProcessRecordsFromKafka` to not fire during forced TSDB head compaction. #11006
 * [ENHANCEMENT] Alerts: Add alerts for invalid cluster validation labels. #11255 #11282
 * [BUGFIX] Dashboards: fix "Mimir / Tenants" legends for non-Kubernetes deployments. #10891
+* [BUGFIX] Alerts: Enhance the `MimirRolloutStuck` alert, so it checks whether rollout groups as a whole (and not spread across instances) are changing or stuck. #11288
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
 * [ENHANCEMENT] Dashboards: Add panels to the `Mimir / Tenants` and `Mimir / Top Tenants` dashboards showing the rate of gateway requests. #10978
 * [ENHANCEMENT] Alerts: Improve `MimirIngesterFailsToProcessRecordsFromKafka` to not fire during forced TSDB head compaction. #11006
 * [ENHANCEMENT] Alerts: Add alerts for invalid cluster validation labels. #11255 #11282
+* [ENHANCEMENT] Alerts: Enhance the `MimirRolloutStuck` alert, so it checks whether rollout groups as a whole (and not spread across instances) are changing or stuck. #11288
```
colega, line 62: I would say this is a bugfix: the alert has a false positive when some of the kubernetes metrics are exposed by different ksm-metrics instances.

aknuds1: OK, will change it.

aknuds1: Done.

colega (approved):

PR #11300, opened 2025-04-24
CI: retry integration and unit tests

dimitarvdimitrov:
#### What this PR does

this helps with flaky tests

related to https://github.com/grafana/mimir-squad/issues/2944

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/.github/workflows/test-build-deploy.yml
+++ b/.github/workflows/test-build-deploy.yml
@@ -170,10 +170,18 @@ jobs:
         run: |
           mkdir -p /go/src/github.com/grafana/mimir
           ln -s $GITHUB_WORKSPACE/* /go/src/github.com/grafana/mimir
-      - name: Run Tests
-        run: |
-          echo "Running unit tests (group ${{ matrix.test_group_id }} of ${{ matrix.test_group_total }}) with Go version: $(go version)"
-          ./.github/workflows/scripts/run-unit-tests-group.sh --index ${{ matrix.test_group_id }} --total ${{ matrix.test_group_total }}
+      # We use retry logic for tests to handle flaky tests that might fail intermittently.
+      # The retry action will run the test command up to max_attempts times if it fails.
+      # This helps reduce CI failures caused by temporary issues, timing problems, or race conditions.
+      - name: Run Tests with retry
+        uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3.0.2
+        with:
+          timeout_minutes: 180 # 3 hours so that we rely on the timeout in the makefile
+          max_attempts: 2
+          retry_wait_seconds: 0
+          command: |
+            echo "Running unit tests (group ${{ matrix.test_group_id }} of ${{ matrix.test_group_total }}) with Go version: $(go version)"
+            ./.github/workflows/scripts/run-unit-tests-group.sh --index ${{ matrix.test_group_id }} --total ${{ matrix.test_group_total }}
 
   test-docs:
     uses: ./.github/workflows/test-docs.yml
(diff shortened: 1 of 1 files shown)


colega:
I would vote 100% against this. Please let's dont.

Retrying flaky tests is just hiding the real problem and making a future problem bigger.

If there's a flaky test, let's just delete it, or whatever.

dimitarvdimitrov:
if we delete a test, then we never run it again. So if it were to catch a bug in a month's time, we wouldn't know because we deleted it.  How is that not hiding the problem too?Retrying is a middle ground for still having tests without blocking people.

What if I add some automation to detect who was the last editor of a test, create a github issue and tag the person? As long as the test is flaky, we can keep pinging the person on the issue. In the meantime we retry the test.

Alternatively, maybe we can also come up with some way of running newly added tests in a PR many times before allowing a merge. That way we may flaky tests as before they are added to `main`.

colega:
> If we delete a test, then we never run it again. So if it were to catch a bug in a month's time, we wouldn't know because we deleted it. How is that not hiding the problem too?

Yes sure, that's why we should fix them, but retrying isn't fixing them. And we aren't even retrying the failed test we're retrying just all of them, which is worse.

> What if I add some automation to detect who was the last editor of a test, create a github issue and tag the person? As long as the test is flaky, we can keep pinging the person on the issue. In the meantime we retry the test.

Automation or just an engineer pointing with a finger, but how will we know now which tests are flaky?

Can we at least export metrics on the flaky tests and create issues for those?

dimitarvdimitrov:
> Yes sure, that's why we should fix them, but retrying isn't fixing them. And we aren't even retrying the failed test we're retrying just all of them, which is worse.

i get that this isn't a complete solution, but i don't get how retries are a regression either. Is you concern that this way we'll only have more flaky tests?



> Automation or just an engineer pointing with a finger, but how will we know now which tests are flaky?

I made a [dashboard](https://ops.grafana-ops.net/d/b93d10bb-8a89-4944-a39e-75a112db706a/golang-failing-tests?orgId=1&from=now-30d&to=now&timezone=utc&var-query0=&var-repo_name=&var-repo=grafana-mimir&var-parent_test_name=&var-Filters=) which tracks how many branches a test has failed on. More distinct branches == more flaky. Every week we can pick the top 3 from the dashboard, find the last 2 committers of the test function, open an issue, tag the 2 committers. If we keep tagging them every week, maybe they fix it?

colega:
> Is you concern that this way we'll only have more flaky tests?

One concern is that we'll have more flaky tests, it will be easier to introduce one since even new tests are retried here. Another one is based on the assumption that their number will grow, eventually you'll have to increase the retries number, but that increase will probably be exponential compared to the number of flaky tests because of how probabilities and combinatorial works.

> I made a [dashboard](https://ops.grafana-ops.net/d/b93d10bb-8a89-4944-a39e-75a112db706a/golang-failing-tests?orgId=1&from=now-30d&to=now&timezone=utc&var-query0=&var-repo_name=&var-repo=grafana-mimir&var-parent_test_name=&var-Filters=) which tracks how many branches a test has failed on.

TIL (it would be nice to have this data public)

> More distinct branches == more flaky. 

With retries we can clearly identify flaky tests if they fail and then they succeed on the retries. IIRC go1.24 improved the json format of test runs, so maybe we could do some jq-foo magic?

dimitarvdimitrov:
> With retries we can clearly identify flaky tests if they fail and then they succeed on the retries. IIRC go1.24 improved the json format of test runs, so maybe we could do some jq-foo magic?

I access the data in loki, so i can't do much analysis beyond transforming a single log line and running a metrics query.

I think the multi-branch thing is still good enough. If a single failure per test per week is still annoying, we can extend the window to something like the last 2 weeks or 30 days.

colega:
Discussed offline, I'm not blocking on merging this, given you're going to follow up and pursue people who added the flaky tests.

narqo:
> I think the multi-branch thing is still good enough.

As an alternative, we could only do the retries for PRs, leaving the normal execution in the main and release branches. That is, if the PRs are green but the tests in the `main` branch keep failing randomly, this is the signal of a "healthy" repo, that can be tracked, similar to TCO and CVEs, or other signals.

iirc, this is _kinda_ how [Uber's Testopedia](https://www.uber.com/en-DE/blog/flaky-tests-overhaul/) came about.

dimitarvdimitrov:
> As an alternative, we could only do the retries for PRs, leaving the normal execution in the main and release branches. That is, if the PRs are green but the tests in the `main` branch keep failing randomly, this is the signal of an "un-healthy" project.

do you mean we'd retry only tests added by PRs? And tests that fail, but are not changed in this PR would not be retried and we'd alert on them, etc?

narqo:
> do you mean we'd retry only tests added by PRs?

No. What I meant was that this workflow currently runs on,

- a push to branches ("main", "release-*"), or
- when a new PR is opened/updated ([ref](https://github.com/grafana/mimir/blob/17c718c387bbc58c42278b5f8d2d91711271c081/.github/workflows/test-build-deploy.yml#L2-L10)).

Having a retry is mainly valuable for the PRs, because an unrelated flake distract the PR's review, and is blocking for a human. But we don't actively care about flakes that happen, when the workflow is running against the `main` branch (_I think, we don't_).

What I suggested above was to NOT do the retries in the main and release branches. Only do the retries for the PRs.

dimitarvdimitrov:
> What I suggested above was to NOT do the retries in the main and release branches. Only do the retries for the PRs.

now i understand what you said. But what's the benefit of not retrying on the main & release branches?

You started this thread from my "I think the multi-branch thing is still good enough." What I meant by "the multi-branch thing" is how we determine if a test is flaky - if it fails on more than 1 branch, then it is flaky. The more branches if fails on, the flakier. Are we talking about the same thing?

PR #11301, opened 2025-04-24
fix(parquet reader): fix handling for matcher on unexisting column

npazosmendez:


francoposa (approved):

pkg/storage/parquet/reader.go
```diff
@@ -246,14 +246,17 @@ func (pr *ParquetReader) ScanRows(ctx context.Context, rows [][]int64, full bool
 	span, _ := opentracing.StartSpanFromContext(ctx, "ParquetReader.ScanRows")
 	defer span.Finish()
 
-	mtx := sync.Mutex{}
-	results := make([][]int64, len(pr.f.RowGroups()))
-
 	ci, ok := pr.columnIndexMap[m.Name]
 	if !ok {
-		return results, nil
+		if m.Matches("") {
```
MichaHoffmann, line 251: if a column does not exist then any "not" matcher should be ignored here; not only if it matches empty.

`i_dont_exist!="some_val"` does not constrain the result set

jesusvazquez: I found this bug by running `sum(rate(container_cpu_usage_seconds_total{namespace="mimir-dev-11", foo!="bar"}[$__rate_interval]))` query. Will try to submit a PR with the fix.

jesusvazquez: https://github.com/grafana/mimir/pull/11319

PR #11303, opened 2025-04-24
Compactor: Use Stable Set of Tenants on Each BlocksCleaner Run

dmwilson-grafana:
#### What this PR does

This PR fixes an issue where `MimirBucketIndexNotUpdated` can fire even though the index has been updated within the alert threshold (typically 2100s). This most often occurs when multiple pods are emitting a metric for a tenant and the pod that most recently updated the index is terminated.

As implemented, `BlocksCleaner` checks ownership multiple times per execution. 

* `instrumentBucketIndexUpdate` Scans object storage, finds owned tenants, reads the current bucket index, and writes a metric, `cortex_bucket_index_last_successful_update_timestamp_seconds`.

* `runCleanup`  Re-scans object storage and calls `ownUser(id)` before updating the index for any given tenant. 

The set of tenants a pod owns can change between these three steps. This PR makes each run use a single set of users sourced from a single call to `refreshOwnedUsers`.

As implemented, the following scenario would cause alerts to fire. Assume we're starting at `t=1600`. 

1. `c-1` runs `instrumentBucketIndexUpdate`, owns `tenant_1` and writes `upd_timestamp{user="tenant_1", pod="c-1"} 0`.
2. `c-2` runs `instrumentBucketIndexUpdate` and also writes `upd_timestamp{user="tenant_1", pod="c2"} 0`.
5. `c-1` and `c-2` start `runCleanup`. Both pods agree the tenant belongs to`c-2`.
6. `c-2` updates the index for `tenant_1` and writes the metric `upd_timestamp{user="tenant_1", pod="c-2"} 1600`.
7. At time t=2000,`c-2` goes down. Now `upd_timestamp{user="tenant_1", pod="c-1"} 0` is the only metric available to the alert. 
8. Multiple Cases:
* If `c-1` refreshes users, takes ownership of `tenant_1`, and updates the index before t=2100, then no alert will fire.
*  if `c-2` recovers, runs `instrumentBucketIndexUpdate`, reads the index from object storage, and writes `upd_timestamp{user="tenant_1", pod="c2"} 1600`, then no alert will fire.
*  If t=2100 passes before 6a. or 6b. occur, `MimirBucketIndexNotUpdated` will fire even though the index was updated just 500s ago.

By requiring a stable set of users, we ensure that any BlocksCleaner that emits a metric will update the index for that tenant. With this change, we don't need 6a. or 6b. to happen to prevent an incorrect firing, because `c-1` updates the index and writes `upd_timestamp{user="tenant_1", pod="c-1"} 1600` in step 4.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/compactor/blocks_cleaner.go
```diff
@@ -189,14 +189,29 @@ func (c *BlocksCleaner) runCleanup(ctx context.Context, async bool) {
...
+		users     []string
+		isDeleted map[string]bool
+		err       error
+	)
+
+	// Use a single set of tenants while BlocksCleaner is in Starting state. This ensures that a BlockCleaner updates
+	// the bucket-index.json for all tenants discovered on startup. When multiple BlockCleaners come up at once,
+	// tenants can be updated by multiple BlocksCleaners on startup. However, ownership will be delegated to a single
+	// BlocksCleaner instance once the sharding ring is stable.
+	if c.State() == services.Starting && len(c.lastOwnedUsers) > 0 {
```
dimitarvdimitrov, line 202: why can't we do this for all states, not just starting?

dmwilson-grafana: I think keeping the set of users stable on start fixes the issue we saw while minimizing extra work. I'm assuming that: 

1. Keeping a stable set of users throughout a `cleanAllUsers` call will  lead to duplicate work more often.
2. Multiple compactors will be coming up during `services.Starting`, but not so much during a `cleanAllUsers` when the cleaner is in `services.Running`.

I suppose that if compactor_0 comes up at t=0, waits 15m, begins running a cleanAllUsers at t=900 while compactors_1...compactor_5 also come up at t=900, then we could have the same issue.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 * [BUGFIX] Query-frontend: Fix an issue where transient errors could be inadvertently cached. #11198
 * [BUGFIX] Ingester: read reactive limiters should activate and deactivate when the ingester changes state. #11234
 * [BUGFIX] Query-frontend: Fix an issue where errors from date/time parsing methods did not include the name of the invalid parameter. #11304
+* [BUGFIX] Compactor: Fix issue where `MimirBucketIndexNotUpdated` can fire even though the index has been updated within the alert threshold. #11303
```
tacole02, line 56:
```suggestion
* [BUGFIX] Compactor: Fix issue where `MimirBucketIndexNotUpdated` can fire even though the index has been updated in the alert threshold. #11303
```

tacole02 (approved):

pkg/compactor/blocks_cleaner.go
dimitarvdimitrov: i was more concerned about keeping the code simpler. Right now it's hard to understand where the list of users is coming from. 

WDYT about passing the list of users down to `runCleanup` and to `instrumentBucketIndexUpdate`? We'd call `refreshownedUsers` in `ticker` and `starting` and then pass down the tenants (maybe as a struct instead of a slice and a map? idk). That way you can follow the call chain to see where list of users is derived and don't have to think about sharing memory and order of calls

tacole02 (approved):

pkg/compactor/blocks_cleaner.go
```diff
@@ -36,6 +36,8 @@ import (
 const (
 	defaultDeleteBlocksConcurrency       = 16
 	defaultGetDeletionMarkersConcurrency = 16
+	cleanUsersServiceStarting            = "clean_up_users_starting"
```
dimitarvdimitrov, line 39:
```suggestion
	cleanUsersServiceStarting            = "clean_up_users_during_startup"
```

dimitarvdimitrov (approved):
nice one, thanks

PR #11306, opened 2025-04-24
Grafana Alertmanager: Prevent invalid config and use correct template validation on Apply

JacobsonMT:
A combination of two issues was causing a bug when attempting to use the new template function introduced with [custom webhook payloads](https://github.com/grafana/mimir/pull/11195) in a Grafana AM tenant:
 - The `SetUserGrafanaConfig` API endpoint was not validating templates (or receivers).
 - `ApplyConfig` was validating against Cloud AM template functions for Grafana AM tenants (though it was correctly using Grafana AM template functions to create the integrations).

Together, these allowed a few ways to get invalid config persisted to the grafana config bucket. This invalid config would not successfully Apply so would cause the config to not sync until the invalid config was corrected.

Additionally, since `ApplyConfig` was using the wrong template validation, even valid config could cause the above sync issue if attempting to use template functions available in Grafana AM but not in Cloud AM (such as the new webhook payload functions).

This PR:
- Does some small refactoring to reduce redundant function params and email config creation.
- Adds validation to the `SetUserGrafanaConfig` endpoint (specifically template and receiver).
- Adds template count and size limits as warnings alongside `SetUserGrafanaConfig` validation.
- Ensures `ApplyConfig` uses the correct template loader for Grafana AM receivers/tenants.
- Adds regression tests for the described issues.

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -435,18 +435,29 @@ func clusterWait(position func() int, timeout time.Duration) func() time.Duratio
 
 // ApplyConfig applies a new configuration to an Alertmanager.
 func (am *Alertmanager) ApplyConfig(conf *definition.PostableApiAlertingConfig, tmpls []alertingTemplates.TemplateDefinition, rawCfg string, tmplExternalURL *url.URL, staticHeaders map[string]string) error {
-	templates := make([]string, 0, len(tmpls))
-	for _, tmpl := range tmpls {
-		templates = append(templates, tmpl.Template)
+	cfg := definition.GrafanaToUpstreamConfig(conf)
+
+	emailCfg := alertingReceivers.EmailSenderConfig{
+		AuthPassword:  string(cfg.Global.SMTPAuthPassword),
+		AuthUser:      cfg.Global.SMTPAuthUsername,
+		CertFile:      cfg.Global.HTTPConfig.TLSConfig.CertFile,
+		ContentTypes:  []string{"text/html"},
+		EhloIdentity:  cfg.Global.SMTPHello,
+		ExternalURL:   tmplExternalURL.String(),
+		FromAddress:   cfg.Global.SMTPFrom,
+		FromName:      "Grafana",
+		Host:          cfg.Global.SMTPSmarthost.String(),
+		KeyFile:       cfg.Global.HTTPConfig.TLSConfig.KeyFile,
+		SkipVerify:    !cfg.Global.SMTPRequireTLS,
+		StaticHeaders: staticHeaders,
+		SentBy:        fmt.Sprintf("Mimir v%s", version.Version),
 	}
 
-	tmpl, err := loadTemplates(templates, WithCustomFunctions(am.cfg.UserID))
+	integrationsMap, err := am.buildIntegrationsMap(emailCfg, conf.Receivers, tmpls, tmplExternalURL)
 	if err != nil {
 		return err
 	}
-	tmpl.ExternalURL = tmplExternalURL
 
-	cfg := definition.GrafanaToUpstreamConfig(conf)
 	am.api.Update(&cfg, func(_ model.LabelSet) {})
 
 	// Ensure inhibitor is set before being called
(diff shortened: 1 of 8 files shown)


pkg/alertmanager/alertmanager.go
```diff
@@ -664,9 +653,36 @@ func (am *Alertmanager) buildIntegrationsMap(gCfg *config.GlobalConfig, nc []*de
...
+	// templates, we load our best guess as to the appropriate one (Grafana vs Cloud) here to ensure the definitions
+	// are valid.
+	// This might appear different from the above dynamic approach that depends on receiver types, and it is, but
+	// currently AMs should not have mixed receiver types. So, this is a safe (and necessary) workaround.
+	if tmpl == nil && gTmpl == nil {
+		_, err := am.loadTemplates(templates, WithCustomFunctions(am.cfg.UserID))
+		if err != nil {
+			return nil, err
+		}
+	}
```
santihernandezc, line 666: I'm fine with this workaround for now, but relying on an optional, experimental argument to choose which set of functions we're using to validate templates is not ideal :grimacing:

pkg/alertmanager/api_grafana.go
```diff
@@ -419,3 +430,62 @@ func (am *MultitenantAlertmanager) DeleteUserGrafanaConfig(w http.ResponseWriter
 	w.WriteHeader(http.StatusOK)
 	util.WriteJSONResponse(w, successResult{Status: statusSuccess})
 }
+
+// ValidateUserGrafanaConfig validates the Grafana Alertmanager configuration.
+func validateUserGrafanaConfig(logger log.Logger, cfg alertspb.GrafanaAlertConfigDesc, limits Limits, user string) error {
+	// We don't have a valid use case for empty configurations. If a tenant does not have a
+	// configuration set and issue a request to the Alertmanager, we'll a) upload an empty
+	// config and b) immediately start an Alertmanager instance for them if a fallback
+	// configuration is provisioned.
```
santihernandezc, line 439: Nit: Something along the lines of:
```suggestion
	// configuration set and issue a request to the Alertmanager, we'll a) upload an empty
	// config and b) on the next configuration sync, start an Alertmanager instance for them if a fallback
	// configuration is provisioned.
```

Or omit the "immediately" part. It's not true now, but it will be if we implement _strict initialization_.

pkg/alertmanager/api_grafana.go
```diff
@@ -419,3 +430,62 @@ func (am *MultitenantAlertmanager) DeleteUserGrafanaConfig(w http.ResponseWriter
...
+
+	if maxSize := limits.AlertmanagerMaxTemplateSize(user); maxSize > 0 {
+		// Permissive limit to first verify impact on existing users.
+		// TODO: Make this strict.
+		for name, tmpl := range grafanaConfig.Templates {
+			if size := len(tmpl.Body); size > maxSize {
+				level.Warn(logger).Log("msg", "template too big", "user", user, "template", name, "size", size, "limit", maxSize)
+			}
+		}
+	}
```
santihernandezc, line 469: Nice :sunglasses::+1:

santihernandezc (approved):
Nice work! :rocket:

pkg/alertmanager/api_grafana_test.go
```diff
@@ -410,21 +490,21 @@ func TestMultitenantAlertmanager_SetUserGrafanaConfig(t *testing.T) {
 			).WithContext(ctx)
 
 			am.SetUserGrafanaConfig(rec, req)
-			require.Equal(t, tc.expStatusCode, rec.Code)
+			assert.Equal(t, tc.expStatusCode, rec.Code)
```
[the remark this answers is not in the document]
JacobsonMT: Was useful during testing to see the message body when the code didn't match, figured I'd keep it that way

pkg/alertmanager/alertmanager.go
JacobsonMT: Agreed. Not much else to do though without a much larger refactor. At least it only ever runs if a tenant has templates but no receivers, which is rare given the default receiver needed in the notification policy.

PR #11307, opened 2025-04-25
[WIP] parquet-mimir fix tests and lints

francoposa:
#### What this PR does

* some minor lint fixes
* what looks like big changes in parquet-converter `starting`, `running`, and `stopping` funcs, but they are mostly just rearranged to mirror those in the compactor so that it's less weird to debug 
* some tweaks to the expected values for some ring member metrics in integration tests.
  * there is still a bunch of expected ring member metric values to fix in integration tests because with the current configuration we are:
  * still starting the store-gateways with its 512 ring tokens
  * _not_  initializing the the store-gateway queryable, which would normally also see the 512 ring tokens for the store-gateway sharding ring and contribute a tombstone on shutdown
  * now starting the parquet-converter, with _its_ 512 tokens and a tombstone on shutdown

There is a bunch to figure out there for making those integration tests make sense, obviously the most correct answer is to have a config option & test cases to handle this correctly in integration tests (when parquet is on, we would not start the store-gateways, etc.) but I think this is fine for this PR.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/tsdb/bucketindex/loader.go
+++ b/pkg/storage/tsdb/bucketindex/loader.go
@@ -7,6 +7,7 @@ package bucketindex
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 )
 
 type LoaderConfig struct {
+	ExtraMetricsPrefix    string
 	CheckInterval         time.Duration
 	UpdateOnStaleInterval time.Duration
 	UpdateOnErrorInterval time.Duration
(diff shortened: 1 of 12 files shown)


pkg/storage/tsdb/bucketindex/loader.go
```diff
@@ -24,6 +25,7 @@ import (
 )
 
 type LoaderConfig struct {
+	ExtraMetricsPrefix    string
```
francoposa, line 28: this avoids duplicate metrics registration when we have multiple bucket index loaders initialized

PR #11311, opened 2025-04-25
Alerts: Add Lookback in BucketIndexNotUpdated

dmwilson-grafana:
#### What this PR does

This PR fixes an issue where `BucketIndexNotUpdated` alerts can fire even though the index has been updated within the alert threshold (typically 2100s). If multiple compactors are publishing a metric for a tenant and the compactor which has most  recently updated the index goes down, the alert will fire despite the update.

Consider the following example where `compactor-12` goes down at 3:44, `compactor-25` and `compactor-33` come up at 3:52, and `compactor-33` goes down again around 4:07. Although `compactor-33` updated the index around 3:56,   the alert fires at 4:09 because it could only reference the timestamp published by `compactor-25` which corresponded to the 3:33 index update. 

Screenshot 2025-04-25 at 2 36 20 PM

With this PR's change to the query, the metric from `compactor-33` prevents the alert from firing for up to 15m after the pod goes down. At 4:08, the metric would take the minimum of  (2200s, 800s) and remain silent.

Screenshot 2025-04-25 at 2 48 04 PM

see also - compactor changes to mitigate: https://github.com/grafana/mimir/pull/11303/ 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

NickAnge (approved):
Thank you.

operations/mimir-mixin/alerts/blocks.libsonnet
```diff
@@ -223,10 +223,10 @@
           // Alert if the bucket index has not been updated for a given user. The default update interval is 900 seconds
           // so we alert if we've missed two updates plus a 300 second buffer to avoid false-positives. It's important
           // that this alert fire before queriers start to return errors because the bucket index is too old (3600 seconds
-          // by default).
+          // by default). Allow a 15m lookback to include compactors that may have recently updated the index and then been terminated.
```
dimitarvdimitrov, line 226: did you choose 15m because of the default cleanup interval? Could be worth mentioning this in the comment

dimitarvdimitrov (approved):
thanks, only one small suggestion about the comment

PR #11318, opened 2025-04-26
Trim unused fields from blockbuilder scheduler protos.

seizethedave:
#### What this PR does

Remove five unused fields from the block-builder scheduler protos. We didn't end up using these, and as the code that makes the scheduler fully functional was only merged this week it's unlikely anyone is running the blockbuilder<>scheduler paired. In our own dev environment we'll just do a one-time restart of the pair of services to cut past any proto incompatibility.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥

PR #11319, opened 2025-04-26
fix(parquet reader): Scan rows fix for MatchNotEqual and MatchNotRegexp

jesusvazquez:
Follow up PR to https://github.com/grafana/mimir/pull/11301

Found this bug while adding `foo!="bar"` matcher to avoid the cache.


MichaHoffmann (approved):

pkg/storage/parquet/reader.go
```diff
@@ -245,9 +245,14 @@ func (pr *ParquetReader) ScanRows(ctx context.Context, rows [][]int64, full bool
...
-	if !ok {
-		if m.Matches("") {
+	ci, validColumn := pr.columnIndexMap[m.Name]
+	if !validColumn {
+		// If the column does not exist we can ignore the matcher under certain conditions.
+		// * `!=` and `!~` matchers can always be ignored.
+		// * `=` matcher can be ignored if the value is empty.
+		if m.Type == labels.MatchNotEqual ||
+			m.Type == labels.MatchNotRegexp ||
+			m.Type == labels.MatchEqual && m.Matches("") {
```
npazosmendez, line 255: I don't understand the need for this change, doesn't the plan `m.Matches("")` cover all cases? I tried running the new test you added with the previous version and it passes

jesusvazquez: You could have the `foo!=bar` case that would not fall into `m.Matches("")`, `foo` is still an invalid column so it should just be skipped.

npazosmendez: If `foo` does not exist it would enter this `if !validColumn`, and `m.Matches("")` would return true for `!=bar`. Also, a `!=` is not always noop for non-existing columns, it could be a `!=""`. The `m.Matches("")` should catch all those cases on its own.

But, maybe I'm not seeing it. Could you add a test that fails with the current implementation?

jesusvazquez: We paired and checked this was right.

PR #11321, opened 2025-04-27
ruler: Rules API to return a warning if evaluation is disabled for tenant

narqo:
#### What this PR does

This PR updates ruler's `/api/v1/rules` and `/api/v1/alerts` APIs to reply with an error if tenant's rule evaluation is disabled in the limits.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 * [ENHANCEMENT] OTLP: Add support for converting OTel explicit bucket histograms to Prometheus native histograms with custom buckets using the `distributor.otel-convert-histograms-to-nhcb` flag. #11077
 * [ENHANCEMENT] Add configurable per-tenant `limited_queries`, which you can only run at or less than an allowed frequency. #11097
 * [ENHANCEMENT] Ingest-Storage: Add `ingest-storage.kafka.producer-record-version` to allow control Kafka record versioning. #11244
+* [ENHANCEMENT] Ruler: Update `/api/v1/rules` and `/api/v1/alerts` to reply with a client error if evaluation is disabled for tenant. #11321
```
tacole02, line 44:
```suggestion
* [ENHANCEMENT] Ruler: Update `/api/v1/rules` and `/api/v1/alerts` to reply with a client error if evaluation is disabled for the tenant. #11321
```

tacole02 (approved):

dimitarvdimitrov (commented):
SGTM. We chatted on slack about this. 

I understand you're doing this only for the JSON endpoints that are also present in prometheus. The other endpoints don't really have support for this extra information and it's not super clear if we want to also disable configuration for disabled rules. It makes more sense for the JSON endpoints because they're supposed to return the currently evaluating rules and their state. When rules are disabled, these endpoints don't make that much sense.

_I know i'm annoying, but it would have been helpful to see this in the description_

PR #11332, opened 2025-04-28
GitHub actions: restrict permissions for some workflows

pracucci:
#### What this PR does

Default permissions (when not specified) may be broader than necessary. In this PR I'm restricting the permissions of some actions. I'm not sure changes in `helm-release` are correct (that means it has enough permissions) but it's something we can later fix if broken.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):

duricanikolic (approved):
LGTM

PR #11343, opened 2025-04-29
mark-blocks: support multi-tenant input

andyasp:
#### What this PR does

A use case came up where blocks from multiple tenants needed to be marked. The input format for `mark-blocks` made this tedious since `--tenant` is required and it enforces a single tenant per run.

To make this easier, these changes maintain the old behavior when `--tenant` is provided, but now it is optional. When it is not provided the input is expected to be of the form `tenantID/blockID` rather than `blockID`.

I've updated the unit tests and manually tested these changes and they behaved as expected.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/tools/mark-blocks/README.md
+++ b/tools/mark-blocks/README.md
@@ -4,10 +4,10 @@ This program creates or removes markers for blocks.
 
 ## Flags
 
-- `--tenant` (required) The tenant that owns the blocks
 - `--mark-type` (required) Mark type to create or remove, valid options: `deletion`, `no-compact`
-- `--blocks` (optional) Comma separated list of blocks IDs. If non-empty, `--blocks-file` is ignored
-- `--blocks-file` (optional) File containing a block ID per-line. Defaults to standard input (`-`). Ignored if `--blocks` is non-empty
+- `--tenant` (optional) Tenant ID of the owner of the block(s). If empty (the default) then each block is assumed to be of the form tenantID/blockID, otherwise blockID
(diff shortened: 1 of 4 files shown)


tools/mark-blocks/README.md
```diff
@@ -17,6 +17,25 @@ This program creates or removes markers for blocks.
 
 Each supported object storage service also has an additional set of flags (see examples in [Running](##Running)).
 
+## Input formats
+
+For convenience, this tool supports two input formats controlled by the `--tenant` flag.
+
+If `--tenant` is empty (the default), then a tenant must be specified for each block provided in either `--blocks` or `--blocks-file`. For example, an input file could be of the form:
```
56quarters, line 24: Should we add an example for when multi-tenancy is disabled and we use the fallback tenant?

56quarters (approved):
Nice

tools/mark-blocks/README.md
56quarters: Actually, I guess I don't know how that would even work offhand. `--tenant=anonymous` ?

andyasp: Yeah, we don't assume `anonymous` here and never did for this tool. It used to fail if `--tenant` wasn't specified.

You'd either have to specify `--tenant=anonymous` or denote it in the input like `--blocks="anonymous/<blockID1>,anonymous/<blockID2>`, which aligns with the other examples.

PR #11344, opened 2025-04-30
feature(queries): enable duration arithmetic

krajorama:
#### What this PR does

Enables use of https://github.com/prometheus/prometheus/pull/16249 feature in Mimir.

Fixes N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added. TBD


pkg/mimirtool/commands/analyse.go
```diff
@@ -19,6 +19,7 @@ type AnalyzeCommand struct{}
 func (cmd *AnalyzeCommand) Register(app *kingpin.Application, envVars EnvVarNames) {
 	analyzeCmd := app.Command("analyze", "Run analysis against your Prometheus, Grafana, and Grafana Mimir to see which metrics are being used and exported.")
 	analyzeCmd.Flag("enable-experimental-functions", "If set, enables parsing experimental PromQL functions.").BoolVar(&parser.EnableExperimentalFunctions)
+	analyzeCmd.Flag("enable-experimental-duration-expressions", "If set, enables parsing experimental duration expressions.").BoolVar(&parser.ExperimentalDurationExpr)
```
krajorama, line 22: Not sure if this should be an option of not.

56quarters: If we're enabling it unconditionally in Mimir, we should enable it unconditionally here. Experimental functions are turned on or off per user so you want to behavior of `mimirtool` to match (hence the option).

pkg/streamingpromql/engine_test.go
```diff
@@ -170,6 +170,11 @@ func TestNewInstantQuery_Strings(t *testing.T) {
 // Test cases that are not supported by the streaming engine are commented out (or, if the entire file is not supported, .disabled is appended to the file name).
 // Once the streaming engine supports all PromQL features exercised by Prometheus' test cases, we can remove these files and instead call promql.RunBuiltinTests here instead.
 func TestUpstreamTestCases(t *testing.T) {
+	origExperimentalDurationExpr := parser.ExperimentalDurationExpr
+	parser.ExperimentalDurationExpr = true
```
56quarters, line 174: I think this should be in the `init()` call above since we turn it on unconditionally in Mimir, we don't really want to toggle it on or off.

56quarters (approved):
LGTM with minor suggestions

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [CHANGE] gRPC/HTTP clients: Rename metric `cortex_client_request_invalid_cluster_validation_labels_total` to `cortex_client_invalid_cluster_validation_label_requests_total`. #11237
 * [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. Limitations: Created timestamp is ignored, per series metadata is merged on metric family level automatically, ingestion might fail if client sends ProtoBuf fields out of order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0` depending on the detected Remote-Write protocol. #11100 #11101 #11192 #11143
 * [FEATURE] Query-frontend: expand `query-frontend.cache-errors` and `query-frontend.results-cache-ttl-for-errors` configuration options to cache non-transient response failures for instant queries. #11120
+* [FEATURE] Querier, query-frontend, ruler: enable experimental support for duration expressions in PromQL, which are simple arithmetics on numbers in offset and range   specification. #11344
```
tacole02, line 16:
```suggestion
* [FEATURE] Querier, query-frontend, ruler: Enable experimental support for duration expressions in PromQL, which are simple arithmetics on numbers in offset and range specification. #11344
```

docs/sources/mimir/configure/about-versioning.md
```diff
@@ -217,6 +217,7 @@ The following features are currently experimental:
   - Enable PromQL experimental functions per-tenant (`-query-frontend.enabled-promql-experimental-functions` and the `enabled_promql_experimental_functions` per-tenant limit)
   - Support for cluster validation via `-query-frontend.client-cluster-validation.label` or `-common.client-cluster-validation.label`.
     Requests with invalid cluster validation labels are tracked via the `cortex_client_invalid_cluster_validation_label_requests_total` metric.
+  - Support for duration expression in PromQL, which are simple arithmetics on numbers in offset and range specification.
```
tacole02, line 220:
```suggestion
  - Support for duration expressions in PromQL, which are simple arithmetics on numbers in offset and range specification.
```

tacole02 (approved):

tacole02 (approved):

dimitarvdimitrov:
@krajorama i'm late to this, but i think we should add some tests to check we're actually supporting this. I see this PR enabled the feature, but is there anything which exercises it? Do we know if it works with query sharding for example?

krajorama:
> @krajorama i'm late to this, but i think we should add some tests to check we're actually supporting this. I see this PR enabled the feature, but is there anything which exercises it? Do we know if it works with query sharding for example?

This turned out to be very true, non-sharded expression work as far as I can see, but sharded has issues.

PR #11345, opened 2025-04-30
feat: Add the option of a lower read consistency level

mbarbon:
[Note that this MR contains a large amount of changes to vendor/ as it depends on [this dskit PR](https://github.com/grafana/dskit/pull/686)]

Using a lower read consistency level allows trading lower consistency for higher availability.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (commented):
A quick comment about the correctness (I left the rest of the team comment whether it's something desired for Mimir or not). If we want to allow partial query results, then it's not just about how we build the replication set but also we have to change the quorum logic, implemented in `ring.DoBatchWithOptions()` (quorum is implemented here).

mbarbon:
> A quick comment about the correctness (I left the rest of the team comment whether it's something desired for Mimir or not). If we want to allow partial query results, then it's not just about how we build the replication set but also we have to change the quorum logic, implemented in `ring.DoBatchWithOptions()` (quorum is implemented here).

@pracucci could you point out the logic that needs to be updated? I'm asking because I already looked at that code and I thought it did not need to be changed...

PR #11348, opened 2025-04-30
Set explicit permissions for all publish documentation workflows

jdbaldry:


56quarters (approved):
LGTM

PR #11354, opened 2025-05-01
querier: add support dynamic replication in consistency checks

dimitarvdimitrov:
#### What this PR does

the querier today retries only 3 times each block. This works ok with replication factor=3. But with dynamic replication it feels like we’re giving up too early. There might be other store-gateways that can serve the block, but we don’t even attempt them.

This PR changes the blockStoreQuerier to retry until the maximum replication factor. This doesn't add additional overhead for non-replicated blocks because `blocksStoreReplicationSet.GetClientsFor` would return an error after the third retry anyway and we'd short-circuit here https://github.com/grafana/mimir/blob/efc8a65a733bebba8a5fd4305d55a50393534d75/pkg/querier/blocks_store_queryable.go#L560-L567

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 * [ENHANCEMENT] Ingest-Storage: Add `ingest-storage.kafka.producer-record-version` to allow control Kafka record versioning. #11244
 * [ENHANCEMENT] Ruler: Update `<prometheus-http-prefix>/api/v1/rules` and `<prometheus-http-prefix>/api/v1/alerts` to reply with HTTP error 422 if rule evaluation is completely disabled for the tenant. If only recording rule- or alerting rule evaluation is disabled for the tenant, the response now includes a corresponding warning. #11321
 * [ENHANCEMENT] Add tenant configuration block `ruler_alertmanager_client_config` which allows the Ruler's Alertmanager client options to be specified on a per-tenant basis. #10816
+* [ENHANCEMENT] Store-gateway: retry querying blocks from store-gateways with dynamic replication until trying all possible store-gateways. #11354
 * [BUGFIX] OTLP: Fix response body and Content-Type header to align with spec. #10852
 * [BUGFIX] Compactor: fix issue where block becomes permanently stuck when the Compactor's block cleanup job partially deletes a block. #10888
 * [BUGFIX] Storage: fix intermittent failures in S3 upload retries. #10952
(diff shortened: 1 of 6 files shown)


pkg/querier/blocks_store_queryable.go
```diff
@@ -257,11 +250,17 @@ func NewBlocksStoreQueryableFromConfig(querierCfg Config, gatewayCfg storegatewa
 		return nil, errors.Wrap(err, "failed to create store set")
 	}
 
+	maxReplication := gatewayCfg.ShardingRing.ReplicationFactor
+	if gatewayCfg.DynamicReplication.Enabled {
+		maxReplication *= gatewayCfg.DynamicReplication.Multiple
```
56quarters, line 255: To keep the logic together, maybe add a `MaxReplication()` method to the dynamic replication interface/object above. It'd also be nice if the no-op one returned the default replication factor so we don't use extra retries when they're not needed.

56quarters (approved):
LGTM but I think the `maxReplication` logic would be better in the dynamic replication interface instead of the consistency tracker.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 * [ENHANCEMENT] Add configurable per-tenant `limited_queries`, which you can only run at or less than an allowed frequency. #11097
 * [ENHANCEMENT] Ingest-Storage: Add `ingest-storage.kafka.producer-record-version` to allow control Kafka record versioning. #11244
 * [ENHANCEMENT] Ruler: Update `<prometheus-http-prefix>/api/v1/rules` and `<prometheus-http-prefix>/api/v1/alerts` to reply with HTTP error 422 if rule evaluation is completely disabled for the tenant. If only recording rule- or alerting rule evaluation is disabled for the tenant, the response now includes a corresponding warning. #11321
+* [ENHANCEMENT] Store-gateway: retry querying blocks from store-gateways with dynamic replication until trying all possible store-gateways. #11354
```
tacole02, line 45:
```suggestion
* [ENHANCEMENT] Store-gateway: Retry querying blocks from store-gateways with dynamic replication until trying all possible store-gateways. #11354
```

tacole02 (approved):

pkg/querier/blocks_store_queryable.go
dimitarvdimitrov: have a look at https://github.com/grafana/mimir/pull/11354/commits/8293fb6119a79372df7682e90fe4cc66064e21ff

i wasn't sure what's the best way to wire it up. Another option would be to expose this on the `BlocksStoreSet` interface, but i think that's a bit out of its concern. PTAL 🙏

PR #11358, opened 2025-05-01
block-builder: Add dashboard panels for the scheduler

codesome:
#### What this PR does

Adds panels for the scheduler

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir-mixin/config.libsonnet
+++ b/operations/mimir-mixin/config.libsonnet
@@ -40,6 +40,7 @@
       alertmanager_im: 'alertmanager-im',
       ingester: 'ingester',
       block_builder: 'block-builder',
+      block_builder_scheduler: 'block-builder-scheduler',
       distributor: 'distributor',
       querier: 'querier',
       query_frontend: 'query-frontend',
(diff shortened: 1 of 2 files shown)


seizethedave (commented):
Looks good!
- I kinda like how @narqo in his private dashboard had rendered both outstanding and assigned jobs in one chart as the two are very related. Saves space, too. WDYT?

narqo:
> rendered both outstanding and assigned jobs in one chart as the two are very related

I wanted to suggest the same. This is simialr to how compactor does it (ref [compactor's dashboard][1] in the mixin).

[1]: https://github.com/grafana/mimir/blob/80caecfbd53aad103bcc2d937e55f144f555b25c/operations/mimir-mixin/dashboards/compactor.libsonnet#L94-L111

codesome:
Made the following updates:
* Put outstanding and active jobs in the same panel
* Consolidated 2 rows into a single row, so we have 4 panels in the same row which is still very viewable.

narqo (approved):
🔥 Works for me (we can always improve it later, as we're getting more experience with operating the block-builder components).

PR #11359, opened 2025-05-01
Ruler: Fix Alertmanager URL tenant config mapping in deployments when runtime config is disabled

alexweav:
#### What this PR does

In the process of https://github.com/grafana/mimir/pull/10816, several ruler options are being migrated to per-tenant configs.

Normally, when yaml configuration containing the old values under `ruler:` is provided, we want to internally map it to the tenant default limits to preserve backward compatibility. This is typically done in `initRuntimeConfig()`.

This PR fixes a bug where this mapping fails on deployments with pure static config and no runtime config, and with the ruler alertmanager still configured via yaml in the old method. (i.e. Mimir deployments which use `-config.file` and not `-runtime-config.file`) In such deployments we never launch the runtime config service whatsoever, but we still need the mapping to happen as it might have been statically configured.

The fix is to simply move the mapping earlier in the process, it needs to happen before any shortcuts are taken.

I've noted this potential pitfall in the code and in the migration guide so that future changes others don't run into it.

This is covered already by GEM e2e tests, hence I didn't create a duplicate here. It's already under coverage. Mimir's integration test framework tends to set up Mimir via CLI options, which are not affected by the bug.

n/a

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/internal/contributing/how-to-convert-config-to-per-tenant-limit.md
+++ b/docs/internal/contributing/how-to-convert-config-to-per-tenant-limit.md
@@ -171,6 +171,8 @@ Until deprecation is complete, we need to ensure that a user who is still settin
    }
    ```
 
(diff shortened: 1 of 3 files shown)


56quarters (approved):
LGTM

PR #11367, opened 2025-05-02
distributor: ha tracker memberlist flaky test add notify itnerval

NickAnge:
#### What this PR does

This should 

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov (approved):
LGTM as a temporary solution to make tests less annoying. But I think we should fix the race condition in the code and then undo this temporary fix

PR #11369, opened 2025-05-02
querier: improve e2e tests for error parsing

dimitarvdimitrov:
#### What this PR does

The existing test wasn't catching a bug (https://github.com/prometheus/prometheus/pull/16544) because 1) it was only querying a single store-gateway and 2) blocks were returning disjoin series.

This PR improves the test suite to make it more realistic:
- 3 ingesters instead of 1
- 2 store-gateways instead of 1
- overlapping series in each block so that we can merge series from different sources

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov:
marking this as ready for review as we've received the fix from upstream

narqo (approved):
🔥

PR #11393, opened 2025-05-05
ingest-storage: Use pooled timeseries slices when deserializing

alexweav:
#### What this PR does

The pusherConsumer deserializes directly into a `&mimirpb.WriteRequest{}` by calling the generated `Unmarshal`. This path will [fetch invidiual PreallocTimeseries from the pool](https://github.com/grafana/mimir/blob/68907424027dcd77f50406fa6285503e3b6cad4c/pkg/mimirpb/timeseries.go#L269), but [not the timeseries slice itself](https://github.com/grafana/mimir/blob/68907424027dcd77f50406fa6285503e3b6cad4c/pkg/mimirpb/mimir.pb.go#L7515), which is subject to the usual extra allocations when appending.

This PR uses a slice from the pool in addition to the contained objects.

There are 2 possible paths after we take from the pool, depending on configuration.
- The sequentialStoragePusher, which [forwards the request straight to the ingester](https://github.com/grafana/mimir/blob/main/pkg/storage/ingest/pusher.go#L223). The ingester [frees the slice and its contents](https://github.com/grafana/mimir/blob/68907424027dcd77f50406fa6285503e3b6cad4c/pkg/ingester/ingester.go#L3984). Our contents came from the pool but the slice did not, so prior to this PR we are putting the externally created slice back into the sync.Pool on each request. sync.Pool has an internal safeguard against it growing infinitely, but it causes extra GC churn.
- The parallelStoragePusher, which [rearranges the timeseries into batches](https://github.com/grafana/mimir/blob/68907424027dcd77f50406fa6285503e3b6cad4c/pkg/storage/ingest/pusher.go#L399-L407). The batch slice is already [fetched from the pool](https://github.com/grafana/mimir/blob/68907424027dcd77f50406fa6285503e3b6cad4c/pkg/storage/ingest/pusher.go#L585). That means we can free our serialization slice (but not the contents!) when we're done rearranging it. The ingester will free the batch slice and all the timeseries as usual.

The main purpose of this PR is more of a refactor than an optimization. The block builder already uses a timeseries slice from the pool here. By aligning the allocation/freeing pattern, I can proceed to the same deserialization logic in both services, which I'd like to do for record format V2. The reduced allocations are just a bonus.

TL;DR
```
go test ./pkg/storage/ingest/... -run $^ -bench BenchmarkPusherConsumer -benchtime=10s -benchmem
```
Before:
```
BenchmarkPusherConsumer/sequential_pusher-16         	   20348	    592474 ns/op	  826381 B/op	     933 allocs/op
BenchmarkPusherConsumer/parallel_pusher-16           	    7610	   1519643 ns/op	 1020897 B/op	    1952 allocs/op
```
After:
```
BenchmarkPusherConsumer/sequential_pusher-16         	   31896	    385991 ns/op	   28848 B/op	     463 allocs/op
BenchmarkPusherConsumer/parallel_pusher-16           	    9435	   1246078 ns/op	  208315 B/op	    1502 allocs/op
```

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storage/ingest/pusher.go
```diff
@@ -395,6 +397,10 @@ func labelAdaptersHash(b []byte, ls []mimirpb.LabelAdapter) ([]byte, uint64) {
 // PushToStorage ignores SkipLabelNameValidation because that field is only used in the distributor and not in the ingester.
 // PushToStorage aborts the request if it encounters an error.
 func (p *parallelStorageShards) PushToStorage(ctx context.Context, request *mimirpb.WriteRequest) error {
+	// We're moving Timeseries into batches, each batch has a fresh timeseries slice.
+	// We're done with the slice in the request here, the contents will live on in the batch and be freed when the batch is freed.
+	defer mimirpb.ReuseTimeseriesSliceDangerous(request.Timeseries)
```
seizethedave, line 402: `Consume` is responsible for allocating from the pool, then `PushToStorage` frees back to the pool with the assumption that `Consume` is the caller. Not super familiar with the details, but can Consume do the freeing for symmetry? Like right after it calls `c.pushToStorage(...)`.

alexweav: It would definitely be cleaner to just free where it was allocated. Let's see if I can get it into that form or if I run into issues.

The trickiest part I see so far is we still have some implementation specific behavior making it hard to factor out. The parallelStoragePusher might collapse N requests into 1 batch, so you have N slices to free in that path, but only 1 slice to free in the sequential path.

pkg/storage/ingest/pusher_test.go
```diff
@@ -1550,3 +1550,54 @@ func setupQueue(t *testing.T, capacity, batchSize int, series []mimirpb.Prealloc
...
+	}
+
+	b.Run("sequential pusher", func(b *testing.B) {
+		kcfg := KafkaConfig{}
+		flagext.DefaultValues(&kcfg)
+		kcfg.IngestionConcurrencyMax = 0
+		metrics := newPusherConsumerMetrics(prometheus.NewPedanticRegistry())
+		c := newPusherConsumer(pusher, kcfg, metrics, log.NewNopLogger())
+
+		for range b.N {
```
dimitarvdimitrov, line 1581: nit to call `b.Reset()` before starting the loop. Also in the loop below

pkg/storage/ingest/pusher_test.go
```diff
@@ -1550,3 +1550,54 @@ func setupQueue(t *testing.T, capacity, batchSize int, series []mimirpb.Prealloc
...
+	})
+
+	b.Run("parallel pusher", func(b *testing.B) {
+		kcfg := KafkaConfig{}
+		flagext.DefaultValues(&kcfg)
+		kcfg.IngestionConcurrencyMax = 2
+		kcfg.IngestionConcurrencyEstimatedBytesPerSample = 1
+		kcfg.IngestionConcurrencyTargetFlushesPerShard = 1
+		kcfg.IngestionConcurrencyBatchSize = 5
+		kcfg.IngestionConcurrencyQueueCapacity = 5
```
dimitarvdimitrov, line 1594: This config looks a bit unrealistic. Are the defaults insufficient?

dimitarvdimitrov (commented):
nice one. Not sure if you know it, but there's [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat)  which helps display comparisons between benchmarks runs more nicely.

pkg/storage/ingest/pusher.go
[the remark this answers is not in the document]
alexweav: Good call. I think in the current code it's still safe, as ReuseSlice does the same thing already. It returns each timeseries to the pool, then returns `ts[:0]` to the slice pool without clearing, I imagine the leftover array of `ts` still has pointers to the freed timeseries :thinking: 

But I don't see a good reason to not do this, it's way safer. I think we can clear the slice as we pull the timeseries out of it, let me update the test image with that change.

pkg/storage/ingest/pusher_test.go
alexweav: The odd settings defintely caused the parallel pusher to allocate way more than usual since the values were so low. But, the delta caused by this change is still about the same:
Before:
```
BenchmarkPusherConsumer/sequential_pusher-16         	   19382	    627954 ns/op	  826590 B/op	     934 allocs/op
BenchmarkPusherConsumer/parallel_pusher-16           	   18906	    634400 ns/op	  827177 B/op	     940 allocs/op
```
After:
```
BenchmarkPusherConsumer/sequential_pusher-16         	   30938	    396641 ns/op	   28835 B/op	     463 allocs/op
BenchmarkPusherConsumer/parallel_pusher-16           	   28466	    418158 ns/op	   29447 B/op	     469 allocs/op
```

Benchstat (nice suggestion!)
```
                                    │   old.txt   │               new.txt               │
PusherConsumer/sequential_pusher-16   616.5µ ± 1%   405.2µ ± 4%  -34.28% (p=0.000 n=10)
PusherConsumer/parallel_pusher-16     629.3µ ± 1%   414.2µ ± 1%  -34.17% (p=0.000 n=10)
geomean                               622.9µ        409.7µ       -34.23%

                                    │    old.txt    │               new.txt                │
                                    │     B/op      │     B/op      vs base                │
PusherConsumer/sequential_pusher-16   807.32Ki ± 0%   28.20Ki ± 0%  -96.51% (p=0.000 n=10)
PusherConsumer/parallel_pusher-16     807.87Ki ± 0%   28.78Ki ± 0%  -96.44% (p=0.000 n=10)
geomean                                807.6Ki        28.49Ki       -96.47%

                                    │  old.txt   │              new.txt               │
                                    │ allocs/op  │ allocs/op   vs base                │
PusherConsumer/sequential_pusher-16   934.0 ± 0%   463.0 ± 0%  -50.43% (p=0.000 n=10)
PusherConsumer/parallel_pusher-16     940.0 ± 0%   469.0 ± 0%  -50.11% (p=0.000 n=10)
geomean                               937.0        466.0       -50.27%
```

pkg/storage/ingest/pusher.go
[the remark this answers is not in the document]
alexweav: I changed the approach a bit.:
- We zero out the each timeseries in the orignial slice once it's transferred to the batch with no errors.
- We now only re-use the slice if we successfully transferred and cleared every single timeseries in the request. If there are errors/leftovers in the slice it won't be returned to the pool at all. This should make it much safer and less brittle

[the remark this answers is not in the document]
alexweav: I think you're right that this would have at least contributed to a future bug; there was a test that relied on inspecting the slice after it had been pushed to storage (thus freed). I didn't see any code paths outside tests that did this, but someone could easily add one in the future.

pracucci (commented):
While reviewing this code, I think we have a race condition. When timeseries slice is released to the pool in the `PushToStorage()`, we break any read to the write request happening after that. For example, look here:
https://github.com/grafana/mimir/blob/f266d6be320964088a589ca578e850e3edda2ac3/pkg/storage/ingest/pusher.go#L468

We access the `WriteRequest` but at that point the timeseries slice is already released to the pool.

pkg/storage/ingest/pusher_test.go
dimitarvdimitrov: this looks like a massive gain; curious to see how it fares in a real cluster

pkg/storage/ingest/pusher.go
```diff
@@ -397,15 +399,19 @@ func labelAdaptersHash(b []byte, ls []mimirpb.LabelAdapter) ([]byte, uint64) {
...
-		hashBuf, shard = labelAdaptersHash(hashBuf, ts.Labels)
+		hashBuf, shard = labelAdaptersHash(hashBuf, request.Timeseries[i].Labels)
 		shard = shard % uint64(p.numShards)
 
-		if err := p.shards[shard].AddToBatch(ctx, request.Source, ts); err != nil {
+		if err := p.shards[shard].AddToBatch(ctx, request.Source, request.Timeseries[i]); err != nil {
 			return fmt.Errorf("encountered a non-client error when ingesting; this error was for a previous write request for the same tenant: %w", err)
 		}
+		// We're transferring ownership of the timeseries to the batch, clear the slice as we go so we can reuse it.
+		request.Timeseries[i] = mimirpb.PreallocTimeseries{}
```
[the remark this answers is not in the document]
dimitarvdimitrov: there was a thread about this earlier: https://github.com/grafana/mimir/pull/11393#discussion_r2077569916

I asked to do it so that we don't keep a ref to the same timeseries in two different places. It was safe to keep the ref, but it opens a potential future bug where we reuse the slice without appending new items to it.

Previously after we reuse the timeseries, we wouldn't be holding refs to the items in the slice. After this change we would be.

I'm not sure what's safer - copy the timeseries or move them to a different slice? Having them in a single place _seems_ a bit safer to me, but i don't have a lot of experience with memory management.

pkg/storage/ingest/pusher_test.go
alexweav: https://raintank-corp.slack.com/archives/C067K8L8NFM/p1746486368603179?thread_ts=1746486205.638989&cid=C067K8L8NFM

pkg/storage/ingest/pusher.go
```diff
@@ -460,12 +466,13 @@ func (p *parallelStorageShards) run(queue *batchingQueue) {
 		p.metrics.batchAge.Observe(time.Since(wr.startedAt).Seconds())
 		p.metrics.timeSeriesPerFlush.Observe(float64(len(wr.Timeseries)))
 		processingStart := time.Now()
+		requestContents := requestContents(wr.WriteRequest)
```
pracucci, line 470: What if we do `wr.WriteRequest = nil` here? Given it will be unsafe to use, better a panic rather than data corruption, if in the future we'll try to read it.

pracucci: Same in `pusherConsumer.Consume()`.

pkg/storage/ingest/pusher.go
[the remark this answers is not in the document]
pracucci: Ok I see your rationale. I'm fine keeping this.

pracucci (approved):
Thanks for addressing my feedback. I left another comment about nil-ify `WriteRequest` to have a panic in case of bugs rather than data corruption (which is much harder to debug). I still think that we should rename all `PushToStorage()` to `PushToStorageAndReleaseWriteRequest()` to clarify that the write request after that call is not safe anymore and shouldn't be used. I know it's just a naming thing, but it could give a great hint to the next engineer that will touch this code.

pkg/storage/ingest/pusher.go
```diff
@@ -93,10 +93,12 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) (returnEr
 			}
 
 			parsed := parsedRecord{
-				ctx:          r.ctx,
-				tenantID:     r.tenantID,
-				WriteRequest: &mimirpb.WriteRequest{},
-				index:        index,
+				ctx:      r.ctx,
+				tenantID: r.tenantID,
+				WriteRequest: &mimirpb.WriteRequest{
```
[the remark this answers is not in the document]
alexweav: That's what I did in my dependent PR :smile: see https://github.com/grafana/mimir/pull/11406/files#r2076276808
Decided to use this way here because it makes it more visibly obvious in review what's changing.

Updated, folks have seen it now

pkg/storage/ingest/pusher.go
alexweav: I cleared it at the spot where it's freed (it feels like a logical extension of freeing). The rename makes it clear at the call site too. Is that enough? :thinking:

pracucci (approved):
LGTM, thanks!

alexweav:
Just writing it down somewhere, this is probably the most relevant place.

While testing an image with this change in dev, a continuous test failure happened. I do not think it's related to this change.

- the failure (05/09): [explore](https://admin-dev-us-central-0.grafana-dev.net/grafana/explore?schemaVersion=1&panes=%7B%22djc%22:%7B%22datasource%22:%22loki-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D~%5C%22.%2Acontinuous.%2A%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%7C%3D%20%5C%22results_cache%3Dfalse%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D,%7B%22refId%22:%22B%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D%5C%22ingester%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%21%3D%20%5C%22per-tenant%20limit%5C%22%20%21%3D%20%5C%22WriteTo%20failed%5C%22%20%21%3D%20%5C%22duplicate-timestamp%5C%22%20%21%3D%20%5C%22license%5C%22%20%21%3D%20%5C%22too-old%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22,%22hide%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-30d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)
  - you can see it failed frequently before the rollout too.
- An ingester was stuck terminating for days before I deployed. (zone-b-2)
- That ingester logs WAL corruption consistently: [explore](https://admin-dev-us-central-0.grafana-dev.net/grafana/explore?schemaVersion=1&panes=%7B%22djc%22:%7B%22datasource%22:%22loki-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D~%5C%22.%2Acontinuous.%2A%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%7C%3D%20%5C%22results_cache%3Dfalse%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22,%22hide%22:true%7D,%7B%22refId%22:%22B%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D%5C%22ingester%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%21%3D%20%5C%22per-tenant%20limit%5C%22%20%21%3D%20%5C%22WriteTo%20failed%5C%22%20%21%3D%20%5C%22duplicate-timestamp%5C%22%20%21%3D%20%5C%22license%5C%22%20%21%3D%20%5C%22too-old%5C%22%20%7C%3D%20%5C%22WAL%5C%22%20%21%3D%20%5C%22exemplar%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22,%22hide%22:false%7D%5D,%22range%22:%7B%22from%22:%221746670114000%22,%22to%22:%221746757445000%22%7D%7D%7D&orgId=1)
- first failure at 02:21 UTC, for samples at 12:58. [explore](https://admin-dev-us-central-0.grafana-dev.net/grafana/explore?schemaVersion=1&panes=%7B%22djc%22:%7B%22datasource%22:%22loki-dev%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D~%5C%22.%2Acontinuous.%2A%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%7C%3D%20%5C%22results_cache%3Dfalse%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22,%22hide%22:false%7D,%7B%22refId%22:%22B%22,%22expr%22:%22%7Bnamespace%3D%5C%22mimir-dev-14%5C%22,%20container%3D%5C%22ingester%5C%22%7D%20%7C%3D%20%5C%22level%3Dwarn%5C%22%20%21%3D%20%5C%22per-tenant%20limit%5C%22%20%21%3D%20%5C%22WriteTo%20failed%5C%22%20%21%3D%20%5C%22duplicate-timestamp%5C%22%20%21%3D%20%5C%22license%5C%22%20%21%3D%20%5C%22too-old%5C%22%20%7C%3D%20%5C%22WAL%5C%22%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-dev%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22,%22hide%22:true%7D%5D,%22range%22:%7B%22from%22:%221746670114000%22,%22to%22:%221746757445000%22%7D%7D%7D&orgId=1)

The delay in sample corruption correlates with the WAL issues, that ingester was way behind for a long time. I would expect if this PR broke anything, that the continuous test would immediately fail on the sample that just got written, not have a several hour delay between affected samples and the test failing.

Thus, I don't think it's related.

Issue #11180, opened 2025-04-10
Flaky TestPartitionReader_ShouldNotBufferRecordsInTheKafkaClientWhenDone


56quarters:
The test timed out while running:

```
panic: test timed out after 30m0s
	running tests:
		TestPartitionReader_ShouldNotBufferRecordsInTheKafkaClientWhenDone/with_fetch_concurrency (29m10s)
goroutine 223928 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:2484 +0x605
created by time.goFunc
	/usr/local/go/src/time/sleep.go:215 +0x45
goroutine 1 [chan receive, 29 minutes]:
testing.(*T).Run(0xc000182380, {0x1ac0c27, 0x42}, 0x1ae18f8)
	/usr/local/go/src/testing/testing.go:1859 +0x91e
testing.runTests.func1(0xc000182380)
	/usr/local/go/src/testing/testing.go:2279 +0x86
testing.tRunner(0xc000182380, 0xc0000a5988)
	/usr/local/go/src/testing/testing.go:1792 +0x226
testing.runTests(0xc00019a030, {0x25ed9e0, 0x3e, 0x3e}, {0x640?, 0x700?, 0x25fbaa0?})
	/usr/local/go/src/testing/testing.go:2277 +0x96d
testing.(*M).Run(0xc0003654a0)
	/usr/local/go/src/testing/testing.go:2142 +0xeeb
go.uber.org/goleak.VerifyTestMain({0x1c9d9e0, 0xc0003654a0}, {0xc00033ac40, 0x4, 0x4})
	/__w/mimir/mimir/vendor/go.uber.org/goleak/testmain.go:53 +0x65
github.com/grafana/mimir/pkg/util/test.VerifyNoLeakTestMain(0xc0003654a0, {0x0, 0x0, 0x261e280?})
	/__w/mimir/mimir/pkg/util/test/leak.go:19 +0xbd
github.com/grafana/mimir/pkg/storage/ingest.TestMain(...)
	/__w/mimir/mimir/pkg/storage/ingest/writer_test.go:32
main.main()
	_testmain.go:173 +0x17a
goroutine 19 [select]:
github.com/golang/glog.(*fileSink).flushDaemon(0x25fbb38)
	/__w/mimir/mimir/vendor/github.com/golang/glog/glog_file.go:380 +0x179
created by github.com/golang/glog.init.1 in goroutine 1
[...]

PR #11397, opened 2025-05-06
Fix deadlock in TestPartitionReader_ShouldNotBufferRecordsInTheKafkaClientWhenDone

pracucci:
#### What this PR does

I looked at https://github.com/grafana/mimir/issues/11180. I found the root cause of the deadlock, the deadlock is caused by a failure in metrics assertion, for which I have no clue yet (I would need to see the actual assertion failing, instead of the test timeout panic). This PR fixes the test timeout panic, which is a first step.

**What's the change in this PR?**

The `time.After(3 * time.Second)` was impossible to reach, because `blockedTicker` was always triggering earlier (100ms). This means that if an assertion failed, we didn't interrupt the consumer after 3s, but it was looping forever.

Part of https://github.com/grafana/mimir/issues/11180

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


56quarters (approved):
Wow, very subtle. Nice find

aknuds1 (approved):

PR #11398, opened 2025-05-06
querier: fix NopDynamicReplication.MaxReplicationFactor

dimitarvdimitrov:
#### What this PR does

follow-up of #11354. Thanks @56quarters for spotting this

This method should be returning the full replication factor including the factors from the sharding ring and the dynamic replication config. Without that retries here will not work properly when dynamic replication is off

https://github.com/grafana/mimir/blob/1140f7c6191b08a3f9078f6e45af166a1cf57e77/pkg/querier/blocks_store_queryable.go#L571

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/blocks_store_replicated_set_test.go
+++ b/pkg/querier/blocks_store_replicated_set_test.go
@@ -427,7 +427,7 @@ func TestBlocksStoreReplicationSet_GetClientsFor_ShouldSupportRandomLoadBalancin
 
 	limits := &blocksStoreLimitsMock{storeGatewayTenantShardSize: 0}
 	reg := prometheus.NewPedanticRegistry()
-	s, err := newBlocksStoreReplicationSet(r, randomLoadBalancing, storegateway.NewNopDynamicReplication(), limits, grpcclient.Config{}, log.NewNopLogger(), reg)
+	s, err := newBlocksStoreReplicationSet(r, randomLoadBalancing, storegateway.NewNopDynamicReplication(3), limits, grpcclient.Config{}, log.NewNopLogger(), reg)
 	require.NoError(t, err)
 	require.NoError(t, services.StartAndAwaitRunning(ctx, s))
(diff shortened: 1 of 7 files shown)


pkg/storegateway/sharding_strategy_test.go
```diff
@@ -398,7 +398,7 @@ func TestShuffleShardingStrategy(t *testing.T) {
 			t.Cleanup(func() { assert.NoError(t, closer.Close()) })
 
 			// Initialize dynamic replication if enabled
-			var dynamicReplication DynamicReplication = NewNopDynamicReplication()
+			var dynamicReplication DynamicReplication = NewNopDynamicReplication(3)
```
56quarters, line 401:
```suggestion
			var dynamicReplication DynamicReplication = NewNopDynamicReplication(testData.replicationFactor)
```

pkg/querier/blocks_store_replicated_set_test.go
```diff
@@ -427,7 +427,7 @@ func TestBlocksStoreReplicationSet_GetClientsFor_ShouldSupportRandomLoadBalancin
 
 	limits := &blocksStoreLimitsMock{storeGatewayTenantShardSize: 0}
 	reg := prometheus.NewPedanticRegistry()
-	s, err := newBlocksStoreReplicationSet(r, randomLoadBalancing, storegateway.NewNopDynamicReplication(), limits, grpcclient.Config{}, log.NewNopLogger(), reg)
+	s, err := newBlocksStoreReplicationSet(r, randomLoadBalancing, storegateway.NewNopDynamicReplication(3), limits, grpcclient.Config{}, log.NewNopLogger(), reg)
```
56quarters, line 430: I'm not sure if it matters but it seems like this should be set from the `ringCfg` above.

56quarters (approved):
LGTM but I think there are few tests that need to be changed.

56quarters (approved):

PR #11405, opened 2025-05-06
fix: Add support for mounting TLS certificates in provisioner

### LINKED ISSUE #11400: Idea: Support mTLS for the provisioner job

When attempting to deploy the provisioner job in an environment that requires TLS between pod communication, the provisioner needs to mount certificate files to be able to validate the admin-api certificates.  The extraVolumes is missing in the chart that provides the ability to mount the certificate files.

Add extraVolumes to the provisioner job definition.

We are unable to use a different means of enforcing TLS between pods and without the ability to mount the extraVolume, there is no way to get the certificates into the running deployment.

We've already made this change in deployment_tools and tested it in our environment to confirm it works before backporting that change here.

Small (<= 1 month dev)

_No response_

### Proposer?

_No response_

rbrady:
#### What this PR does

This PR adds extraVolumes to the provisioner definition to ensure it can properly mount TLS certificates.  This PR also corrects the initContainer to use a single container with a call to /usr/bin/provisioner for each tenant, instead of an initContainer per tenant. 

Fixes #11400

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -28,6 +28,7 @@ Entries should be ordered as follows:
 Entries should include a reference to the Pull Request that introduced the change.
 
 ## main / unreleased
+
 * [BUGFIX] Added extraVolumes to provisioner to support mounting TLS certificates. #11400
 * [CHANGE] KEDA Autoscaling: Changed toPromQLLabelSelector from object to list of strings, adding support for all PromQL operators. #10945
 * [CHANGE] Memcached: Set a timeout of `500ms` for the `ruler-storage` cache instead of the default `200ms`. #11231
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -28,7 +28,7 @@ Entries should be ordered as follows:
 Entries should include a reference to the Pull Request that introduced the change.
 
 ## main / unreleased
-
+* [BUGFIX] Added extraVolumes to provisioner to support mounting TLS certificates. #11400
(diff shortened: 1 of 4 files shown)


operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -28,7 +28,7 @@ Entries should be ordered as follows:
 Entries should include a reference to the Pull Request that introduced the change.
 
 ## main / unreleased
-
+* [BUGFIX] Added extraVolumes to provisioner to support mounting TLS certificates. #11400
```
tacole02, line 31: is `extraVolumes` spelled correctly?

tacole02 (approved):

operations/helm/charts/mimir-distributed/CHANGELOG.md
dimitarvdimitrov, line 31:
```suggestion

* [BUGFIX] Added extraVolumes to provisioner to support mounting TLS certificates. #11400
```

operations/helm/charts/mimir-distributed/templates/provisioner/provisioner-job.yaml
```diff
@@ -115,4 +117,10 @@ spec:
             secretName: {{ .Values.tokengenJob.adminTokenSecret }}
         - name: bootstrap
           emptyDir: {}
-{{- end -}} 
\ No newline at end of file
+        {{- if .Values.provisioner.extraVolumes }}
```
dimitarvdimitrov, line 120: can you document this in the values.yaml, please?

rbrady: done

dimitarvdimitrov (approved):
LGTM

PR #11406, opened 2025-05-06
ingest-storage: Add consumer support for a new record version 2, based on Remote Write 2.0

alexweav:
#### What this PR does

Introduces a new record version `V2`, and implements consumer-side support for it.

It's inspired by Prometheus Remote Write 2.0, but not strictly compatible with it. A notable difference is that the symbol refs are shifted up by 64. This reserves space for a future common symbols table, allowing us to completely avoid sending certain commonly used symbols through Kafka.

The format is internal to Mimir, it doesn't need to exactly match any external standard. Aside from within test code, there is currently no way to produce records of the V2 format. Therefore, we're still able to iterate on the format in the future and make breaking changes. I'm not adding this to the changelog for that reason; when we decide to "lock down" the format, the changelog will be updated at that time.

The implementation heavily leverages @krajorama's work in https://github.com/grafana/mimir/pull/11100 and inherits several optimizations from it.

Rel https://github.com/grafana/mimir-squad/issues/2253

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -104,7 +104,7 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) (returnEr
 			}
 
 			// We don't free the WriteRequest slices because they are being freed by a level below.
-			err := DeserializeRecordContent(r.content, parsed.WriteRequest, r.version)
+			err := DeserializeRecordContent(r.content, parsed.PreallocWriteRequest, r.version)
 			if err != nil {
 				parsed.err = fmt.Errorf("parsing ingest consumer write request: %w", err)
 			}
--- a/pkg/storage/ingest/pusher.go
+++ b/pkg/storage/ingest/pusher.go
@@ -104,7 +104,7 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) (returnEr
 			}
 
 			// We don't free the WriteRequest slices because they are being freed by a level below.
-			err := parsed.Unmarshal(r.content)
+			err := DeserializeRecordContent(r.content, parsed.WriteRequest, r.version)
 			if err != nil {
 				parsed.err = fmt.Errorf("parsing ingest consumer write request: %w", err)
 			}
--- a/pkg/storage/ingest/version.go
+++ b/pkg/storage/ingest/version.go
@@ -14,6 +14,7 @@ import (
 const (
 	RecordVersionHeaderKey = "Version"
 	LatestRecordVersion    = 2
+	V2RecordSymbolOffset   = 64
 )
 
 func ValidateRecordVersion(version int) error {
@@ -105,5 +106,6 @@ func deserializeRecordContentV1(content []byte, wr *mimirpb.PreallocWriteRequest
 
 func deserializeRecordContentV2(content []byte, wr *mimirpb.PreallocWriteRequest) error {
 	wr.UnmarshalFromRW2 = true
+	wr.RW2SymbolOffset = V2RecordSymbolOffset
 	return wr.Unmarshal(content)
 }
--- a/pkg/storage/ingest/version.go
+++ b/pkg/storage/ingest/version.go
@@ -85,7 +85,7 @@ func (v versionOneRecordSerializer) ToRecords(partitionID int32, tenantID string
 	return records, nil
 }
 
-func DeserializeRecordContent(content []byte, wr *mimirpb.WriteRequest, version int) error {
+func DeserializeRecordContent(content []byte, wr *mimirpb.PreallocWriteRequest, version int) error {
 	switch version {
 	case 0:
 		// V0 is body-comptaible with V1.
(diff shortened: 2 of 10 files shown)


pkg/storage/ingest/version_test.go
```diff
@@ -44,3 +47,203 @@ func TestRecordVersionHeader(t *testing.T) {
...
+			Type:             mimirpb.COUNTER,
+			MetricFamilyName: "test_metric_total",
+			Help:             "Help for test_metric_total",
+			Unit:             "seconds",
+		}}
+		require.Equal(t, expMetadata, wr.Metadata)
+	})
+}
+
+func BenchmarkDeserializeRecordContent(b *testing.B) {
```
alexweav, line 196: For those curious, the v2 benchmark gives one extra allocation per timeseries over v1. This PR doesn't try to add any optimizations beyond what's already there. This is on par with my [initial test results](https://github.com/grafana/mimir-squad/issues/2253#issuecomment-2769535640).
```
go test ./pkg/storage/ingest/... -run ^$ -bench BenchmarkDeserializeRecordContent -benchtime=1000x -benchmem
```
```
BenchmarkDeserializeRecordContent/deserialize_v1-16         	    1000	   5242524 ns/op	12498059 B/op	   39975 allocs/op
BenchmarkDeserializeRecordContent/deserialize_v2-16         	    1000	   4888400 ns/op	12886412 B/op	   49978 allocs/op
```

pkg/storage/ingest/pusher.go
```diff
@@ -66,7 +66,7 @@ func (c pusherConsumer) Consume(ctx context.Context, records []record) (returnEr
 	}(time.Now())
 
 	type parsedRecord struct {
-		*mimirpb.WriteRequest
+		*mimirpb.PreallocWriteRequest
```
alexweav, line 69: This PR depends on https://github.com/grafana/mimir/pull/11393. I inlined those changes into this PR until that's merged.

This is logically equivalent to:
```
WriteRequest: &mimirpb.WriteRequest{
	Timeseries: mimirpb.PreallocTimeseriesSliceFromPool(),
},
```
PreallocWriteRequest.Unmarshal() calls PreallocTimeseriesSliceFromPool()

pkg/util/test/rw2.go
```diff
@@ -99,32 +99,38 @@ func AddHistogramSeries(
...
-	symbols.symbols[sym] = symbols.count
+	symbols.symbols[sym] = symbols.offset + symbols.count
 	symbols.count++
-	return symbols.count - 1
+	return symbols.offset + symbols.count - 1
 }
 
 func (symbols *SymbolTableBuilder) GetSymbols() []string {
 	res := make([]string, len(symbols.symbols))
 	for sym, i := range symbols.symbols {
```
pracucci, line 132: This logic looks broken to me. The map iteration order is random in golang. Each time you may return symbols in a different order.

pkg/storage/ingest/version.go
```diff
@@ -84,3 +85,27 @@ func (v versionOneRecordSerializer) ToRecords(partitionID int32, tenantID string
 	}
 	return records, nil
 }
+
+func DeserializeRecordContent(content []byte, wr *mimirpb.PreallocWriteRequest, version int) error {
+	switch version {
+	case 0:
+		// V0 is body-comptaible with V1.
```
pracucci, line 92: [nit]
```suggestion
		// V0 is body-compatible with V1.
```

pkg/storage/ingest/version_test.go
pracucci: Please use `benchstat` next time to compare benchmarks 😉  Run each benchmark at least 6 times to get better results from `benchstat`.

pkg/storage/ingest/version_test.go
```diff
@@ -44,3 +47,203 @@ func TestRecordVersionHeader(t *testing.T) {
...
+	require.NoError(b, err)
+
+	b.Run("deserialize v1", func(b *testing.B) {
+		for range b.N {
+			wr := &mimirpb.PreallocWriteRequest{}
+			err := DeserializeRecordContent(v1bytes, wr, 1)
+			if err != nil {
+				b.Fatal(err)
+			}
+			defer mimirpb.ReuseSlice(wr.Timeseries)
```
pracucci, line 235: What's the value of this defer?

_Same comment applies below._

pracucci (commented):
Overall LGTM (modulo a comment about the symbols builder that I think is wrong). Waiting for the rebase once the other PR is merged.

pkg/util/test/rw2.go
alexweav: That's okay; the map is (string->index). We use the map values to reconstruct the order, we don't care about the iteration. It may look like {"b": 1, "a": 0, "c": 2} and come back in that order, and we'd still put "a" in position 0.

All this modulo a shift by `offset`. It's probably a naming thing that's misleading; `i` looks like the index of the iteration but it's not.

pracucci: Sorry, I was drunk when I read the code :)

pkg/storage/ingest/version_test.go
alexweav: Nothing! copy paste :smile:

bboreham:
> symbol refs are shifted up by 64. This reserves space for a future common symbols table

* similar to: https://github.com/prometheus/prometheus/pull/15988

It's mildly illuminating to look at the relative CPU performance and choice of symbols.

PR #11409, opened 2025-05-07
MQE: clarify logic and remove outdated comments

charleskorn:
#### What this PR does

This PR makes three minor changes to MQE:

* it removes some outdated comments in `operators.mergeOneSideFloats`
* it clarifies the logic in `Query.populateVectorFromInstantVectorOperator`
* it ensures the query result is not returned to the pool multiple times if `Query.Close` is called multiple times

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/query.go
+++ b/pkg/streamingpromql/query.go
@@ -692,6 +692,12 @@ func (q *Query) populateVectorFromInstantVectorOperator(ctx context.Context, o t
 			return nil, err
 		}
 
+		if len(d.Floats)+len(d.Histograms) > 1 {
+			types.PutInstantVectorSeriesData(d, q.memoryConsumptionTracker)
+			return nil, fmt.Errorf("expected exactly one sample for series %s, but got %v floats, %v histograms", s.Labels.String(), len(d.Floats), len(d.Histograms))
+		}
+
+		// In addition to the two cases below, a series may also have no data points, in which case we don't need to do anything.
 		if len(d.Floats) == 1 && len(d.Histograms) == 0 {
 			point := d.Floats[0]
 			v = append(v, promql.Sample{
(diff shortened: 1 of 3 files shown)


pkg/streamingpromql/query.go
```diff
@@ -709,14 +709,11 @@ func (q *Query) populateVectorFromInstantVectorOperator(ctx context.Context, o t
 
 			// Remove histogram from slice to ensure it's not mutated when the slice is reused.
 			d.Histograms[0].H = nil
+		} else if len(d.Floats) == 0 && len(d.Histograms) == 0 {
+			// A series may have no data points.
+			// There's nothing to do, fall through to return 'd' to the pool below.
```
lamida, line 714: I find it is confusing to have this empty block which at the end need comment explanation like this. I suggest to rewrite by splitting different responsibilities on what different if branches have.

I see that we have if branches to check for 3 things:
- Initiate points for floats points or histogram points
- Whether to return from the loop when there are more than 1 sample for either floats or histograms
- Return d to the pool, which actually we do for all of different data size cases

There are 4 possible if conditions:
- len(d.Floats) > 1 || len(d.Histograms) > 1 --> return d to the pool and then return error
- len(d.Floats) == 1 && len(d.Histograms) == 0 --> initiate floats points
- len(d.Floats) == 0 && len(d.Histograms) == 1 --> initiate histogram points
- len(d.Floats) == 0 && len(d.Histograms) == 0 --> can be implicit 

In summary:
```
if len(d.Floats) > 1 || len(d.Histograms) > 1 {
  types.PutInstantVectorSeriesData(d, q.memoryConsumptionTracker)
  return nil, fmt.Errorf("expected exactly one sample for series %s, but got %v floats, %v histograms", s.Labels.String(), len(d.Floats), len(d.Histograms))
}

if len(d.Floats) == 1 && len(d.Histograms) == 0 {
  // initiate points like now
} else if len(d.Floats) == 0 && len(d.Histograms) == 1 {
  // initiate points like now too
}
// return d for all valid one sample cases
types.PutInstantVectorSeriesData(d, q.memoryConsumptionTracker)
```

lamida (changes requested):
I suggest to restructure the if else branches

pkg/streamingpromql/query.go
charleskorn: Good point, I've addressed this in 3ce8ab3

lamida (approved):
LGTM

PR #11411, opened 2025-05-07
distributor(ha-tracker): WatchPrefix - handle error on type assertion…

NickAnge:
… gracefully

#### What this PR does

When replica, ok := value.(*ReplicaDesc) failed to do the type assertion, we just log a warn message but return true, so WatchPrefix keep working and checking notifications.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -357,7 +357,8 @@ func (h *defaultHaTracker) loop(ctx context.Context) error {
 	h.client.WatchPrefix(ctx, "", func(key string, value interface{}) bool {
 		replica, ok := value.(*ReplicaDesc)
 		if !ok {
-			return false
+			level.Warn(h.logger).Log("msg", "failed to process value while watching for changes", "key", key)
+			return true
 		}
 		h.processKVStoreEntry(key, replica)
 		return true


pkg/distributor/ha_tracker.go
```diff
@@ -357,7 +357,8 @@ func (h *defaultHaTracker) loop(ctx context.Context) error {
 	h.client.WatchPrefix(ctx, "", func(key string, value interface{}) bool {
 		replica, ok := value.(*ReplicaDesc)
 		if !ok {
-			return false
+			level.Error(h.logger).Log("msg", "failed to process value", "return type", fmt.Sprintf("%T", value), "key", key)
```
pracucci, line 360:
```suggestion
			level.Error(h.logger).Log("msg", "unexpected data type receive when watching for HA tracker updates", "return type", fmt.Sprintf("%T", value), "key", key)
```

pracucci (approved):

pkg/distributor/ha_tracker.go
```diff
@@ -357,7 +357,8 @@ func (h *defaultHaTracker) loop(ctx context.Context) error {
...
+			return true
```
pracucci, line 361:
```suggestion
			// Always return true to ensure WatchPrefix() is never interrupted, otherwise the HA tracker stops to receive updates.
			return true
```

PR #11416, opened 2025-05-07
Add TotalSamplesPerStep stats to the MQE

Konstantinov-Innokentii:
#### What this PR does
Adds TotalSamplesPerStep stats to the MQE, simular to [prometheus implementation](https://github.com/prometheus/prometheus/pull/10369)

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/prometheus/prometheus/promql/promqltest"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/util/annotations"
+	promstats "github.com/prometheus/prometheus/util/stats"
 	"github.com/stretchr/testify/require"
 	"github.com/uber/jaeger-client-go"
 
@@ -3230,6 +3231,7 @@ func TestCompareVariousMixedMetricsComparisonOps(t *testing.T) {
 
 func TestQueryStats(t *testing.T) {
 	opts := NewTestEngineOpts()
+	opts.CommonOpts.EnablePerStepStats = true
 	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), nil, log.NewNopLogger())
 	require.NoError(t, err)
 
@@ -3249,14 +3251,14 @@ func TestQueryStats(t *testing.T) {
 			native_histogram_series {{schema:0 sum:2 count:4 buckets:[1 2 1]}} {{sum:2 count:4 buckets:[1 2 1]}}
 	`)
 
-	runQueryAndGetTotalSamples := func(t *testing.T, engine promql.QueryEngine, expr string, isInstantQuery bool) int64 {
+	runQueryAndGetSamplesStats := func(t *testing.T, engine promql.QueryEngine, expr string, isInstantQuery bool) *promstats.QuerySamples {
 		var q promql.Query
 		var err error
-
+		opts := promql.NewPrometheusQueryOpts(true, 0)
 		if isInstantQuery {
-			q, err = engine.NewInstantQuery(context.Background(), storage, nil, expr, end)
+			q, err = engine.NewInstantQuery(context.Background(), storage, opts, expr, end)
 		} else {
-			q, err = engine.NewRangeQuery(context.Background(), storage, nil, expr, start, end, time.Minute)
+			q, err = engine.NewRangeQuery(context.Background(), storage, opts, expr, start, end, time.Minute)
 		}
 
 		require.NoError(t, err)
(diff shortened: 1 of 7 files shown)


pkg/streamingpromql/types/stats.go
```diff
@@ -21,10 +21,84 @@ type QueryStats struct {
...
+}
+
+type stepStat struct {
+	T int64
+	V int64
 }
 
 const timestampFieldSize = int64(unsafe.Sizeof(int64(0)))
 
+func (qs *QueryStats) InitStepTracking(start, end, interval int64) {
```
charleskorn, line 39: Might be easier to pass the `QueryTimeRange` here, then you can use the various helper methods it provides for getting the number of steps, converting between timestamps and step indices etc.

pkg/streamingpromql/types/stats.go
```diff
@@ -21,10 +21,84 @@ type QueryStats struct {
...
+	qs.TotalSamples += samples
+
+	if qs.TotalSamplesPerStep != nil {
+		i := int((t - qs.startTimestamp) / qs.interval)
+		qs.TotalSamplesPerStep[i] += samples
+	}
+}
+
+// TODO: ikonstantinov: think about better name
+func (qs *QueryStats) NewChild(enablePerStepStats bool) *QueryStats {
```
charleskorn, line 77: Is `enablePerStepStats` here ever not the same as `qs.EnablePerStepStats`?

pkg/streamingpromql/operators/subquery.go
```diff
@@ -115,9 +126,36 @@ func (s *Subquery) NextStepSamples() (*types.RangeVectorStepData, error) {
...
+		temp := s.samplesProcessedInChildQuery(s.stepData.RangeStart, s.stepData.RangeEnd)
+		s.ParentQueryStats.IncrementSamplesAtTimestamp(s.stepData.StepT, temp)
+	}
+
 	return s.stepData, nil
 }
 
+// samplesProcessedInChildQuery returns the number of samples processed in child query that corresponds to the current parent query step.
+func (s *Subquery) samplesProcessedInChildQuery(start, end int64) int64 {
+	if s.ChildQueryStats == nil || s.ChildQueryStats.TotalSamplesPerStep == nil {
```
charleskorn, line 139: We don't need this check here and on line 129.

pkg/streamingpromql/operators/subquery.go
```diff
@@ -115,9 +126,36 @@ func (s *Subquery) NextStepSamples() (*types.RangeVectorStepData, error) {
...
+		return 0
+	}
+
+	sum := int64(0)
+	childStartT := s.ChildQueryTimeRange.StartT
+	childInterval := s.ChildQueryTimeRange.IntervalMilliseconds
+	totalChildSteps := len(s.ChildQueryStats.TotalSamplesPerStep)
+	
+	for i := 0; i < totalChildSteps; i++ {
```
charleskorn, line 148: I think we can be smarter here: we can determine the first value of `i` based on the value of `start`, and determine when to terminate the loop based on the value of `end` (rather than always starting at `i == 0` and advancing until we reach `start`, then breaking when we reach `end`)

pkg/streamingpromql/query.go
```diff
@@ -868,13 +878,27 @@ func (q *Query) Statement() parser.Statement {
 	return q.statement
 }
 
-func (q *Query) Stats() *stats.Statistics {
-	return &stats.Statistics{
-		Timers: stats.NewQueryTimers(),
-		Samples: &stats.QuerySamples{
-			TotalSamples: q.stats.TotalSamples,
+// TODO: ikonstantinov: with the PerStepStats introduction, Stats doesn't really implement the same interface as Prometheus' Stats, because it doesn't report PerStepStats properly
```
charleskorn, line 881: Can you elaborate on this?

pkg/streamingpromql/types/stats.go
Konstantinov-Innokentii: If there is no concern of keeping implementation simular to prometheus, it sounds like a good idea, thanks

pkg/streamingpromql/query.go
Konstantinov-Innokentii: MQE `Query.Stats()` returns `promstats.QuerySamples`. ` promstats.QuerySamples` has a TotalSamplesPerStepMap method, which returns PerStepStats in` {Timestamp: Value,..}` format.  Take a look at the code from prometheus:
```
func (qs *QuerySamples) TotalSamplesPerStepMap() *TotalSamplesPerStep {
	if !qs.EnablePerStepStats {
		return nil
	}

	ts := TotalSamplesPerStep{}
	for _, s := range qs.totalSamplesPerStepPoints() {
		ts[s.T] = int(s.V)
	}
	return &ts
}

func (qs *QuerySamples) totalSamplesPerStepPoints() []stepStat {
	if !qs.EnablePerStepStats {
		return nil
	}

	ts := make([]stepStat, len(qs.TotalSamplesPerStep))
	for i, c := range qs.TotalSamplesPerStep {
		ts[i] = stepStat{T: qs.startTimestamp + int64(i)*qs.interval, V: c}
	}
	return ts
}
```
Problem here, is that it's uses qs.startTimestamp and qs.interval fields to build a TotalSamplesPerStepMap. Fields are unexported, so we can't set them directly in the  `Query.Stats()` method in MQE. Only way to set them is to call `InitStepTracking`, but it will allocate an array we don't need, because stats are already counted, we just want to convert them to prometheus type to satisfy Query interface.
```
func (qs *QueryStats) InitStepTracking(start, end, interval int64) {
	if !qs.EnablePerStepStats {
		return
	}

	numSteps := int((end-start)/interval) + 1
	qs.TotalSamplesPerStep = make([]int64, numSteps)
	qs.startTimestamp = start
	qs.interval = interval
}
```
So, that's why I'm saying that it's not really implements Stats() method, because currently we can't return correct TotalSamplesPerStepMap.

pkg/streamingpromql/types/stats.go
charleskorn: Right, makes sense, thanks for clarifying.

pkg/streamingpromql/types/stats.go
charleskorn: Everything else has diverged, so we might as well make our lives as easy as possible 🙂

pkg/streamingpromql/query.go
charleskorn: Right, thanks for explaining. I think what you've got in this PR is fine.

Konstantinov-Innokentii: I think I'll stick to not comparing the timestamps in the TestQueryStatisticsUpstream to make life bit easier, if it's fine with you.

pkg/streamingpromql/operators/subquery.go
```diff
@@ -115,9 +123,34 @@ func (s *Subquery) NextStepSamples() (*types.RangeVectorStepData, error) {
...
 }
 
+// samplesProcessedInSubquery returns the number of samples processed in a time range selected by parent query step from subquery results.
+func (s *Subquery) samplesProcessedInSubquery(start, end int64) int64 {
+	if s.subqueryStats == nil || s.subqueryStats.TotalSamplesPerStep == nil || len(s.subqueryStats.TotalSamplesPerStep) == 0 {
+		return 0
+	}
+
+	startIndex := 0
+	if start >= s.SubqueryTimeRange.StartT {
```
charleskorn, line 138: Why do we need this condition? (same applies for check on `endIndex >= len(s.subqueryStats.TotalSamplesPerStep)` below)

pkg/streamingpromql/operators/subquery.go
```diff
@@ -63,6 +69,8 @@ func (s *Subquery) NextSeries(ctx context.Context) error {
 	// Release the previous series' slices now, so we are likely to reuse the slices for the next series.
 	s.floats.Release()
 	s.histograms.Release()
+	// Clean the subquery stats to not to double count samples from previous series in a next series.
```
charleskorn, line 72: [nit]

```suggestion
	// Clear the subquery stats to not to double count samples from previous series in a next series.
```

pkg/streamingpromql/types/stats.go
```diff
@@ -21,10 +21,63 @@ type QueryStats struct {
 	// For example, if a query is running with a step of 30s with a range vector selector with range 45s,
 	// then samples in the overlapping 15s are counted twice.
 	TotalSamples int64
+
+	TotalSamplesPerStep []int64
+
+	EnablePerStepStats bool
+	timerange          QueryTimeRange
```
charleskorn, line 28: [nit] 
```suggestion
	timeRange          QueryTimeRange
```

pkg/streamingpromql/types/stats.go
```diff
@@ -21,10 +21,63 @@ type QueryStats struct {
...
+}
+
+type stepStat struct {
+	T int64
+	V int64
 }
 
 const timestampFieldSize = int64(unsafe.Sizeof(int64(0)))
```
charleskorn, line 37: [nit] I would move this down to be just above `EquivalentFloatSampleCount`, to keep the related bits of code next to each other.

pkg/streamingpromql/types/stats.go
```diff
@@ -21,10 +21,63 @@ type QueryStats struct {
...
 	// then samples in the overlapping 15s are counted twice.
 	TotalSamples int64
+
+	TotalSamplesPerStep []int64
+
+	EnablePerStepStats bool
+	timerange          QueryTimeRange
+}
+
+type stepStat struct {
```
charleskorn, line 31: Is this struct used anywhere?

pkg/streamingpromql/engine_test.go
```diff
@@ -3266,100 +3274,677 @@ func TestQueryStats(t *testing.T) {
...
+			require.Equal(t, testCase.expectedTotalSamples, prometheusSamplesStats.TotalSamples, "invalid test case: expected total samples does not match value from Prometheus' engine")
+			require.Equal(t, testCase.expectedTotalSamplesPerStep, prometheusSamplesStats.TotalSamplesPerStep, "invalid test case: expected per stepsamples does not match value from Prometheus' engine")
+
+			mimirSamplesStats := runQueryAndGetSamplesStats(t, mimirEngine, testCase.expr, testCase.isInstantQuery)
+			require.Equal(t, testCase.expectedTotalSamples, mimirSamplesStats.TotalSamples)
+			require.Equal(t, testCase.expectedTotalSamplesPerStep, mimirSamplesStats.TotalSamplesPerStep)
+
+			mimirSamplesStatsWithPlanning := runQueryAndGetSamplesStats(t, mimirEngineWithPlanning, testCase.expr, testCase.isInstantQuery)
+			require.Equal(t, testCase.expectedTotalSamples, mimirSamplesStatsWithPlanning.TotalSamples)
+			require.Equal(t, testCase.expectedTotalSamplesPerStep, mimirSamplesStatsWithPlanning.TotalSamplesPerStep)
```
charleskorn, line 3419: [nit] Might be good to add some messages to these assertions to clarify which are with planning enabled and which are without.

pkg/streamingpromql/engine_test.go
```diff
@@ -3266,100 +3274,677 @@ func TestQueryStats(t *testing.T) {
...
+
+		return q.Stats().Samples
+	}
+
+	testCases := map[string]struct {
+		expr                        string
+		isInstantQuery              bool
+		expectedTotalSamples        int64
+		expectedTotalSamplesPerStep []int64
+		SkipCompareWithPQE          bool
```
charleskorn, line 3476: [nit] This is the convention we've used elsewhere:
```suggestion
		skipCompareWithPrometheus          bool
```

We've also made it a string in some places, to force ourselves to explain why we're skipping the comparison.

pkg/streamingpromql/engine_test.go
```diff
@@ -3266,100 +3274,677 @@ func TestQueryStats(t *testing.T) {
...
+
+		res := q.Exec(context.Background())
+		require.NoError(t, res.Err)
+
+		return q.Stats().Samples
+	}
+
+	cases := []struct {
+		Query               string
+		SkipMaxCheck        bool
```
charleskorn, line 3614: Is this field (`SkipMaxCheck`) used?

charleskorn (commented):
Thanks for working on this!

pkg/streamingpromql/operators/subquery.go
Konstantinov-Innokentii: For the startIndex check, it's a `dense_series{}[1m:5m]` case.
ParentQueryTimeRange is (600.000, 600.000). It gives a stepData Range (540.000, 600.000) 
But a SubQueryTimeRange is 600.000, 600.000. Which will give the incorrect start index when doing
`int((start-s.SubqueryTimeRange.StartT)/s.SubqueryTimeRange.IntervalMilliseconds) + 1`.

For the endIndex I don't have a case, but clamping it just to make sure we are always withing a stats array boundaries.

pkg/streamingpromql/operators/subquery.go
```diff
@@ -115,9 +123,34 @@ func (s *Subquery) NextStepSamples() (*types.RangeVectorStepData, error) {
...
+
+	startIndex := 0
+	if start >= s.SubqueryTimeRange.StartT {
+		startIndex = int((start-s.SubqueryTimeRange.StartT)/s.SubqueryTimeRange.IntervalMilliseconds) + 1
+	}
+
+	endIndex := int((end - s.SubqueryTimeRange.StartT) / s.SubqueryTimeRange.IntervalMilliseconds)
+	if endIndex >= len(s.subqueryStats.TotalSamplesPerStep) {
+		endIndex = len(s.subqueryStats.TotalSamplesPerStep) - 1
+	}
```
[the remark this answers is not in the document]
Konstantinov-Innokentii: Yep, looks like we can, thanks

pkg/streamingpromql/operators/subquery.go
```diff
@@ -115,9 +123,34 @@ func (s *Subquery) NextStepSamples() (*types.RangeVectorStepData, error) {
...
 	s.stepData.RangeEnd = rangeEnd
 
+	s.parentQueryStats.IncrementSamplesAtTimestamp(s.stepData.StepT, s.samplesProcessedInSubquery(s.stepData.RangeStart, s.stepData.RangeEnd))
+
 	return s.stepData, nil
 }
 
+// samplesProcessedInSubquery returns the number of samples processed in a time range selected by parent query step from subquery results.
+func (s *Subquery) samplesProcessedInSubquery(start, end int64) int64 {
+	if s.subqueryStats == nil || s.subqueryStats.TotalSamplesPerStep == nil || len(s.subqueryStats.TotalSamplesPerStep) == 0 {
```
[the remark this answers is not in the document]
charleskorn: If we expect that `s.subqueryStats` will never be nil, then I think we should remove these checks - otherwise they might hide a bigger problem.

pkg/streamingpromql/types/operator.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/prometheus/prometheus/promql/parser/posrange"
 )
 
+type PrepareParams struct {
+	QueryStats *QueryStats
+}
+
+func (p *PrepareParams) Clone() PrepareParams {
+	return PrepareParams{}
```
charleskorn: Shouldn't this set `QueryStats`?

pkg/streamingpromql/types/operator.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/prometheus/prometheus/promql/parser/posrange"
 )
 
+type PrepareParams struct {
+	QueryStats *QueryStats
+}
+
+func (p *PrepareParams) Clone() PrepareParams {
```
charleskorn: Given we always work with a pointer to `PrepareParams`, it might be nicer to return a `*PrepareParams` here.

It'd be good to then introduce a test that `Clone` returns an equivalent value (ie. all fields have the same values), but is a different instance of `PrepareParams`.

pkg/streamingpromql/types/stats.go
```diff
@@ -21,6 +24,75 @@ type QueryStats struct {
...
+func NewQueryStats(timeRange QueryTimeRange, enablePerStepStats bool, memoryConsumptionTracker *limiter.MemoryConsumptionTracker) (*QueryStats, error) {
+	qs := &QueryStats{
+		EnablePerStepStats:       enablePerStepStats,
+		timeRange:                timeRange,
+		memoryConsumptionTracker: memoryConsumptionTracker,
+	}
+	if enablePerStepStats {
+		s, err := Int64SlicePool.Get(qs.timeRange.StepCount, qs.memoryConsumptionTracker)
+		if err != nil {
+			return nil, fmt.Errorf("failed to allocate TotalSamplesPerStep: %w", err)
```
charleskorn: There's no need to wrap the error here, and in fact we shouldn't, as this message will be returned to the user, and `TotalSamplesPerStep` won't mean anything to them. The error message returned by `Get` will be user-friendly already.

pkg/streamingpromql/types/stats.go
```diff
@@ -21,6 +24,75 @@ type QueryStats struct {
...
+	}
+}
+
+// Clear resets the TotalSamples counter to 0 and zeroes out all entries in the
+// TotalSamplesPerStep slice, preserving its length and capacity for reuse.
+func (qs *QueryStats) Clear() {
+	qs.TotalSamples = 0
+	if qs.TotalSamplesPerStep != nil {
+		clear(qs.TotalSamplesPerStep)
+	}
```
charleskorn: I don't think we need the `if` here, [the spec for `clear`](https://go.dev/ref/spec#Clear) says it'll work fine with `nil` slices:

>If the map or slice is `nil`, `clear` is a no-op.

pkg/streamingpromql/operators/subquery.go
```diff
@@ -130,6 +163,17 @@ func (s *Subquery) ExpressionPosition() posrange.PositionRange {
 	return s.expressionPosition
 }
 
+func (s *Subquery) Prepare(params types.PrepareParams) {
+	s.parentQueryStats = params.QueryStats
+	// Child Samples will be counted later when processing subquery results.
+	s.subqueryStats = types.NewQueryStats(s.SubqueryTimeRange, true)
+
+	paramsCopy := types.PrepareParams{
```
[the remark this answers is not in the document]
charleskorn: Fair point, happy for you to keep things as-is if you'd prefer.

pkg/streamingpromql/engine_test.go
```diff
@@ -3292,100 +3306,641 @@ func TestQueryStats(t *testing.T) {
...
+		require.NoError(t, res.Err)
+
+		return q.Stats().Samples
+	}
+
+	cases := []struct {
+		Query                       string
+		Start                       time.Time
+		End                         time.Time
+		Interval                    time.Duration
```
charleskorn: [nit] We've tended not to expose these fields in other similar test cases:

```suggestion
		query                       string
		start                       time.Time
		end                         time.Time
		interval                    time.Duration
```

pkg/streamingpromql/engine_test.go
```diff
@@ -3292,100 +3306,641 @@ func TestQueryStats(t *testing.T) {
...
+		expectedTotalSamples        int64
+		expectedTotalSamplesPerStep []int64
+		// ...WithPlanner expectations are optional and should be set only if query with planning reports different stats
+		expectedTotalSamplesWithPlanner        int64
+		expectedTotalSamplesPerStepWithPlanner []int64
```
charleskorn, line 3630: Could you please add some test cases that involve common subexpression elimination but don't involve subqueries? eg. something like `sum(foo) + sum(foo)`

pkg/streamingpromql/query.go
```diff
@@ -81,12 +81,17 @@ func (e *Engine) newQuery(ctx context.Context, queryable storage.Queryable, opts
 		return nil, fmt.Errorf("could not get memory consumption limit for query: %w", err)
 	}
 
+	memoryConsumptionTracker := limiter.NewMemoryConsumptionTracker(maxEstimatedMemoryConsumptionPerQuery, e.queriesRejectedDueToPeakMemoryConsumption)
+	stats, err := types.NewQueryStats(timeRange, e.enablePerStepStats && opts.EnablePerStepStats(), memoryConsumptionTracker)
+	if err != nil {
+		return nil, fmt.Errorf("could not get create stats for query: %w", err)
```
charleskorn: Same comment as in `stats.go`: there's no need to wrap the error here, and it might lead to a confusing error message for the user as well.

pkg/streamingpromql/operators/subquery.go
charleskorn: > For the startIndex check, it's a `dense_series{}[1m:5m]` case.

Right, in that case, what if we did something like `start = max(start, s.SubqueryTimeRange.StartT)`? Then we could always use `PointIndex` below.

(It'd also be good to add a comment here either way so I don't ask the same question again in two months' time 🙂)

>For the endIndex I don't have a case, but clamping it just to make sure we are always withing a stats array boundaries.

If there's not a use case where we need to clamp it, I'd prefer not to clamp it, as that may hide a bug elsewhere.

pkg/streamingpromql/types/operator.go
Konstantinov-Innokentii: Oh, it's just a leftover, I decided not to use Clone() at all, keeping things simple for now,  thanks for catching

pkg/streamingpromql/types/operator.go
Konstantinov-Innokentii: Same as above

pkg/streamingpromql/query.go
Konstantinov-Innokentii: I have two questions:
1. We are wrapping error just above:
```
maxEstimatedMemoryConsumptionPerQuery, err := e.limitsProvider.GetMaxEstimatedMemoryConsumptionPerQuery(ctx)
if err != nil {
	return nil, fmt.Errorf("could not get memory consumption limit for query: %w", err)
}
```
Should we "unwrap" it then?

2. Does it make sense to log something in this case? Right now, we only fail to create QueryStats if the estimated memory is exceeded. Should we log an error when that happens to make it easier to debug potential query failures due to high memory usage?

charleskorn: > 1. We are wrapping error just above:

This is a different scenario: in this case, we failed to retrieve the memory consumption limit for the query, which should never happen in practice. So wrapping the error here is fine and helpful, as we can't guarantee the returned error has enough context about the problem.

> 2. Does it make sense to log something in this case?

We haven't done this elsewhere. I'd be very surprised if a query failed at this point due to the memory consumption limit - this is typically set to 1 GB and so the query would need to be for a large number of steps to hit the limit here.

pkg/streamingpromql/operators/subquery.go
```diff
@@ -130,8 +167,24 @@ func (s *Subquery) ExpressionPosition() posrange.PositionRange {
 	return s.expressionPosition
 }
 
+func (s *Subquery) Prepare(ctx context.Context, params *types.PrepareParams) error {
+	s.parentQueryStats = params.QueryStats
+	// Final aggregation into parent stats happens in samplesProcessedInSubquery().
+	// Child samples always counted per step to count parent stats correctly.
+	childStats, err := types.NewQueryStats(s.SubqueryTimeRange, true, s.memoryConsumptionTracker)
+	if err != nil {
+		return fmt.Errorf("could not create stats for subquery: %w", err)
```
charleskorn: We shouldn't wrap this error, for the same reasons as elsewhere: the only error `NewQueryStats` can return is a limit exceeded error, and so we want to return that as-is to users.

charleskorn (approved):
LGTM modulo comment below and [this comment](https://github.com/grafana/mimir/pull/11416#discussion_r2119812558) which I believe is unresolved

pkg/streamingpromql/engine_test.go
Konstantinov-Innokentii: It's [there](https://github.com/grafana/mimir/pull/11416/files#diff-df3817badf68588818b73b443096f0edd778d0019a2bc135e6a7f6eeb218946cR3514). I added test to our testcases, not upstream

pkg/streamingpromql/operators/subquery.go
Konstantinov-Innokentii: pushed the fix

pkg/streamingpromql/engine_test.go
charleskorn: Completely missed that, thanks.

PR #11417, opened 2025-05-07
mqe: Allow usage of Mimir Query Engine in query-frontend

56quarters:
#### What this PR does

Introduce configuration to allow the Mimir Query Engine to be used instead of the Prometheus engine in the query-frontend when queries are rewritten to be shardable or subqueries are spun off.

Notable changes:

* New configuration for query-frontend to allow MQE to be used with an optional fallback to the Prometheus engine (same as in queriers).
* Context timeouts are now checked at least once in `Selector` and also every 128 series instead of only every 128 series.
* Query-frontend tests that used a query engine now run with both Prometheus and MQE engines. Results are expected to be the same in most cases, excluding some minor differences like error messages.

N/A

#### Notes to reviewers

There are a lot of whitespace changes in test files: click "hide whitespace changes" when reviewing.

#### Checklist

- [X] Tests updated.
- [x] Documentation added.
- [X] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

### DIFF
--- a/pkg/frontend/querymiddleware/querysharding.go
+++ b/pkg/frontend/querymiddleware/querysharding.go
@@ -39,7 +39,7 @@ const shardingTimeout = 10 * time.Second
 type querySharding struct {
 	limit Limits
 
-	engine            *promql.Engine
+	engine            promql.QueryEngine
 	next              MetricsQueryHandler
 	logger            log.Logger
 	maxSeriesPerShard uint64
@@ -62,7 +62,7 @@ type queryShardingMetrics struct {
 // Finally we can translate the embedded vector selector back into subqueries in the Queryable and send them in parallel to downstream.
 func newQueryShardingMiddleware(
 	logger log.Logger,
-	engine *promql.Engine,
+	engine promql.QueryEngine,
 	limit Limits,
 	maxSeriesPerShard uint64,
 	registerer prometheus.Registerer,
@@ -157,7 +157,7 @@ func (s *querySharding) Do(ctx context.Context, r MetricsQueryRequest) (Response
 	return ExecuteQueryOnQueryable(ctx, r, s.engine, shardedQueryable, annotationAccumulator)
 }
 
-func ExecuteQueryOnQueryable(ctx context.Context, r MetricsQueryRequest, engine *promql.Engine, queryable storage.Queryable, annotationAccumulator *AnnotationAccumulator) (Response, error) {
+func ExecuteQueryOnQueryable(ctx context.Context, r MetricsQueryRequest, engine promql.QueryEngine, queryable storage.Queryable, annotationAccumulator *AnnotationAccumulator) (Response, error) {
 	qry, err := newQuery(ctx, r, engine, lazyquery.NewLazyQueryable(queryable))
 	if err != nil {
 		return nil, apierror.New(apierror.TypeBadData, err.Error())
@@ -205,7 +205,7 @@ func ExecuteQueryOnQueryable(ctx context.Context, r MetricsQueryRequest, engine 
 	}, nil
 }
 
-func newQuery(ctx context.Context, r MetricsQueryRequest, engine *promql.Engine, queryable storage.Queryable) (promql.Query, error) {
+func newQuery(ctx context.Context, r MetricsQueryRequest, engine promql.QueryEngine, queryable storage.Queryable) (promql.Query, error) {
 	switch r := r.(type) {
 	case *PrometheusRangeQueryRequest:
 		return engine.NewRangeQuery(
--- a/pkg/frontend/querymiddleware/querysharding_test.go
+++ b/pkg/frontend/querymiddleware/querysharding_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/grafana/mimir/pkg/mimirpb"
 	"github.com/grafana/mimir/pkg/querier"
 	"github.com/grafana/mimir/pkg/storage/sharding"
+	"github.com/grafana/mimir/pkg/streamingpromql/compat"
 	"github.com/grafana/mimir/pkg/util"
 )
 
@@ -680,7 +681,7 @@ func TestQuerySharding_Correctness(t *testing.T) {
 
 	// Add native histogram series.
 	for i := 0; i < numNativeHistograms; i++ {
-		gen := factor(float64(i) * 0.1)
+		gen := factor(float64(i) * 0.5)
 		if i >= numNativeHistograms-numStaleNativeHistograms {
 			// Wrap the generator to inject the staleness marker between minute 10 and 20.
 			gen = stale(start.Add(10*time.Minute), start.Add(20*time.Minute), gen)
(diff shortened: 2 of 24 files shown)


pkg/frontend/querymiddleware/querysharding.go
```diff
@@ -238,7 +238,12 @@ func newQuery(ctx context.Context, r MetricsQueryRequest, engine *promql.Engine,
 			r.GetQuery(),
 			util.TimeFromMillis(r.GetStart()).Add(1*time.Nanosecond),
 			util.TimeFromMillis(r.GetEnd()),
-			time.Duration(r.GetStep())*time.Millisecond,
+			// TODO(56quarters): Need to actually figure out what the Prometheus engine
```
charleskorn, line 241: Why do we need to handle remote read requests here at all? As far as I can see, this method is only used for the sharding and subquery spin-off middlewares, which should only be used for range and instant queries.

pkg/mimir/modules.go
```diff
@@ -797,16 +800,39 @@ func (t *Mimir) initQueryFrontendTopicOffsetsReaders() (services.Service, error)
...
 
-	engineOpts, _ := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, t.ActivityTracker, util_log.Logger, promqlEngineRegisterer)
+		if t.Cfg.Querier.EnableQueryEngineFallback {
+			eng = streamingpromqlcompat.NewEngineWithFallback(streamingEngine, promql.NewEngine(promOpts), nil, util_log.Logger)
+		} else {
+			eng = streamingEngine
+		}
+	default:
+		panic(fmt.Sprintf("invalid config not caught by validation: unknown PromQL engine '%s'", t.Cfg.Querier.QueryEngine))
+	}
```
charleskorn, line 826: We have similar logic in `querier.New` - should we extract this to a shared method somewhere?

pkg/frontend/querymiddleware/querysharding_test.go
```diff
@@ -1635,48 +1640,83 @@ func TestQuerySharding_ShouldReturnErrorInCorrectFormat(t *testing.T) {
...
+			engineType:         querier.PrometheusEngine,
+			engineShardingOpts: []engineOpt{withMaxSamples(1)},
+			queryable:          queryablePrometheusExecErr,
+			expError:           apierror.Newf(apierror.TypeExec, "expanding series: %s", querier.NewMaxQueryLengthError(744*time.Hour, 720*time.Hour)),
 		},
+
+		// MQE equivalents when applicable. For example, MQE doesn't have a sample limit
+		// and returns different error messages for timeouts when executing a query.
+
 		{
```
56quarters: This test currently fails because of the way that timeouts are checked in the Prometheus engine vs the Mimir engine. In Prometheus, the context is checked for timeouts [after the query has been evaluated](https://github.com/grafana/mimir/blob/cc1c7bfb93e2e4774b58cfc68ff48fedd7265c12/vendor/github.com/prometheus/prometheus/promql/engine.go#L824-L826) and a few other places. In contrast, we only check for timeouts [every 128 series](https://github.com/grafana/mimir/blob/cc1c7bfb93e2e4774b58cfc68ff48fedd7265c12/pkg/streamingpromql/operators/selectors/selector.go#L110-L115). I'm not sure what to do about this. Should we check for timeouts in more places? Remove the test?

pkg/frontend/querymiddleware/querysharding_test.go
```diff
@@ -1635,48 +1640,83 @@ func TestQuerySharding_ShouldReturnErrorInCorrectFormat(t *testing.T) {
...
-			name:             "downstream - storage internal error",
-			engineDownstream: engine,
-			engineSharding:   engineSampleLimit,
-			queryable:        queryableInternalErr,
-			expError:         apierror.New(apierror.TypeInternal, "some internal error"),
+			name:               "sharding - timeout",
+			engineType:         querier.MimirEngine,
+			engineShardingOpts: []engineOpt{withTimeout(50 * time.Millisecond)},
+			expError:           apierror.New(apierror.TypeTimeout, "context deadline exceeded"),
```
56quarters: ~This fails because we return a different error compared to the Prometheus engine. I think we should probably just update the test to the expected MQE error message?~ Nevermind, forgot I already updated the message.

pkg/mimir/modules.go
56quarters: I was avoiding this because of the panic and the number of parameters required means this would be kinda gross. I can see how it looks.

pkg/frontend/querymiddleware/querysharding_test.go
56quarters: Another option would be to move our existing check for context cancellation before we increment the series index. That guarantees we check the context at least once for all selectors _and_ after 128 series. WDYT?

pkg/frontend/querymiddleware/querysharding.go
56quarters: Fixed by you in #11583

pkg/frontend/querymiddleware/querysharding_test.go
charleskorn: Swapping the order so we check for context cancellation before we increment the series index sounds good to me.

56quarters: > Another option would be to move our existing check for context cancellation before we increment the series index. That guarantees we check the context at least once for all selectors _and_ after 128 series. WDYT?

This is the approach I went with.

pkg/frontend/querymiddleware/querysharding_test.go
```diff
@@ -1635,48 +1637,79 @@ func TestQuerySharding_ShouldReturnErrorInCorrectFormat(t *testing.T) {
...
-			expError:         apierror.New(apierror.TypeTimeout, "query timed out in expression evaluation"),
-			queryable:        queryableSlow,
+			name:       "downstream - storage prometheus execution error",
+			engineType: querier.PrometheusEngine,
+			queryable:  queryablePrometheusExecErr,
+			expError:   apierror.Newf(apierror.TypeExec, "expanding series: %s", querier.NewMaxQueryLengthError(744*time.Hour, 720*time.Hour)),
+		},
+
+		// MQE equivalents when applicable. For example, MQE doesn't have a sample limit
+		// and returns different error messages for timeouts when executing a query.
```
charleskorn: Rather than duplicating the tests like this, what if we added a flag to disable the test cases that aren't relevant for MQE, and to provide alternative error messages if MQE is being used?

pkg/frontend/querymiddleware/spin_off_subqueries_test.go
```diff
@@ -24,8 +24,6 @@ import (
 )
 
 func TestSubquerySpinOff_Correctness(t *testing.T) {
-	t.Parallel()
-
```
charleskorn, line 28: Is there a reason why you've removed this? (same applies to other places where `t.Parallel()` has been removed)

charleskorn (commented):
Overall LGTM.

Do we need to do anything to support the `X-Mimir-Force-Prometheus-Engine` header?

pkg/frontend/querymiddleware/spin_off_subqueries_test.go
56quarters: Locally, they didn't make things faster and I was having trouble getting the spin-off tests to complete without timing out.

56quarters: I reduced the data used in spin-off tests so I'll add this back now.

56quarters:
> Do we need to do anything to support the `X-Mimir-Force-Prometheus-Engine` header?

I didn't know about that header so we probably do :facepalm: . I'll look into it.

pkg/frontend/querymiddleware/querysharding_test.go
56quarters: I've added a flag to skip MQE for the sample test and stop comparing the bodies generated by the errors (this was not a great idea anyway) and instead now just compare the error types and HTTP status codes.

charleskorn (approved):

pkg/mimir/modules.go
```diff
@@ -796,16 +799,39 @@ func (t *Mimir) initQueryFrontendTopicOffsetsReaders() (services.Service, error)
 // to optimize Prometheus query requests.
 func (t *Mimir) initQueryFrontendTripperware() (serv services.Service, err error) {
 	promqlEngineRegisterer := prometheus.WrapRegistererWith(prometheus.Labels{"engine": "query-frontend"}, t.Registerer)
+	promOpts, mqeOpts := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, t.ActivityTracker, util_log.Logger, promqlEngineRegisterer)
+	// Disable concurrency limits for sharded queries spawned by the query-frontend.
+	promOpts.ActiveQueryTracker = nil
```
charleskorn, line 804: We should also set `mqeOpts.EagerLoadSelectors = true` here now that https://github.com/grafana/mimir/pull/11624 is merged.

56quarters: Will do that now :+1:

PR #11429, opened 2025-05-08
Distributor: simplify code for validateSeries

bboreham:
#### What this PR does

Simplifies the code.
The returned `bool shouldRemove` was only ever true if the returned `err` was non-nil, so we don't need the `bool`.

#### Checklist

- [x] Tests updated.
- NA Documentation added.


narqo (approved):
The new one looks more logical and is simpler, indeed 🔥

aknuds1 (approved):
Makes perfect sense, thanks!

PR #11437, opened 2025-05-11
Query-frontend: Avoid some re-parsing of PromQL

bboreham:
#### What this PR does

Avoid turning the parsed expression into a string only to immediately parse it again.

Based on observations like this, which is a heap alloc_space profile of query-frontend, that `ParseExpr` was getting called over and over again.

image

Here's `GetQuery` from the same profile:

image

So this change should save ~20% of garbage.

#### Checklist

- NA Tests updated.
- NA Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/remote_read.go
+++ b/pkg/frontend/querymiddleware/remote_read.go
@@ -282,6 +282,16 @@ func (r *remoteReadQueryRequest) GetQuery() string {
 	return r.promQuery
 }
 
+func (r *remoteReadQueryRequest) GetParsedQuery() parser.Expr {
+	if r.promQuery != "" {
+		expr, err := parser.ParseExpr(r.promQuery)
+		if err == nil {
+			return expr
+		}
+	}
+	return nil
+}
+
 func (r *remoteReadQueryRequest) GetHeaders() []*PrometheusHeader {
 	return nil
 }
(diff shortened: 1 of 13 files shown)


pkg/frontend/querymiddleware/remote_read.go
```diff
@@ -282,6 +282,16 @@ func (r *remoteReadQueryRequest) GetQuery() string {
 	return r.promQuery
 }
 
+func (r *remoteReadQueryRequest) GetParsedQuery() parser.Expr {
+	if r.promQuery != "" {
+		expr, err := parser.ParseExpr(r.promQuery)
+		if err == nil {
```
bboreham, line 288: I think this is not expected to error because we generate the PromQL.
Perhaps someone can think of a better way to check?

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -2025,13 +2024,14 @@ func Test_evaluateAtModifier(t *testing.T) {
...
 	} {
 		t.Run(tt.in, func(t *testing.T) {
 			t.Parallel()
+			expr, err := parser.ParseExpr(tt.in)
+			require.NoError(t, err)
 			expectedExpr, err := parser.ParseExpr(tt.expected)
 			require.NoError(t, err)
-			out, err := evaluateAtModifierFunction(tt.in, start, end)
+			out, err := evaluateAtModifierFunction(expr, start, end)
 			if tt.err != nil {
```
bboreham, line 2035: There are now no test cases where the function errors.

56quarters (approved):
Thank you, this has been driving me nuts but I never got around to fixing it.

PR #11448, opened 2025-05-12
query-frontend: Move step alignment metric to step align middleware

56quarters:
#### What this PR does

Move the metric emitted for non-step aligned queries to the middleware responsible for adjusting query start and end times when enabled.

N/A

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/step_align.go
+++ b/pkg/frontend/querymiddleware/step_align.go
@@ -18,26 +18,32 @@ import (
 )
 
 type stepAlignMiddleware struct {
-	next    MetricsQueryHandler
-	limits  Limits
-	logger  log.Logger
-	aligned *prometheus.CounterVec
+	next       MetricsQueryHandler
+	limits     Limits
+	logger     log.Logger
+	notAligned prometheus.Counter
+	adjusted   *prometheus.CounterVec
 }
 
 // newStepAlignMiddleware creates a middleware that aligns the start and end of request to the step to
 // improve the cacheability of the query results based on per-tenant configuration.
 func newStepAlignMiddleware(limits Limits, logger log.Logger, registerer prometheus.Registerer) MetricsQueryMiddleware {
-	aligned := promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
-		Name: "cortex_query_frontend_queries_step_aligned_total",
+	notAligned := promauto.With(registerer).NewCounter(prometheus.CounterOpts{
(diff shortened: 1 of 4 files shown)


pkg/frontend/querymiddleware/step_align.go
```diff
@@ -48,33 +54,35 @@ func (s *stepAlignMiddleware) Do(ctx context.Context, r MetricsQueryRequest) (Re
...
-
-			updatedReq, err := r.WithStartEnd(start, end)
-			if err != nil {
-				return nil, err
-			}
-
-			return s.next.Do(ctx, updatedReq)
+	start := (r.GetStart() / r.GetStep()) * r.GetStep()
+	end := (r.GetEnd() / r.GetStep()) * r.GetStep()
+	stepAligned := start == r.GetStart() && end == r.GetEnd()
```
charleskorn, line 59: [nit] Might be clearer to do something like `if stepAligned { return s.next.Do(ctx, r) }` here to avoid the need for the conditions below

charleskorn (approved):

pkg/frontend/querymiddleware/step_align.go
56quarters: I've reversed the conditionals so we return early in all cases unless we are adjusting the start/end.

pkg/frontend/querymiddleware/step_align.go
```diff
@@ -18,26 +18,32 @@ import (
...
+	logger     log.Logger
+	notAligned prometheus.Counter
+	adjusted   *prometheus.CounterVec
 }
 
 // newStepAlignMiddleware creates a middleware that aligns the start and end of request to the step to
 // improve the cacheability of the query results based on per-tenant configuration.
 func newStepAlignMiddleware(limits Limits, logger log.Logger, registerer prometheus.Registerer) MetricsQueryMiddleware {
-	aligned := promauto.With(registerer).NewCounterVec(prometheus.CounterOpts{
-		Name: "cortex_query_frontend_queries_step_aligned_total",
```
dimitarvdimitrov, line 31: is this metric now gone? so you replaced the positive one with a negative one? wasn't clear from the description. Might make sense to mention it in a changelog entry too

dimitarvdimitrov: oh, it was renamed to `cortex_query_frontend_non_step_aligned_queries_adjusted_total`, not "reversed"

PR #11456, opened 2025-05-14
Make `maxLimit` configurable per tenant for the `/cardinality/label_names` and `/cardinality/label_values` endpoints.

zhuoyuan-liu:
#### What this PR does

This PR introduces a new configurable parameter `cardinality_api_max_series_limit` to make the maximum number of series returned by cardinality API requests ( `/cardinality/label_names` and `/cardinality/label_values` ) configurable on a per-tenant basis.

The changes include:

1. Adding a new field `CardinalityAPIMaxSeriesLimit` to the `Limits` struct in the cardinality section
2. Registering a new flag `querier.cardinality-api-max-series-limit` with a default value of 500 (which was the previous hard-coded value)
3. Adding a getter method `CardinalityAPIMaxSeriesLimit` to the `Overrides` struct
4. Modifying the cardinality handlers to pass the tenant ID and overrides to the request decoders
5. Updating the `extractLimit` function to use the tenant-specific limit instead of the hard-coded value

This PR is based on the feedback from https://github.com/grafana/mimir/pull/11025#pullrequestreview-2735479417

Fixes https://github.com/grafana/mimir/issues/10328

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 * [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. Limitations: Created timestamp is ignored, per series metadata is merged on metric family level automatically, ingestion might fail if client sends ProtoBuf fields out of order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0` depending on the detected Remote-Write protocol. #11100 #11101 #11192 #11143
 * [FEATURE] Query-frontend: expand `query-frontend.cache-errors` and `query-frontend.results-cache-ttl-for-errors` configuration options to cache non-transient response failures for instant queries. #11120
 * [FEATURE] Querier, query-frontend, ruler: Enable experimental support for duration expressions in PromQL, which are simple arithmetics on numbers in offset and range specification. #11344
+* [FEATURE] Querier: Make the max series limit for cardinality API requests configurable on a per-tenant basis with the `cardinality_api_max_series_limit` option. #11456
```
tacole02, line 17:
```suggestion
* [FEATURE] Querier: Make the maximum series limit for cardinality API requests configurable on a per-tenant basis with the `cardinality_api_max_series_limit` option. #11456
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3706,6 +3706,11 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -querier.label-values-max-cardinality-label-names-per-request
 [label_values_max_cardinality_label_names_per_request: <int> | default = 100]
 
+# Maximum number of series that can be requested in a single cardinality API
```
tacole02, line 3709:
```suggestion
# Maximum number of series that you can request in a single cardinality API
```

tacole02 (approved):

zhuoyuan-liu:
Hi @narqo @pracucci , could you please take a look at this PR? This PR is based on the feedback from https://github.com/grafana/mimir/pull/11025#pullrequestreview-2735479417

LasseHels:
@narqo @pracucci @dimitarvdimitrov Bump. We'd be happy to get a review of this :heart:

LasseHels:
@narqo @npazosmendez @56quarters Bump. Any chance of a review on this?

pkg/util/validation/limits.go
```diff
@@ -203,6 +203,7 @@ type Limits struct {
 	CardinalityAnalysisEnabled                    bool `yaml:"cardinality_analysis_enabled" json:"cardinality_analysis_enabled"`
 	LabelNamesAndValuesResultsMaxSizeBytes        int  `yaml:"label_names_and_values_results_max_size_bytes" json:"label_names_and_values_results_max_size_bytes"`
 	LabelValuesMaxCardinalityLabelNamesPerRequest int  `yaml:"label_values_max_cardinality_label_names_per_request" json:"label_values_max_cardinality_label_names_per_request"`
+	CardinalityAPIMaxSeriesLimit                  int  `yaml:"cardinality_api_max_series_limit" json:"cardinality_api_max_series_limit"`
```
56quarters: Can you add an `experimental` tag here so that we can make changes before finalizing this setting?

56quarters: The limit isn't really limiting the number of series returned. Can you change the setting (and CLI flag, and methods) to `cardinality_analysis_max_results`.

56quarters (commented):
LGTM with some small changes. Thanks!

pkg/cardinality/request.go
```diff
@@ -65,16 +67,21 @@ func (r *LabelNamesRequest) String() string {
...
 	if err := r.ParseForm(); err != nil {
 		return nil, err
 	}
 
-	return DecodeLabelNamesRequestFromValues(r.Form)
+	return DecodeLabelNamesRequestFromValuesWithUser(r.Form, userID, overrides)
 }
 
 // DecodeLabelNamesRequestFromValues is like DecodeLabelNamesRequest but takes url.Values in input.
 func DecodeLabelNamesRequestFromValues(values url.Values) (*LabelNamesRequest, error) {
```
dimitarvdimitrov, line 77: this is used in caching, but the actual handler that serves the request uses `DecodeLabelNamesRequest`. I think this means that we will store the request under the same cache key with or without the limit. I think this will lead to confusing results (for example serving 500 results to a query which explicitly requested 1000 results)

zhuoyuan-liu: Hi @dimitarvdimitrov,
Thanks for the comments — I see the issue now: the cache doesn’t take tenant-specific limits into account.

One approach would be to update the current cache generator to use new decoder functions that are tenant-aware and include tenant-specific limits. However, this might require introducing a new decoder specifically for cardinality limits, instead of using the default decoder.

Another option would be to revisit the previous PR, since this configuration only controls how much data is returned to the user — not the actual cardinality calculation.

Could you please take another look at the previous https://github.com/grafana/mimir/pull/11025#pullrequestreview-2735479417?

pkg/cardinality/request.go
```diff
@@ -140,16 +147,21 @@ func (r *LabelValuesRequest) String() string {
...
+	return DecodeLabelValuesRequestFromValuesWithUser(r.Form, userID, overrides)
 }
 
 // DecodeLabelValuesRequestFromValues is like DecodeLabelValuesRequest but takes url.Values in input.
 func DecodeLabelValuesRequestFromValues(values url.Values) (*LabelValuesRequest, error) {
+	return DecodeLabelValuesRequestFromValuesWithUser(values, "", nil)
+}
+
+// DecodeLabelValuesRequestFromValuesWithUser is like DecodeLabelValuesRequestFromValues but also accepts the userID and overrides.
+func DecodeLabelValuesRequestFromValuesWithUser(values url.Values, userID string, overrides *validation.Overrides) (*LabelValuesRequest, error) {
```
dimitarvdimitrov: here can we just take a limit instead of accepting the overrides? the limit we take will be the upper bound of the limit in the request. Same with `DecodeLabelValuesRequest`

pkg/cardinality/request.go
dimitarvdimitrov: actually i thought this is worse than it actually is. Today the cache key doesn't include the limit in it. I think it should. I think that would be a good-enough fix for now. The trick is that after increasing/decreasing the tenant limit, the cache entry will be stale. That's ok.

you didn't change the cache key in this PR, so i'm fine with doing this in a follow-up PR. I can open it.

dimitarvdimitrov: i'm wrong again: the limit _is_ in the cache  key https://github.com/grafana/mimir/blob/a979e2b34f638528ca48f781ae08b0aceed675b5/pkg/cardinality/request.go#L59-L61

so this is basically ok

dimitarvdimitrov:
you also need to run `make reference-help doc` to fix some of the unit tests and the linter. The integration test is a flake

pkg/cardinality/request.go
zhuoyuan-liu: Hi @dimitarvdimitrov,
I believe the issue is that the cache doesn’t take tenant-specific (max) limits into account. The code above reflects limits coming from user input. So even if we replace the overrides with an int, the cache function still won’t be able to retrieve the (max) limits from the request as it doesn't include the overrides config.

pkg/cardinality/request.go
dimitarvdimitrov: does this make sense @zhuoyuan-liu? can you do it?

zhuoyuan-liu: Hi @dimitarvdimitrov,
Thanks for the review, and I tried to make this change and realized the following problem:

For each of the requests, we can get the userID and the user limits, but not the max limits. So, even though we changed the function to accept the maxlimit, for the cache, it cannot get the maxlimit.

pkg/cardinality/request.go
```diff
@@ -228,7 +237,7 @@ func extractSelector(values url.Values) (matchers []*labels.Matcher, err error)
 }
 
 // extractLimit parses and validates request param `limit` if it's defined, otherwise returns default value.
-func extractLimit(values url.Values, userID string, overrides *validation.Overrides) (limit int, err error) {
+func extractLimit(values url.Values, requestLimit int) (limit int, err error) {
```
dimitarvdimitrov: nitpick on naming: this should be something like tenantMaxLimit. It's not the limit in the request, it's the limit that will bound the request

pkg/cardinality/request.go
```diff
@@ -72,16 +72,21 @@ func DecodeLabelNamesRequest(r *http.Request, userID string, overrides *validati
 		return nil, err
 	}
 
-	return DecodeLabelNamesRequestFromValuesWithUser(r.Form, userID, overrides)
+	requestLimit := 0
+	if overrides != nil {
+		requestLimit = overrides.CardinalityAPIMaxSeriesLimit(userID)
+	}
+
```
dimitarvdimitrov: sorry i wasn't clear enough. I mean move this overrides checking outside of the decoding logic. For example do it in `LabelNamesCardinalityHandler`  and in `LabelValuesCardinalityHandler`

pkg/cardinality/request.go
dimitarvdimitrov: yes, you are correct. but i think this is ok - changing the tenant overrides, doesn't invalidate the cache and risks returning stale and incorrect results for the time being. Do you disagree?

pkg/cardinality/request.go
dimitarvdimitrov: I think we're mixing the two threads. I replied regarding the caching in the other thread

pkg/cardinality/request.go
zhuoyuan-liu: Hi @dimitarvdimitrov ,
I haven’t looked into the cache logic in detail, so please correct me if I’m wrong.

One edge case I’m thinking about is when we increase the max limit to, say, 5000, and a request comes in with a limit of 1000. However, when the request hits the cache, it might throw a parsing error if the cache still defaults to a max limit of 500.

Otherwise, I already fixed rest of the comments

dimitarvdimitrov: wouldn't we just serve the cached result with the 500 results?

zhuoyuan-liu: Good point. I believe that we shouldn't cache the result if it's a single query with a large amount of data. 
Could you please review and let me know if there is anything I need to update or change?

dimitarvdimitrov (dismissed):
LGTM

Thank you for your effort on this @zhuoyuan-liu and apologies for the huge delay.

dimitarvdimitrov:
there is one more thing that I forgot - can you push a change to the ./CHANGELOG.md and about-versioning.md? This is an enhancement and it would be nice to mention it to end users/operators in releases

zhuoyuan-liu:
Thanks, I think the Changelog was accidentally removed while rebasing the branch. I have added them back. https://github.com/grafana/mimir/pull/11456/commits/7b6d3ea31ea06ec062e745c863867ac625f56f1a.

dimitarvdimitrov (approved):
Thank you again!

PR #11462, opened 2025-05-16
HA Tracker: fix old entries with zero valued ElectedAt timestamp

pracucci:
#### What this PR does

In this PR there are two changes:

1. Simplify the `defaultHaTracker.updateKVStore()` `else` branch, removing a superfluous extra condition (the only case where the code wouldn't enter the extra condition `if desc == nil || (desc.DeletedAt > 0)` is if `DeletedAt` is negative, which can never happen
2. Add a small logic to set `ElectedAt` timestamp in case its value is zero. In the past we had a bug causing zero value to be stored. If a replica has never changed since then, the `ElectedAt` timestamp is still zero (we've seen it in prod). In this case we can just set it to "now": even if it's not true that it has been elected "now" (it was in the past), we do fix it to avoid misunderstandings when you look at the HA tracker web page.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


NickAnge (approved):

colega (approved):

duricanikolic (approved):
LGTM
Thank you

PR #11464, opened 2025-05-16
Wait for HA tracker's updateKVLoop to finish with a context

colega:
#### What this PR does

When [a bug](https://github.com/grafana/mimir/pull/11463) occurs, we shouldn't wait forever on a waitgroup silently. This will stop waiting after a reasonable amount of time and will stop the service.

Follow up on https://github.com/grafana/mimir/pull/11463

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


colega:
I don't think this is worth a changelog entry, but I will create one if the reviewers think it's needed.

pkg/distributor/ha_tracker.go
```diff
@@ -371,7 +372,12 @@ func (h *defaultHaTracker) loop(ctx context.Context) error {
 		return true
 	})
```
pracucci, line 374: [nit] Above, when we call `WatchPrefix()` I would add a comment to explain that it blocks until the context has done or the callback function returns false (which can't happen because we never return false).

pracucci, line 374: [nit] If the code reach this point and the context has not done, then we have an unexpected issue. We could check it and log an error log.

pracucci (approved):

pkg/distributor/ha_tracker.go
colega: I added a comment there.

colega: I can add something in a separate PR.

colega:
@pracucci I changed the code a little bit, do you mind taking another look before I merge?

pkg/util/sync/wait_with_timeout.go
```diff
@@ -0,0 +1,24 @@
...
+func WaitWithTimeout(wg *sync.WaitGroup, timeout time.Duration) error {
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-time.After(timeout):
+		return fmt.Errorf("WaitGroup.Wait() timed out after %s", timeout)
```
dimitarvdimitrov, line 20: there's a race between waiting the wait group and returning an error. not sure if this won't cause problems. maybe document that WaitWithTimeout may return an error even if the WaitGroup.Wait() returns before the timeout expires

colega: Ugh... I don't think that's worth documenting, just the call to `WaitWithTimeout` already makes it being imprecise. The idea of waiting on waitgroup with a timeout is to avoid blocking forever on something we expect to take a limited amount of time, not to be precise on how long it took (waitgroups would be a bad choice for that)

pkg/distributor/ha_tracker.go
```diff
@@ -371,7 +373,11 @@ func (h *defaultHaTracker) loop(ctx context.Context) error {
...
 
-	wg.Wait()
+	if err := mimirsync.WaitWithTimeout(&wg, 10*time.Minute); err != nil {
```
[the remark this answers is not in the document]
colega: Sure, let's make it 10s.
```suggestion
	if err := mimirsync.WaitWithTimeout(&wg, 10*time.Second); err != nil {
```

Anyway the point here is not to handle the process termination stage, but "we shouldn't be here" scenario. But I agree that in any case we should be done within a few millis so 10s should work.

PR #11490, opened 2025-05-19
[WIP] bring in prometheus/parquet-common code to new package

francoposa:
replace efficient-go errors with pkg/errors; satisfy mimir-prometheus ChunkSeries interface

TODO license headers

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


aknuds1 (commented):
I think it would be preferable to use Renovate's PRs for updating [objstore](https://github.com/grafana/mimir/pull/10664) and [cloud.google.com/go/storage](https://github.com/grafana/mimir/pull/11470). Plus, I don't think this PR should be removing the `toolchain` directive and its explanatory comment.

pkg/parquet/search/parquet_queriable_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/thanos-io/objstore/providers/filesystem"
+
+	"github.com/grafana/mimir/pkg/parquet/schema"
+	"github.com/grafana/mimir/pkg/parquet/storage"
+)
+
+func TestPromQLAcceptance(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping, because 'short' flag was set")
+	}
```
npazosmendez, line 33: This test took ~15min in the CI. Is that a concern? Do we want to make it opt-in rather than opt-out? I'm mostly thinking about the impact on other mimir devs, not us

npazosmendez (approved):

go.mod
```diff
@@ -1,10 +1,6 @@
 module github.com/grafana/mimir
 
-go 1.23.8
-
-// Please note that this directive is ignored when building with the Mimir build image,
-// that will always use its bundled toolchain.
-toolchain go1.24.2
+go 1.24.2
```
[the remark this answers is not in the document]
francoposa: @aknuds1 I bumped the toolchain version up to 1.24.2 since that is what we are actually using, and then go mod tidy removes it because it matches the main go version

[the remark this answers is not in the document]
aknuds1: @francoposa OK then it makes sense :) Bit of a pain IMO that `go mod tidy` can remove it, so we lose our documentation of how it's ignored by our build process.

francoposa:
@aknuds1 The toolchain directive came back anyway when I copied over the go.mod from main to fix conflicts, so we should be good to go

I also used the env var approach to prevent the super long PromQL acceptance tests from running in CI

pkg/parquet/search/parquet_queriable_test.go
56quarters: Our CI health is already pretty terrible. This should be opt-in, 15m is a non-starter.

francoposa: It's fixed already to be opt-in with env var

PR #11501, opened 2025-05-21
MQE: enable by default for queriers

charleskorn:
#### What this PR does

This PR enables MQE by default for queriers.

Note that I've left all of the MQE-related CLI flags as experimental, given we'll likely remove them at some point and only allow using MQE.

I've also removed the integration test that covers the case where the "max samples" limit is hit introduced in #499. MQE does not support this limit, and the functionality is covered by unit tests that were also introduced in that PR.

#10067

#### Checklist

- [n/a] Tests updated.
- [x] Documentation added.


56quarters (approved):
Makes sense to me :tada:

56quarters (approved):
LGTM

PR #11502, opened 2025-05-21
Upgrade Go requirement to v1.24.2 - fix build

aknuds1:
#### What this PR does

Build is currently failing, as gRPC v1.72.1 requires Go v1.24. To fix the build, I'm upgrading the Go requirement to v1.24.2.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


go.mod
```diff
@@ -163,9 +159,11 @@ require (
 	github.com/sercand/kuberesolver/v6 v6.0.0 // indirect
 	github.com/shopspring/decimal v1.2.0 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
+	github.com/spiffe/go-spiffe/v2 v2.5.0 // indirect
 	github.com/tklauser/go-sysconf v0.3.12 // indirect
 	github.com/tklauser/numcpus v0.6.1 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
+	github.com/zeebo/errs v1.4.0 // indirect
```
charleskorn, line 166: Why are these changes required?

go.mod
```diff
@@ -1,10 +1,6 @@
 module github.com/grafana/mimir
 
-go 1.23.8
-
-// Please note that this directive is ignored when building with the Mimir build image,
-// that will always use its bundled toolchain.
-toolchain go1.24.2
+go 1.24.3
```
charleskorn, line 3: Should this match the Go version in the build image?

And should we keep the `toolchain` line?

aknuds1: > Should this match the Go version in the build image?

Ah yes, nicely spotted. 

> And should we keep the toolchain line?

I would prefer it, although `toolchain` is dropped by `go mod tidy` automatically when upgrading.

go.mod
aknuds1: They are made automatically by `go mod tidy`. It makes some vendoring changes somehow because `gRPC` v1.72.1 is simply broken with Go v1.23.

go.mod
aknuds1: I tried adding it back, but `go mod tidy` strips it away consistently.

go.mod
```diff
@@ -1,10 +1,6 @@
 module github.com/grafana/mimir
 
-go 1.23.8
-
-// Please note that this directive is ignored when building with the Mimir build image,
-// that will always use its bundled toolchain.
-toolchain go1.24.2
+go 1.24.2
```
aknuds1, line 3: `go mod tidy` consistently removes `toolchain` :(

go.mod
aknuds1: Fixed.

go.mod
krajorama: I think toolchain is redundant if the version is the same as the Go version.

aknuds1: Yes, I'm just unhappy to lose the explanatory comment for when `toolchain` gets reintroduced.

krajorama (approved):
LGTM

PR #11504, opened 2025-05-21
fix(alertmanager): fix typos in alertmanager template test

timonegk:
#### What this PR does

Fixes two typos in `alertmanager_template_test.go`.

#### Checklist

- [x] Tests updated.
- [ ] ~Documentation added.~
- [ ] ~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.~

### DIFF
--- a/pkg/alertmanager/alertmanager_template_test.go
+++ b/pkg/alertmanager/alertmanager_template_test.go
@@ -47,7 +47,7 @@ func Test_withCustomFunctions(t *testing.T) {
 					GeneratorURL: "http://localhost:9090?foo=bar",
 				},
 			},
-			template:    `{{ queryFormGeneratorURL (index .Alerts 0).GeneratorURL }}`,
+			template:    `{{ queryFromGeneratorURL (index .Alerts 0).GeneratorURL }}`,
 			expectError: true,
 		},
 		{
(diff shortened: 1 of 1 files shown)


dimitarvdimitrov (commented):
thanks for the contribution! Nice catch. I'm a bit worried that the test was passing even with the wrong template. Is there anything we can do to make the test fail without your change? perhaps have stronger assertions in the test loop?

dimitarvdimitrov:
we're curious - how did you come across this?

timonegk:
> Is there anything we can do to make the test fail without your change? perhaps have stronger assertions in the test loop?

As far as I can see from the testify docs, you could use [`EqualError`](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualError) to assert that a specific error is returned instead of asserting that any error is returned. If you want, I can update the PR accordingly.

> we're curious - how did you come across this?

I wanted to use `queryFromGeneratorURL` and `grafanaExploreURL` in a template and was looking through the code to find if I could also reference data sources by name instead of by id. While looking at the examples in the test I found the typos.

dimitarvdimitrov:
> As far as I can see from the testify docs, you could use [`EqualError`](https://pkg.go.dev/github.com/stretchr/testify/assert#EqualError) to assert that a specific error is returned instead of asserting that any error is returned. If you want, I can update the PR accordingly.

that would be great, thank you!

timonegk:
Hi @dimitarvdimitrov, I changed the test to use `assert.ErrorContains`. I also had to change the test "error on URL decoding query in GeneratorURL" because the test actually only failed because of the typo before.
When I change the template to include the typo again, the tests now fail, for example with the error `Error "template: :1: function \"queryFormGeneratorURL\" not defined" does not contain "query not found in the generator URL"`.

dimitarvdimitrov (approved):
thanks! LGTM

PR #11508, opened 2025-05-21
Alerts: Increase Threshold on DistributorGcUsesTooMuchCpu

dmwilson-grafana:
#### What this PR does

We've found that the current threshold for the `DistributorGcUsesTooMuchCpu` alert is unnecessarily low. The alert currently fires when `go_cpu_classes_gc_total_cpu_seconds_total / (go_cpu_classes_total_cpu_seconds_total - go_cpu_classes_idle_cpu_seconds_total)` exceeds 0.1, this PR sets this value to be configurable with a default of 0.1. In practice, we see Mimir cells have near-constant push and forwarding latencies, and inflight requests until around 15% GC CPU.

Consider the following plot of performance metrics vs. GC CPU:

Screenshot 2025-05-21 at 12 58 12 PM

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pr00se (approved):

lindeskar:
Hi and thank you for adding this!

Any chance the default value can be raised to 15%? As you mention the Distributors seem to hover between 10-15% even with low pressure.

PR #11520, opened 2025-05-24
Block-builder-scheduler: fix bug where startup job is erroneously skipped based on commit movement

seizethedave:
#### What this PR does

This PR fixes a bug in handling "initial" jobs computed at startup. Here's an example of the bug:

1. scheduler knows that its committed offset is 10.
1. a job J1 is computed over offsets (10, 20) and is assigned to a worker 1. it is in progress.
1. the scheduler restarts and learns about J1 in progress.
1. this new scheduler scans for initial jobs for all partitions, and it computes a job J2 (10, 30)
1. we only allow one outstanding job per partition so J2 goes into a feeder queue.
1. worker 1 completes J1. scheduler moves the committed offset to 20.
1. Now that there's room in the job queue, we can move J2 over there. But, there was a bug where we were [skipping feeder-queue jobs if `j.start < committed`](https://github.com/grafana/mimir/blob/cd4127dd571cc03c6edb3277ccd086b1e0045365/pkg/blockbuilder/scheduler/scheduler.go#L307). (then we'd lose offsets 20 to 30.) It should instead skip the feeder queue job if `j.end <= committed`.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/scheduler/scheduler.go
+++ b/pkg/blockbuilder/scheduler/scheduler.go
@@ -304,7 +304,7 @@ func (s *BlockBuilderScheduler) enqueuePendingJobs() {
 			// The job discovery process happens concurrently with ongoing job
 			// completions. Now that we have the lock, ignore this job if it's
 			// older than our committed offset.
-			if c, ok := s.committed.Lookup(s.cfg.Kafka.Topic, partition); ok && or.start < c.At {
+			if c, ok := s.committed.Lookup(s.cfg.Kafka.Topic, partition); ok && or.end <= c.At {
 				ps.pendingJobs.Remove(e)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -304,7 +304,7 @@ func (s *BlockBuilderScheduler) enqueuePendingJobs() {
 			// The job discovery process happens concurrently with ongoing job
 			// completions. Now that we have the lock, ignore this job if it's
 			// older than our committed offset.
-			if c, ok := s.committed.Lookup(s.cfg.Kafka.Topic, partition); ok && or.start < c.At {
+			if c, ok := s.committed.Lookup(s.cfg.Kafka.Topic, partition); ok && or.end <= c.At {
 				ps.pendingJobs.Remove(e)
```
narqo, line 308: How about we add an info log line here, so this behaviour was easier to debug in the future

narqo (approved):
Awesome, thank you for fixing it 🔥

PR #11545, opened 2025-05-26
ruler: Terminate on failires during initial sync

narqo:
#### What this PR does

If the ruler fails to complete its initial sync, it still transitions to the running state. In this case, the tenants, whose rules have been synced, observe a delay in their rule evaluation.

This PR addresses that by making sure the ruler fails to start on a failure during initial sync. During a rolling update, if a failure happens, the running rulers will continue to evaluate the rules, until the new replica manages to transition into the running state.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/ruler.go
+++ b/pkg/ruler/ruler.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
+	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/concurrency"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/grpcclient"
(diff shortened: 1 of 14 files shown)


tacole02 (approved):

pkg/ruler/ruler.go
```diff
@@ -539,37 +549,62 @@ func (r *Ruler) run(ctx context.Context) error {
...
+		}
+
+		lastErr = r.syncRules(ctx, nil, rulerSyncReasonInitial, true)
+		if lastErr == nil {
+			return nil
+		}
+		boff.Wait()
+	}
+	if lastErr == nil {
+		return boff.Err()
```
dimitarvdimitrov, line 596: can we wrap the error again here?

pkg/ruler/ruler_test.go
```diff
@@ -1208,6 +1211,63 @@ func TestRuler_NotifySyncRulesAsync_ShouldTriggerRulesSyncingOnAllRulersWhenEnab
 	}
 }
 
+func TestRuler_InitialSyncFails(t *testing.T) {
```
dimitarvdimitrov, line 1214: technically we're not testing that it retries. is it difficult to add that test too?

dimitarvdimitrov (commented):
looks good

PR #11556, opened 2025-05-27
Revert "Use OTel library from tracing while keeping Jaeger exporter config"

colega:
Reverts grafana/mimir#11249

Tracing doesn't seem to work properly, I overlooked something while testing previously. Need to revert while I'm investigating as there are quite a lot of changes here to just keep moving forward.


colega:
No need to revert, it was an issue in another place

PR #11569, opened 2025-05-28
ruler: ignore rulers in non-operation states in forEachRulerInTheRing

fionaliao:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/golang/snappy v1.0.0
 	github.com/google/gopacket v1.1.19
 	github.com/gorilla/mux v1.8.1
-	github.com/grafana/dskit v0.0.0-20250508185919-68d09ac9016e
+	github.com/grafana/dskit v0.0.0-20250521185944-1f3d678623e9
 	github.com/grafana/e2e v0.1.2-0.20250428181430-708d63bcc673
 	github.com/hashicorp/golang-lru v1.0.2 // indirect
 	github.com/influxdata/influxdb/v2 v2.7.11
--- a/go.sum
+++ b/go.sum
@@ -551,8 +551,8 @@ github.com/grafana-tools/sdk v0.0.0-20220919052116-6562121319fc h1:PXZQA2WCxe85T
 github.com/grafana-tools/sdk v0.0.0-20220919052116-6562121319fc/go.mod h1:AHHlOEv1+GGQ3ktHMlhuTUwo3zljV3QJbC0+8o2kn+4=
 github.com/grafana/alerting v0.0.0-20250506135548-3c7640592aa8 h1:z4FqXZJvYrt5HV0glX6eIf40E62mS7tdAF2zJ2Y0W4g=
 github.com/grafana/alerting v0.0.0-20250506135548-3c7640592aa8/go.mod h1:pMfhRxL2LZ3Pm8iy7VcVsb9CLYuBtjFYbf1oxgx7yFA=
-github.com/grafana/dskit v0.0.0-20250508185919-68d09ac9016e h1:9JDfa1EkyMFR4410dfZURYNGxPJYNYer9C90rh8nJ/k=
-github.com/grafana/dskit v0.0.0-20250508185919-68d09ac9016e/go.mod h1:aT6thv8yV0shmqbbRbtEwxK9a+H/8fWDOMm2IRx8xx0=
+github.com/grafana/dskit v0.0.0-20250521185944-1f3d678623e9 h1:R7NWRuAC9dVex4WA1V1dSrCZlN8ssyrNjnMBUXb9iMw=
+github.com/grafana/dskit v0.0.0-20250521185944-1f3d678623e9/go.mod h1:qAaZ2E+2GyDJVJVOz5wibs3mt+Uq/c4W+py9WNQMUu0=
 github.com/grafana/e2e v0.1.2-0.20250428181430-708d63bcc673 h1:Va04sDlP33f1SFUHRTho4QJfDlGTz+/HnBIpYwlqV9Q=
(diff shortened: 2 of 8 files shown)


pkg/ruler/ruler_test.go
```diff
@@ -623,39 +687,56 @@ func TestGetRules(t *testing.T) {
 
 			// Sync rules on each ruler.
 			for _, r := range rulerAddrMap {
-				r.syncRules(ctx, nil, rulerSyncReasonInitial, true)
+				// Use rulerSyncReasonPeriodic to ensure rules are distributed among ACTIVE rulers
+				r.syncRules(ctx, nil, rulerSyncReasonPeriodic, true)
```
fionaliao, line 691: TODO: use rulerSyncReasonPeriodic for ACTIVE rulers, otherwise reason initial to properly simular what really happens

pkg/ruler/ruler_test.go
```diff
@@ -447,62 +451,248 @@ func compareRuleGroupDescToStateDesc(t *testing.T, expected *rulespb.RuleGroupDe
...
+				),
+				// While ruler4 registers some tokens for all users, it doesn't actually evaluate any rules.
+				// ruler1 and ruler2 are selected for subring for user1 and user2, even though ruler2 is in the JOINING
+				// state (this means all the rules are evaluated on ruler1).
+				// This does form part of user3's subring but user3's single rule is evaluated on ruler3.
+				"ruler4": append(
+					[]uint32{userToken("user1", 2) + 1, userToken("user2", 2) + 1, userToken("user3", 1) + 1},
+					generateTokenForGroups(2, rules[2])...,
+				),
+			},
```
fionaliao, line 602: Wanted to check if the result of this test case is expected - when getting the shuffle sharded subring for a user, non-ACTIVE rulers are **not** filtered out. Rules do end up being evaluated on the ACTIVE rulers in the subring though. 

The issue I see is that it's possible during rollout or scaling for all the rulers in a shuffle sharded subring to be in a non-ACTIVE state, in which case all rules will not be evaluated for a user.

This was happening before my PR as well. This could be fixed (if it is an issue) by moving `GetSubringForOperationStates(op)` before calling `ShuffleShard()` on the ring, so the shuffle sharded ring is made from only the instances in the requested state. Would need to be careful to update the logic in all the places that gets the shuffled sharded subring.

pkg/ruler/ruler_test.go
```diff
@@ -526,12 +716,11 @@ func TestGetRules(t *testing.T) {
...
 
-			for rID, r := range expectedRulesByRuler {
+			activeRulers := 0
+			for rID := range tc.tokensByRuler {
 				createAndStartRuler(rID)
-				for user, rules := range r {
-					allRulesByUser[user] = append(allRulesByUser[user], rules...)
-					allRulesByRuler[rID] = append(allRulesByRuler[rID], rules...)
-					allTokensByRuler[rID] = generateTokenForGroups(rules, 1)
+				if tc.rulerState[rID] == ring.ACTIVE {
```
pracucci: [nit] The `rulerState` contract is that if it's not specified then it should be considered ACTIVE. Here we don't check specifically if the state is set of a given ruler. However, it works anyway because `ring.ACTIVE` value is `0`, and if the ruler state is not specified in `rulerState` then we just get the zero value, which is `0`. Don't know if this was intentional or not. It works, but maybe we could make it even more clear by checking if the state is explicitly set in `rulerState`, like you do it below.

pkg/ruler/ruler_test.go
```diff
@@ -556,41 +749,96 @@ func TestGetRules(t *testing.T) {
...
+					require.Equal(t, len(userRules), len(actualRules.Groups), "rules are not equal for %s, %s", u, rID)
+
+					var actualRuleDescs []*rulespb.RuleGroupDesc
+					for _, g := range actualRules.Groups {
+						actualRuleDescs = append(actualRuleDescs, g.Group)
+					}
+					sort.Slice(actualRuleDescs, func(i, j int) bool {
+						return actualRuleDescs[i].Name < actualRuleDescs[j].Name
+					})
+					require.Equal(t, expectedRuleDescs, actualRuleDescs)
```
pracucci: [nit] I think `expectedRuleDescs` is sorted nearly by a coincidence (due to how fixtures are created, in order). To keep the test a bit stronger, you may consider removing the sorting at all, and use `require.ElementsMatch()` (it asserts that 2 slices contain the same elements, even if in different order).

```suggestion
					require.ElementsMatch(t, expectedRuleDescs, actualRuleDescs)
```

pkg/ruler/ruler_test.go
pracucci: I was looking exactly at this test and then I noticed your comment. I'm glad we both put our 👀 on this.

I agree this behaviour in related from your PR. Is it desired? It depends. Generally speaking, this is the expected behaviour of shuffle sharding in ingesters (where shuffle sharding has been introduced at the beginning): the goal is to not reshard series among different ingesters during a rolling update. I also think it's a desired property for the store-gateways shuffle sharding.

Is it desired for rulers? I don't know right now (I need to think about it more). Synching rules take time. Given rulers unregister from the ring during a restart, changing this logic would realistically just cover the time it takes for the initial sync (when the ruler is in JOINING state). Moving rule groups to the remaining ACTIVE ruler in the shard (if any) or to another ruler outside the original tenant's shard may not make much difference. The only real difference may be in case the shard size = 1 (if the shard is composed by the JOINING ruler, then no rules would be evaluated while JOINING). However, synching rules on other rules will take time too, so not sure would be a significant improvement changing this logic.

PR #11575, opened 2025-05-28
Grafana Alertmanager: Fix template validation on Apply for empty receivers

JacobsonMT:
Empty receivers were defaulting to using Mimir AM template functions for parsing. This causes issues when Grafana-only template functions are defined.

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -637,7 +637,7 @@ func (am *Alertmanager) buildIntegrationsMap(emailCfg alertingReceivers.EmailSen
 	for _, rcv := range nc {
 		var integrations []*nfstatus.Integration
 		var err error
-		switch rcvType := rcv.Type(); rcvType {
+		switch rcv.Type() {
 		case definition.GrafanaReceiverType:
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/alertmanager.go
```diff
@@ -637,7 +637,8 @@ func (am *Alertmanager) buildIntegrationsMap(emailCfg alertingReceivers.EmailSen
 	for _, rcv := range nc {
 		var integrations []*nfstatus.Integration
 		var err error
-		if rcv.Type() == definition.GrafanaReceiverType {
+		rcvType := rcv.Type()
```
santihernandezc, line 640: Wdyt about using a `switch` statement with cases for Grafana and Mimir receivers, adding a `default` statement with a comment that makes it clear that we're not building a `tmpl` object for empty receivers?

```go
switch rcv.Type() {
    case GrafanaReceiverType:
        // Logic here...
    case AlertmanagerReceiverType:
        // Logic here...
    default:
        // Comment explaining that we're not creating a tmpl for this case.
}
```

santihernandezc (approved):

PR #11583, opened 2025-05-30
Don't try to create a `Query` instance just to compute stats for remote read requests

charleskorn:
#### What this PR does

This PR modifies the behaviour of the query stats middleware to no longer create a `promql.Query` instance when computing query stats for remote read requests.

This is completely unnecessary, and also [doesn't work as expected with MQE](https://github.com/grafana/mimir/pull/11417#discussion_r2092556045).

I haven't added a changelog entry given this is just a refactoring.

#11417

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


56quarters (approved):
Nicely done

PR #11591, opened 2025-06-02
Renovate: Don't update grafana/mimir-otlptranslator

aknuds1:
#### What this PR does

Configure Renovate to not update our prometheus/otlptranslator fork: github.com/grafana/mimir-otlptranslator. The motivation is that Renovate may try to update `replace` directives of forked dependencies to use the latest upstream revision instead.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


charleskorn:
> Configure Renovate to not update our prometheus/otlptranslator fork: github.com/grafana/mimir-otlptranslator. The motivation is that Renovate may try to update `replace` directives of forked dependencies to use the latest upstream revision instead.

Do you have an example of this? I couldn't find any PRs that did this.

aknuds1:
> Do you have an example of this? I couldn't find any PRs that did this.

@charleskorn I don't have any examples for grafana/mimir-otlptranslator (yet), but it's a general problem I've seen many times with Renovate. See #11585 for a recent example. It seems as if Renovate thinks the latest upstream revision is the latest also for the fork.

charleskorn:
> > Do you have an example of this? I couldn't find any PRs that did this.
> 
> @charleskorn I don't have any examples for grafana/mimir-otlptranslator (yet), but it's a general problem I've seen many times with Renovate. See #11585 for a recent example. It seems as if Renovate thinks the latest upstream revision is the latest also for the fork.

In #11585, wasn't it updating the `replace` statement with the latest commit from `main` from my fork, which happened to be the latest commit on `main` from upstream as well? I've since merged my changes onto the `main` branch of my fork, so this shouldn't keep happening.

aknuds1:
> In https://github.com/grafana/mimir/pull/11585, wasn't it updating the replace statement with the latest commit from main from my fork, which happened to be the latest commit on main from upstream as well? I've since merged my changes onto the main branch of my fork, so this shouldn't keep happening.

@charleskorn That sounds like a plausible explanation. However, based on my experience of observing Renovate opening PRs to change our `replace` directives to a different revision than we want, I think we should disable updating of pinned dependencies for safety, i.e. avoiding accidental downgrades. Since we've manually pinned to a certain revision, the safe procedure in my experience is to manually update the pin as well.

charleskorn:
> @charleskorn That sounds like a plausible explanation. However, based on my experience of observing Renovate opening PRs to change our `replace` directives to a different revision than we want, I think we should disable updating of pinned dependencies for safety, i.e. avoiding accidental downgrades. Since we've manually pinned to a certain revision, the safe procedure in my experience is to manually update the pin as well.

It seems like we have two cases:

* `replace` directives used to set a particular version (eg. what we were doing with gRPC until recently)
* `replace` directives used to swap the upstream for a fork

To me, it sounds like we should only be asking Renovate to ignore dependencies used in the first case.

aknuds1:
> It seems like we have two cases:
> * replace directives used to set a particular version (eg. what we were doing with gRPC until recently)
> * replace directives used to swap the upstream for a fork
> To me, it sounds like we should only be asking Renovate to ignore dependencies used in the first case.

@charleskorn I get your point, but in my personal opinion I still find it safer to let a human update `replace` directives also for forks, so we don't end up accidentally losing our changes.

charleskorn (approved):

PR #11600, opened 2025-06-02
parquet-converter: Introduce metrics and ring test

jesusvazquez:
#### What this PR does

This PR does the following:
* Introduce a unit test to verify that the converters starting, joining the ring and converting the blocks.
* Adds metrics to the converter to track total conversions, failures, durations and processed users.
* Small refactors in parquet-converter
  * Extract the `ConvertBlock` function to an interface `blockConverter` so that we can mock it for faster testing.
  * Extract block conversion and write conversion mark to its own function `processBlock` so that we can properly instrument it.


pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,210 @@
...
+				converterCfg.ShardingRing.WaitStabilityMaxDuration = 0
+				converterCfg.ShardingRing.WaitActiveInstanceTimeout = 30 * time.Second
+				converterCfg.ConversionInterval = 1 * time.Second
+
+				c, reg := prepareWithMockConverter(t, converterCfg, objstore.WithNoopInstr(bucketClient))
+				t.Cleanup(func() {
+					if c.State() == services.Failed {
+						// Don't try to stop a service that already failed
+						t.Logf("Converter %s failed with error: %v", instanceID, c.FailureCase())
+						return
```
npazosmendez: Should we fail the test in this case? Also below

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,210 @@
...
+				instanceID := fmt.Sprintf("converter-%d", i)
+
+				converterCfg := prepareConfig(t)
+				converterCfg.ShardingRing.Common.KVStore.Mock = ringStore
+				converterCfg.ShardingRing.Common.InstanceID = instanceID
+				converterCfg.ShardingRing.Common.InstanceAddr = fmt.Sprintf("127.0.0.%d", i)
+				converterCfg.ShardingRing.WaitStabilityMinDuration = 0
+				converterCfg.ShardingRing.WaitStabilityMaxDuration = 0
+				converterCfg.ShardingRing.WaitActiveInstanceTimeout = 30 * time.Second
+				converterCfg.ConversionInterval = 1 * time.Second
```
npazosmendez: I would leave the `ConversionInterval` that the `prepareConfig` returns (100ms), which should speed up the test.
```suggestion
```

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,210 @@
...
+			time.Sleep(2 * time.Second) // Wait for converters to do the initial pass through the blocks.
+
+			// At this point we expect that all converters have joined the ring, discovered all tenants, and converted
+			// blocks for all users.
+			metrics := registries.BuildMetricFamiliesPerTenant()
+			// All converters should have discovered the same number of tenants.
+			assert.Equal(t, float64(numUsers*numConverters), metrics.GetSumOfGauges("cortex_parquet_converter_tenants_discovered"), "number of discovered tenants didnt match")
+			// The number of blocks synced should be equal to the number of blocks in the TSDB proving that despite the
+			// number of converters the blocks are sharded across them.
+			assert.Equal(t, float64(numBlocks), metrics.GetSumOfCounters("cortex_parquet_converter_blocks_converted_total"), "number of blocks converted didnt match")
```
npazosmendez: We can use `test.Poll` here for more test robustness and speed, avoiding the `time.Sleep(2 * time.Second)`

pkg/parquetconverter/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+
+package parquetconverter
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+type parquetConverterMetrics struct {
+	blocksConverted       prometheus.Counter
```
npazosmendez, line 11: We can get this count from `conversionDuration`, I would drop this one

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,210 @@
...
+			t.Cleanup(cancel)
+
+			// Create isolated bucket and blocks for this test case
+			bucketClientOnDisk, storageDir := mimir_testutil.PrepareFilesystemBucket(t)
+			now := time.Now()
+			mockTSDB(t, path.Join(storageDir, "user-1"), 24, numBlocks/2, now.Add(-24*time.Hour).Unix()*1000, now.Unix()*1000)
+			mockTSDB(t, path.Join(storageDir, "user-2"), 24, numBlocks/2, now.Add(-24*time.Hour).Unix()*1000, now.Unix()*1000)
+
+			// Uploading the blocks to in memory storage - otherwise we run out of file descriptors during the test
+			// The default limit for fsd is 1024 on linux.
```
npazosmendez, line 89: Did this really happen for this test? :open_mouth:  Sounds like a bug somewhere, we should investigate

npazosmendez (commented):
This will be very nice, thanks

I'm thinking we should move the entire `for _, m := range metas` loop (except the context error check) inside the `processBlock`. Mostly to get visibility on other errors that happen before the conversion itself (downloading, checking marks, etc), register those as failed conversions.

pkg/parquetconverter/parquet_converter_ring_test.go
jesusvazquez: Nop, this is taken from another unit test for store gateways. The limit is probably at CI level when the whole suite runs, hence the idea of running in memory instead 👍

pkg/parquetconverter/metrics.go
jesusvazquez: I tried this but because its a native histogram the only way to access the count is using histogram_count like this:
```
histogram_count(sum(rate(cortex_parquet_converter_conversion_duration_seconds[$__rate_interval]))
```

And the utilities to do metrics.GetSumOfCounters() doesn't support this so instead I'm going to keep the counter as it is right now 👍

pkg/parquetconverter/parquet_converter_ring_test.go
jesusvazquez: TIL and looks pretty nice indeed
```golang
test.Poll(t, 10*time.Second, true, func() interface{} {
  numConvertedBlocks := metrics.GetSumOfCounters("cortex_parquet_converter_blocks_converted_total")
  return numConvertedBlocks == numBlocks
})

```

pkg/parquetconverter/parquet_converter_ring_test.go
jesusvazquez: Sorry this was me debugging a past issue. Something that happens sometimes is that the parent converter context is cancelled earlier than the test cleanup runs so running `StopAndAwaitTerminated` on an already stopped converter yields an error.

I've left it like this:
```golang
	err := services.StopAndAwaitTerminated(context.Background(), c)
	if err != nil && errors.Is(err, context.Canceled) {
		return
	}
	assert.NoError(t, err)

```

I don't think we should fail the tests on the context cancelation race because afterall the cleanup function is a best effort and the service can be stopped either by us synchronously here or by the parent context that initialized the converter before.

pkg/parquetconverter/parquet_converter.go
```diff
@@ -259,40 +288,58 @@ func (c *ParquetConverter) running(ctx context.Context) error {
...
+func (c *ParquetConverter) processBlock(ctx context.Context, userID string, meta *block.Meta, uBucket objstore.InstrumentedBucket, logger log.Logger) {
+	start := time.Now()
+	var err error
+	defer func() {
+		duration := time.Since(start)
+		c.metrics.conversionDuration.Observe(duration.Seconds())
+		c.metrics.blocksConverted.Inc()
+		if err != nil {
+			c.metrics.blocksConvertedFailed.Inc()
+		}
```
npazosmendez: It might be a bit confusing that we are counting failed conversions (i.e. not converted) in `blocksConverted`. We could do 
```suggestion
		if err != nil {
			c.metrics.blocksConvertedFailed.Inc()
		} else {
		        c.metrics.blocksConverted.Inc()
		}
```

I believe this would be equivalent to how compactors track metrics too: https://github.com/grafana/mimir/blob/15c7117f2c5aeee690f628c01d8ae8605410f3c0/pkg/compactor/bucket_compactor.go#L958-L973

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,215 @@
...
+				instanceID := fmt.Sprintf("converter-%d", i)
+
+				converterCfg := prepareConfig(t)
+				converterCfg.ShardingRing.Common.KVStore.Mock = ringStore
+				converterCfg.ShardingRing.Common.InstanceID = instanceID
+				converterCfg.ShardingRing.Common.InstanceAddr = fmt.Sprintf("127.0.0.%d", i)
+				converterCfg.ShardingRing.WaitStabilityMinDuration = 0
+				converterCfg.ShardingRing.WaitStabilityMaxDuration = 0
+				converterCfg.ShardingRing.WaitActiveInstanceTimeout = 30 * time.Second
+				converterCfg.ConversionInterval = 1 * time.Second
```
npazosmendez: nit: bringing back my other comment, I'd remove this to speed up the test
```suggestion
```

npazosmendez (approved):
Some final suggestions, otherwise LGTM

francoposa:
Just noting that the integration test here is also failing other places and people are working on it, not need to look at it

pkg/parquetconverter/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		Help: "Time taken to convert a block to parquet format",
+		// Buckets
+		// 1 0.008s  (8ms)
+		// 2 0.032s  (32ms)
+		// 3 0.128s  (128ms)
+		// 4 0.512s  (512ms)
+		// 5 2.048s  (2048ms)
+		// 6 8.192s  (8192ms)
+		// 7 32.768s (32768ms)
+		// 8 +Inf
```
francoposa: Do we think this is a good spread to capture the block conversion times? The conversions for larger blocks took a lot longer than 32s on my machine with 16 cores working on it

pkg/parquetconverter/parquet_converter.go
```diff
@@ -259,40 +288,58 @@ func (c *ParquetConverter) running(ctx context.Context) error {
...
+			}
+		}
+	}
+}
 
-					if mark.Version == CurrentVersion {
-						continue
-					}
+// processBlock handles the conversion of a single block with proper metrics tracking.
+func (c *ParquetConverter) processBlock(ctx context.Context, userID string, meta *block.Meta, uBucket objstore.InstrumentedBucket, logger log.Logger) {
```
francoposa, line 299: general cleanup: the logger is on the `ParquetConverter` struct, we do not need to pass it as an arg to its methods

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+			for i := 1; i <= numConverters; i++ {
+				instanceID := fmt.Sprintf("converter-%d", i)
+
+				converterCfg := prepareConfig(t)
+				converterCfg.ShardingRing.Common.KVStore.Mock = ringStore
+				converterCfg.ShardingRing.Common.InstanceID = instanceID
+				converterCfg.ShardingRing.Common.InstanceAddr = fmt.Sprintf("127.0.0.%d", i)
+				converterCfg.ShardingRing.WaitStabilityMinDuration = 0
+				converterCfg.ShardingRing.WaitStabilityMaxDuration = 0
```
francoposa: I suspect this is only passing consistently due to the minimal latency of these mocked out instances all running and communicating in the same process.
The corresponding values in the store-gateway test are
```
gatewayCfg.ShardingRing.WaitStabilityMinDuration = 4 * time.Second
gatewayCfg.ShardingRing.WaitStabilityMaxDuration = 30 * time.
```

The point is for the initial wait to give them time to join and split up the work instead of one running off converting for the entire working set because it thinks it is the only instance.
The docstring helps a bit here too:
```
// TestStoreGateway_InitialSyncWithWaitRingTokensStability tests the store-gateway cold start case.
// When several store-gateways start up at once, we expect each store-gateway to only load
// their own blocks, regardless which store-gateway joined the ring first or last (even if starting
// at the same time, they will join the ring at a slightly different time).
```

pkg/parquetconverter/parquet_converter_ring_test.go
```diff
@@ -0,0 +1,215 @@
...
+				require.NoError(t, g.AwaitRunning(ctx))
+			}
+
+			// At this point we expect that all converters have joined the ring, discovered all tenants, and converted
+			// blocks for all users.
+
+			// All converters should have discovered the same number of tenants.
+			test.Poll(t, 10*time.Second, true, func() interface{} {
+				metrics := registries.BuildMetricFamiliesPerTenant()
+				users := metrics.GetSumOfGauges("cortex_parquet_converter_tenants_discovered")
```
francoposa, line 151: Could be saved for a later PR: this number makes more sense in conjunction with the corresponding metric like `cortex_compactor_tenants_skipped` which is incremented when a discovered tenant is skipped because it is not owned by the compactor due to the sharding.

jesusvazquez: Thought about this but the sharding for the converter is per block not per tenant. Would you want a counter with blocks discarded?

pkg/parquetconverter/parquet_converter.go
jesusvazquez: Thanks saw that but also we were instantiating this logger with the user ID right before the calls https://github.com/grafana/mimir/blob/6b922d78fe406ab4b7e91c5af70dd4c270ec55f0/pkg/parquetconverter/parquet_converter.go#L253 so I was thinking maybe we want to keep using that logger?

pkg/parquetconverter/parquet_converter.go
jesusvazquez: Okey will do it this way. Originally I was trying to have a "global" counter and an error counter, pretty much how it works for requests. But since we're talking about conversions might make sense to only have successful and failed ones separately.

pkg/parquetconverter/metrics.go
jesusvazquez: Ha! you are right, was thinking about adjusting later. Going to try with:
```
❯ go run main.go  0.500 2 10
// Buckets
// 1 0.500s  (500ms)
// 2 1.000s  (1000ms)
// 3 2.000s  (2000ms)
// 4 4.000s  (4000ms)
// 5 8.000s  (8000ms)
// 6 16.000s  (16000ms)
// 7 32.000s  (32000ms)
// 8 64.000s  (64000ms)
// 9 128.000s  (128000ms)
// 10 256.000s  (256000ms)
// 11 +Inf
```

pkg/parquetconverter/parquet_converter_ring_test.go
jesusvazquez: Yes lets do this, I think originally i was working with these timeouts but i was running into a flakky test of not having time to convert all blocks, this is why i changed it to not wait. But now that I'm using test.Poll that problem isnt there.

I also checked how fast they were starting:
```
parquet-converter ring is ACTIVE in the ring after 0.112763
parquet-converter ring is ACTIVE in the ring after 0.237433
parquet-converter ring is ACTIVE in the ring after 0.337752
parquet-converter ring topology is stable after 2.238654
parquet-converter ring topology is stable after 2.339001
parquet-converter ring topology is stable after 3.113959
```

So I'm going to leave it at 2 seconds for min duration and 30 seconds for max duration.

jesusvazquez:
Going to go ahead and merge because I want to try this. Happy to make further changes later on.

PR #11601, opened 2025-06-02
fix(promql): histogram_count returns wrong value

### LINKED ISSUE #11148: Bug: Native histogram operations like histogram_count emitting negative values
### What is the bug?

Hi Team,

We are seeing cases where histogram_count are giving negative values intermittently.

We also recently upgraded to 2.16.0 to see if that alleviates the issue as it includes fix related to [aggr_over_time](https://github.com/prometheus/prometheus/pull/15711) and [counter reset detection](https://github.com/prometheus/prometheus/pull/15987) but issue still persists.

### How to reproduce it?

Start Mimir 2.16.0
One of rule is based on query `avg_over_time(histogram_count(sum(rate(query:without_instance{bushido_mode="shogun",bushido_tier=~"production.*"}[5m])))[5m:])` which trigger alerts with a negative value

However, when we visualize the same query in a dashboard it shows expected values (non-negative), so it's only seen during rule evaluation.

histogram_count should never emit negative values

Kubernetes

_No response_

krajorama:
DO NOT PICK, WIP

#### What this PR does

Reproduce inconsistency between native histogram `count` of  `sum(rate(histogram_metric[10m]))` and query of the `count` by `histogram_count(sum(rate(histogram_metric[10m])))`

Fixes #11148 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/integration/querier_test.go
+++ b/integration/querier_test.go
@@ -17,6 +17,7 @@ import (
 	e2edb "github.com/grafana/e2e/db"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
+//	"github.com/prometheus/prometheus/model/histogram"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


krajorama:
This test fails locally and in CI on 2.16, however if I use the Mimir query engine then it doesn't fail.

The native histogram result has count==0.006805555555555557 whereas the `histogram_count` gives 0.

krajorama:
Fixing upstream: https://github.com/prometheus/prometheus/pull/16682

PR #11602, opened 2025-06-02
operations/mimir: Upstream Jsonnet changes from r334

aknuds1:
#### What this PR does

Upstream operations/mimir Jsonnet changes from r334:

1. Don't depend on Memcached on startup 
1. Set max buffered bytes in the Kafka client used by the ruler

The first is upstreamed through upgrading to latest dskit, in which the corresponding flag `-dns-ignore-startup-failures`'s default [changes to true](https://github.com/grafana/dskit/pull/705). Therefore, we don't need to set the flag via Jsonnet any longer.

I've tested applying the PR's changes, and can see there are no differences versus our r334 cloud deployment (given that Jsonnet corresponding to change 1 is kept until the right Mimir release).
 
Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 * [ENHANCEMENT] Querier: Include chunks streamed from store-gateway in Mimir Query Engine memory estimate of query memory usage. #11453 #11465
 * [ENHANCEMENT] Querier: Include chunks streamed from ingester in Mimir Query Engine memory estimate of query memory usage. #11457
 * [ENHANCEMENT] Query-frontend: Add retry mechanism for remote reads, series, and cardinality prometheus endpoints #11533
+* [ENHANCEMENT] Ruler: Ignore rulers in non-operation states when getting and syncing rules #11569
 * [BUGFIX] OTLP: Fix response body and Content-Type header to align with spec. #10852
 * [BUGFIX] Compactor: fix issue where block becomes permanently stuck when the Compactor's block cleanup job partially deletes a block. #10888
 * [BUGFIX] Storage: fix intermittent failures in S3 upload retries. #10952
@@ -74,6 +75,7 @@
 * [BUGFIX] Ruler: Failures during initial sync must be fatal for the service's startup. #11545
 * [BUGFIX] Querier and query-frontend: Fix issue where aggregation functions like `topk` and `quantile` could return incorrect results if the scalar parameter is not a constant and Prometheus' query engine is in use. #11548
 * [BUGFIX] Querier and query-frontend: Fix issue where range vector selectors could incorrectly ignore samples at the beginning of the range. #11548
+* [BUGFIX] Querier: Fix rare panic if a query is canceled while a request to ingesters or store-gateways has just begun. #11613
 
 ### Mixin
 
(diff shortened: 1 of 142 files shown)


operations/mimir/querier.libsonnet
```diff
@@ -28,6 +28,9 @@
       // We request high memory but the Go heap is typically very low (< 100MB) and this causes
       // the GC to trigger continuously. Setting a ballast of 256MB reduces GC.
       'mem-ballast-size-bytes': 1 << 28,  // 256M
+
+      // Don't depend on Memcached on startup.
+      'blocks-storage.bucket-store.metadata-cache.memcached.dns-ignore-startup-failures': true,
```
charleskorn: Should we make this the default value for the CLI flag, rather than setting it in Jsonnet and Helm?

(same comment applies below)

aknuds1: I was honestly not sure about this @charleskorn. Should we go for it?

aknuds1: Changing the defaults is probably fine, when I think about it. It does make things simpler.

aknuds1: I realized it's not that simple after all @charleskorn - `-dns-ignore-startup-failures` is a dskit flag. dskit being an open source project, I don't think we can just go and change its flag defaults. WDYT?

CHANGELOG.md
```diff
@@ -95,6 +95,8 @@
 * [CHANGE] Increase the allowed number of rule groups for small, medium_small, and extra_small user tiers by 20%. #11152
 * [CHANGE] Update rollout-operator version to 0.26.0. #11232
 * [CHANGE] Memcached: Set a timeout of `500ms` for the `ruler-storage` cache instead of the default `200ms`. #11231
+* [CHANGE] Memcached: Ignore initial DNS resolution failure, i.e. don't depend on Memcached on startup. #11602
```
```suggestion
* [CHANGE] Memcached: Ignore initial DNS resolution failure, meaning don't depend on Memcached on startup. #11602
```

CHANGELOG.md
```diff
@@ -95,6 +95,8 @@
...
+* [CHANGE] Ruler: If ingest storage is enabled, set the max buffered bytes in the Kafka client used by the ruler based on the expected max rule evaluation response size, clamping it between 1 GB (default) and 4 GB. #11602
```
```suggestion
* [CHANGE] Ruler: If ingest storage is enabled, set the maximum buffered bytes in the Kafka client used by the ruler based on the expected maximum rule evaluation response size, clamping it between 1 GB (default) and 4 GB. #11602
```

CHANGELOG.md
aknuds1: Can I ask why we shouldn't be using i.e.?

tacole02: Sure, good question! It's just better to always be as specific as possible and not use Latin abbreviations, as many people confuse their meanings. Here's a reference to Google's documentation style guide that supports this idea: https://developers.google.com/style/word-list

operations/mimir/querier.libsonnet
56quarters: > I realized it's not that simple after all @charleskorn - -dns-ignore-startup-failures

It's experimental and added by @tinitiuset specifically for Mimir so that's fine if we want to do it.

If you're going to add the flag to Mimir, can you move this to `config.libsonnet` with all the other cache settings?

aknuds1: @56quarters Thanks, in that case it shouldn't be a problem to upstream the new default to dkskit. I'll do it.

aknuds1: @56quarters @charleskorn [dskit PR](https://github.com/grafana/dskit/pull/705).

aknuds1: Thanks for the review @charleskorn! The dskit PR is merged.

charleskorn (approved):

PR #11613, opened 2025-06-03
Fix issue where queriers can panic if the query is cancelled while a `Select` call is running

charleskorn:
#### What this PR does

This PR fixes an issue where queriers can panic if the query is cancelled or fails while a `Select` call is running. 

This happens because queriers wrap the ingester and store-gateway `Querier` instances with a `LazyQuerier`. This can result in `Close` being called while the `Select` call is still running, creating a race to read from / write to `streamReaders` and causing `Close` to call `FreeBuffer()` on a nil reader instance if `Select` has not finished appending the new readers to `streamReaders`.

An alternative solution would be to modify `LazyQuerier` to wait for the `Select` to finish before calling `Close` on the wrapped `Querier`, but this may cause delays in closing the query if the `Select` does not complete promptly.

#11136

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/querier/blocks_store_queryable.go
+++ b/pkg/querier/blocks_store_queryable.go
@@ -56,6 +56,8 @@ import (
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 )
 
+var errAlreadyClosed = errors.New("querier already closed")
+
 // BlocksStoreSet is the interface used to get the clients to query series on a set of blocks.
 type BlocksStoreSet interface {
 	services.Service
@@ -308,6 +310,7 @@ func (q *BlocksStoreQueryable) Querier(mint, maxt int64) (storage.Querier, error
 		consistency:              q.consistency,
 		logger:                   q.logger,
 		queryStoreAfter:          q.queryStoreAfter,
+		streamReadersMtx:         &sync.Mutex{},
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/querier/distributor_queryable.go
```diff
@@ -195,6 +199,18 @@ func (q *distributorQuerier) streamingSelect(ctx context.Context, minT, maxT int
 	}
 
 	// Store the stream readers so we can free their buffers when we're done using this Querier.
```
aknuds1: Can this be moved to above line 214? It looks left behind to me.

pkg/querier/blocks_store_queryable.go
```diff
@@ -326,7 +329,9 @@ type blocksStoreQuerier struct {
 	// "now - queryStoreAfter" so that most recent blocks are not queried.
 	queryStoreAfter time.Duration
 
-	streamReaders []*storeGatewayStreamReader
+	streamReadersMtx *sync.Mutex
```
aknuds1: Just curious, why use a Mutex pointer? Wouldn't a value type Mutex be easier to deal with and more idiomatic?

aknuds1 (approved):
LGTM, but please see comments (especially the comment which I think should be moved).

pkg/querier/distributor_queryable.go
charleskorn: Good catch, fixed.

pkg/querier/blocks_store_queryable.go
charleskorn: Yep, you're right, I've fixed this in 77d9518.

PR #11619, opened 2025-06-03
Fix Jsonnet installation help for recent Go versions

tcard:
#### What this PR does

Minor fix in Jsonnet installation instructions from several files.


56quarters (approved):
Nice!

PR #11627, opened 2025-06-04
Reinstate mimir-otlptranslator fork dependency

aknuds1:
#### What this PR does

The mimir-otlptranslator fork dependency (via `replace` directive) was temporarily dropped in #11607. It should be reinstated, since it was agreed on management level that Mimir will maintain its own fork to be able to ensure that no breaking changes enter the dependency chain. The fork is currently identical with upstream, so there are no code changes.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo (approved):
🔥

PR #11643, opened 2025-06-05
block-builder: remove standalone mode

narqo:
#### What this PR does

Now, as we have block-builder-scheduler, we can remove the standalone mode, and continue iterating on top of its architecture.

This PR _mechanically_ removes everything related the standalone mode. My idea is that after this one gets merged, we can go ahead and start looking at if there are any test scenarios missing, and gradually add those.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -51,10 +51,6 @@ type BlockBuilder struct {
 	// the current job iteration number. For tests.
 	jobIteration atomic.Int64
 
-	assignedPartitionIDs []int32
-	// fallbackOffsetMillis is the milliseconds timestamp after which a partition that doesn't have a commit will be consumed from.
-	fallbackOffsetMillis int64
-
 	blockBuilderMetrics blockBuilderMetrics
 	tsdbBuilderMetrics  tsdbBuilderMetrics
 }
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -221,13 +189,12 @@ func (b *BlockBuilder) runningPullMode(svcCtx context.Context) error {
 		}
 
 		// Once we've gotten a job, we attempt to complete it even if the context is cancelled.
-
-		if _, err := b.consumeJob(context.Background(), key, spec); err != nil {
+		if _, err := b.consumeJob(context.WithoutCancel(ctx), key, spec); err != nil {
```
seizethedave, line 192: Oh, nice.

pkg/blockbuilder/metrics.go
```diff
@@ -37,8 +28,9 @@ func newBlockBuilderMetrics(reg prometheus.Registerer) blockBuilderMetrics {
 	}, []string{"success"})
 
 	m.processPartitionDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
-		Name:                        "cortex_blockbuilder_process_partition_duration_seconds",
-		Help:                        "Time spent processing one partition.",
+		Name: "cortex_blockbuilder_process_partition_duration_seconds",
```
seizethedave, line 31: I think this one is now redundant with the similar `consume_job` metric. It's in some dashboards, though.

seizethedave (approved):
Yes! Living the dream.

pkg/blockbuilder/metrics.go
narqo: Good point, I will clean up the dashboards after

PR #11646, opened 2025-06-05
[WIP] parquet index reader

francoposa:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storage/parquet/index/reader_lazy_loader.go
```diff
@@ -0,0 +1,382 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/thanos-io/thanos/blob/main/pkg/block/indexheader/lazy_binary_reader.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Thanos Authors.
+
+package index
+
```
francoposa, line 7: for now this is just a copy of `lazy_binary_reader.go` with the methods updated to implement the `Reader` interface from this package, meaning I just dropped the `LabelValuesOffsets`, `PostingsOffsets`, and changed `IndexVersion` to return a string version because it will give us more flexibility to experiment with multiple versions than an integer version number.

PR #11647, opened 2025-06-05
ruler: Fix QueryOffset being ignored when recording xor alerting rule evaluation is disabled

alexweav:
#### What this PR does

When evaluation of one rule type is disabled but the other is enabled, the sync logic makes a filtered copy of each rule group. The copy logic ignores newly added fields `QueryOffset` and `AlignEvaluationTimeOnInterval`. We then sync these groups with the fields blapped.

This PR changes the copy + tests to consider all fields as of today, and also be tolerant of newly added fields so it doesn't break again in the future.

n/a

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 * [BUGFIX] Querier and query-frontend: Fix issue where aggregation functions like `topk` and `quantile` could return incorrect results if the scalar parameter is not a constant and Prometheus' query engine is in use. #11548
 * [BUGFIX] Querier and query-frontend: Fix issue where range vector selectors could incorrectly ignore samples at the beginning of the range. #11548
 * [BUGFIX] Querier: Fix rare panic if a query is canceled while a request to ingesters or store-gateways has just begun. #11613
+* [BUGFIX] Ruler: Fix QueryOffset and AlignEvaluationTimeOnInterval being ignored when recording xor alerting rule evaluation is disabled. #11647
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -79,6 +79,7 @@
 * [BUGFIX] Querier and query-frontend: Fix issue where aggregation functions like `topk` and `quantile` could return incorrect results if the scalar parameter is not a constant and Prometheus' query engine is in use. #11548
 * [BUGFIX] Querier and query-frontend: Fix issue where range vector selectors could incorrectly ignore samples at the beginning of the range. #11548
 * [BUGFIX] Querier: Fix rare panic if a query is canceled while a request to ingesters or store-gateways has just begun. #11613
+* [BUGFIX] Ruler: Fix QueryOffset and AlignEvaluationTimeOnInterval being ignored when recording xor alerting rule evaluation is disabled. #11647
```
narqo: > recording **xor** alerting rule evaluation

What does the "xor" refer to here?

dimitarvdimitrov (approved):

CHANGELOG.md
alexweav: "Exclusive or," the bug only happens when you have recording disabled and alerting enabled, or vice versa, it does not happen when both are disabled. I'll improve the wording a bit before merging it

PR #11654, opened 2025-06-06
MQE: label each source of memory consumption to make it easier to identify where something is returned to the pool multiple times

charleskorn:
#### What this PR does

This PR extends `limiting.MemoryConsumptionTracker` to track memory consumption by source.

This allows us to include more information in `Estimated memory consumption of this query is negative` panics, and also makes it easier to debug failures when MQE is running in pedantic mode during tests. 

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/util/limiter/memory_consumption.go
+++ b/pkg/util/limiter/memory_consumption.go
@@ -58,23 +58,47 @@ const (
 	memoryConsumptionSourceCount = TopKBottomKRangeQuerySeriesSlices + 1
 )
 
-var memoryConsumptionSourceNames = map[MemoryConsumptionSource]string{
-	IngesterChunks:                      "ingester chunks",
-	StoreGatewayChunks:                  "store-gateway chunks",
-	FPointSlices:                        "[]promql.FPoint",
-	HPointSlices:                        "[]promql.HPoint",
-	Vectors:                             "promql.Vector",
-	Float64Slices:                       "[]float64",
-	IntSlices:                           "[]int",
-	IntSliceSlice:                       "[][]int",
-	Int64Slices:                         "[]int64",
-	BoolSlices:                          "[]bool",
-	HistogramPointerSlices:              "[]*histogram.FloatHistogram",
-	SeriesMetadataSlices:                "[]SeriesMetadata",
-	BucketSlices:                        "[]promql.Buckets",
-	QuantileGroupSlices:                 "[]aggregations.qGroup",
-	TopKBottomKInstantQuerySeriesSlices: "[]topkbottom.instantQuerySeries",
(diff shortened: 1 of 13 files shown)


pkg/util/limiter/memory_consumption.go
```diff
@@ -42,6 +85,8 @@ type MemoryConsumptionTracker struct {
 	currentEstimatedMemoryConsumptionBytes uint64
```
tcard, line 109: **suggestion:** Given this should always be `sum(values(currentEstimatedMemoryConsumptionBySource))` maybe it's worth replacing it with a function? Less room for inconsistency bugs.

tcard (approved):

pkg/util/limiter/memory_consumption.go
```diff
@@ -34,6 +35,48 @@ func AddMemoryTrackerToContext(ctx context.Context, tracker *MemoryConsumptionTr
...
+	SeriesMetadataSlices
+	BucketSlices
+	QuantileGroupSlices
+	TopKBottomKInstantQuerySeriesSlices
+	TopKBottomKRangeQuerySeriesSlices
+
+	memoryConsumptionSourceCount = TopKBottomKRangeQuerySeriesSlices + 1
+)
+
+var memoryConsumptionSourceNames = map[MemoryConsumptionSource]string{
```
aknuds1: I think it would be preferable to implement a `String` method on `MemoryConsumptionSource`, where you use a `switch` statement, that can also handle (panic) if an unrecognized instance is encountered.

56quarters: Done in ff4c2898745b0d18a800588a1870db08c2306ec6

pkg/util/limiter/memory_consumption.go
56quarters: I'm going to put this off for now since it would required larger changes and @charleskorn is out for the weekend and I'd like to get this merged.

pkg/util/limiter/memory_consumption.go
56quarters: I decided not to panic and instead add a test to make sure we don't introduce a variant without a `String()` representation. I don't want to panic in the code that is used when we're already panicking.

pkg/util/limiter/memory_consumption.go
charleskorn: I agree that would reduce the possibility of inconsistency bugs, but given we need to check the total memory consumption every time `IncreaseMemoryConsumption` is called, and that method can be called thousands of times for a single query, performance takes a higher priority.

PR #11658, opened 2025-06-06
add livenessProbe and readinessProbe to memcached in helm chart

KyriosGN0:
#### What this PR does

add supports for a readiness Probe and livenss Probe, both have commented out values in the values.yaml for reference 

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/compare-helm-with-jsonnet/helm/06-pods/kustomization.yaml
+++ b/operations/compare-helm-with-jsonnet/helm/06-pods/kustomization.yaml
@@ -17,7 +17,7 @@ patches:
 
   # TODO(logiraptor): Helm adds a grpc port to the overrides exporter, Jsonnet does not.
   - target:
-      name: mimir-overrides-exporter 
+      name: mimir-overrides-exporter
       kind: Deployment
     patch: |-
       - op: remove
@@ -83,3 +83,20 @@ patches:
     patch: |-
       - op: remove
         path: /spec/template/metadata/namespace
+
+
+  # Helm sets readiness probe with tcp socket and few other fields, Jsonnet does only http get withouth successThreshold, timeoutSeconds, periodSeconds and failureThreshold
+  - target:
+      kind: 'StatefulSet'
+      name: 'mimir-.+-cache'
+    patch: |-
+      - op: remove
+        path: /spec/template/spec/containers/0/readinessProbe/tcpSocket
+      - op: remove
+        path: /spec/template/spec/containers/0/readinessProbe/successThreshold
+      - op: remove
+        path: /spec/template/spec/containers/0/readinessProbe/timeoutSeconds
+      - op: remove
+        path: /spec/template/spec/containers/0/readinessProbe/periodSeconds
+      - op: remove
+        path: /spec/template/spec/containers/0/readinessProbe/failureThreshold


operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -2472,6 +2472,22 @@ memcached:
...
 
+  readinessProbe: {}
+#    tcpSocket:
+#      port: 11211
+#    initialDelaySeconds: 5
+#    periodSeconds: 5
+#    timeoutSeconds: 3
+#    failureThreshold: 6
+#    successThreshold: 1
+
```
dimitarvdimitrov: is there any reason not to have a sensible default instead of having tiias opt-in?

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -2472,6 +2472,22 @@ memcached:
...
+  livenessProbe: {}
```
dimitarvdimitrov: we don't set liveness anywhere for the rest of the components. Are there specific problems you're trying to solve?

dimitarvdimitrov:
can you also add a static test in https://github.com/grafana/mimir/blob/ed3e678c490e8cdf34585d64106b06af5a64bf9a/operations/helm/charts/mimir-distributed/ci/offline so we can track if there are regressions?

56quarters:
Note that there was previously [another PR that proposed this](https://github.com/grafana/mimir/pull/9990). As mentioned in that PR, a liveness probe does not make sense for Memcached. Readiness I could see making sense but there would need to be a justification for it.

KyriosGN0:
> Note that there was previously [another PR that proposed this](https://github.com/grafana/mimir/pull/9990). As mentioned in that PR, a liveness probe does not make sense for Memcached. Readiness I could see making sense but there would need to be a justification for it.

i agree with the liveness probe and will remove it
however if a memcahed instance has reached it connection limit, trying to open more connection (because without readiness probe other compoents will try to reach it) will just cause errors in the services and possibly increase cpu in the memcached instance
readiness probe is exactly used for such problem (reach max connection -> remove from service -> close connection as work is done -> get added back to the service)

operations/helm/charts/mimir-distributed/values.yaml
KyriosGN0: no, i will update the values to have the defaults as they are commented

operations/helm/charts/mimir-distributed/values.yaml
KyriosGN0: will remove it

KyriosGN0:
@dimitarvdimitrov there is a problem with the jsonnet, currently i can't add the readiness probe to Jsonnet as the memcached lib doesn't support probes. 
should i add a kustomize file to ignore the probe?

dimitarvdimitrov:
> @dimitarvdimitrov there is a problem with the jsonnet, currently i can't add the readiness probe to Jsonnet as the memcached lib doesn't support probes. should i add a kustomize file to ignore the probe?

you're navigating the repo very well! I think you can a kustomization in this file https://github.com/grafana/mimir/blob/bdd443bd6d5b83d65a614cf4cf1d34048347f4b7/operations/compare-helm-with-jsonnet/helm/06-pods/kustomization.yaml. That should make CI pass. 


I also see check-changelog failing. That's a "bug" in the github action; it still has rough edges because it's new. This PR should fix it https://github.com/grafana/mimir/pull/11777. In the meantime we can add a label to make the action pass

KyriosGN0:
@dimitarvdimitrov fixed the jsonnet diff, for some reason it looks like jsonnet is adding a readiness probe with httpGet, so i worked around that a bit.

56quarters (changes requested):
> i agree with the liveness probe and will remove it
> however if a memcahed instance has reached it connection limit, trying to open more connection (because without readiness probe other compoents will try to reach it) will just cause errors in the services and possibly increase cpu in the memcached instance
> readiness probe is exactly used for such problem (reach max connection -> remove from service -> close connection as work is done -> get added back to the service)

I haven't been able to find a definitive answer one way or another but I think this may break the way Memcached works with Mimir.

From what I can tell from https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

> Also, the Pod needs to be ready in order to have a record unless publishNotReadyAddresses=True is set on the Service.

I think this means that if a Memcached pod becomes "not ready" its DNS record will be removed from the headless service that it is a part of. 

The way keys are hashed to Memcached servers in Mimir requires an ordered list of all cache servers. In order to not invalidate all keys when the list of servers changes, servers can only be added or removed (for example the STS scaling up and down) at the end of the server list.  I.e. if you have `memcached-1`, `memcached-2`, `memcached-3`, `memcached-4` and you remove `memcached-2` -- all keys that hashed to `memcached-3` and `memcached-4` will be sent to the wrong servers and result in cache misses.

We need to confirm how DNS records are affected with a pod becomes not ready.

KyriosGN0:
> > i agree with the liveness probe and will remove it
> > however if a memcahed instance has reached it connection limit, trying to open more connection (because without readiness probe other compoents will try to reach it) will just cause errors in the services and possibly increase cpu in the memcached instance
> > readiness probe is exactly used for such problem (reach max connection -> remove from service -> close connection as work is done -> get added back to the service)
> 
> I haven't been able to find a definitive answer one way or another but I think this may break the way Memcached works with Mimir.
> 
> From what I can tell from [kubernetes.io/docs/concepts/services-networking/dns-pod-service](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/)
> 
> > Also, the Pod needs to be ready in order to have a record unless publishNotReadyAddresses=True is set on the Service.
> 
> I think this means that if a Memcached pod becomes "not ready" its DNS record will be removed from the headless service that it is a part of.
> 
> The way keys are hashed to Memcached servers in Mimir requires an ordered list of all cache servers. In order to not invalidate all keys when the list of servers changes, servers can only be added or removed (such as the STS scaling up and down) at the end of the server list. For example if you have `memcached-1`, `memcached-2`, `memcached-3`, `memcached-4` and you remove `memcached-2` -- all keys that hashed to `memcached-3` and `memcached-4` will be sent to the wrong servers and result in cache misses.
> 
> We need to confirm how DNS records are affected with a pod becomes not ready.

i have found this https://github.com/kubernetes/kubernetes/issues/118952, it seems to confirm that if we use `publishNotReadyAddress` it will work as expected but the readiness check is useless, the only reason to do it is for folks who have compliance issues for pods without readiness check

56quarters:
> i have found this [kubernetes/kubernetes#118952](https://github.com/kubernetes/kubernetes/issues/118952), it seems to confirm that if we use `publishNotReadyAddress` it will work as expected but the readiness check is useless, the only reason to do it is for folks who have compliance issues for pods without readiness check

Right now, I don't want to add this as-is for the previously mentioned reasons: if it works how it's supposed to work, it breaks caching. If we work around that with adding `publishNotReadyAddress=true` to the service, the readiness probe doesn't solve any problems. Am I missing something?

KyriosGN0:
> If we work around that with adding `publishNotReadyAddress=true` to the service, the readiness probe doesn't solve any problems. Am I missing something?

No, i will close this PR then

PR #11664, opened 2025-06-06
Don't leave read-only mode if there is a forced compaction pending or in progress

pr00se:
#### What this PR does

Stop ingesters leaving read-only mode while a forced compaction is pending or in progress. Leaving read-only mode during the compaction will cause write errors, because writes are blocked.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/downscale.go
+++ b/pkg/ingester/downscale.go
@@ -52,6 +52,14 @@ func (i *Ingester) PrepareInstanceRingDownscaleHandler(w http.ResponseWriter, r 
 		i.circuitBreaker.push.deactivate()
 
 	case http.MethodDelete:
+		// Don't leave read-only mode if there's a forced compaction pending or in progress.
+		if len(i.forceCompactTrigger) > 0 || i.forcedCompactionInProgress.Load() {
+			msg := "cannot clear read-only mode while forced compaction is in progress"
+			level.Warn(i.logger).Log("msg", msg)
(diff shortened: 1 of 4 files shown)


periklis (approved):
LGTM

aldernero (approved):

pkg/ingester/downscale.go
```diff
@@ -52,6 +52,14 @@ func (i *Ingester) PrepareInstanceRingDownscaleHandler(w http.ResponseWriter, r
 		i.circuitBreaker.push.deactivate()
 
 	case http.MethodDelete:
+		// Don't leave read-only mode if there's a forced compaction pending or in progress.
+		if len(i.forceCompactTrigger) > 0 || i.forcedCompactionInProgress.Load() {
```
colega, line 56: I think there are a couple of race conditions here:
- One trivial one is that maybe we've consumed the `forceCompactTrigger` channel but didn't set the `forceCompactionInProgress` mark yet.
- Another tricky one, is the one that I learned the other day that `len(channel)` [is not _really_ atomic](https://stackoverflow.com/a/79021746/3990767).

I think that in order to handle this properly, we need to have another atomic value indicating that `forceCompactTriggered`, increment that before even trying to send to the channel, and decrement that after the force-compaction callback has been returned (or ingester shut down, whatever comes first, i.e., in the defer of that `run()` function).

Issue #11272, opened 2025-04-21
Idea: Add limit on minimum rule evaluation interval


alexweav:
### What is the problem you are trying to solve?

You can use any valid prometheus time for the rule evaluation interval, even as low as 1 second.
It's almost never useful to go that low, as it's way lower than a typical scrape interval. When a user does go this low, the system is likely not doing what they think it's doing. The group is very likely to miss evaluations during normal operation. It also issues a wasteful amount of queries when multiplied across many rules and groups.

### Which solution do you envision (roughly)?

Adding a per-tenant limit on minimum evaluation interval can protect the system by providing a more reasonable bound on the number of ruler queries. It also guides users who are unfamiliar with the Prometheus model toward best practices.

### How long do you think this would take to be developed?

Not sure

PR #11665, opened 2025-06-06
ruler: Add new limit on minimum rule evaluation interval

alexweav:
#### What this PR does

Introduces a new per-tenant limit, `ruler_min_rule_evaluation_interval`, which allows administrators to limit the speed at which rules may evaluate.

It defaults to zero, zero indicates no limit.

When the limit is raised and a rule already exists with an interval below the limit:
- The rule is still evaluated, but artificially slowed down to match the limit.
- The rule can't be edited until the limit is fixed.

This means administrators can use this as a knob to throttle a tenant's rule evaluations at runtime if needed. We still do a best-effort evaluation to avoid disrupting the tenant.

Rel https://github.com/grafana/mimir/issues/11272

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -680,13 +680,13 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if err := a.ruler.AssertMaxRulesPerRuleGroup(userID, namespace, len(rg.Rules)); err != nil {
-		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+		level.Warn(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
 	if err := a.ruler.AssertMinRuleEvaluationInterval(userID, time.Duration(rg.Interval)); err != nil {
-		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+		level.Warn(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
@@ -703,7 +703,7 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		}
 
 		if err := a.ruler.AssertMaxRuleGroups(userID, namespace, len(rgs)+1); err != nil {
-			level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
+			level.Warn(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
--- a/pkg/ruler/api.go
+++ b/pkg/ruler/api.go
@@ -685,6 +685,12 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if err := a.ruler.AssertMinRuleEvaluationInterval(userID, time.Duration(rg.Interval)); err != nil {
+		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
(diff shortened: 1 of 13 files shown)


pkg/ruler/api.go
```diff
@@ -685,6 +685,12 @@ func (a *API) CreateRuleGroup(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if err := a.ruler.AssertMinRuleEvaluationInterval(userID, time.Duration(rg.Interval)); err != nil {
+		level.Error(logger).Log("msg", "limit validation failure", "err", err.Error(), "user", userID)
```
colega: I don't think this should be logged as an error, this should be a warning at most: nothing failed, it's just a user who sent a wrong request.

alexweav: Makes sense! I only chose it because all the other "limit validation failure" log lines are also at error level.

I changed all 3 to warn :+1:

chencs (approved):
Thank you for doing this! One question -- we're changing the evaluation interval of rules, so `cortex_prometheus_rule_group_interval_seconds` is no longer necessarily accurate, correct?

alexweav:
@chencs interesting question, I suppose it depends on what you consider to be the true interval :smile: If you create a group with a 1s interval and an admin raises the limit to 10s, the interval is 1s if you ask the config and 10s if you ask the ruler, it's not consistent.

Right now this PR prefers the actively running state over the config when they disagree. That metric is calculated inside the running prometheus rules group instance, after the rule sync happens, and based on the rule that's actually loaded in memory. So, it'll reflect the changed value, not the value in the configuration. In the example above it would change to 10s because it's actually executing at 10s.

IMO this is consistent with the current ruler behavior. Putting this PR aside - right now, when you change an interval in the rule configuration, that metric doesn't update directly. It updates later on when the new rule gets synced and actually executes. This feels like precedent to prefer the rule in memory over the config.

chencs:
>  So, it'll reflect the changed value, not the value in the configuration. In the example above it would change to 10s because it's actually executing at 10s.

This answers my question and the "functional" interval is what I'd want to know from an operator perspective. Thank you!

alexweav:
For some extra confirmation of that, here's the metric updating when I change a tenant's min interval to 45s, with a 30s rule defined:
2025-06-09-153808_2057x341_scrot

alexweav:
@colega and @chencs I've discovered a bug in testing and reset the reviews. Users can leave `interval` completely blank or set it to zero. Doing this means to run the rule at `ruler.evaluation-interval` which is by default 1 minute. This PR begins blocking such rules in the API, because 0 is less than the `min_rule_evaluation_interval`!

It also can cause previously created rules with a blank interval to accidentally speed up, because it adjusts the 0 to `30s` or `45s` or whatever. Since 0 used to mean `1m`, the group now runs faster than intended.

I've pushed a fix. Instead, we now:
- Allow rules with 0/blank interval through the API as a special case, just like before.
- 0 means to use max(`ruler.evaluation-interval`, `min_rule_evaluation_interval`). This is non-breaking, and the limit is always strictly followed.
- In the rare event that a tenant `min_rule_evaluation_interval` is greater than `ruler.evaluation-interval`, which is nonsense, we log a warning about the misconfiguration, and keep respecting the max of the two.

chencs:
One more question -- I _believe_ that the interval being changed during rule sync means that the following scenario works, but want to double check:
1. with no minimum interval, tenant creates a rule with a small interval (e.g. 15s)
2. minimum interval is set higher than the rule interval (e.g. 1m) --> tenant cannot update the rule without changing the interval
3. minimum interval is then removed or set <= than the rule interval (e.g., 0 or 15s) --> tenant can now update the rule without changing the interval again

alexweav:
@chencs Yup, that's exactly what happens. Though, note the API effects are separate from the sync - so they read the limit independently. (The rule might be created on pod 1's API, but be scheduled on pod 2, so the sync is eventually consistent)

The API on all pods immediately respect the limit when it changes, whether up or down. If a rule is in breach of the limit at request time, it can't be updated without changing the interval. If the limit changes to allow it, it works again. Each request checks the limit, every time

chencs (approved):
Nice catch on that bug!

PR #11683, opened 2025-06-09
MQE: Add labels.Label of SeriesMetadata to MQE memory tracker

lamida:
#### What this PR does

Add labels.Label of SeriesMetadata to MQE memory tracker. 

The main change on this PR is the introduction of `IncreaseMemoryConsumptionForLabels` and `DecreaseMemoryConsumptionForLabels` method in memory consumption tracker.

`IncreaseMemoryConsumptionForLabels` will have to be called when passing actual `labels.Labels` under SeriesMetadata. 

`DecreaseMemoryConsumptionForLabels` can be called when putting back `SeriesMetadataSlicePool`.

As mentioned in [this comment](https://github.com/grafana/mimir/pull/11219#discussion_r2041860118), a change is also would be required to have more optimal way on getting `labels.Labels` size.

This the PR in Prometheus to get ByteSize(): https://github.com/prometheus/prometheus/pull/16717

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -1438,8 +1438,8 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			expr:          "sum(some_metric)",
 			shouldSucceed: true,
 
-			rangeQueryExpectedPeak: 9*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
-			rangeQueryLimit:        9*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
+			rangeQueryExpectedPeak: 8*types.FPointSize + 16*types.Float64Size + 8*types.BoolSize + types.SeriesMetadataSize,
+			rangeQueryLimit:        8*types.FPointSize + 16*types.Float64Size + 8*types.BoolSize + types.SeriesMetadataSize + 1,
 
 			instantQueryExpectedPeak: 9*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 			instantQueryLimit:        9*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
@@ -1605,7 +1605,7 @@ func TestMemoryConsumptionLimit_MultipleQueries(t *testing.T) {
 	opts := NewTestEngineOpts()
 	opts.CommonOpts.Reg = reg
 
-	limit := 3*8*types.FPointSize + 8*types.SeriesMetadataSize // Allow up to three series and its SeriesMetadatawith five points (which will be rounded up to 8, the nearest power of 2)
+	limit := 32*types.FPointSize + 4*types.SeriesMetadataSize + 3*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize())
 	engine, err := NewEngine(opts, NewStaticQueryLimitsProvider(limit), stats.NewQueryMetrics(reg), NewQueryPlanner(opts), log.NewNopLogger())
 	require.NoError(t, err)
 
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -1397,13 +1397,14 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			shouldSucceed: true,
 
 			// Each series has five samples, which will be rounded up to eight (the nearest power of two) by the bucketed pool,
-			// and we have five series and each of the series has SeriesMetadata. Five labelstring are also added at the end.
+			// and we have five series and each of the series has SeriesMetadata.
+			// Five out of SeriesMetadata has labels.Labels with each of them having the same ByteSize.
 			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 			rangeQueryLimit:        0,
 
 			// At peak, we'll hold all the output samples plus one series, which has one sample.
 			// The output contains five samples with SeriesMetadata, which will be rounded up to eight (the nearest power of two).
-			// Five labelstring are also added at the end.
+			// Five out of SeriesMetadata has labels.Labels with each of them having the same ByteSize.
 			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 			instantQueryLimit:        0,
 		},
@@ -1412,13 +1413,13 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			shouldSucceed: true,
 
 			// Each series has five samples with SeriesMetadata, which will be rounded up to 8 (the nearest power of two) by the bucketed pool, and we have five series.
-			// Five labelstring are also added at the end.
+			// Five out of SeriesMetadata has labels.Labels with each of them having the same ByteSize.
 			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 			rangeQueryLimit:        5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 
 			// At peak, we'll hold all the output samples plus one series, which has one sample.
 			// The output contains five samples with SeriesMetadata, which will be rounded up to 8 (the nearest power of two).
-			// Five labelstring are also added at the end.
+			// Five out of SeriesMetadata has labels.Labels with each of them having the same ByteSize.
 			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 			instantQueryLimit:        types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(labels.FromStrings(labels.MetricName, "some_metric", "idx", "i").ByteSize()),
 		},
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"testing"
 	"time"
+	"unsafe"
 
 	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
@@ -49,6 +50,9 @@ import (
 
 var (
 	spanExporter = tracetest.NewInMemoryExporter()
+	// TODO: move these two vars to appropriate places
+	nameLength = len("__name__")
+	stringSize = unsafe.Sizeof("")
 )
 
 func init() {
@@ -1398,12 +1402,12 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 
 			// Each series has five samples, which will be rounded up to 8 (the nearest power of two) by the bucketed pool,
 			// and we have five series and each of the series has SeriesMetadata.
-			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize,
+			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
 			rangeQueryLimit:        0,
 
 			// At peak, we'll hold all the output samples plus one series, which has one sample.
 			// The output contains five samples with SeriesMetadata, which will be rounded up to 8 (the nearest power of two).
-			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize),
+			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
 			instantQueryLimit:        0,
 		},
 		"limit enabled, but query does not exceed limit": {
@@ -1411,13 +1415,13 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			shouldSucceed: true,
 
 			// Each series has five samples with SeriesMetadata, which will be rounded up to 8 (the nearest power of two) by the bucketed pool, and we have five series.
-			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize,
-			rangeQueryLimit:        5*8*types.FPointSize + 8*types.SeriesMetadataSize,
+			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
+			rangeQueryLimit:        5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
 
 			// At peak, we'll hold all the output samples plus one series, which has one sample.
 			// The output contains five samples with SeriesMetadata, which will be rounded up to 8 (the nearest power of two).
-			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize),
-			instantQueryLimit:        types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize),
+			instantQueryExpectedPeak: types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
+			instantQueryLimit:        types.FPointSize + 8*(types.VectorSampleSize+types.SeriesMetadataSize) + 5*uint64(stringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
 		},
 		"limit enabled, and query exceeds limit": {
 			expr:          "some_metric",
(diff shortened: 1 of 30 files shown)


pkg/streamingpromql/operators/functions/histogram_function.go
```diff
@@ -233,6 +233,7 @@ func (h *HistogramFunction) SeriesMetadata(ctx context.Context) ([]types.SeriesM
 	h.remainingGroups = make([]*bucketGroup, 0, len(groups))
 	for _, g := range groups {
 		seriesMetadata = append(seriesMetadata, types.SeriesMetadata{Labels: g.labels.DropMetricName()})
+		h.memoryConsumptionTracker.IncreaseMemoryConsumptionForLabels(g.labels)
```
charleskorn: Shouldn't this be this?

```suggestion
		h.memoryConsumptionTracker.IncreaseMemoryConsumptionForLabels(g.labels.DropMetricName())
```

There are a bunch of places where we append a series and then update the memory consumption estimate. What if we introduced a wrapper `append` method that adds the labels to the slice and updates the memory consumption tracker? Then inconsistencies like this wouldn't be possible.

pkg/streamingpromql/types/limiting_pool.go
```diff
@@ -223,6 +223,11 @@ func (p *LimitingBucketedPool[S, E]) Put(s S, tracker *limiter.MemoryConsumption
 	}
 
 	tracker.DecreaseMemoryConsumption(uint64(cap(s))*p.elementSize, p.source)
+	if seriesMetadataSlice, isSeriesMetadataSlice := any(s).([]SeriesMetadata); isSeriesMetadataSlice {
+		for _, sm := range seriesMetadataSlice {
+			tracker.DecreaseMemoryConsumptionForLabels(sm.Labels)
+		}
+	}
```
charleskorn: Rather than adding this type check here, I would add an extra method as a field on `LimitingBucketedPool` that is called by `Put`. Then, when we create the `LimitingBucketedPool` for `SeriesMetadata` slices, we can pass a method that adjusts the memory consumption estimate based on the labels in the slice before we put the slice in the pool.

pkg/util/limiter/memory_consumption.go
```diff
@@ -54,8 +58,9 @@ const (
 	QuantileGroupSlices
 	TopKBottomKInstantQuerySeriesSlices
 	TopKBottomKRangeQuerySeriesSlices
+	SeriesMetadataLabels
```
charleskorn: [nit] I would just call this `Labels` or something like that - there's nothing that specifically links it to `SeriesMetadata`.

pkg/streamingpromql/engine_test.go
```diff
@@ -1451,26 +1455,26 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 
 			// Each series has five samples, which will be rounded up to 8 (the nearest power of two) by the bucketed pool,
 			// and we have five series and each of the series has SeriesMetadata.
-			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize,
+			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize + 5*uint64(limiter.StringSize)*uint64(nameLength+len("some_metric")+len("idx")+len("1")),
```
charleskorn: Once you've added a the size method you mention below to `labels.Labels`, I would expect these tests to use that as well.

pkg/streamingpromql/types/data.go
```diff
@@ -20,6 +20,26 @@ type SeriesMetadata struct {
 	Labels labels.Labels
 }
 
+type SeriesMetadataSlice []SeriesMetadata
+
+func (s SeriesMetadataSlice) Append(tracker *limiter.MemoryConsumptionTracker, lbs labels.Labels) (SeriesMetadataSlice, error) {
```
charleskorn: Do we need both `Append` and `AppendToOther`? Can't we have just one append method?

lamida: Yes we can. Let me do a quick refactoring.

lamida: Updated in 16a2ff0ec0d657de3e7e242eed6c3d00e6e85158

pkg/util/limiter/memory_consumption.go
lamida: Updated in 925309493132bf43c3f1cc8683b48b755041aa3f

pkg/streamingpromql/types/data.go
```diff
@@ -20,6 +20,19 @@ type SeriesMetadata struct {
 	Labels labels.Labels
 }
 
+type SeriesMetadataSlice []SeriesMetadata
+
+// Append appends this SeriesMetadataSlice with the provided otherSeriesMetadata.
+func (s SeriesMetadataSlice) Append(tracker *limiter.MemoryConsumptionTracker, otherSeriesMetadata ...SeriesMetadata) (SeriesMetadataSlice, error) {
```
charleskorn: The use of the special `SeriesMetadataSlice` type here feels a bit unnatural to me.

What if the signature was something like this?

```go
func AppendSeriesMetadata(tracker *limiter.MemoryConsumptionTracker, base []SeriesMetadata, others ...SeriesMetadata) (SeriesMetadataSlice, error) {

```

pkg/util/limiter/memory_consumption.go
```diff
@@ -181,3 +188,22 @@ func (l *MemoryConsumptionTracker) CurrentEstimatedMemoryConsumptionBytes() uint
 
 	return l.currentEstimatedMemoryConsumptionBytes
 }
+
+// IncreaseMemoryConsumptionForLabels attempts to increase the current memory consumption based on labels.
+func (l *MemoryConsumptionTracker) IncreaseMemoryConsumptionForLabels(lbs labels.Labels) error {
+	for _, lb := range lbs {
+		// TODO: Update labels.Labels to get size of bytes directly instead of calculating it here.
```
charleskorn: Are you planning to add this to `prometheus/prometheus` before merging this PR? 

If not, we'll need to modify this method to work with `stringlabels`.

pkg/streamingpromql/types/data.go
lamida: Yes we can change the method into function like that to make this more idiomatic. I don't have a strong opinion on that. Will change that as suggested.

pkg/util/limiter/memory_consumption.go
lamida: I will try to make the change on Prometheus first before merging the PR. I will try to draft the Prometheus PR today.

pkg/streamingpromql/types/limiting_pool.go
lamida: Updated in b6fa7d3c37891f325cf1e61c16917a50d0f07353

charleskorn: Thanks. It'd be good to see a test for this as well.

go.mod
```diff
@@ -344,7 +344,7 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v1.8.2-0.20250617052252-226f974e041b
+replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v1.8.2-0.20250619151923-05b1e6ec6c46
```
lamida: Note to reviewer: I am pointing this to https://github.com/grafana/mimir-prometheus/tree/lamida/bytes-string-labels while waiting https://github.com/prometheus/prometheus/pull/16717 get approved.

go.sum
```diff
@@ -573,8 +573,8 @@ github.com/grafana/memberlist v0.3.1-0.20250428154222-f7d51a6f6700 h1:0t7iOQ5ZkB
 github.com/grafana/memberlist v0.3.1-0.20250428154222-f7d51a6f6700/go.mod h1:Ri9p/tRShbjYnpNf4FFPXG7wxEGY4Nrcn6E7jrVa//4=
 github.com/grafana/mimir-otlptranslator v0.0.0-20250527173959-2573485683d5 h1:H4Del07v9UAYJgky9oeRY1oNqs6dh8lmHRUWiGCKXoE=
 github.com/grafana/mimir-otlptranslator v0.0.0-20250527173959-2573485683d5/go.mod h1:v1PzmPjSnNkmZSDvKJ9OmsWcmWMEF5+JdllEcXrRfzM=
-github.com/grafana/mimir-prometheus v1.8.2-0.20250617052252-226f974e041b h1:PazUdsbdAx9U0B7nIT0LWYogw/ezlO/x/HL93HVMzX4=
-github.com/grafana/mimir-prometheus v1.8.2-0.20250617052252-226f974e041b/go.mod h1:wxNDaTwYlAYXHIlsXbCj1xQZik2CB+GGqYW6LUJq6s4=
+github.com/grafana/mimir-prometheus v1.8.2-0.20250619151923-05b1e6ec6c46 h1:OGq1T/qsxmpLXnuwUctv+LW9r3/wefwaZ1MoQBBvFnA=
+github.com/grafana/mimir-prometheus v1.8.2-0.20250619151923-05b1e6ec6c46/go.mod h1:wxNDaTwYlAYXHIlsXbCj1xQZik2CB+GGqYW6LUJq6s4=
```
lamida: Note to reviewer: I am pointing this to https://github.com/grafana/mimir-prometheus/tree/lamida/bytes-string-labels while waiting https://github.com/prometheus/prometheus/pull/16717 get approved.

go.mod
charleskorn: Are you hoping to merge this PR before that Prometheus PR is accepted?

If so, we'll need to merge your change into mimir-prometheus `main` (otherwise it'll be lost next time we sync mimir-prometheus with upstream and bring that into Mimir).

pkg/streamingpromql/operators/functions/label.go
```diff
@@ -75,6 +80,7 @@ func LabelReplaceFactory(dstLabelOp, replacementOp, srcLabelOp, regexOp types.St
 		lb := labels.NewBuilder(labels.EmptyLabels())
 
 		for i := range seriesMetadata {
+			tracker.DecreaseMemoryConsumptionForLabels(seriesMetadata[i].Labels)
```
charleskorn: Should we only bother adjusting the memory consumption if the pattern matches? (ie. move this `DecreaseMemoryConsumptionForLabels` call and the later `IncreaseMemoryConsumptionForLabels` call inside the `if indexes != nil { ...` block)

pkg/streamingpromql/operators/binops/and_unless_binary_operation.go
```diff
@@ -143,6 +143,7 @@ func (a *AndUnlessBinaryOperation) computeAndSeriesMetadata(leftMetadata []types
 			// This series doesn't match any series from the right side.
 			// Discard the group.
 			a.leftSeriesGroups[seriesIdx] = nil
+			a.MemoryConsumptionTracker.DecreaseMemoryConsumptionForLabels(leftMetadata[seriesIdx].Labels)
 		} else {
 			leftMetadata[nextOutputSeriesIndex] = leftMetadata[seriesIdx]
 			nextOutputSeriesIndex++
```
charleskorn, line 149: Before we return `leftMetadata[:nextOutputSeriesIndex]` below, we should probably clear out any `SeriesMetadata` elements in the remainder of the slice, to avoid holding onto `labels.Labels` instances we don't need.

(this was true before your changes here, but your changes made it more obvious that this was missing)

pkg/streamingpromql/engine_test.go
```diff
@@ -1396,28 +1396,31 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			expr:          "some_metric",
 			shouldSucceed: true,
 
-			// Each series has five samples, which will be rounded up to 8 (the nearest power of two) by the bucketed pool,
-			// and we have five series and each of the series has SeriesMetadata.
-			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize,
+			// Each series has five samples, which will be rounded up to eight (the nearest power of two) by the bucketed pool,
+			// and we have five series and each of the series has SeriesMetadata. Five labelstring are also added at the end.
```
charleskorn: The change to the comment here isn't clear to me - could you explain what you mean by "Five labelstring are also added at the end." here?

pkg/streamingpromql/engine_test.go
```diff
@@ -1435,51 +1438,29 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			expr:          "sum(some_metric)",
 			shouldSucceed: true,
 
-			// Each series has five samples, which will be rounded up to 8 (the nearest power of two) by the bucketed pool.
-			// At peak we'll hold in memory:
-			//  - the running total for the sum() (two floats (due to kahan) and a bool at each step, with the number of steps rounded to the nearest power of 2),
-			//  - the next series from the selector
-			//  - the labels for the output series
```
charleskorn, line 1442: Is there a reason why you've removed this comment? If it is outdated, then it might be helpful to replace it with an updated explanation.

Same comment applies below.

go.mod
lamida: I am fine for waiting this PR to be merged after prometheus PR is merged, unless if you think we should do before that.

pkg/streamingpromql/engine_test.go
lamida: Maybe shouldn't use "added at the end". I tried to tell the peak memory consumption also include 5 labelstrings. Let me update that.

pkg/streamingpromql/engine_test.go
lamida: I just start to find hard to track the comment here because one small change to **what** we track, **where** we track or change in code logic will immediately make the comment outdated. Furthermore we now have `currentEstimatedMemoryConsumptionBySource` in `MemoryConsumptionTracker` which make easier if we want peek what are the breakdown of the total bytes. 

I am proposing to just remove all the comments in this test.

pkg/streamingpromql/engine_test.go
charleskorn: I think "added at the end" is a little confusing, as is the use of the term "labelstring", which isn't a term we use elsewhere.

pkg/streamingpromql/engine_test.go
charleskorn: OK, makes sense to me.

pkg/streamingpromql/operators/functions/label.go
lamida: Yes that works. Updated in ca761e1848a7e15b2520c1684e4eeb23e009d190.

pkg/streamingpromql/operators/binops/and_unless_binary_operation.go
lamida: Updated in 8794c29e5a159f42bda7e043cfe374792b328dd5

pkg/streamingpromql/operators/binops/and_unless_binary_operation.go
```diff
@@ -143,13 +143,19 @@ func (a *AndUnlessBinaryOperation) computeAndSeriesMetadata(leftMetadata []types
...
 			a.lastLeftSeriesIndexToRead = seriesIdx
 		}
 	}
 
+	// Clear up labels that we don't need anymore.
+	for i := nextOutputSeriesIndex; i < len(leftMetadata); i++ {
+		leftMetadata[i].Labels = nil
+	}
+
```
charleskorn: Couldn't we use `clear(leftMetadata[nextOutputSeriesIndex:])`?

lamida: Didn't think on that. Updated as suggested in  c664a09f715c1008beace3485b210f8e96377818

go.mod
lamida: @charleskorn on the second thought, I think let's merge the promethes change to mimir-prometheus first. I created this PR: https://github.com/grafana/mimir-prometheus/pull/908. Let me know if that is not the right way to do that.

pkg/streamingpromql/engine_test.go
lamida: Tried to update in 8add546e30caaf1b48567398c1c889b3b88305b1.

go.mod
charleskorn: Let's give the Prometheus PR another day or two, it'll save us a bit of faffing about if it's merged soon.

pkg/streamingpromql/engine_test.go
```diff
@@ -1396,28 +1396,32 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			expr:          "some_metric",
 			shouldSucceed: true,
 
-			// Each series has five samples, which will be rounded up to 8 (the nearest power of two) by the bucketed pool,
+			// Each series has five samples, which will be rounded up to eight (the nearest power of two) by the bucketed pool,
 			// and we have five series and each of the series has SeriesMetadata.
```
charleskorn: Maybe this would be clearer?

```suggestion
			// and we have five series and each of the series has labels of the same size.
```

(same applies below)

pkg/streamingpromql/engine_test.go
```diff
@@ -1396,28 +1396,32 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
...
-			rangeQueryExpectedPeak: 5*8*types.FPointSize + 8*types.SeriesMetadataSize,
+			// Five out of SeriesMetadata has labels.Labels with each of them having the same ByteSize.
```
charleskorn: No longer required with suggestion above:

```suggestion
```

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -125,6 +125,12 @@ func (a *Aggregation) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
 	}
 
 	defer types.SeriesMetadataSlicePool.Put(&innerSeries, a.MemoryConsumptionTracker)
+	// TODO: place this as part of Put call above.
+	defer func() {
+		for _, s := range innerSeries {
+			a.MemoryConsumptionTracker.DecreaseMemoryConsumptionForLabels(s.Labels)
+		}
+	}()
```
charleskorn: Is this still needed? Looks like this should be handled by the `Put` hook now.

pkg/streamingpromql/operators/test_operator.go
```diff
@@ -39,6 +39,10 @@ func (t *TestOperator) SeriesMetadata(_ context.Context) ([]types.SeriesMetadata
 	seriesPool = seriesPool[:len(t.Series)]
```
charleskorn: [nit, realise this was here before] `seriesPool` doesn't make a lot of sense in this context - how about just `series` or `metadata` or something like that?

pkg/streamingpromql/query.go
```diff
@@ -207,6 +207,12 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
 			return &promql.Result{Err: err}
 		}
 		defer types.SeriesMetadataSlicePool.Put(&series, q.memoryConsumptionTracker)
+		// TODO: place this as part of Put call above.
+		defer func() {
+			for _, s := range series {
+				q.memoryConsumptionTracker.DecreaseMemoryConsumptionForLabels(s.Labels)
+			}
+		}()
```
charleskorn: Is this still needed? Seems like it would be covered by the hook as well.

pkg/streamingpromql/operators/aggregations/aggregation.go
lamida: Remnant from rebase process. I will remove that.

pkg/streamingpromql/query.go
lamida: Remnant from rebase process. I will remove that.

pkg/streamingpromql/operators/test_operator.go
lamida: Updated in 541526b08f8f373d450fce12822a89267fab9bf0

lamida:
@charleskorn I will check again why the test is failing in CI (while it works fine in local). But I think the PR is ready for another review.

charleskorn:
> @charleskorn I will check again why the test is failing in CI (while it works fine in local). But I think the PR is ready for another review.

Are you running the tests with `-tags stringlabels` locally?

lamida:
@charleskorn no I was not. Just running with default `make test` target. Adding the tags fail the test in local too. Fixing it now. Thanks!

charleskorn (approved):
🚀

pkg/streamingpromql/engine_test.go
dimitarvdimitrov: +1 for keeping the comment. Each time the code changes, you also need to update the assertion, so at least then someone will know why the assertion is changing (and can update the comment). The test is failing on main now and I have no idea where these numbers don't add up

PR #11686, opened 2025-06-10
store-gateway: download sparse headers on startup when lazy loading is enabled

dimitarvdimitrov:
Without this the store-gateway would only download the sparse header and store it on disk when the block is queried. For clusters with many tenants, this means the store-gateway spends roughly 3 minutes downloading sparse headers and this contributes (not sure how much) to a spike in latency after store-gateway scale-out

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/indexheader/lazy_binary_reader.go
+++ b/pkg/storage/indexheader/lazy_binary_reader.go
@@ -6,6 +6,7 @@
 package indexheader
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
@@ -26,6 +27,7 @@ import (
 
 	streamindex "github.com/grafana/mimir/pkg/storage/indexheader/index"
 	"github.com/grafana/mimir/pkg/storage/tsdb/block"
+	"github.com/grafana/mimir/pkg/util/atomicfs"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/storage/indexheader/lazy_binary_reader.go
```diff
@@ -123,34 +125,48 @@ func NewLazyBinaryReader(
...
 		}
 
-		level.Debug(logger).Log("msg", "built index-header file", "path", path, "elapsed", time.Since(start))
+		level.Debug(logger).Log("msg", "built index-header file", "path", indexHeaderPath, "elapsed", time.Since(start))
+	}
+
+	sparseHeaderPath := filepath.Join(dir, id.String(), block.SparseIndexHeaderFilename)
+
+	if _, err := os.Stat(sparseHeaderPath); err != nil {
+		bucketSparseHeaderBytes, err := tryDownloadSparseHeader(ctx, logger, bkt, id)
```
dmwilson-grafana: Two questions: 

1. Do sparse headers and index headers need to be downloaded sequentially? If not, do we want to do these downloads in parallel to save on startup time?
2. Can we save on startup time by building sparse headers from the downloaded index headers vs. downloading from object storage?

dmwilson-grafana:
I'd update the [comment](https://github.com/grafana/mimir/pull/11686/files#diff-78ee27df00399db0bd60407169d200d392224a2938ed14ed847802232d0894adL118-L121) on `NewLazyBinaryReader` to reference sparse headers as well.

pkg/storage/indexheader/lazy_binary_reader.go
dimitarvdimitrov: > 1. Do sparse headers and index headers need to be downloaded sequentially? If not, do we want to do these downloads in parallel to save on startup time?

fair, i added it in this commit https://github.com/grafana/mimir/pull/11686/commits/1933622cc871e79fd52101baaee60f8fd4ca90d3


> 2. Can we save on startup time by building sparse headers from the downloaded index headers vs. downloading from object storage?

constructing from the index header is usually many order of magnitude slower than downloading from the bucket. That's the reason we started persisting sparse headers in the first place.

PR #11689, opened 2025-06-11
Set explicit version for Minio in development environments to ensure `development/tools/upload-block` works with latest `aws` CLI

charleskorn:
#### What this PR does

`development/tools/upload-block` uses the AWS CLI to upload files to the local Minio instance used in our Docker Compose environments.

Recent versions of the AWS CLI trigger the issue described in https://github.com/minio/minio/issues/19670, which is fixed in recent versions of Minio.

However, we had not set an explicit Docker image tag when pulling Minio, so there's no guarantee that Minio is any particular version.

So in this PR, I'm pinning Minio to the most recent stable version, to ensure this fix is included.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/development/mimir-ingest-storage/docker-compose.jsonnet
+++ b/development/mimir-ingest-storage/docker-compose.jsonnet
@@ -126,7 +126,7 @@ std.manifestYamlDoc({
 
   minio:: {
     minio: {
-      image: 'minio/minio',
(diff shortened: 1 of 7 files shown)


narqo (approved):
🔥

PR #11690, opened 2025-06-11
MQE: fix possible panic while evaluating a query containing `group_left` or `group_right`

charleskorn:
#### What this PR does

This PR fixes an issue where queriers could panic with `estimated memory consumption of this query is negative` while evaluating a query containing a one-to-many or many-to-one binary operation (aka `group_left` / `group_right`).

Specifically, the issue is triggered if all of the series with the same group matching labels on the "one" side are read and then the operator is closed before all of the corresponding output series have been emitted. (This can happen if, for example, the query contains further binary operations that only need a subset of the output series.)

In this case, the `matchGroup`'s `presence` slice would be returned to the pool multiple times: once in `updateOneSidePresence`, and again when `oneSide.Close` was called by the first remaining output series in that group.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
+++ b/pkg/streamingpromql/operators/binops/grouped_vector_vector_binary_operation.go
@@ -683,6 +683,7 @@ func (g *GroupedVectorVectorBinaryOperation) updateOneSidePresence(side *oneSide
 
 	if matchGroup.oneSideCount == 0 {
 		types.IntSlicePool.Put(matchGroup.presence, g.MemoryConsumptionTracker)
+		matchGroup.presence = nil
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


charleskorn:
> All instances of `types.SomeSlicePool.Put(slice, ..)` is always followed by setting `slice = nil`. Should we nil the slice inside the Put instead to avoid missed setting that up like this?

We can't nil the slice inside the `Put`, but we might be able to do something if `Put` returns `nil` - if there's a linting rule to ensure that return values aren't ignored, then this would help catch issues like these. However, last time I looked into this, such a linter didn't exist.

bboreham:
Another way to go would be to have `Put` take a pointer to the slice, then you could overwrite it.

This would be a bit more ugly in the code, but perhaps easier to use?

PR #11691, opened 2025-06-11
Enforce that methods on the same type have the same receiver name

charleskorn:
#### What this PR does

This PR enables a linting rule that enforces that all methods on the same type have the same receiver name, and fixes instances where this is not already the case.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


lamida (approved):
LGTM

PR #11710, opened 2025-06-11
Block-builder-scheduler: better handling of gaps in job specs

seizethedave:
#### What this PR does

* Adds scheduling gap detection as a new warning log and metric (`cortex_blockbuilder_scheduler_gaps_detected`).
* Improves "startup mode:" Startup mode now computes the "latest planned offset" from observed job reports. The LPO is the largest observed endOffset such that there are no "gaps" between jobs starting from the partition's commit offset.
* Improves initial job planning by resuming from the latest planned offset rather than the committed offset.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


seizethedave:
I've closed in favor of #11867

PR #11712, opened 2025-06-12
Don't emit unnecessary "Ingester.QueryStream" trace spans

charleskorn:
#### What this PR does

This PR improves the clarity of traces emitted during queries by removing unnecessary "Ingester.QueryStream" trace spans emitted by ingesters that are nested beneath "cortex.Ingester/QueryStream" trace spans also emitted by ingesters.

Instead, any events that were previously attached to the "Ingester.QueryStream" span are now attached to the "cortex.Ingester/QueryStream" span. 

The only major difference is that the list of tenant IDs will no longer be on a trace span emitted by ingesters, but this information is available on parent spans emitted by other components (eg. the "Distributor.queryIngesterStream" span emitted by queriers).

Before:

Screenshot 2025-06-12 at 2 39 13 pm

After:

Screenshot 2025-06-12 at 2 40 06 pm

I haven't added a changelog entry given this is a minor change to tracing.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -2203,7 +2203,8 @@ const queryStreamBatchMessageSize = 1 * 1024 * 1024
 func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_QueryStreamServer) (err error) {
 	defer func() { err = i.mapReadErrorToErrorWithStatus(err) }()
 
-	spanlog, ctx := spanlogger.New(stream.Context(), i.logger, tracer, "Ingester.QueryStream")
+	ctx := stream.Context()
+	spanlog := spanlogger.FromContext(ctx, i.logger)
 	defer spanlog.Finish()
 
 	userID, err := tenant.TenantID(ctx)


pkg/ingester/ingester.go
```diff
@@ -2203,7 +2203,8 @@ const queryStreamBatchMessageSize = 1 * 1024 * 1024
 func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_QueryStreamServer) (err error) {
 	defer func() { err = i.mapReadErrorToErrorWithStatus(err) }()
 
-	spanlog, ctx := spanlogger.New(stream.Context(), i.logger, tracer, "Ingester.QueryStream")
+	ctx := stream.Context()
+	spanlog := spanlogger.FromContext(ctx, i.logger)
 	defer spanlog.Finish()
```
dimitarvdimitrov, line 2208: do we still want to finish this span here? wouldn't the parent span be finished?

colega: +1

charleskorn: 🤦 Good catch, I've fixed this in #11726.

PR #11713, opened 2025-06-12
MQE: include trace ID in panic message if memory consumption estimate becomes negative

charleskorn:
#### What this PR does

This PR adds the trace ID (if any) to the panic message used if the estimated memory consumption of a query becomes negative.

#11615, #11654

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/util/limiter/memory_consumption.go
+++ b/pkg/util/limiter/memory_consumption.go
@@ -163,10 +163,10 @@ func (l *MemoryConsumptionTracker) DecreaseMemoryConsumption(b uint64, source Me
 	defer l.mtx.Unlock()
 
 	if b > l.currentEstimatedMemoryConsumptionBySource[source] {
-		traceID, sampled := tracing.ExtractSampledTraceID(l.ctx)
+		traceID, ok := tracing.ExtractTraceID(l.ctx)
 		traceDescription := ""
 
-		if sampled {
+		if ok {
 			traceDescription = fmt.Sprintf(" (trace ID: %v)", traceID)
 		}
 


pkg/util/limiter/memory_consumption.go
```diff
@@ -159,7 +163,14 @@ func (l *MemoryConsumptionTracker) DecreaseMemoryConsumption(b uint64, source Me
 	defer l.mtx.Unlock()
 
 	if b > l.currentEstimatedMemoryConsumptionBySource[source] {
-		panic(fmt.Sprintf("Estimated memory consumption of all instances of %s in this query is negative. This indicates something has been returned to a pool more than once, which is a bug. The affected query is: %v", source, l.queryDescription))
+		traceID, sampled := tracing.ExtractSampledTraceID(l.ctx)
+		traceDescription := ""
+
+		if sampled {
+			traceDescription = fmt.Sprintf(" (trace ID: %v)", traceID)
+		}
```
colega: Given you're going to panic, I think you should log the trace ID even if it wasn't sampled. You may still find that trace in some logs from other components that would help you debugging. I'd just take `trace.SpanFromContext(l.ctx).SpanContext().TraceID()`

charleskorn: Is there an advantage in using `trace.SpanFromContext(l.ctx).SpanContext().TraceID()` over using `tracing.ExtractTraceID(l.ctx)` here?

colega: You can also use `tracing.ExtractTraceID()` 👍

colega: Also: https://github.com/grafana/dskit/pull/715
When that's done, you can just remove the `sampled` condition.

charleskorn: Changed in 7b0dcfa.

PR #11719, opened 2025-06-12
Query-frontend: accurate tracking of samples processed from cache

Konstantinov-Innokentii:
#### What this PR does
This PR leverages perStepStats to accurately track samples processed from cache, without approximation. 
It's a follow-up for https://github.com/grafana/mimir/pull/11164 and https://github.com/grafana/mimir/pull/11416

#### How it works

- Added `GetStats() `and `SetStats()` MetricsReqiestInterface  and its implementations. It allows to manipulate a "stats" query param, which is required for MQE and Prometheus Query Engine to return PerStepStats.
- This param is always set in the [SplitAndCacheMiddleware](https://github.com/grafana/mimir/pull/11719/files#diff-655764486d839037085bd64ff1e3f08c12902660a2a083608ac67f55c07e3434R155), so each request will have a PerStepStats gathered.
- PerStepStats are [stored](https://github.com/grafana/mimir/pull/11719/files) in Extents instead of approximation.
- PerStepStats from cache are [merged](https://github.com/grafana/mimir/pull/11719/files#diff-655764486d839037085bd64ff1e3f08c12902660a2a083608ac67f55c07e3434R326) with the PerStepStats from a downstream request and sum is reported in the `cortex_query_samples_processed_cache_adjusted_total`
- Bunch of [integration tests ](https://github.com/grafana/mimir/pull/11719/files#diff-5d39b940451aba1e23d7e664ca574f63bbe521344e67670e9beefe9b082c348c)for that 

**Note**, that this functionality is under the `query-frontend.cache-samples-processed-stats` feature flag, so It's needed to enable this flag progressively for each release wave, to keep PerStepStats collection after the release. It also affects behaviour of the `cortex_query_samples_processed_cache_adjusted_total` metric. If `query-frontend.cache-samples-processed-stats` is not enabled if will report 0 samples.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -7514,7 +7514,7 @@
           "kind": "field",
           "name": "cache_samples_processed_stats",
           "required": false,
-          "desc": "Cache Statistics of processed samples on results cache.",
+          "desc": "Cache statistics of processed samples on results cache.",
           "fieldValue": null,
           "fieldDefaultValue": false,
           "fieldFlag": "query-frontend.cache-samples-processed-stats",
--- a/cmd/mimir/help-all.txt.tmpl
+++ b/cmd/mimir/help-all.txt.tmpl
@@ -2330,7 +2330,7 @@ Usage of ./cmd/mimir/mimir:
   -query-frontend.cache-results
     	Cache query results.
   -query-frontend.cache-samples-processed-stats
-    	Cache Statistics of processed samples on results cache.
+    	Cache statistics of processed samples on results cache.
   -query-frontend.cache-unaligned-requests
     	Cache requests that are not step-aligned.
   -query-frontend.client-cluster-validation.label string
--- a/cmd/mimir/help.txt.tmpl
+++ b/cmd/mimir/help.txt.tmpl
@@ -582,7 +582,7 @@ Usage of ./cmd/mimir/mimir:
   -query-frontend.cache-results
     	Cache query results.
   -query-frontend.cache-samples-processed-stats
-    	Cache Statistics of processed samples on results cache.
+    	Cache statistics of processed samples on results cache.
   -query-frontend.log-queries-longer-than duration
     	Log queries that are slower than the specified duration. Set to 0 to disable. Set to < 0 to enable on all queries.
   -query-frontend.max-queriers-per-tenant int


cmd/mimir/config-descriptor.json
dimitarvdimitrov, line 1: do we need these to be configurable? is there a tradeoff for having them always enabled?

pkg/frontend/querymiddleware/codec.go
```diff
@@ -138,6 +138,8 @@ type MetricsQueryRequest interface {
 	GetHints() *Hints
 	// GetLookbackDelta returns the lookback delta for the request.
 	GetLookbackDelta() time.Duration
+	// GetStats returns the stats parameter for the request.
+	GetStats() string
```
dimitarvdimitrov, line 143: can we use an enum type here instead of a string?

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -474,18 +473,24 @@ func newAccumulator(base Extent) (*accumulator, error) {
...
+func toExtent(ctx context.Context, req MetricsQueryRequest, res Response, queryTime time.Time, perStepStats []stats.StepStat) (Extent, error) {
 	marshalled, err := types.MarshalAny(res)
 	if err != nil {
 		return Extent{}, err
 	}
+
+	extentPerStepStats := make([]StepStat, len(perStepStats))
+	for i, stat := range perStepStats {
+		extentPerStepStats[i] = StepStat(stat)
+	}
```
dimitarvdimitrov, line 485: do you have any numbers on how much this would increase the cache size? should we try to delta-encode the per-step stats for example?

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -271,10 +284,11 @@ func (s *splitAndCacheMiddleware) Do(ctx context.Context, req MetricsQueryReques
 				downstreamRes := splitReq.downstreamResponses[downstreamIdx]
 				downstreamStats := splitReq.downstreamStatistics[downstreamIdx]
 				if !isResponseCachable(downstreamRes) {
+					fmt.Println("Response not cachable")
```
dimitarvdimitrov: leftover debug statement; there's one above too

pkg/querier/stats/stats.proto
```diff
@@ -15,7 +15,7 @@ import "google/protobuf/duration.proto";
 option (gogoproto.marshaler_all) = true;
 option (gogoproto.unmarshaler_all) = true;
 
-message Stats {
+message StatsData {
```
dimitarvdimitrov: this looks unnecessary

pkg/frontend/querymiddleware/codec.go
```diff
@@ -365,8 +369,9 @@ func (c prometheusCodec) decodeRangeQueryRequest(r *http.Request) (MetricsQueryR
 	var options Options
 	decodeOptions(r, &options)
 
+	stats := reqValues.Get("stats")
```
dimitarvdimitrov, line 377: there is already an HTTP header which requests stats for the `server-timing` response header, can we reuse that? 

https://github.com/grafana/mimir/blob/8d8a0a82dc8b65222d389e0dfe8c291ec7869edb/pkg/frontend/transport/handler.go#L46

Konstantinov-Innokentii: I'm not sure. The problem is that Prometheus Query Engine will ReportPerStep stats under two conditions:
1. Feature flag is enabled
2. Request has a stats="all" query param.
https://prometheus.io/docs/prometheus/latest/feature_flags/#per-step-stats

Yes, we currently use MQE, but it might be the case that it's disabled in favour of Prometheus Query Engine, because of a bug or something, so that's why I had to implement what Prometheus wants.

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: This PR will not increase the cache size; it changes the method of calculating cachedStats. However, they have been stored in a cache for ~ a month now after [this](https://github.com/grafana/mimir/pull/11164) was merged and released.
Speaking about numbers, it's protobuf encoded in cache, so I didn't do any real math around using delta-encoding, assuming that it's good enough for the start.

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: I'm not sure, Prometheus interface is to pass stats="all" to enable PerStepStats collection.

pkg/querier/stats/stats.proto
Konstantinov-Innokentii: Well, I could use some help with that.  The story is that I need to add a mutex to make Stats.PerStepStats array concurrency-safe, especially when merging stats.
To achieve this, I [renamed](https://github.com/grafana/mimir/pull/11719/files#diff-8ca2404f80bab2e8d434ef41516c62c98fa2d642471ba0d8e29006c38eecb5deR18) the Protobuf message  StatsData and [wrapped](https://github.com/grafana/mimir/pull/11719/files#diff-25b9719633031df951dfb7a15155940c34efc0f34c2f8b18947bc8335fb3d465R46) it in a new Stats struct that includes a mutex.
However, frontend.proto uses the Stats message in the ClientToFrontend message, which is leading to some lint [errors](https://github.com/grafana/mimir/actions/runs/15630408459/job/44033201239?pr=11719#step:8:11).
The solution I'm considering is to rename StatsData back to Stats and introduce a new WrappedStats type that embeds Stats and includes a mutex. This way, the Protobuf-generated Stats would be used only when sending data between the querier and the frontends, while the rest of the code would interact with WrappedStats.
The downside is that I’d need to write code to convert between Stats and WrappedStats when encoding and decoding responses.

cmd/mimir/config-descriptor.json
Konstantinov-Innokentii: I mostly followed Prometheus and Cortex implementations here. My thoughts however, is that PerStepStats are not really needed by default for OSS or GEM users, so I put it behind feature flag.

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: maybe i'm missing something. Isn't the `MetricsQueryRequest` interface something that we define? What prometheus interface do you mean?

Konstantinov-Innokentii: I mean the contract between Prometheus and clients to get PerStepStats - https://prometheus.io/docs/prometheus/latest/feature_flags/#per-step-stats.
I'm sure that's me missing something - what's pros of using enum here?

pkg/frontend/querymiddleware/results_cache.go
dimitarvdimitrov: yeah i wanted to understand if it's good enough for the start. a 90-day query on my monitor has 1957 points. So that'd be 32K additional data we store in the cache. For a `count(vector(1))` query that's ~29% of the 110KB JSON response. Do we really need to store all timestamps though? We can store start,end,interval and it's as good as knowing all the timestamps.

did you monitor cache performance after this change?

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: Bit more details - here is some Prom code, which builds QueryOptions from request:
```
func extractQueryOpts(r *http.Request) (promql.QueryOpts, error) {
	var duration time.Duration

	if strDuration := r.FormValue("lookback_delta"); strDuration != "" {
		parsedDuration, err := parseDuration(strDuration)
		if err != nil {
			return nil, fmt.Errorf("error parsing lookback delta duration: %w", err)
		}
		duration = parsedDuration
	}

	return promql.NewPrometheusQueryOpts(r.FormValue("stats") == "all", duration), nil
}
```
and the `r.FormValue("stats") == "all"` goes to **enablePerStepStats** param of the `NewPrometheusQueryOpts`

We reuse this code in Querier, so that's why we have to pass "stats".

pkg/querier/stats/stats.proto
dimitarvdimitrov: you can use a cusotm go type in the proto definition like here https://github.com/grafana/mimir/blob/04aa4a2c850465d38bc65d2d44e8923c6febb17c/pkg/frontend/querymiddleware/model.proto#L41

you'd still declare the proto message in the .proto file, but the gRPC service will be returning your custom type, which in turn would embed the generated protobuf type. 

This is practically the same as what you described as a second option.

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: i didn't know prometheus supports this too. So is your plan to also return these stats in the client response too?

It's unfortunate both projects implemented the same. Can we at least make it less confusing to use? I'm thinking that we can treat the header and the query param would mean the same thing - include `Server-Timing` response and also return per-step statistics.

pkg/querier/stats/stats.pb.go
```diff
@@ -54,19 +54,21 @@ type Stats struct {
 	SamplesProcessed uint64 `protobuf:"varint,11,opt,name=samples_processed,json=samplesProcessed,proto3" json:"samples_processed,omitempty"`
 	// The number of subqueries that were spun off as actual range queries in order to execute the full query
 	SpunOffSubqueries uint32 `protobuf:"varint,12,opt,name=spun_off_subqueries,json=spunOffSubqueries,proto3" json:"spun_off_subqueries,omitempty"`
+	// SamplesProcessedPerStep represents the number of samples scanned per step timestamp while evaluating a query.
+	SamplesProcessedPerStep []StepStat `protobuf:"bytes,13,rep,name=samples_processed_per_step,json=samplesProcessedPerStep,proto3" json:"samples_processed_per_step"`
```
dimitarvdimitrov, line 58: Have you considered having a different format for this? Right now if we wanted to add another type of statistics that we want to track per step, we would have to duplicate this array. So we'd have two arrays: one for SamplesProcessed and one for the next statistic. At least the timestamps would repeat.

WDYT about having single array where every element is a timestamp and a multiple values which have names. So in this case the field here would be called PerStepStatistics. And then the StepStats would be something like {Timestamp, SamplesProcessed}. (again - not sure if we need to store every timestamp or if start,end,step are enough)

cmd/mimir/config-descriptor.json
dimitarvdimitrov: isn't `querier.per-step-stats-enabled` also controlled by the query parameter - why also have a config option? Only `query-frontend.cache-samples-processed-stats` is something that's not controllable per-query and could use its own config option.

Konstantinov-Innokentii: > isn't querier.per-step-stats-enabled also controlled by the query parameter - why also have a config option?

To report PerStep stats, Prometheus wants **both** config flag and the query parameter passed.
Under the hood, the `per-step-stats-enabled ` flag is passed down to the QueryEngine. When creating a new query, Prometheus reads the `r.Form("stats")`and if parameter is passed and engine has it's enabled too, PerStepStas will be reported.


In Mimir we reuse prometheus code which reads the `r.Form("stats")`–  I described it [here](https://github.com/grafana/mimir/pull/11719#discussion_r2145165720). So, it seems we anyway need to pass it, but it's true that we have some options around the `per-step-stats-enabled` flag.
1. Introduce `querier.per-step-stats-enabled` 
2. hardcode `enablePerStepStas=true`when creating an Engine in Querier.

I find the config a bit more elegant - there are no hardcoded values and it mimics Prometheus behaviour.

pkg/querier/stats/stats.proto
Konstantinov-Innokentii: Thanks, I'll try that.

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: Could you please give more context about what "X-Mimir-Response-Query-Stats" is doing? I'm not sure I get the idea of "treat the header and the query param would mean the same thing" 

> So is your plan to also return these stats in the client response too?

It depends on what do you mean when saying client response.
Prometheus returns PerStepStats in it's response, alongside with some other stats. Here is an [example](https://github.com/prometheus/prometheus/pull/10369) of it. I'll call it _PrometheusResponse_.
Mimir wraps this into its own structs when passing it from querier to query-frontend. Also, mimir doesn't; use PrometheusResponse.stats, but use querier stats instead:
```
type ClientToFrontend struct {
	HttpResponse *httpgrpc.HTTPResponse `protobuf:"bytes,1,opt,name=httpResponse,proto3" json:"httpResponse,omitempty"`
	ClientID     string                 `protobuf:"bytes,2,opt,name=clientID,proto3" json:"clientID,omitempty"`
	Stats        *stats.Stats           `protobuf:"bytes,3,opt,name=stats,proto3" json:"stats,omitempty"`
}
```
So, that's why I'm using querier [Stats](https://github.com/grafana/mimir/pull/11719/files#diff-25b9719633031df951dfb7a15155940c34efc0f34c2f8b18947bc8335fb3d465R46) to track PerStepStats and pass them from Querier to Frontend as well.
The puzzle here is that when you pass "stats" parameter in the request, it will make prometheus code to render stats inside a _PrometheusResponse_, which is unnecessary, because we don't use them at all. So, I modified the [statsRenderer](https://github.com/grafana/mimir/pull/11719/files#diff-57b872fd2ee77f0d08b0240d7827df78209df1e331f3e5dbd56741dbf7e2c0b1R19).

I draw a little diagram, when was crawling through it
image

cmd/mimir/config-descriptor.json
dimitarvdimitrov: Right now, the reason to add a config option is for consistency in operating Mimir and operating Prometheus. We shouldn't be trying to have consistency there - they are two very different systems and have diverged a long time ago.

i'd be happy to add a config option, but we have to understand it first. If there is a practical reason to add the option, sure; otherwise, we're just piling up options, which even we can't recommend when to have on and when to have off.

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: the code you included is code which parses an `*http.Request`. It parses it into a QueryOpts struct that the rest of the code can use. Our `MetricsQueryRequest` should be a bit higher level than just HTTP requests. 

> We reuse this code in Querier, so that's why we have to pass "stats".

We can still pass a `stats=all` query parameter when we're serializing the request back into an `*http.Request`.

Some nice things that come with using a proper type as opposed to string-typing is that you can easily see what are the possible values and validate them. Is an empty string valid? You can check by seeing the enum. It also prevents rare bugs where we pass a random string as the stats.

It's also a bit self-documenting. When I read this code, I thought stats would be the actual statistics (e.g. `"samples_processed=10"`), but it's actually a request to compute statistics. (This makes me wonder if there isn't also a better name for `GetStats()` and `WithStats()`, but I can't think of one)

cmd/mimir/config-descriptor.json
Konstantinov-Innokentii: If consistency here is not a concern, I can definitely adjust querier to always enable per step stats in engine and rely on a stats param only 👍🏻

pkg/querier/stats_renderer.go
```diff
@@ -15,6 +14,23 @@ func StatsRenderer(ctx context.Context, s *promql_stats.Statistics, param string
 	mimirStats := stats.FromContext(ctx)
 	if mimirStats != nil && s != nil {
 		mimirStats.AddSamplesProcessed(uint64(s.Samples.TotalSamples))
+		mimirStats.AddSamplesProcessedPerStep(totalSamplesPerStepPoints(s.Samples))
 	}
-	return prom_api.DefaultStatsRenderer(ctx, s, param)
+	// Mimir doesn't use stats from Prometheus Response in favour of mimirStats above.
+	return nil
```
dimitarvdimitrov, line 20: we still claim that our querier has a prometheus-compatible interface. If we want this to stay true, we have to keep these stats :(

At the same time, I see that we will transport the stats twice, which is not great, so I'm tempted to make this breaking change now and document it as caveat until 3.0. I added it to "2025-04 Mimir 3.0 Features and Changes" as a proposal where we can break this compatibility promise. Do you want to bring this up during a weekly call where we can discuss it?

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: `x-mimir-response-query-stats` packages up all the query stats we normally log in the query frontend in an HTTP response header called server-timing. The idea was to allow clients like Grafana to visualize them.

I meant that the header `x-mimir-response-query-stats` and `stats=all` have the same effect: enrich toggle the `server-timing` response header and the `stats` in the JSON response. WDYT?

pkg/querier/stats_renderer.go
Konstantinov-Innokentii: I didn't know that; thanks for pointing it out. With that knowledge, I wonder if we should do the same as Cortex did—do not pass PerStep stats in the querier stats, but pass them in the PrometheusResponse and store in Extent response. Here is the [PR](https://github.com/cortexproject/cortex/pull/4708) for the reference.
The idea behind not doing that if a first [PR](https://github.com/grafana/mimir/pull/11164) with approximate calculations, was not to introduce a mess with some piece of stats passed in the queries stats and some in the Prometheus response, but passing them around twice seems like an even worse idea.

Konstantinov-Innokentii: Small question – what level of compatibility are we aiming to provide? [Above](https://github.com/grafana/mimir/pull/11719#discussion_r2144543534), we agreed not to introduce a feature flag, so perStepStat collection will be controlled solely by the **stats** parameter. Strictly speaking, this makes the querier not fully compatible.

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: I did a pass on making it enum and There is one caveat with this approach that I would like to hear what you think about.

```
type StatsParam string
const (
	// StatsNone means no stats collection
	StatsNone StatsParam = ""
	// StatsAll means collect all stats including PerStepStats
	StatsAll StatsParam = "all"
	// StatsTrue means collect all stats except PerStepStats.
        // Prometheus doesn't specify concrete value, taking everything which is non empty string.
	StatsTrue StatsParam = "true" 
)

// String returns the string representation of StatsParam
func (s StatsParam) String() string {
	return string(s)
}
```
But, Prometheus doesn't specify values for the stat parameter.  If` stats != ""` it will return some stats in the response and if `stats == "all"` it will return some stats + PerStepStats. It leads to some weird code like:
```
func ParseStatsParam(s string) StatsParam {
	switch s {
	case "all":
		return StatsAll
	case "":
		return StatsNone
	default:
		return StatsTrue
	}
}
```
Two things I don't really like with this are:
1. Some extra-behaviour of rewriting the value of stats param
2. If prometheus will ever change this behaviour, let's say will return stats on stats=true, instead of any non-empty value, this code will not work properly.

I don't have a strong point about what is better; neither option is ideal, so I'm happy to proceed with the option you prefer.

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: I kept an eye on the Query results cache hit ratio  when was testing it in dev cell and in ops after the rollout, (AFAIK it was released in ops on 04-24) and I didn't notice any sudden hit rate drops. Is there anything else we can check?
Also, looking on the Query results cache hit ratio  in random prod cells, it seems ok too

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: in a DM we agreed to keep them as separate concepts because we always want prometheus stats for internal billing, but maybe not always want the header stats

pkg/frontend/querymiddleware/results_cache.go
dimitarvdimitrov: hit rate over a week or two is probably good enough if you manage to smooth it over.

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: > in a DM we agreed to keep them as separate concepts because we always want prometheus stats for internal billing, but maybe not always want the header stats

I guess it's related to https://github.com/grafana/mimir/pull/11719#discussion_r2145687264?

pkg/querier/stats_renderer.go
dimitarvdimitrov: ~~Taking them from the Prometheus response seems less surprising. Will this be hard to implement given the groundwork from https://github.com/grafana/mimir/pull/11164?~~ [^1]

[^1]: I think I prefer if we depend less on Prometheus types and maintain our own version of stats.

An alternative would be to strip the stats for internal responses. Here we make a differentiation between requests from the query-frontend/query-scheduler and requests coming directly to the querier. https://github.com/grafana/mimir/blob/699a122aba83da75ccef884d63af468da4f5127e/pkg/mimir/modules.go#L620-L666

But I'm not convinced we should be keeping this compatibility promise.

> what level of compatibility are we aiming to provide?

we're only looking at API-level compatibility - so the querier should feel like a Prometheus server to users/clients.

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: > 1. Some extra-behaviour of rewriting the value of stats param

We're only parsing and validating the value, we're not attaching different meaning to it or adding new values, are we?

> `StatsTrue` ... But, Prometheus doesn't specify values for the stat parameter


Doesn't it only support `all`? I don't see where it returns "some stats" without PerStepStats. Also why do we need the `StatsTrue` value? I was looking at this code https://github.com/grafana/mimir/blob/f2d8b55d3ed30e95a039131d2d27b2208ede674a/vendor/github.com/prometheus/prometheus/web/api/v1/api.go#L542

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: in a DM we agreed to keep them as separate concepts because we always want prometheus stats for internal billing, but maybe not always want the header stats

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: We had a call and agreed that enum here will not help, since prometheus supports not only "all", but any non-empty string.

pkg/querier/stats/stats.pb.go
Konstantinov-Innokentii: We had a chat about it. Passing PerStepStats is in fact adding just adding one more series to the response, so we shouldn't be too concerned about it.

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: As I mentioned above, we had a chat and figured out that it's same as one more series in the response, which is fine. We spent some time monitoring cache performance and it seems to be OK.

pkg/querier/stats/stats.proto
Konstantinov-Innokentii: Works like a charm, thanks!

pkg/querier/stats_renderer.go
dimitarvdimitrov: we agreed to break the goal attempt in the querier because mimir 3.0 will anyways break it and it's not that significant of an API behaviour. In addition to that, the query-frontend hasn't ever supported the query stats and it hasn't been raised as a bug or a concern

pkg/querier/stats/stats.go
```diff
@@ -42,8 +43,13 @@ func IsEnabled(ctx context.Context) bool {
 	return FromContext(ctx) != nil
 }
 
+type QueryStats struct {
```
dimitarvdimitrov: let's call these SafeStats or something similar because they can be used safely concurrently

development/mimir-microservices-mode/docker-compose.jsonnet
```diff
@@ -38,6 +38,9 @@ std.manifestYamlDoc({
 
     // If true, a query-tee instance with a single backend is started.
     enable_query_tee: false,
+
+    // If true, enable per-step stats
+    enable_cached_samples_stats: true,
```
dimitarvdimitrov: these fields here are jsonnet-only config - this doesn't actually change mimir if you don't use it when setting `$.ingester_args` for example.

I don't think we need special jsonnet config, do we? We can just add the CLI flag.

integration/query_frontend_samples_processed_test.go
```diff
@@ -0,0 +1,524 @@
...
+
+	"github.com/grafana/mimir/integration/e2emimir"
+)
+
+func TestQueryFrontendStatsFromResultsCacheShouldBeSameWhenWholeQueryIsCached(t *testing.T) {
+	now := time.Now().Round(time.Second)
+
+	testCases := []struct {
+		name                   string
+		splitQueriesByInterval string
```
dimitarvdimitrov: nit, but can this be time.duration similar to the start and end

integration/query_frontend_samples_processed_test.go
```diff
@@ -0,0 +1,524 @@
...
+			},
+		},
+		{
+			name:                   "split query",
+			splitQueriesByInterval: "10m",
+			query:                  "test_series{}",
+			queryStart:             now.Add(-60 * time.Minute),
+			queryEnd:               now.Add(-20 * time.Minute),
+			setupSeries: func(t *testing.T, writeClient *e2emimir.Client) {
+				pushSeries(t, writeClient, now.Add(-60*time.Minute), now.Add(-10*time.Minute), "test_series")
```
dimitarvdimitrov, line 61: nit to pass start and end as args too, so that they don't have to appear twice in each test case

integration/query_frontend_samples_processed_test.go
```diff
@@ -0,0 +1,524 @@
...
+	e2ecache "github.com/grafana/e2e/cache"
+	e2edb "github.com/grafana/e2e/db"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/integration/e2emimir"
+)
+
+func TestQueryFrontendStatsFromResultsCacheShouldBeSameWhenWholeQueryIsCached(t *testing.T) {
```
dimitarvdimitrov, line 23: let's try to make this a unit test. it takes 1m on my laptop and will slow down CI. It should be easy to add it like a unit test  such as `TestTripperware_Metrics` - see if you can extend this one

pkg/frontend/querymiddleware/codec.go
```diff
@@ -157,6 +160,14 @@ type MetricsQueryRequest interface {
 	WithEstimatedSeriesCountHint(uint64) (MetricsQueryRequest, error)
 	// AddSpanTags writes information about this request to an OpenTracing span
 	AddSpanTags(span trace.Span)
+	// WithStats clones the current request with a a provided value for the stats parameter.
+	// stats parameter controls prometheus query stats collection in the querier.
+	// Note, that these are not returned in the PrometheusResponse, but collected and used for querier.Stats.
+	// Possible values are:
+	// - any non-empty string: stats are collected
```
dimitarvdimitrov: we don't actually respond to any of those other random values, right? they're only responded to by some various implementations of prometheus. But we inject our own `StatsRenderer` and it ignores the params. And for us only `stats="all"` should make a difference. (which again reminds me of the enum, am I going in circles? :D)

this is something i'm fine if we continue discussing after merging this PR and/or something that you can address in a follow-up PR. I mean this only as a way to merge the important changes of this PR before the weekly release next week

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@
   * `cortex_ingest_storage_writer_kafka_request_throttled_seconds`: measures how long Kafka requests have been throttled by the Kafka client.
 * [ENHANCEMENT] Distributor: Add per-user `cortex_distributor_sample_delay_seconds` to track delay of ingested samples with regard to wall clock. #11573
 * [ENHANCEMENT] Distributor: added circuit breaker to not produce Kafka records at all if the context is already canceled / expired. This applied only when experimental ingest storage is enabled. #11768
+* [ENHANCEMENT] Query-frontend: accurate tracking of samples processed from cache #11719
```
dimitarvdimitrov: can you also mention this change in docs/sources/mimir/references/http-api/_index.md

also ok to address in a followup

integration/query_frontend_samples_processed_test.go
```diff
@@ -0,0 +1,524 @@
+//go:build requires_docker
```
dimitarvdimitrov, line 1: all the comments in this file can be addressed in a follow-up PR

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -474,24 +473,30 @@ func newAccumulator(base Extent) (*accumulator, error) {
...
 }
 
-func toExtent(ctx context.Context, req MetricsQueryRequest, res Response, queryTime time.Time, samplesProcessed uint64) (Extent, error) {
+func toExtent(ctx context.Context, req MetricsQueryRequest, res Response, queryTime time.Time, perStepStats []stats.StepStat) (Extent, error) {
 	marshalled, err := types.MarshalAny(res)
 	if err != nil {
 		return Extent{}, err
 	}
+
+	extentPerStepStats := make([]StepStat, len(perStepStats))
```
dimitarvdimitrov, line 482: i'm curious to see if this causes memory spikes every now and then in some special cases

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -757,3 +738,111 @@ func mergeSamplesProcessedPerStep(a, b []StepStat) []StepStat {
...
+
+// heapItem represents an item in the min-heap for merging step stats
+type heapItem struct {
+	value    StepStat
+	arrayIdx int
+	itemIdx  int
+}
+
+// mergeManySamplesProcessedPerStep merges multiple sorted arrays of StepStat using a min-heap.
+func mergeManySamplesProcessedPerStep(arrays ...[]StepStat) []StepStat {
```
dimitarvdimitrov: nit - the type these have is already a `slice` so, `arrays` is a bit like repeating ourselves. let's call these what they are, something like `slicesToMerge` for example

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -304,6 +313,28 @@ func (s *splitAndCacheMiddleware) Do(ctx context.Context, req MetricsQueryReques
...
+			convertedSamplesProcessed[i] = StepStat{
+				Timestamp: stat.Timestamp,
+				Value:     stat.Value,
+			}
+		}
+		perStepStats = append(perStepStats, convertedSamplesProcessed)
+	}
+
+	if details := QueryDetailsFromContext(ctx); details != nil {
+		total := mergeManySamplesProcessedPerStep(perStepStats...)
```
dimitarvdimitrov: nit/stupid question but why do we need to merge them and create a brand new slice when all we want to do is to sum the elements of that slice. Can we sum them instead of adding them to the slices?

pkg/frontend/querymiddleware/results_cache.go
```diff
@@ -757,3 +738,111 @@ func mergeSamplesProcessedPerStep(a, b []StepStat) []StepStat {
...
+				heapifyDown(heap, 0)
+			}
+		}
+	}
+
+	return merged
+}
+
+// buildMinHeap builds a min-heap in-place
+func buildMinHeap(heap []heapItem) {
```
dimitarvdimitrov: let's use the `container/heap` package instead of implementing our own heap.

pkg/frontend/querymiddleware/results_cache_test.go
```diff
@@ -137,14 +135,14 @@ func mkExtentWithEvenPerStepSamplesProcessed(start, end int64, step int64, sampl
 
 func mxEvenlyDistributedExtentPerStepStats(start, end int64, step int64, samplesPerStep int64) []StepStat {
```
dimitarvdimitrov: what does `mx` mean?

pkg/frontend/querymiddleware/results_cache_test.go
```diff
@@ -137,14 +135,14 @@ func mkExtentWithEvenPerStepSamplesProcessed(start, end int64, step int64, sampl
...
 	numSteps := int((end-start)/step) + 1
-	stats := make([]StepStat, numSteps)
+	s := make([]StepStat, numSteps)
```
dimitarvdimitrov: nit: we don't need this change, right?

pkg/frontend/querymiddleware/results_cache_test.go
```diff
@@ -891,7 +944,8 @@ func TestFilterRecentCacheExtents(t *testing.T) {
 			},
 			shouldTruncate: []bool{false, true},
 			// First extent - 1 hour with 1000ms step is a 3601 data points - 36010 samples - not truncated
-			// Second extent - 2 hours with 1000ms step is a 7201 data points - 72010 samples - truncated to 54010 samples
+			// Second extent - 2 hours with 1000ms step is a 7201 data points - 72010 samples.
+			// Truncated to 54010 samples.
```
dimitarvdimitrov: nit but could be useful to say why it's truncated to 54010 exactly

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -143,18 +146,26 @@ func (s *splitAndCacheMiddleware) Do(ctx context.Context, req MetricsQueryReques
 		return nil, apierror.New(apierror.TypeBadData, err.Error())
 	}
 
+	isCacheEnabled := s.cacheEnabled && (s.shouldCacheReq == nil || s.shouldCacheReq(req))
+	maxCacheFreshness := validation.MaxDurationPerTenant(tenantIDs, s.limits.MaxCacheFreshness)
+	maxCacheTime := int64(model.Now().Add(-maxCacheFreshness))
+	cacheUnalignedRequests := validation.AllTrueBooleansPerTenant(tenantIDs, s.limits.ResultsCacheForUnalignedQueryEnabled)
+
+	// Force queier to track PerStepStats, so they could be cached.
+	if s.cacheSamplesProcessedStats {
```
dimitarvdimitrov, line 155: i'm not sure if this shouldn't be called something like `computeSamplesProcessedStats` - it controls whether we compute the stats at all, not only whether we don't cache them. To me cache means that the thing we're caching already exists, we just choose to store it in the cache. 

nitpick i guess

pkg/frontend/querymiddleware/stats.go
```diff
@@ -161,9 +161,9 @@ type QueryDetails struct {
 
 	ResultsCacheMissBytes int
 	ResultsCacheHitBytes  int
-	// SamplesProcessedFromCache represents the total number of samples processed by queriers to produce the result
-	// that has been stored in the query-frontend cache and then fetched to produce the current full query result.
-	SamplesProcessedFromCache uint64
```
dimitarvdimitrov, line 166: there's a test which still references SamplesProcessedFromCache both in the name and in a comment

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -304,6 +313,28 @@ func (s *splitAndCacheMiddleware) Do(ctx context.Context, req MetricsQueryReques
 		}
 	}
 
+	samplesProcessed := queryStats.LoadSamplesProcessedPerStep()
+	if len(samplesProcessed) > 0 {
+		// TODO: reuse stats.StepStat proto in Extent, instead of defining duplicate
```
dimitarvdimitrov: are you still doing this later?

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -304,6 +313,28 @@ func (s *splitAndCacheMiddleware) Do(ctx context.Context, req MetricsQueryReques
 		}
 	}
 
+	samplesProcessed := queryStats.LoadSamplesProcessedPerStep()
```
dimitarvdimitrov: nit: can we tuck these new changes away in a function? `Do` as broad as it sounds, is already doing a lot

pkg/frontend/transport/handler.go
```diff
@@ -151,7 +151,8 @@ func NewHandler(cfg HandlerConfig, roundTripper http.RoundTripper, log log.Logge
 		}, []string{"user"})
 		h.querySamplesProcessedCacheAdjusted = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "cortex_query_samples_processed_cache_adjusted_total",
-			Help: "Number of samples processed to execute a query taking in account the original number of samples processed for parts of the query results looked up from the cache. Note, that this number is approximate.",
+			Help: "Number of samples processed to execute a query taking in account the original number of samples processed for parts of the query results looked up from the cache. " +
+				"Note, that it will be reported proprely only in conjunction with `-query-frontend.cache-samples-processed-stats` enabled. Otherwise, it will be 0.",
```
dimitarvdimitrov: typo

dimitarvdimitrov: also use present tense

pkg/frontend/transport/handler.go
```diff
@@ -329,20 +330,19 @@ func (f *Handler) reportQueryStats(
...
-		samplesProcessedFromCache = details.SamplesProcessedFromCache
+		samplesProcessedCacheAdjusted = details.SamplesProcessedCacheAdjusted
 	}
 	wallTime := stats.LoadWallTime()
 	numSeries := stats.LoadFetchedSeries()
 	numBytes := stats.LoadFetchedChunkBytes()
 	numChunks := stats.LoadFetchedChunks()
 	numIndexBytes := stats.LoadFetchedIndexBytes()
-	sharded := strconv.FormatBool(stats.GetShardedQueries() > 0)
+	sharded := strconv.FormatBool(stats.LoadShardedQueries() > 0)
```
dimitarvdimitrov, line 344: nice catch, but changes like these are sometimes easier to review in a separate PR, especially when it's bundled with a lot of changes like this PR

pkg/querier/stats/stats.go
```diff
@@ -52,194 +58,213 @@ func (s *Stats) AddWallTime(t time.Duration) {
...
-func (s *Stats) LoadSpunOffSubqueries() uint32 {
+func (s *QueryStats) LoadSpunOffSubqueries() uint32 {
 	if s == nil {
 		return 0
 	}
 	return atomic.LoadUint32(&s.SpunOffSubqueries)
 }
 
+func (s *QueryStats) AddSamplesProcessedPerStep(points []StepStat) {
+	s.mx.Lock()
```
dimitarvdimitrov: can we use `atomic.Pointer` here instead of a mutex? That way we don't need two structs and the slice is accessed in the same way as the rest of the fields

dimitarvdimitrov: this works for me 


```go
func TestAtomicPointer(t *testing.T) {
	type testStruct struct {
		field []StepStat
	}

	strct := &testStruct{}
	newStats := []StepStat{{}, {}}
	atomic.StorePointer((*unsafe.Pointer)(unsafe.Pointer(&(strct.field))), unsafe.Pointer(&newStats))

	stats := (*[]StepStat)(atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&(strct.field)))))
	assert.Len(t, *stats, 2)
}
```

pkg/querier/stats/stats_test.go
```diff
@@ -170,6 +175,10 @@ func TestStats_Merge(t *testing.T) {
 		stats2.AddSplitQueries(11)
 		stats2.AddQueueTime(10 * time.Second)
 		stats2.AddSamplesProcessed(20)
+		stats2.AddSamplesProcessedPerStep([]StepStat{
```
dimitarvdimitrov, line 178: in this test case it's not clear what we're testing  - that the stats are deduplicated or that we pick the first or the last or just any of them

pkg/querier/stats/stats.go
```diff
@@ -52,194 +58,213 @@ func (s *Stats) AddWallTime(t time.Duration) {
...
 
-func (s *Stats) LoadSpunOffSubqueries() uint32 {
+func (s *QueryStats) LoadSpunOffSubqueries() uint32 {
 	if s == nil {
 		return 0
 	}
 	return atomic.LoadUint32(&s.SpunOffSubqueries)
 }
 
+func (s *QueryStats) AddSamplesProcessedPerStep(points []StepStat) {
```
dimitarvdimitrov: shouldn't we be merging the points here? i think we share a `stats` instance for multiple queries, so the promql engine can invoke our `StatsRenderer` concurrently (that's why all fields are accessed atomically too i think)

pkg/frontend/transport/handler.go
Konstantinov-Innokentii: Some tests were panicing without it, after I wrapped protogenerated stats into the "QueryStats".

development/mimir-microservices-mode/docker-compose.jsonnet
Konstantinov-Innokentii: But this changes when you run mimir locally in a microservice mode, right? My goal was always to enable the feature in local development.

integration/query_frontend_samples_processed_test.go
Konstantinov-Innokentii: Do you mean all the tests in this file or `TestQueryFrontendStatsFromResultsCacheShouldBeSameWhenWholeQueryIsCached`? These tests are logically connected. One way to speed them up is not to set up the Mimir on every test case run but create it once for Test. Downside is the very confusing test code: metric values are always increasing, so every next test case needs to be aware of that.
 Speaking about TestTripperware_Metrics - afaik it doesn't use cache currently. My thoughts behind making it an integration tests were to cover this functionality as close to real-life as possible, without tons of mocks.

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: That's correct. It affects only the PerStepStats collection and reporting. Even if someone provided "stats=all" it will not return anything. I hate this comment as well :)

pkg/frontend/querymiddleware/results_cache_test.go
Konstantinov-Innokentii: It's a typo, supposed to be "mk"

pkg/frontend/querymiddleware/results_cache_test.go
Konstantinov-Innokentii: Yep, it's leftover of trying to reuse stats from querier and stats variable shadowed the import, I'll fix it, thanks

pkg/frontend/querymiddleware/split_and_cache.go
Konstantinov-Innokentii: Yep, it seems Marco found the solution, but I better do it after I merge this one.

pkg/querier/stats/stats.go
Konstantinov-Innokentii: Makes sense

integration/query_frontend_samples_processed_test.go
Konstantinov-Innokentii: You mean pass seriesStart and seriesEnd as test case arguments? I didn't do it because some tests require several push_series calls:
```
{
			name:                   "series with gap inside of a query range",
			splitQueriesByInterval: 10 * time.Minute,
			query:                  "test_series",
			queryStart:             now.Add(-60 * time.Minute),
			queryEnd:               now.Add(-20 * time.Minute),
			setupSeries: func(t *testing.T, writeClient *e2emimir.Client) {
				pushSeries(t, writeClient, now.Add(-60*time.Minute), now.Add(-50*time.Minute), "test_series")
				pushSeries(t, writeClient, now.Add(-40*time.Minute), now.Add(-20*time.Minute), "test_series")
			},
},
```

CHANGELOG.md
Konstantinov-Innokentii: I'm happy to add it, but I'm unsure what you want me to mention there. From what I see, it's a API doc, but we didn't change user-facing apis?

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: I believe it's not, at least not a big spike. PerStepStats is just one more series in the Response. Converting them to a different type is like allocating an array for another series. Not great, but once I figure out how to import from stats.proto, this code will be removed to avoid duplicating StepStat type.
Is there a better way to address it now?

pkg/frontend/querymiddleware/split_and_cache.go
Konstantinov-Innokentii: It's not stupid at all, in fact it makes a lot of sense, thanks!

pkg/frontend/querymiddleware/results_cache.go
Konstantinov-Innokentii: I figured out that we don't need heap at all.

development/mimir-microservices-mode/docker-compose.jsonnet
dimitarvdimitrov: i dont think so. this is still just a jsonnet field that is not used. see how enable_query_tee is used for example

integration/query_frontend_samples_processed_test.go
dimitarvdimitrov: they can become like this

```go
{
			name:                   "series with gap inside of a query range",
			splitQueriesByInterval: 10 * time.Minute,
			query:                  "test_series",
			queryStart:             now.Add(-60 * time.Minute),
			queryEnd:               now.Add(-20 * time.Minute),
			setupSeries: func(t *testing.T, writeClient *e2emimir.Client, start, end time.Time) {
				pushSeries(t, writeClient, start, start.Add(10*time.Minute), "test_series")
				pushSeries(t, writeClient, start.Add(10 * time.Minute), end, "test_series")
			},
},
```

integration/query_frontend_samples_processed_test.go
dimitarvdimitrov: i mean all tests. only `TestQueryFrontendStatsFromResultsCacheShouldBeSameWhenWholeQueryIsCached` took a minute. I think we can move these to unit tests where they'd take ~second without losing the strength of the assertions

pkg/frontend/querymiddleware/codec.go
dimitarvdimitrov: can you change the comment then :D

pkg/querier/stats/stats.go
Konstantinov-Innokentii: I rewrote it, so now it uses merge, not overwrite. I'm not sure merge will work concurrently safe with atomic.Pointer

CHANGELOG.md
dimitarvdimitrov: oh fair point, this wasn't changed in this PR, but I still think it should be mentioned in the reference docs. especially its interaction with the per step caching flag

pkg/frontend/querymiddleware/results_cache.go
dimitarvdimitrov: oh no worries, i was just saying i'm curious, didn't mean this as a suggestion for a change right now

Konstantinov-Innokentii:
> i left a few nitpicks, but also a few important comments like if we shouldn't be extending instead of overwriting the slice of stats in the querier stats.

Thanks!
> All of the nitpicks you can assume i'm ok to address in a follow-up PR or if it's a convo, we can continue after merging.

I believe I addressed most of them, except the one about integration tests, but we can continue the convo after merging the PR
> Did you have a chance to see the local dev setup instrutions I sent you yesterday? It would be nice to run this locally and see that the right metrics are incremented for sharded, split queries. You can even give it a go in a dev cell to see if it matches napkin math (like approximates the ratio between the stats for cached/uncached bytes in "query stats" logs - results_cache_miss_bytes and results_cache_hit_bytes)

Yep, I ran it locally and tested this. That's why I wrote integration tests, not to keep doing it after every change :) Plan is to test in dev once you are OK with the PR

pkg/querier/stats/stats.go
Konstantinov-Innokentii: Done, thanks

integration/query_frontend_samples_processed_test.go
Konstantinov-Innokentii: Done, thanks

pkg/frontend/querymiddleware/results_cache_test.go
Konstantinov-Innokentii: Added more details, thanks

pkg/frontend/querymiddleware/split_and_cache.go
Konstantinov-Innokentii: I'm not fully happu with name, but let me give my perspective - w anyway computing Sample Processed, so it's more like `computeSamplesProcessedPerStep`. On the other hand, we compute them only to count SamplesProcessed from cache correctly, so that's why I ended up with cacheSamplesProcessedStats - to highlight feature, not the implementation

CHANGELOG.md
Konstantinov-Innokentii: Sorry, but what is the reference docs?

development/mimir-microservices-mode/docker-compose.jsonnet
Konstantinov-Innokentii: Right, so I believe the change in [development/mimir-microservices-mode/config/mimir.yaml](https://github.com/grafana/mimir/pull/11719/files#diff-bd01f80874c5c272fa139e62b87aac503a98af5273c284b0f85ed8f86872bd76) is enough?

pkg/frontend/querymiddleware/codec.go
Konstantinov-Innokentii: Did one more pass

CHANGELOG.md
dimitarvdimitrov: the reference docs are the ones in docs/sources/mimir/references/http-api/_index.md : `docs....references`

pkg/frontend/querymiddleware/split_and_cache.go
dimitarvdimitrov: fair

pkg/querier/stats/stats.go
dimitarvdimitrov: yeah it won't i wrote this comment before realziignwe should merge not overwrite

dimitarvdimitrov (approved):
nice one, I think we can merge it once you take care of the failing CI

Konstantinov-Innokentii:
Testing in mimir-dev-12, please do not merge yet :)

CHANGELOG.md
```suggestion
* [ENHANCEMENT] Query-frontend: Accurate tracking of samples processed from cache. #11719
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1804,6 +1804,10 @@ results_cache:
 # CLI flag: -query-frontend.query-result-response-format
 [query_result_response_format: <string> | default = "protobuf"]
 
+# Cache Statistics of processed samples on results cache.
```
```suggestion
# Cache statistics of processed samples on results cache.
```

PR #10785, opened 2025-03-03
Alertmanager: Strict initialization

santihernandezc:
Like https://github.com/grafana/mimir/pull/10691 but for all tenants.

We can later remove `grafana-alertmanager-conditionally-skip-tenant-suffix` in favor of `alertmanager.grafana-alertmanager-grace-period`, having the same behavior for all tenants.

PR #11720, opened 2025-06-12
Alertmanager: stop using isGrafanaTenant

titolins:
## Context
- `isGrafanaTenant` was added as a helper method for strict AM initialization
- But it relies on a specific flag to be set (`alertmanager.grafana-alertmanager-conditionally-skip-tenant-suffix`)
- If this is not set, the method will return false regardless if it's a Grafana tenant or not
- On https://github.com/grafana/mimir/pull/11609, we started using this in the template validation logic ([here](https://github.com/grafana/mimir/pull/11609/files#diff-8e36fbeb7a03da16fee1b491a126581d93b5dec3b8bd4e4d4398aed38232bbacR678))
- This causes the Grafana AM to error when applying the configuration if the user is using templates with functions which are not supported by Mimir templates
- There's a bool in the `amConfig` (`usingGrafanaConfig`) which indicates whether the tenant is using a Grafana or Mimir config
- Use that instead of the helper method

Note: I was trying to avoid doing clean up in this PR as we're planning to backport it, but since the linter was complaining about it being unused I thought it made sense to remove the flag here as well as to avoid further confusion

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -137,7 +137,7 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet, logger 
 
 	f.BoolVar(&cfg.EnableAPI, "alertmanager.enable-api", true, "Enable the alertmanager config API.")
 	f.BoolVar(&cfg.GrafanaAlertmanagerCompatibilityEnabled, "alertmanager.grafana-alertmanager-compatibility-enabled", false, "Enable routes to support the migration and operation of the Grafana Alertmanager.")
-	f.DurationVar(&cfg.GrafanaAlertmanagerIdleGracePeriod, "alertmanager.grafana-alertmanager-grace-period", defaultGrafanaAlertmanagerGracePeriod, "Duration to wait before shutting down an idle Alertmanager for a tenant that matches grafana-alertmanager-conditionally-skip-tenant-suffix and is using an unpromoted or default configuration.")
+	f.DurationVar(&cfg.GrafanaAlertmanagerIdleGracePeriod, "alertmanager.grafana-alertmanager-grace-period", defaultGrafanaAlertmanagerGracePeriod, "Duration to wait before shutting down an idle Alertmanager using an unpromoted or default configuration when strict initialization is enabled.")
 	f.IntVar(&cfg.MaxConcurrentGetRequestsPerTenant, "alertmanager.max-concurrent-get-requests-per-tenant", 0, "Maximum number of concurrent GET requests allowed per tenant. The zero value (and negative values) result in a limit of GOMAXPROCS or 8, whichever is larger. Status code 503 is served for GET requests that would exceed the concurrency limit.")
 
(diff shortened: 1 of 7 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -138,7 +137,6 @@ func (cfg *MultitenantAlertmanagerConfig) RegisterFlags(f *flag.FlagSet, logger
 
 	f.BoolVar(&cfg.EnableAPI, "alertmanager.enable-api", true, "Enable the alertmanager config API.")
 	f.BoolVar(&cfg.GrafanaAlertmanagerCompatibilityEnabled, "alertmanager.grafana-alertmanager-compatibility-enabled", false, "Enable routes to support the migration and operation of the Grafana Alertmanager.")
-	f.StringVar(&cfg.GrafanaAlertmanagerTenantSuffix, "alertmanager.grafana-alertmanager-conditionally-skip-tenant-suffix", "", "Skip starting the Alertmanager for tenants matching this suffix unless they have a promoted, non-default Grafana Alertmanager configuration or they are receiving requests.")
 	f.DurationVar(&cfg.GrafanaAlertmanagerIdleGracePeriod, "alertmanager.grafana-alertmanager-grace-period", defaultGrafanaAlertmanagerGracePeriod, "Duration to wait before shutting down an idle Alertmanager for a tenant that matches grafana-alertmanager-conditionally-skip-tenant-suffix and is using an unpromoted or default configuration.")
```
santihernandezc: We should update this toggle too (probably should've done it in [my PR](https://github.com/grafana/mimir/pull/10785/files) :sweat_smile:)

```suggestion
	f.DurationVar(&cfg.GrafanaAlertmanagerIdleGracePeriod, "alertmanager.grafana-alertmanager-grace-period", defaultGrafanaAlertmanagerGracePeriod, "Duration to wait before shutting down an idle Alertmanager using an unpromoted or default configuration when strict initialization is enabled.")
```

pkg/alertmanager/multitenant.go
```diff
@@ -822,8 +819,7 @@ func isGrafanaConfigUsable(cfg alertspb.GrafanaAlertConfigDesc) bool {
 // canSkipTenant returns true if the tenant can be skipped, either because strict initialization is enabled,
 // or because the tenant ID matches the provided Grafana tenant suffix.
 func (am *MultitenantAlertmanager) canSkipTenant(userID string) bool {
-	return am.cfg.StrictInitialization ||
-		am.cfg.GrafanaAlertmanagerTenantSuffix != "" && strings.HasSuffix(userID, am.cfg.GrafanaAlertmanagerTenantSuffix)
+	return am.cfg.StrictInitialization
 }
```
santihernandezc: There's no need for this helper.

pkg/alertmanager/multitenant.go
```diff
@@ -768,7 +765,7 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 	}
 
 	// Check if this tenant can be skipped (Grafana suffix matching).
```
santihernandezc:
```suggestion
	// Check if tenants can be skipped (strict initialization enabled).
```

We can also change the variable name to something like `strictInitEnabled`.

titolins: renamed now 👍

PR #11729, opened 2025-06-13
parquet: Scaffolding for parquet bucket store Series()

jesusvazquez:
This PR is a follow up to https://github.com/grafana/mimir/pull/11724 adding the scaffolding for the Series method.

I've mirrored the structure of bucket.go 's Series() but changed some methods for placeholders that we discussed might need to change due to the parquet specific requirements

### DIFF
--- a/pkg/storegateway/parquet_bucket_stores.go
+++ b/pkg/storegateway/parquet_bucket_stores.go
@@ -11,12 +11,11 @@ import (
 	"github.com/grafana/dskit/services"
 	"github.com/oklog/ulid/v2"
 	"github.com/pkg/errors"
+	"github.com/prometheus-community/parquet-common/storage"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/model/labels"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-
-	"github.com/prometheus-community/parquet-common/storage"
 
 	"github.com/grafana/mimir/pkg/storage/sharding"
 	"github.com/grafana/mimir/pkg/storegateway/hintspb"


pkg/storegateway/parquet_bucket_stores.go
```diff
@@ -35,27 +51,259 @@ func NewParquetBucketStores(
...
+	for _, shard := range shards {
+		resHints.AddQueriedBlock(shard.BlockID)
+		shard.MarkQueried()
+	}
+	if err := ss.sendHints(srv, resHints); err != nil {
+		return err
+	}
+
+	streamingSeriesCount := 0
+	if req.StreamingChunksBatchSize > 0 {
```
npazosmendez, line 117: Since we are not going to bother supporting the the non-streamed version, I propose we do this check here:

```go
if req.StreamingChunksBatchSize == 0 {
	return status.Error(codes.Unimplemented, "only streaming is supported")
}
```


And after that proceed knowing we have to stream. That would avoid this big nested block, and also remove the repeated checks in lines 154 and 181. Then we can also discard these methods:

- `createParquetSeriesSetWithChunks`
- `sendSeriesChunks`

pkg/storegateway/parquet_bucket_stores.go
```diff
@@ -35,27 +49,259 @@ func NewParquetBucketStores(
...
+		reqBlockMatchers, err = storepb.MatchersToPromMatchers(reqHints.BlockMatchers...)
+		if err != nil {
+			return status.Error(codes.InvalidArgument, errors.Wrap(err, "translate request hints labels matchers").Error())
+		}
+	}
+
+	logSeriesRequestToSpan(srv.Context(), ss.logger, req.MinTime, req.MaxTime, matchers, reqBlockMatchers, shardSelector, req.StreamingChunksBatchSize)
+
+	shards := ss.openParquetShardsForReading(ctx, req.SkipChunks, req.MinTime, req.MaxTime, reqBlockMatchers, stats)
+	defer ss.closeParquetShards(shards)
```
npazosmendez, line 95: Non actionable comment, just to keep in mind for later: in these functions (or somewhere below them) is where we'll want to, probably, cache file descriptors for local files. Just mentioning it because I see that `closeParquetShards` is already calling Close on the shard openers. We can rework that later depending on which part of the code takes care of caching. cc @francoposa

npazosmendez (approved):
LGTM modulo the linter complains 

About my other comment on streaming: we've talked offline and agreed to cleanup that later, since part of it is still needed for when only Labels are requested.

pkg/storegateway/parquet_bucket_stores.go
francoposa: So how this works is calling "Close" does not actually close the readers it just decrements some sync waitgroups/counters to denote that the reader not being used anymore.

The actual opening and closing of file handles is managed by the "reader pools" which do the lazy opening and then closing after some idle timeout.
Those pools have their own close/shutdown methods which are called when the store-gateway is stopping. Those are what actually release any file handles that haven't been dropped by the idle timeout.

PR #11731, opened 2025-06-13
writes dashboard: add Influx write requests row

alexgreenbank:
#### What this PR does

The new Influx endpoint in the distributor is effectively a sibling of the OTLP endpoint, and so I have made similar additions to those that exist for the OTLP endpoint.

This PR adds a new row to 'Writes' Dashboard showing the stats for metrics arriving by the Influx endpoint. (Based heavily on the existing instrumentation for the OTLP endpoint.) This adds the appropriate routes to the queries and to the SLO based queries.

Also adds and entry for `err-mimir-distributor-max-influx-request-size` to the runbook.

Documentation for the Influx endpoint in the API reference (akin to https://grafana.com/docs/mimir/latest/references/http-api/#otlp) to be added in a separate PR.

It's designed look exactly like the `Gateway - OTLP write requests` row in the current dashboard. The new row looks like this (data is from `cortex-dev-01` as that is the only cell that has traffic (Snythethic Monitoring) to this cell.

image

#### Checklist

- [ ] ~Tests updated.~
- [x] Documentation added.


krajorama (approved):
stamp

PR #11732, opened 2025-06-13
querier remote read: parallel execution of multiple queries

dimitarvdimitrov:
## Why

This change helps match the behavior for non-chunked responses and makes working with the querier easier in downstream projects like GEM. Previously, queries in the streaming chunked response mode were processed sequentially, which was inconsistent with the non-chunked mode.

### Core implementation:

- Modified `remoteReadStreamedXORChunks` to process queries concurrently using goroutines
- Results are collected and streamed back in the original query order
- Error handling maintains fail-fast behavior - if any query fails, the entire request fails

### Test improvements:
- Updated existing tests to support both single and multiple query scenarios
- Restructured `TestRemoteReadHandler_StreamedXORChunks` for easier assertions
  - Introduced `expectedResult` struct combining responses and expected time ranges
  - Each test case now defines queries, expected results, and time ranges in one place

### DIFF
--- a/pkg/querier/remote_read_test.go
+++ b/pkg/querier/remote_read_test.go
@@ -353,11 +353,6 @@ type queryStartEnd struct {
 
 // findQueryIndexByMatchers finds the index of the query that matches the given matchers
 func findQueryIndexByMatchers(queries []*prompb.Query, matchers []*labels.Matcher) int {
-	// If there's only one query, return index 0
-	if len(queries) == 1 {
-		return 0
-	}
-
 	// For multiple queries, match by comparing matchers
 	for i, query := range queries {
 		if matchersEqual(query.Matchers, matchers) {


pkg/querier/remote_read.go
```diff
@@ -147,14 +147,72 @@ func remoteReadStreamedXORChunks(
...
+				return
+			}
+
+			// The streaming API has to provide the series sorted.
+			seriesSet := querier.Select(ctx, true, hints, matchers...)
+			resultCh <- queryResult{idx: i, series: seriesSet, err: nil}
+		}(i, qr)
+	}
+
+	// Collect results and maintain order
```
charleskorn, line 180: Should we have some limits on the number of queries evaluated in parallel, or the amount of data they can return? Otherwise we could be holding a lot of data in memory here and potentially OOM.

pkg/querier/remote_read.go
```diff
@@ -147,14 +147,72 @@ func remoteReadStreamedXORChunks(
...
 			if code/100 != 4 {
-				level.Error(logger).Log("msg", "error while processing remote read request", "err", err)
+				level.Error(logger).Log("msg", "error while streaming remote read response", "err", err)
 			}
 			http.Error(w, err.Error(), code)
-			return
+			break
 		}
 	}
 	w.WriteHeader(http.StatusOK)
```
charleskorn: Is this call effective if we've already written some data to `w` in `streamChunkedReadResponses` above?

Same for the `http.Error` call above.

pkg/querier/remote_read_test.go
```diff
@@ -111,53 +113,107 @@ func (p *partiallyFailingSeriesSet) Warnings() annotations.Annotations {
...
+				{
+					StartTimestampMs: 6,
+					EndTimestampMs:   10,
+					Matchers: []*prompb.LabelMatcher{
+						{Type: prompb.LabelMatcher_EQ, Name: labels.MetricName, Value: "metric2"},
+					},
+				},
+			},
+			expectedQueriedStart: -1, // Skip time range verification for concurrent queries
+			expectedQueriedEnd:   -1,
```
charleskorn: Any reason not to check this for concurrent queries?

pkg/querier/remote_read_test.go
```diff
@@ -208,280 +298,514 @@ func TestRemoteReadHandler_Samples(t *testing.T) {
...
+
+			// Use elements matching to verify the expected time ranges
+			var expectedTimeRanges []queryStartEnd
+			for _, expected := range testData.expectedResults {
+				expectedTimeRanges = append(expectedTimeRanges, queryStartEnd{
+					start: expected.start,
+					end:   expected.end,
 				})
 			}
+			require.ElementsMatch(t, expectedTimeRanges, actualQueryTimeRanges)
```
charleskorn: Why `ElementsMatch` and not `Equal` here? Don't we care about the order?

pkg/querier/remote_read_test.go
```diff
@@ -568,74 +892,123 @@ func TestRemoteReadErrorParsing(t *testing.T) {
...
+			[]model.SamplePair{{Timestamp: 0, Value: 0}, {Timestamp: 1, Value: 1}, {Timestamp: 2, Value: 2}, {Timestamp: 3, Value: 3}},
+			[]mimirpb.Histogram{mimirpb.FromHistogramToHistogramProto(4, test.GenerateTestHistogram(4))},
+		),
+	})
+
 	testCases := map[string]struct {
-		getQuerierErr error
-		seriesSet     storage.SeriesSet
+		getQuerierErr []error             // Changed to slice to support multiple queries
+		seriesSet     []storage.SeriesSet // Changed to slice to support multiple queries
```
charleskorn: These comments don't seem useful.

pkg/querier/remote_read_test.go
dimitarvdimitrov: This is only the time ranges. Since we make the requests in parallel, we can't guarantee which request comes first.

pkg/querier/remote_read.go
dimitarvdimitrov: this was a leftover from another set of changes. Restored to `return` now

pkg/querier/remote_read.go
dimitarvdimitrov: for our use in GEM, we need all queries to run concurrently. I can see where you're coming from. Is it ok if I open a follow-up PR to add concurrency controls?

pkg/querier/remote_read_test.go
charleskorn: Can we change the way this assertion works to make sure the right time range is associated with the right set of matchers?

pkg/querier/remote_read.go
charleskorn: I think I'm missing something sorry: won't calling `WriteHeader` here (or `http.Error` above) be ineffective because `streamChunkedReadResponses` writes to `w` before these calls happen?

pkg/querier/remote_read.go
charleskorn: >Is it ok if I open a follow-up PR to add concurrency controls?

Sounds good to me.

pkg/querier/remote_read.go
dimitarvdimitrov: oh i didn't fully comprehend what you were saying, sorry. yes, you're right. pushed a change. we can't really do much about this, but at least I added comments and put the status code in an earlier place. See https://github.com/grafana/mimir/pull/11732/commits/e9495d66cbef7b787a7fe2fcf0372238eba832e0

dimitarvdimitrov: actually, i removed the status setting altogether, because the tests I had asserted that an error before we start streaming will be reflected in the status code. `http.ResponseWriter.Write()` will set the status code to 200 on the first `Write()` if it's not set already

this is the new commit https://github.com/grafana/mimir/pull/11732/commits/84664c4b9486078dcfae1bc0277e52c123cf835b

pkg/querier/remote_read_test.go
dimitarvdimitrov: pushed this too in https://github.com/grafana/mimir/pull/11732/commits/2af221bbc676aa0c6e334cffa7a616261f0097b3

pkg/querier/remote_read_test.go
```diff
@@ -208,279 +325,566 @@ func TestRemoteReadHandler_Samples(t *testing.T) {
...
+	start int64
+	end   int64
+}
+
+// findQueryIndexByMatchers finds the index of the query that matches the given matchers
+func findQueryIndexByMatchers(queries []*prompb.Query, matchers []*labels.Matcher) int {
+	// If there's only one query, return index 0
+	if len(queries) == 1 {
+		return 0
+	}
```
charleskorn: Shouldn't we still check the matchers are what we expect them to be in this case?

dimitarvdimitrov: yess, fixed in https://github.com/grafana/mimir/pull/11732/commits/53e154cedb2b07920f0dabb807a4ccf4b39af231

pkg/querier/remote_read.go
zenador: Just for my own understanding, is there a reason why we don't use `concurrency.ForEachJob` (from dskit) here? It would make it easier to implement concurrency controls, and cancellation of all goroutines if any fail, plus it's easier to maintain code if we reuse more unified methods wherever possible.

dimitarvdimitrov: It's a really good point @zenador  I think I just copied the approach from the previous presentation, but I'll open a PR to use that. it will be easier to implement concurrency limit too

dimitarvdimitrov: PTAL https://github.com/grafana/mimir/pull/11892

PR #11735, opened 2025-06-13
fix split between Parquet Stores and individual tenant Store

francoposa:
fix split between Parquet Stores and each tenant's Store - those two were accidentially combined in the previous PR.

The `Stores` (one object, plural name) keeps track of each tenant's `Store` and just dispatches to them to satisfy query calls.
The `Stores` also coordinates blocks syncing, keeping track of allowed and disallowed tenants and dispatching sync calls to each tenant's `Store`

The individual `Store` does most of the actual work.

Also broke out a new file to hold the parquet version of `bucketBlock`, now called `parquetBlockWithMeta`. It was named `parquetShardWithMetadata`, but it appears every usage of this code has punted on the idea of having more than 1 shard per block now, so I figured we would try to keep the names as analogous to the main store-gateway types as possible for now.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/parquet_bucket_block.go
+++ b/pkg/storegateway/parquet_bucket_block.go
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/thanos-io/thanos/blob/main/pkg/store/bucket.go
+// Provenance-includes-license: Apache-2.0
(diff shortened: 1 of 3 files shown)


jesusvazquez (approved):
LGTM! Thanks for fixing the inconsistencies!

PR #11741, opened 2025-06-16
MQE: ensure pooled slices not reused

charleskorn:
#### What this PR does

This PR introduces a linter to catch issues like https://github.com/grafana/mimir/pull/11690. 

It does this by changing `LimitingBucketedPool.Put` to return a `nil` slice, and adding a linter to enforce that this return value is used.

This isn't foolproof. For example, the linter doesn't enforce that the return value is stored in the first argument to `Put` (eg. `b = pool.Put(a, memoryConsumptionTracker)` isn't flagged).

#11690

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


charleskorn:
Closing in favour of #11877.

PR #11328, opened 2025-04-28
GitHub actions: set 'persist-credentials: false' for all actions/checkout jobs

pracucci:
#### What this PR does

As part of an effort to harden our GitHub actions workflows, in this PR I'm setting `persist-credentials: false` ([doc](https://github.com/actions/checkout)) for all actions/checkout jobs.

**How it works**:
By default, the auth token is persisted in the local git config. This behaviour enables our CI scripts to run authenticated git commands. The token is removed during post-job cleanup. Setting `persist-credentials: false` we opt-out.

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11761, opened 2025-06-17
Use app credentials when pushing updated build image from GitHub Actions workflow

charleskorn:
#### What this PR does

This PR fixes two issues in the 'build and push build image' workflow:

* it would fail with `fatal: could not read Username for 'https://github.com/': No such device or address` following https://github.com/grafana/mimir/pull/11328, as we were relying on the persisted credentials to push a commit containing the new build image tag
* it would not trigger a CI run after pushing the updated build image tag because it used the default `GITHUB_TOKEN` and so would not trigger another workflow run, as described in the [docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


.github/workflows/push-mimir-build-image.yml
```diff
@@ -109,6 +109,26 @@ jobs:
             echo "isDifferent=true" >> $GITHUB_OUTPUT
           fi
 
+      # This job uses "Mimir bot" instead of "github-actions bot" (secrets.GITHUB_TOKEN)
+      # because any events triggered by the later don't spawn GitHub actions.
```
aknuds1: [Nit]
```suggestion
      # because any events triggered by the latter don't spawn GitHub actions.
```

PR #11767, opened 2025-06-17
Update mimir-prometheus @ 226f974e041bed1c679ebd12145f61216af581c6

Konstantinov-Innokentii:
#### What this PR does

Brings changes form https://github.com/grafana/mimir-prometheus/pull/891

**Note**, that the changes introducing new promql functions -  ts_of_(min|max|last)_over_time. 
It's experimental, and TestQuerySharding_FunctionCorrectness was failing with `ts_of_(min|max|last)_over_time should be tested,` so I added them to `fnToIgnore`.

#### Checklist

- [N/A] Tests updated.
- [N/A ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/querysharding_test.go
+++ b/pkg/frontend/querymiddleware/querysharding_test.go
@@ -1016,6 +1016,7 @@ func TestQuerySharding_FunctionCorrectness(t *testing.T) {
 		{fn: "timestamp"},
 		{fn: "label_replace", args: []string{`"fuzz"`, `"$1"`, `"foo"`, `"b(.*)"`}},
 		{fn: "label_join", args: []string{`"fuzz"`, `","`, `"foo"`, `"bar"`}},
+		{fn: "ts_of_last_over_time", rangeQuery: true, allowedErr: compat.NotSupportedError{}},
 	}
 	testsForFloatsOnly := []queryShardingFunctionCorrectnessTest{
 		{fn: "abs"},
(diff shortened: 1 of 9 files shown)


Konstantinov-Innokentii:
Struggling with TestOurUpstreamTestCasesAreInSyncWithUpstream, trying to figure out why it doesn't respect the 
`# Unsupported by streaming engine` annotation

pkg/frontend/querymiddleware/querysharding_test.go
```diff
@@ -1200,10 +1200,13 @@ func testQueryShardingFunctionCorrectness(t *testing.T, queryable storage.Querya
...
-		"scalar": {},
-		"vector": {},
-		"pi":     {},
+		"time":                 {},
+		"scalar":               {},
+		"vector":               {},
+		"pi":                   {},
+		"ts_of_min_over_time":  {},
+		"ts_of_last_over_time": {},
+		"ts_of_max_over_time":  {},
```
charleskorn: I don't think this is correct. I think we should add these functions to `testsForBoth` in `TestQuerySharding_FunctionCorrectness`.

Konstantinov-Innokentii: Could you please explain why? I mean, if it's experimental, we are not supporting it at all, since MQE is not supporting it?

charleskorn: It would still be supported through Prometheus' engine if experimental functions were enabled.

Konstantinov-Innokentii: But when I add these functions to "testsForBoth", I'm getting `not supported by streaming engine: 'ts_of_min_over_time' function`, am I missing something?

charleskorn: Add `allowedErr: compat.NotSupportedError{}`, like for the other experimental functions.

Konstantinov-Innokentii: Thanks! Also, I fugured out that only `ts_of_last_over_time` should be in `testsForBoth`, while `min/max_over_time `probably should to the `testsForFloatsOnly`, like it's done for last and min/max_over_time. Same happens in the upstream tests. Otherwise, tests are failing. Does it make sense?

PR #11766, opened 2025-06-17
Improve ingest storage Kafka records production observability

pracucci:
#### What this PR does

In this PR I'm doing few changes to improve Kafka records production observability (used by ingest storage):

- Added `cortex_ingest_storage_writer_produce_remaining_deadline_seconds` metric, to measure the remaining deadline (in seconds) when records are requested to be produced.
- Added `cortex_ingest_storage_writer_produce_records_enqueue_duration_seconds` metric, to measure how long it takes to enqueue produced Kafka records in the client.
- Renamed `cortex_ingest_storage_writer_produce_requests_total` to `cortex_ingest_storage_writer_produce_records_enqueued_total` (reminder: `cortex_ingest_storage_writer_produce_records_total` also exists and is tracked by the franz-go client; it's the number of records sent over the wire). This metric has been renamed to clarify it's about records and not requests. The tracked value doesn't change.
- Renamed `cortex_ingest_storage_writer_produce_failures_total` to `cortex_ingest_storage_writer_produce_records_failed_total`. This metric has been renamed to clarify it's about records and not requests. The tracked value doesn't change.

**Dashboard changes previews**:

preview-ruler

preview-produced-records

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11768, opened 2025-06-17
Add circuit breaker to not produce Kafka records at all if the context is already canceled / expired

pracucci:
#### What this PR does

In this PR I'm proposing to add a circuit breaker to not produce Kafka records at all if the context is already canceled / expired. Rationale in the code comment.

This PR is on top of https://github.com/grafana/mimir/pull/11766.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/writer_test.go
+++ b/pkg/storage/ingest/writer_test.go
@@ -326,10 +326,14 @@ func TestWriter_WriteSync(t *testing.T) {
 		// Once the 1st Produce request is received by the server but still processing (there's a 1s sleep),
 		// issue two more requests. One with a short context timeout (expected to expire before the next Produce
 		// request will be sent) and one with no timeout.
-		secondRequestCtx, cancelSecondRequest := context.WithTimeout(ctx, 00*time.Millisecond)
-		t.Cleanup(cancelSecondRequest)
+		var cancelSecondRequest context.CancelFunc
 
 		runAsyncAfter(&wg, firstRequestReceived, func() {
+			var secondRequestCtx context.Context
+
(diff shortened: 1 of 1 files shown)


pkg/storage/ingest/writer_client.go
```diff
@@ -206,6 +206,21 @@ func (c *KafkaProducer) ProduceSync(ctx context.Context, records []*kgo.Record)
...
+	// As a safety mechanism, we want to make sure that the context is not already canceled or its deadline exceeded.
+	// The reason is that once we buffer records to the Kafka client (later in this function), these records will be
+	// sent to the Kafka backend regardless the context is canceled or not. There's no way, in the Kafka client, to
+	// pull out records from a batch buffer. So, if the context is canceled, we circuit break instead of buffering
+	// records that may be sent to the Kafka backend, but that the caller will not know about because context is done.
+	if err := ctx.Err(); err != nil {
+		recordsCount := float64(len(records))
+
+		c.produceRecordsEnqueuedTotal.Add(recordsCount)
+		c.produceRecordsFailedTotal.WithLabelValues(produceErrReason(err)).Add(recordsCount)
```
vlad-diachenko: Should we use something other than `timeout` in this case? It would be useful to have a separate reason here, maybe `cancelled-before-producing`

pracucci: That's a very good point. I like it.

pracucci: Changed in https://github.com/grafana/mimir/pull/11768/commits/c14c713bbcf60687a999c8c468082dd8b867ee1a

PR #11770, opened 2025-06-17
Track cortex_ingest_storage_writer_latency_seconds for failures too

pracucci:
#### What this PR does

During a recent incident, I've learned that `cortex_ingest_storage_writer_latency_seconds` is very misleading if records production times out, because this metric only tracks the latency for successfully produced records. In this PR I'm changing it to track latency both on success and error, but distinguishing between the two cases using a `outcome` label.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * [CHANGE] Querier: Use Mimir Query Engine (MQE) by default. Set `-querier.query-engine=prometheus` to continue using Prometheus' engine. #11501
 * [CHANGE] Memcached: Ignore initial DNS resolution failure, meaning don't depend on Memcached on startup. #11602
 * [CHANGE] Ingester: The `-ingester.stream-chunks-when-using-blocks` CLI flag and `ingester_stream_chunks_when_using_blocks` runtime configuration option have been deprecated and will be removed in a future release. #11711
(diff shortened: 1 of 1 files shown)


vlad-diachenko (approved):

pkg/storage/ingest/writer_test.go
```diff
@@ -95,7 +95,7 @@ func TestWriter_WriteSync(t *testing.T) {
 
 		// Check metrics.
 		assert.NoError(t, promtest.GatherAndCompare(reg, strings.NewReader(fmt.Sprintf(`
-			# HELP cortex_ingest_storage_writer_sent_bytes_total Total number of bytes sent to the ingest storage.
+			# HELP cortex_ingest_storage_writer_sent_bytes_total Total number of bytes produced to the Kafka backend.
```
aknuds1, line 98: Can you verify in these tests that `outcome` is `success`?

aknuds1 (approved):
Just wondering whether we can verify the `outcome` label in tests.

pkg/storage/ingest/writer_test.go
pracucci: It's a bit tricky because the outcome is just on a histogram, that is quite difficult to assert on. There was no assertion on the histogram even before.

PR #11773, opened 2025-06-17
jsonnet: Allow configuring OTLP_TRACES_ENDPOINT and remove $.jaeger_mixin

colega:
#### What this PR does

This replaces the existing `$.jaeger_mixin` in jsonnet with `$.tracing_env_mixin` that would set `OTEL_` environment variables.

Follow up on https://github.com/grafana/mimir/issues/2708

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir/rollout-operator.libsonnet
```diff
@@ -65,7 +65,7 @@
     container.mixin.readinessProbe.httpGet.withPort(8001) +
     container.mixin.readinessProbe.withInitialDelaySeconds(5) +
     container.mixin.readinessProbe.withTimeoutSeconds(1) +
-    $.jaeger_mixin,
+    $.tracing_env_mixin,
```
narqo, line 68: > `rollout_operator_container`

Does rollout-operator support OTLP env also?

narqo (approved):
Overall, works for me 🔥

operations/mimir/rollout-operator.libsonnet
colega: Good point, if it doesn't, I'll fix it now.

dimitarvdimitrov (approved):
can you make the same changes in the helm chart please 🙏 🥺

operations/mimir/rollout-operator.libsonnet
colega: Ok, here we go: https://github.com/grafana/rollout-operator/pull/234

colega: I'll draft this PR until we start using that version of rollout-operator.

colega: Updating rollout-operator in https://github.com/grafana/mimir/pull/11787

colega: That was merged.

colega:
@dimitarvdimitrov I'll do Helm in a separate PR.

colega:
@dimitarvdimitrov turns out Helm doesn't have `JAEGER_` env var set up for tracing, it's agnostic:
```
~/w/github.com/grafana/mimir
$ grep -R -i jaeger operations
operations/helm/charts/mimir-distributed/CHANGELOG.md:* [CHANGE] All: Environment variable `JAEGER_REPORTER_MAX_QUEUE_SIZE` is no longer set. Components will use OTel's default value of `2048` unless explicitly configured. You can still configure `JAEGER_REPORTER_MAX_QUEUE_SIZE` if you configure tracing using Jaeger env vars, and you can always set `OTEL_BSP_MAX_QUEUE_SIZE` OTel configuration. #11700
operations/helm/charts/mimir-distributed/CHANGELOG.md:* [ENHANCEMENT] Add `jaegerReporterMaxQueueSize` Helm value for all components where configuring `JAEGER_REPORTER_MAX_QUEUE_SIZE` makes sense, and override the Jaeger client's default value of 100 for components expected to generate many trace spans. #7068 #7086 #7259
```

Users can configure OTel tracing using the [extraEnv](https://github.com/grafana/mimir/blob/f89b840defefd42ca6bb40222df1b8f43a1e02fc/operations/helm/charts/mimir-distributed/values.yaml#L52-L54) setting and following the `configure-tracing.md` docs. Given OTel already offers standarized env vars that users can most likely just inherit from the other Helm charts they're running, I'd rather not add a solution that renders those standard env vars from custom values in the config.

PR #11774, opened 2025-06-17
hatracker: Add timeouts to per-tenant runtime config

### LINKED ISSUE #4066: Make ha_tracker_failover_timeout configurable per-tenant

The HA-tracker inside distributor will look to change tenant if it hasn't received a message inside `ha_tracker_failover_timeout`, which defaults to 30s.
Suppose a tenant is set up with one target, scraping at 1 minute intervals, then this default does not suit: we should be able to configure the timeout per-tenant. 

Add `ha_tracker_failover_timeout` to runtime config.  Note the existing ones are a bit shorter, e.g. `ha_cluster_label`.

If one tenant has multiple clusters and only a few of them send very slowly, then we might need more fine-grained configuration.

Have it auto-tune - see comment from @pracucci below.

tcard:
#### What this PR does

Adds update and failover timeouts to per-tenant runtime config, so that they different tenants can have specific timeouts that match their specific message frequency.

Fixes #4066 

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/mimir/modules.go
```diff
@@ -424,12 +424,13 @@ func (t *Mimir) initRuntimeConfig() (services.Service, error) {
 	}
 
 	serv, err := NewRuntimeManager(&t.Cfg, "mimir-runtime-config", prometheus.WrapRegistererWithPrefix("cortex_", t.Registerer), util_log.Logger)
-	if err == nil {
-		// TenantLimits just delegates to RuntimeConfig and doesn't have any state or need to do
-		// anything in the start/stopping phase. Thus we can create it as part of runtime config
-		// setup without any service instance of its own.
-		t.TenantLimits = newTenantLimits(serv)
+	if err != nil {
```
tcard, line 427: Unrelated change, but this error was being returned way below at line 453 and thus a bunch of code would be running unnecessarily.

pkg/distributor/ha_tracker.go
```diff
@@ -173,8 +166,31 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
...
+	UpdateTimeout          time.Duration `yaml:"ha_tracker_update_timeout" category:"advanced"`
+	UpdateTimeoutJitterMax time.Duration `yaml:"ha_tracker_update_timeout_jitter_max" category:"advanced"`
+	// We should only failover to accepting samples from a replica
+	// other than the replica written in the KVStore if the difference
+	// between the stored timestamp and the time we received a sample is
+	// more than this duration
+	FailoverTimeout time.Duration `yaml:"ha_tracker_failover_timeout" category:"advanced"`
+}
+
+func (cfg *HATrackerTimeoutsConfig) overrideFromLimits(user string, limits haTrackerLimits) error {
```
colega: This being a pointer receiver that mutates the config made me check the usage to ensure that it's thread safe. I assume the reason to do it this way is because we depend on the `validation` package and can't make `validation` depend on this package to use `HATrackerTimeoutConfig`? Maybe we can move the config there? Or maybe extract it to a different package?

pkg/distributor/ha_tracker.go
```diff
@@ -602,7 +622,34 @@ func (h *defaultHaTracker) checkReplica(ctx context.Context, userID, cluster, re
...
+		defaultHaTracker: h,
+		userID:           userID,
+		cfg: HATrackerTimeoutsConfig{
+			UpdateTimeout:          h.cfg.UpdateTimeout,
+			UpdateTimeoutJitterMax: h.cfg.UpdateTimeoutJitterMax,
+			FailoverTimeout:        h.cfg.FailoverTimeout,
+		},
+		updateTimeoutJitter: h.updateTimeoutJitter,
+	}
+	err := uh.cfg.overrideFromLimits(userID, h.limits)
```
colega: We're setting the `uh.cfg` from `h.cfg`, but I think that `uh.cfg.overrideFromLimits` overwrites all values, regardless whatever is set, isn't that true?

pkg/distributor/ha_tracker.go
tcard: Removed the smelly mutation in [c8b56e7](https://github.com/grafana/mimir/pull/11774/commits/c8b56e7dd347c7fedad8de7ec195dd302a2afb76). Now everything happens in the same place, hopefully more clear now.

Also, I realized `Set...` wasn't actually doing what I wanted (save allocations), so I just made it return the pointers instead.

pkg/distributor/ha_tracker.go
tcard: See c8b56e7dd347c7fedad8de7ec195dd302a2afb76, the behavior should now be more clear.

pkg/util/validation/limits.go
```diff
@@ -106,15 +106,21 @@ func IsLimitError(err error) bool {
...
+	// See distributor.HATrackerTimeoutsConfig.UpdateTimeout
+	HATrackerUpdateTimeout *time.Duration `yaml:"ha_tracker_update_timeout" json:"ha_tracker_update_timeout" category:"advanced" doc:"description=Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp."`
+	// See distributor.HATrackerTimeoutsConfig.UpdateTimeoutJitterMax
+	HATrackerUpdateTimeoutJitterMax *time.Duration `yaml:"ha_tracker_update_timeout_jitter_max" json:"ha_tracker_update_timeout_jitter_max" category:"advanced" doc:"description=Maximum jitter applied to the update timeout, in order to spread the HA heartbeats over time."`
+	// See distributor.HATrackerTimeoutsConfig.FailoverTimeout
+	HATrackerFailoverTimeout                    *time.Duration      `yaml:"ha_tracker_failover_timeout" json:"ha_tracker_failover_timeout" category:"advanced" doc:"description=If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout."`
```
colega: Even though I understand your intention about the pointers here, I would stick to consistency with the rest of the fields where zero value just means "unset". WDYT?

Also, can we use `model.Duration` for consistency?

tcard: Done both at a92525ef4e9b4a4207fb6dc43e6d64b252725701.

pkg/distributor/ha_tracker.go
```diff
@@ -141,50 +138,44 @@ func (r *ReplicaDesc) Clone() memberlist.Mergeable {
...
-	f.DurationVar(&cfg.UpdateTimeout, "distributor.ha-tracker.update-timeout", 15*time.Second, "Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp.")
-	f.DurationVar(&cfg.UpdateTimeoutJitterMax, "distributor.ha-tracker.update-timeout-jitter-max", 5*time.Second, "Maximum jitter applied to the update timeout, in order to spread the HA heartbeats over time.")
-	f.DurationVar(&cfg.FailoverTimeout, "distributor.ha-tracker.failover-timeout", 30*time.Second, "If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout")
 	f.BoolVar(&cfg.EnableElectedReplicaMetric, "distributor.ha-tracker.enable-elected-replica-metric", false, "Enable the elected_replica_status metric, which shows the current elected replica. It is disabled by default due to the possible high cardinality of the metric.")
 
+	// Even though those are deprecated in favor of Limits, we need to keep the flags
```
tcard: @colega Could you let me know if this is clear enough? I realize it's hacky but I don't see how to make things fully backwards-compatible otherwise.

pkg/mimir/mimir_config_test.go
```diff
@@ -262,6 +263,65 @@ custom_client_cluster_validation:
...
+	}, {
+		name: "keeps values in limit",
+		yamlConfig: `
+limits:
+  ha_tracker_update_timeout: 13s
+  ha_tracker_update_timeout_jitter_max: 3s
+  ha_tracker_failover_timeout: 26s
+  `,
+	}, {
+		name: "prefers values in limit",
```
tcard: @colega I realize you suggested failing on a conflict like this, but given CLI flags have defaults, and we can't really tell apart CLI defaults from values from config files, we'll always have a conflict if `distributor.ha_tracker` is set.

I think it isn't worth making things more complicated to handle this case.

pkg/util/validation/limits.go
```diff
@@ -106,15 +109,25 @@ func IsLimitError(err error) bool {
...
+	IngestionBurstSize   int     `yaml:"ingestion_burst_size" json:"ingestion_burst_size"`
+	IngestionBurstFactor float64 `yaml:"ingestion_burst_factor" json:"ingestion_burst_factor" category:"experimental"`
+	AcceptHASamples      bool    `yaml:"accept_ha_samples" json:"accept_ha_samples"`
+	HAClusterLabel       string  `yaml:"ha_cluster_label" json:"ha_cluster_label"`
+	HAReplicaLabel       string  `yaml:"ha_replica_label" json:"ha_replica_label"`
+	HAMaxClusters        int     `yaml:"ha_max_clusters" json:"ha_max_clusters"`
+	// We should only update the timestamp if the difference
+	// between the stored timestamp and the time we received a sample at
+	// is more than this duration.
+	HATrackerUpdateTimeout          model.Duration `yaml:"ha_tracker_update_timeout" json:"ha_tracker_update_timeout" category:"advanced" doc:"flag=distributor.ha-tracker.update-timeout|description=Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp."`
```
tcard: Added a new `doc.flag` parameter to be able to override in docs which CLI flag populates this parameter, since we do it with custom code (`migrateParameter`).

tcard: No longer necessary after [2ab77ef](https://github.com/grafana/mimir/pull/11774/commits/2ab77eff772829566cb16468afbb0d43747561ca).

pkg/mimir/mimir_config_test.go
tcard: After discussing with @colega, we now prefer `distributor.ha_tracker` and log a warning. [0aef3ac](https://github.com/grafana/mimir/pull/11774/commits/0aef3acf489d5ff53095122cb4bef8db46743c5f)

pkg/distributor/ha_tracker.go
tcard: No longer necessary after [2ab77ef](https://github.com/grafana/mimir/pull/11774/commits/2ab77eff772829566cb16468afbb0d43747561ca).

pkg/util/validation/limits.go
```diff
@@ -295,6 +308,12 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
...
 	f.StringVar(&l.HAReplicaLabel, "distributor.ha-tracker.replica", "__replica__", "Prometheus label to look for in samples to identify a Prometheus HA replica.")
+	l.HATrackerUpdateTimeout = model.Duration(15 * time.Second)
+	f.Var(&l.HATrackerUpdateTimeout, "distributor.ha-tracker.update-timeout", "Update the timestamp in the KV store for a given cluster/replica only after this amount of time has passed since the current stored timestamp.")
+	l.HATrackerUpdateTimeoutJitterMax = model.Duration(5 * time.Second)
+	f.Var(&l.HATrackerUpdateTimeoutJitterMax, "distributor.ha-tracker.update-timeout-jitter-max", "Maximum jitter applied to the update timeout, in order to spread the HA heartbeats over time.")
+	l.HATrackerFailoverTimeout = model.Duration(30 * time.Second)
+	f.Var(&l.HATrackerFailoverTimeout, "distributor.ha-tracker.failover-timeout", "If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout")
```
colega:
```suggestion
	f.Var(&l.HATrackerFailoverTimeout, "distributor.ha-tracker.failover-timeout", "If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout.")
```

pkg/distributor/ha_tracker.go
```diff
@@ -435,7 +420,8 @@ const (
 func (h *defaultHaTracker) updateKVLoop(ctx context.Context) {
 	cleanupTick := time.NewTicker(util.DurationWithJitter(cleanupCyclePeriod, cleanupCycleJitterVariance))
 	defer cleanupTick.Stop()
-	tick := time.NewTicker(h.cfg.UpdateTimeout)
+	updateTimeout, _, _ := h.limits.DefaultHATrackerTimeouts()
```
colega: Nit, since we only use one value from this method, and it's the only usage, maybe we should make it like this?
```suggestion
	updateTimeout := h.limits.DefaultHATrackerUpdateTimeout()
```

pkg/distributor/ha_tracker.go
```diff
@@ -173,18 +159,13 @@ func (cfg *HATrackerConfig) RegisterFlags(f *flag.FlagSet) {
...
-	minFailureTimeout := cfg.UpdateTimeout + cfg.UpdateTimeoutJitterMax + time.Second
-	if cfg.FailoverTimeout < minFailureTimeout {
-		return fmt.Errorf(errInvalidFailoverTimeout, cfg.FailoverTimeout, minFailureTimeout)
-	}
-
-	return nil
+// DeprecatedHATrackerTimeoutsConfig is kept for backwards-compatibility in the
+// YAML config files. Values are copied to limits config, and must be accesed
+// through haTrackerLimits.HATrackerTimeouts.
+type DeprecatedHATrackerTimeoutsConfig struct {
```
colega: This should have a comment saying when we should remove it.

```suggestion
// through haTrackerLimits.HATrackerTimeouts.
// TODO: Remove in Mimir 2.18.0
type DeprecatedHATrackerTimeoutsConfig struct {
```

pkg/distributor/ha_tracker.go
tcard: Done at 6171400077.

tcard:
> we need to mention this deprecation in `about-versioning.md`.

@colega Done at [d5d1339](https://github.com/grafana/mimir/pull/11774/commits/d5d13399fc3f4c1e700ce5c423492a214240e0ac).

colega (approved):
Thank you for addressing all my feedback.

PR #11785, opened 2025-06-18
Block-builder-scheduler: Bug: assigned job skipped if start < commit < end

seizethedave:
#### What this PR does

This fixes a bug in block-builder-scheduler where coming out of startup mode, a job may be skipped if the partition's commit offset moves past the job's _start offset_ but before the job's _end offset_. Those offsets between the commit and end offset would be skipped and never consumed by block-builders. The fix is to only ignore the job if it is **wholly** behind the commit offset. (Same bug as #11520 but at assignment time rather than job creation time.)

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
Thank you 🔥

narqo:
I've rebased the branch to resolve a conflict in the changelog

PR #11753, opened 2025-06-16
Skip changelog check for automated PRs

colega:
## Summary
- Skip changelog check for automated PRs with specific labels
- Add appropriate labels to automated workflows

## Changes
- **changelog-check.yml**: Skip validation for PRs with `dependency-update`, `vendored-mimir-prometheus-update`, or `helm-weekly-release` labels
- **helm-weekly-release-pr.yaml**: Add `helm-weekly-release` label to weekly helm release PRs
- **update-vendored-mimir-prometheus.yml**: Add `vendored-mimir-prometheus-update` label to vendored prometheus update PRs  
- **renovate.json5**: Configure Renovate to add `dependency-update` label to all dependency update PRs

PR #11790, opened 2025-06-19
ci: update changelog check workflow to skip backports

narqo:
#### What this PR does

This is a follow-up to https://github.com/grafana/mimir/pull/11753

I've realized that currently, the CI fails on a backporting PR. Such a PR typically, doesn't need a dedicated changelog entry, ie. it's upstream PR either had one, or skipped doing it explicitly.

I think, rather than asking a reviewer of the backporting PR to set the "changelog-not-needed" label, we can just skip these PRs.

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -15,8 +15,9 @@ jobs:
         id: check-label
         run: |
           LABELS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels[].name')
-          
-          if echo "$LABELS" | grep -qE "(changelog-not-needed|dependency-update|vendored-mimir-prometheus-update|helm-weekly-release)"; then
+
+          # "grep -w" to check for the whole words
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog-check.yml
```diff
@@ -15,8 +15,9 @@ jobs:
         id: check-label
         run: |
           LABELS=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '.labels[].name')
-          
-          if echo "$LABELS" | grep -qE "(changelog-not-needed|dependency-update|vendored-mimir-prometheus-update|helm-weekly-release)"; then
+
+          # "grep -w" to check for the whole words
+          if echo "$LABELS" | grep -w -qE "(changelog-not-needed|dependency-update|vendored-mimir-prometheus-update|helm-weekly-release|backport)"; then
```
narqo, line 20: Note to a reviewer: I've added `-w` to the `grep` call so it checked the whole word, and not a substring. E.g. 

- `backport` should match
- `backport rXXX` should NOT match

colega (approved):

PR #11795, opened 2025-06-20
OTLP: Add support for converting OTel scope metadata to metric labels

aknuds1:
#### What this PR does

Add experimental support to OTLP endpoint for converting OTel scope metadata (name, version, schema URL, and attributes) to metric labels, prefixed with `otel_scope_`. Enabled via `-distributor.otel-convert-scope-metadata` flag. #

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


aknuds1:
> while resource attributes are promoted to labels, scope metadata is converted to labels, but what's the difference? Can we use the same language for both things?

@colega I didn't think about that until now, thanks for raising it. I'll think it over and get back to you :)

aknuds1:
Thanks @colega - [made the change in Prometheus](https://github.com/prometheus/prometheus/pull/16760) - just waiting for it to trickle down to Mimir.

aknuds1:
@colega Take another look, please.

pkg/distributor/otel.go
```diff
@@ -484,10 +498,13 @@ func otelMetricsToMetadata(addSuffixes bool, md pmetric.Metrics) []*mimirpb.Metr
 func otelMetricsToTimeseries(
 	ctx context.Context,
 	converter *otlpMimirConverter,
-	addSuffixes, enableCTZeroIngestion, enableStartTimeQuietZero bool,
+	addSuffixes,
+	enableCTZeroIngestion,
+	enableStartTimeQuietZero,
+	keepIdentifyingResourceAttributes,
+	convertHistogramsToNHCB,
+	promoteScopeMetadata bool,
```
colega: Nit: I would consider passing this as a struct (unexported, just defined in this package) because there's a high chance of these bools accidentally changing order or shifting and not being properly covered by a test.

colega (approved):
LGTM but consider my nit please.

pkg/distributor/otel.go
colega: Or maybe just pass the `limits` to that function.

aknuds1: Will take a look tomorrow probably :) Thanks!

aknuds1: Done (passing options).

PR #11796, opened 2025-06-20
Alertmanager: Save the original config string when updating Grafana configuration

alexander-akhmetov:
#### What this PR does

This PR changes the handler that saves Grafana Alertmanager configurtion to preserve the original JSON string during unmarshaling, instead of re-marshaling the parsed Go struct before saving it.

If we do a JSON marshalling round-trip and the Grafana config contains an upstream Alertmanager configuration, secrets, [example](https://github.com/prometheus/alertmanager/blob/0ce3cfb962db3cbb1649d3e816a49a13c4036cd1/config/config.go#L56), in this configuration are lost during marshalling and saved to the storage as redacted strings. 

Instead this PR makes the handler use JSON unmarshalling to validate the configuration, and save the original JSON to the storage.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/alertmanager/api_grafana.go
+++ b/pkg/alertmanager/api_grafana.go
@@ -28,18 +28,19 @@ import (
 )
 
 const (
-	errMalformedGrafanaConfigInStore = "error unmarshalling Grafana configuration from storage"
-	errMarshallingState              = "error marshalling Grafana Alertmanager state"
-	errMarshallingStateJSON          = "error marshalling JSON Grafana Alertmanager state"
-	errMarshallingGrafanaConfigJSON  = "error marshalling JSON Grafana Alertmanager config"
-	errReadingState                  = "unable to read the Grafana Alertmanager state"
-	errDeletingState                 = "unable to delete the Grafana Alertmanager State"
-	errStoringState                  = "unable to store the Grafana Alertmanager state"
-	errReadingGrafanaConfig          = "unable to read the Grafana Alertmanager config"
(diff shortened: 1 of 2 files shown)


pkg/alertmanager/api_grafana.go
```diff
@@ -28,18 +28,19 @@ import (
...
-	errReadingGrafanaConfig          = "unable to read the Grafana Alertmanager config"
-	errDeletingGrafanaConfig         = "unable to delete the Grafana Alertmanager config"
-	errStoringGrafanaConfig          = "unable to store the Grafana Alertmanager config"
-	errBase64DecodeState             = "unable to base64 decode Grafana Alertmanager state"
-	errUnmarshalProtoState           = "unable to unmarshal protobuf for Grafana Alertmanager state"
+	errMalformedGrafanaConfigInStore  = "error unmarshalling Grafana configuration from storage"
+	errMarshallingState               = "error marshalling Grafana Alertmanager state"
+	errMarshallingStateJSON           = "error marshalling JSON Grafana Alertmanager state"
+	errMarshallingGrafanaConfigJSON   = "error marshalling JSON Grafana Alertmanager config"
+	errUnmarshallingGrafanaConfigJSON = "error unmarshalling JSON Grafana Alertmanager config"
```
alexander-akhmetov, line 35: Added `errUnmarshallingGrafanaConfigJSON`

yuri-tceretian (approved):
LGTM

stevesg (approved):

PR #11803, opened 2025-06-21
Add alert for high volume level 1 blocks queried

dimitarvdimitrov:
## Summary
- Add warning alert that fires when level 1 blocks are queried for more than 1 hour
- Indicates potential compactor performance issues when store-gateway serves non-compacted blocks

## Test plan
- Verify alert compiles correctly in mixin
- Monitor for false positives in production environments

### DIFF
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -610,7 +610,7 @@ This alert fires when the store-gateway is querying level 1 blocks for more than
 
 How it **works**:
 
-- Level 1 blocks are deduplicated 2-hour blocks and contain less optimized data compared to higher-level blocks
+- Level 1 blocks are 2-hour blocks shipped by ingesters, not yet optimized by the compactor
 - When the compactor falls behind, store-gateways must serve queries from these less efficient level 1 blocks instead of well-compacted higher-level blocks
 - This can lead to increased query latency and resource usage
 - Out-of-order blocks are excluded from this alert as they may be shipped for previous dates and queried before being compacted into bigger blocks
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -613,6 +613,7 @@ How it **works**:
 - Level 1 blocks are deduplicated 2-hour blocks and contain less optimized data compared to higher-level blocks
 - When the compactor falls behind, store-gateways must serve queries from these less efficient level 1 blocks instead of well-compacted higher-level blocks
 - This can lead to increased query latency and resource usage
+- Out-of-order blocks are excluded from this alert as they may be shipped for previous dates and queried before being compacted into bigger blocks
 
 How to **investigate**:
 
@@ -625,6 +626,8 @@ How to **fix**:
 
 - Scale up the compactor horizontally if it's CPU or memory constrained
 - Check and increase compactor resources if needed
+- Preventatively check store-gateway CPU/memory/disk usage and scale out if constrained
+- Verify store-gateway auto-scaling has sufficient headroom to handle increased load during compactor catch-up
 - If the issue persists, investigate for corrupted blocks that might be blocking compaction
 
 ### MimirCompactorHasNotSuccessfullyRunCompaction
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -606,7 +606,7 @@ How to **investigate**:
 
 ### MimirHighVolumeLevel1BlocksQueried
 
-This alert fires when the store-gateway is querying level 1 blocks for more than 1 hour, indicating that the compactor may not be keeping up with compaction work.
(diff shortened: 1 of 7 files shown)


docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -604,6 +604,29 @@ How to **investigate**:
 - Ensure ingesters are successfully shipping blocks to the storage
 - Look for any error in the compactor logs
 
+### MimirHighVolumeLevel1BlocksQueried
+
+This alert fires when the store-gateway is querying level 1 blocks for more than 6 hours, indicating that the compactor may not be keeping up with compaction work.
+
+How it **works**:
+
+- Level 1 blocks are the result of the first compaction level and contain less optimized data compared to higher-level blocks
```
dimitarvdimitrov:
```suggestion
- Level 1 blocks are deduplicated 2-hour blocks and contain less optimized data compared to higher-level blocks
```

56quarters (approved):
LGTM

operations/mimir-mixin/alerts/blocks.libsonnet
```diff
@@ -235,6 +235,21 @@
             message: '%(product)s bucket index for tenant {{ $labels.user }} in %(alert_aggregation_variables)s has not been updated since {{ $value | humanizeDuration }}.' % $._config,
           },
         },
+        {
+          // Alert if there's sustained querying of level 1 blocks, which indicates the compactor
+          // is not keeping up and the store-gateway is serving blocks that aren't well compacted.
+          alert: $.alertName('HighVolumeLevel1BlocksQueried'),
+          'for': '6h',
+          expr: |||
+            sum(rate(cortex_bucket_store_series_blocks_queried_sum{component="store-gateway",level="1",%s}[%s])) > 0
```
pracucci: Please sum by `%(alert_aggregation_labels)s`

dimitarvdimitrov:
> I backtested this alert at Grafana and it would fire continuously for several cells. I don't think it will be much useful as is.

This is mostly because of out-of-order blocks. I excluded them from the alert. This is the query I tested with and it looks more normal now. It would have fired during a legitimate case of compactor slowdown between Jun 18 and Jun 20 (I couldn't get to the bottom of it though :( )


```
count_over_time(
    (sum by(cluster, out_of_order, namespace) (rate(cortex_bucket_store_series_blocks_queried_sum{component="store-gateway",level="1", out_of_order="false", job=~".*/(store-gateway.*|cortex|mimir|mimir-backend.*)"}[5m])) > 0)
    [6h:1m]
) >= 60 * 6
```

tacole02 (approved):
Thanks for the great docs, @dimitarvdimitrov !

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -604,6 +604,32 @@ How to **investigate**:
 - Ensure ingesters are successfully shipping blocks to the storage
 - Look for any error in the compactor logs
 
+### MimirHighVolumeLevel1BlocksQueried
+
+This alert fires when the store-gateway is querying level 1 blocks for more than 6 hours, indicating that the compactor may not be keeping up with compaction work.
+
+How it **works**:
+
+- Level 1 blocks are deduplicated 2-hour blocks and contain less optimized data compared to higher-level blocks
```
pracucci: What do you mean by "deduplicated" here? They contain duplicated data, that's why they're less optimised, no?

dimitarvdimitrov: i thought level 1 are just 2-hour blocks which have been through compaction once?

but now I realize that this may not mean that they have been deduplicated. Is it that they've gone through the split phase of the compactor only?

@pracucci

pracucci: In Prometheus there's no L0. The first block compacted by ingesters is L1. Every step in the compaction after that has higher level, even the splitting phase because it's just yet another compaction. Querying level=1 blocks means querying block uploaded by ingesters, which I thought was what we wanted here.

dimitarvdimitrov: that's strange because we see level 0 in some blocks 
Screenshot 2025-06-27 at 00 35 45

dimitarvdimitrov: i think you're right. The level zero blocks don't have a source and they're always not out of order. This makes me think that we're only recording a zero-valued block meta and this is resulting in level zero blocks in the metric. I'll look into why this is. In the meantime, I'll fix the runbook.

dimitarvdimitrov: figured it out - see https://github.com/grafana/mimir/pull/11891

edgarasg:
Can You please update https://grafana.com/docs/mimir/latest/manage/mimir-runbooks/ ?

dimitarvdimitrov:
> Can You please update [grafana.com/docs/mimir/latest/manage/mimir-runbooks](https://grafana.com/docs/mimir/latest/manage/mimir-runbooks/) ?

The `latest` tag will be updated with 2.17.0, coming out later this month or next month. You can find the `next` tag version here: https://grafana.com/docs/mimir/next/manage/mimir-runbooks/#mimirhighvolumelevel1blocksqueried

PR #11828, opened 2025-06-23
Skip inclusion check if we're going to no-op anyway in ShardAwareDeduplicateFilter

alexweav:
#### What this PR does

When trying to add a successor, we no-op if the given block is already a successor.
In this path there's no need to do the `isIncludedIn` check. We can assume it's already included by way of it already being a successor.

n/a

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pracucci (commented):
I think this change is logically correct, but I don't know how to evaluate whether this is better or worse. `addSuccessorIfPossible()` is called by `findDuplicates()` with a different block each time, so I don't see how we could enter the `if _, ok := b.successors[other.meta.ULID]` in practice (I've the feeling it's just a safety check).

alexweav:
After a profile this turned out to not affect the hot path too much.

PR #11829, opened 2025-06-24
Fix issue regenerating OTLP files when updating vendored mimir-prometheus

charleskorn:
#### What this PR does

This PR fixes an issue introduced in https://github.com/grafana/mimir/pull/11792 where the OTLP file generation step would fail because `gopatch` is not installed.

#11792

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


dimitarvdimitrov (commented):
thanks

PR #11415, opened 2025-05-07
ruler: add support for remote rule evaluation over HTTP

mbarbon:
We're working on a multi-region Mimir deployment. Our approach is going to be to run two separate Mimir clusters, with a proxy (testing with [promxy](https://github.com/jacksontj/promxy) right now) to provide an unified view.
 
This is working work OK in our initial tests, one of the missing pieces is the ability for `ruler` to query the proxy. This PR adds the option to do this.

#### Checklist

- [x] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.

PR #11833, opened 2025-06-24
ruler: change ruler.RemoteQuerier.client to an http.RoundTripper

mbarbon:
In preparation to adding an option to support remote rule evaluation over HTTP.

#### Checklist

- [x] Tests updated.

### DIFF
--- a/pkg/ruler/grpcroundtripper.go
+++ b/pkg/ruler/grpcroundtripper.go
@@ -1,0 +1,94 @@
+package ruler
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/httpgrpc"
+	"github.com/grafana/dskit/middleware"
+	"github.com/grafana/mimir/pkg/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"google.golang.org/grpc"
+)
+
+// buffer exists to avoid extra copies of body data when adaping httpgrpc.HTTPClient to an http.RoundTripper
+type buffer struct {
+	buff []byte
+	io.ReadCloser
+}
+
+func wrapBuffer(b []byte) io.ReadCloser {
+	return &buffer{buff: b, ReadCloser: io.NopCloser(bytes.NewReader(b))}
+}
+
+func readAll(r io.Reader) ([]byte, error) {
+	if b, ok := r.(*buffer); ok {
+		return b.buff, nil
+	}
+
+	return io.ReadAll(r)
+}
+
+type grpcRoundTripper struct {
+	client httpgrpc.HTTPClient
+}
+
+func newGrpcRoundTripper(client httpgrpc.HTTPClient) *grpcRoundTripper {
+	return &grpcRoundTripper{client: client}
+}
+
+func (g *grpcRoundTripper) RoundTrip(httpReq *http.Request) (*http.Response, error) {
+	body, err := readAll(httpReq.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	grpcRequest := &httpgrpc.HTTPRequest{
+		Method:  httpReq.Method,
+		Url:     httpReq.URL.Path,
+		Body:    body,
+		Headers: httpgrpc.FromHeader(httpReq.Header),
+	}
+
+	grpcResp, err := g.client.Handle(httpReq.Context(), grpcRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	httpResp := &http.Response{
+		StatusCode: int(grpcResp.Code),
+		Body:       wrapBuffer(grpcResp.Body),
+		Header:     make(http.Header),
+	}
+	httpgrpc.ToHeader(grpcResp.Headers, httpResp.Header)
+
+	return httpResp, nil
+}
+
+// dialQueryFrontend creates and initializes a new httpgrpc.HTTPClient taking a QueryFrontendConfig configuration.
+func dialQueryFrontendGRPC(cfg QueryFrontendConfig, prometheusHTTPPrefix string, reg prometheus.Registerer, logger log.Logger) (http.RoundTripper, *url.URL, error) {
+	invalidClusterValidation := util.NewRequestInvalidClusterValidationLabelsTotalCounter(reg, "ruler-query-frontend", util.GRPCProtocol)
+	opts, err := cfg.GRPCClientConfig.DialOption(
+		[]grpc.UnaryClientInterceptor{
+			middleware.ClientUserHeaderInterceptor,
+		},
+		nil,
+		util.NewInvalidClusterValidationReporter(cfg.GRPCClientConfig.ClusterValidation.Label, invalidClusterValidation, logger),
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	opts = append(opts, grpc.WithDefaultServiceConfig(serviceConfig))
+	opts = append(opts, grpc.WithStatsHandler(otelgrpc.NewClientHandler()))
+
+	// nolint:staticcheck // grpc.Dial() has been deprecated; we'll address it before upgrading to gRPC 2.
+	conn, err := grpc.Dial(cfg.Address, opts...)
+	if err != nil {
+		return nil, nil, err
+	}
+	return newGrpcRoundTripper(httpgrpc.NewHTTPClient(conn)), &url.URL{Path: prometheusHTTPPrefix}, nil
+}
--- a/pkg/ruler/remotequerier.go
+++ b/pkg/ruler/remotequerier.go
@@ -22,7 +22,6 @@ import (
 	"github.com/grafana/dskit/grpcclient"
 	"github.com/grafana/dskit/grpcutil"
 	"github.com/grafana/dskit/httpgrpc"
-	"github.com/grafana/dskit/middleware"
 	"github.com/grafana/dskit/user"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -30,13 +29,10 @@ import (
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/storage/remote"
-	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"golang.org/x/time/rate"
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 
 	"github.com/grafana/mimir/pkg/querier/api"
-	"github.com/grafana/mimir/pkg/util"
 	"github.com/grafana/mimir/pkg/util/grpcencoding/s2"
 	"github.com/grafana/mimir/pkg/util/spanlogger"
 	"github.com/grafana/mimir/pkg/util/version"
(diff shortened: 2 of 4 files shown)


pkg/ruler/remotequerier.go
```diff
@@ -200,30 +177,31 @@ func (q *RemoteQuerier) Read(ctx context.Context, query *prompb.Query, sortSerie
...
 	level.Debug(log).Log("msg", "remote read successfully performed", "qs", query)
 
-	var contentType string
-	for _, h := range resp.GetHeaders() {
-		if strings.ToLower(h.GetKey()) == "content-type" {
-			contentType = h.GetValues()[0]
-			break
-		}
-	}
+	contentType := getHeader(resp.Header, "Content-Type")
```
dimitarvdimitrov: we can use `resp.Header.Get("Content-Type")` instead of te heklper function

pkg/ruler/grpcroundtripper.go
```diff
@@ -0,0 +1,94 @@
+package ruler
```
dimitarvdimitrov: can you name this file grpc_roundtripper.go?

dimitarvdimitrov: and the grpcRoundTripper here might be a better fit for the httpgrpcutil package

pkg/ruler/remotequerier.go
```diff
@@ -411,12 +394,9 @@ func getHeader(headers []*httpgrpc.Header, name string) string {
 // injectHTTPGrpcReadConsistencyHeader reads the read consistency level from the ctx and, if defined, injects
 // it as an HTTP header to the list of input headers. This is required to propagate the read consistency
 // through the network when issuing an HTTPgRPC request.
-func injectHTTPGrpcReadConsistencyHeader(ctx context.Context, headers []*httpgrpc.Header) []*httpgrpc.Header {
+func injectHTTPGrpcReadConsistencyHeader(ctx context.Context, headers http.Header) http.Header {
```
dimitarvdimitrov: nit on naming - this is no longer an HTTPGrpc function. It's just HTTP

pkg/ruler/remotequerier.go
```diff
@@ -178,16 +155,16 @@ func (q *RemoteQuerier) Read(ctx context.Context, query *prompb.Query, sortSerie
 		return nil, errors.Wrapf(err, "unable to marshal read request")
 	}
 
-	req := httpgrpc.HTTPRequest{
+	req := http.Request{
```
dimitarvdimitrov: can we use `http.NewRequest` here? I think it does a bit more than what this function does right now

pkg/ruler/grpcroundtripper.go
dimitarvdimitrov: i found out that there's very similar code in pkg/frontend/transport/roundtripper.go. Do you think you can move the code from the frontend to httpgrpcutil and reuse it here? I see the interface there is slightly different - `GrpcRoundTripper`. but we can write a small shim on top which converts a `httpgrpc.HTTPClient` into a `GrpcRoundTripper` and have two constructors for a `grpcRoundTripperAdapter`

pkg/ruler/remotequerier.go
```diff
@@ -152,7 +129,7 @@ func NewRemoteQuerier(
 		timeout:                            timeout,
 		retryLimiter:                       rate.NewLimiter(rate.Limit(maxRetryRate), 1),
 		middlewares:                        middlewares,
-		promHTTPPrefix:                     prometheusHTTPPrefix,
+		promHTTPUrl:                        prometheusHTTPUrl,
```
dimitarvdimitrov: Url -> URL I think. Since it's also HTTP not Http

pkg/ruler/remotequerier.go
dimitarvdimitrov: same for createRequest below

pkg/ruler/grpcroundtripper.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	opts = append(opts, grpc.WithDefaultServiceConfig(serviceConfig))
+	opts = append(opts, grpc.WithStatsHandler(otelgrpc.NewClientHandler()))
+
+	// nolint:staticcheck // grpc.Dial() has been deprecated; we'll address it before upgrading to gRPC 2.
+	conn, err := grpc.Dial(cfg.Address, opts...)
+	if err != nil {
+		return nil, nil, err
+	}
+	return newGrpcRoundTripper(httpgrpc.NewHTTPClient(conn)), &url.URL{Path: prometheusHTTPPrefix}, nil
```
dimitarvdimitrov: here shouldn't we return a URL which also contains the address?

pkg/ruler/grpcroundtripper.go
```diff
@@ -0,0 +1,94 @@
...
+	buff []byte
+	io.ReadCloser
+}
+
+func wrapBuffer(b []byte) io.ReadCloser {
+	return &buffer{buff: b, ReadCloser: io.NopCloser(bytes.NewReader(b))}
+}
+
+func readAll(r io.Reader) ([]byte, error) {
+	if b, ok := r.(*buffer); ok {
```
dimitarvdimitrov: the approach the query-frontned package took was to retun a struct which implements `Bytes()`. I think that'll work a bit better once this is in the httpgrpcutil package

pkg/ruler/remotequerier.go
```diff
@@ -295,31 +278,31 @@ func (q *RemoteQuerier) createRequest(ctx context.Context, query string, ts time
...
-			return httpgrpc.HTTPRequest{}, err
+			return http.Request{}, err
 		}
 	}
 
 	return req, nil
 }
 
-func (q *RemoteQuerier) sendRequest(ctx context.Context, req *httpgrpc.HTTPRequest, logger log.Logger) (*httpgrpc.HTTPResponse, error) {
+func (q *RemoteQuerier) sendRequest(ctx context.Context, req *http.Request, logger log.Logger) (*http.Response, error) {
```
dimitarvdimitrov: let's drop the `ctx` argument and use `req.Context()` and ensure the request we get here has the correct context. I think this also requires to change up a bit the order of functions that call sendRequest - we should first wrap the context in a timeout, then create a new request which the timeout context, then pass the request to sendRequest

pkg/ruler/remotequerier.go
```diff
@@ -266,21 +244,26 @@ func (q *RemoteQuerier) query(ctx context.Context, query string, ts time.Time, l
...
 	}
-	if resp.Code/100 != 2 {
-		return promql.Vector{}, httpgrpc.Errorf(int(resp.Code), "unexpected response status code %d: %s", resp.Code, string(resp.Body))
+	defer resp.Body.Close()
+	body, err := readAll(resp.Body)
+	if err != nil {
+		return promql.Vector{}, httpgrpc.Errorf(int(resp.StatusCode), "error reading response body for status code %d: %s", resp.StatusCode, err)
+	}
+	if resp.StatusCode/100 != 2 {
+		return promql.Vector{}, httpgrpc.Errorf(int(resp.StatusCode), "unexpected response status code %d: %s", resp.StatusCode, string(body))
```
dimitarvdimitrov: unnecessary `int` conversion here and above

dimitarvdimitrov (commented):
thanks for taking this up! I think this will work out. I've got mostly minor comments. The tests aren't compiling also, can you take a look?

pkg/ruler/grpcroundtripper.go
mbarbon: It does not need the address as the GRPC dialer above already has the address.

I chose not to put the address in the URL as it is not used.

pkg/ruler/grpcroundtripper.go
mbarbon: I tried mapping `GrpcRoundTripper` to `httpgrpc.HTTPClient`. 

It did not work well. As far as I can tell `GrpcRoundTripper` is meant to be used for the server side, and makes some assumptions that do not hold for a client. In the end it seemed cleaner to have two implementation sharing some code.

Let me know what you think.

mbarbon:
Thanks for the review! I updated the MR with your suggestions.

pkg/ruler/grpcroundtripper.go
dimitarvdimitrov: what about when using http in #11415

pkg/ruler/remotequerier.go
```diff
@@ -178,52 +157,52 @@ func (q *RemoteQuerier) Read(ctx context.Context, query *prompb.Query, sortSerie
...
-	level.Debug(log).Log("msg", "remote read successfully performed", "qs", query)
 
-	var contentType string
-	for _, h := range resp.GetHeaders() {
-		if strings.ToLower(h.GetKey()) == "content-type" {
-			contentType = h.GetValues()[0]
-			break
-		}
+	if resp.StatusCode/100 != 2 {
+		return nil, httpgrpc.Errorf(int(resp.StatusCode), "unexpected response status code %d: %s", resp.StatusCode, string(body))
```
dimitarvdimitrov: There's still an unnecessary integer conversion here and above

pkg/ruler/grpcroundtripper.go
dimitarvdimitrov: can't we do something like this 


```diff
diff --git a/pkg/util/httpgrpcutil/grpc_roundtripper.go b/pkg/util/httpgrpcutil/grpc_roundtripper.go
index d39ec2aff9..8c800e78f5 100644
--- a/pkg/util/httpgrpcutil/grpc_roundtripper.go
+++ b/pkg/util/httpgrpcutil/grpc_roundtripper.go
@@ -30,7 +30,9 @@ func AdaptGrpcRoundTripperToHTTPRoundTripper(r GrpcRoundTripper) http.RoundTripp
 }
 
 func AdaptHTTPGrpcClientToHTTPRoundTripper(c httpgrpc.HTTPClient) http.RoundTripper {
-	return &httpGrpcClientAdapter{client: c}
+	return &grpcRoundTripperAdapter{
+		roundTripper: &httpgrpcClientAdapter{client: c},
+	}
 }
 
 // This adapter wraps GrpcRoundTripper and converted it into http.RoundTripper
@@ -38,16 +40,20 @@ type grpcRoundTripperAdapter struct {
 	roundTripper GrpcRoundTripper
 }
 
-// This adapter wraps httpgrp.HTTPClient and converted it into GrpcRoundTripper
-type httpGrpcClientAdapter struct {
-	client httpgrpc.HTTPClient
-}
-
 type buffer struct {
 	buff []byte
 	io.ReadCloser
 }
 
+type httpgrpcClientAdapter struct {
+	client httpgrpc.HTTPClient
+}
+
+func (h *httpgrpcClientAdapter) RoundTripGRPC(ctx context.Context, request *httpgrpc.HTTPRequest) (*httpgrpc.HTTPResponse, io.ReadCloser, error) {
+	resp, err := h.client.Handle(ctx, request)
+	return resp, nil, err
+}
+
 func newBuffer(b []byte) io.ReadCloser {
 	return &buffer{buff: b, ReadCloser: io.NopCloser(bytes.NewReader(b))}
 }
@@ -62,6 +68,12 @@ func (a *grpcRoundTripperAdapter) RoundTrip(r *http.Request) (*http.Response, er
 		return nil, err
 	}
 
+	// for client requests, use the request Path as GRPC Url, as RequestURI is not set
+	if r.URL == nil || r.URL.Path == "" {
+		return nil, errors.New("error mapping HTTP to GRPC (missing URL or URL.Path)")
+	}
+	req.Url = r.URL.Path
+
 	resp, body, err := a.roundTripper.RoundTripGRPC(r.Context(), req)
 	if err != nil {
 		var ok bool
@@ -84,41 +96,12 @@ func (a *grpcRoundTripperAdapter) RoundTrip(r *http.Request) (*http.Response, er
 	}
 	httpgrpc.ToHeader(resp.Headers, httpResp.Header)
 
-	setContentLenght(resp, httpResp)
-
-	return httpResp, nil
-}
-
-func (a *httpGrpcClientAdapter) RoundTrip(r *http.Request) (*http.Response, error) {
-	req, err := httpgrpc.FromHTTPRequest(r)
-	if err != nil {
-		return nil, err
-	}
-
-	// for client requests, use the request Path as GRPC Url, as RequestURI is not set
-	if r.URL == nil || r.URL.Path == "" {
-		return nil, errors.New("Error mapping HTTP to GRPC (missing URL or URL.Path)")
-	}
-	req.Url = r.URL.Path
-
-	resp, err := a.client.Handle(r.Context(), req)
-	if err != nil {
-		return nil, err
-	}
-
-	httpResp := &http.Response{
-		StatusCode: int(resp.Code),
-		Body:       newBuffer(resp.Body),
-		Header:     http.Header{},
-	}
-	httpgrpc.ToHeader(resp.Headers, httpResp.Header)
-
-	setContentLenght(resp, httpResp)
+	setContentLength(resp, httpResp)
 
 	return httpResp, nil
 }
 
-func setContentLenght(resp *httpgrpc.HTTPResponse, httpResp *http.Response) {
+func setContentLength(resp *httpgrpc.HTTPResponse, httpResp *http.Response) {
 	contentLength := -1
 	if len(resp.Body) > 0 {
 		contentLength = len(resp.Body)

```

pkg/ruler/remotequerier.go
```diff
@@ -96,39 +94,20 @@ func (c *QueryFrontendConfig) Validate() error {
...
-	opts = append(opts, grpc.WithStatsHandler(otelgrpc.NewClientHandler()))
-
-	// nolint:staticcheck // grpc.Dial() has been deprecated; we'll address it before upgrading to gRPC 2.
-	conn, err := grpc.Dial(cfg.Address, opts...)
-	if err != nil {
-		return nil, err
-	}
-	return httpgrpc.NewHTTPClient(conn), nil
+func DialQueryFrontend(cfg QueryFrontendConfig, prometheusHTTPPrefix string, reg prometheus.Registerer, logger log.Logger) (http.RoundTripper, *url.URL, error) {
+	return dialQueryFrontendGRPC(cfg, prometheusHTTPPrefix, reg, logger)
```
dimitarvdimitrov, line 98: nit: do we need a whole function in a new file for this? can't we restore the old version with the modifications you made?

pkg/ruler/remotequerier.go
```diff
@@ -386,37 +377,29 @@ func (q *RemoteQuerier) sendRequest(ctx context.Context, req *httpgrpc.HTTPReque
...
+	req.Header.Set(user.OrgIDHeaderName, orgID)
 	return nil
 }
 
-func getHeader(headers []*httpgrpc.Header, name string) string {
-	for _, h := range headers {
-		if h.Key == name && len(h.Values) > 0 {
-			return h.Values[0]
-		}
+func getHeader(headers http.Header, name string) string {
```
dimitarvdimitrov: i think we can replace calls to this function with just `headers.Get(name)`

pkg/ruler/remotequerier.go
mbarbon: In the follow-up MR that adds HTTP support, it's going to have a separate branch for the HTTP case. which is going to call the function defined [here](https://github.com/grafana/mimir/pull/11415/files#diff-e301b12cae358e9e4c37b6770e27da4792ff2f999ef687f6fd3bbb1a2bfe4359).

Do you prefer the code for both GRPC and HTTP in `remotequerier.go`?

pkg/ruler/grpcroundtripper.go
mbarbon: In that case it will be set up by a different function and it will have the address.

pkg/ruler/grpcroundtripper.go
mbarbon: Done

pkg/ruler/remotequerier.go
dimitarvdimitrov: oh, sure, let's have a separate function like you've done it

dimitarvdimitrov (approved):
LGTM, sorry for dropping the ball on this

dimitarvdimitrov (approved):

mbarbon:
> LGTM, sorry for dropping the ball on this

Thanks for seeing it through!

PR #11848, opened 2025-06-24
Parameterize metric/label name validation scheme

### LINKED ISSUE #11503: Implement classical name validation mode without disabling through global variable
Implement the current, classical, metric and label name validation mode without disabling through the `prometheus/common` global variable, as is done currently.

See @ywwg's [comment below](https://github.com/grafana/mimir/issues/11503#issuecomment-2959360706) for full context.

juliusmh:
#### What this PR does

Direct and indirect references to the global name validation scheme were removed in favor of a per-tenant override.

- Distributors have a validation middleware that ensures metric and label names are valid.
- Rulers validate rule(group)s using this naming scheme.
- Queriers use UTF8 validation *everywhere*.

TODO:

- [x] Reach consensus on whether `streamingpromqlcompat.NameValidatingEngine` is the right approach: Decided to do UTF8 validation in query path.
- [x] Get Alertmanager updated with fix, stop using our own fork

Depends on
- https://github.com/grafana/mimir-prometheus/pull/946
- https://github.com/grafana/prometheus-alertmanager/pull/116

Fixes: #11503

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/mimirtool/rules/rules.go
```diff
@@ -255,7 +256,8 @@ func (r RuleNamespace) Validate(groupNodes []rulefmt.RuleGroupNode) []error {
 func ValidateRuleGroup(g rwrulefmt.RuleGroup, node rulefmt.RuleGroupNode) []error {
 	var errs []error
 	for i, r := range g.Rules {
-		for _, err := range r.Validate(node.Rules[i]) {
+		// TODO(juliusmh):
+		for _, err := range r.Validate(node.Rules[i], validation.LegacyNamingScheme) {
```
juliusmh: Q: How should we handle validation checks in mimirtool?

aknuds1: I can think about it after the initial review :D

pkg/cardinality/request.go
```diff
@@ -145,27 +146,38 @@ func (r *LabelValuesRequest) String() string {
 
 // DecodeLabelValuesRequest decodes the input http.Request into a LabelValuesRequest.
 // The input http.Request can either be a GET or POST with URL-encoded parameters.
-func DecodeLabelValuesRequest(r *http.Request, tenantMaxLimit int) (*LabelValuesRequest, error) {
+func DecodeLabelValuesRequest(
+	r *http.Request,
+	tenantMaxLimit int,
+	tenantNameValidationScheme validation.NamingScheme,
```
aknuds1: I think `nameValidationScheme` makes more sense, since the configuration parameter doesn't inherently pertain to the tenant. It _can_ be overridden per tenant, but doesn't have to be. Also, `tenantNameValidationScheme` sounds as if we are validating the tenant name, which is confusing :)
```suggestion
	nameValidationScheme validation.NamingScheme,
```

pkg/cardinality/request.go
```diff
@@ -145,27 +146,38 @@ func (r *LabelValuesRequest) String() string {
...
+		r.Form,
+		tenantMaxLimit,
+		tenantNameValidationScheme,
+	)
 }
 
 // DecodeLabelValuesRequestFromValues is like DecodeLabelValuesRequest but takes url.Values in input.
 func DecodeLabelValuesRequestFromValues(values url.Values) (*LabelValuesRequest, error) {
-	return DecodeLabelValuesRequestFromValuesWithTenantMaxLimit(values, 0)
+	return DecodeLabelValuesRequestFromValuesWithTenantLimits(values, 0, validation.LegacyNamingScheme)
```
aknuds1: Why do we hard code the naming scheme here, can't it be passed in?

pkg/cardinality/request_test.go
```diff
@@ -98,7 +99,7 @@ func TestDecodeLabelValuesRequest(t *testing.T) {
 		req, err := http.NewRequest("GET", "http://localhost?"+params.Encode(), nil)
 		require.NoError(t, err)
 
-		actual, err := DecodeLabelValuesRequest(req, 0)
+		actual, err := DecodeLabelValuesRequest(req, 0, validation.LegacyNamingScheme)
```
aknuds1: I think that on the query side we can use `validation.UTF8NamingScheme`, because there's no need to enforce name validity rules (except for them being valid UTF-8) when querying. For example, what if the tenant has enabled the UTF-8 validation scheme in the past, and then flipped the switch to the legacy validation scheme? Why should we prevent them for querying for their older metrics?

pkg/distributor/validate_test.go
```diff
@@ -73,13 +79,24 @@ func TestValidateLabels(t *testing.T) {
...
+		namingScheme:               prom_validation.LegacyNamingScheme,
+	}
+
+	utf8Config := validateLabelsCfg{
+		maxLabelNamesPerSeries:     3,
+		maxLabelNamesPerInfoSeries: 4,
+		maxLabelNameLength:         24,
+		maxLabelValueLength:        24,
+		namingScheme:               prom_validation.UTF8NamingScheme,
+	}
```
aknuds1: I think it's better to ensure that `utf8Config` doesn't diverge more from `legacyConfig` than necessary:
```suggestion
	utf8Config := legacyConfig
	utf8Config.namingScheme = prom_validation.UTF8NamingScheme
```

pkg/mimir/modules.go
```diff
@@ -827,6 +827,8 @@ func (t *Mimir) initQueryFrontendTripperware() (serv services.Service, err error
 		panic(fmt.Sprintf("invalid config not caught by validation: unknown PromQL engine '%s'", t.Cfg.Querier.QueryEngine))
 	}
 
+	eng = streamingpromqlcompat.NameValidatingEngine(eng, t.Overrides)
```
aknuds1: Just wondering currently why we need to wrap the engine, can't e.g. a parameter be passed to `promql.NewEngine`?

pkg/querier/querier.go
```diff
@@ -197,6 +197,7 @@ func New(cfg Config, limits *validation.Overrides, distributor Distributor, quer
 		panic(fmt.Sprintf("invalid config not caught by validation: unknown PromQL engine '%s'", cfg.QueryEngine))
 	}
 
+	eng = compat.NameValidatingEngine(eng, limits)
```
aknuds1: Also wondering as above why this wrapping is necessary, but haven't looked deeply into it yet.

pkg/streamingpromql/compat/name_validating_engine.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/grafana/mimir/pkg/util/validation"
+)
+
+type nameValidatingEngine struct {
+	engine promql.QueryEngine
+	limits *validation.Overrides
+}
+
+// NameValidatingEngine creates a new promql.QueryEngine that wraps engine and overrides query options
+// with the tenants naming scheme from limits.
```
aknuds1: I just think the following is clearer, since the scheme isn't necessarily overridden for the tenant, it just _can_ be.

```suggestion
// with the name validation scheme from limits.
```

pkg/streamingpromql/compat/name_validating_engine.go
```diff
@@ -0,0 +1,88 @@
...
+// Returns legacy validation scheme if at least one tenant uses legacy validation.
+func (e nameValidatingEngine) getNamingScheme(ctx context.Context) (prom_validation.NamingScheme, error) {
+	tenantIDs, err := tenant.TenantIDs(ctx)
+	if err != nil {
+		return "", err
+	}
+	for _, tenantID := range tenantIDs {
+		if e.limits.NameValidationScheme(tenantID) == prom_validation.LegacyNamingScheme {
+			return prom_validation.LegacyNamingScheme, nil
+		}
```
aknuds1: Isn't it bad to enforce the legacy name validation scheme on the query path? All it will do is prevent the user from querying any metrics they ingested in the past with the UTF-8 name validation scheme?

pkg/streamingpromql/operators/aggregations/count_values_test.go
```diff
@@ -222,7 +223,16 @@ func TestCountValues_GroupLabelling(t *testing.T) {
...
-			aggregator := NewCountValues(inner, labelName, types.NewInstantQueryTimeRange(timestamp.Time(0)), testCase.grouping, testCase.without, memoryConsumptionTracker, posrange.PositionRange{})
+			aggregator := NewCountValues(
+				inner,
+				labelName,
+				types.NewInstantQueryTimeRange(timestamp.Time(0)),
+				testCase.grouping,
+				testCase.without,
+				memoryConsumptionTracker,
+				posrange.PositionRange{},
+				validation.LegacyNamingScheme,
```
aknuds1: Is there any point in not using the UTF-8 naming scheme instead?

pkg/util/validation/limits.go
```diff
@@ -1371,6 +1372,13 @@ func (o *Overrides) SubquerySpinOffEnabled(userID string) bool {
 	return o.getOverridesForUser(userID).SubquerySpinOffEnabled
 }
 
+// NameValidationScheme returns the validation scheme to use for a particular tenant.
+// Must be one of validation.LegacyNamingScheme or validation.UTF8NamingScheme.
```
aknuds1: Should we even document possible values for `validation.UTF8NamingScheme`? Those are defined in the `validation` package aren't they?
```suggestion
```

aknuds1 (commented):
OK, gave it a quick review. Nice work! Please check my questions :)

pkg/mimir/modules.go
juliusmh: I decided to go with a wrapper instead, mainly because it means less changes to the query path and it's easier to implement. This also drove the decision to add name validation scheme to `promql.QueryOpts` rather than `promql.Engine`.

At the moment queriers/frontends use a single engine to process all incoming queries. In Mimir, we use one engine in query frontends, another one is used by queriers. The alternative is to pass a `QueryEngineFactory` instead of a `QueryEngine` and either:

- create a new engine for each request _or_
- create legacy- and utf8-engines and choose the appropriate one

The question is: do we want validation scheme per-engine or per-query?

pkg/streamingpromql/compat/name_validating_engine.go
juliusmh: This PR tries to maintain the current behavior, and currently, we validate label names in `label_join`, `label_replace` and `count_values` with legacy scheme. Queries that don't use those functions support UTF8, today. This can be seen in the [tests](https://github.com/grafana/mimir/pull/11855) I did. 

The question we have to answer: is there any scenario in which data returned by a querier or frontend is fed back to mimir without validation (e.g. to ingesters directly, or with label validation disabled)? If the answer is no, we can safely remove the validation. I'm assuming the answer is yes though, based on [this comment](https://github.com/grafana/mimir/issues/11503#issuecomment-2965189971).

pkg/cardinality/request_test.go
juliusmh: Mainly because at the moment, Mimir does validate this. However, you're correct that we will not be able to access those metrics after switching from utf8 to legacy. I altered package cardinality and querier to use utf8 validation.

pkg/cardinality/request.go
```diff
@@ -262,10 +263,10 @@ func extractLabelNames(values url.Values) ([]model.LabelName, error) {
 
 	labelNames := make([]model.LabelName, 0, len(labelNamesParams))
 	for _, labelNameParam := range labelNamesParams {
-		labelName := model.LabelName(labelNameParam)
-		if !labelName.IsValid() {
+		if !validation.UTF8NamingScheme.IsValidLabelName(labelNameParam) {
```
charleskorn: Should the naming scheme used here be configurable?

pkg/mimir/modules.go
charleskorn: >The question is: do we want validation scheme per-engine or per-query?

If the goal is to make the scheme configurable per tenant, then making it per-query makes sense to me. Otherwise we'd need an engine instance per configuration scheme.



> I decided to go with a wrapper instead, mainly because it means less changes to the query path and it's easier to implement. This also drove the decision to add name validation scheme to `promql.QueryOpts` rather than `promql.Engine`.

Could you elaborate? It seems like adding it to `promql.QueryOpts` would be a more natural fit rather than creating a wrapper like this (at least for Mimir, not sure about Prometheus).

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -18,12 +19,17 @@ import (
 
 type InstantVectorFunctionOperatorFactory func(
 	args []types.Operator,
-	memoryConsumptionTracker *limiter.MemoryConsumptionTracker,
-	annotations *annotations.Annotations,
-	expressionPosition posrange.PositionRange,
-	timeRange types.QueryTimeRange,
+	params InstantVectorFunctionOperatorParams,
```
charleskorn: I would make this a pointer to avoid the potential of copying a large struct every time the method is called:

```suggestion
	params *InstantVectorFunctionOperatorParams,
```

pkg/mimir/modules.go
aknuds1: @charleskorn For context, I believe that we should make the metric/label name validation scheme configurable _per-tenant_. The actual configurability, IMO, should be done in a follow-up PR though, so we can focus this PR on just dropping the global variable usage.

I also think that for the most part, queries shouldn't enforce the legacy naming scheme. My reasoning is that if the user at one point has ingested metrics with the less restrictive UTF-8 naming scheme, and then flips the switch back to the "legacy" naming scheme, why should the query engine fail queries against the metrics ingested with the UTF-8 naming scheme? @juliusmh has made me aware that, on the other hand, there are certain parts of the query engine (label replace?) that need to use the same validation scheme as the ingest path. WDYT @charleskorn?

pkg/streamingpromql/compat/name_validating_engine.go
aknuds1: @charleskorn WDYT?

pkg/streamingpromql/operators/aggregations/count_values_test.go
aknuds1: In retrospect, maybe we can assume `validation.LegacyNamingScheme` in this PR, and switch to `validation.UTF8NamingScheme` in the PR that enables user configuration of the naming scheme.

pkg/cardinality/request.go
aknuds1: I am wondering in this case whether this code should use a parameter (instead of hard coding), but not let it be user configurable since, as I explain above, I'm not sure we want to enforce the legacy naming scheme in queries. WDYT @charleskorn?

pkg/streamingpromql/operators/aggregations/count_values_test.go
juliusmh: Are we talking about this unit test or the `count_values` function in general? At runtime, MQE will use the injected/per-tenant scheme (this is just hardcoded in the test)

aknuds1: I mean the hard coding in the test. When we support `validation.UTF8NamingScheme`, I suspect that would be a more natural default in tests.

juliusmh: I switched to utf8 and added two new test cases demonstrating this works.

pkg/mimir/modules.go
charleskorn: > @juliusmh has made me aware that, on the other hand, there are certain parts of the query engine (label replace?) that need to use the same validation scheme as the ingest path.

This is correct. `label_replace` and `label_join` will need to use the same validation scheme. 

> I also think that for the most part, queries shouldn't enforce the legacy naming scheme. My reasoning is that if the user at one point has ingested metrics with the less restrictive UTF-8 naming scheme, and then flips the switch back to the "legacy" naming scheme, why should the query engine fail queries against the metrics ingested with the UTF-8 naming scheme

I'm not aware of anywhere apart from `label_replace` and `label_join` that will enforce label name validation, so I don't believe this is an issue - is there something I've missed?

pkg/streamingpromql/compat/name_validating_engine.go
charleskorn: I believe that data will always be validated when written back by rulers. They don't skip validation.

However, I believe we should validate names in at least `label_replace` and `label_join`, as not validating may lead to unexpected results:

* functions that consume the results of `label_replace` and `label_join` may have undefined behaviour with invalid label names
* a user testing a recording or alerting rule as a query may not realise that their rule will later fail to write samples when evaluated by rulers

pkg/cardinality/request.go
charleskorn: I don't know enough about this code path to know what makes sense here. Does Prometheus apply validation to the label names here as well?

pkg/mimir/modules.go
aknuds1: @charleskorn I just want to make sure we don't introduce configurable metric/label name validation in the query engine except for where we need to (`label_replace` and `label_join` as I understand it). At most we should verify that names are valid UTF-8 (apart from the noted exceptions).

pkg/cardinality/request.go
aknuds1: > Does Prometheus apply validation to the label names here as well?

@juliusmh Did you check?

In any case, this is an example of read path isn't it, where we don't want to enforce the "legacy" naming scheme? If so, I think it could a non-configurable parameter/argument, just to avoid hard coding.

juliusmh: Afaik, the `api/v1/cardinality/*` endpoints are Mimir specific. 

The closest is probably db stats in prometheus admin API (calculated [here](https://github.com/grafana/mimir-prometheus/blob/6693a913b24cca33b98a2d98f1f954d8415a4583/tsdb/index/postings.go#L198)), where you can get some analysis on label name/value cardinality, which, as far as I can see, doesn't involve any validation. But I'm not super confident in that analysis.

pkg/mimir/modules.go
juliusmh: What about `count_values`, it returns a series with the aggregation label?

pkg/querier/cardinality_analysis_handler_test.go
```diff
@@ -705,10 +705,6 @@ func TestLabelValuesCardinalityHandler_ParseError(t *testing.T) {
 				url:                  "/label_values",
 				expectedErrorMessage: "'label_names[]' param is required",
 			},
-			"label_names param is invalid": {
-				url:                  "/label_values?label_names[]=olá",
-				expectedErrorMessage: "invalid 'label_names' param 'olá'",
-			},
```
aknuds1: The test case just needs an adjustment, to a string which isn't valid UTF-8:
```suggestion
			"label_names param is invalid": {
				url:                  "/label_values?label_names[]=\xff\xfe",
				expectedErrorMessage: "invalid 'label_names' param '\xff\xfe'",
			},
```

aknuds1 (changes requested):
I think this is starting to look good. We should keep that test case though.

pkg/mimir/modules.go
charleskorn: Yep, good catch, you're correct.

pkg/streamingpromql/planning.go
```diff
@@ -441,7 +441,7 @@ func findFunction(name string) (functions.Function, bool) {
 // Materialize converts a query plan into an executable query.
 func (e *Engine) Materialize(ctx context.Context, plan *planning.QueryPlan, queryable storage.Queryable, opts promql.QueryOpts) (promql.Query, error) {
 	if opts == nil {
-		opts = promql.NewPrometheusQueryOpts(false, 0)
+		opts = promql.NewPrometheusQueryOpts(false, 0, model.UTF8Validation)
```
juliusmh: > @juliusmh Should q.nameValidationScheme be used here?

What is `q` ? At this point we don't have a query yet. We could give the engine `e` a configurable default. FYI, prometheus uses global `model.NameValidationScheme` here.

pkg/streamingpromql/query.go
```diff
@@ -59,7 +61,7 @@ type Query struct {
 
 func (e *Engine) newQuery(ctx context.Context, queryable storage.Queryable, opts promql.QueryOpts, timeRange types.QueryTimeRange, originalExpression string) (*Query, error) {
 	if opts == nil {
-		opts = promql.NewPrometheusQueryOpts(false, 0)
+		opts = promql.NewPrometheusQueryOpts(false, 0, model.UTF8Validation)
```
juliusmh: > @juliusmh Should e.nameValidationScheme be used here?

We can add some default to `Engine`. Using `UTF8Validation` here matches prometheus behavior (which uses the global). In practice, it doesn't matter much because `opts` is never `nil`.

juliusmh:
As discussed @aknuds1 :
- Synced with https://github.com/grafana/mimir/pull/12109
- Fixed failing tests

@charleskorn :
> Could you elaborate? It seems like adding it to promql.QueryOpts would be a more natural fit rather than creating a wrapper like this (at least for Mimir, not sure about Prometheus).

For promql, the main problem is that query opts are directly constructed and passed to query engine in prometheus' [v1.NewApi](https://github.com/grafana/mimir/blob/be35f7d5d9b8a0345ad7a16eb938928523b92e3d/vendor/github.com/prometheus/prometheus/web/api/v1/api.go#L600). Unless we want to rewrite that handler we need to somehow "modify" `promql.QueryOptions`, IMO the `QueryEngine` interface is the only viable place. 

You're right, for MQE this can be implemented differently, see [this commit](https://github.com/grafana/mimir/pull/11848/commits/35af54397548c7e62378913da04d579e0215a256). The wrapper, is/was a way of unifying this logic for promql/mimir query engine paths.

pkg/streamingpromql/planning.go
aknuds1: Ah you are right @juliusmh, I didn't notice that `q` is made later.

pkg/streamingpromql/query.go
aknuds1: @juliusmh Reading this again, `validationScheme` is defined on line 77. Could it be defined before line 63 and be used in the call to `NewPrometheusQueryOpts`? Just wondering whether it makes sense.

dimitarvdimitrov:
copying some feedback form slack on doing validation in the query engine:



> a user testing a recording or alerting rule as a query may not realise that their rule will later fail to write samples when evaluated by rulers
> […]
> or switches to legacy

so let’s say a user has had UTF8 for a bit, but for some reason (?) we want to turn it back off. Are we ok also breaking their query evaluations that work on the UTF8-ingested metrics?

Now their dashbaords will be broken and some alerts may not work until the blocker for enabling UTF8 is enabled. This sounds more intrusive and inconvenient than surprising them with failing rule evals. It makes something which was supposed to be async suddenly a top priority.

Always having UTF8 validation in the query engine looks safer and I agree with Bryan that it’s also simpler.

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -208,9 +208,10 @@ func TestCardinalityQueryRequestValidationRoundTripper(t *testing.T) {
 			expectedErrType: apierror.TypeBadData,
 		},
 		{
-			// non-utf8 label name will be rejected even when we transition to UTF-8 label names
+			// TODO: Check
+			// non-legacy label name will be accepted
```
aknuds1: I don't think this test should change. Should it?
```suggestion
			// non-utf8 label name will be rejected even when we transition to UTF-8 label names

```

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -208,9 +208,10 @@ func TestCardinalityQueryRequestValidationRoundTripper(t *testing.T) {
...
 			url:             cardinalityLabelValuesPathSuffix + "?label_names[]=\\xbd\\xb2\\x3d\\xbc\\x20\\xe2\\x8c\\x98",
-			expectedErrType: apierror.TypeBadData,
+			expectedErrType: "",
```
aknuds1, line 232:
```suggestion
			expectedErrType: apierror.TypeBadData,
```

aknuds1 (dismissed):
There's a test change that I don't think should be made.

pkg/frontend/querymiddleware/request_validation_test.go
juliusmh: I think the test was wrong due to escape backslashes. I changed the case to test non-utf8 label name as the comment suggests. Also added a test case here ensuring regular utf8 names pass.

aknuds1: Thanks @juliusmh

go.mod
dimitarvdimitrov, line 1: i assume you'd want to merge the two PRs linked in the description before merging _this_ PR. Don't forget to do that :)

pkg/distributor/distributor.go
```diff
@@ -1099,15 +1093,22 @@ func (d *Distributor) prePushRelabelMiddleware(next PushFunc) PushFunc {
...
 
+		validationScheme := d.limits.ValidationScheme(userID)
+
 		var removeTsIndexes []int
 		lb := labels.NewBuilder(labels.EmptyLabels())
 		for tsIdx := 0; tsIdx < len(req.Timeseries); tsIdx++ {
 			ts := req.Timeseries[tsIdx]
 
-			if mrc := d.limits.MetricRelabelConfigs(userID); len(mrc) > 0 {
+			if mrcs := d.limits.MetricRelabelConfigs(userID); len(mrcs) > 0 {
```
dimitarvdimitrov: nit to name `mrcs` something intelligible like `relabelConfigs`

pkg/distributor/distributor.go
```diff
@@ -1099,15 +1093,22 @@ func (d *Distributor) prePushRelabelMiddleware(next PushFunc) PushFunc {
...
+				for _, mrc := range mrcs {
```
dimitarvdimitrov: this currently runs for every incoming sample; can we make a change somewhere where the relabel configs are loaded so we set the validation scheme once instead of on every ingested sample for every relabel config?

pkg/distributor/validate_test.go
```diff
@@ -236,10 +262,34 @@ func TestValidateLabels(t *testing.T) {
...
+		},
+		{
+			metric:                   map[model.LabelName]model.LabelValue{model.MetricNameLabel: "foo", "name😀": "value", "team": "b"},
+			skipLabelNameValidation:  false,
+			skipLabelCountValidation: false,
+			config:                   utf8Config,
+		},
 	} {
-		err := validateLabels(s, cfg, userID, "custom label", mimirpb.FromMetricsToLabelAdapters(c.metric), c.skipLabelNameValidation, c.skipLabelCountValidation, cast, ts)
+		err := validateLabels(s, c.config, userID, "custom label", mimirpb.FromMetricsToLabelAdapters(c.metric), c.skipLabelNameValidation, c.skipLabelCountValidation, cast, ts)
```
dimitarvdimitrov: i'm tempted to add more test cases for utf8 config. I think a lot of the existing cases would work in both validation schemes. Maybe have a slice of schemes to run each test cases in? or have two expected errors - one for legacy validation scheme and one for utf8

pkg/distributor/validate_test.go
```diff
@@ -249,7 +299,7 @@ func TestValidateLabels(t *testing.T) {
 	require.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(`
 			# HELP cortex_discarded_samples_total The total number of samples that were discarded.
 			# TYPE cortex_discarded_samples_total counter
-			cortex_discarded_samples_total{group="custom label",reason="label_invalid",user="testUser"} 1
+			cortex_discarded_samples_total{group="custom label",reason="label_invalid",user="testUser"} 2
```
dimitarvdimitrov: nitpick to create this count out of the test cases instead of hardcoding it here. e.g. counting the number of expected errors. Would be amazing if you can do the same below; not sure how hard it would be

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -209,9 +209,13 @@ func TestCardinalityQueryRequestValidationRoundTripper(t *testing.T) {
 		},
 		{
 			// non-utf8 label name will be rejected even when we transition to UTF-8 label names
-			url:             cardinalityLabelValuesPathSuffix + "?label_names[]=\\xbd\\xb2\\x3d\\xbc\\x20\\xe2\\x8c\\x98",
+			url:             cardinalityLabelValuesPathSuffix + "?label_names[]=\xbd\xb2\x3d\xbc\x20\xe2\x8c\x98",
```
dimitarvdimitrov, line 227: why was the previous url still rejected even though it was ascii-only? was it because of the slashes? let's keep it as a test case in addition to the real UTF8 case you added/fixed 🤷

dimitarvdimitrov, line 227: also can you make the smae changes to `TestMetricsQueryRequestValidationRoundTripper`  and `TestLabelsQueryRequestValidationRoundTripper`? maybe also in `TestPrometheusCodec_DecodeEncodeMultipleTimes_Labels`

pkg/mimir/promexts.go
dimitarvdimitrov, line 1: is there anything which depended on legacy validation in this package? The closes i could find is the runtime config, but i couldn't see how exactly

pkg/mimirtool/commands/analyse_rulefiles.go
```diff
@@ -22,8 +22,8 @@ type RuleFileAnalyzeCommand struct {
 }
 
 func (cmd *RuleFileAnalyzeCommand) run(_ *kingpin.ParseContext) error {
-
-	output, err := AnalyzeRuleFiles(cmd.RuleFilesList)
+	// TODO: Get scheme from CLI flag.
```
dimitarvdimitrov, line 25: there's still this TODO outstanding. Can't we just always use utf8? what's the benefit of restricting the validation scheme in this mimirtool command

pkg/mimirtool/commands/rules.go
dimitarvdimitrov, line 1: let's of TODOs here too. I think some of the commands make sense to have a validation scheme - `prepare`, `lint`, `check` - because they're supposed to help you catch rules which the backend won't like before you send them to the backend. But for the rest like `load`, `get`, `diff`, `sync`, `print` can we use always use utf8?

it would also be nice to add some tests for this cli flag and the parameters you already added to some functions

pkg/querier/cardinality_analysis_handler_test.go
```diff
@@ -706,8 +706,8 @@ func TestLabelValuesCardinalityHandler_ParseError(t *testing.T) {
 				expectedErrorMessage: "'label_names[]' param is required",
 			},
 			"label_names param is invalid": {
-				url:                  "/label_values?label_names[]=olá",
```
dimitarvdimitrov, line 709: did `olá` become valid with your changes?

dimitarvdimitrov (commented):
i can't find any blockers. Maybe it would have been easier to get this through if you took it package by package

dimitarvdimitrov:
oh, also the mimirtool changes i think need a changelog entry because they're a change in behaviour

aknuds1:
> oh, also the mimirtool changes i think need a changelog entry because they're a change in behaviour

@dimitarvdimitrov Oh, let me look that over.

pkg/mimirtool/commands/analyse_rulefiles.go
aknuds1: @dimitarvdimitrov The idea is to leave the configurability (CLI flags etc) to the follow-up PR. The TODOs are for the next PR :)

go.mod
aknuds1: The Alertmanager PR is merged, only the mimir-prometheus PR remains. We will merge that first :)

pkg/frontend/querymiddleware/request_validation_test.go
aknuds1: @dimitarvdimitrov The previous URL was simply invalid according to the "legacy" Prometheus name validation scheme. The backslashes should be enough for the label name to be invalid.

Aren't there enough cases for the legacy scheme?

pkg/mimir/promexts.go
aknuds1: @dimitarvdimitrov I believe I just configured `model.LegacyValidation` in every package `init` function that I found to be necessary, when upgrading the Prometheus version where `model.NameValidationScheme` was introduced. To ensure that we didn't suddenly change the name validation scheme in Mimir.

pkg/mimirtool/commands/rules.go
aknuds1: @dimitarvdimitrov There actually isn't any `mimirtool` flag yet. We are postponing that until the follow-up PR, the TODOs are for resolving in said PR. Make sense?

The reason is that this PR is supposed to be refactoring only, and not change behaviour.

pkg/querier/cardinality_analysis_handler_test.go
aknuds1: Yes, it does become valid, because the query path validation scheme becomes UTF-8 instead of legacy. This makes me realize that the latter is actually a behavioral change (although something we want I think, from our discussions).

aknuds1:
> oh, also the mimirtool changes i think need a changelog entry because they're a change in behaviour

@dimitarvdimitrov I think there may have been a misunderstanding here? We are postponing `mimirtool` (and other) behavioral changes until a follow-up PR.

pkg/frontend/querymiddleware/request_validation_test.go
aknuds1: @juliusmh Could you look into Dimitar's above request?

pkg/distributor/validate_test.go
aknuds1: @juliusmh Could you address the nitpick?

pkg/distributor/validate_test.go
aknuds1: @juliusmh WDYT?

pkg/distributor/distributor.go
aknuds1: @juliusmh WDYT?

pkg/distributor/distributor.go
aknuds1: @juliusmh Can you address the nitpick?

pkg/frontend/querymiddleware/request_validation_test.go
dimitarvdimitrov: i don't see any test cases for invalid strings

pkg/mimir/promexts.go
dimitarvdimitrov: why was it necessary here? we only turned it off, but didn't parametarize it. so if something depended on this init function it's now got the validation flipped

pkg/frontend/querymiddleware/request_validation_test.go
aknuds1: On second look, you are right @dimitarvdimitrov. Could you add a case @juliusmh?

pkg/mimirtool/commands/rules.go
dimitarvdimitrov: aha i see. it's confusing that the tests run with utf8 validation, but the main code runs with legacy validation

dimitarvdimitrov:
> > oh, also the mimirtool changes i think need a changelog entry because they're a change in behaviour
> 
> @dimitarvdimitrov I think there may have been a misunderstanding here? We are postponing `mimirtool` (and other) behavioral changes until a follow-up PR.

you're right 👍

pkg/mimirtool/commands/rules.go
aknuds1: Yeah, we will be introducing support for UTF-8 validation in a follow-up.

pkg/mimir/promexts.go
aknuds1: I believe I just mechanically figured out where the `prometheus/common` package was imported, and where I had to force `LegacyValidation` in `init` functions. I didn't analyze what would actually depend on it. I just did it from a robustness perspective.

aknuds1: When working on this PR, we figured out what depends on `model.NameValidationScheme` by deleting that global, and working our way backwards.

pkg/distributor/distributor.go
juliusmh: Was thinking about this when implementing. I would place this check in util/limits.validate, but the validation scheme doesn't have a yaml parameter (yet). 

I moved it up so it's not run for each series (but still for each sample). Do you have any other suggestion? @aknuds1

pkg/distributor/validate_test.go
juliusmh: I refactored the whole test to run each case for legacy and utf8 and compute the desired metrics using the test cases. This wasn't hard, but lengthy.

pkg/frontend/querymiddleware/request_validation_test.go
juliusmh: I don't really understand this comment. We have cases for
Invalid names: `\xbd\xb2\x3d` , valid legacy/utf8 name: `foo` and valid utf8 name: `some.label`

juliusmh: `LabelsQueryRequestValidationRoundTripper` and the multiple encode test both represent `labelMatchers` as string. Should I add validation, if yes, like this?

```go
for _, matcher := range labelMatcherSets {
  if _, err := parser.ParseMetricSelector(matcher); err != nil {
    return nil, apierror.New(apierror.TypeBadData, fmt.Sprintf("invalid label matcher: %s", matcher))
  }
}
```

juliusmh:
Given this was a bit chaotic, quick summary of what was addressed:
- refactored distributor's validation tests
- added more tests to `MetricsQueryRequestValidationRoundTripper`


and what is still open:
- [Here](https://github.com/grafana/mimir/pull/11848#discussion_r2225265876), I don't see which case is missing.
- [Here](https://github.com/grafana/mimir/pull/11848#discussion_r2225277677), should I add labelMatcher/selector validation to `LabelsQueryRequestValidationRoundTripper`? We didn't do that previously.
- [Here](https://github.com/grafana/mimir/pull/11848#discussion_r2225237252), Move this relabel config modification to a better place. Do you have any ideas where that could be?
- I added some Changelog messages, but needs refining

pkg/distributor/distributor.go
dimitarvdimitrov: can't you include this logic in `limits.MetricRelabelConfigs`? it already knows the right validation scheme

```
		validationScheme := d.limits.ValidationScheme(userID)
		relabelConfigs := d.limits.MetricRelabelConfigs(userID)
```

pkg/frontend/querymiddleware/request_validation_test.go
dimitarvdimitrov: you're right

dimitarvdimitrov: oh, i'm surprised we don't do any validation on this request. do we delegate it to prometheus?

pkg/mimir/promexts.go
dimitarvdimitrov: but like you can't just make this change and hope it works 😄 i checked the package and i couldn't find anything which either directly or transitively needs validation

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Grafana Mimir
 
+* [CHANGE] Query-frontend: Add support for utf8 label/metric names in `/api/v1/cardinality/{label_values|label_values|active_series}` endpoints. #11848.
+* [CHANGE] Querier: Add support for utf8 metric names, support utf8 label names in `label_join`, `label_replace` and `count_values`. #11848.
```
```suggestion
* [CHANGE] Querier: Add support for utf8 metric/label names in `label_join`, `label_replace` and `count_values` PromQL functions. #11848.
```


idk, maybe it's not clear to everyone what this means

dimitarvdimitrov:
thanks @juliusmh . i think only the last thread still needs resolving. i left a suggestion

pkg/frontend/querymiddleware/request_validation_test.go
juliusmh: As far as I can see, the request is not validated in `querymiddleware`. It looks like prometheus' `web/api/v1.API` already does validation on the query name, for label values, label names and series endpoints.

pkg/distributor/distributor.go
juliusmh: Did that, please take a look at https://github.com/grafana/mimir/pull/11848/commits/61faf174f73d6a7d5fcf79f7a144402c5c416a22

pkg/frontend/querymiddleware/request_validation_test.go
aknuds1: Ah yes, I wonder if I was looking at an outdated revision.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Grafana Mimir
 
+* [CHANGE] Query-frontend: Add support for utf8 label/metric names in `/api/v1/cardinality/{label_values|label_values|active_series}` endpoints. #11848.
+* [CHANGE] Querier: Add support for utf8 metric names, support utf8 label names in `label_join`, `label_replace` and `count_values` PromQL functions. #11848.
```
```suggestion
* [CHANGE] Query-frontend: Add support for UTF-8 metric and label names in `/api/v1/cardinality/{label_values|label_values|active_series}` endpoints. #11848.
* [CHANGE] Querier: Add support for UTF-8 metric and label names in `label_join`, `label_replace`, and `count_values` PromQL functions. #11848.
```

tacole02 (approved):

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -70,6 +70,21 @@ func TestMetricsQueryRequestValidationRoundTripper(t *testing.T) {
 			url:             instantQueryPathSuffix + "?query=up&start=123&end=456&step=60s",
 			expectedErrType: "",
 		},
+		{
+			// accepts utf label names
```
aknuds1: Better to be specific:
```suggestion
			// accepts utf-8 label names
```

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -70,6 +70,21 @@ func TestMetricsQueryRequestValidationRoundTripper(t *testing.T) {
...
+			url:             instantQueryPathSuffix + `?query=up{"test.label"="test"}`,
+			expectedErrType: "",
+		},
+		{
+			// accepts utf metric name
```
aknuds1: Better to be specific:
```suggestion
			// accepts utf-8 metric name
```

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -70,6 +70,21 @@ func TestMetricsQueryRequestValidationRoundTripper(t *testing.T) {
...
+			url:             instantQueryPathSuffix + `?query={"test.label"}`,
+			expectedErrType: "",
+		},
+		{
+			// invalid utf-string
```
aknuds1: Better to be specific:
```suggestion
			// invalid utf-8 string
```

pkg/frontend/querymiddleware/request_validation_test.go
```diff
@@ -132,7 +147,7 @@ func TestLabelsQueryRequestValidationRoundTripper(t *testing.T) {
 			expectedErrType: "",
 		},
 		{
-			url:             labelNamesPathSuffix + "?match=up{}&start=123&end=456&limit=100",
+			url:             labelNamesPathSuffix + "?match[]=up{}&start=123&end=456&limit=100",
```
aknuds1: Why do you make this change? Accidental?

juliusmh: Thanks, it's unrelated! The query param is called `match[]`.

dimitarvdimitrov (approved):

go.mod
```diff
@@ -343,7 +343,7 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v1.8.2-0.20250721161615-ca0aa6ff0922
+replace github.com/prometheus/prometheus => github.com/grafana/mimir-prometheus v1.8.2-0.20250724124959-0d9a87207337
```
aknuds1: Please update to mimir-prometheus@main

```console
go mod edit -replace github.com/prometheus/prometheus=github.com/grafana/mimir-prometheus@main && go mod tidy && go mod vendor
```

PR #11855, opened 2025-06-25
Integration test: distributor/querier/ruler legacy name validation

juliusmh:
#### What this PR does

Add integration test(s) that ensure distributor/ruler/querier use legacy name validation.

There are a couple of interesting things to see here:
- ruler doesn't validate `RuleGroup.Labels`, only `RuleGroup.Rules.Labels`
- ruler doesn't validate `Rule.Expr`
- MQE and PromQL engine don't validate label matchers

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/integration/legacy_name_validation_test.go
+++ b/integration/legacy_name_validation_test.go
@@ -1,0 +1,388 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/cortexproject/cortex/blob/master/integration/alertmanager_test.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Cortex Authors.
+//go:build requires_docker
+
+package integration
+
+import (
+	"net/http"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/grafana/e2e"
+	e2edb "github.com/grafana/e2e/db"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/rulefmt"
+	"github.com/prometheus/prometheus/prompb"
+	promRW2 "github.com/prometheus/prometheus/prompb/io/prometheus/write/v2"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/integration/e2emimir"
+)
+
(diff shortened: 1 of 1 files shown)


integration/legacy_name_validation_test.go
```diff
@@ -0,0 +1,388 @@
...
+	"github.com/prometheus/prometheus/prompb"
+	promRW2 "github.com/prometheus/prometheus/prompb/io/prometheus/write/v2"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/integration/e2emimir"
+)
+
+// TestLegacyNameValidation_DistributorWrite ensures that distributors discard invalid
+// metric and label names in the write path.
+func TestLegacyNameValidation_DistributorWrite(t *testing.T) {
```
dimitarvdimitrov, line 28: this looks like a copied version of `testDistributorCases`. Did you try to reuse that?

integration/legacy_name_validation_test.go
```diff
@@ -0,0 +1,388 @@
...
+// invalid metric and label names.
+func TestLegacyNameValidation_Querier(t *testing.T) {
+	for _, queryEngine := range []string{"mimir", "prometheus"} {
+		t.Run(queryEngine, func(t *testing.T) {
+			testLegacyNameValidationQuerier(t, queryEngine)
+		})
+	}
+}
+
+func testLegacyNameValidationQuerier(t *testing.T, queryEngine string) {
```
dimitarvdimitrov, line 186: it might be more valuable to test this on the query-frontend. That's the component which will first receive the queries.

dimitarvdimitrov (commented):
overall these make sense, but I'm wondering if a unit test wouldn't be a better fit for this. It can be run with much more inputs and takes much less time. There's already some unit tests for this kind of assertions in pkg/distributor/validate_test.go (TestDistributor_ValidateSeries), for the ruler in pkg/ruler/api_test.go (TestAPI_CreateRuleGroup) and the query-frontend in pkg/frontend/v2/frontend_test.go (TestFrontendEnqueueFailures)

juliusmh:
A similar test was merged as part of https://github.com/grafana/mimir/pull/12215. 
Closing, as not needed.

PR #11867, opened 2025-06-25
Block-builder-scheduler: Job monitor.

seizethedave:
#### What this PR does

This PR:

* introduces a per-partition "planned offset" which is the offset up to where planning has been done for that partition. (It is the latest planned job's end offset.) (3449dd5b09dba066902d1663be23bcfde2e7eb96)
* committed offsets (along with the planned offset) is moved from a scheduler-wide map to a field in `partitionState`. (a65b4f8ac08d689bcf099ac3b9f9b59c9d7d5ce9)
* implements gap monitoring among jobs planned and completed. A detected gap increments a new metric: `cortex_blockbuilder_scheduler_job_gap_detected`, but otherwise does not prohibit offset movement. (ff125405c04507f6b690a7e8f7e14de3960baf15)
* adds more rigor to the transition between startup mode and regular mode: (61e5ae5be4c5339226b942b79d2b46da19bf4fb5)
  * jobs are imported and honored such that their successive (start, end) pairs form a gap-free path from the partition's commit offset. (job<sub>i</sub>'s start offset should be equal to job<sub>i-1</sub>'s end offset.)
  * such contiguous jobs in the transition will move the partition's planned offset forward and will influence where we begin planning new jobs for the partition. 
* Fixes a data race when enqueueing new jobs. (cef55a72b28f2050ab420121d5e79487e4d5e915)
* Fix a bug where a made-up job could be reported as completed, and the committed offset would move forward. (9b299541cc51f0f9e38f25dd4929bf8dcee49ba7)

Behavioral notes:
* Today we only allow one job per partition. But, there are ways (trust me, dude) involving multiple scheduler restarts where a partition may have multiple jobs in progress. In the future, we plan to have explicit support for multiple jobs per partition. This PR starts to acknowledge that future by supposing at startup time that multiple jobs may exist in a partition.
* If a partition's jobs observed at startup time do not form a gap-free path, then we will reject some of those jobs. When the scheduler enters normal operation mode, those rejected jobs will be met with "job not found" errors. In the future, we may tell a worker to abandon the job. Today, they'll just be wasted consumption work. I mention this in interest of completeness but this will probably be an edge case resting on the very precipice of another rare edge case.

#### Checklist

- [x] Tests updated.
- n/a Documentation added.
- n/a [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

### DIFF
--- a/pkg/blockbuilder/scheduler/scheduler.go
+++ b/pkg/blockbuilder/scheduler/scheduler.go
@@ -42,6 +42,9 @@ type BlockBuilderScheduler struct {
 	observations        obsMap
 	observationComplete bool
 	partState           map[int32]*partitionState
+
+	// for synchronizing tests.
+	onScheduleUpdated func()
 }
 
 func New(
@@ -58,6 +61,8 @@ func New(
 
 		observations: make(obsMap),
 		partState:    make(map[int32]*partitionState),
+
+		onScheduleUpdated: func() {},
 	}
 	s.Service = services.NewBasicService(s.starting, s.running, s.stopping)
 	return s, nil
@@ -137,6 +142,7 @@ func (s *BlockBuilderScheduler) running(ctx context.Context) error {
 			}
 
 			s.updateSchedule(ctx)
+			s.onScheduleUpdated()
 
 		case <-ctx.Done():
 			return nil
--- a/pkg/blockbuilder/scheduler/scheduler_test.go
+++ b/pkg/blockbuilder/scheduler/scheduler_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/dskit/services"
 	"github.com/prometheus/client_golang/prometheus"
 	promtest "github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
@@ -39,13 +40,16 @@ func mustSchedulerWithKafkaAddr(t *testing.T, addr string) (*BlockBuilderSchedul
 	cli := mustKafkaClient(t, addr)
 	cfg := Config{
 		Kafka: ingest.KafkaConfig{
-			Topic: "ingest",
+			Address:      addr,
+			Topic:        "ingest",
+			FetchMaxWait: 10 * time.Millisecond,
 		},
 		ConsumerGroup:       "test-builder",
 		SchedulingInterval:  1000000 * time.Hour,
 		JobSize:             1 * time.Hour,
 		MaxJobsPerPartition: 1,
 	}
+
 	reg := prometheus.NewPedanticRegistry()
 	sched, err := New(cfg, test.NewTestingLogger(t), reg)
 	sched.adminClient = kadm.NewClient(cli)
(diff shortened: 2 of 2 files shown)


pkg/blockbuilder/scheduler/scheduler_test.go
```diff
@@ -286,121 +304,145 @@ func TestObservations(t *testing.T) {
...
+	mkJob(inProgress, "w103", 5, "ingest/5/12000", 33, 12000, 13000, maybeBadEpoch, errBadEpoch)
+	mkJob(inProgress, "w104", 5, "ingest/5/12000", 34, 12000, 13000, nil, nil)
+
+	// Partition 6 has a complete job, but wasn't among the offsets we learned
+	// from Kafka. We'll drop this as the start offset is nonzero, but not-found
+	// completed jobs are ignored on update.
+	mkJob(complete, "w0", 6, "ingest/6/500", 48, 500, 600, nil, nil)
+	// Partition 7 has an in-progress job, but wasn't among the offsets we
+	// learned from Kafka. We'll drop this one, too.
+	mkJob(inProgress, "w1", 7, "ingest/7/92874", 52, 92874, 93874, nil, errJobNotFound)
```
seizethedave, line 457: this was `complete` by mistake before.

pkg/blockbuilder/scheduler/metrics.go
```diff
@@ -64,5 +65,9 @@ func newSchedulerMetrics(reg prometheus.Registerer) schedulerMetrics {
 			Name: "cortex_blockbuilder_scheduler_pending_jobs",
 			Help: "The number of jobs in the pending queues.",
 		}, []string{"partition"}),
+		jobGapDetected: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "cortex_blockbuilder_scheduler_job_gap_detected",
+			Help: "The number of times an unexpected gap was detected between jobs.",
+		}, []string{"offset"}),
```
narqo: (nit) After passing through the whole code, it's turned out this label isn't about a topic's "offset", but the internal's state (can be "planned" or "committed"). Can we name the label with something that doesn't overlap with existing Kafka terms, e.g. `phase="planned|committed"`, or something.

(nit 2) Can we add the `partition` label to be consistent with other metrics.

CHANGELOG.md
```diff
@@ -104,6 +104,7 @@
 * [ENHANCEMENT] Query-frontend: Accurate tracking of samples processed from cache. #11719
 * [ENHANCEMENT] Compactor, distributor, ruler, scheduler and store-gateway: Makes `-<component-ring-config>.auto-forget-unhealthy-periods` configurable for each component. Deprecates the `-store-gateway.sharding-ring.auto-forget-enabled` flag. #11923
 * [ENHANCEMENT] otlp: Stick to OTLP vocabulary on invalid label value length error. #11889
+* [ENHANCEMENT] Block-builder-scheduler: detect gaps in scheduled and completed jobs. #11867
```
```suggestion
* [ENHANCEMENT] Block-builder-scheduler: Detect gaps in scheduled and completed jobs. #11867
```

pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -192,7 +192,7 @@ func (s *BlockBuilderScheduler) updateSchedule(ctx context.Context) {
 		s.metrics.partitionEndOffset.WithLabelValues(partStr).Set(float64(o.Offset))
 
 		s.mu.Lock()
-		ps := s.getPartitionState(o.Partition)
+		ps := s.getPartitionState(o.Topic, o.Partition)
```
narqo: _It's unrelated to the changes in this PR but_,

I wonder if there is maybe a race around the access to `ps` down here. I.e. we don't hold the lock, when we do `ps.addPendingJob`, which modifies the `ps.pendingJobs`. But there is also the `enqueuePendingJobs` goroutine that reads-modifies the `pendingJobs` list. I might be wrong here.

narqo (approved):
I think, I got the general shape of the changes. Left a couple of nitpicking comments, but overall this works for me. Let's test it 🔥

tacole02 (approved):

pkg/blockbuilder/scheduler/scheduler.go
seizethedave: Oy. Oversight that I'm calling `addPendingJob` outside the lock.

seizethedave: fixed cef55a72b28f2050ab420121d5e79487e4d5e915

seizethedave: Locally, I wrote a test that instantiates and runs the scheduler using its Service interface with its various goroutines, and I strung up some client simulator goroutines and ran the whole thing with the race detector and the thing you pointed out is the only race it found. 🍀 

I'll stash this test (f77efeeae3) as it needs more work before it's ready to be merged.

PR #11868, opened 2025-06-25
docs/how-to-decrease-tenant-shard-size

tacole02:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/sources/mimir/configure/configure-shuffle-sharding/index.md
+++ b/docs/sources/mimir/configure/configure-shuffle-sharding/index.md
@@ -129,17 +129,21 @@ If you’re running a Grafana Mimir cluster with shuffle sharding disabled, and 
 
 #### Limitation: Decreasing the tenant shard size
 
-The current shuffle sharding implementation in Grafana Mimir has a limitation that prevents you from safely decreasing the tenant shard size when you enable ingesters’ shuffle sharding on the read path.
+The current shuffle sharding implementation in Grafana Mimir has a limitation that prevents you from abruptly decreasing the tenant shard size when you enable ingesters’ shuffle sharding on the read path. Decreasing the tenant shard size without following the complete procedure might result in query failures.
 
-If a tenant’s shard decreases in size, there is currently no way for the queriers and rulers to know how large the tenant shard was previously, and as a result, they potentially miss an ingester with data for that tenant.
-The blocks-storage.tsdb.retention-period, which is used to select the ingesters that might have received series since 'now - blocks-storage.tsdb.retention-period', doesn't work correctly for finding tenant shards if the tenant shard size is decreased.
+When you decrease a tenant's shard size, the queriers and rulers can't determine how large the tenant shard was previously. As a result, they could potentially miss an ingester with data for a given tenant.
+The `blocks-storage.tsdb.retention-period` setting, which is used to select the ingesters that might have received series since the value of 'now - blocks-storage.tsdb.retention-period', doesn't work correctly for finding tenant shards if you decrease the tenant shard size.
 
-Although decreasing the tenant shard size is not supported, consider the following workaround:
+To safely decrease the tenant shard size, follow these steps.
 
 1. Disable shuffle sharding on the read path via `-querier.shuffle-sharding-ingesters-enabled=false`.
 1. Decrease the configured tenant shard size.
 1. Wait for at least the amount of time specified via `-blocks-storage.tsdb.retention-period`.
 1. Re-enable shuffle sharding on the read path via `-querier.shuffle-sharding-ingesters-enabled=true`.
+
+{{< admonition type="note" >}}
+The procedure for decreasing the tenant shard size is the same as for enabling shuffle sharding. Enabling shuffle sharding on the ingesters effectively decreases the tenant shard size.
+{{< /admonition >}}
 
 ### Query-frontend and query-scheduler shuffle sharding
 


docs/sources/mimir/configure/configure-shuffle-sharding/index.md
```diff
@@ -129,18 +129,22 @@ If you’re running a Grafana Mimir cluster with shuffle sharding disabled, and
 
 #### Limitation: Decreasing the tenant shard size
 
-The current shuffle sharding implementation in Grafana Mimir has a limitation that prevents you from safely decreasing the tenant shard size when you enable ingesters’ shuffle sharding on the read path.
+The current shuffle sharding implementation in Grafana Mimir has a limitation that prevents you from abruptly decreasing the tenant shard size when you enable ingesters’ shuffle sharding on the read path. Decreasing the tenant shard size without following the complete procedure might result in query failures.
```
francoposa: I know this is not new copy but this part is confusing:
> when you enable ingesters’ shuffle sharding on the read path
This makes it sound like it is part of the process, but it was likely already on for anyone using shuffle sharding and if it was not already on, they would not want to turn it on until after this process is over.

It should be passive to indicate that it has likely already been enabled in the past something "when ingester shuffle sharding is/has been enabled on the read path".

docs/sources/mimir/configure/configure-shuffle-sharding/index.md
```diff
@@ -129,18 +129,22 @@ If you’re running a Grafana Mimir cluster with shuffle sharding disabled, and
...
 
-If a tenant’s shard decreases in size, there is currently no way for the queriers and rulers to know how large the tenant shard was previously, and as a result, they potentially miss an ingester with data for that tenant.
-The blocks-storage.tsdb.retention-period, which is used to select the ingesters that might have received series since 'now - blocks-storage.tsdb.retention-period', doesn't work correctly for finding tenant shards if the tenant shard size is decreased.
+When you decrease a tenant's shard size, the queriers and rulers can't determine how large the tenant shard was previously. As a result, they could potentially miss an ingester with data for a given tenant.
```
francoposa: I think we are missing a very brief explanation of *why* this occurs, which is the whole reason we need this process in the first place.

When the `-querier.shuffle-sharding-ingesters-enabled=false`, the queriers check all ingesters for the blocks, when it is true, they check only those that are supposed to be assigned to that tenant via the shuffle-shard.

May not be worth including, but when it is off, you may get slightly increased latency and CPU usage as the queriers have to cycle through asking more ingesters for the data.

tacole02: I've added in these details, let me know what you think.

I left out the part about increased latency/CPU usage when it's disabled, as that doesn't seem directly related to the process for safely decreasing the tenant shard size, but correct me if I'm wrong.

tacole02:
Thanks for your feedback, @francoposa . I've incorporated your points. Please let me know how this looks!

docs/sources/mimir/configure/configure-shuffle-sharding/index.md
```diff
@@ -122,25 +122,28 @@ Keeping ingesters shuffle sharding enabled only on the write path does not lead
...
 
-If a tenant’s shard decreases in size, there is currently no way for the queriers and rulers to know how large the tenant shard was previously, and as a result, they potentially miss an ingester with data for that tenant.
-The blocks-storage.tsdb.retention-period, which is used to select the ingesters that might have received series since 'now - blocks-storage.tsdb.retention-period', doesn't work correctly for finding tenant shards if the tenant shard size is decreased.
+Decreasing the tenant shard size without following this procedure might result in query failures, because the queriers and rulers can't determine how large the tenant shard was previously. As a result, they could potentially miss an ingester with data for a given tenant. The `blocks-storage.tsdb.retention-period` setting, which is used to select the ingesters that might have received series since the value set in `now - blocks-storage.tsdb.retention-period`, doesn't work correctly for finding tenant shards if you decrease the tenant shard size.
```
dimitarvdimitrov: is it failures or incorrect results?

tacole02: This is a good distinction to make. @francoposa could you please clarify?

francoposa: It is failures, they manifest as 500s with the querier logging `err-mimir-store-consistency-check-failed`.

francoposa: Actually wait - that is how it works for store-gateways. Let me check the ingester path

francoposa: Looked into this - the ingester queryable does not do the consistency check so the failure would take the form of some or all of the data being missing from the series sent back to the querier.

I would phrase it as "incorrect or incomplete results".
E.g. instant query with a range vector returns the exact data points, so IMO that would be more "incomplete"  - with data points missing - than "incorrect".
"Incorrect" would come into play when things like rates & sums are aggregated over that window with missing data points.

docs/sources/mimir/configure/configure-shuffle-sharding/index.md
```diff
@@ -122,25 +122,28 @@ Keeping ingesters shuffle sharding enabled only on the write path does not lead
 
 If you’re running a Grafana Mimir cluster with shuffle sharding disabled, and you want to enable it for the ingesters, use the following rollout strategy to avoid missing querying for any series currently in the ingesters:
 
-1. Explicitly disable ingesters shuffle-sharding on the read path via `-querier.shuffle-sharding-ingesters-enabled=false` since this is enabled by default.
+1. Explicitly disable ingesters shuffle-sharding on the read path via `-querier.shuffle-sharding-ingesters-enabled=false`, since this is enabled by default.
```
francoposa: Nit: for a little bit more clarity or ease of understanding I would replace usages of
> ingesters shuffle-sharding on the read path
with
> shuffle-sharding on the ingester read path

We have 3 types of "shuffle-sharding on the read path"  - on the ingesters, queriers, and store-gateways.
In this scenario we only focused on the difference between the "shuffle sharding on the ingester read path" vs "shuffle sharding on the ingester write path".

tacole02: That makes sense to me. Thanks for the suggestion!

tacole02:
Thanks for the feedback, @francoposa and @dimitarvdimitrov . I've updated the docs accordingly. LMK how this looks 😄

docs/sources/mimir/configure/configure-shuffle-sharding/index.md
dimitarvdimitrov: that tracks, thanks for double-checking

francoposa (approved):
LGTM, just I think that compound modifier “previously assigned ingester set” should he hyphenated, “previously-assigned ingester set”.

tacole02:
Good catch, @francoposa ! I fixed it before merging.

PR #11877, opened 2025-06-27
MQE: alternative way to ensure pooled slices are not reused

charleskorn:
#### What this PR does

This PR is an alternative to #11741, which implements an idea from @bboreham [here](https://github.com/grafana/mimir/pull/11690#issuecomment-2999379812).

Compared to #11741, this avoids a need for a linter, at the cost of a slightly less obvious behaviour and slightly uglier code.

https://github.com/grafana/mimir/pull/11690

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


56quarters (approved):
I prefer this approach instead of https://github.com/grafana/mimir/pull/11741 since this does not require a linter and when we make a mistake and pass a non-pointer to `Put()` the type system catches it.

PR #11189, opened 2025-04-11
MQE: initial implementation of common subexpression elimination

charleskorn:
#### What this PR does

This PR implements common subexpression elimination in MQE.

Prior to this PR, when given an expression like `sum(foo) / (sum(foo) + sum(bar))`, MQE would evaluate `sum(foo)` twice. This is not necessary: we can instead compute `sum(foo)` once and use the result in both places, and that's what this PR implements. 

This improves overall query performance, and also saves CPU time in queriers, ingesters and store-gateways. However, the drawback is that queriers must buffer unused parts of common results until they are consumed by all places that need them. For queries where the size of the common result is small, this is not an issue, but in other cases this can increase the peak memory consumption of queries significantly. 

This could be addressed somewhat by consuming from both sides of binary operations concurrently, but this is out of scope for this PR - for now, given the substantial performance improvements and relative cost of CPU and memory, we can provision more memory for queriers if needed. In any case, query memory consumption will be no worse than Prometheus' engine.

##### Benchmark results

Overall, latency improves roughly in line with the ratio of eliminated selectors to original selectors. (For example, if there were two selectors originally, and they're both the same, then latency improves around 50%, or if there were three selectors and two are the same, then latency improves around 33%.)

<details>

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/streamingpromql/benchmarks
cpu: Apple M1 Pro
                                                                                │    Mimir     │        MimirWithQueryPlanner         │
                                                                                │    sec/op    │    sec/op      vs base               │
Query/sum(a_1),_instant_query-10                                                   144.7µ ± 3%    146.8µ ±  1%        ~ (p=0.180 n=6)
Query/sum(a_1),_range_query_with_100_steps-10                                      149.8µ ± 1%    153.6µ ±  2%   +2.51% (p=0.004 n=6)
Query/sum(a_1),_range_query_with_1000_steps-10                                     202.9µ ± 1%    208.0µ ±  0%   +2.47% (p=0.002 n=6)
Query/sum(a_100),_instant_query-10                                                 757.4µ ± 1%    768.9µ ±  2%   +1.52% (p=0.015 n=6)
Query/sum(a_100),_range_query_with_100_steps-10                                    1.270m ± 1%    1.273m ±  1%        ~ (p=0.394 n=6)
Query/sum(a_100),_range_query_with_1000_steps-10                                   5.541m ± 1%    5.538m ±  1%        ~ (p=0.937 n=6)
[...]

PR #11879, opened 2025-06-27
MQE: add tests for common subexpression elimination for expressions with multiple children, and add metric on number of selectors eliminated

charleskorn:
#### What this PR does

This PR makes two changes to common subexpression elimination:

* Tests: I was worried we weren't handling the case where a expression has multiple children correctly, so added some tests to check. Turns out this case is OK, but I think it'd be valuable to retain so it's not broken in the future.
* Metric: the existing `cortex_mimir_query_engine_common_subexpression_elimination_duplication_nodes_introduced` metric is useful for gauging the impact of CSE, but doesn't tell us how much load on ingesters and store-gateways we've avoided. The new metric tells us the number of selectors eliminated and therefore gives us a better idea of the number of avoided ingester and store-gateway requests.

#11189 

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
+++ b/pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
@@ -30,6 +30,7 @@ import (
 
 type OptimizationPass struct {
 	duplicationNodesIntroduced prometheus.Counter
+	selectorsEliminated        prometheus.Counter
 }
 
 func NewOptimizationPass(reg prometheus.Registerer) *OptimizationPass {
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
```diff
@@ -50,7 +55,7 @@ func (e *OptimizationPass) Apply(_ context.Context, plan *planning.QueryPlan) (*
 	paths := e.accumulatePaths(plan)
 
 	// For each path: find all the other paths that terminate in the same selector, then inject a duplication node
-	err := e.groupAndApplyDeduplication(paths, 0)
```
56quarters, line 53: Stylistic choice but: If the `groupAndApplyDeduplication()` method was changed to return a count of the number of selectors eliminated, we could remove the "have eliminated" boolean and increment the metric here single time. I don't feel strongly either way.

charleskorn:
Given the approval, I'm going to merge this as-is so it's part of this week's release while we discuss https://github.com/grafana/mimir/pull/11879#discussion_r2172182055

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
charleskorn: Good idea. How's https://github.com/grafana/mimir/pull/11900?

PR #11890, opened 2025-06-27
fix(ingester): Keep read only when any number of compactions in progress

### LINKED ISSUE #11850: Bug: Ingesters can exit read-only mode during TSDB compactions
### What is the bug?

There is a bug in how Mimir handles TSDB head compaction that was previously reported and partially resolved by https://github.com/grafana/mimir/pull/11664. That PR fixed the issue for TSDB compactions started when the `force` flag is set as an argument to `Ingester.CompactBlocks`. This issue deals with TSDB compactions that are triggered by a tenant's TSDB being idle.

When an ingester has a tenant running TSDB  head compaction, the state of the tenant's `userTSDB` is set to `forceCompacting` ([ref](https://github.com/grafana/mimir/blob/b7ef8854f464ce30e2687040a9b2acf8cd208f76/pkg/ingester/user_tsdb.go#L221)). When a write is sent to a TSDB that is in `forceCompacting`, the write is conditionally rejected if the `forcedCompactionMaxTime` is larger than the push's minimum timestamp. ([ref](https://github.com/grafana/mimir/blob/b7ef8854f464ce30e2687040a9b2acf8cd208f76/pkg/ingester/user_tsdb.go#L484)). Because TSDB compactions triggered by idleness set `forcedCompactionMaxTime` to `math.MaxInt64`, all writes to these DBs will be rejected.

At Grafana Labs, we have observed cases where  the following sequence occurs: 

- A StatefulSet scales down and puts an ingester replica into  `readOnly` mode. 
- After ~1h, the user's TSDB begins a compaction due to idleness. 
    - Note: This process starts between 60m-75m of no-writes due to jitter (see: https://github.com/grafana/mimir/issues/2699).  More precisely, between 1-1.25 * `blocks-storage.tsdb.head-compaction-idle-timeout`.
- Shortly after head compaction begins, the STS scales back up and the replica rejoins the ring in an active state. 

This final step should not be allowed. If multiple zones are compacting due to idleness at the same time, all pushes with series sharded to that ingester will fail. As noted on https://github.com/grafana/mimir/pull/11664.

> read-only mode while a forced compaction is pending or in progress. Leaving read-only mode during the compaction will cause write errors, because writes are blocked.

The ingester shouldn't be able to leave read-only mode because a `forcedCompaction` (triggered by DB idleness) is in progress. This fails because atomics in `Ingester.compactBlocks` are only set when `force` is set.

A scale-down and scale-up separated by approx. `head_compaction_idle_timeout` is just one way this bug can appear. Consider the following situation:

- A tenant receives frequent writes from 00:00 to 01:00
- At 01:00, the tenant stops writing all series
- Between 02:05 and 02:10, all three TSDBs begin head compaction. For argument's sake - we'll assume this takes 5m.
- At 2:10, writes restart. With all three zones running compaction, writes will fail until 2/3 finish.
 
This version of the bug may not be resolved in the same way as the version described above.

periklis:
#### What this PR does
This pull requests extends the guardrails for ingesters leaving the read-only state when compaction is in progress. The PR #11664 adds a single guardrail for force compactions only. In contrast this PR extends to any compactions in progress (idle, force). This PR provides a fix for the situation where an ingester was in read-only while a downscale was in progress and it started to compact because of idleness. When doing this early compaction job, the ingesters received an upscale event in following left read-only mode without the early compaction having finished.

In detail the approach taken is to repurpose the ingester-wide atomic `i.forcedCompactionInProgress (bool)` to be an atomic counter `i.numCompactionsInProgress` for total amount of times the `compactBlocks` method is called. Each call of this method is incrementing before and decrementing the counter after compaction of the collection of TSDBs hosted. The rationale for a counter instead of a single boolean flag is that `compactBlocks` is called concurrently in many places for different reasons, i.e.:
- Idle compaction is a timer (plus jitter) based routine
- Force compaction is called by the Flush handler directly or during downscales by the rollout-operator.

Fixes #11850

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -120,6 +120,7 @@
 * [BUGFIX] Ruler: Fix rare panic when the ruler is shutting down. #11781
 * [BUGFIX] Block-builder-scheduler: Fix data loss bug in job assignment. #11785
 * [BUGFIX] Compactor: start tracking `-compactor.max-compaction-time` after the initial compaction planning phase, to avoid rare cases where planning takes longer than `-compactor.max-compaction-time` and so actual compaction never runs for a tenant. #11834
+* [BUGFIX] Ingester: Fix issue where ingesters could leave read-only mode during idle compactions, resulting in write errors. #11890
```
```suggestion
* [BUGFIX] Ingester: Fix issue where ingesters can exit read-only mode during idle compactions, resulting in write errors. #11890
```

pkg/ingester/ingester.go
```diff
@@ -3269,14 +3269,15 @@ func (i *Ingester) compactionServiceInterval() (firstInterval, standardInterval
 
 // Compacts all compactable blocks. Force flag will force compaction even if head is not compactable yet.
 func (i *Ingester) compactBlocks(ctx context.Context, force bool, forcedCompactionMaxTime int64, allowed *util.AllowList) {
+	i.numCompactionsInProgress.Inc()
+	defer i.numCompactionsInProgress.Dec()
+
 	// Expose a metric tracking whether there's a forced head compaction in progress.
 	// This metric can be used in alerts and when troubleshooting.
 	if force {
```
dmwilson-grafana: It seems we don't need the `forcedCompactionInProgress` metric anymore. It may be useful to replace it with a `compactionsInProgress` metric.

pkg/ingester/downscale.go
```diff
@@ -52,9 +52,9 @@ func (i *Ingester) PrepareInstanceRingDownscaleHandler(w http.ResponseWriter, r
 		i.circuitBreaker.push.deactivate()
 
 	case http.MethodDelete:
-		// Don't leave read-only mode if there's a forced compaction pending or in progress.
-		if len(i.forceCompactTrigger) > 0 || i.forcedCompactionInProgress.Load() {
-			msg := "cannot clear read-only mode while forced compaction is in progress"
+		// Don't leave read-only mode if there is any compaction pending or in progress.
+		if i.numCompactionsInProgress.Load() != 0 {
```
dmwilson-grafana, line 56: See Oleg's point from [a previous review](https://github.com/grafana/mimir/pull/11664/files#r2135191777). Right now it's possible for the following actions to happen: 

- trigger a force compaction 
- this code runs + ingester leaves read-only mode 
- force compaction starts - `numCompactionsInProgress` set.

This might not be an issue, just something to consider. This sequence is quite unlikely and even then, we're successfully fulfilling a forced compaction that was requested (and forced compactions are allowed when the ingester isn't in readOnly).

CHANGELOG.md
```diff
@@ -120,6 +120,7 @@
 * [BUGFIX] Ruler: Fix rare panic when the ruler is shutting down. #11781
 * [BUGFIX] Block-builder-scheduler: Fix data loss bug in job assignment. #11785
 * [BUGFIX] Compactor: start tracking `-compactor.max-compaction-time` after the initial compaction planning phase, to avoid rare cases where planning takes longer than `-compactor.max-compaction-time` and so actual compaction never runs for a tenant. #11834
+* [BUGFIX] Ingester: Fix issue where ingesters can exit read-only mode during idle compactions, resulting in write errors. #11890
```
dmwilson-grafana, line 123: nit: I'd just add `#11890` to the previous BUGFIX line (L119).

dmwilson-grafana (approved):

PR #11895, opened 2025-06-27
Bring BufferHolder back to PrometheusResponse

Konstantinov-Innokentii:
#### What this PR does

BufferHolder was added in this [PR](https://github.com/grafana/mimir/pull/11136) and mistakenly removed  later ina different [PR](https://github.com/grafana/mimir/pull/11568/files#diff-f460500f40606393884c046e42540c09ff5c7cd6146d2b1f59c6b3d7f2b320e6R85). I confirmed with @aknuds1 that it should be there.

#### Checklist

- [N/A] Tests updated.
- [N/A] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/model.pb.go
+++ b/pkg/frontend/querymiddleware/model.pb.go
@@ -83,6 +83,9 @@ func (m *PrometheusHeader) GetValues() []string {
 }
 
 type PrometheusResponse struct {
+	// Keep reference to buffer for unsafe references.
+	github_com_grafana_mimir_pkg_mimirpb.BufferHolder
+
 	Status    string              `protobuf:"bytes,1,opt,name=Status,proto3" json:"status"`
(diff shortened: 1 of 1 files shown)


56quarters (approved):
Could we add something that would have caught this at build time? I'm assuming it was removed because the author of the PR that removed it had a different version of `protoc` installed. Could we build protos during CI and error if they are any changes to the working tree?

Konstantinov-Innokentii:
> Could we add something that would have caught this at build time? I'm assuming it was removed because the author of the PR that removed it had a different version of `protoc` installed. Could we build protos during CI and error if there are any changes to the working tree?

We already doing it, by running make check-protos in CI. The thing is that BufferHolder is applied using the `apply_expected_diffs.sh` script, which applies the content of `.expdiff` file. 
It seems there is a corner case, when changes are staged, apply_expected_diffs.sh will not apply the expdiff file.

PR #11930, opened 2025-07-01
MQE: Add support for sort_by_label and sort_by_label_desc functions

56quarters:
#### What this PR does

Add support for the experimental PromQL functions sort_by_label and sort_by_label_desc in the Mimir Query Engine.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -496,6 +496,48 @@ func TimestampFunctionOperatorFactory(args []types.Operator, memoryConsumptionTr
...
+			l, ok := args[i].(types.StringOperator)
+			if !ok {
+				var suffix string
+				if i == 1 {
+					suffix = "nd"
+				} else if i == 2 {
+					suffix = "rd"
+				} else {
+					suffix = "th"
+				}
```
lamida: Somewhat nits, I understand this is the initial draft. But just want to note that the suffix won't handle 21st, 22nd, 23rd, 31st, etc. But I guess nobody will sort with so that many labels anyway. ¯\_(ツ)_/¯

56quarters: D'oh, forgot about those. I'll just remove it or rephrase the message.

pkg/streamingpromql/operators/functions/sort_by_label.go
```diff
@@ -0,0 +1,85 @@
...
+
+	s.allData = make([]types.InstantVectorSeriesData, len(allSeries))
+	for idx := range allSeries {
+		d, err := s.inner.NextSeries(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		s.allData[idx] = d
+	}
```
charleskorn: I don't think it's necessary to load all the data upfront for `sort_by_label` / `sort_by_label_desc` given we're sorting only by the series labels.

I'm imagining we'd do something along the lines of:

1. load all input series' metadata
2. sort the list of series based on their labels, and keep track of which output series corresponds to which output series (eg. the series that sorts first and is returned first is the 3rd input series, the second output series is the 5th input series etc.), and retain this list for use in `NextSeries`
3. use a `InstantVectorOperatorBuffer` to handle buffering series as we need them in `NextSeries`

This might reduce the value of introducing `sortBase` given there will be far less shared code.

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -496,6 +496,39 @@ func TimestampFunctionOperatorFactory(args []types.Operator, memoryConsumptionTr
...
+			l, ok := args[i].(types.StringOperator)
+			if !ok {
+				// Should be caught by the PromQL parser, but we check here for safety.
+				return nil, fmt.Errorf("expected a string for argument %d, got %T", i+1, args[i])
+			}
+
+			labels = append(labels, l.GetValue())
+		}
+
+		return NewSortByLabel(inner, labels, descending, memoryConsumptionTracker, expressionPosition), nil
```
charleskorn: If this is a range query, should we just return `inner` here given `sort_by_label` doesn't do anything for range queries?

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -997,6 +997,49 @@ eval range from 0 to 4m step 1m sort_desc(test_metric)
 
 clear
 
+# Test sort_by_label / sort_by_label_desc.
```
charleskorn, line 1000: Would be good to see some tests for range queries as well.

pkg/streamingpromql/operators/functions/sort_by_label.go
56quarters: I've implemented this in 91a4403a0f581709efab3e6271c050e7f238d145

pkg/streamingpromql/operators/functions/factories.go
56quarters: Done in 6f501e635c1299a271b6d238ee0f73019c72564e

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -496,6 +496,46 @@ func TimestampFunctionOperatorFactory(args []types.Operator, memoryConsumptionTr
...
+	functionName := "sort_by_label"
+	if descending {
+		functionName = "sort_by_label_desc"
+	}
+
+	return func(args []types.Operator, memoryConsumptionTracker *limiter.MemoryConsumptionTracker, _ *annotations.Annotations, expressionPosition posrange.PositionRange, timeRange types.QueryTimeRange) (types.InstantVectorOperator, error) {
+		if len(args) < 2 {
+			// Should be caught by the PromQL parser, but we check here for safety.
+			return nil, fmt.Errorf("expected at least 2 arguments for %s, got %v", functionName, len(args))
+		}
```
charleskorn: Is it valid to not provide any labels, something like `sort_by_label(foo)`?

pkg/streamingpromql/operators/functions/sort_by_label.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Create a mapping of each series to its original position before sorting.
+	s.originalPositions = make(map[uint64]int, len(innerMetadata))
+	for i, series := range innerMetadata {
+		s.originalPositions[series.Labels.Hash()] = i
+	}
+
+	slices.SortFunc(innerMetadata, compareForLabels(s.labels, s.descending))
+
+	s.sortedMetadata = innerMetadata
```
charleskorn: As a general rule, it's not safe to retain the returned metadata slice like this - the rule we've been following is that the slice returned by `SeriesMetadata` becomes owned by the caller, and that caller is free to modify it (or return it to a pool) as it sees fit.

pkg/streamingpromql/operators/functions/sort_by_label.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, err
+	}
+
+	// Create a mapping of each series to its original position before sorting.
+	s.originalPositions = make(map[uint64]int, len(innerMetadata))
+	for i, series := range innerMetadata {
+		s.originalPositions[series.Labels.Hash()] = i
+	}
+
+	slices.SortFunc(innerMetadata, compareForLabels(s.labels, s.descending))
```
charleskorn: What if we created a slice containing the indices of the inner series, and swapped them around as we swap the series around during sorting? Then we could avoid the use of a map and needing to hash the labels, and we don't need to retain the labels for use in `NextSeries` either.

eg. imagine the input series are `[d, c, a, b]`. We'd create an initial indices slice of `[0, 1, 2, 3]`, then as we sort the series, we'd update that slice as well, with a final result of `[2, 3, 1, 0]`.

Then in `NextSeries` we can get the next series index from that slice.

56quarters: Good idea, I'll give this a shot.

pkg/streamingpromql/operators/functions/factories.go
56quarters: From the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/functions/#sort_by_label), it doesn't seem like it's valid but I'll test locally to confirm.

pkg/streamingpromql/operators/functions/sort_by_label.go
56quarters: Fixed in 31c4a6aa571d9c95ac198b6c27b21ef69b9f2b12

pkg/streamingpromql/operators/functions/sort_by_label.go
56quarters: Implemented in 31c4a6aa571d9c95ac198b6c27b21ef69b9f2b12. Is that close to what you were thinking?

pkg/streamingpromql/operators/functions/factories.go
56quarters: Turns out Prometheus will let you not include any labels. It just sorts by all labels in that case. I'll adjust our behavior to match.

pkg/streamingpromql/testdata/ours/functions.test
56quarters: Added in 7e3a39562b00f9dcb86b72d9796eb28c8f87d300

pkg/streamingpromql/operators/functions/factories.go
56quarters: Fixed in e146309e11c693eb6942aff20752af993405ce70

pkg/streamingpromql/operators/functions/sort_by_label.go
```diff
@@ -53,63 +52,83 @@ func (s *SortByLabel) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
...
 	}
 
-	// Create a mapping of each series to its original position before sorting.
-	s.originalPositions = make(map[uint64]int, len(innerMetadata))
-	for i, series := range innerMetadata {
-		s.originalPositions[series.Labels.Hash()] = i
+	// Store the original indexes of each piece of metadata and then sort them alongside
+	// the metadata. This allows us to return series in sorted order when requesting them
+	// from the inner operator by their original index.
+	indexes := make([]int, len(innerMetadata))
```
charleskorn: We should be able to take this slice from a pool (and we'd then need to return it to the pool in `Close`)

pkg/streamingpromql/operators/functions/sort_by_label.go
```diff
@@ -0,0 +1,152 @@
...
+	return s.expressionPosition
+}
+
+func (s *SortByLabel) Prepare(ctx context.Context, params *types.PrepareParams) error {
+	return s.inner.Prepare(ctx, params)
+}
+
+func (s *SortByLabel) Close() {
+	// Closing the buffer also closes its source operator which is our `inner`.
+	s.buffer.Close()
```
charleskorn: If `SeriesMetadata` wasn't called for some reason (or it returned an error before `s.buffer` is set), then we'll leave `inner` unclosed. All operators should be safe to close multiple times, so calling `inner.Close` here should be safe to do.

charleskorn (commented):
LGTM, two suggestions:

tacole02 (approved):

pkg/streamingpromql/operators/functions/sort_by_label.go
56quarters: Done in d5ff77f3902f7a9c3ef6dc6eac2c58f5f23daefe. It would correct to put the call to `s.inner.Close()` in the `else` statement but I'm not sure it's needed since we can call `.Close()` twice without issue.

pkg/streamingpromql/operators/functions/sort_by_label.go
56quarters: Done in d5ff77f3902f7a9c3ef6dc6eac2c58f5f23daefe

pkg/streamingpromql/operators/functions/functions.proto
```diff
@@ -67,6 +67,8 @@ enum Function {
   FUNCTION_SIN = 58;
   FUNCTION_SINH = 59;
   FUNCTION_SORT = 60;
+  FUNCTION_SORT_BY_LABEL = 76;
```
56quarters, line 70: @charleskorn I elected to keep these in sorted order but picked the next unused ordinals. LMK if this isn't the right thing to do.

charleskorn: LGTM, thanks Nick.

PR #11935, opened 2025-07-01
Account for weight in memory HPA for resource-scaled objects

chencs:
#### What this PR does
This PR fixes an inconsistency where the `weight` for `ResourceScaledObject`s is honored when building the CPU HPA query, but not when building the memory HPA query.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir/autoscaling.libsonnet
```diff
@@ -331,29 +331,28 @@
...
-                max by (pod) (up{container="%(container)s",namespace="%(namespace)s"%(extra_matchers)s}) > 0
-              ) or vector(0)
-            )[15m:]
+              sum by (pod) (max_over_time(kube_pod_container_resource_requests{container="%(container)s", namespace="%(namespace)s", resource="memory"%(extra_matchers)s}[15m]))
+              and
+              max by (pod) (changes(kube_pod_container_status_restarts_total{container="%(container)s", namespace="%(namespace)s"%(extra_matchers)s}[15m]) > 0)
+              and
+              max by (pod) (kube_pod_container_status_last_terminated_reason{container="%(container)s", namespace="%(namespace)s", reason="OOMKilled"%(extra_matchers)s})
+              or vector(0)
+            )
```
aknuds1: What's the purpose of this change? Is it just to get the comment into the rendered query?

operations/mimir-tests/test-autoscaling-custom-target-utilization-generated.yaml
```diff
@@ -2066,15 +2069,6 @@ spec:
...
         )
-        +
-        sum(
-          sum by (pod) (max_over_time(kube_pod_container_resource_requests{container="distributor", namespace="default", resource="memory"}[15m]))
-          and
-          max by (pod) (changes(kube_pod_container_status_restarts_total{container="distributor", namespace="default"}[15m]) > 0)
-          and
-          max by (pod) (kube_pod_container_status_last_terminated_reason{container="distributor", namespace="default", reason="OOMKilled"})
-          or vector(0)
-        )
```
aknuds1: Is this supposed to be dropped, or is it by mistake?

operations/mimir/autoscaling.libsonnet
aknuds1: On reading it again, this looks like a possible bug to me. I.e. the following part of the `memoryHPAQuery` query becomes dependent on the `with_ready_trigger` argument being false:

```jsonnet
// Add pods that were terminated due to an OOM in the memory calculation.
|||
  +
  sum(
    sum by (pod) (max_over_time(kube_pod_container_resource_requests{container="%(container)s", namespace="%(namespace)s", resource="memory"%(extra_matchers)s}[15m]))
    and
    max by (pod) (changes(kube_pod_container_status_restarts_total{container="%(container)s", namespace="%(namespace)s"%(extra_matchers)s}[15m]) > 0)
    and
    max by (pod) (kube_pod_container_status_last_terminated_reason{container="%(container)s", namespace="%(namespace)s", reason="OOMKilled"%(extra_matchers)s})
    or vector(0)
  )
|||

```

tacole02 (approved):

operations/mimir/autoscaling.libsonnet
chencs: The goal was to correctly set the memory query into parentheses so that `metricWithWeight` would adjust the entire value and not just one portion of the memory query.

I've fixed this by having `metricWithWeight` handle putting queries into parentheses; it means that some queries (e.g., the CPU HPA query) will have unnecessary parentheses, but I don't think it should break anything.

operations/mimir/autoscaling.libsonnet
```diff
@@ -66,7 +66,7 @@
   // metricWithWeight will multiply the metric by the weight provided if it's different from 1.
   local metricWithWeight(metric, weight) =
```
aknuds1: Maybe out of scope, but when reviewing this PR it seems to me that this function should be called `queryWithWeight` and the `metric` arg should be `query`, since in fact a query is modified. It's much less confusing to the reviewer when they know the function acts on a query :)

aknuds1 (approved):
LGTM, with a suggestion for readability that's probably out of scope for this PR.

operations/mimir/autoscaling.libsonnet
chencs: Done!

PR #11939, opened 2025-07-02
Move query pruning and experimental functions middlewares earlier in query-frontend pipeline

charleskorn:
#### What this PR does

This PR moves the query pruning and experimental functions middlewares earlier in the query-frontend pipeline.

This has a number of benefits:

* these middlewares now run before splitting and sharding, so they are only applied once to the entire query rather than once per split and sharded leg
* these middlewares no longer run after sharding, which will make it easier to migrate sharding to a MQE query planner optimisation pass in the future
* the experimental functions middleware now runs immediately after the error caching middleware, so queries that will be rejected are rejected and cached before doing as much work as possible

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
   * `memberlist.max-concurrent-writes`: `5`
   * `memberlist.acquire-writer-timeout`: `1s`
     These defaults perform better but may cause long-running packets to be dropped in high-latency networks.
-* [CHANGE] Query-frontend: apply query pruning and check for disabled experimental functions earlier in query processing. #11939
+* [CHANGE] Query-frontend: Apply query pruning and check for disabled experimental functions earlier in query processing. #11939
 * [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. Limitations: Created timestamp is ignored, per series metadata is merged on metric family level automatically, ingestion might fail if client sends ProtoBuf fields out of order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0` depending on the detected Remote-Write protocol. #11100 #11101 #11192 #11143
 * [FEATURE] Query-frontend: expand `query-frontend.cache-errors` and `query-frontend.results-cache-ttl-for-errors` configuration options to cache non-transient response failures for instant queries. #11120
 * [FEATURE] Query-frontend: Allow use of Mimir Query Engine (MQE) via the experimental CLI flags `-query-frontend.query-engine` or `-query-frontend.enable-query-engine-fallback` or corresponding YAML. #11417 #11775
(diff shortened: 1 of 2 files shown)


pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -445,6 +445,34 @@ func newQueryMiddlewares(
...
+			queryRangeMiddleware,
+			newInstrumentMiddleware("pruning", metrics),
+			pruneMiddleware,
+		)
+		queryInstantMiddleware = append(
+			queryInstantMiddleware,
+			newInstrumentMiddleware("pruning", metrics),
+			pruneMiddleware,
+		)
+	}
```
charleskorn, line 474: Note to reviewers: I can see pros and cons to moving this immediately after the splitting and caching middleware.

Pros: if the entire result has already been cached, then we'll skip the work of pruning the expression

Cons: if at least two parts of the query haven't been cached, we'll end up applying pruning to the same expression multiple times

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
   * `memberlist.max-concurrent-writes`: `5`
   * `memberlist.acquire-writer-timeout`: `1s`
     These defaults perform better but may cause long-running packets to be dropped in high-latency networks.
+* [CHANGE] Query-frontend: apply query pruning and check for disabled experimental functions earlier in query processing. #11939
```
dimitarvdimitrov: Realistically, this shouldn't have a noticeable effect on the user, should it?

pkg/frontend/querymiddleware/roundtrip.go
dimitarvdimitrov: > Cons: if at least two parts of the query haven't been cached, we'll end up applying pruning to the same expression multiple times

I don't exactly get this. Why two parts? Isn't one uncached extent enough?

CHANGELOG.md
```suggestion
* [CHANGE] Query-frontend: Apply query pruning and check for disabled experimental functions earlier in query processing. #11939
```

pkg/frontend/querymiddleware/roundtrip.go
charleskorn: >Why two parts? Isn't one uncached extent enough?

If the query is split into multiple extents, then the split query for each extent passes through all of middlewares after the cache/split middleware independently. So if multiple extents aren't cached, and the pruning middleware runs after the cache/split middleware, then pruning would be applied to each extent independently.

CHANGELOG.md
charleskorn: It should be almost imperceptible, unless the query is being split into multiple extents, in which case there may be a very slight performance benefit.

charleskorn:
Given the approvals, I'm going to merge this, but if there are any further questions I'm happy to discuss them.

pkg/frontend/querymiddleware/roundtrip.go
dimitarvdimitrov: So you meant twice within the same HTTP request. I think I understood twice within different HTTP requests. So that's where the confusion came from. Thanks for explaining

PR #11952, opened 2025-07-02
[ingest-storage] Update ingester autoscaling settings

pr00se:
#### What this PR does

- Remove the scaling trigger based on average series, and use only the trigger based on max owned series. We have found it to perform better in production at Grafana Labs.
- Change the ingester downscale behavior to allow 15% downscale within 30 minutes rather than just 10%.

Fixes #N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir/ingest-storage-ingester-autoscaling.libsonnet
```diff
@@ -10,16 +10,12 @@
     ingest_storage_ingester_autoscaling_min_replicas_per_zone: error 'you must set ingest_storage_ingester_autoscaling_min_replicas_per_zone in the _config in namespace %s' % $._config.namespace,
     ingest_storage_ingester_autoscaling_max_replicas_per_zone: error 'you must set ingest_storage_ingester_autoscaling_max_replicas_per_zone in the _config in namespace %s' % $._config.namespace,
```
pr00se, line 12: only real change is in this file

56quarters (approved):
LGTM. Do we need to update the default configuration settings in ingesters to use owned series to go along with this?

pr00se:
> LGTM. Do we need to update the default configuration settings in ingesters to use owned series to go along with this?

I don't think so. ingest storage is still experiemental, and the while the trigger this deletes was named "average active series", the [actual query](https://github.com/grafana/mimir/pull/11952/files#diff-bd35861d132cac1cacaa7e906e854d84e283106f32fa6eefad8b9ee1c9fedc84L46) was already using owned series.

PR #11953, opened 2025-07-02
rw2: Validate that the first symbol of the RW2 Symbols field is an empty string

alexweav:
#### What this PR does

Per the [Remote-Write 2.0 spec](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/#symbols):

> The first element of the symbols table MUST be an empty string, which is used to represent empty or unspecified values such as when Metadata.unit_ref or Metadata.help_ref are not provided.

Obviously Prometheus [follows its own spec](https://github.com/prometheus/prometheus/blob/74aca682b77d66f5761bd68459f3b570ec34ce2d/prompb/io/prometheus/write/v2/symbols.go#L27-L29), but other clients might send a request that's out of spec. If a client does this, we accept their request, but store garbage. Typically, this looks like mangling their metric metadata into a random label name used elsewhere in the request, possibly for an unrelated series.

We cannot decipher the meaning of such a request. Say a metadata unit has a reference value of 0, we can't tell whether the client intended to set it to the first symbol or to leave it blank. We don't have enough info to try to interpret it, in some cases we will always mangle metadata.

The ingest-storage record format V2 had a compounded version of this bug. In that format, we reserve the first N symbols for common symbols. An unsupplied metadata field serializes as the value `0`, which is in the common symbols space. We add a mandate that common symbols must start with empty string just like regular symbols, for correctness. Even if common symbols are not supplied, we'll still understand the meaning of `0`. It's redundant - in practice, rejecting such requests at the distributor shields this layer from this problem.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/compat_rw2_test.go
+++ b/pkg/mimirpb/compat_rw2_test.go
@@ -306,6 +306,44 @@ func TestRW2Unmarshal(t *testing.T) {
 		require.ErrorContains(t, err, "invalid")
 	})
 
+	t.Run("zero refs translate to empty string despite offset", func(t *testing.T) {
+		syms := test.NewSymbolTableBuilderWithCommon(nil, 256, nil)
+		writeRequest := &rw2.Request{
+			Timeseries: []rw2.TimeSeries{
+				{
+					LabelsRefs: []uint32{syms.GetSymbol("__name__"), syms.GetSymbol("test_metric_total"), syms.GetSymbol("job"), syms.GetSymbol("test_job")},
+					Samples: []rw2.Sample{
+						{
+							Value:     123.456,
+							Timestamp: 1234567890,
+						},
+					},
+					Exemplars: []rw2.Exemplar{
+						{
+							Value:      123.456,
+							Timestamp:  1234567890,
+							LabelsRefs: []uint32{syms.GetSymbol("__name__"), syms.GetSymbol("test_metric_total"), syms.GetSymbol("traceID"), syms.GetSymbol("1234567890abcdef")},
+						},
+					},
+					Metadata: rw2.Metadata{
+						Type:    rw2.Metadata_METRIC_TYPE_COUNTER,
+						HelpRef: syms.GetSymbol("test_metric_help"),
+						// UnitRef: left default!
+					},
+				},
+			},
+		}
+		data, err := writeRequest.Marshal()
+		require.NoError(t, err)
+
+		received := PreallocWriteRequest{}
+		received.UnmarshalFromRW2 = true
+		received.RW2SymbolOffset = 256
+		err = received.Unmarshal(data)
+		require.NoError(t, err)
+		require.Equal(t, "", received.Metadata[0].Unit)
+	})
+
 	t.Run("common symbol out of bounds", func(t *testing.T) {
 		commonSyms := []string{"__name__"}
 		syms := test.NewSymbolTableBuilderWithCommon(nil, 256, commonSyms)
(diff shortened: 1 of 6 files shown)


pkg/util/test/rw2.go
```diff
@@ -108,6 +108,11 @@ func NewSymbolTableBuilder(symbols []string) *SymbolTableBuilder {
 }
 
 func NewSymbolTableBuilderWithCommon(symbols []string, offset uint32, commonSymbols []string) *SymbolTableBuilder {
+	// RW2.0 Spec: The first element of the symbols table MUST be an empty string.
+	if len(symbols) == 0 || symbols[0] != "" {
+		symbols = append([]string{""}, symbols...)
+	}
+
```
alexweav, line 115: Preventing this generator from building requests for tests that are out of spec

CHANGELOG.md
```diff
@@ -130,6 +130,7 @@
 * [BUGFIX] Block-builder-scheduler: Fix data loss bug in job assignment. #11785
 * [BUGFIX] Compactor: start tracking `-compactor.max-compaction-time` after the initial compaction planning phase, to avoid rare cases where planning takes longer than `-compactor.max-compaction-time` and so actual compaction never runs for a tenant. #11834
 * [BUGFIX] Ingester: Fix issue where ingesters can exit read-only mode during idle compactions, resulting in write errors. #11890
+* [BUGFIX] Distributor: Validate the RW2 symbols field, reject invalid requests that do not have empty string as the first symbol.
```
alexweav: Note: Ingest storage record format V2 is not stable. It's currently impossible for a user to produce records of that format, hence I didn't add that part of the fix to the changelog, as it can't affect anyone yet.

pkg/storage/ingest/version.go
```diff
@@ -23,6 +23,7 @@ var (
 	// The first `V2RecordSymbolOffset` symbols are reserved for this table.
 	// Note: V2 is not yet stabilized.
 	V2CommonSymbols = []string{
+		"",
```
pracucci, line 28: [nit] Should we add a comment to explain why?

pkg/mimirpb/compat_rw2_test.go
```diff
@@ -323,6 +365,31 @@ func TestRW2Unmarshal(t *testing.T) {
 		err = received.Unmarshal(data)
 		require.ErrorContains(t, err, "invalid")
 	})
+
+	t.Run("messages where the first symbol is empty string are rejected", func(t *testing.T) {
```
pracucci: Isn't the opposite? We error out if the first symbol is not an empty string.

```suggestion
	t.Run("messages where the first symbol is not empty string are rejected", func(t *testing.T) {
```

pracucci (approved):
Good job spotting it!

If a request fail to unmarshal because it doesn't follow the specs it should be a 4xx and not a 5xx. I think with the current code it will generate a 4xx. Can you take a look, please? I also suggest to add a test case to `TestDistributor_Push` testing the case of an invalid RW2 request.

CHANGELOG.md
```diff
@@ -130,6 +130,7 @@
 * [BUGFIX] Block-builder-scheduler: Fix data loss bug in job assignment. #11785
 * [BUGFIX] Compactor: start tracking `-compactor.max-compaction-time` after the initial compaction planning phase, to avoid rare cases where planning takes longer than `-compactor.max-compaction-time` and so actual compaction never runs for a tenant. #11834
 * [BUGFIX] Ingester: Fix issue where ingesters can exit read-only mode during idle compactions, resulting in write errors. #11890
+* [BUGFIX] Distributor: Validate the RW2 symbols field, reject invalid requests that do not have empty string as the first symbol. #11953
```
```suggestion
* [BUGFIX] Distributor: Validate the RW2 symbols field and reject invalid requests that don't have an empty string as the first symbol. #11953
```

tacole02 (approved):

pkg/mimirpb/compat_rw2_test.go
alexweav: ah! yes.
triple checked the test is still testing the right thing, just the title is wrong.

alexweav:
The reason why I think the status code looks confusing, is that this PR does the validation inside the actual deserialization layer, not among the other request validation. That's because the RW2 symbols substitution also happens during deserialization.  If the symbols are bad, deserialization can succeed and produce an incorrect result, hence it felt cleaner to have the unmarshal just fail rather than create a footgun if a caller forgets to check the validity of the output.

`TestDistributor_Push` is also too low level to cover it, it operates in terms of requests that are already deserialized and symbols have already been substituted. Instead I extended the e2e distributor test to cover it.

Maybe it makes sense to also validate it lower down in the distributor? It's redundant, but it's not expensive to check, and the alternative is to store incorrect data. WDYT?

pracucci (approved):
Good point about the distribution unit test being too low level: you're totally right. The integration test you added is the best way to test it. I wouldn't go further than it: looks good enough! 🚀

PR #11936, opened 2025-07-02
MQE: encode query plan node type as an integer, rather than a string

charleskorn:
#### What this PR does

This PR improves the way query plans are encoded, replacing the use of a string like `type.googleapis.com/BinaryExpressionDetails` to identify the type of each node with an enum.

This significantly reduces the size of encoded payloads, which was the primary goal of this change. Previously, very simple expressions could result in very large encoded query plans (eg. `(a - b) + (c - d) + (e * f) + (g * h)`, which is 37 characters, would have resulted in a 1.12 KB encoded query plan).

It also significantly improves encoding and decoding time, although this was already fast in absolute terms.

Benchmark results:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/streamingpromql
cpu: Apple M1 Pro
                                                                                                                               │ original.txt  │            with-int.txt            │
                                                                                                                               │    sec/op     │   sec/op     vs base               │
PlanEncodingAndDecoding/foo/encode-10                                                                                             385.6n ± 18%   301.0n ± 1%  -21.95% (p=0.002 n=6)
PlanEncodingAndDecoding/foo/decode-10                                                                                             434.9n ±  1%   339.1n ± 1%  -22.03% (p=0.002 n=6)
PlanEncodingAndDecoding/foo{env="test",_region="us"}/encode-10                                                                    429.9n ±  0%   334.2n ± 1%  -22.25% (p=0.002 n=6)
PlanEncodingAndDecoding/foo{env="test",_region="us"}/decode-10                                                                    624.8n ±  0%   526.0n ± 0%  -15.81% (p=0.002 n=6)
PlanEncodingAndDecoding/sum(foo)/encode-10                                                                                        673.2n ±  2%   480.6n ± 0%  -28.61% (p=0.002 n=6)
PlanEncodingAndDecoding/sum(foo)/decode-10                                                                                        779.5n ±  1%   575.7n ± 0%  -26.14% (p=0.002 n=6)
PlanEncodingAndDecoding/sum_by_(env)_(foo)/encode-10                                                                              682.2n ±  1%   503.2n ± 1%  -26.23% (p=0.002 n=6)
PlanEncodingAndDecoding/sum_by_(env)_(foo)/decode-10                                                                              829.2n ±  0%   617.6n ± 0%  -25.52% (p=0.002 n=6)
[...]

PR #11955, opened 2025-07-03
MQE: encode query plan functions as an integer, rather than a string, and simplify function handling

charleskorn:
#### What this PR does

This PR builds on #11936, and adopts a similar approach for functions.

This further reduces query plan size, and improves encoding and decoding performance slightly:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/streamingpromql
cpu: Apple M1 Pro
                                                                                                                               │ with-int.txt │     with-function-as-int.txt      │
                                                                                                                               │    sec/op    │   sec/op     vs base              │
PlanEncodingAndDecoding/sum(rate(foo[5m]))/encode-10                                                                              684.7n ± 0%   679.5n ± 0%  -0.75% (p=0.004 n=6)
PlanEncodingAndDecoding/sum(rate(foo[5m]))/decode-10                                                                              864.4n ± 0%   832.9n ± 0%  -3.64% (p=0.002 n=6)
PlanEncodingAndDecoding/sum_by_(zone)_(label_replace(rate(foo[5m]),_"zone",_"$1",_"pod",_"ingester-zone-(.*)-\\d+"))/encode-10    1.487µ ± 0%   1.468µ ± 0%  -1.24% (p=0.002 n=6)
PlanEncodingAndDecoding/sum_by_(zone)_(label_replace(rate(foo[5m]),_"zone",_"$1",_"pod",_"ingester-zone-(.*)-\\d+"))/decode-10    1.893µ ± 0%   1.853µ ± 0%  -2.09% (p=0.002 n=6)
PlanEncodingAndDecoding/max_over_time(rate(foo[5m:])[10m:])_+_max_over_time(rate(bar[5m:])[10m:])/encode-10                       2.769µ ± 0%   2.720µ ± 1%  -1.77% (p=0.002 n=6)
PlanEncodingAndDecoding/max_over_time(rate(foo[5m:])[10m:])_+_max_over_time(rate(bar[5m:])[10m:])/decode-10                       3.272µ ± 2%   3.185µ ± 0%  -2.64% (p=0.002 n=6)
PlanEncodingAndDecoding/label_join(foo,_"abc",_"-")_+_label_join(bar,_"def",_",")/encode-10                                       1.990µ ± 2%   1.970µ ± 0%  -0.98% (p=0.002 n=6)
PlanEncodingAndDecoding/label_join(foo,_"abc",_"-")_+_label_join(bar,_"def",_",")/decode-10                                       2.162µ ± 0%   2.113µ ± 2%  -2.24% (p=0.002 n=6)

                                                                                                                               │ with-int.txt │      with-function-as-int.txt      │
                                                                                                                               │      B       │     B       vs base                │
PlanEncodingAndDecoding/sum(rate(foo[5m]))/encode-10                                                                              100.00 ± 0%   96.00 ± 0%  -4.00% (p=0.002 n=6)
PlanEncodingAndDecoding/sum_by_(zone)_(label_replace(rate(foo[5m]),_"zone",_"$1",_"pod",_"ingester-zone-(.*)-\\d+"))/encode-10     301.0 ± 0%   284.0 ± 0%  -5.65% (p=0.002 n=6)
PlanEncodingAndDecoding/max_over_time(rate(foo[5m:])[10m:])_+_max_over_time(rate(bar[5m:])[10m:])/encode-10                        367.0 ± 0%   333.0 ± 0%  -9.26% (p=0.002 n=6)
PlanEncodingAndDecoding/label_join(foo,_"abc",_"-")_+_label_join(bar,_"def",_",")/encode-10                                        267.0 ± 0%   247.0 ± 0%  -7.49% (p=0.002 n=6)

```

The main drawback of this approach is that it makes handling additional functions (eg. those introduced by Grafana's Adaptive Metrics query rewriting) a little more awkward: we must either declare those functions in `functions.proto`, or at least reserve a value for them, to ensure that we don't end up with conflicting functions with the same value.

This PR also makes two other changes to how functions are handled:

* it removes the special cases for `absent` and `absent_over_time`
* it unifies the handling of functions that return scalars and instant vectors

I would recommend reviewing each commit separately.

#11936

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/planning.go
```diff
@@ -416,16 +418,17 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
 	}
 }
 
-func isKnownFunction(name string) bool {
-	if _, ok := functions.InstantVectorFunctionOperatorFactories[name]; ok {
-		return true
+func findFunction(name string) (functions.Function, bool) {
+	f, ok := functions.FromPromQLName(name)
```
dimitarvdimitrov, line 422: woudln't we always find it in `RegisteredFunctions`? Why do the protobuf lookup first?

charleskorn: The key of `RegisteredFunctions` is a `functions.Function`, so we need to translate the name into a `functions.Function` to look it up in `RegisteredFunctions` first.

dimitarvdimitrov: sorry, it was a stupid question, i had misread this function

charleskorn: No worries 🙂

PR #11956, opened 2025-07-03
MQE: adjust behaviour of binary operations with `on (__name__)` to match Prometheus

charleskorn:
#### What this PR does

This PR changes the behaviour of MQE's implementation of binary operations to match Prometheus' query engine.

Specifically, it changes MQE to match how Prometheus handles binary operations with `on (__name__)`. Prometheus always drops the metric name label, even if it is listed in `on`, as discussed in https://github.com/prometheus/prometheus/issues/16631. Changing this in Prometheus is considered a breaking change, so [I've updated the Prometheus documentation](https://github.com/prometheus/prometheus/pull/16716) to explicitly state that the current behaviour is expected, and in this PR I'm updating MQE to match.

https://github.com/prometheus/prometheus/issues/16631

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


lamida (approved):

lamida (approved):
Didn't think about hot path, but indeed this new change looks more efficient.

PR #11957, opened 2025-07-03
Replace namespace with job label in golang_alerts

edgarasg:
#### What this PR does
Increasing scope of golang_alerts since labels as `namespace` are usually related to cloud deployments.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,7 @@
 * [ENHANCEMENT] Dashboards: Add "per-query memory consumption" and "fallback to Prometheus' query engine" panels to the Queries dashboard. #11626
 * [ENHANCEMENT] Alerts: Add `MimirGoThreadsTooHigh` alert. #11836 #11845
 * [ENHANCEMENT] Dashboards: Add autoscaling row for ruler query-frontends to `Mimir / Remote ruler reads` dashboard. #11838
+* [CHANGE] Alerts: Replace namespace with job label in golang_alerts. #11957
 * [BUGFIX] Dashboards: fix "Mimir / Tenants" legends for non-Kubernetes deployments. #10891
 * [BUGFIX] Dashboards: fix Query-scheduler RPS panel legend in "Mimir / Reads". #11515
 * [BUGFIX] Recording rules: fix `cluster_namespace_deployment:actual_replicas:count` recording rule when there's a mix on single-zone and multi-zone deployments. #11287
--- a/operations/mimir-mixin/alerts/alerts.libsonnet
+++ b/operations/mimir-mixin/alerts/alerts.libsonnet
@@ -975,7 +975,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
             alert: $.alertName('GoThreadsTooHigh'),
             expr: |||
               # We filter by the namespace because go_threads can be very high cardinality in a large organization.
-              max by(%(alert_aggregation_labels)s, %(per_instance_label)s) (go_threads{namespace=~".*(cortex|mimir).*"} > %(threshold)s)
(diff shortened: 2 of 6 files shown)


tacole02 (approved):

operations/mimir-mixin/alerts/alerts.libsonnet
```diff
@@ -980,7 +980,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
             alert: $.alertName('GoThreadsTooHigh'),
             expr: |||
               # We filter by the namespace because go_threads can be very high cardinality in a large organization.
-              max by(%(alert_aggregation_labels)s, %(per_instance_label)s) (go_threads{namespace=~".*(cortex|mimir).*"} > %(threshold)s)
+              max by(%(alert_aggregation_labels)s, %(per_instance_label)s) (go_threads{job=~".*(cortex|mimir).*"} > %(threshold)s)
```
narqo, line 978: This will break the Jsonnet-based deployments, where workload names don't have a fixed prefix.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Grafana Mimir
 
+* [CHANGE] Mixin: Replace namespace with job label in golang_alerts.
```
narqo: The mixins have a separate section down in the changelog (look for `### Mixin`). Could you move this there, after we finalize what needs a change for this PR.

narqo:
Thank you for the PR. From the description, I'm not fully understand the problem [..]

> labels as "namespace" are usually related to cloud deployments

What do you include in the "cloud deployments" — is this anything that doesn't run on Kubernetes?

narqo:
@pracucci you added the alert recently in #11836. Could you have a look at this PR?

edgarasg:
> Thank you for the PR. From the description, I'm not fully understand the problem [..]
> 
> > labels as "namespace" are usually related to cloud deployments
> 
> What do you include in the "cloud deployments" — is this anything that doesn't run on Kubernetes?

Correct. For ex. the "baremetal" deployment.

operations/mimir-mixin/alerts/alerts.libsonnet
edgarasg: Then maybe we can just add `baremetal` to the namespace regex?

CHANGELOG.md
edgarasg: Done.

operations/mimir-mixin/alerts/alerts.libsonnet
pracucci: > This will break the Jsonnet-based deployments, where workload names don't have a fixed prefix.

The job label pattern is expected to be `<namespace>/<deployment|statefulset>` so I think that works.

pracucci: > Then maybe we can just add baremetal to the namespace regex?

I don't think it will work. Baremetal deployments typically have no namespace.

pracucci:
@edgarasg Can you better explain me what specific problem you're trying to solve here? Can you give me more details about why the alert doesn't work to you? I guess you don't have the `namespace` label, but I would like to understand it a bit better.

edgarasg:
Yes, we run Mimir on VMs and basically we don't have `namespace` label with value `mimir`.

pracucci:
> Yes, we run Mimir on VMs and basically we don't have `namespace` label with value `mimir`.

And does `job=~".*(cortex|mimir).*` match for you?

edgarasg:
Yes, we have job labels separated by `mimir-read`/`write`/`backend`.

pracucci (approved):
LGTM, thanks!

pracucci:
I pushed a commit to recompile the alerts. I've run:
```
make build-mixin build-helm-tests
```

PR #11962, opened 2025-07-03
MQE: add tracing for CSE

dimitarvdimitrov:
#### What this PR does

I was looking at a trace and couldn't figure out if CSE was enabled and if it did what I expected.

I also added some more details around what has been deduplicated. The details shouldn't be too noisy, since it looks like `.Describe()` only returns the description top-level node and doesn't recurse.

Fixes #<issue number>

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
+++ b/pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
@@ -71,7 +71,7 @@ func (e *OptimizationPass) Apply(ctx context.Context, plan *planning.QueryPlan) 
 	e.selectorsEliminated.Add(float64(selectorsEliminated))
 
 	spanLog := spanlogger.FromContext(ctx, log.NewNopLogger())
-	spanLog.DebugLog("msg", "attempted common subexpression elimination", "inspected_selectors", len(paths), "eliminated_selectors", selectorsEliminated)
+	spanLog.DebugLog("msg", "attempted common subexpression elimination", "selectors_inspected", len(paths), "selectors_eliminated", selectorsEliminated)
 
 	return plan, nil
 }


pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
```diff
@@ -263,6 +268,7 @@ func (e *OptimizationPass) introduceDuplicateNode(group []*path, duplicatePathLe
 	duplicatedExpression, _ := firstPath.NodeAtOffsetFromLeaf(duplicatePathLength - 1)
 	duplicate := &Duplicate{Inner: duplicatedExpression, DuplicateDetails: &DuplicateDetails{}}
 	e.duplicationNodesIntroduced.Inc()
+	spanlogger.FromContext(ctx, log.NewNopLogger()).DebugLog("msg", "eliminated common subexpression", "expression_root", duplicatedExpression.Describe())
```
charleskorn: I'm not sure this is going to be particularly helpful - `Describe()` will return  `"sum"` for an aggregation like `sum(foo)`.

Rather than adding this log line like this, what if we rely on the debugging line added above to know if CSE did anything, and rely on the analysis endpoint and UI for delving deeper into what happened?

dimitarvdimitrov: there's an analysis UI? 👀

charleskorn: _nods_ It's very rough and not yet publicly available: https://github.com/grafana/mqe-query-analysis-ui

dimitarvdimitrov: > Rather than adding this log line like this, what if we rely on the debugging line added above to know if CSE did anything, and rely on the analysis endpoint and UI for delving deeper into what happened?

Yeah, this makes sense.  I want to leave the summary logline which says how many expressions have been eliminated and we can use the analysis UI.

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
```diff
@@ -68,6 +70,9 @@ func (e *OptimizationPass) Apply(_ context.Context, plan *planning.QueryPlan) (*
 	e.selectorsInspected.Add(float64(len(paths)))
 	e.selectorsEliminated.Add(float64(selectorsEliminated))
 
+	spanLog := spanlogger.FromContext(ctx, log.NewNopLogger())
+	spanLog.DebugLog("msg", "attempted common subexpression elimination", "inspected_selectors", len(paths), "eliminated_selectors", selectorsEliminated)
```
charleskorn: [nit] Let's keep the log labels consistent with the metric names:

```suggestion
	spanLog.DebugLog("msg", "attempted common subexpression elimination", "selectors_inspected", len(paths), "selectors_eliminated", selectorsEliminated)
```

charleskorn (approved):
LGTM modulo nit, thanks for adding this

PR #11963, opened 2025-07-03
MQE: Fix estimations in TestMemoryConsumptionLimit_SingleQueries

dimitarvdimitrov:
#### What this PR does

follow-up of #11683

The memory estimations have peaks during different stages of query processing when running with `-tags stringlabels` and when running without.

When running without `stringlabels`, strings appear cheaper (because the plain `ByteSize()` doesn't account for slice overhead). Because of this, the peak is during floating point processing.

When running with `stringlabels`, strings appear more expensive. Then the peak is during labels processing.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine_test.go
+++ b/pkg/streamingpromql/engine_test.go
@@ -1593,9 +1593,10 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 						assert.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(rejectedMetrics(1)), "cortex_querier_queries_rejected_total"))
 					}
 
-					spanStubs := spanExporter.GetSpans()
-					assert.Len(t, spanStubs, 1)
-					spanStub := spanStubs[0]
+					var spanStub tracetest.SpanStub
+					if spanStubs := spanExporter.GetSpans(); assert.Len(t, spanStubs, 1) {
+						spanStub = spanStubs[0]
+					}
 					assertEstimatedPeakMemoryConsumption(t, reg, spanStub, expectedPeakMemoryConsumption, queryType)
 				})
 			}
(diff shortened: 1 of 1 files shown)


pkg/streamingpromql/engine_test.go
```diff
@@ -1563,15 +1586,15 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 					span.End()
 
 					if testCase.shouldSucceed {
-						require.NoError(t, res.Err)
-						require.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(rejectedMetrics(0)), "cortex_querier_queries_rejected_total"))
+						assert.NoError(t, res.Err)
```
dimitarvdimitrov, line 1589: I made these changes so that I can see what assertions are failing on the peak memory limit after the error assertion.

pkg/streamingpromql/engine_test.go
```diff
@@ -1563,15 +1586,15 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
...
 					} else {
-						require.ErrorContains(t, res.Err, globalerror.MaxEstimatedMemoryConsumptionPerQuery.Error())
-						require.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(rejectedMetrics(1)), "cortex_querier_queries_rejected_total"))
+						assert.ErrorContains(t, res.Err, globalerror.MaxEstimatedMemoryConsumptionPerQuery.Error())
+						assert.NoError(t, testutil.GatherAndCompare(reg, strings.NewReader(rejectedMetrics(1)), "cortex_querier_queries_rejected_total"))
 					}
 
 					spanStubs := spanExporter.GetSpans()
-					require.Len(t, spanStubs, 1)
+					assert.Len(t, spanStubs, 1)
```
56quarters: This one seems like it should probably stay a `require` since it will immediately panic after if the assertion doesn't hold.

56quarters (approved):
Nicely done!

pkg/streamingpromql/engine_test.go
dimitarvdimitrov: I still wanted to descend into the big memory assertions, even if the spans aren't there. pushed this https://github.com/grafana/mimir/pull/11963/commits/aad053466bfda0e2eaa341f8879d59c84a6e6621

PR #11968, opened 2025-07-03
Increase store-gateway -> querier max gRPC message size to 200MB

pr00se:
#### What this PR does

Increase the max store-gateway gRPC response size (and corresponding receive size on the queriers) to allow responses to e.g.: high-cardinality label values queries.

Fixes #N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/operations/mimir/store-gateway.libsonnet
+++ b/operations/mimir/store-gateway.libsonnet
@@ -35,10 +35,9 @@
 
       // Relax pressure on KV store when running at scale.
       'store-gateway.sharding-ring.heartbeat-period': '1m',
-    } +
-    (if 'store_gateway_grpc_max_query_response_size_bytes' in $._config then {
+
       'server.grpc-max-send-msg-size-bytes': $._config.store_gateway_grpc_max_query_response_size_bytes,
-    } else {}) +
+    } +
     (if !$._config.store_gateway_lazy_loading_enabled then {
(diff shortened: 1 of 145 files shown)


operations/mimir/store-gateway.libsonnet
```diff
@@ -36,6 +36,9 @@
       // Relax pressure on KV store when running at scale.
       'store-gateway.sharding-ring.heartbeat-period': '1m',
     } +
+    (if 'store_gateway_grpc_max_query_response_size_bytes' in $._config then {
```
56quarters: Why make this conditional on the config existing when it's explicitly set in `config.libsonnet`?

56quarters (dismissed):
Change seems fine but I don't really understand the conditional nature of it.

operations/mimir/store-gateway.libsonnet
pr00se: I guess I was considering it an optional config, so wanted to hedge against it possibly not being set at all. If you think it's unnecessary, I'll happily remove it (and the similar change in the helm charts).

56quarters: Yeah, I don't think we need to worry about it not existing since it's set in the config file.

56quarters (approved):
LGTM

PR #11971, opened 2025-07-03
Add more query planner AST optimisation passes

zenador:
#### What this PR does

Add more query planner AST optimisation passes:

1. Removing parts of queries that are toggled off (moving this functionality from query pruning middleware)
2. Propagating matchers across binary operations to avoid retrieving unused series
3. Reordering histogram functions for a more efficient execution order (left as frontend middleware for now, to be moved over later after query sharding is migrated)

Started a long time ago, but got stuck on how to implement (2) properly so that it can handle more complicated cases.

N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


zenador:
After discussing with Charles, we've decided to move all the query pruning functionality over to AST optimisation passes (https://github.com/grafana/mimir-squad/issues/3090).

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1787,8 +1787,18 @@ results_cache:
 # (experimental) True to enable pruning dead code (eg. expressions that cannot
 # produce any results) and simplifying expressions (eg. expressions that can be
 # evaluated immediately) in queries.
-# CLI flag: -query-frontend.prune-queries
-[prune_queries: <boolean> | default = false]
+# CLI flag: -query-frontend.prune-queries-toggle
+[prune_queries_toggle: <boolean> | default = false]
+
+# (experimental) True to enable rewriting queries to optimise query execution by
```
```suggestion
# (experimental) Set to true to enable rewriting queries to optimize query execution by
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -1787,8 +1787,18 @@ results_cache:
...
+# copying matchers where relevant across binary expressions.
+# CLI flag: -query-frontend.prune-queries-matcher-propagate
+[prune_queries_matcher_propagate: <boolean> | default = false]
+
+# (experimental) True to enable rewriting histogram queries for a more efficient
```
tacole02, line 1803:
```suggestion
# (experimental) Set to true to enable rewriting histogram queries for a more efficient
```

tacole02 (approved):

docs/sources/mimir/configure/configuration-parameters/index.md
zenador: No longer used since query pruning is now removed in here

docs/sources/mimir/configure/configuration-parameters/index.md
zenador: No longer used since query pruning is now removed in here

zenador:
@charleskorn A few points of discussion...

1) After trying to migrate the query pruning middleware functionality to the query planner, I found one reason we might want to leave some functionality in the middleware instead. The pruning middleware can be run before the query sharding middleware, whereas it looks like the AST optimisation passes are done after the query has been sharded. This is particularly relevant for the histogram reordering.

When we have this in the middleware, `histogram_sum(sum(foo))` gets re-ordered to `sum(histogram_sum(foo))`, which is passed to the sharding where it is rewritten to `sum(__embedded_queries__{__queries__="{\"Concat\":[{\"Expr\":\"sum(histogram_sum(metric_native_histogram{__query_shard__=\\\"1_of_2\\\"}))\"},{\"Expr\":\"sum(histogram_sum(metric_native_histogram{__query_shard__=\\\"2_of_2\\\"}))\"}]}"})` and does not get re-ordered afterwards. This works as intended.

When it's in the query planner, `histogram_sum(sum(foo))` is passed to the sharding where it is rewritten to `histogram_sum(sum(__embedded_queries__{__queries__="{\"Concat\":[{\"Expr\":\"sum(metric_native_histogram{__query_shard__=\\\"1_of_2\\\"})\"},{\"Expr\":\"sum(metric_native_histogram{__query_shard__=\\\"2_of_2\\\"})\"}]}"}))`, and then gets re-ordered to `sum(histogram_sum(__embedded_queries__{__queries__="{\"Concat\":[{\"Expr\":\"sum(metric_native_histogram{__query_shard__=\\\"1_of_2\\\"})\"},{\"Expr\":\"sum(metric_native_histogram{__query_shard__=\\\"2_of_2\\\"})\"}]}"}))` which throws errors like `vector cannot contain metrics with the same labelset`. I could fix this by avoiding doing any re-ordering on embedded queries which should make it work without any errors, but then we would not get the optimisation of the re-ordering.

Are there any plans or is there any way to make some of these AST optimisation passes run before the query sharding? If not, the histogram reordering might be better off in the pruning middleware.

2) Also, when trying to allow matcher propagation in aggregate queries like `sum by (cluster) (metric_1{cluster!="test-1"}) / sum by (cluster) (metric_2)`, it starts making query splitting tests fail. Haven't investigated in detail and don't really understand query splitting yet, but could this be a similar issue to (1)?

3) For now I'm reusing astmapper from querymiddleware as I find it helps us avoid a lot of extra code to implement the recursive mapping, but please let me know if it'd be problematic having that as a dependency.

charleskorn:
> 1. After trying to migrate the query pruning middleware functionality to the query planner, I found one reason we might want to leave some functionality in the middleware instead. The pruning middleware can be run before the query sharding middleware, whereas it looks like the AST optimisation passes are done after the query has been sharded. This is particularly relevant for the histogram reordering.
>
>     ...
>
>     Are there any plans or is there any way to make some of these AST optimisation passes run before the query sharding? If not, the histogram reordering might be better off in the pruning middleware.

Good catch. Yep, my plan is that query sharding will becoming a query plan optimisation pass, and I'm expecting this will happen in the next few months. For now, we could either:

* leave the histogram reordering out of this PR and return to it once we've migrated query sharding
* leave the histogram reordering as a query-frontend middleware



> 2\. Also, when trying to allow matcher propagation in aggregate queries like `sum by (cluster) (metric_1{cluster!="test-1"}) / sum by (cluster) (metric_2)`, it starts making query splitting tests fail. Haven't investigated in detail and don't really understand query splitting yet, but could this be a similar issue to (1)?

Not sure about this one sorry - let me know if you need some help digging into it.



> 3\. For now I'm reusing astmapper from querymiddleware as I find it helps us avoid a lot of extra code to implement the recursive mapping, but please let me know if it'd be problematic having that as a dependency.

I think that's fine for now.

pkg/streamingpromql/planning.go
```diff
@@ -50,8 +50,11 @@ func NewQueryPlanner(opts EngineOpts) *QueryPlanner {
 	// we introduce query-frontend-specific optimization passes like sharding and splitting for two reasons:
 	//  1. We want to avoid a circular dependency between this package and the query-frontend package where most of the logic for these optimization passes lives.
 	//  2. We don't want to register these optimization passes in queriers.
+	planner.RegisterASTOptimizationPass(&ast.PruneToggles{})          // Do this first to ensure that toggled off expressions are removed before any other optimization passes are applied.
 	planner.RegisterASTOptimizationPass(&ast.SortLabelsAndMatchers{}) // This is a prerequisite for other optimization passes such as common subexpression elimination.
 	planner.RegisterASTOptimizationPass(&ast.CollapseConstants{})
+	planner.RegisterASTOptimizationPass(&ast.ReorderHistogramAgg{})
+	planner.RegisterASTOptimizationPass(&ast.PropagateMatchers{})
```
charleskorn: It'd be good to have feature toggles for each of these new optimisation passes, so we can disable them if we run into any issues.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
```diff
@@ -0,0 +1,72 @@
...
+	newExpr := &parser.AggregateExpr{
+		Op: agg.Op,
+		Expr: &parser.Call{
+			Func: call.Func,
+			Args: []parser.Expr{agg.Expr},
+		},
+		Param:    agg.Param,
+		Grouping: agg.Grouping,
+		Without:  agg.Without,
+	}
```
charleskorn: This will drop the original expression position information, which means any annotations emitted during evaluation won't include position information.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,91 @@
...
+		`up * ignoring(foo) group_left down`: `up * ignoring(foo) group_left down`,
+		`up{foo="bar"} * 1`:                  `up{foo="bar"} * 1`,
+		`up * down`:                          `up * down`,
+		`up{foo="bar"} * down{foo="bar"}`:    `up{foo="bar"} * down{foo="bar"}`,
+		`up{foo="bar"} * down`:               `up{foo="bar"} * down{foo="bar"}`,
+		`up * down{foo="bar"}`:               `up{foo="bar"} * down{foo="bar"}`,
+		`up{foo="bar"} * down{bar="foo"}`:    `up{foo="bar", bar="foo"} * down{foo="bar", bar="foo"}`,
+		`up{foo="bar"} * down{foo!="bar2"}`:  `up{foo="bar", foo!="bar2"} * down{foo="bar", foo!="bar2"}`,
+		`up{foo="bar"} * ignoring(foo) down`: `up{foo="bar"} * ignoring(foo) down`,
+		`up{foo="bar", bar="foo", boo="far", far="boo"} * ignoring(foo, bar) down`: `up{foo="bar", bar="foo", boo="far", far="boo"} * ignoring(foo, bar) down{boo="far", far="boo"}`,
```
charleskorn: [nit] Reusing the label names as values in the matchers is a bit confusing, might be worth using different values to make it unambiguous.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,91 @@
...
+		`sum by (foo) (up) / sum by (foo) (down)`:                                  `sum by (foo) (up) / sum by (foo) (down)`,
+		`sum by (foo) (up{foo!="bar"}) / sum by (foo) (down)`:                      `sum by (foo) (up{foo!="bar"}) / sum by (foo) (down{foo!="bar"})`,
+		`sum by (foo) (up) / sum by (foo) (down{boo="far"})`:                       `sum by (foo) (up) / sum by (foo) (down{boo="far"})`,
+		`sum by (foo) (up{foo!="bar", boo="far"}) / sum by (foo) (down)`:           `sum by (foo) (up{foo!="bar", boo="far"}) / sum by (foo) (down{foo!="bar"})`,
+		`sum by (foo) (up) / sum(down)`:                                            `sum by (foo) (up) / sum(down)`,
+		`sum by (foo) (up{foo!="bar"}) / sum(down)`:                                `sum by (foo) (up{foo!="bar"}) / sum(down)`,
+		`sum by (foo) (up{foo!="bar"}) / sum by (foo) (down{boo="far"})`:           `sum by (foo) (up{foo!="bar"}) / sum by (foo) (down{foo!="bar", boo="far"})`,
+		`sum by (foo) (up{foo!="bar"}) / sum by (boo) (down{boo="far"})`:           `sum by (foo) (up{foo!="bar"}) / sum by (boo) (down{boo="far"})`,
```
charleskorn: Some other cases that might be interesting:

* aggregations with multiple grouping labels
* expressions with aggregations where matchers can be propagated from the RHS to the LHS
* aggregations with `without` rather than `by`
* expressions with aggregations where the binary operation uses `on` or `ignoring`
* aggregations where the selector is wrapped in another function (eg. `sum by (env) (abs(foo{env="prod"})) / sum by (env) (sin(bar))` becomes `sum by (env) (sin(bar{env="prod"}))` on the RHS
* `topk` and `bottomk`: I think the behaviour would be correct as it stands, but I think these deserve special tests so we don't break this in the future, because something like `topk by (env) (foo{env="prod"}) / bottomk by (region) (bar{region="eu"})` can't be rewritten to `topk by (env) (foo{env="prod", region="eu"}) / bottomk by (region) (bar{region="eu", env="prod"})`, even though all labels are returned by `topk` and `bottomk`

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -0,0 +1,84 @@
...
+		`foo`:                `foo`,
+		`foo[1m]`:            `foo[1m]`,
+		`foo{bar="baz"}[1m]`: `foo{bar="baz"}[1m]`,
+		`foo{bar="baz"}`:     `foo{bar="baz"}`,
+		`quantile(0.9,foo)`:  `quantile(0.9,foo)`,
+		`histogram_quantile(0.5, rate(bar1{baz="blip"}[30s]))`: `histogram_quantile(0.5, rate(bar1{baz="blip"}[30s]))`,
+		`count(up)`:                                `count(up)`,
+		`avg(rate(foo[1m]))`:                       `avg(rate(foo[1m]))`,
+		`vector(0)`:                                `vector(0)`,
+		`up < -Inf`:                                `up < -Inf`,
```
charleskorn: Might be worth adding a comment explaining why this can't be simplified to just `up`

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -0,0 +1,84 @@
...
+		`avg(rate(foo[1m])) < (-Inf)`:              `avg(rate(foo[1m])) < (-Inf)`,
+		`Inf * -1`:                                 `Inf * -1`,
```
charleskorn: Seeing examples like this makes me wonder if it'd make the logic in this optimisation pass simpler (or would cover more cases) if we require that the collapse constants pass runs first.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -0,0 +1,84 @@
...
+		`avg(rate(foo[1m])) < (-1 * +Inf)`:         `avg(rate(foo[1m])) < (-1 * +Inf)`,
+		`(-1 * -Inf) < avg(rate(foo[1m]))`:         `(-1 * -Inf) < avg(rate(foo[1m]))`,
+		`vector(0) < -Inf or avg(rate(foo[1m]))`:   `vector(0) < -Inf or avg(rate(foo[1m]))`,
```
charleskorn: Can't this be simplified to `avg(rate(foo[1m]))`?

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
zenador: Okay, will work on this. There are a lot of functions though, we'd need to go through them and consider which can work with matcher propagation. Should we just whitelist some functions like `abs` to start?

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
zenador: Yeah, this was written to run independently so it includes some mini collapse constants handling, but we could remove some logic and some test cases now that we're moving it into a series of AST optimisation passes. Is there a reason why `CollapseConstants` isn't already running first? It's currently after `SortLabelsAndMatchers`. Do we want to move `CollapseConstants` first and simplify this?

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
zenador: You might know this already but since it was quite some time ago, and to provide context for others if they're looking:

The original toggle query pruning, along with the original way we used to toggle between classic and native histograms in our dashboards, was to compare expressions we wanted to toggle to +/-Inf. Then there was a change in Prometheus such that this worked differently for native histograms and floats, so we couldn't use this method anymore, especially as we only work on parsing the expression and have no idea if the underlying series are floats or native histograms or both. So we changed the method in the dashboards to use `and on() (vector(1/-1)==1)` which works with both floats and native histograms, and updated the query pruning logic to match. We left the original unit tests in there, but changed the expected output to the unchanged input to verify that we were no longer rewriting according to the old logic.

I ported over the tests from the current query pruning middleware unchanged. So yes, by right `vector(0) < -Inf or avg(rate(foo[1m]))` can be simplified to `avg(rate(foo[1m]))` (because it is a known float vector and not some unknown series), but that is not the goal of the current toggle query pruning, so it's unhandled. AFAIK, since our dashboards were updated to the new logic, there's no reason we should be receiving such queries anymore. It is true that it can be optimised, but would it be adding unnecessary code to maintain if it is unlikely to be used?

So should we add the old query pruning based on comparison with +/-Inf back, but make it a pared down version that only works with known vectors and not unknown series? I'd put it separately from the new query pruning logic using `and on()` since they are different.

Either way we should probably remove the old unit tests from here to avoid confusion.

WDYT?

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
zenador: Related to the thread below.

zenador:
Okay, I'm moving histogram reordering back to the pruning middleware for now so we can start testing the logic earlier and then move it back later after query sharding is migrated. I left the code in the ast package since it'd be moved there eventually and the structure is pretty similar anyway, is that okay? Also, I'm not sure if I should rename the pruning middleware to rewriting middleware or something or if that'd be confusing.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
charleskorn: Having some kind of allow list of functions makes sense to me. It'd be good if we had a test that failed if new functions were added to PromQL in the future, so we're forced to either add them to the list of allowed functions or explicitly mark them as not supported. I think it'd also be OK to do this as a follow-up PR.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
charleskorn: I think it'd be fine to have `CollapseConstants` run first, especially if that means we don't need to duplicate logic. It'd be good to include a note in `planning.go` that we expect `CollapseConstants` to run before this optimisation pass.

cmd/mimir/config-descriptor.json
```diff
@@ -7543,12 +7565,12 @@
...
           "kind": "field",
-          "name": "prune_queries",
+          "name": "prune_queries_histogram",
           "required": false,
-          "desc": "True to enable pruning dead code (eg. expressions that cannot produce any results) and simplifying expressions (eg. expressions that can be evaluated immediately) in queries.",
+          "desc": "True to enable rewriting histogram queries for a more efficient order of execution.",
           "fieldValue": null,
           "fieldDefaultValue": false,
-          "fieldFlag": "query-frontend.prune-queries",
+          "fieldFlag": "query-frontend.prune-queries-histogram",
```
charleskorn: Maybe something like `query-frontend.enable-rewriting-histogram-queries` would be clearer? Using the term "prune" here might be confusing given nothing is removed.

pkg/frontend/querymiddleware/prune.go
```diff
@@ -10,22 +10,24 @@ import (
...
 type pruneMiddleware struct {
 	next   MetricsQueryHandler
 	logger log.Logger
+	cfg    Config
 }
 
 // newPruneMiddleware creates a middleware that optimises queries by rewriting them to prune away
 // unnecessary computations.
-func newPruneMiddleware(logger log.Logger) MetricsQueryMiddleware {
+func newPruneMiddleware(logger log.Logger, cfg Config) MetricsQueryMiddleware {
```
charleskorn: Same comment as for the CLI flag above - I'm not sure "prune" is the clearest name for what this does now.

pkg/frontend/querymiddleware/prune.go
```diff
@@ -62,12 +64,16 @@ func (p *pruneMiddleware) pruneQuery(ctx context.Context, query string) (string,
 		return "", false, apierror.New(apierror.TypeBadData, DecorateWithParamName(err, "query").Error())
 	}
 	origQueryString := expr.String()
+	prunedQuery := expr
```
charleskorn: I realise this was like this before, but rather than checking if the query changed based on converting it back to a string, could we have the mapper return a flag that indicates if it made any changes?

pkg/frontend/querymiddleware/prune.go
```diff
@@ -62,12 +64,16 @@ func (p *pruneMiddleware) pruneQuery(ctx context.Context, query string) (string,
...
 
-	mapper := astmapper.NewQueryPruner(ctx, p.logger)
-	prunedQuery, err := mapper.Map(expr)
-	if err != nil {
-		return "", false, err
+	if p.cfg.PruneQueriesHistogram {
```
charleskorn: Rather than checking this here, what if we didn't add this middleware altogether if this is disabled? Then if there's an issue with some of the surrounding logic (ie. outside this `if`), then that is disabled as well when this feature is disabled.

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -130,6 +130,10 @@ func (cfg *Config) cardinalityBasedShardingEnabled() bool {
 	return cfg.TargetSeriesPerShard > 0
 }
 
+func (cfg *Config) isPruningQueriesEnabled() bool {
```
charleskorn, line 137: Why is this needed? Why not use `cfg.PruneQueriesHistogram` directly?

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+
+	"github.com/prometheus/prometheus/promql/parser"
+
+	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
+)
+
+// ReorderHistogramAgg optimizes queries by reordering histogram functions and aggregations
+// for more efficient execution.
+type ReorderHistogramAgg struct{}
```
charleskorn: [nit] I would use the full word "Aggregation" here, rather than abbreviating it - we use the full term everywhere else in MQE.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type reorderHistogramAgg struct {
+	ctx context.Context
+}
+
+func (mapper *reorderHistogramAgg) MapExpr(expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
+	if err := mapper.ctx.Err(); err != nil {
+		return nil, false, err
+	}
```
charleskorn: Is this used to abort mapping if the query is cancelled? I'm wondering if this is something that could be moved into the `ASTMapper` type so all mappers behave the same way. (I would do this as a separate PR and merge it before this one, to keep this PR as small as possible.)

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
```diff
@@ -0,0 +1,78 @@
...
+// for more efficient execution.
+type ReorderHistogramAgg struct{}
+
+func (r *ReorderHistogramAgg) Name() string {
+	return "Histogram aggregation reordering"
+}
+
+func (r *ReorderHistogramAgg) Apply(ctx context.Context, expr parser.Expr) (parser.Expr, error) {
+	ASTExprMapper := NewMapperReorderHistogramAgg(ctx)
+	return ASTExprMapper.Map(expr)
```
charleskorn: It'd be nice if we could create the mapper once and reuse it for all requests - I think if we make the change I suggested below for the context, then this becomes possible.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg_test.go
```diff
@@ -0,0 +1,54 @@
...
+		`histogram_count(avg(foo))`:           `avg(histogram_count(foo))`,
+		`avg(histogram_count(foo))`:           `avg(histogram_count(foo))`,
+		`histogram_count(sum(rate(foo[5m])))`: `sum(histogram_count(rate(foo[5m])))`,
+		`sum(histogram_count(rate(foo[5m])))`: `sum(histogram_count(rate(foo[5m])))`,
+		`histogram_count(avg(rate(foo[5m])))`: `avg(histogram_count(rate(foo[5m])))`,
+		`avg(histogram_count(rate(foo[5m])))`: `avg(histogram_count(rate(foo[5m])))`,
+		`(((histogram_sum(sum(foo)))))`:       `(((sum(histogram_sum(foo)))))`,
+		`histogram_sum(sum(foo+bar))`:         `sum(histogram_sum(foo+bar))`,
+		`histogram_sum(sum(foo)+sum(bar))`:    `histogram_sum(sum(foo)+sum(bar))`,
+		"histogram_sum(sum by (job) (foo))":   "sum by (job) (histogram_sum(foo))",
```
charleskorn: Would be good to add a test with `without`, to make sure that's not lost when we swap the order around.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
charleskorn: I'd completely forgotten about that, thanks for the reminder. I think your suggestion to remove the test cases for the old style of queries makes sense, and I think it also makes sense to add a comment clarifying that the goal of this is to optimise a very specific query pattern and leave the functionality as-is.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -0,0 +1,84 @@
...
+		`((((-1 * -Inf) < avg(rate(foo[4m]))) unless avg(rate(foo[3m]))) and avg(rate(foo[2m]))) or avg(rate(foo[1m]))`: `((((-1 * -Inf) < avg(rate(foo[4m]))) unless avg(rate(foo[3m]))) and avg(rate(foo[2m]))) or avg(rate(foo[1m]))`,
+		`avg(rate(foo[1m])) or avg(rate(bar[1m]))`:                                                                      `avg(rate(foo[1m])) or avg(rate(bar[1m]))`,
+		// The const expression is on the wrong side.
+		`(vector(0) == 1) and on() (avg(rate(foo[1m])))`: `(vector(0) == 1) and on() (avg(rate(foo[1m])))`,
+		// Matching on labels.
+		`(avg(rate(foo[1m]))) and on(id) (vector(0) == 1)`: `(avg(rate(foo[1m]))) and on(id) (vector(0) == 1)`,
+		// Not "on" expression.
+		`(avg(rate(foo[1m]))) and ignoring() (vector(0) == 1)`: `(avg(rate(foo[1m]))) and ignoring() (vector(0) == 1)`,
+		// Pruned expressions.
+		`(avg(rate(foo[1m]))) and on() (vector(0) == 1)`:     `(vector(0) == 1)`,
```
charleskorn: Something that could be interesting to try in a follow-up PR (or as a follow-up issue to pick up later): what if we replaced these expressions we know won't evaluate to anything with a special function (`empty()` perhaps?)? 

That might help simplify the logic in this optimisation pass: we could first go through the AST and replace all conditions that can't evaluate to anything (eg. `vector(3) == 4.5`) with `empty()`, and then go through and simplify the `and`, `or` and `unless` expressions (eg. `foo and empty()` becomes `empty()`, `foo or empty()` becomes `foo`, `foo unless empty()` becomes `foo`).

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -1,33 +1,36 @@
...
-	logger log.Logger
+func (p *PruneToggles) Name() string {
+	return "Toggled off expressions pruning"
 }
 
-func newQueryPruner(ctx context.Context, logger log.Logger) *queryPruner {
-	return &queryPruner{
-		ctx:    ctx,
-		logger: logger,
+func (p *PruneToggles) Apply(ctx context.Context, expr parser.Expr) (parser.Expr, error) {
```
charleskorn, line 33: Same comment as for the reordering pass: it'd be good if we can avoid creating the mapper for every query if we can.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, false, err
+	}
+
+	e, ok := expr.(*parser.BinaryExpr)
+	if !ok {
+		return expr, false, nil
+	}
+
+	_, _, boolResult := mapper.propagateMatchersInBinaryExpr(e)
+	return e, boolResult, nil
```
charleskorn: It's not clear to me what `boolResult` here represents - is there another name that would make this clearer?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+type vectorSelectorWrapper struct {
+	vs        *parser.VectorSelector
+	labelsSet map[string]struct{}
+	whitelist bool
+}
+
+func (mapper *propagateMatchers) propagateMatchersInBinaryExpr(e *parser.BinaryExpr) ([]*vectorSelectorWrapper, []*labels.Matcher, bool) {
+	if e.Op == parser.LOR || e.Op == parser.LUNLESS {
```
charleskorn: Why not `unless`? If the query is something like `foo{env="prod"} unless bar`, then we can propagate the `env` matcher from the LHS to the RHS. We can't propagate matchers in the other direction though.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,188 @@
...
+	e, ok := expr.(*parser.BinaryExpr)
+	if !ok {
+		return expr, false, nil
+	}
+
+	_, _, boolResult := mapper.propagateMatchersInBinaryExpr(e)
+	return e, boolResult, nil
+}
+
+type vectorSelectorWrapper struct {
```
charleskorn: It's not clear to me what this represents - is there another name that would make this clearer?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	for _, vsR := range vssR {
+		vsR.vs.LabelMatchers = combineMatchers(vsR.vs.LabelMatchers, newMatchersR, vsR.labelsSet, vsR.whitelist)
+	}
+	vss := append(vssL, vssR...)
+	matchers := combineMatchers(newMatchersL, newMatchersR, map[string]struct{}{}, false)
+	return vss, matchers, true
+}
+
+func (mapper *propagateMatchers) extractVectorSelectors(expr parser.Expr) ([]*vectorSelectorWrapper, []*labels.Matcher, bool) {
```
charleskorn: It'd be good to explain what each of the return parameters here represents.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
zenador: Is https://github.com/grafana/mimir/pull/12238 what you meant? This is also what you mean by "make the change I suggested below for the context" right?

charleskorn: >This is also what you mean by "make the change I suggested below for the context" right?

Yep, #12238 is what I had in mind for that comment.

>Is https://github.com/grafana/mimir/pull/12238 what you meant?

That covers passing the context to `Map` so we don't have to create a new mapper for every request, but I was also imagining we'd move this detection of context cancellation into the shared AST mapping types (eg. `ASTExprMapper`) so implementations of `ExprMapper` don't need to do this themselves.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Added comment

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Can't think of any sorry, added a comment to explain its purpose for now.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
zenador: Moved CollapseConstants first. The original logic for query pruning involved a lot of collapsing constants so this would have helped then, but it looks like the new one doesn't really so nothing much to cut down there. Still, it's nice to have CollapseConstants put first so that future logic for any passes can rely on it.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
zenador: These cases should be covered now. Allowed some functions to start with, we can make them more thorough in the separate PRs or follow-up.

pkg/frontend/querymiddleware/prune.go
zenador: Hmm the `Map` signature doesn't have room to indicate a change, and it seems too much to change the signature everywhere just for this. If we recreated the mapper each time then we could store the flag in the struct, but now that we want to keep reusing the mapper, I'm not sure how to do this.

pkg/frontend/querymiddleware/rewrite.go
```diff
@@ -0,0 +1,81 @@
...
+	apierror "github.com/grafana/mimir/pkg/api/error"
+	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
+	"github.com/grafana/mimir/pkg/streamingpromql/optimize/ast"
+	"github.com/grafana/mimir/pkg/util/spanlogger"
+)
+
+type rewriteMiddleware struct {
+	next            MetricsQueryHandler
+	logger          log.Logger
+	cfg             Config
```
charleskorn: Is the config used anywhere? I think this can be removed.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -44,6 +44,10 @@ func (mapper *propagateMatchers) MapExpr(expr parser.Expr) (mapped parser.Expr,
 	return e, boolResult, nil
 }
 
+// vectorSelectorWrapper is a struct to hold additional information about vector selectors,
+// mainly the set of matchers that are allowed to propagate inwards and outwards, and whether
```
charleskorn: `the set of matchers that are allowed to propagate inwards and outwards`: there's only one set of labels in the struct, which does it represent?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// vectorSelectorWrapper is a struct to hold additional information about vector selectors,
+// mainly the set of matchers that are allowed to propagate inwards and outwards, and whether
+// that set is a whitelist or blacklist, which are used only for aggregate expressions and the
+// expressions that contain them.
+type vectorSelectorWrapper struct {
+	vs          *parser.VectorSelector
+	labelsSet   map[string]struct{}
+	whitelist   bool
```
charleskorn: We try not to use the terms "whitelist" and "blacklist" - are there other terms we could use instead? (include/exclude perhaps?)

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -84,6 +88,9 @@ func (mapper *propagateMatchers) propagateMatchersInBinaryExpr(e *parser.BinaryE
 	return vss, matchers, true
 }
 
+// extractVectorSelectors returns the vector selectors found in the expression (wrapped with additional
+// info, see vectorSelectorWrapper), along with the label matchers associated with them, and a boolean
+// indicating whether any vector selectors were found.
```
charleskorn: Do we need the boolean? If no vector selectors are found, won't the slice of `vectorSelectorWrapper` be empty?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
charleskorn: `enrichedVectorSelector` or `vectorSelectorInfo` perhaps?

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -10,7 +10,9 @@ import (
 	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
 )
 
-// PruneToggles optimizes queries by pruning expressions that are toggled off.
+// PruneToggles optimizes queries by pruning expressions that are toggled off, only
+// targeting a very specific query pattern that is commonly used to toggle between
+// classic and native histograms in our dashboards.
```
charleskorn, line 15: [nit] Would be good to include an example of that pattern here.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg.go
charleskorn, line 1: Could you please rename this file (and the test file) to match the struct's name?

pkg/streamingpromql/optimize/ast/reorder_histogram_agg_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReorderHistogramAgg(t *testing.T) {
```
charleskorn: It'd be good to verify that the expressions produce equivalent results - what if we evaluated the optimised and unoptimised expressions and confirmed they produce the same results?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -75,10 +75,16 @@ func (mapper *propagateMatchers) propagateMatchersInBinaryExpr(e *parser.BinaryE
 	}
 
 	matchingLabelsSet := makeStringSet(e.VectorMatching.MatchingLabels)
-	newMatchersL := mapper.getMatchersToPropagate(matchersR, matchingLabelsSet, e.VectorMatching.On)
 	newMatchersR := mapper.getMatchersToPropagate(matchersL, matchingLabelsSet, e.VectorMatching.On)
-	for _, vsL := range vssL {
-		vsL.vs.LabelMatchers = combineMatchers(vsL.vs.LabelMatchers, newMatchersL, vsL.labelsSet, vsL.whitelist)
+	var newMatchersL []*labels.Matcher
+	if e.Op == parser.LUNLESS {
+		// For LUNLESS, we do not propagate matchers from the right-hand side to the left-hand side.
```
charleskorn, line 90: Might be good to add an example here explaining why we can't do this.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -133,9 +133,36 @@ func (mapper *propagateMatchers) extractVectorSelectors(expr parser.Expr) ([]*ve
...
+	// If the function is eligible for matcher propagation, return the index of the argument
+	// that contains the vector selector. Otherwise, return -1.
+	switch funcName {
+	// Mathematical
+	case "abs", "sgn", "sqrt", "exp", "deriv", "ln", "log2", "log10", "ceil", "floor", "round", "clamp", "clamp_min", "clamp_max":
+		return 0
+	// Trigonometric
+	case "acos", "acosh", "asin", "asinh", "atan", "atanh", "cos", "cosh", "sin", "sinh", "tan", "tanh", "deg", "rad":
+		return 0
+	// Comparisons
```
charleskorn: [nit] The functions below aren't comparisons, and won't contain vector selectors - so this comment and the one above the `switch` aren't quite right.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPropagateMatchers(t *testing.T) {
+	testCases := map[string]string{
```
charleskorn: Could you please add some test cases that include functions over range vector selectors (eg. `rate(foo{env="prod"}[5m]) + rate(bar[10m])`, `rate(foo{env="prod"}) + bar`)?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,278 @@
...
+			vs.whitelist = true
+		}
+		return
+	}
+	vs.labelsSet = groupingSet
+	vs.whitelist = whitelist
+	vs.containsAgg = true
+}
+
+func getOverlap(set1, set2 map[string]struct{}) map[string]struct{} {
```
charleskorn: [nit] In mathematics, the term for this would be "intersection"

charleskorn (commented):
I'll review this again once https://github.com/grafana/mimir/pull/12238 is merged (the volume of changes here is making things hard to review), but this is looking good to me.

pkg/frontend/querymiddleware/prune.go
charleskorn: In that case, I think it's worth creating a new mapper each time - formatting expressions as strings is expensive, and I'd prefer to avoid this if we can.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Hmm I thought I'd differentiate between the case where there's an inner aggregation but no by/without, vs no inner aggregation, but I suppose that can be simplified away.

charleskorn: Ah, then maybe the comment here is out of date - the comment says the boolean indicates if any vector selectors were found.

zenador: Oops sorry I mixed this up, I thought you were commenting on `containsAgg bool` in `vectorSelectorWrapper`, do you think we should try to remove that?

As for this one, I can try and remove it. Can't think of any case offhand where it might be a problem to simplify it, but not sure if there's some funny edge case we're missing.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: The same set would apply to both inwards and outwards right? I've updated the comment to make that clearer.

pkg/frontend/querymiddleware/prune.go
zenador: Just wondering, would it be better if we just don't log it then, since we'd need to format them to print to log? We don't log our existing optimization passes anyway right?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
charleskorn: > Oops sorry I mixed this up, I thought you were commenting on `containsAgg bool` in `vectorSelectorWrapper`, do you think we should try to remove that?

If we need it, let's keep it. My original comment was on the return value from this function.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
charleskorn: Ah, I misunderstood the comment - thanks for clarifying it.

pkg/frontend/querymiddleware/prune.go
charleskorn: I think it's OK to pay the cost of stringifying it to log the expression if it has changed, but it seems unnecessary to do it just to check if we made any changes given we can figure that out much more cheaply.

pkg/streamingpromql/optimize/ast/reorder_histogram_agg_test.go
zenador: Added tests on sample data in `mimir/pkg/frontend/querymiddleware/ast_optimization_pass_test.go`

zenador:
Split into:
- https://github.com/grafana/mimir/pull/12303
- https://github.com/grafana/mimir/pull/12304
- https://github.com/grafana/mimir/pull/12305

PR #11977, opened 2025-07-04
write path: handle created timestamp

### LINKED ISSUE #11976: Idea: handle created timestamp field in Remote-Write (PRW) 2.0

Leftover from #11100 .
Related to #9072 .

Mimir should handle the created timestamp field in Remote Write 2.0 according to spec.

If the value of the created timestamp field in RW2.0 is not zero, we should inject a zero sample (float or float/integer native histogram) at the created timestamp.

Out of order created timestamp is ignored without error, as well as created timestamp in the future.
Duplicate sample value is handled as with any other sample. Exact match ok, diff is an error.

Created timestamps were defined here: https://github.com/prometheus/proposals/blob/main/proposals/0029-created-timestamp.md

Remote Write 2.0 spec: https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/

Handling created timestamps for native histograms: https://prometheus.io/docs/specs/native_histograms/#created-timestamp-handling

We already have an alternative with the [silent zero approach](https://github.com/grafana/mimir-prometheus/blob/d3be9d0e3cf53e9a67bed0710eac1ef18b5a99ae/storage/remote/otlptranslator/prometheusremotewrite/helper.go#L605) introduced in https://github.com/grafana/mimir-prometheus/pull/689.

However it is a custom solution that has some problems and isn't upstreamed.

_No response_

Small (<= 1 month dev)

_No response_

### Proposer?

_No response_

krajorama:
#### What this PR does

Modify the ingester and block builder to accept and pass the Created timestamp field from incoming write requests to TSDB.

Note1: if samples aren't ordered by time in a remote write requests, then the result is undefined.

Note2: as with regular samples, Mimir will have non-deterministic behavior if there's a sample and a created timestamp with the same timestamp ingested on two different ingesters for whatever reason. During query and block building the resolution of such conflicting samples has always been non-deterministic, as there is no way to know at that point which sample is the correct one.

Fixes #11976

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/integration/distributor_test.go
+++ b/integration/distributor_test.go
@@ -148,17 +148,17 @@ func testDistributorWithCachingUnmarshalData(t *testing.T, cachingUnmarshalDataE
 				{
 					Timeseries: []promRW2.TimeSeries{
 						{
-							LabelsRefs: []uint32{0, 1},
+							LabelsRefs: []uint32{1, 2},
 							Samples:    []promRW2.Sample{{Timestamp: queryStart.Add(1 * time.Second).UnixMilli(), Value: 100}},
 							Metadata: promRW2.Metadata{
 								Type:    promRW2.Metadata_METRIC_TYPE_COUNTER,
-								HelpRef: 2,
-								UnitRef: 3,
+								HelpRef: 3,
+								UnitRef: 4,
 							},
 							CreatedTimestamp: queryStart.UnixMilli(),
 						},
 					},
-					Symbols: []string{"__name__", "foobarC_CT_total", "some helpC_CT", "someunitC_CT"},
+					Symbols: []string{"", "__name__", "foobarC_CT_total", "some helpC_CT", "someunitC_CT"},
 				},
 			},
 			queries: map[string]model.Matrix{


CHANGELOG.md
```diff
@@ -31,7 +31,8 @@
...
 * [CHANGE] Query-frontend: Apply query pruning and check for disabled experimental functions earlier in query processing. #11939
-* [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. Limitations: Created timestamp is ignored, per series metadata is merged on metric family level automatically, ingestion might fail if client sends ProtoBuf fields out of order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0` depending on the detected Remote-Write protocol. #11100 #11101 #11192 #11143
+* [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. Limitations: if samples in a write request are not ordered by time, the created timestamp might be dropped, per series metadata is merged on metric family level automatically, ingestion might fail if client sends ProtoBuf fields out of order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0` depending on the detected Remote-Write protocol. #11100 #11101 #11192 #11143 #11977
```
```suggestion
* [FEATURE] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol. This feature includes some limitations. If samples in a write request aren't ordered by time, the created timestamp might be dropped. Additionally, per-series metadata is automatically merged on the metric family level, and ingestion might fail if the client sends ProtoBuf fields out-of-order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0`, depending on the detected remote-write protocol. #11100 #11101 #11192 #11143 #11977
```

tacole02: I tried to break this up a bit for clarity. Let me know if the meaning is preserved.

tacole02: Also, I changed the second occurrence of `remote-write` to lower-case, since we aren't reference a specific remote-write protocol in this case.

CHANGELOG.md
```diff
@@ -31,7 +31,8 @@
...
+* [FEATURE] Ingester/Block-builder: handle the Created timestamp field of write requests. #11977
```
```suggestion
* [FEATURE] Ingester/Block-builder: Handle the timestamp field for write requests. #11977
```

tacole02 (approved):

pkg/blockbuilder/tsdb.go
```diff
@@ -139,7 +139,27 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record, lastBlockMax
 		// and NOT the stable hashing because that's what TSDB expects. We don't need stable hashing in block builder.
 		ref, copiedLabels := app.GetRef(nonCopiedLabels, hash)
 
+		ingestCreatedTimestamp := ts.CreatedTimestamp > 0
+
 		for _, s := range ts.Samples {
+			if ingestCreatedTimestamp && ts.CreatedTimestamp < s.TimestampMs &&
+				(processEverything || !recordAlreadyProcessed || ts.CreatedTimestamp >= lastBlockMax) &&
+				(processEverything || ts.CreatedTimestamp < blockMax) &&
+				(!nativeHistogramsIngestionEnabled || len(ts.Histograms) == 0 || ts.Histograms[0].Timestamp >= s.TimestampMs) {
```
narqo: Sorry, I didn't know you were working on expanding the tsdb parts at the moment. I've recently merged a refactoring to the blockbuilder, that got rig of "legacy" `blockMax`, `processEverything`, etc. But now you have to resolve the conflicts here.

Let me know, if I can share any extra details from why/what was changed in https://github.com/grafana/mimir/pull/11987

krajorama: Does that mean I don't have to care about block boundaries anymore? Meaning I just process created timestamp like in the ingester where block boundaries don't matter .

krajorama: I've simplified the logic and test. I did leave some of the structure of the builder test because it gives pretty good coverage of case and I've done a lot of work to have it :)

krajorama:
@tacole02 Thank you for the suggested changes, I've made a small tweak on top. PTAL if you have time, thanks.

jesusvazquez (approved):
LGTM the test coverage is great so I think we're good.

I was wondering if it would be possible to have less code duplication between how the ingester does the CT appends and how the blockbuilder does it but i dont think we can do it without making this PR much larger and even then I don't know if it would be worth it.

Great work Krajo

pkg/blockbuilder/tsdb_test.go
```diff
@@ -627,3 +628,314 @@ func defaultLimitsTestConfig() validation.Limits {
...
+				{TS: lastEnd + 200, Val: 0},
+				{TS: lastEnd + 300, Val: 4},
+				{TS: lastEnd + 400, Val: 5},
+				{TS: lastEnd + 500, Val: 6},
+				{TS: lastEnd + 600, Val: 7},
+				{TS: lastEnd + 700, Val: 0},
+				{TS: lastEnd + 800, Val: 8},
+				{TS: currEnd - 200, Val: 9},
+				{TS: currEnd - 100, Val: 0},
+				{TS: currEnd + 200, Val: 10},
```
narqo, line 771: To confirm I understand the test: all samples with `Val: 0` here are from the created timestamps, right?

pkg/blockbuilder/tsdb.go
```diff
@@ -132,7 +132,25 @@ func (b *TSDBBuilder) Process(ctx context.Context, rec *kgo.Record) (err error)
...
+					// If the cached reference exists, we try to use it.
+					_, err = app.AppendCTZeroSample(ref, copiedLabels, s.TimestampMs, ts.CreatedTimestamp)
+				} else {
+					// Copy the label set because TSDB may retain it.
+					copiedLabels = mimirpb.CopyLabels(nonCopiedLabels)
+					ref, err = app.AppendCTZeroSample(0, copiedLabels, s.TimestampMs, ts.CreatedTimestamp)
+				}
+				if err != nil && !errors.Is(err, storage.ErrOutOfOrderCT) && !errors.Is(err, storage.ErrOutOfOrderSample) {
+					level.Warn(b.logger).Log("msg", "failed to store zero float sample for created timestamp", "tenant", userID, "err", err)
+				}
```
narqo: I've noticed that the code in ingester, handles this error as "soft-error", ie. `errProcessor.ProcessErr(err, ...)`. I believe the result of that is that the sample will be counted towards discarded samples metric.

This is a nit, but do we want to count a discarded created sample in the block-builder as well?

narqo: Also, I've noticed that the logic about skipping the sample when its created timestamp fails is different between the ingester and the block-builder (ref the comment below).

pkg/ingester/ingester.go
```diff
@@ -1556,6 +1558,23 @@ func (i *Ingester) pushSamplesToAppender(userID string, timeseries []mimirpb.Pre
...
+				} else {
+					// Copy the label set because both TSDB and the active series tracker may retain it.
+					copiedLabels = mimirpb.CopyLabels(nonCopiedLabels)
+					ref, err = app.AppendCTZeroSample(0, copiedLabels, s.TimestampMs, ts.CreatedTimestamp)
+				}
+				if err == nil {
+					stats.succeededSamplesCount++
+				} else if !errors.Is(err, storage.ErrOutOfOrderCT) && !errors.Is(err, storage.ErrOutOfOrderSample) {
+					errProcessor.ProcessErr(err, ts.CreatedTimestamp, ts.Labels)
+					continue
```
narqo: To double-check, the `continue` here will cause the sample `s` to be skipped, if there was an error in the append of its created sample. I think, the piece of code in the block-builder doesn't skip the sample. Is this on purpose?

narqo:
> I was wondering if it would be possible to have less code duplication between how the ingester does the CT appends and how the blockbuilder does it but i dont think we can do it without making this PR much larger and even then I don't know if it would be worth it.

I did a couple of attempts trying to unify these pieces of the ingester and the block-builder, but that, indeed, resulted in a fairly large refactoring of ingester's internals, which was a bit scary. So every time, I gave up. I still want to do that. Although, tbf, an extra doubt I have currently, comes from the parquet exploration, that folks do in parallel. If parquet is the way, I'm not sure the block-builder will keep building tsdb blocks at all.

pkg/blockbuilder/tsdb_test.go
krajorama: yes exactly

pkg/ingester/ingester.go
krajorama: no, I think that's a fair point, shouldn't skip

krajorama: I've copy pasted the error handling, I'll try to add a unit test to cover this case.

krajorama: I've modified the code and added tests, but the native histogram one fails as there's a bug in upstream Prometheus. So I need to fix the bug there, and update mimir-prometheus.

pkg/blockbuilder/tsdb.go
krajorama: > This is a nit, but do we want to count a discarded created sample in the block-builder as well?

Just add `discardedSamples++` ?

narqo: Yes, just `discardedSamples` will do that

pkg/ingester/ingester.go
krajorama: Actually the bug is not new https://github.com/prometheus/prometheus/pull/16332

pkg/blockbuilder/tsdb.go
krajorama: done, added test as well

tacole02 (approved):
Looks great! Thanks, @krajorama !

zenador:

krajorama:
> Looks great! Thanks, @krajorama !

Thanks @tacole02 

Unfortunately I  have to drop the changes to the existing CHANGELOG entry as the release has been cut in the meantime :( So I'm moving the text to the unreleased section as enhancement.

PR #11980, opened 2025-07-04
Parquet converter: Implement priority queue for block conversion

jesusvazquez:
This PR redesigns the parquet converter to use a non-blocking priority queue that prioritises recently uploaded blocks for conversion.

* Priority Queue Implementation:
  - Replaces blocking nested loops with a thread-safe priority queue using container/heap
  - Blocks are prioritized by ULID timestamp, ensuring older blocks are processed first
* Separate block discovery:
  - There is a new discovery goroutine that periodically discovers users and blocks, enqueuing them for processing
    - If the block was previously processed it will be marked as converted and skipped the next time its discovered.
    - There is a new configuration flag `parquet-converter.max-block-age` that allows us to have a rolling window of blocks so we dont queue up all the work at once. We can set this to 30 days and only blocks up to 30 days old will be converted, when the work is completed we can go and increase that window again.
  - There is a new processing goroutine that continuously consumes from the priority queue and converts blocks
  - Main Loop remains responsive and handles only service lifecycle events
* New metrics
  - Since we added a priority queue, I added 5 new metrics for queue monitoring:
    - cortex_parquet_converter_queue_size - Current queue depth
    - cortex_parquet_converter_queue_wait_time_seconds - Time blocks spend queued
    - cortex_parquet_converter_queue_items_enqueued_total - Total blocks enqueued
    - cortex_parquet_converter_queue_items_processed_total - Total blocks processed
    - cortex_parquet_converter_queue_items_dropped_total - Total blocks dropped when queue closed

The idea here is that by looking at the queue metrics we can have an idea of how much scaling up we need to deal with the pending work. Also, before this PR we had no idea of how much work was left to be done but now we will.

### DIFF
--- a/pkg/parquetconverter/parquet_converter.go
+++ b/pkg/parquetconverter/parquet_converter.go
@@ -85,6 +85,7 @@ type Config struct {
 
 	ConversionInterval time.Duration `yaml:"conversion_interval"`
 	DiscoveryInterval  time.Duration `yaml:"discovery_interval"`
+	MaxBlockAge        time.Duration `yaml:"max_block_age"`
 
 	ShardingRing RingConfig `yaml:"sharding_ring"`
 }
@@ -97,6 +98,7 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
 	f.Var(&cfg.DisabledTenants, "parquet-converter.disabled-tenants", "Comma-separated list of tenants that cannot have their TSDB blocks converted into Parquet. If specified, and the Parquet-converter would normally pick a given tenant to convert the blocks to Parquet (via -parquet-converter.enabled-tenants or sharding), it is ignored instead.")
 	f.DurationVar(&cfg.ConversionInterval, "parquet-converter.conversion-interval", time.Minute, "The frequency at which the conversion runs.")
 	f.DurationVar(&cfg.DiscoveryInterval, "parquet-converter.discovery-interval", 5*time.Minute, "The frequency at which user and block discovery runs.")
+	f.DurationVar(&cfg.MaxBlockAge, "parquet-converter.max-block-age", 0, "Maximum age of blocks to convert. Blocks older than this will be skipped. Set to 0 to disable age filtering.")
 	f.StringVar(&cfg.DataDir, "parquet-converter.data-dir", "./data-parquet-converter/", "Directory to temporarily store blocks during conversion. This directory is not required to persist between restarts.")
 }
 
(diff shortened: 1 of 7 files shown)


pkg/parquetconverter/parquet_converter.go
```diff
@@ -83,6 +84,8 @@ type Config struct {
 	DataDir string `yaml:"data_dir"`
 
 	ConversionInterval time.Duration `yaml:"conversion_interval"`
```
npazosmendez, line 86: We are not using `ConversionInterval` anymore I think, we can delete the field and flag.

pkg/parquetconverter/parquet_converter.go
```diff
@@ -115,6 +120,10 @@ type ParquetConverter struct {
 
 	blockConverter blockConverter
 	metrics        parquetConverterMetrics
+
+	conversionQueue *priorityQueue
+
+	queuedBlocks sync.Map // map[ulid.ULID]time.Time - persistent cache of blocks that have been queued
```
npazosmendez, line 126: nit: i think this docstring was from a previous version
```suggestion
	queuedBlocks sync.Map // map[ulid.ULID]time.Time - blocks that are in the priority queue or being processed
```

pkg/parquetconverter/parquet_converter.go
```diff
@@ -236,65 +246,147 @@ func newRingAndLifecycler(cfg RingConfig, logger log.Logger, reg prometheus.Regi
...
-						continue
-					}
-
-					c.processBlock(ctx, u, m, uBucket, ulogger)
-				}
+			if _, alreadyQueued := c.queuedBlocks.Load(m.ULID); alreadyQueued {
+				continue
 			}
+
+			mark, err := ReadConversionMark(ctx, m.ULID, uBucket, ulogger)
```
npazosmendez, line 368: We could have a cache of blocks that are known to be already converted, to avoid the constant bucket storage calls

pkg/parquetconverter/metrics.go
```diff
@@ -42,5 +47,33 @@ func newParquetConverterMetrics(reg prometheus.Registerer) parquetConverterMetri
...
+
+	m.queueItemsProcessed = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "cortex_parquet_converter_queue_items_processed_total",
+		Help: "Total number of blocks processed from the conversion queue",
+	}, []string{"user"})
+
+	m.queueItemsDropped = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "cortex_parquet_converter_queue_items_dropped_total",
+		Help: "Total number of blocks dropped from the conversion queue when full",
+	}, []string{"user"})
```
npazosmendez, line 76: AFAICT drops are not about about a full queue, but about the queue being closed, which happens only when the converter is shutdown. I don't think it's worth tracking this, I would delete this metric.

PR #11988, opened 2025-07-06
kafkatool: Add the `consumer-group delete-offsets` command to `kafkatool`

gotjosh:

#### What this PR does

Adds a new command to `kafkatool` - `consumer-group delete-offset` as a way to delete all the committed offsets from a consumer group.

Fixes N/A

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/tools/kafkatool/consumer_group.go
+++ b/tools/kafkatool/consumer_group.go
@@ -44,6 +44,11 @@ func (c *ConsumerGroupCommand) Register(app *kingpin.Application, getKafkaClient
 	copyOffsetCmd.Flag("to-group", "Consumer group to which the offset should be copied.").Required().StringVar(&c.copyToGroup)
 	copyOffsetCmd.Flag("topic", "Topic.").Required().StringVar(&c.topic)
 	copyOffsetCmd.Flag("partition", "Partition ID.").Required().Int32Var(&c.partitionID)
+
+	deleteOffsetCmd := cmd.Command("delete-offsets", "Deletes all the offsets for a given topic and partition.").Action(c.deleteOffsets)
+	deleteOffsetCmd.Flag("group", "Consumer group name.").Required().StringVar(&c.group)
+	deleteOffsetCmd.Flag("topic", "Topic.").Required().StringVar(&c.topic)
+	deleteOffsetCmd.Flag("partition", "Partition ID.").Required().Int32Var(&c.partitionID)
 }
 
 func (c *ConsumerGroupCommand) listOffsets(_ *kingpin.ParseContext) error {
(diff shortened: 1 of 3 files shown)


tools/kafkatool/consumer_group.go
```diff
@@ -44,6 +44,11 @@ func (c *ConsumerGroupCommand) Register(app *kingpin.Application, getKafkaClient
 	copyOffsetCmd.Flag("to-group", "Consumer group to which the offset should be copied.").Required().StringVar(&c.copyToGroup)
 	copyOffsetCmd.Flag("topic", "Topic.").Required().StringVar(&c.topic)
 	copyOffsetCmd.Flag("partition", "Partition ID.").Required().Int32Var(&c.partitionID)
+
+	deleteOffsetCmd := cmd.Command("delete-offsets", "Deletes all the offsets for a given topic and partition.").Action(c.deleteOffsets)
```
pracucci: I find it a bit misleading naming "offsets" plural. Since it takes in input group, topic and partition, the offset that will be deleted is just 1. I would rename "offsets" to "offset" in every place related to this command (flag name, flag description, code).

tools/kafkatool/consumer_group.go
```diff
@@ -90,6 +95,36 @@ func (c *ConsumerGroupCommand) copyOffset(_ *kingpin.ParseContext) error {
...
+	toDelete.Add(topic, partitionID)
+
+	results, err := adm.DeleteOffsets(context.Background(), group, toDelete)
+	if err != nil {
+		return err
+	}
+
+	err, ok := results.Lookup(topic, partitionID)
+	if !ok {
+		return fmt.Errorf("unable to determine the result of the delete offsets operations for consumer gorup %s, topic %s, and partition %d", group, topic, partitionID)
```
pracucci:
```suggestion
		return fmt.Errorf("unable to determine the result of the delete offsets operations for consumer group %s, topic %s, and partition %d", group, topic, partitionID)
```

tools/kafkatool/consumer_group.go
```diff
@@ -90,6 +95,36 @@ func (c *ConsumerGroupCommand) copyOffset(_ *kingpin.ParseContext) error {
...
+func deleteConsumerGroupOffsets(adm *kadm.Client, group, topic string, partitionID int32, printer Printer) error {
+	toDelete := kadm.TopicsSet{}
+	toDelete.Add(topic, partitionID)
+
+	results, err := adm.DeleteOffsets(context.Background(), group, toDelete)
+	if err != nil {
+		return err
+	}
+
+	err, ok := results.Lookup(topic, partitionID)
```
pracucci, line 114: I manually tested it against Warpstream and I got the error `unable to determine the result of the delete offsets operations for...` but the deletion worked. The reason why I got the error is because `results` only contains errors. If there's no error, then the lookup will find nothing, but it doesn't mean it was not successful. You should just remove the `if` branch checking for `ok`.

pracucci (approved):
Approved with a couple of things to fix, please.

tools/kafkatool/consumer_group.go
gotjosh: Fixed in https://github.com/grafana/mimir/pull/11988/commits/5f083704cd817664453af71063b300817974ec30

tools/kafkatool/consumer_group.go
gotjosh: Makes sense - fixed in https://github.com/grafana/mimir/pull/11988/commits/5f083704cd817664453af71063b300817974ec30

tools/kafkatool/consumer_group.go
gotjosh: I decided, to still check it but also assert for `err != nil` in the same statement in https://github.com/grafana/mimir/pull/11988/commits/5f083704cd817664453af71063b300817974ec30

gotjosh: I manually tested both scenarios:

```
/mimir # kafkatool --kafka-address=kafka_1:9092 consumer-group delete-offset --group=mimir-write-zone-a-1 --topic=mimir-ingest --partition=1
successfully deleted committed offset for consumer group mimir-write-zone-a-1, topic mimir-ingest and partition 1
```

and

```
/mimir # kafkatool --kafka-address=kafka_1:9092 consumer-group delete-offset --group=mimir-write-zone-a-1 --topic=mimir-ingest --partition=333
kafkatool: error: unable to delete committed offset for consumer group mimir-write-zone-a-1, topic mimir-ingest and partition 333: UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
```

PR #11989, opened 2025-07-06
MQE: fix issue where a query with a common subexpression could return incorrect results

charleskorn:
#### What this PR does

This PR fixes an issue where common subexpression elimination (CSE) can cause a query to return incorrect query results.

The issue would occur if the ring buffer used to buffer data was in a state where the head of the buffer is at the end of the slice, and the tail was at the beginning of the slice. In this case, calling `Remove` for the item at the end of the slice would return the correct value, as would the subsequent call for the item at the beginning of the slice, but all following calls would return incorrect values.

#11189 

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


charleskorn:
Backport to 2.17 isn't necessary, as the release is based on the r350 branch and this fix was already backported there.

PR #12014, opened 2025-07-07
Alertmanager: Add an endpoint to get the full state

santihernandezc:
### Description

This PR adds an endpoint through which Grafana can retrieve the Alertmanager's current state.

As the state is not immediately persisted in the database, we attempt to fetch the state from the running Alertmanager. This way we return the latest nflog entries and silences (at least from a specific Alertmanager's perspective, there's still the chance that some entries have not yet been received via gossip).

The remote Alertmanager feature can use this to roll back to _remote secondary_ mode using the latest Alertmanager state, thus avoiding duplicate notifications and missing silences.

### How to test it

You can spin up Mimir in read-write mode to test this PR.
1. Go to `development/mimir-read-write-mode`
2. In `config/mimir.yaml`, set `grafana_alertmanager_compatibility_enabled: true` under the `alertmanager` section
3. Run `./compose-up`
4. (Optional) Create some silences and alerts to populate the state
6. Get the full state using `curl http://localhost:8780/api/v1/grafana/full_state`

### DIFF
--- a/pkg/alertmanager/api_grafana.go
+++ b/pkg/alertmanager/api_grafana.go
@@ -28,7 +28,6 @@ import (
 )
 
 const (
-	errGettingFullState               = "error getting full state from the running Grafana Alertmanager"
 	errMalformedGrafanaConfigInStore  = "error unmarshalling Grafana configuration from storage"
 	errMarshallingState               = "error marshalling Grafana Alertmanager state"
 	errMarshallingStateJSON           = "error marshalling JSON Grafana Alertmanager state"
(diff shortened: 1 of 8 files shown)


pkg/alertmanager/api_grafana.go
```diff
@@ -168,6 +171,60 @@ type errorResult struct {
...
+	if err != nil {
+		level.Error(logger).Log("msg", errMarshallingState, "err", err, "user", userID)
+		w.WriteHeader(http.StatusInternalServerError)
+		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errMarshallingState, err.Error())})
+		return
+	}
+
+	util.WriteJSONResponse(w, successResult{
+		Status: statusSuccess,
+		Data:   &UserGrafanaState{State: base64.StdEncoding.EncodeToString(bytes)},
```
santihernandezc: Not really a "Grafana" state, but there's no need to create a separate struct for this.
I'll make a follow-up PR to improve state structs. There's no need to have a Grafana-specific state, and the `Promoted` field is of no use.

pkg/alertmanager/api_grafana.go
```diff
@@ -168,6 +171,60 @@ type errorResult struct {
...
+		level.Error(logger).Log("msg", errNoOrgID, "err", err.Error())
+		w.WriteHeader(http.StatusUnauthorized)
+		util.WriteJSONResponse(w, errorResult{Status: statusError, Error: fmt.Sprintf("%s: %s", errNoOrgID, err.Error())})
+		return
+	}
+
+	// Try to get the full state from the running Alertmanager, using the database as a fallback.
+	am.alertmanagersMtx.Lock()
+	userAM, ok := am.alertmanagers[userID]
+	am.alertmanagersMtx.Unlock()
```
yuri-tceretian: I wonder if safer approach would be to keep the lock until we're done reading the state because once we release it there is a chance the Alertmanager gets stopped, which might do some mangling with the state.

stevesg (commented):
The problem here is that the replica might not have the requesting user assigned to it - we would need to contact the correct replica.

The way this is implemented, you're at the mercy of which replica the load balancer picks, as to whether you get in-memory state, or stale state from object storage.

I'd prefer it to be deterministic and either:
- Always fetch state from memory, routing to correct replica
- Always fetch from object storage, accept the downsides

santihernandezc:
@stevesg I just updated the PR to always fetch the state from the running Alertmanager. The distributor should route the request to a replica that owns the tenant.

fayzal-g (approved):

PR #12024, opened 2025-07-08
Update IngesterHasNotShippedBlocks to be block-builder aware

narqo:
#### What this PR does

In #11872 we added the `cortex_blockbuilder_tsdb_last_successful_compact_and_upload_timestamp_seconds` metric, that the block-builder exposes to track its shipment of blocks.

In this PR we update the ingester alerts to not trigger if the mentioned metric exists. This is so we could safely stop the ingesters from uploading the blocks to the object store, while the block-builder is doing that.

Also, the PR updates related runbooks. It removes the references to alerts and metrics of the legacy stand-alone mode, and adds the new one.

### DIFF
--- a/docs/sources/mimir/manage/mimir-runbooks/_index.md
+++ b/docs/sources/mimir/manage/mimir-runbooks/_index.md
@@ -1757,55 +1757,23 @@ How to **fix**:
 
   1. Once ingesters are stable, revert the temporarily config applied in the previous step.
 
-### MimirBlockBuilderNoCycleProcessing
-
-This alert fires when the block-builder stops reporting any processed cycles for an unexpectedly long time.
-
-How it **works**:
-
-- The block-builder periodically consumes a portion of the backlog from Kafka partition, and processes the consumed data into TSDB blocks. The block-builder calls these periods "cycles".
-- If the block-builder doesn't process any cycles for an extended period of time, this could indicate that a block-builder instance is stuck and cannot complete cycle processing.
-
-How to **investigate**:
-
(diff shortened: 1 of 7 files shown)


docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -1757,69 +1757,61 @@ How to **fix**:
 
   1. Once ingesters are stable, revert the temporarily config applied in the previous step.
 
-### MimirBlockBuilderNoCycleProcessing
+### MimirBlockBuilderLagging
```
tcard, line 1760: **question:** I see this alert is based in metrics published by block-builder itself, rather than metrics produced by Kafka (or Warpstream). This means that if block-builder is in trouble, we won't get this alert. Couldn't we have used metrics from the origin instead?

tcard (approved):

docs/sources/mimir/manage/mimir-runbooks/_index.md
narqo: > [..] Couldn't we have used metrics from the origin instead?

We could but that requires to have these per-consumer-group and per-partition metrics unified in an origin. That is, iirc, Kafka ecosystem has `kafka-exporter`, that emits `kafka_consumergroup_lag`. But internally we don't use kafka-exporter with WS. And WS itself doesn't have them either.

Note: historically, I wasn't a fan of tracking the lag from inside the block-builder, because this duplicated what `kafka-exporter` did. I think it's fine now.

Issue #11404, opened 2025-05-06
Bug: mimirtool remote-read export http: read on closed response body most of the time


pierremahot:
### What is the bug?

got issue when using mimirtool remote-read export fail with strange close reponse body error
it work sometime but depending on the timewindows it fail without reason.
all data look available with gafana querying the time window.

```
go run cmd/mimirtool/main.go remote-read export --selector '{__name__="xxx"}' --tsdb-path ./local-tsdb 
INFO[0000] Created remote read client using endpoint 'https://mimir.xxx.xx/prometheus/api/v1/read' 
INFO[0000] Using existing TSDB in path './local-tsdb' 
INFO[0000] Querying time from=2025-05-01T12:00:00+02:00 to=2025-05-01T13:00:00+02:00 with selector={__name__="xxxx"} 
INFO[0000] Store TSDB blocks in './local-tsdb' 
INFO[0000] BLOCK ULID                  MIN TIME                       MAX TIME                       DURATION     NUM SAMPLES  NUM CHUNKS   NUM SERIES   SIZE 
INFO[0000] 01JTK6GFCKDHPSHW1JAB8C8VN1  2025-05-01 10:00:13 +0000 UTC  2025-05-01 10:59:30 +0000 UTC  59m16.773s   4500         75           75           14KiB571B 
main: error: process blocks: input: http: read on closed response body, try --help
exit status 1
```

### How to reproduce it?

may be linked to data don't know how to reproduce. 

### What did you think would happen?

no error

### What was your environment?

 mimir-distributed-5.7.0
mimir 2.16.0
mimirtool 2.16.1

pierremahot:
New findings it fail when i query odd hours on my mimir.
Can only query pair hours on my mimir.

pracucci:
> New findings it fail when i query odd hours on my mimir.

This looks odd to me. If you look at Mimir logs, do you see any detail about why the request is failing?

pierremahot:
when testing I only see success my request with 200 OK on the mimi gateway and query stats with status success on the mimir frontend component.

no error on mimir side.

```
10.0.2.122 - mimir [12/May/2025:07:45:37 +0000]  200 "POST /prometheus/api/v1/read HTTP/1.1" 37652 "-" "mimirtool/2.16.0 (version=2.16.0, branch=release-2.16, revision=b4f36dac3a)" "88.172.71.62"
[...]

PR #12025, opened 2025-07-09
Fix issues with `mimirtool remote-read export` and long time ranges

charleskorn:
#### What this PR does

This PR fixes a number of issues in `mimirtool remote-read export` that could occur when querying a range longer than an hour, or that crossed block boundaries (which were hardcoded to 2h in size). In particular, queries that crossed block boundaries could omit some output blocks.

Key changes:

- Output blocks will now be 2h in length, not 1h, and this can be configured with `--block-size`.
- Time ranges that cross block boundaries are now handled correctly, including for cases where some series appear in only some output blocks.
- Long time ranges are now broken up into intervals that match the output block size, reducing the amount of data that must be held in memory by `mimirtool` and simplifying the logic of producing blocks in this case. 
  - An alternative solution would be to create multiple block writers, one for each output block, and append to the correct one based on the samples received, but this would require all received samples to be held in memory until the last one is received.
- Details of blocks are now reliably printed to the console, rather than sometimes being incomplete or omitted entirely, or failing with `read/write on closed pipe`.

Related to #11404

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -183,7 +183,10 @@
 ### Mimirtool
 
 * [FEATURE] Add `--enable-experimental-functions` flag to commands that parse PromQL to allow parsing experimental functions such as `sort_by_label()`.
+* [ENHANCEMENT] Add `--block-size` CLI flag to `remote-read export` that allows setting the output block size. #12025
 * [BUGFIX] Fix issue where `remote-read` doesn't behave like other mimirtool commands for authentication. #11402
+* [BUGFIX] Fix issue where `remote-read export` could omit some samples if the query time range spanned multiple blocks. #12025
+* [BUGFIX] Fix issue where `remote-read export` could omit some output blocks in the list printed to the console, or fail with `read/write on closed pipe`. #12025
 
 ### Mimir Continuous Test
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,8 +185,8 @@
 * [FEATURE] Add `--enable-experimental-functions` flag to commands that parse PromQL to allow parsing experimental functions such as `sort_by_label()`.
 * [ENHANCEMENT] Add `--block-size` CLI flag to `remote-read export` that allows setting the output block size. #12025
 * [BUGFIX] Fix issue where `remote-read` doesn't behave like other mimirtool commands for authentication. #11402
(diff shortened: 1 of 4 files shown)


pkg/mimirtool/commands/remote_read.go
```diff
@@ -109,6 +121,10 @@ func (c *RemoteReadCommand) Register(app *kingpin.Application, envVars EnvVarNam
 	exportCmd.Flag("tsdb-path", "Path to the folder where to store the TSDB blocks, if not set a new directory in $TEMP is created.").
 		Default("").
 		StringVar(&c.tsdbPath)
+
+	exportCmd.Flag("block-size", "Maximum block size to create. Requests that span multiple blocks will be split.").
+		Default((time.Duration(tsdb.DefaultBlockDuration) * time.Millisecond).String()).
+		DurationVar(&c.blockSize)
```
narqo: (nit) I think, prometheus calls this `block-duration` and "block size" refers to size in bytes

CHANGELOG.md
```diff
@@ -183,7 +183,10 @@
 ### Mimirtool
 
 * [FEATURE] Add `--enable-experimental-functions` flag to commands that parse PromQL to allow parsing experimental functions such as `sort_by_label()`.
+* [ENHANCEMENT] Add `--block-size` CLI flag to `remote-read export` that allows setting the output block size. #12025
 * [BUGFIX] Fix issue where `remote-read` doesn't behave like other mimirtool commands for authentication. #11402
+* [BUGFIX] Fix issue where `remote-read export` could omit some samples if the query time range spanned multiple blocks. #12025
```
```suggestion
* [BUGFIX] Fix issue where `remote-read export` could omit some samples if the query time range spans multiple blocks. #12025
```

CHANGELOG.md
```diff
@@ -183,7 +183,10 @@
...
+* [BUGFIX] Fix issue where `remote-read export` could omit some output blocks in the list printed to the console, or fail with `read/write on closed pipe`. #12025
```
```suggestion
* [BUGFIX] Fix issue where `remote-read export` could omit some output blocks in the list printed to the console or fail with `read/write on closed pipe`. #12025
```

tacole02 (approved):

pkg/mimirtool/commands/remote_read.go
charleskorn: Good point, fixed in dadfbb9.

Issue #12033, opened 2025-07-09
store-gateway - querier distributed deadlock


dimitarvdimitrov:
We were speaking with @charleskorn  about the possible deadlock between queriers and store-gateways.

### Background 

Queriers send many requests to multiple store-gateways and only proceed with a query once all store-gateways have started processing the query and have returned some preliminary results.

https://github.com/grafana/mimir/blob/ca1441dc1dcb1967a281be434286920474b00d17/pkg/querier/blocks_store_queryable.go#L942-L943

store-gateways enforce a concurrency limit on the in-flight queries

https://github.com/grafana/mimir/blob/74c9a8955e89ad41565b35f05c2ffdf1f93c89a4/pkg/storegateway/bucket.go#L741

### Problem

So it’s possible that at some point queries from multiple queriers end up waiting for each other to start executing at the store-gateway. Meanwhile in the store-gateway there are actively occupied concurrency slots which are sitting idle. Another reason this slowdown can happen is if there is one slower store-gateway which is piling up its requests.

In case of a single slow store-gateway This will lead to $O({NumQueriers})$ queries being blocked in every other store-gateway. This can either severy reduce the throughput of the read path (now each store-gateway has $O({NumQueriers})$ idle concurrency slots or it can cause a deadlock if the number of queriers is higher than the number of concurrency slots.

### Potential solutions

1. Propagate trace ID or query start time with each query to the store-gateway. store-gateways will give preference to queries which are children of queries which have started earlier. The idea is to preserve the first-come-first-served QoS that the scheduler tries to give.
2. Disable concurrency limits; instead use a reactive instance limit to prevent overloading store-gateways (e.g. something which shuts off queries when memory runs too high)
3. Increase concurrency limits: https://github.com/grafana/mimir/issues/7608

### Next steps

Add metrics to confirm this is happening

dimitarvdimitrov:
> ### Next steps
> Add metrics to confirm this is happening

* https://github.com/grafana/mimir/pull/12034

dimitarvdimitrov:
> ### Potential solutions

an idea came to me after chatting with @kavirajk 

We can take inspiration from HTTP and QUIC is to instead of limit concurrency on a request, we would limit concurrency on an **active** request. This is a request for which we can either compute the next message to send. This means that all the queries which are stuck being idle do not actually consume a lot of resources (requests hold up to 5000 series and their chunks in memory, which is in the order of 10MB)

dimitarvdimitrov:
[...]

PR #12034, opened 2025-07-09
Add concurrency gate bottleneck metrics for store-gateway queries

dimitarvdimitrov:
This PR adds metrics to detect and measure concurrency gate bottlenecks in store-gateway queries, addressing the observability need described in #12033.

## Problem
As described in #12033, there's a potential distributed deadlock where queries from multiple queriers can end up waiting to execute at store-gateways, with concurrency slots sitting idle.

## Solution
This PR adds two new native histogram metrics:

- **cortex_querier_storegateway_concurrency_wait_seconds**: Tracks the longest time any store-gateway waits to acquire its concurrency gate
- **cortex_querier_storegateway_concurrency_efficiency**: Measures query efficiency as 1 - (blocked_time / total_query_time)

## Implementation
- Store-gateway captures gate request/acquired timestamps and sends them via gRPC
- Querier aggregates timings across all store-gateways and reports metrics
- Uses native histograms for better precision and storage efficiency

These metrics provide the observability needed to confirm whether the concurrency bottleneck described in #12033 is happening.

Addresses #12033


pkg/storegateway/stats.go
```diff
@@ -172,6 +176,14 @@ func (s queryStats) merge(o *queryStats) *queryStats {
 	s.streamingSeriesIndexHeaderLoadDuration += o.streamingSeriesIndexHeaderLoadDuration
 	s.streamingSeriesConcurrencyLimitWaitDuration += o.streamingSeriesConcurrencyLimitWaitDuration
 
+	// For timestamps, take the latest one (not sum). This should only be set once for each request, so merging doesn't make a lot of sense anyways.
```
charleskorn, line 179: If these should only be set once for each request, should we panic if it ever happens that they're set multiple times?

pkg/querier/blocks_store_queryable.go
```diff
@@ -136,7 +140,51 @@ func newBlocksStoreQueryableMetrics(reg prometheus.Registerer) *blocksStoreQuery
 			Name: "cortex_querier_query_storegateway_chunks_total",
 			Help: "Number of chunks received from store gateways at query time.",
 		}),
+
+		concurrencyWait: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+			Name:                            "cortex_querier_storegateway_concurrency_wait_seconds",
+			Help:                            "Time from query start until the last store-gateway started processing the query.",
```
charleskorn, line 146: Is this more accurate? "query start" seems ambiguous to me, and "last store-gateway" isn't quite correct:

```suggestion
			Help:                            "Longest wait time amongst all store-gateways that serviced a request, where wait time is the time between the request arriving at that store-gateway and when it began processing the request.",
```

dimitarvdimitrov:
another approach which will be easier to reason about, but less broad is to detect dependency cycles within a single store-gateway. This should be easy to do, but will miss the distributed deadlock detection. 

At the same time I couldn't find indications of deadlock actually happening (unexplained delays, too many retries of requests, high CPU without progress on queries). It's possible that the 5s timeout is saving us.

PR #12035, opened 2025-07-10
Allow cost attribution label config to include output labels

seizethedave:
#### What this PR does

This PR enables configuration of an output label for each defined cost attribution label. A new `cost_attribution_labels_structured` YAML element is added to the `Limits` struct with the structure:

```go
type Label struct {
	Input, Output string
}
```

This allows both an "input" label that refers to metrics actually in use in the system, and an "output" label which is what will be emitted at scrape time in the usage metrics.

`cost_attribution_labels_structured` is preferred over `cost_attribution_labels` if both are present.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/costattribution/manager.go
```diff
@@ -282,7 +294,7 @@ func (m *Manager) updateTracker(userID string) (*SampleTracker, *ActiveSeriesTra
 		m.stmtx.Unlock()
 	}
 
-	if !at.hasSameLabels(lbls) || at.maxCardinality != newMaxCardinality || st.cooldownDuration != newCooldownDuration {
+	if !at.hasSameLabels(lbls) || at.maxCardinality != newMaxCardinality || at.cooldownDuration != newCooldownDuration {
```
seizethedave: that was a copy/paste error, wasn't it? (Cursor caught it.)

pkg/util/validation/limits.go
```diff
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/time/rate"
 	"gopkg.in/yaml.v3"
 
+	costattributionmodel "github.com/grafana/mimir/pkg/costattribution/model"
```
colega: Can we rename the package to `costattributionmodel` so we don't need to alias it?
It will also make the code more readable in other places where `model.Label` isn't really desciptive (is it the Prometheus model? Mimir model? etc.)

pkg/util/validation/limits.go
```diff
@@ -363,6 +365,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.CostAttributionLabels, costAttributionLabelsFlag, "Defines labels for cost attribution. Applies to metrics like cortex_distributor_received_attributed_samples_total. To disable, set to an empty string. For example, 'team,service' produces metrics such as cortex_distributor_received_attributed_samples_total{team='frontend', service='api'}.")
```
colega: How about we remove this flag? It's experimental, and we're not using it anyway.

seizethedave: It is currently in-use for a handful of statically configured tenants. I think we'll need a migration period for those.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [FEATURE] Distributor: Add experimental `-distributor.otel-native-delta-ingestion` option to allow primitive delta metrics ingestion via the OTLP endpoint. #11631
 * [FEATURE] MQE: Add support for experimental `sort_by_label` and `sort_by_label_desc` PromQL functions. #11930
 * [FEATURE] Ingester/Block-builder: Handle the created timestamp field for remote-write requests. #11977
+* [FEATURE] Cost attribution: Labels specified in limit configuration may specify an output label in order to override emitted label names. #12035
```
```suggestion
* [FEATURE] Cost attribution: Labels specified in the limit configuration may specify an output label in order to override emitted label names. #12035
```

pkg/util/validation/limits.go
colega: I don't think the flag is used, but the yaml config, right? I mean we can remove the flag and leave the yaml config.

seizethedave: Aha. You're right.

PR #12036, opened 2025-07-10
Enforce consistent formatting of Protobuf files

charleskorn:
#### What this PR does

This PR adds Buf to our build image, and uses it to enforce consistent formatting of Protobuf files.

While the generated Go files have changed, all of these changes are in the file descriptor, and have no impact on Mimir's behaviour.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


Makefile
```diff
@@ -509,6 +509,14 @@ format-promql-tests:
 check-promql-tests: format-promql-tests
 	@./tools/find-diff-or-untracked.sh $(PROMQL_TESTS) || (echo "Please format PromQL test files by running 'format-promql-tests'" && false)
 
+.PHONY: format-protobuf
+format-protobuf:
```
dimitarvdimitrov, line 513: can we make this execute after someone runs `make protos`?

charleskorn: It would certainly be possible, but I'd personally find that behaviour a bit surprising (eg. imagine if building the binary also formatted the Go code).

charleskorn:
> any idea why the descriptors changed? it's a bit sus

The descriptors are sensitive to changes in the Protobuf files, but the different descriptors produce semantically equivalent results.

For example, the change in the `distributorpb` descriptor is due to the order of imports changing. The amount of change in the descriptor's bytes can also be misleading as they are gzipped.

This is the dummy test I used to verify this, it'll dump the contents of a file's descriptor (`fileDescriptor_c518e33639ca565d` in this case):

```go
func TestDump(t *testing.T) {
	reader, err := gzip.NewReader(bytes.NewReader(fileDescriptor_c518e33639ca565d))
	require.NoError(t, err)

	uncompressed, err := io.ReadAll(reader)
	require.NoError(t, err)

	desc := &descriptor.FileDescriptorProto{}
	err = proto.Unmarshal(uncompressed, desc)
	require.NoError(t, err)

	marshaler := &jsonpb.Marshaler{
		Indent: "  ",
	}
	require.NoError(t, marshaler.Marshal(os.Stdout, desc))
}
```

Makefile
dimitarvdimitrov: My problem was that you'd almost always have linter fail after you work in these files. Why wait until ci only to have the linter fail?

We can fix it later if it's actually annoying

dimitarvdimitrov (approved):

PR #12038, opened 2025-07-10
fix(frontend): queries with duration expression

krajorama:
#### What this PR does

The sharding, splitting middlewares do not handle duration expressions in PromQL queries. Any offset, range, step that would be calculated from a duration expression like (e.g. `metric offset (2+2)`) is treated as if the expression wasn't there and the value of the offset, range, step was 0/default (e.g. `metric offset 0`). 

Currently all such expressions can be pre-calculated based on the query step. Add a middleware to pre-calculate the offsets, ranges, steps.

Fixes N/A

#### Checklist

- [x] Tests updated.
- [N/A] Documentation added.


zenador:

pkg/frontend/querymiddleware/durations.go
```diff
@@ -0,0 +1,174 @@
...
+func (d *durationsMiddleware) Do(ctx context.Context, req MetricsQueryRequest) (Response, error) {
+	// Log the instant query and its timestamp in every error log, so that we have more information for debugging failures.
+	logger := log.With(d.logger, "query", req.GetQuery(), "query_timestamp", req.GetStart())
+
+	spanLog, ctx := spanlogger.New(ctx, logger, tracer, "durationsMiddleware.Do")
+	defer spanLog.Finish()
+
+	expr, err := parser.ParseExpr(req.GetQuery())
+	if err != nil {
+		level.Warn(spanLog).Log("msg", "failed to parse query", "err", err)
```
jesusvazquez: Just to clarify, are we using `Warn` error level to avoid noisy errors?

pkg/frontend/querymiddleware/durations.go
```diff
@@ -0,0 +1,174 @@
...
+		level.Warn(spanLog).Log("msg", "failed to update the query expression after durations expressions were evaluated", "err", err)
+		return nil, apierror.New(apierror.TypeInternal, DecorateWithParamName(err, "query").Error())
+	}
+
+	return d.next.Do(ctx, req)
+}
+
+type durationVisitor struct{}
+
+// Verify that duration expressions only contain durations that Mimir supports.
```
jesusvazquez:
```suggestion
// Visit verifies that duration expressions only contain durations that Mimir supports.
```
I just run the linter in my head since there is already a comment 🙈

jesusvazquez (approved):
LGTM, left a few small nits

pkg/frontend/querymiddleware/durations.go
krajorama: I went with not logging anything like https://github.com/grafana/mimir/blob/47f7604146a054bed21802a03fb4b14919e98c6c/pkg/frontend/querymiddleware/stats.go#L66

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * [BUGFIX] Query-frontend: Allow limit parameter to be 0 in label queries to explicitly request unlimited results. #12054
 * [BUGFIX] Distributor: Fix a possible panic in the OTLP push path while handling a gRPC status error. #12072
 * [BUGFIX] Tracing: Skip tracing configuration when no tracing environment variables were provided. #12074
+* [BUGFIX] Query-frontend: evaluate experimental duration expressions before sharding, splitting, caching otherwise the result is not correct. #12038
```
```suggestion
* [BUGFIX] Query-frontend: Evaluate experimental duration expressions before sharding, splitting, and caching. Otherwise, the result is not correct. #12038
```

pkg/frontend/querymiddleware/durations.go
```diff
@@ -0,0 +1,190 @@
...
+
+	origQuery := req.GetQuery()
+	expr, err := parser.ParseExpr(origQuery)
+	if err != nil {
+		// This middleware focuses on duration expressions, so if the query is
+		// not valid, we just fall through to the next handler.
+		return req, nil
+	}
+
+	expr, err = promql.PreprocessExpr(expr, time.UnixMilli(req.GetStart()), time.UnixMilli(req.GetEnd()), time.Duration(req.GetStep())*time.Second)
```
charleskorn, line 65: `PreprocessExpr` will also resolve the value of timestamps in expressions like `foo @ start()`. We already have logic to replace `start()` and `end()` with the actual query timestamp in the query splitting logic (`evaluateAtModifierFunction` in `split_and_cache.go`), so now we're duplicating work.

Using `PreprocessExpr` to do this seems like the most robust way to do this - should we remove this from the split and cache middleware?

charleskorn (commented):
>The sharding, splitting middlewares do not handle duration expressions in PromQL queries. Currently all such expressions can be pre-calculated based on the query step.

Could you please elaborate on what doesn't work here?

krajorama:
> > The sharding, splitting middlewares do not handle duration expressions in PromQL queries. Currently all such expressions can be pre-calculated based on the query step.
> 
> Could you please elaborate on what doesn't work here?

updated the PR

pkg/frontend/querymiddleware/durations.go
krajorama: I didn't actually catch that. I've added tests for `start()` and `end()`.
It probably makes sense to remove from the split and cache, but it's up to you. Do you want me to do it in this PR?

charleskorn: >Do you want me to do it in this PR?

I'm happy for it to be a follow-up PR.

charleskorn (approved):

PR #12040, opened 2025-07-10
feature(parquet): Implement store-gateway limits

jesusvazquez:
This PR is based on the upstream work https://github.com/prometheus-community/parquet-common/pull/81

The idea is to implement a set of basic quota limiters that can protect us against potential bad queries for the gateways.

Note we had to bring bits of the code available in the querier in upstream because we have our own chunk querier in Mimir.

### DIFF
--- a/pkg/storegateway/bucket_stores_test.go
+++ b/pkg/storegateway/bucket_stores_test.go
@@ -670,6 +670,40 @@ func generateStorageBlock(t *testing.T, storageDir, userID string, metricName st
 	require.NoError(t, db.Snapshot(userDir, true))
 }
 
+func generateStorageBlockWithMultipleSeries(t *testing.T, storageDir, userID string, metricName string, numSeries int, minT, maxT int64, step int) {
+	// Create a directory for the user (if doesn't already exist).
+	userDir := filepath.Join(storageDir, userID)
+	if _, err := os.Stat(userDir); err != nil {
+		require.NoError(t, os.Mkdir(userDir, os.ModePerm))
+	}
+
+	// Create a temporary directory where the TSDB is opened,
+	// then it will be snapshotted to the storage directory.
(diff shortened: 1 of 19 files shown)


pkg/storegateway/parquet_bucket_store_test.go
```diff
@@ -254,3 +255,56 @@ func createTestParquetBucketStore(t *testing.T, userID string, bkt objstore.Buck
 	require.NoError(t, err)
 	return store
 }
+func TestParquetBucketStores_RowLimits(t *testing.T) {
```
jesusvazquez, line 262: This test is just making sure that the default and a high limit allow query execution. It doesnt test that a low limit makes the query stop because it requires modifying `generateParquetStorageBlock(t, storageDir, bkt, userID, metricName, 10, 100, 15)` to generate multiple series blocks and that would make the change bigger plus that is tested in upstream.

jesusvazquez: addressed as part of https://github.com/grafana/mimir/pull/12040/commits/03de86c1de0a1d393a4a89a022676be896e45a11

pkg/storegateway/bucket_stores_test.go
```diff
@@ -670,6 +670,40 @@ func generateStorageBlock(t *testing.T, storageDir, userID string, metricName st
 	require.NoError(t, db.Snapshot(userDir, true))
 }
 
+func generateStorageBlockWithMultipleSeries(t *testing.T, storageDir, userID string, metricName string, numSeries int, minT, maxT int64, step int) {
```
npazosmendez: I would move this to one of our `parquet_..test.go` files to avoid conflicts

npazosmendez (approved):
Nice -- changes LGTM, but there's a test failing after a new label was added to the block creation helper

PR #11834, opened 2025-06-24
Start tracking -compactor.max-compaction-time after the initial compaction planning phase

pracucci:
#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #11819, opened 2025-06-23
Optimise ShardAwareDeduplicateFilter

pracucci:
#### What this PR does

This PR has been be authored by @colega and myself.

We had an issue where a tenant compaction was lagging behind, and their number blocks reached the monster number of 250K (from a baseline of 10K). When that happened, we found out that the compactor planning was extremely slow (3 hours), and that the slow down came from `ShardAwareDeduplicateFilter`.

In this PR we introduce some small optimizations to `ShardAwareDeduplicateFilter` for tenants with a very large number of blocks. The impact on performance is not impressive, but we got a -33% reduction (from 3h30m to 2h15m) for the real world scenario.

Existing benchmark result:

```

goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/compactor
cpu: Apple M3 Pro
                                            │  before.txt   │ after-with-bloom-filter-and-without-list.txt │
                                            │    sec/op     │        sec/op          vs base               │
DeduplicateFilter_Filter/Block-10/#00-11       17.64µ ± 14%             25.23µ ± 3%  +43.04% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-10/#01-11       53.33µ ±  1%             76.36µ ± 1%  +43.18% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-100/#00-11      383.6µ ±  4%             294.2µ ± 1%  -23.30% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-100/#01-11     1153.2µ ±  1%             882.1µ ± 1%  -23.51% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-1000/#00-11    28.346m ±  1%             7.578m ± 6%  -73.26% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-1000/#01-11     90.15m ±  1%             23.02m ± 1%  -74.46% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-10000/#00-11     6.820 ±  5%              1.237 ± 1%  -81.86% (p=0.002 n=6)
DeduplicateFilter_Filter/Block-10000/#01-11    20.710 ±  2%              3.723 ± 1%  -82.02% (p=0.002 n=6)
geomean                                        10.52m                   5.018m       -52.31%

                                            │  before.txt  │ after-with-bloom-filter-and-without-list.txt │
                                            │     B/op     │         B/op           vs base               │
[...]

Issue #11801, opened 2025-06-21
store-gateway panics exceeding 10000 threads


dimitarvdimitrov:
## Problem
Store Gateway panics with `fatal error: thread exhaustion; runtime: program exceeds 10000-thread limit` after hitting Go's 10000 thread limit. Analysis shows 9986 goroutines blocked on syscalls, all performing symbol lookups.

## Root Cause

After startup, the symbols cache is empty, causing concurrent queries to trigger simultaneous symbol lookups that exhaust the golang thread limit.

[From the docs:](https://pkg.go.dev/runtime/debug#SetMaxThreads)
> A Go program creates a new thread only when a goroutine is ready to run but all the existing threads are blocked in system calls, cgo calls, or are locked to other goroutines due to use of runtime.LockOSThread.

We have a code path which calls LookupSymbol for every symbol on a series. This is what caused the saturated syscalls (`multiLookupStringify`).

## Flame graphs

Flame graph from the goroutine dump during the panic. `multiLookupStringify` highlighted of all threads that end in syscall

<img width="1610" alt="Image" src="https://github.com/user-attachments/assets/67c0e642-1dab-4134-b680-4161980cb25e" />

Summary of goroutine stacks showing 9986 gorotuines in Syscall

```
Top 20 most frequent functions:
  63477: runtime.gopark
  38071: runtime.selectgo
  35377: runtime.goexit
  34740: github.com/grafana/mimir/pkg/storegateway.(*mergedSeriesChunkRefsSet).ensureItemAvailableToRead
  24387: runtime.semacquire1
  24376: github.com/grafana/mimir/pkg/storegateway.(*mergedSeriesChunkRefsSet).ensureCursors
  16991: golang.org/x/sync/errgroup.(*Group).add.func1
  16976: github.com/grafana/mimir/pkg/storage/indexheader.NewLazyBinaryReader.gowrap1
  16859: github.com/grafana/mimir/pkg/storage/indexheader.(*LazyBinaryReader).controlLoop
  11794: golang.org/x/sync/errgroup.(*Group).Wait
  11742: github.com/grafana/mimir/pkg/storegateway.(*mergedSeriesChunkRefsSet).Next
  11648: sync.runtime_SemacquireWaitGroup
  11585: runtime.goparkunlock
  11580: sync.(*WaitGroup).Wait
  10488: github.com/grafana/mimir/pkg/storegateway.flattenedSeriesChunkRefsIterator.Next
[...]

PR #12048, opened 2025-07-10
store-gateway: Fix goroutine leak in LabelValues when using errgroup …

philk:
#### What this PR does

We've recently started seeing intermittent panics from our store-gateways in our production environment, and I believe this may be the cause. In total transparency I used Gemini to find this as I'm not very familiar with this part of the codebase, but this certainly seems like a valid bug, if potentially not also the root cause. I've spent a bit of time trying to come up with a test case that can reproduce it, but that's been a challenge and I wanted to get eyes on this.

First thousand lines of the crash, the entire logs are >100k lines since it's every goroutine:
[Logs-2025-07-08 18_10_49.txt](https://github.com/user-attachments/files/21168977/Logs-2025-07-08.18_10_49.txt)

AI Slop Explanation:
The ensureIndexHeaderLoaded call was using the original context instead of the errgroup context (gctx), which could cause goroutines to get stuck in select operations when other goroutines in the same errgroup failed or were canceled. This led to goroutine leaks as seen in the stack trace where goroutines were stuck waiting on gate.Start() calls.

The fix ensures that ensureIndexHeaderLoaded respects the errgroup's cancellation, allowing proper cleanup when any goroutine in the group fails.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added. (N/A)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Block-builder-scheduler: Detect gaps in scheduled and completed jobs. #11867
 * [ENHANCEMENT] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol has been updated. Created timestamps are now supported. This feature includes some limitations. If samples in a write request aren't ordered by time, the created timestamp might be dropped. Additionally, per-series metadata is automatically merged on the metric family level. Ingestion might fail if the client sends ProtoBuf fields out-of-order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0`, depending on the detected remote-write protocol. #11977
 * [BUGFIX] Distributor: Validate the RW2 symbols field and reject invalid requests that don't have an empty string as the first symbol. #11953
+* [BUGFIX] Store-gateway: Pass correct errGroup context to prevent thread exhaustion panic
```
narqo:
```suggestion
* [BUGFIX] Store-gateway: Pass correct errGroup context to prevent thread exhaustion panic when handling LabelValues #12048
```

narqo (commented):
Thank you for the PR. From the first look, this indeed looks like a bug that slipped through in #8528. Kudos to gemini for sifting through the stack trace. BUT, so far I'm not convinced this bug was what caused the "thread exhaustion".

In the attached stacktrace, there were seem to only two goroutines, seating in the `storegateway.(*BucketStore).LabelValues`. These goroutines were seating the `errGroup.Wait`. I don't see any other blocked goroutine, that originated from inside storegateway.

I suspect the stacktrace was truncated. Do you happen to have the full one?

Note that we recently investigated another case of the `thread exhaustion` panics in the store-gateway https://github.com/grafana/mimir/issues/11801 Could there be a chance that the one you'd seen came from the same problem?

philk:
Had to remind myself how to use logcli...
Here's all 600k lines for ya 😅 

[stacktrace.txt.gz](https://github.com/user-attachments/files/21185112/stacktrace.txt.gz)

As for #11801 here's what that query looks like on my cluster right before those logs where all of our store-gateways crashed (17:04:51 UTC)
Screenshot 2025-07-11 at 9 56 20 AM

narqo:
> [..] here's what that query looks like on my cluster right before those logs where all of our store-gateways crashed (17:04:51 UTC)

I am not 100% sure what the screenshot shows. Is this the average number of blocks queried by store-gateway instances? 50 to 250 could be a bit too much.

In our internal investigations for #11801, the key insight, that convinced us that the overall number of syscalls was the cause, was the graph, that showed a skew in the number of blocks per-tenant. That is the [compactor's][1] "Tenants with largest number of blocks" graph showed this 👇

Screenshot 2025-07-11 at 20 30 14

<details>

<summary>PromQL</summary>

```
max by (user) (cortex_bucket_blocks_count{job=~".+compactor.+"})
```

</details>

Note, although I didn't include the actual tenants in the screenshot, the data shows a steady increase in number of blocks in one tenant, that caused by a lagging compactor. This was what caused the store-gateways to read almost 10x more blocks per query, than what was the norm for the cluster.

>  Here's all 600k lines for ya 

In the stack trace there are more around 8500 goroutines in the `syscall.Syscall`. All seem to originated from the `indexheader` reader, which also suggests the same symptoms as in #11801.

Could you check, how was the compactor doing during around the day, when the problem occurred in your cluster?

[1]: https://grafana.com/docs/mimir/latest/manage/monitor-grafana-mimir/dashboards/compactor/

philk:
Oh it's definitely related to entirely too many blocks, we ran into #11834 and didn't have good alerts in place to catch it. Let's just say things got out of control...

Screenshot 2025-07-14 at 10 53 22 AM

Our metadata sync for the compactors was taking ~28-30h `msg="successfully synchronized block metadata" duration=29h6m26.4033835s`. So of course compactions were basically never being run. Kind of looks like #11819 is also relevant for us...really looking forward to 2.17 😁 

We tripled our number of compactors and disabled max-compaction-time and we've been slowly recovering for a couple weeks now. I'm sure you all are already familiar with a number of issues we were facing because of this like: "why are these parts of the system just growing in usage and not behaving?". Turns out compaction is quite important 😅  I'd be curious why performance improvements aren't linear as the blocks go down, things just really improved in the last few days.

(Don't worry about that blue line that's rising, that's a mimir-continuous-test tenant and we're going to adjust our compactors again now that we've caught up on the primary tenant)

Screenshot 2025-07-14 at 10 50 06 AM
Screenshot 2025-07-14 at 10 50 36 AM

dimitarvdimitrov:
the alert from #11803 should also help detect this falling-behind compaction earlier

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Block-builder-scheduler: Detect gaps in scheduled and completed jobs. #11867
 * [ENHANCEMENT] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol has been updated. Created timestamps are now supported. This feature includes some limitations. If samples in a write request aren't ordered by time, the created timestamp might be dropped. Additionally, per-series metadata is automatically merged on the metric family level. Ingestion might fail if the client sends ProtoBuf fields out-of-order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0`, depending on the detected remote-write protocol. #11977
 * [BUGFIX] Distributor: Validate the RW2 symbols field and reject invalid requests that don't have an empty string as the first symbol. #11953
+* [BUGFIX] Store-gateway: Pass correct errGroup context to prevent thread exhaustion panic when handling LabelValues #12048
```
narqo: Since this wasn't the cause of the panics, let's not mention it

```suggestion
* [BUGFIX] Store-gateway: Fix potential goroutine leak by passing the scoped context in LabelValues. #12048
```

philk:
Yup, easy rebase. Looks good to go now.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 * [BUGFIX] Distributor: Check `max_inflight_push_requests_bytes` before decompressing incoming requests. #11967
 * [BUGFIX] Query-frontend: Allow limit parameter to be 0 in label queries to explicitly request unlimited results. #12054
 * [BUGFIX] Distributor: Fix a possible panic in the OTLP push path while handling a gRPC status error. #12072
+* [BUGFIX] Tracing: Skip tracing configuration when no tracing environment variables were provided. #12074
```
tacole02, line 48:
```suggestion
* [BUGFIX] Tracing: Skip tracing configuration when no tracing environment variables are provided. #12074
```

tacole02 (approved):

zenador:

PR #12052, opened 2025-07-11
Refactor MultitenantAlertmanager to use amConfig with alerting templates

yuri-tceretian:
#### What this PR does
This PR proposes two changes:
1. refactors configuration structure `amConfig`. The structure does not use protobuf's structures. All fields are copied from the protobuf, and type of field template has changed to `[]definition.PostableApiTemplate` from alerting module. 
I chose this approach to avoid changing the Mimir Alertmanager's protobuf model and instead make boundaries where that model gets converted to `amConfig` which is then used everywhere downstream.
2. Also, I decided to refactor `MultitenantAlertmanager` and change the configuration cache map to store entire `amConfig` instead of its part, which `alertspb.AlertConfigDesc` was. This will potentially have memory impact on Grafana alertmanager that stores more information than Mimir's. However, this will let us to fix the logic that determines whether configuration has updated (currently, it ignores changes in smtp config and urls). 

This refactoring is needed for propagating templates from Grafana with information about their kind. This will be added in a follow-up PR.

Related to https://github.com/grafana/alerting-squad/issues/1152

### DIFF
--- a/pkg/alertmanager/config.go
+++ b/pkg/alertmanager/config.go
@@ -23,7 +23,7 @@ import (
 
 // createUsableGrafanaConfig creates an amConfig from a GrafanaAlertConfigDesc.
 // If provided, it assigns the global section from the Mimir config to the Grafana config.
-// The amConfig.emailConfig field can be used to create Grafana email integrations.
+// The amConfig.EmailConfig field can be used to create Grafana email integrations.
 func createUsableGrafanaConfig(logger log.Logger, gCfg alertspb.GrafanaAlertConfigDesc, rawMimirConfig string) (amConfig, error) {
 	externalURL, err := url.Parse(gCfg.ExternalUrl)
 	if err != nil {
@@ -141,9 +141,9 @@ func createUsableGrafanaConfig(logger log.Logger, gCfg alertspb.GrafanaAlertConf
 		User:               gCfg.User,
 		RawConfig:          string(rawCfg),
 		Templates:          tmpl,
-		tmplExternalURL:    externalURL,
-		usingGrafanaConfig: true,
-		emailConfig:        emailCfg,
+		TmplExternalURL:    externalURL,
+		UsingGrafanaConfig: true,
+		EmailConfig:        emailCfg,
 	}, nil
 }
 
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -764,7 +764,7 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 		User:            cfgs.Mimir.User,
 		RawConfig:       cfgs.Mimir.RawConfig,
 		Templates:       templateDescToPostableApiTemplate(cfgs.Mimir.Templates, definition.MimirTemplateKind),
-		tmplExternalURL: am.cfg.ExternalURL.URL,
+		TmplExternalURL: am.cfg.ExternalURL.URL,
 	}
 
 	// Check if tenants can be skipped (strict initialization enabled).
@@ -840,7 +840,7 @@ func (am *MultitenantAlertmanager) syncStates(ctx context.Context, cfg amConfig)
 
 	// If we're not using Grafana configuration, we shouldn't use Grafana state.
 	// Update the flag accordingly.
-	if !cfg.usingGrafanaConfig {
+	if !cfg.UsingGrafanaConfig {
 		if ok && userAM.usingGrafanaState.CompareAndSwap(true, false) {
 			level.Debug(am.logger).Log("msg", "Grafana state unpromoted", "user", cfg.User)
 		}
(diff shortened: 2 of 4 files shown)


pkg/alertmanager/api_grafana.go
```diff
@@ -512,22 +512,14 @@ func validateUserGrafanaConfig(logger log.Logger, cfg alertspb.GrafanaAlertConfi
...
-	tmpls := make([]alertingTemplates.TemplateDefinition, 0, len(grafanaConfig.Templates))
-	for _, tmpl := range grafanaConfig.Templates {
-		tmpls = append(tmpls, alertingTemplates.TemplateDefinition{
-			Name:     tmpl.Filename,
-			Template: tmpl.Body,
-			Kind:     alertingTemplates.GrafanaKind,
-		})
-	}
-	factory, err := alertingTemplates.NewFactory(tmpls, logger, "http://localhost", user) // use fake URL to avoid errors.
+	factory, err := alertingTemplates.NewFactory(alertingNotify.PostableAPITemplatesToTemplateDefinitions(grafanaConfig.Templates), logger, "http://localhost", user) // use fake URL to avoid errors.
```
santihernandezc: Just a suggestion to make it more readable:
```suggestion
	factory, err := alertingTemplates.NewFactory(
		alertingNotify.PostableAPITemplatesToTemplateDefinitions(grafanaConfig.Templates),
		logger,
		"http://localhost", // Use a fake URL to avoid errors.
		user,
	)
```

pkg/alertmanager/config.go
```diff
@@ -134,10 +134,27 @@ func createUsableGrafanaConfig(logger log.Logger, gCfg alertspb.GrafanaAlertConf
 		}
 	}
 
+	// the map can only contain templates of Grafana kind.
```
santihernandezc:
```suggestion
	// The map can only contain Grafana templates.
```
or
```suggestion
	// The map can only contain templates of the Grafana kind.
```

pkg/alertmanager/multitenant.go
```diff
@@ -334,7 +334,7 @@ type MultitenantAlertmanager struct {
 	alertmanagers    map[string]*Alertmanager
 	// Stores the current set of configurations we're running in each tenant's Alertmanager.
 	// Used for comparing configurations as we synchronize them.
-	cfgs map[string]alertspb.AlertConfigDesc
+	cfgs map[string]amConfig
```
santihernandezc, line 337: I wouldn't put the entire `amConfig` struct here.

My biggest concern is the `emailConfig` field, which is [at least 216 bytes](https://go.dev/play/p/j8bb3uCJ9tA). Changes in SMTP configuration are **extremely** rare, and it's enough to unpromote-promote a configuration (or make any other change) to update the email sender config in the Alertmanager.

If we have 10K tenants and a replication factor of 3, that's ~6.5MB of data we're not using. I don't think this is worth the extra overhead.

If we want to check for changes in all config fields, perhaps it's better to just keep a hash of the config.

yuri-tceretian: Yea, I wouldn't put any of that in the cache where just a hash is sufficient. I am working on a change to calculate hash of the structure. Rare is not impossible, and therefore I would consider it as a bug because if it changes, the config won't be applied.

yuri-tceretian: Here is the PR that improves this https://github.com/grafana/mimir/pull/12052

pkg/alertmanager/multitenant.go
```diff
@@ -891,10 +893,12 @@ func (am *MultitenantAlertmanager) syncStates(ctx context.Context, cfg amConfig)
...
 
 type amConfig struct {
-	alertspb.AlertConfigDesc
-	tmplExternalURL    *url.URL
-	usingGrafanaConfig bool
-	emailConfig        alertingReceivers.EmailSenderConfig
+	User               string
+	RawConfig          string
+	Templates          []definition.PostableApiTemplate
+	TmplExternalURL    *url.URL
```
JacobsonMT, line 899: NIT: Do we need to export `tmplExternalURL`, `UsingGrafanaConfig`, or `EmailConfig`?

JacobsonMT (commented):
Code looks good, one small nit. Want to give it a test locally before I hit approve

pkg/alertmanager/multitenant.go
yuri-tceretian: Yes. The struct is private anyway. This is needed in a follow-up PR where I calculate fingerprint.

santihernandezc (approved):
LGTM!

pkg/alertmanager/multitenant.go
santihernandezc: @yuri-tceretian that's this PR :laughing: 
I'll trust that there's another PR to address this :ok_hand:

yuri-tceretian: https://github.com/grafana/mimir/pull/12087

PR #12053, opened 2025-07-11
Improve docs for blocked requests, blocked queries and limited queries config options

charleskorn:
#### What this PR does

This PR improves how the configuration of blocked requests, blocked queries and limited queries is displayed in the docs.

In particular:

* all fields are now shown as they would be for any other config block, rather than relying on a comment
* an example has been added for how to configure a blocked request

To achieve this, I've made two changes to the config doc generator:

* maps of structs are now shown expanded, instead of as `map of string to package.SomeStruct`
* examples (if available) are now shown for all types of fields

I've also changed the implementation of `ExampleDoc` for each type to use the Go type rather than a generic map, so these should stay in-sync with the actual configuration fields. I changed the blocked requests, blocked queries and limited queries slices to not use pointers, as the pointers seemed unnecessary and keeping them would have required further changes to the config doc generator.

(none)

#### Checklist

- [n/a] Tests updated.
- [x] Documentation added.

### DIFF
--- a/docs/sources/mimir/configure/configuration-parameters/index.md
+++ b/docs/sources/mimir/configure/configuration-parameters/index.md
@@ -3759,6 +3759,12 @@ blocked_requests:
 
     [method: <string> | default = ""]
 
+    [query_params:]
+      <string>:
+        [value: <string> | default = ""]
+
+        [is_regexp: <boolean> | default = ]
+
 # Mutate incoming queries to align their start and end with their step to
 # improve result caching.
 # CLI flag: -query-frontend.align-queries-with-step
--- a/docs/sources/mimir/configure/configuration-parameters/index.md
+++ b/docs/sources/mimir/configure/configuration-parameters/index.md
@@ -3720,20 +3720,44 @@ The `limits` block configures default and per-tenant limits imposed by component
 #   Setting the pattern to ".*" and regex to true blocks all queries.
 #   blocked_queries:
 #       - pattern: rate(metric_counter[5m])
+#         regex: false
 #         reason: because the query is misconfigured
-[blocked_queries: <list of pattern (string), regex (bool), and, optionally, reason (string)> | default = ]
+blocked_queries:
+  -
+    [pattern: <string> | default = ""]
+
+    [regex: <boolean> | default = ]
+
+    [reason: <string> | default = ""]
 
 # (experimental) List of queries to limit and duration to limit them for.
 # Example:
 #   The following configuration limits the query "rate(metric_counter[5m])" to
 #   running, at most, every minute.
 #   limited_queries:
-#       - allowed_frequency: 1m
-#         query: rate(metric_counter[5m])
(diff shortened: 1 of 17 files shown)


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3720,20 +3720,50 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+#   parameter is set to 100.
+#   blocked_requests:
+#       - path: /foo
+#         method: GET
+#         query_params:
+#           limit:
+#               value: "100"
+blocked_requests:
+  -
+    [path: <string> | default = ""]
```
dimitarvdimitrov: can we remove the first new line? it looks a bit crooked like this. I wouldn't mind removing all extra new lines either

dimitarvdimitrov (commented):
thanks, left a single nitpick

docs/sources/mimir/configure/configuration-parameters/index.md
charleskorn: How does it look after 8894aba and 6c1cd74? I've added descriptions to everything and changed it to start immediately after the `-` bullet.

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3720,20 +3720,61 @@ The `limits` block configures default and per-tenant limits imposed by component
...
 #   blocked_queries:
 #       - pattern: rate(metric_counter[5m])
+#         regex: false
 #         reason: because the query is misconfigured
-[blocked_queries: <list of pattern (string), regex (bool), and, optionally, reason (string)> | default = ]
+blocked_queries:
+  - # PromQL expression pattern to match.
+    [pattern: <string> | default = ""]
+
+    # If true, pattern is treated as a regular expression. If false, pattern is
```
```suggestion
    # If true, the pattern is treated as a regular expression. If false, the pattern is
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3720,20 +3720,61 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+    [method: <string> | default = ""]
+
+    # Query parameters to match. Requests must have all of the provided query
+    # parameters to be considered a match.
+    [query_params:]
+      <string>:
+        # Value to match.
+        [value: <string> | default = ""]
+
+        # If true, value is treated as a regexp pattern. If false, value is
```
```suggestion
        # If true,the  value is treated as a regular expression pattern. If false, the value is
```

docs/sources/mimir/configure/configuration-parameters/index.md
charleskorn: Fixed in 67fe1f7

docs/sources/mimir/configure/configuration-parameters/index.md
charleskorn: Fixed in 67fe1f7

Issue #11831, opened 2025-06-24
Mimir API doesn't accept limit=0 unlike Prometheus API


dimitarvdimitrov:
## Problem

Mimir's label values API doesn't accept `limit=0` parameter, while Prometheus API does. According to [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values), `limit=0` means "disabled" (no limit).

## Context

- This was changed in PR #7722
- The change was intentional and includes unit tests
- However, it breaks compatibility with Prometheus API behavior
- This has caused issues in production environments

## Expected Behavior

Mimir should accept `limit=0` and treat it as "no limit" to maintain API compatibility with Prometheus.

## Current Behavior

Mimir rejects `limit=0` parameter.

## References

- Prometheus API docs: https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values
- Related PR: https://github.com/grafana/mimir/pull/7722

Reported by: @itsmylife

PR #12054, opened 2025-07-11
Optimise label names API by removing __name__!="" matcher when redundant

pracucci:
#### What this PR does

This PR introduces a new labels query optimizer middleware for the query-frontend that automatically removes
the ` __name__!=""` matcher from label names queries when the matcher is redundant. The PR also includes a bug
  fix for the limit parameter validation in label queries: the value 0 was not supported previously, but it's actually a valid parameter in Prometheus (means unlimited).

I've tested the following matchers, comparing the result with and without the optimizer enabled and I've got the same exact results:
- No matchers
- With `{}` matcher (invalid)
- With `{__name__!=""}` matcher
- With `{__name__!="", container="mimir-write"}` matcher
- With `{__name__!="", __name__="up"}` matcher
- With `{__name__!=""}` and `{__name__=~"a.*"}` matchers
- With `{__name__!="", __name__=~"a.*"}` matcher

This PR only address the label names API. I will then open a PR for the label values.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -5025,7 +5025,7 @@
           "kind": "field",
           "name": "labels_query_optimizer_enabled",
           "required": false,
-          "desc": "Enable labels query optimizations. When enabled, the query-frontend rewrites labels queries in an idempotent way to improve their performance.",
+          "desc": "Enable labels query optimizations. When enabled, the query-frontend may rewrite labels queries to improve their performance.",
           "fieldValue": null,
           "fieldDefaultValue": false,
           "fieldFlag": "query-frontend.labels-query-optimizer-enabled",
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -5023,6 +5023,17 @@
         },
         {
           "kind": "field",
+          "name": "labels_query_optimizer_enabled",
+          "required": false,
+          "desc": "Enable labels query optimizations. When enabled, the query-frontend rewrites labels queries in an idempotent way to improve their performance.",
+          "fieldValue": null,
+          "fieldDefaultValue": false,
+          "fieldFlag": "query-frontend.labels-query-optimizer-enabled",
+          "fieldType": "boolean",
+          "fieldCategory": "experimental"
+        },
+        {
+          "kind": "field",
           "name": "cardinality_analysis_enabled",
           "required": false,
           "desc": "Enables endpoints used for cardinality analysis.",
--- a/pkg/frontend/querymiddleware/codec_test.go
+++ b/pkg/frontend/querymiddleware/codec_test.go
@@ -505,6 +505,7 @@ func TestPrometheusCodec_DecodeEncodeLabelsQueryRequest(t *testing.T) {
 		propagateHeaders          []string
 		url                       string
 		headers                   http.Header
+		expectedURL               string
 		expectedStruct            LabelsSeriesQueryRequest
 		expectedGetLabelName      string
 		expectedGetStartOrDefault int64
@@ -513,8 +514,9 @@ func TestPrometheusCodec_DecodeEncodeLabelsQueryRequest(t *testing.T) {
 		expectedLimit             uint64
 	}{
 		{
-			name: "label names with start and end timestamps, no matcher sets",
-			url:  "/api/v1/labels?end=1708588800&start=1708502400",
+			name:        "label names with start and end timestamps, no matcher sets",
+			url:         "/api/v1/labels?end=1708588800&start=1708502400",
+			expectedURL: "/api/v1/labels?end=1708588800&start=1708502400",
 			expectedStruct: &PrometheusLabelNamesQueryRequest{
 				Path:             "/api/v1/labels",
 				Start:            1708502400 * 1e3,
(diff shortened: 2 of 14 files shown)


pkg/frontend/querymiddleware/labels_query_optimizer.go
```diff
@@ -0,0 +1,173 @@
...
+		if len(matchers) == 0 {
+			return rawMatcherSets, false, nil
+		}
+
+		optimizedMatchers := make([]*labels.Matcher, 0, len(matchers))
+
+		for _, matcher := range matchers {
+			// Filter out `__name__!=""` matcher because all series in Mimir have a metric name
+			// so this matcher matches all series but very expensive to run.
+			if matcher.Name == labels.MetricName && matcher.Type == labels.MatchNotEqual && matcher.Value == "" {
```
bboreham, line 153: We could also do `foo=~".*"`.

bboreham (approved):
lgtm

pkg/frontend/querymiddleware/labels_query_optimizer.go
pracucci: Good point. Will do in the follow up PR, where I add label names support too.

cmd/mimir/config-descriptor.json
```diff
@@ -5021,6 +5021,17 @@
...
           "fieldCategory": "experimental"
         },
+        {
+          "kind": "field",
+          "name": "labels_query_optimizer_enabled",
+          "required": false,
+          "desc": "Enable labels query optimizations. When enabled, the query-frontend rewrites labels queries in an idempotent way to improve their performance.",
+          "fieldValue": null,
+          "fieldDefaultValue": false,
+          "fieldFlag": "query-frontend.labels-query-optimizer-enabled",
```
dimitarvdimitrov, line 5031: should go in about-versioning.md too, right?

pkg/frontend/querymiddleware/codec_test.go
```diff
@@ -664,21 +675,34 @@ func TestPrometheusCodec_DecodeEncodeLabelsQueryRequest(t *testing.T) {
...
+				LabelName: "job",
+				Start:     1708502400 * 1e3,
+				End:       1708588800 * 1e3,
+				Limit:     0,
+			},
+			expectedGetLabelName:      "job",
+			expectedLimit:             0,
+			expectedGetStartOrDefault: 1708502400 * 1e3,
+			expectedGetEndOrDefault:   1708588800 * 1e3,
 		},
```
dimitarvdimitrov, line 692: looks like this fixes https://github.com/grafana/mimir/issues/11831 cc @itsmylife

pkg/util/validation/limits.go
```diff
@@ -447,6 +448,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
 	f.Var(&l.EnabledPromQLExperimentalFunctions, "query-frontend.enabled-promql-experimental-functions", "Enable certain experimental PromQL functions, which are subject to being changed or removed at any time, on a per-tenant basis. Defaults to empty which means all experimental functions are disabled. Set to 'all' to enable all experimental functions.")
 	f.BoolVar(&l.Prom2RangeCompat, "query-frontend.prom2-range-compat", false, "Rewrite queries using the same range selector and resolution [X:X] which don't work in Prometheus 3.0 to a nearly identical form that works with Prometheus 3.0 semantics")
 	f.BoolVar(&l.SubquerySpinOffEnabled, "query-frontend.subquery-spin-off-enabled", false, "Enable spinning off subqueries from instant queries as range queries to optimize their performance.")
+	f.BoolVar(&l.LabelsQueryOptimizerEnabled, "query-frontend.labels-query-optimizer-enabled", false, "Enable labels query optimizations. When enabled, the query-frontend rewrites labels queries in an idempotent way to improve their performance.")
```
dimitarvdimitrov: does the idempotency matter to whoever is operating mimir? the way i read this is that the same optimization can be applied multiple times and will always lead to the same result.

pracucci: What I wanted to explain in the description is that it doesn't change the actual query result. I can remove that part from the CLI flag. I think it just makes it harder to read.

pracucci: Done in https://github.com/grafana/mimir/pull/12054/commits/83776c2cb1cb986777feeab4ee52a3867281696f

cmd/mimir/config-descriptor.json
pracucci: Yes, done in https://github.com/grafana/mimir/pull/12054/commits/83776c2cb1cb986777feeab4ee52a3867281696f

pkg/util/validation/limits.go
dimitarvdimitrov: another thing you can improve is to help people understand when they'd want to turn this on and when they don't. In this case they want it off when they're pushing metrics without `__name__` labels. Sorry i didn't say this the first time around

pracucci: > In this case they want it off when they're pushing metrics without __name__ labels

As discussed offline, in the distributors we enforce that every metric has `__name__`. I think that currently there's no way to write metrics without name.

The reason why it's disabled by default is because I want to gain confidence about it, but my idea would be to switch the default to true once we feel comfortable.

pkg/frontend/querymiddleware/labels_query_optimizer.go
pracucci: Proposed here: https://github.com/grafana/mimir/pull/12080

PR #12060, opened 2025-07-11
distributor: Expermental support for producing ingest-storage records in format V2

alexweav:
#### What this PR does

This PR adds experimental support for `ingest-storage.kafka.producer-record-version=2`. Consumer side support exists already, this allows toggling producers to write the new format.

This implements translation of RW1.0 to the new RW2.0-esque warpstream wire format.

We are not considering the format 100% stabilized, still. We continue to reserve the right to make backward-incompatible changes to the format, particularly the common symbols space.

In order to keep the PR size reasonable (it's already big), this PR omits a few key optimizations from https://github.com/grafana/mimir/compare/main...alexweav/experiment-rw-v2 and focuses more on the plain implementation of the format. Namely, we don't yet pool the RW2.0 WriteRequest fields yet, similar to `PreallocTimeseries`, nor do we use yoloStrings when marshalling. These will be saved for a future PR.

The symbols table is an exception - it has been optimized to be significantly faster than the Prometheus implementation, with near-zero allocations.

The PR is still a net positive even without those optimizations, testing it in a small cell (2.2m in-memory series) still reduces distributor CPU:
2025-07-18-124932_742x298_scrot

Not adding any changelog entries yet. I really don't want to advertise using it yet, I still think the format is likely to break (by changing the common symbols space). We'll add a changelog entry pointing to all these PRs once we're ready to roll it out further.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/mimirpb/symbols_test.go
```diff
@@ -0,0 +1,123 @@
...
+		las[i/2].Value = symbols[labelRefs[i+1]]
+		if las[i/2].Name == labels.MetricName {
+			name = las[i/2].Value
+		}
+	}
+	return las, name
+}
+
+func BenchmarkSymbolizer(b *testing.B) {
+	b.Run("prom symbolizer: 10k labels unique values", func(b *testing.B) {
```
bboreham: Individual sends from distributors to ingesters typically have 7 series, each with say 20 labels, so 280 strings total. That's the sort of ball-park we should be benchmarking.

pracucci: Agree on few series per per-partition write request. Let's say `max_samples_per_send` is 2000 and requests are full. Then we have per-partition write requests within this range:
- A tenant with shard size = 1: 2000 symbols in the per-partition write request
- A tenant with shard size = all ingesters: with 100 partitions that's 200 samples per per-partition write request, with 300 partitions (which is pretty large scale) that's 7, the number Bryan mentioned

alexweav: Gotcha, thank you for the context.

I want to move the benchmark up a level, using one benchmark to cover the whole flow for writes including both marshalling and unmarshalling. When I do that, I'll also make the request look more realistic.

alexweav: I've updated this benchmark to be more realistic (a case for both bryan's and marco's scenarios). It's also not coupled to the implementation so much anymore, thus easier to use with `benchstat` and iterate on.

pkg/mimirpb/symbols_test.go
```diff
@@ -0,0 +1,238 @@
...
+		las[i/2].Name = symbols[labelRefs[i]]
+		las[i/2].Value = symbols[labelRefs[i+1]]
+		if las[i/2].Name == labels.MetricName {
+			name = las[i/2].Value
+		}
+	}
+	return las, name
+}
+
+func BenchmarkSymbolizer(b *testing.B) {
```
alexweav, line 153: Benchmark results for - the upstream Prom symbolizer, the FastSymbolTable without memory pooling, and the FastSymbolTable with pooling:

```
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/mimirpb
cpu: AMD Ryzen 7 PRO 5850U with Radeon Graphics
                                                                   │ prom-symbol-table.txt │        fast-symbol-table.txt        │    fast-symbol-table-pooling.txt    │
                                                                   │        sec/op         │   sec/op     vs base                │   sec/op     vs base                │
Symbolizer/2k_labels_unique_values,_common_symbols-16                          273.3µ ± 1%   316.3µ ± 2%  +15.74% (p=0.000 n=10)   129.3µ ± 3%  -52.67% (p=0.000 n=10)
Symbolizer/7_series_with_some_shared_label_keys,_common_symbols-16             16.80µ ± 1%   23.95µ ± 1%  +42.57% (p=0.000 n=10)   14.99µ ± 6%  -10.78% (p=0.000 n=10)
geomean                                                                        67.75µ        87.03µ       +28.46%                  44.02µ       -35.02%

                                                                   │ prom-symbol-table.txt │         fast-symbol-table.txt          │   fast-symbol-table-pooling.txt    │
                                                                   │         B/op          │      B/op       vs base                │    B/op     vs base                │
Symbolizer/2k_labels_unique_values,_common_symbols-16                       318960.00 ± 0%   250749.00 ± 0%  -21.39% (p=0.000 n=10)   50.50 ± 3%  -99.98% (p=0.000 n=10)
Symbolizer/7_series_with_some_shared_label_keys,_common_symbols-16           17976.00 ± 0%    15560.00 ± 0%  -13.44% (p=0.000 n=10)   24.00 ± 0%  -99.87% (p=0.000 n=10)
geomean                                                                       73.95Ki          61.00Ki       -17.51%                  34.81       -99.95%

                                                                   │ prom-symbol-table.txt │        fast-symbol-table.txt        │   fast-symbol-table-pooling.txt    │
                                                                   │       allocs/op       │  allocs/op   vs base                │ allocs/op   vs base                │
Symbolizer/2k_labels_unique_values,_common_symbols-16                          44.000 ± 0%   30.000 ± 0%  -31.82% (p=0.000 n=10)   1.000 ± 0%  -97.73% (p=0.000 n=10)
Symbolizer/7_series_with_some_shared_label_keys,_common_symbols-16             21.000 ± 0%   12.000 ± 0%  -42.86% (p=0.000 n=10)   1.000 ± 0%  -95.24% (p=0.000 n=10)
geomean                                                                         30.40         18.97       -37.58%                  1.000       -96.71%
```

alexweav: I included how the table performs without any pooling for a more apples to apples comparison.

Ironically, `FastSymbolsTable` is slightly slower in that benchmark :laughing: but the diff is entirely because it's doing extra work searching the common symbols, which the prom implementation does not have support for. With common symbols disabled it's slightly faster. Still, fewer allocations is nice.

And, when using all the pooling features like we would in real life, it's significantly faster. The goal was to explain why I didn't just pool prom symbolizers, but rather rolled our own.

pkg/mimirpb/symbols.go
```diff
@@ -0,0 +1,129 @@
...
+	t.commonSymbols = commonSymbols
+}
+
+func (t *FastSymbolsTable) Symbolize(str string) uint32 {
+	if str == "" {
+		// 0 means empty string, even if an offset is provided.
+		return 0
+	}
+	if t.commonSymbols != nil {
+		// TODO: CommonSymbols is bounded size, it small enough to where linear search is faster?
```
alexweav, line 103: I'm still on the fence about it. Especially if we change the length of the common symbols space, it would completely invalidate any decision here, so I think we should defer it for now.

pkg/mimirpb/symbols_test.go
```diff
@@ -0,0 +1,238 @@
...
+				b.Fatalf("unexpected number of symbols: %d", len(symbols))
+			}
+			reuseSymbolsSlice(symbols)
+			reuseSymbolsTable(st)
+		}
+	})
+}
+
+var (
+	benchmarkCommonSymbols = []string{
```
alexweav, line 218: copied due to package dependency cycles. it's just local to a benchmark, we don't really care much if it drifts

pkg/storage/ingest/version.go
```diff
@@ -139,6 +144,33 @@ func (v versionOneRecordSerializer) ToRecords(partitionID int32, tenantID string
...
+type versionTwoRecordSerializer struct{}
+
+func (v versionTwoRecordSerializer) ToRecords(partitionID int32, tenantID string, req *mimirpb.WriteRequest, maxSize int) ([]*kgo.Record, error) {
+	reqv2, err := mimirpb.FromWriteRequestToRW2Request(req, V2CommonSymbols, V2RecordSymbolOffset)
+	defer mimirpb.ReuseRW2(reqv2)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to convert RW1 request to RW2")
+	}
+
+	// TODO: V1 contains logic that splits large records up across multiple records if they exceed maxSize.
```
pracucci, line 156: This is a requirement to go to ops or prod. I'm fine doing it in a separate PR, but don't rollout without the splitting in any cell where we can't afford to break writes. The reason is that Kafka records payload have a hard limit in size. There are some edge cases where the serialised request size goes above that limit, and we need to split it.

pkg/storage/ingest/version_test.go
```diff
@@ -205,6 +206,51 @@ func TestDeserializeRecordContent(t *testing.T) {
...
+
+		resultReq := &mimirpb.PreallocWriteRequest{}
+		err = DeserializeRecordContent(record.Value, resultReq, 2)
+		require.NoError(t, err)
+
+		require.Equal(t, req.Timeseries, resultReq.Timeseries[0:2])
+		require.Nil(t, resultReq.SymbolsRW2)
+		require.Nil(t, resultReq.TimeseriesRW2)
+		// Metadata order is currently not preserved by the RW2 deser layer.
+		assertMetadataEqualUnordered(t, req.Metadata, resultReq.Metadata)
```
pracucci: Isn't it enough? `ElementsMatch()` ignores ordering.

```suggestion
		assert.ElementsMatch(t, req.Metadata, resultReq.Metadata)
```

pkg/storage/ingest/version_test.go
```diff
@@ -205,6 +206,51 @@ func TestDeserializeRecordContent(t *testing.T) {
...
+		require.Len(t, records, 1)
+		record := records[0]
+
+		require.Equal(t, 2, ParseRecordVersion(record))
+
+		resultReq := &mimirpb.PreallocWriteRequest{}
+		err = DeserializeRecordContent(record.Value, resultReq, 2)
+		require.NoError(t, err)
+
+		require.Equal(t, req.Timeseries, resultReq.Timeseries[0:2])
```
pracucci, line 247: I have a doubt here. You're excluding the last 3 timeseries because they're the metadata, but the series (with no samples) are still preserved in `resultReq.Timeseries`. This unmarshalling logic was introduced in https://github.com/grafana/mimir/pull/11100 by @krajorama, and I'm wondering if that's the desired behaviour (I mean, keeping the "metadata" series in `Timeseries`). If no, we should address it in a follow up PR, if yes then I would improve this unit test to actually compare the entire `resultReq.Timeseries`.

From this unit test perspective, just asserting on a subset of the data doesn't look super robust to me, because there could also be some spurious data at the end of the `Timeseries` slice that we're not checking here.

pkg/mimirpb/symbols.go
```diff
@@ -0,0 +1,129 @@
...
+			return make([]string, 0, baseSymbolsSliceCapacity)
+		},
+	}
+)
+
+func symbolsTableFromPool() *FastSymbolsTable {
+	return symbolsTablePool.Get().(*FastSymbolsTable)
+}
+
+func reuseSymbolsTable(t *FastSymbolsTable) {
```
pracucci, line 41: A recurring mistake we do is using pools that can be subject to poising by large data type recycled. I recommend you to define a max size (hardcoded) of the map to be considered eligible for pooling. If it's bigger than that, then do not recycle it.

Same comment applies to `reuseSymbolsSlice()`.

pkg/mimirpb/symbols.go
pracucci: When I benchmarked it for `FastRegexMatcher`, the sweet spot there was 16 items. Looking up a map of <= 16 items was slower than iterating a slice. Above that size, the map lookup was faster.  But that's something that should be easy to benchmark. Can be discussed separately, no need to do it in this PR.

pkg/mimirpb/symbols.go
```diff
@@ -0,0 +1,129 @@
...
+			if str == t.commonSymbols[i] {
+				return uint32(i)
+			}
+		}
+	}
+
+	if ref, ok := t.symbolsMap[str]; ok {
+		return ref
+	}
+	ref := uint32(len(t.symbolsMap)) + t.offset + 1
```
pracucci: [nit] Do you mind adding a comment to explain that the +1 is because the ref index start from 1, because 0 is reserved for the empty string, please?

pkg/mimirpb/symbols.go
```diff
@@ -0,0 +1,129 @@
...
+func (t *FastSymbolsTable) SymbolsPrealloc(prealloc []string) []string {
+	if cap(prealloc) < t.CountSymbols() {
+		prealloc = make([]string, 0, t.CountSymbols())
+	}
+	for range len(t.symbolsMap) + 1 {
+		prealloc = append(prealloc, "")
+	}
+
+	for k, v := range t.symbolsMap {
+		(prealloc)[v-t.offset] = k
```
pracucci: [nit]
```suggestion
		prealloc[v-t.offset] = k
```

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -0,0 +1,105 @@
...
+			Samples:          ts.Samples,
+			Histograms:       ts.Histograms,
+			Exemplars:        FromExemplarsToExemplarsRW2(ts.Exemplars, symbols),
+			Metadata:         MetadataRW2{},
+			CreatedTimestamp: ts.CreatedTimestamp,
+		})
+	}
+
+	// Represent metadata as extra, empty timeseries rather than attaching it to existing series.
+	// It prevents duplication, and removes the need to match metadata to series, with very low actual size overhead.
```
pracucci: [nit] What do you mean by "it prevents duplication"? Series are actually duplicated when metadata is present. I'm fine with this approach, I just don't understand the statement here.

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -0,0 +1,105 @@
...
+package mimirpb
+
+import "fmt"
+
+func ReuseRW2(req *WriteRequest) {
+	reuseSymbolsSlice(req.SymbolsRW2)
+}
+
+// FromWriteRequestToRW2Request converts a write request with RW1 fields populated to a write request with RW2 fields populated.
+// It makes a new RW2 request, leaving the original request alone - it is still up to the caller to free the provided request.
```
pracucci, line 12: [nit] Clarify that the some references from the input request are preserved, so it's not safe to free / reuse the input the request data structures until the returned request is still in use.

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -0,0 +1,105 @@
...
+	result := MetadataRW2{
+		Type: MetadataRW2_MetricType(metadata.Type),
+	}
+
+	if metadata.Help != "" {
+		result.HelpRef = symbols.Symbolize(metadata.Help)
+	}
+	if metadata.Unit != "" {
+		result.UnitRef = symbols.Symbolize(metadata.Unit)
+	}
```
pracucci: [nit] Is there a significant performance penalty (due to the function call) if we do this 👇 ?

```suggestion
	result.HelpRef = symbols.Symbolize(metadata.Help)
	result.UnitRef = symbols.Symbolize(metadata.Unit)
```

pkg/mimirpb/prealloc_rw2_test.go
```diff
@@ -0,0 +1,463 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
pracucci, line 1: Look at the test `TestSplitWriteRequestByMaxMarshalSize_WriteRequestHasChanged` in this package. I think we want something similar for `FromWriteRequestToRW2Request()` too.

pracucci (approved):
Nice job! I left few comments here and there, but all minor things.

pkg/mimirpb/prealloc_rw2.go
alexweav: Our request might contain 10 TimeSeries with the same `__name__` but different labels. The RW2.0 convention (but not requirement) is to repeat the same metadata on every series with that name, the rationale being that the metadata is cheaper than before.

I'm specifically avoiding this for ingest-storage, omitting any repetition of metadata in order to squeeze extra bits per timeseries out of it.

alexweav: Updated wording.

pkg/mimirpb/symbols.go
alexweav: There is unfortunately no `cap()` of map in golang (https://github.com/golang/go/issues/52157) but metadata like this was part of the reason to pool the entire struct instead of just the map directly.

Since the table uniquely owns the map I think we can make our own "cap" by remembering the maximum size in the symbol table and not clearing that field when we return it to the pool.

pkg/storage/ingest/version.go
alexweav: Let's separate it. If possible I'd like to try and refactor a bit, to reuse what I can from the existing splitting logic.

pkg/storage/ingest/version_test.go
alexweav: Exactly, I want to go touch the deserialization logic to drop the timeseries, if it was only used to carry metadata and no data.

That being said, the missing test coverage was an oversight, I should have put a note here.

pkg/storage/ingest/version.go
alexweav: I am addressing this here: https://github.com/grafana/mimir/pull/12166

pkg/mimirpb/symbols.go
```diff
@@ -0,0 +1,157 @@
...
+	Symbolize(str string) uint32
+	Symbols() []string
+	Reset()
+}
+
+var _ StringSymbolizer = &writev2.SymbolsTable{}
+
+var (
+	minPreallocatedSymbolsPerRequest = 40
+	maxPreallocatedSymbolsPerRequest = 400
```
pracucci: I would make them constants. Also `maxPreallocatedSymbolsPerRequest = 400` looks very low. We have requests with 2K series, and for each series we may have 20-30 labels. I think this number should be much higher, otherwise the symbols table will be never reused. We just want to protect from outliers here.

alexweav: Agreed, I missed that the equivalent in `timeseries.go` was **per timeseries** and not per request

alexweav: Bumped to 100k. (2k series @ max 40 labels per series = 80k, also needs room for metadata).
I plan to do some analysis of real traffic which can help tune this further.

PR #12076, opened 2025-07-14
Query-frontend: fix auth propagation in labels query optimizer

pracucci:
#### What this PR does

In https://github.com/grafana/mimir/pull/12054 and https://github.com/grafana/mimir/pull/12066 I've done a mistake that I haven't noticed when testing it in the local dev env: the `X-Scope-OrgID` was not propagated when a query is optimised. In this PR I'm fixing it by using `user.InjectOrgIDIntoHTTPRequest()`: this follows the same strategy used in other places in the query-frontend (i.e. instant/range queries, and active series sharding). In addition, I'm also adding an integration test, that tests idempotency of the optimizer comparing label names and values query responses with and without optimizer enabled.

https://github.com/grafana/mimir/pull/12054 was merged last week: should we back port the fix to the last week release? No, because the optimizer is experimental and disabled by default. When disabled, no request is rewritten, and so the auth bug doesn't show up.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo (approved):
Looks good to me 🔥

pkg/frontend/querymiddleware/labels_query_optimizer.go
```diff
@@ -125,6 +126,10 @@ func (l *labelsQueryOptimizer) optimizeRequest(ctx context.Context, req *http.Re
 		return l.next.RoundTrip(req)
 	}
 
+	if err := user.InjectOrgIDIntoHTTPRequest(ctx, optimizedReq); err != nil {
+		return nil, apierror.New(apierror.TypeBadData, err.Error())
+	}
```
dimitarvdimitrov, line 131: nitpick to do this in the EncodeLabelsSeriesQueryRequest so we don’t hit the same problem next time we encode these requests

pracucci: https://github.com/grafana/mimir/pull/12100

PR #12077, opened 2025-07-14
[release-2.17] Update dskit to skip tracing configuration when no env is defined (

colega:
Backport of #12074

This updates dskit to include https://github.com/grafana/dskit/pull/723


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.17.0-rc.1
```
zenador: I haven't actually released the rc yet, so this should be under rc.0

colega: Oh, ok.

colega: Updated (moved the PR under the existing feature entry)

zenador: Thanks! The changelog check is failing as the current PR number is not used. I guess in this case we can just apply the changelog not needed label?

zenador (approved):

PR #12080, opened 2025-07-14
Query-frontend: optimise labels API by removing =~'.*' matcher when redundant

pracucci:
#### What this PR does

This is a follow up of https://github.com/grafana/mimir/pull/12054 and https://github.com/grafana/mimir/pull/12066, implementing the suggestion received by @bboreham [here](https://github.com/grafana/mimir/pull/12054#discussion_r2200625105). The idea is to filter out `label=~".*"` matcher when it's redundant: in Prometheus 3 (and Mimir now) `.*` matches any string, including new line, so the `.*` matcher literally matches every series.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/labels_query_optimizer.go
+++ b/pkg/frontend/querymiddleware/labels_query_optimizer.go
@@ -149,13 +149,18 @@ func optimizeLabelsRequestMatchers(rawMatcherSets []string) (_ []string, optimiz
 		hasNonEmptyMatchers := false
 
 		for _, matcher := range matchers {
+			// Before filtering out any matcher we should check if the among the original matchers
+			// there's anyone not matching the empty string. This will be used later to ensure there's
+			// at least one non-empty matcher.
+			if !matcher.Matches("") {
+				hasNonEmptyMatchers = true
+			}
+
 			// Filter out a matcher that matches any string because it matches all series.
 			if matcher.Type == labels.MatchRegexp && matcher.Value == ".*" {
 				optimized = true
 				continue
 			}
-
-			hasNonEmptyMatchers = true
 
 			// Filter out `__name__!=""` matcher because all series in Mimir have a metric name
 			// so this matcher matches all series but very expensive to run.


CHANGELOG.md
```diff
@@ -11,7 +11,7 @@
...
-* [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. The optimizer can be enabled per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076
+* [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. The optimizer can be enabled per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076 #12080
```
```suggestion
* [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. You can enable the optimizer per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076 #12080
```

tacole02 (approved):

pkg/frontend/querymiddleware/labels_query_optimizer.go
```diff
@@ -145,16 +145,18 @@ func optimizeLabelsRequestMatchers(rawMatcherSets []string) (_ []string, optimiz
...
+		hasNonEmptyMatchers := false
 
 		for _, matcher := range matchers {
+			// Filter out a matcher that matches any string because it matches all series.
+			if matcher.Type == labels.MatchRegexp && matcher.Value == ".*" {
+				optimized = true
+				continue
+			}
+
+			hasNonEmptyMatchers = true
```
bboreham: I didn't get why this is here.  Maybe a comment would help?
It seems you're making a distinction between `foo=~".*"` and all other matchers.
Please explain why that is required.

pkg/frontend/querymiddleware/labels_query_optimizer.go
```diff
@@ -166,14 +168,21 @@ func optimizeLabelsRequestMatchers(rawMatcherSets []string) (_ []string, optimiz
 			optimizedMatchers = append(optimizedMatchers, matcher)
 		}
 
-		if len(optimizedMatchers) > 0 {
-			optimizedRawMatcherSets = append(optimizedRawMatcherSets, util.LabelMatchersToString(optimizedMatchers))
-		} else {
-			// It was not an empty matcher (see condition above), and all the matchers have been removed.
+		if !hasNonEmptyMatchers {
```
dimitarvdimitrov, line 176: if the request is `{foo=~".*", bar=~".*"}` then your implementation would forward these to downstream. Shouldn’t it remove both instead?

pracucci: No, because `match[]={foo=~".*", bar=~".*"}` is invalid (they both match the empty string and Prometheus API requires at least non-empty matcher). So I don't want to manipulate the request in that case, but forward it downstream to get the actual error message. If I remove them, it wouldn't be an idempotent change, because removing them means "match all series" which is different than having the API returning an error.

pkg/frontend/querymiddleware/labels_query_optimizer.go
pracucci: I tried to explain it a couple of lines below:
https://github.com/grafana/mimir/pull/12080/files#diff-50affd3ccdca209af579be130df3956b4c72c5e3caaf175f37765e29e6f2366dR172-R174

I guess the intent is not clear.

pkg/frontend/querymiddleware/labels_query_optimizer.go
dimitarvdimitrov: oh, i see the test case and comment now

pkg/frontend/querymiddleware/labels_query_optimizer.go
pracucci: I tried to improve it here: https://github.com/grafana/mimir/pull/12080/commits/99d8a261b266ee3e0aea6ec19afe0b5047490c9b

PR #12082, opened 2025-07-14
Test potential fix for HA tracker failovers on r348

zenador:
Not to be merged, just running tests in CI, using it as a base for vendoring and getting feedback.

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -41,6 +41,7 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/relabel"
+	"github.com/prometheus/prometheus/model/timestamp"
 	"github.com/prometheus/prometheus/scrape"
 	"github.com/prometheus/prometheus/storage"
 	"go.opentelemetry.io/otel"
@@ -785,7 +786,7 @@ func (d *Distributor) stopping(_ error) error {
 // Returns a boolean that indicates whether or not we want to remove the replica label going forward,
 // and an error that indicates whether we want to accept samples based on the cluster/replica found in ts.
 // nil for the error means accept the sample.
-func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica string) (removeReplicaLabel bool, _ error) {
+func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica string, ts int64) (removeReplicaLabel bool, _ error) {
 	// If the sample doesn't have either HA label, accept it.
 	// At the moment we want to accept these samples by default.
 	if cluster == "" || replica == "" {
@@ -799,7 +800,9 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica 
 
 	// At this point we know we have both HA labels, we should lookup
 	// the cluster/instance here to see if we want to accept this sample.
-	err := d.HATracker.checkReplica(ctx, userID, cluster, replica, time.Now())
+	// Convert the timestamp to a time.Time for checking the replica
+	sampleTime := timestamp.Time(ts)
+	err := d.HATracker.checkReplica(ctx, userID, cluster, replica, time.Now(), sampleTime)
 	// checkReplica would have returned an error if there was a real error talking to Consul,
 	// or if the replica is not the currently elected replica.
 	if err != nil { // Don't accept the sample.
@@ -1042,11 +1045,33 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
 		numSamples := 0
 		now := time.Now()
 		group := d.activeGroups.UpdateActiveGroupTimestamp(userID, validation.GroupLabel(d.limits, userID, req.Timeseries), now)
+		sampleTimestamp := timestamp.FromTime(now)
+		if d.limits.HATrackerUseSampleTimeForFailover(userID) {
+			earliestSampleTimestamp := sampleTimestamp
+			latestSampleTimestamp := sampleTimestamp
+			for _, ts := range req.Timeseries {
+				if len(ts.Samples) > 0 {
+					tsms := ts.Samples[0].TimestampMs
+					if tsms < earliestSampleTimestamp {
+						earliestSampleTimestamp = tsms
+					} else if tsms > latestSampleTimestamp {
+						latestSampleTimestamp = tsms
+					}
+				}
+			}
+			delta := now.UnixMilli() - earliestSampleTimestamp
+			if delta > 10000 {
+				delta2 := sampleTimestamp - earliestSampleTimestamp
+				delta3 := latestSampleTimestamp - now.UnixMilli()
+				level.Warn(d.log).Log("msg", "earliest sample timestamp is a lot earlier than now", "delta", delta, "delta2", delta2, "delta3", delta3, "user", userID, "cluster", cluster, "replica", replica, "earliestSampleTime", timestamp.Time(earliestSampleTimestamp), "sampleTime", timestamp.Time(sampleTimestamp), "latestSampleTime", timestamp.Time(latestSampleTimestamp), "now", now)
+			}
+			sampleTimestamp = earliestSampleTimestamp
+		}
 		for _, ts := range req.Timeseries {
 			numSamples += len(ts.Samples) + len(ts.Histograms)
 		}
 
-		removeReplica, err := d.checkSample(ctx, userID, cluster, replica)
+		removeReplica, err := d.checkSample(ctx, userID, cluster, replica, sampleTimestamp)
 		if err != nil {
 			if errors.As(err, &replicasDidNotMatchError{}) {
 				// These samples have been deduped.
--- a/pkg/distributor/ha_tracker.go
+++ b/pkg/distributor/ha_tracker.go
@@ -38,13 +38,14 @@ type haTrackerLimits interface {
 	// HATrackerTimeouts returns timeouts that may be specific for the user.
 	HATrackerTimeouts(user string) (update time.Duration, updateJitterMax time.Duration, failover time.Duration)
 	DefaultHATrackerUpdateTimeout() time.Duration
+	HATrackerUseSampleTimeForFailover(user string) bool
 }
 
 type haTracker interface {
 	services.Service
 	http.Handler
 
-	checkReplica(ctx context.Context, userID, cluster, replica string, now time.Time) error
+	checkReplica(ctx context.Context, userID, cluster, replica string, now, sampleTime time.Time) error
 	cleanupHATrackerMetricsForUser(userID string)
 }
 
@@ -207,10 +208,11 @@ type defaultHaTracker struct {
 
 // For one cluster, the information we need to do ha-tracking.
 type haClusterInfo struct {
-	elected                     ReplicaDesc // latest info from KVStore
-	electedLastSeenTimestamp    int64       // timestamp in milliseconds
-	nonElectedLastSeenReplica   string
-	nonElectedLastSeenTimestamp int64 // timestamp in milliseconds
+	elected                      ReplicaDesc // latest info from KVStore
+	electedLastSeenTimestamp     int64       // timestamp in milliseconds
+	nonElectedLastSeenReplica    string
+	nonElectedLastSeenTimestamp  int64 // timestamp in milliseconds
+	nonElectedLastSeenSampleTime int64 // last seen earliest sample time in milliseconds
 }
 
 // newHaTracker returns a new HA cluster tracker using either Consul,
@@ -453,20 +455,27 @@ func (h *defaultHaTracker) updateKVStoreAll(ctx context.Context, now time.Time) 
 			}
 			var replica string
 			var receivedAt int64
+			var sampleTime time.Time
 			if uh.withinUpdateTimeout(now, entry.electedLastSeenTimestamp) {
 				// We have seen the elected replica recently; carry on with that choice.
 				replica = entry.elected.Replica
 				receivedAt = entry.electedLastSeenTimestamp
+				sampleTime = now
 			} else if uh.withinUpdateTimeout(now, entry.nonElectedLastSeenTimestamp) {
 				// Not seen elected but have seen another: attempt to fail over.
 				replica = entry.nonElectedLastSeenReplica
 				receivedAt = entry.nonElectedLastSeenTimestamp
+				if uh.useSampleTimeForFailover {
+					sampleTime = timestamp.Time(entry.nonElectedLastSeenSampleTime)
+				} else {
+					sampleTime = now
+				}
 			} else {
 				continue // we don't have any recent timestamps
 			}
 			// Release lock while we talk to KVStore, which could take a while.
 			h.electedLock.RUnlock()
-			err := uh.updateKVStore(ctx, cluster, replica, now, receivedAt)
+			err := uh.updateKVStore(ctx, cluster, replica, now, sampleTime, receivedAt, true)
 			h.electedLock.RLock()
 			if err != nil {
 				// Failed to store - log it but carry on
@@ -556,7 +565,7 @@ func (h *defaultHaTracker) cleanupOldReplicas(ctx context.Context, deadline time
 // Updates to and from the KV store are handled in the background, except
 // if we have no cached data for this cluster in which case we create the
 // record and store it in-band.
-func (h *defaultHaTracker) checkReplica(ctx context.Context, userID, cluster, replica string, now time.Time) error {
+func (h *defaultHaTracker) checkReplica(ctx context.Context, userID, cluster, replica string, now, sampleTime time.Time) error {
 	h.electedLock.Lock()
 	if entry := h.clusters[userID][cluster]; entry != nil {
 		var err error
@@ -567,6 +576,7 @@ func (h *defaultHaTracker) checkReplica(ctx context.Context, userID, cluster, re
 			// Sample received is from non-elected replica: record details and reject.
 			entry.nonElectedLastSeenReplica = replica
 			entry.nonElectedLastSeenTimestamp = timestamp.FromTime(now)
+			entry.nonElectedLastSeenSampleTime = timestamp.FromTime(sampleTime)
 			err = newReplicasDidNotMatchError(replica, entry.elected.Replica)
 		}
 		h.electedLock.Unlock()
@@ -582,21 +592,22 @@ func (h *defaultHaTracker) checkReplica(ctx context.Context, userID, cluster, re
 	}
 
 	uh := h.forUser(userID)
-	err := uh.updateKVStore(ctx, cluster, replica, now, now.UnixMilli())
+	err := uh.updateKVStore(ctx, cluster, replica, now, sampleTime, now.UnixMilli(), false)
 	if err != nil {
 		level.Error(h.logger).Log("msg", "failed to update KVStore - rejecting sample", "err", err)
 		return err
 	}
 	// Cache will now have the value - recurse to check it again.
-	return h.checkReplica(ctx, userID, cluster, replica, now)
+	return h.checkReplica(ctx, userID, cluster, replica, now, sampleTime)
 }
 
 type defaultHaTrackerForUser struct {
 	*defaultHaTracker
-	userID              string
-	updateTimeout       time.Duration
-	updateTimeoutJitter time.Duration
-	failoverTimeout     time.Duration
+	userID                   string
+	updateTimeout            time.Duration
+	updateTimeoutJitter      time.Duration
+	failoverTimeout          time.Duration
+	useSampleTimeForFailover bool
 }
 
 func (h *defaultHaTracker) forUser(userID string) defaultHaTrackerForUser {
@@ -613,6 +624,8 @@ func (h *defaultHaTracker) forUser(userID string) defaultHaTrackerForUser {
 		computeJitter = computeUpdateTimeoutJitter
 	}
 	uh.updateTimeoutJitter = computeJitter(updateJitterMax)
+
+	uh.useSampleTimeForFailover = h.limits.HATrackerUseSampleTimeForFailover(userID)
 
 	return uh
 }
@@ -653,7 +666,7 @@ func (h *defaultHaTracker) updateCache(userID, cluster string, desc *ReplicaDesc
 
 // If we do set the value then err will be nil and desc will contain the value we set.
 // If there is already a valid value in the store, return nil, nil.
-func (h *defaultHaTrackerForUser) updateKVStore(ctx context.Context, cluster, replica string, now time.Time, receivedAt int64) error {
+func (h *defaultHaTrackerForUser) updateKVStore(ctx context.Context, cluster, replica string, now, sampleTime time.Time, receivedAt int64, fromAll bool) error {
 	key := fmt.Sprintf("%s/%s", h.userID, cluster)
 	var desc *ReplicaDesc
 	var electedAtTime, electedChanges int64
@@ -676,11 +689,17 @@ func (h *defaultHaTrackerForUser) updateKVStore(ctx context.Context, cluster, re
 
 			// If our replica is different, wait until the failover time
 			if desc.Replica != replica {
-				if now.Sub(timestamp.Time(desc.ReceivedAt)) < h.failoverTimeout {
-					level.Info(h.logger).Log("msg", "replica differs, but it's too early to failover", "user", h.userID, "cluster", cluster, "replica", replica, "elected", desc.Replica, "received_at", timestamp.Time(desc.ReceivedAt))
+				var currentTime time.Time
+				if h.useSampleTimeForFailover {
+					currentTime = sampleTime
+				} else {
+					currentTime = now
+				}
+				if currentTime.Sub(timestamp.Time(desc.ReceivedAt)) < h.failoverTimeout {
+					level.Info(h.logger).Log("msg", "replica differs, but it's too early to failover", "user", h.userID, "cluster", cluster, "replica", replica, "elected", desc.Replica, "received_at", timestamp.Time(desc.ReceivedAt), "current_time", currentTime, "use_sample_time_for_failover", h.useSampleTimeForFailover, "from_all", fromAll)
 					return nil, false, nil
 				}
-				level.Info(h.logger).Log("msg", "replica differs, attempting to update kv", "user", h.userID, "cluster", cluster, "replica", replica, "elected", desc.Replica, "received_at", timestamp.Time(desc.ReceivedAt))
+				level.Info(h.logger).Log("msg", "replica differs, attempting to update kv", "user", h.userID, "cluster", cluster, "replica", replica, "elected", desc.Replica, "received_at", timestamp.Time(desc.ReceivedAt), "current_time", currentTime, "use_sample_time_for_failover", h.useSampleTimeForFailover, "from_all", fromAll)
 				electedAtTime = timestamp.FromTime(now)
 				electedChanges = desc.ElectedChanges + 1
 			}
--- a/pkg/distributor/ha_tracker_test.go
+++ b/pkg/distributor/ha_tracker_test.go
@@ -303,18 +303,18 @@ func TestHaTrackerWithMemberList(t *testing.T) {
 	now := time.Now()
 
 	// Write the first time.
-	err = tracker.checkReplica(context.Background(), "user", cluster, replica1, now)
+	err = tracker.checkReplica(context.Background(), "user", cluster, replica1, now, now)
 	assert.NoError(t, err)
 
 	// Throw away a sample from replica2.
-	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now)
+	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now, now)
 	assert.Error(t, err)
 
 	// Wait more than the overwrite timeout.
 	now = now.Add(failoverTimeoutPlus100ms)
 
 	// Another sample from replica2 to update its timestamp.
-	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now)
+	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now, now)
 	assert.Error(t, err)
 
 	// Update KVStore - this should elect replica 2.
@@ -323,11 +323,11 @@ func TestHaTrackerWithMemberList(t *testing.T) {
 	checkReplicaTimestamp(t, 100*time.Millisecond, tracker, "user", cluster, replica2, now, now)
 
 	// Now we should accept from replica 2.
-	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now)
+	err = tracker.checkReplica(context.Background(), "user", cluster, replica2, now, now)
 	assert.NoError(t, err)
 
 	// We timed out accepting samples from replica 1 and should now reject them.
-	err = tracker.checkReplica(context.Background(), "user", cluster, replica1, now)
+	err = tracker.checkReplica(context.Background(), "user", cluster, replica1, now, now)
 	assert.Error(t, err)
 }
 
@@ -386,7 +386,7 @@ func TestHaTrackerWithMemberlistWhenReplicaDescIsMarkedDeletedThenKVStoreUpdateI
 	now := time.Now()
 
 	// Write the first time.
-	err = tracker.checkReplica(context.Background(), tenant, cluster, replica1, now)
+	err = tracker.checkReplica(context.Background(), tenant, cluster, replica1, now, now)
 	assert.NoError(t, err)
 
 	key := fmt.Sprintf("%s/%s", tenant, cluster)
@@ -407,11 +407,11 @@ func TestHaTrackerWithMemberlistWhenReplicaDescIsMarkedDeletedThenKVStoreUpdateI
 
 	now = now.Add(failoverTimeoutPlus100ms)
 	// check replica2
-	err = tracker.checkReplica(context.Background(), tenant, cluster, replica2, now)
+	err = tracker.checkReplica(context.Background(), tenant, cluster, replica2, now, now)
 	assert.NoError(t, err)
 
 	// check replica1
-	assert.ErrorAs(t, tracker.checkReplica(context.Background(), tenant, cluster, replica1, now), &replicasDidNotMatchError{})
+	assert.ErrorAs(t, tracker.checkReplica(context.Background(), tenant, cluster, replica1, now, now), &replicasDidNotMatchError{})
 }
 
 func TestHATrackerCacheSyncOnStart(t *testing.T) {
@@ -447,7 +447,7 @@ func TestHATrackerCacheSyncOnStart(t *testing.T) {
 	assert.Equal(t, 0, int(mockCountingClient.GetCalls.Load()))
 
 	now = time.Now()
-	err = c.checkReplica(context.Background(), "user", cluster, replicaOne, now)
+	err = c.checkReplica(context.Background(), "user", cluster, replicaOne, now, now)
 	assert.NoError(t, err)
 
 	err = services.StopAndAwaitTerminated(context.Background(), c)
@@ -475,7 +475,7 @@ func TestHATrackerCacheSyncOnStart(t *testing.T) {
 	assert.Equal(t, 1, int(mockCountingClient.GetCalls.Load()))
 
 	now = time.Now()
-	err = c.checkReplica(context.Background(), "user", cluster, replicaTwo, now)
+	err = c.checkReplica(context.Background(), "user", cluster, replicaTwo, now, now)
 	assert.Error(t, err)
 	assert.ErrorIs(t, err, replicasDidNotMatchError{replica: "r2", elected: "r1"})
 }
@@ -506,7 +506,7 @@ func TestHATrackerWatchPrefixAssignment(t *testing.T) {
 	// Write the first time.
 	now := time.Now()
 
-	err = c.checkReplica(context.Background(), "user", cluster, replica, now)
+	err = c.checkReplica(context.Background(), "user", cluster, replica, now, now)
 	assert.NoError(t, err)
 
 	// Check to see if the value in the trackers cache is correct.
@@ -533,18 +533,18 @@ func TestHATrackerCheckReplicaOverwriteTimeout(t *testing.T) {
 	now := time.Now()
 
 	// Write the first time.
-	err = c.checkReplica(context.Background(), "user", "test", replica1, now)
+	err = c.checkReplica(context.Background(), "user", "test", replica1, now, now)
 	assert.NoError(t, err)
 
 	// Throw away a sample from replica2.
-	err = c.checkReplica(context.Background(), "user", "test", replica2, now)
+	err = c.checkReplica(context.Background(), "user", "test", replica2, now, now)
 	assert.Error(t, err)
 
 	// Wait more than the overwrite timeout.
 	now = now.Add(1100 * time.Millisecond)
 
 	// Another sample from replica2 to update its timestamp.
-	err = c.checkReplica(context.Background(), "user", "test", replica2, now)
+	err = c.checkReplica(context.Background(), "user", "test", replica2, now, now)
 	assert.Error(t, err)
 
 	// Update KVStore - this should elect replica 2.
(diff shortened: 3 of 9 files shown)


zenador:
:wave: Hi @NickAnge , I'd like to test this change on a dev cell to see if it addresses the problem in https://github.com/grafana/support-escalations/issues/15064

Overview: The customer is using HA tracker for deduplication and the replicas restart and elect a new leader fairly often. When they send writes, the samples are also fairly delayed. So for example, the leader replica A sends a bunch of samples at midnight, and some of those sample also have a timestamp around midnight. If replica A stays healthy and remains the leader, the samples should continue being spaced out by the scrape interval e.g. 15s. However, if right after midnight replica A restarts, we will try to failover to replica B. If the failover timeout is 1 minute, then we failover at 12:01am. If there is no sample delay, then the next samples received would have a timestamp of 12:01am, and the sample interval here would be about a minute, which is okay.

But with the regularly high sample delay of this customer, let's say we have samples that are about a minute old. So when it failsover at 12:01am, we might receive new samples from replica B at 12:00:01am, and then we end up with a sample interval of a second, and we get 2 samples from different replicas very close together in time. And these are counters, so within a replica they are consistently and strictly increasing or staying the same (assuming no resets). But different replicas may have slightly different values (depending on what time they scrape from their sources etc.) at the same point in time, so when this switch happens, the counter might go down a bit, causing a fake counter reset that shouldn't be there.

So to fix this issue, I'm trying to see if we can do the failover based on the earliest sample time instead, and use a per-tenant feature toggle to control that. So in the above scenario, if the earliest sample time in the batch from replica B is 12:00:01am, we would not failover at 12:01am. We would only failover when the earliest sample time is at least 12:01am, so the interval between samples would not be ridiculously close (lower than scrape interval) and cause these fake counter resets. The tradeoff is that it takes longer to failover and we miss out on more samples in the meantime, but this is a toggle so customers facing this problem can choose this to avoid many fake counter resets which may be problematic.

Here we ignore the timestamps of histograms and exemplars for now because this customer doesn't use them, and other customers aren't complaining about this issue (I guess they have smaller sample delays and/or fewer replica restarts.)

I've tagged you because based on history of the HA tracker file you seem to be quite familiar with how this code works, but please let me know if there's someone else I should be asking instead! I'm not very familiar with the HA tracker or the write path in general so I may be misunderstanding how this works or the root cause of the problem the customer is facing. Do you think this approach is feasible, or do you have other suggestions? Are the changes in the code actually achieving what's intended? One weird thing is that when testing, it looks like the pathway where `from_all=false` is never triggered in the customer's cell, which is very strange to me.

pkg/distributor/distributor.go
```diff
@@ -1042,11 +1045,33 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
...
+					} else if tsms > latestSampleTimestamp {
+						latestSampleTimestamp = tsms
+					}
+				}
+			}
+			delta := now.UnixMilli() - earliestSampleTimestamp
+			if delta > 10000 {
+				delta2 := sampleTimestamp - earliestSampleTimestamp
+				delta3 := latestSampleTimestamp - now.UnixMilli()
+				level.Warn(d.log).Log("msg", "earliest sample timestamp is a lot earlier than now", "delta", delta, "delta2", delta2, "delta3", delta3, "user", userID, "cluster", cluster, "replica", replica, "earliestSampleTime", timestamp.Time(earliestSampleTimestamp), "sampleTime", timestamp.Time(sampleTimestamp), "latestSampleTime", timestamp.Time(latestSampleTimestamp), "now", now)
```
dimitarvdimitrov: can delta1, delta2, delta3 have better names?

pkg/distributor/ha_tracker.go
```diff
@@ -453,20 +455,27 @@ func (h *defaultHaTracker) updateKVStoreAll(ctx context.Context, now time.Time)
...
+				} else {
+					sampleTime = now
+				}
 			} else {
 				continue // we don't have any recent timestamps
 			}
 			// Release lock while we talk to KVStore, which could take a while.
 			h.electedLock.RUnlock()
-			err := uh.updateKVStore(ctx, cluster, replica, now, receivedAt)
+			err := uh.updateKVStore(ctx, cluster, replica, now, sampleTime, receivedAt, true)
```
dimitarvdimitrov, line 478: why pass both times to updateKVStore? can't we change between the time of the sample and now

pkg/distributor/ha_tracker_test.go
dimitarvdimitrov, line 1: would be nice to see some tests for the new functionality

pkg/util/validation/limits.go
```diff
@@ -128,6 +128,7 @@ type Limits struct {
 	// between the stored timestamp and the time we received a sample is
 	// more than this duration
 	HATrackerFailoverTimeout                    model.Duration      `yaml:"ha_tracker_failover_timeout" json:"ha_tracker_failover_timeout" category:"advanced" doc:"description=If we don't receive any samples from the accepted replica for a cluster in this amount of time we will failover to the next replica we receive a sample from. This value must be greater than the update timeout."`
+	HATrackerUseSampleTimeForFailover           bool                `yaml:"ha_tracker_use_sample_time_for_failover" json:"ha_tracker_use_sample_time_for_failover" category:"advanced" doc:"description=Use the sample time for failover instead of the current time. This is useful to prevent samples being too close together during failover when write requests are delayed such that the sample time is earlier than the current time."`
```
dimitarvdimitrov: do we need to explain this twice? isn't the flag below enough? it seems to be enough for `drop_labels`

dimitarvdimitrov (commented):
what would happen if a replica sends continuously out of order samples? would we consider that replica offline and fail over to the other replica? 

By extension if a prometheus replica falls behind would we fail over to a more up-to-date replica? This can miss even more scrape intervals than what the failover causes because now at best we will also discard the samples in the time difference between the replicas. For example, replica_1 falls behind by a minute, so it's samples are at T-1m, i.e. 4 scrapes; this causes us to select replica_2 whose samples are at T; now we've discarded all samples between T-1m and T.

But as much as I think about preventing counter resets, then ensuring this sample continuity is the best I can think of.  I think it's ok to merge this for the older release, but this requires us to keep an eye on the behaviour of the prometheis - if their timesync improves, then we have to tun this off for a tenant.

pkg/distributor/distributor.go
zenador: This logging is just temporary, for me to look at. If we keep this approach, I'll probably simplify these logs (get rid of the extra deltas) or remove them altogether, but if/when we start trying to merge to main, I'll make sure to check that the names of the stuff are more descriptive. I've renamed them, but since the only changes so far are to the logging, docs and tests without changing any functionality, I won't build and test a new image (unless we decide we need to change any functionality here)

pkg/distributor/ha_tracker.go
zenador: Not sure if I understand your question, are you asking if we can check for the new config flag, and then pass `now`/`sampleTime` but not both, without changing the signature? If so:

Within `updateKVStore` we will always use the `now` value, as it is used in multiple places, e.g. for `electedAt` time and to check failover. If the new flag is set to true, we will use `sampleTime` to check failover, but we will still keep `now` for `electedAt`, to avoid changing more than is necessary for the fix.

zenador:
> what would happen if a replica sends continuously out of order samples? would we consider that replica offline and fail over to the other replica?

I don't believe so, I think this change only affects when we failover (makes it slower), as the way the update timeout works is untouched. but that's assuming I understand the code correctly, we can confirm this with other reviewers.

> By extension if a prometheus replica falls behind would we fail over to a more up-to-date replica? This can miss even more scrape intervals than what the failover causes because now at best we will also discard the samples in the time difference between the replicas. For example, replica_1 falls behind by a minute, so it's samples are at T-1m, I.e. 4 scrapes; this causes us to select replica_2 whose samples are at T; now we've discarded all samples between T-1m and T.

Not sure I get what you mean here. By "replica_1 falls behind by a minute", do you mean that all replica_1's samples are about a minute old when it is received? I think in this case if replica_1 is the leader, this will not cause a failover, as long as they are still received within the update timeout, as we use `now` for comparison and not the sample time for this. This PR should not be modifying anything to do with the update timeout, please let me know if you spot anything indicating otherwise.

However, if replica_1 is the leader and it does go down for a while, triggering a failover to replica_2, having the new flag set to true can delay the failover if replica_2 has delayed samples, making the time in between with no samples longer than it would have been. I mention this in an earlier comment (but it's very long so I'll just repeat it here): "The tradeoff is that it takes longer to failover and we miss out on more samples in the meantime, but this is a toggle so customers facing this problem can choose this to avoid many fake counter resets which may be problematic." It's definitely something we should make clear in our docs if we end up doing this.

> I think it's ok to merge this for the older release, but this requires us to keep an eye on the behaviour of the prometheis - if their timesync improves, then we have to tun this off for a tenant.

Agreed, it's only needed if they have a fake resets problem, which should go away if they manage to fix their delayed samples consistently. But we should ask them about it first, to see if they think ensuring no fake resets is more important or failing over quickly is more important (and their preference might change depending on how likely the fake resets are).

pkg/distributor/distributor.go
zenador: Separately, we probably won't need this particular log when the reintroduced `cortex_distributor_sample_delay_seconds` lands in the customer cell (think it should be this week).

zenador:
One potential improvement is to have the new feature toggle as a duration instead of a bool. Then we'd leave the current check for failover timeout alone, and if that duration is more than 0, we add a second check to ensure the delta between the earliest sample time and last received at is more than that duration, and require that both checks pass before we failover. So if the new duration is smaller than the failover timeout we can failover faster but still delay it enough to fix the fake resets. Theoretically this duration would need to be at least as much as the scrape interval to account for replicas scraping in an unsynced manner (or maybe twice that)? If we do 2x to be safe, and the scrape interval is 15s so the new duration will be 30s, it would be the same as the default failover timeout of 30s (so the distinction is unneeded in that common case, but could be useful in other setups).

I'd like to test the current version (simpler but more extreme) first though, just to confirm that the approach of delaying failovers works to solve the fake resets problem, before tuning it further (e.g. removing the increased failover timeout and trying the above).

pkg/distributor/ha_tracker.go
```diff
@@ -676,11 +689,17 @@ func (h *defaultHaTrackerForUser) updateKVStore(ctx context.Context, cluster, re
...
 			if desc.Replica != replica {
-				if now.Sub(timestamp.Time(desc.ReceivedAt)) < h.failoverTimeout {
-					level.Info(h.logger).Log("msg", "replica differs, but it's too early to failover", "user", h.userID, "cluster", cluster, "replica", replica, "elected", desc.Replica, "received_at", timestamp.Time(desc.ReceivedAt))
+				var currentTime time.Time
+				if h.useSampleTimeForFailover {
+					currentTime = sampleTime
+				} else {
+					currentTime = now
+				}
+				if currentTime.Sub(timestamp.Time(desc.ReceivedAt)) < h.failoverTimeout {
```
dimitarvdimitrov, line 698: wait so here we compare the timestamp of the samples with the timestamp of the time when we received the request? so if a replica sends some samples from the future, won't this trigger a failover over a replica which is already in the kv store (but its ReceivedAt is outside the update timeout)?

dimitarvdimitrov:
> Not sure I get what you mean here. By "replica_1 falls behind by a minute", do you mean that all replica_1's samples are about a minute old when it is received? I think in this case if replica_1 is the leader, this will not cause a failover, as long as they are still received within the update timeout, as we use `now` for comparison and not the sample time for this. This PR should not be modifying anything to do with the update timeout, please let me know if you spot anything indicating otherwise

https://github.com/grafana/mimir/pull/12082#discussion_r2228070793: it might be the opposite actually - if a replica runs ahead, will that replace an existing replica?

dimitarvdimitrov:
> Then we'd leave the current check for failover timeout alone, and if that duration is more than 0, we add a second check to ensure the delta between the earliest sample time and last received at is more than that duration, and require that both checks pass before we failover

yeah i'm still not sure about comparing the receivedAt with a sample time. They're two different data points and technically nothing ties them together. Shouldn't we store something like a `MaxSampleTime` in the KV store and compare against that instead? 

(This approach still has the problem of getting samples from the future and triggering a false positive failover.)

pkg/distributor/ha_tracker.go
```diff
@@ -453,20 +455,27 @@ func (h *defaultHaTracker) updateKVStoreAll(ctx context.Context, now time.Time)
...
 			if uh.withinUpdateTimeout(now, entry.electedLastSeenTimestamp) {
 				// We have seen the elected replica recently; carry on with that choice.
 				replica = entry.elected.Replica
 				receivedAt = entry.electedLastSeenTimestamp
+				sampleTime = now
 			} else if uh.withinUpdateTimeout(now, entry.nonElectedLastSeenTimestamp) {
 				// Not seen elected but have seen another: attempt to fail over.
 				replica = entry.nonElectedLastSeenReplica
 				receivedAt = entry.nonElectedLastSeenTimestamp
+				if uh.useSampleTimeForFailover {
```
dimitarvdimitrov, line 468: why is the `if` here as well? i mean `updateKVStore` already checks the same `if` and uses `sampleTime` if necessary.

pkg/distributor/distributor.go
```diff
@@ -1042,11 +1045,31 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
...
+				if len(ts.Samples) > 0 {
+					tsms := ts.Samples[0].TimestampMs
+					if tsms < earliestSampleTimestamp {
+						earliestSampleTimestamp = tsms
+					} else if tsms > latestSampleTimestamp {
+						latestSampleTimestamp = tsms
+					}
+				}
+			}
+			if deltaBetweenEarliestAndNow := now.UnixMilli() - earliestSampleTimestamp; deltaBetweenEarliestAndNow > 10000 {
```
NickAnge, line 1062: Could you please clarify why we went with 10000 Milli ? Should this be based on the failoverTimeout ?

pkg/distributor/distributor.go
```diff
@@ -1042,11 +1045,31 @@ func (d *Distributor) prePushHaDedupeMiddleware(next PushFunc) PushFunc {
 		numSamples := 0
 		now := time.Now()
 		group := d.activeGroups.UpdateActiveGroupTimestamp(userID, validation.GroupLabel(d.limits, userID, req.Timeseries), now)
+		sampleTimestamp := timestamp.FromTime(now)
+		if d.limits.HATrackerUseSampleTimeForFailover(userID) {
+			earliestSampleTimestamp := sampleTimestamp
+			latestSampleTimestamp := sampleTimestamp
+			for _, ts := range req.Timeseries {
+				if len(ts.Samples) > 0 {
```
NickAnge, line 1053: We have it behind a feature flag, so it should be fine. But I wonder how many time series we have for each `req.Timeseries`. Iterating over them on every request—I'm not sure how much that's going to impact the CPU.

NickAnge:
Hello @zenador . Sorry for my late feedback. Missed the GH notification.

> So when it failsover at 12:01am, we might receive new samples from replica B at 12:00:01am, and then we end up with a sample interval of a second

I'm not sure I understand this part. Do you mean that when we fail over to B, we're going to receive samples from the past? For example, if replica A sent the last sample at 12:00 AM and the failover to B happens at 12:01 AM, are we still going to see samples from B with earlier timestamps—like 12:00:01? This happens because replica B has scraped Data at the interval 12:00 - 12:01 ? That minute

> So to fix this issue, I'm trying to see if we can do the failover based on the earliest sample time instead, and use a per-tenant feature toggle to control that. So in the above scenario, if the earliest sample time in the batch from replica B is 12:00:01am, we would not failover at 12:01am

Does that mean we are gonna accept samples only after the failover timestamp ? 



Could you please also explain what do we mean by false counter resets ? I think I miss the terminology

NickAnge:
@zenador , another question is what is their scrape interval. Is it 15seconds?  Do you think by increasing their failover time, would have helped ? I think we have that option right now after https://github.com/grafana/mimir/issues/4066

pkg/distributor/ha_tracker.go
zenador: > here we compare the timestamp of the samples with the timestamp of the time when we received the request?

We compare the earliest timestamp of the samples with the ReceivedAt timestamp we last stored for the current replica (not the time we received this request).

> so if a replica sends some samples from the future, won't this trigger a failover over a replica which is already in the kv store (but its ReceivedAt is outside the update timeout)?

Are you asking about the case where a replica's sending time is within the failover timeout (so it shouldn't failover), but the replica's samples are somehow all in the future (since we are comparing the earliest sample timestamp) and hence outside the failover timeout, so it does the failover when it shouldn't? If so, I think the fix mentioned [here](https://github.com/grafana/mimir/pull/12082#issuecomment-3082901124) and implemented in [here](https://github.com/grafana/mimir/pull/12178) should resolve this. Instead of replacing the current failover timing check, we add an additional one, so we do not perform any failovers that we shouldn't according to the current checks, we only delay/prevent some failovers.

> it might be the opposite actually - if a replica runs ahead, will that replace an existing replica?
(comment from below)

Not too sure what you mean here, but does the change mentioned above address your concern?

pkg/distributor/ha_tracker.go
zenador: Yeah this check isn't strictly necessary, I just didn't want to do the conversion if it wasn't going to be used, to minimise unnecessary work for 99% of cases (since for now we only expect one customer to use this flag). It's really minor though, so if you think we should remove the check to keep things simple I'm happy to make that change.

pkg/distributor/distributor.go
zenador: This part is just for logging purposes to observe what's happening, 10000 ms is just arbitrary. I've done enough logging for now so I've removed this in the next [PR](https://github.com/grafana/mimir/pull/12178).

pkg/distributor/distributor.go
zenador: True, for this customer it looks like it's fine based on the last rollout, but it's something we need to be wary of if we enable it for other tenants in future. See [explore](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/explore?schemaVersion=1&panes=%7B%22rfp%22:%7B%22datasource%22:%22ops-cortex%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%20by%20%28user%29%20%28rate%28cortex_ha_tracker_elected_replica_changes_total%7Bcluster%3D%5C%22prod-us-central-5%5C%22,%20namespace%3D%5C%22cortex-dedicated-06%5C%22%7D%5B1m%5D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D,%7B%22refId%22:%22B%22,%22expr%22:%22sum%20by%20%28container,%20version%29%20%28grafana_metrics_enterprise_build_info%7Bcluster%3D%5C%22prod-us-central-5%5C%22,%20namespace%3D%5C%22cortex-dedicated-06%5C%22,%20container%3D~%5C%22distributor%5C%22%7D%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221753251861169%22,%22to%22:%221753254658021%22%7D%7D%7D&orgId=1) for the timing of when the last patch rolled out, and the [writes resources dashboard](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/d/bc9160e50b52e89e0e49c840fea3d379/mimir-writes-resources?orgId=1&refresh=30s&var-datasource=mimir-ops-03&var-cluster=prod-us-central-5&var-namespace=cortex-dedicated-06&from=2025-07-23T06:24:21.000Z&to=2025-07-23T07:10:58.000Z&timezone=UTC) around the same time. It looks fine as we don't have a marked increase in CPU or memory. The pattern does look noticeably different though (the range expands and the min CPU/mem goes down while the max stays the same) and I'm not sure why. Besides the change in the image, I also increased their failover timeout just to be safe (I want to be sure we can eliminate the fake counter resets first before experimenting with what's the minimal timeouts we can use), so not sure if that also affects it, if the pattern changes because failovers happen less often.

zenador:
@dimitarvdimitrov 

> yeah i'm still not sure about comparing the receivedAt with a sample time. They're two different data points and technically nothing ties them together.

The assumption is that the receivedAt should be the same time or later than the latest sample time in the last batch. I think if the current patch fails to fix the fake resets for the customer, this assumption being false could be one possible reason it doesn't work. But if the assumption holds true, I think just using receivedAt is useful to keep things simple (less processing and minimal changes).

> Shouldn't we store something like a MaxSampleTime in the KV store and compare against that instead?

This approach is definitely safer but more involved. I'm not against this but should we first check to see if it's necessary for this customer, i.e. add a log that appears if the latest sample time is after the now, just to avoid making too many changes we might not need? Or do you think we should err on the side of safety because even if it isn't necessary now, it might be in future? I think if the timestamp is too far in the future (like bad timezone conversion) it causes a lot of other problems that we can't solve just in the logic for HA tracker failover anyway, but I suppose it's possible that in cases of bad time sync, it could be off by just a few minutes or less, and if so we should try to cater to that. But I'm not sure how common this is in a real world scenario that we should account for it.

> (This approach still has the problem of getting samples from the future and triggering a false positive failover.)

With the change to an additional timeout check instead of a replaced timeout check, we will not be introducing any new potential for false positives with this change (but we could still get false positives, I just mean it won't make things any worse than status quo).

zenador:
Hi @NickAnge, np and sorry about that, I forgot GH notifications tend to be lost in the noise, will mention on Slack for future PRs.

> I'm not sure I understand this part. Do you mean that when we fail over to B, we're going to receive samples from the past? For example, if replica A sent the last sample at 12:00 AM and the failover to B happens at 12:01 AM, are we still going to see samples from B with earlier timestamps—like 12:00:01? This happens because replica B has scraped Data at the interval 12:00 - 12:01 ? That minute

Yes that's what I mean, because when replica B sends the request at 12:01 it may include samples from 12:00:01. That’s what appears to be happening in the customer cell (I asked this last time in the ingest channel and received this suggestion to look into), and we can confirm that they do have enough delay between the earliest sample time in the batch and the time we get the batch to potentially cause this issue.

> Does that mean we are gonna accept samples only after the failover timestamp ?

I've changed the logic a bit in https://github.com/grafana/mimir/pull/12178 to make it an additional check instead of replacing the current one. So the existing logic is to only allow the failover if enough time has elapsed between the receivedAt of the current leader and the `now` time when receiving samples from the new replica, and we keep that, but in addition to that, when the flag is configured, we also ensure that there's enough time between that receivedAt and the earliest sample time in the incoming batch.

> Could you please also explain what do we mean by false counter resets ? I think I miss the terminology

Yep sorry, it isn't established terminology. I mean that we are detecting a reset (because the counter decreases when it's supposed to be strictly increasing) when there isn't supposed to be one (because the counter didn't actually reset to 0). Does that make sense?

> Do you think by increasing their failover time, would have helped?

I think increasing failover time alone can mitigate the problem but not solve it entirely. The issue is that the delay in samples (diff between earliest sample time and the time we get the batch) can vary a lot. So for example, they originally had the default failover timeout of 30s. (As mentioned above I'm also increasing that to 1m for now just to be safe, but plan to experiment with reducing it later on.) If the sample delay is 29s, we can end up with samples 1s apart and trigger the problem. In that case, having a 1m timeout would mean that the samples would be 31s apart so it would prevent that problem. But when the sample delay is 59s, then a 1m timeout would trigger it again. We could increase that to 2m, but we'd still get the issue when the sample delay is 1m59s (which is much less likely but still can happen), and we'd also be waiting longer unnecessarily when the sample delay is low. See [explore](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/explore?schemaVersion=1&panes=%7B%22rfp%22:%7B%22datasource%22:%22ops-cortex%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.999,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2299.9%20pct%20%7B%7Bjob%7D%7D%22%7D,%7B%22refId%22:%22B%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.99,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2299%20pct%20%7B%7Bjob%7D%7D%22%7D,%7B%22refId%22:%22C%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.5,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2250%20pct%20%7B%7Bjob%7D%7D%22%7D%5D,%22range%22:%7B%22from%22:%221753431139178%22,%22to%22:%221753432909053%22%7D%7D%7D&orgId=1) for the variation in their sample delays. So I think the best option is to have an additional failover timeout based on sample timestamps (that is off by default and can be turned on for this customer) to ensure we delay failover by no more or no less than what we need to. The cost like you mentioned earlier is the additional processing of the time series, but based on the dashboard I linked in the other comment that should be negligible.

pkg/distributor/ha_tracker.go
dimitarvdimitrov: yeah i think we should keep it as simple as possible

dimitarvdimitrov:
> This approach is definitely safer but more involved. I'm not against this but should we first check to see if it's necessary for this customer

part of my concern is to make this code easier to understand. Right now there are assumptions around how the traffic looks like and what's the distribution of samples in time compared to the current time. We could get rid of some of those assumptions if we add the additional field. It's not great if we're adding complexity which solves a very narrow use case.

At least we should document these assumptions

> But I'm not sure how common this is in a real world scenario that we should account for it.

Don't we have some metrics on this? I think the distributor exports `cortex_distributor_latest_seen_sample_timestamp_seconds` and `cortex_distributor_sample_delay_seconds`

pkg/distributor/ha_tracker.go
zenador: Sure, will do that in the [r349 PR](https://github.com/grafana/mimir/pull/12178)

NickAnge:
> Yep sorry, it isn't established terminology. I mean that we are detecting a reset (because the counter decreases when it's supposed to be strictly increasing) when there isn't supposed to be one (because the counter didn't actually reset to 0). Does that make sense?

Yes it makes sense. Thanks for explaining

> I've changed the logic a bit in https://github.com/grafana/mimir/pull/12178 to make it an additional check instead of replacing the current one. So the existing logic is to only allow the failover if enough time has elapsed between the receivedAt of the current leader and the now time when receiving samples from the new replica, and we keep that, but in addition to that, when the flag is configured, we also ensure that there's enough time between that receivedAt and the earliest sample time in the incoming batch.

I think it was discussed in a different thread but I am not sure so I am gonna repeat it. In the case of replica B sending OOO samples that are close to receivedAt constantly, its gonna delay the failover. Do we know for how long can this happen ? 

Imagine below scenario:
- failoverTIme: 1m
- failoverSampleTimeout: 1m
 Lets say that replica A stops sending samples at 12:00 am. For the duration of 1m its gonna go at the first if of [updateKVstore](https://github.com/grafana/mimir/pull/12178/files#diff-d1ce8bd6bcc9fc513fa38df68d5093d7d42227e7ca1dd2fbfb6f8d259db4645e). Afterwards its gonna compare sampleLatest with receivedAt. What if replicaB constantly sends samples that are < than 1m. Its gonna delay the failover and for that amount of time no samples will be ingested. Is my understanding wrong ? 🤔 


>  think increasing failover time alone can mitigate the problem but not solve it entirely. The issue is that the delay in samples (diff between earliest sample time and the time we get the batch) can.....

Am I reading the explore correctly ? Is it the p50 ~2 second?

zenador:
@dimitarvdimitrov 

[`cortex_distributor_sample_delay_seconds` for this customer](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/explore?schemaVersion=1&panes=%7B%22rfp%22:%7B%22datasource%22:%22ops-cortex%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.999,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2299.9%20pct%20%7B%7Bjob%7D%7D%22%7D,%7B%22refId%22:%22B%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.99,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2299%20pct%20%7B%7Bjob%7D%7D%22%7D,%7B%22refId%22:%22C%22,%22expr%22:%22max%20by%20%28job%29%20%28histogram_quantile%280.5,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%2250%20pct%20%7B%7Bjob%7D%7D%22%7D,%7B%22refId%22:%22D%22,%22expr%22:%22min%20by%20%28job%29%20%28histogram_quantile%280.0,%20cortex_distributor_sample_delay_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22min%20%7B%7Bjob%7D%7D%22%7D%5D,%22range%22:%7B%22from%22:%221753431139178%22,%22to%22:%221753432909053%22%7D%7D%7D&orgId=1): Just to see the rough distribution, nothing below 0

Abandoned querying `cortex_distributor_sample_delay_seconds` overall as it takes very long for 5 minutes and seems very expensive.

[`cortex_distributor_latest_seen_sample_timestamp_seconds` for this customer](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/explore?schemaVersion=1&panes=%7B%22rfp%22:%7B%22datasource%22:%22ops-cortex%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22min%28time%28%29%20-%20cortex_distributor_latest_seen_sample_timestamp_seconds%7Bnamespace%3D%5C%22cortex-dedicated-06%5C%22,%20user%3D%5C%22327%5C%22%7D%29%20%3C%200%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1): Thanks, this metric wasn't really on my radar. If I'm using the metric correctly, according to this, based on the past week we don't really need to worry about this customer (the only current target audience for this change) sending future samples.

[`cortex_distributor_latest_seen_sample_timestamp_seconds` overall](https://admin-ops-eu-south-0.grafana-ops.net/stable-grafana/explore?schemaVersion=1&panes=%7B%22rfp%22:%7B%22datasource%22:%22ops-cortex%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22min%20by%20%28namespace%29%20%28time%28%29%20-%20cortex_distributor_latest_seen_sample_timestamp_seconds%29%20%3C%200%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22ops-cortex%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-5m%22,%22to%22:%22now%22%7D%7D%7D&orgId=1): But future samples do happen for other users (sometimes very drastic). Do you think we should remember the latest future timestamp? It looks like in some cases that would completely disable the failover given the huge time gap, so the change might not be worth it.

> At least we should document these assumptions

Sure, if we decide not to do the future timestamp thing for now, I'll add more comments for this.

zenador:
@NickAnge 

> Its gonna delay the failover and for that amount of time no samples will be ingested. 

Yes, if replicaB consistently sends samples that are very delayed. This is the tradeoff we make when we enable this additional timeout. So it will be 0 / off by default, and we only recommend that customers like this one with a fake resets problem and no overly huge sample delay (e.g. we don't want to delay failover and have missing samples for an hour) enable it, if they would prefer to avoid the fake resets at the cost of taking longer to failover, maybe about half a min to 3 mins for this customer (based on the logs I mention below, assuming both timeouts are set to 30s which I want to test out later on if this works).

Also that only applies if they constantly have very delayed samples at the time of the failover. If it goes up and down, then it can trigger the failover faster when there's a batch with low delay, the first time we get a batch where the earliest sample time is at least the failover sample timeout from when replica A last sent samples. So if the out of order samples are very sporadic, say replica A goes down and the first batch from replica B contains a sample from 2 hours ago, it doesn't mean we have to wait 2 hours for the failover, if the next few batches from replica B only has delays of a few minutes, then we will just wait a few minutes.

If my math/logic is correct, the additional delay we're introducing is failoverSampleTimeout + delayBetweenEarlistSampleTimestampInBatchAndNow (roughly related to cortex_distributor_sample_delay_seconds, let's call this X) - failoverTimeout. So if our failoverSampleTimeout is the same or less than failoverTimeout (I should probably recommend setting a value like this in the docs), then the extra delay should be equal to X.

> Am I reading the explore correctly ? Is it the p50 ~2 second?

It's not any one line (ignore the ruler ones though, I should have excluded those, sorry), it's just showing the rough distributions at p50, p99 and p99.9. It's a bit complicated because this metric records observations for all the samples, but what we really care about is the earliest sample timestamp in a batch from a replica. So that distribution shows us that the sample delay for all samples isn't that high, like the median is  ~2s like you said, but the delay for the earliest sample timestamp in a batch could be a lot higher than that.

I don't think we have a metric for the earliest sample timestamp in a batch, but I did some [logging](https://admin-prod-us-east-0.grafana.net/grafana/explore?schemaVersion=1&panes=%7B%22ouv%22%3A%7B%22datasource%22%3A%22loki-ops%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22q0%22%2C%22editorMode%22%3A%22code%22%2C%22expr%22%3A%22%7Bcluster%3D%5C%22prod-us-central-5%5C%22%2Cnamespace%3D%5C%22cortex-dedicated-06%5C%22%2Ccontainer%3D%5C%22distributor%5C%22%7D+%7C%3D+%5C%22earliest+sample+timestamp+is+a+lot+earlier+than+now%5C%22+%7C%3D+%5C%22cluster%3Dvam-ttd-p-01%5C%22%22%2C%22legendFormat%22%3A%22%22%2C%22range%22%3Atrue%2C%22instant%22%3Afalse%2C%22datasource%22%3A%7B%22type%22%3A%22loki%22%2C%22uid%22%3A%22loki-ops%22%7D%2C%22queryType%22%3A%22range%22%2C%22direction%22%3A%22backward%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%221753274044410%22%2C%22to%22%3A%221753274340250%22%7D%2C%22panelsState%22%3A%7B%22logs%22%3A%7B%22columns%22%3A%7B%220%22%3A%22Time%22%2C%221%22%3A%22Line%22%7D%2C%22visualisationType%22%3A%22table%22%2C%22labelFieldName%22%3A%22labels%22%7D%7D%7D%7D&orgId=1) previously for this customer, and just focusing on a particular cluster, we can see that each batch has quite high X (just look at the delta value divided by 1000 for the number of seconds), for example the first few lines show 164s, 210s, 162s. Then checking [logs](https://admin-prod-us-east-0.grafana.net/grafana/explore?schemaVersion=1&panes=%7B%22ouv%22:%7B%22datasource%22:%22loki-ops%22,%22queries%22:%5B%7B%22refId%22:%22q0%22,%22editorMode%22:%22code%22,%22expr%22:%22%7Bcluster%3D%5C%22prod-us-central-5%5C%22,namespace%3D%5C%22cortex-dedicated-06%5C%22,container%3D%5C%22distributor%5C%22%7D%20%7C%3D%20%5C%22earliest%20sample%20timestamp%20is%20a%20lot%20earlier%20than%20now%5C%22%20%7C%3D%20%5C%22cluster%3Dvam-ttd-p-01%5C%22%22,%22legendFormat%22:%22%22,%22range%22:true,%22instant%22:false,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki-ops%22%7D,%22queryType%22:%22range%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%221753266844000%22,%22to%22:%221753267140000%22%7D,%22panelsState%22:%7B%22logs%22:%7B%22columns%22:%7B%220%22:%22Time%22,%221%22:%22Line%22%7D,%22visualisationType%22:%22table%22,%22labelFieldName%22:%22labels%22%7D%7D%7D%7D&orgId=1) for the same thing 2 hours earlier, we see 45s, 28s, 22s.

NickAnge:
@zenador

> If my math/logic is correct, the additional delay we're introducing is failoverSampleTimeout + delayBetweenEarlistSampleTimestampInBatchAndNow (roughly related to cortex_distributor_sample_delay_seconds, let's call this X) - failoverTimeout. So if our failoverSampleTimeout is the same or less than failoverTimeout (I should probably recommend setting a value like this in the docs), then the extra delay should be equal to X.

Can we prove this from this customer that we have pinned the image ? 

Also, FYI, I will be on PTO next week. I have pinged @aldernero to take a look, since we worked together on the HA tracker, so he can provide his input

zenador:
@NickAnge

The patched image we just ran for r348 on their cell doesn't have the latest changes for a customisable failover sample timeout, just an on/off switch, but when we next apply the patch for r349 with the new separate timeout, we can use the logs to check this.

Okay thank you!

zenador:
@dimitarvdimitrov regarding how the counter can go down during failovers

I think it might be due to differences in scrape duration as mentioned in the post linked previously in slack (I think we get the value when the scrape begins but attach the timestamp when the scrape ends), or perhaps you're right and there could also be some minor clock sync issues involved, though I think theoretically it can still happen even if the clocks are perfectly synced, see this simplified scenario using roughly the scrape durations and offsets found in the post:

1. replica A scrapes at 12:00:00.000am and gets 100, takes 20ms, recorded timestamp 12:00:00.020am
2. replica B scrapes at 12:00:00.700am and gets 102, takes 20ms, recorded timestamp 12:00:00.720am
3. replica A scrapes at 12:30:00.000am and gets 210, takes 1s, recorded timestamp 12:30:01.000am
4. replica B scrapes at 12:30:00.700am and gets 212, takes 20ms, recorded timestamp 12:30:00.720am

So let's say replica B is the leader and the last ingested sample was (4). It then goes offline, and when we try to failover to replica A, we receive a batch from A at 12:30:31am, outside the regular failover timeout of 30s, so we accept the failover, but as there is a lot of sample delay, it includes the sample from (3), so we've ingested (4) and (3) together where the counter goes down slightly over a very small time interval, causing a fake reset

dimitarvdimitrov:
> see this simplified scenario

so i was leading with this question, but my point was going to be that in this case the delay can't be that significant. Getting a delay of 15s in a scenario like this sounds contrived to me.

Where I was leaning towards was that this seems like a clock drift problem at the customer end and it's best solved there too.

dimitarvdimitrov:
i still think we should merge something in the main branch so that you don't have to keep making the same diff all over again and so that we are all sure what code we're talking about (is it the one here? is it on the r349 branch or on the r350 branch?)

dimitarvdimitrov:
this has been a very lengthy PR discussion. Before we merge this into main we need a succinct summary for what this change does, why we need it, who is it aimed at and what tradeoffs it comes with. The discussion here is almost enough volume for a design doc, but a good PR description should work too

zenador:
Superceded by https://github.com/grafana/mimir/pull/12331

PR #12083, opened 2025-07-14
Add new panel to writes dashboard for HA tracker elected replica changes

zenador:
#### What this PR does

Monitor HA tracker elected replica changes in the writes dashboard which could be helpful to debug problems with HA tracker.

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir-mixin/dashboards/writes.libsonnet
```diff
@@ -847,5 +847,24 @@ local filename = 'mimir-writes.json';
...
         { fieldConfig+: { defaults+: { unit: 'percentunit', max: '1' } } }
       ),
+    )
+    .addRow(
+      $.row('HA Tracker')
+      .addPanel(
+        local title = 'Elected replica changes';
+        $.timeseriesPanel('Elected replica changes') +
+        $.queryPanel(
+          'sum by (user) (rate(cortex_ha_tracker_elected_replica_changes_total{%s}[$__rate_interval]))'
```
dimitarvdimitrov: let's add a `> 0` so that the panel doesn't contain a lot of series

operations/mimir-mixin/dashboards/writes.libsonnet
```diff
@@ -847,5 +847,24 @@ local filename = 'mimir-writes.json';
...
+          % $.namespaceMatcher(),
+          '{{user}}',
+        ) +
+        { fieldConfig+: { defaults+: { unit: 'changes/s' } } } +
+        $.panelDescription(
+          title,
+          |||
+            The total number of times the elected replica has changed for a user ID/cluster.
```
dimitarvdimitrov:
```suggestion
            The total number of times the elected replica has changed for a user ID.
```

this doesn't group by cluster, does it?

operations/mimir-mixin/dashboards/writes.libsonnet
```diff
@@ -847,5 +847,24 @@ local filename = 'mimir-writes.json';
         ) +
         { fieldConfig+: { defaults+: { unit: 'percentunit', max: '1' } } }
       ),
+    )
+    .addRow(
+      $.row('HA Tracker')
```
dimitarvdimitrov: can we fit this in the row about "Distributor - key-value store for high-availability (HA) deduplication" and maybe rename that row to just HA tracker?

operations/mimir-mixin/dashboards/writes.libsonnet
zenador: Good catch, forgot to update that part, thanks!

operations/mimir-mixin/dashboards/writes.libsonnet
zenador: I've removed the `by(user)` part so it'd only be 1 series, so will omit `> 0` so we can see that there's a result even if it's 0.

zenador:
Good point, thanks! Will use the total ungrouped by user for the main writes dashboard, and add new panels for tenants and top tenants.

zenador:
Not sure why the lint is complaining about `should use $__rate_interval` when other panels in top tenants also use 5m.

dimitarvdimitrov:
> Not sure why the lint is complaining about `should use $__rate_interval` when other panels in top tenants also use 5m.

me neither, i think using rate_interval is safer. i don't see a reason for hardcoding 5m there

do you have a screenshot of the new panels?

PR #12086, opened 2025-07-14
Propagate gcs object mutation rate limit for rule group uploads

fionaliao:
#### What this PR does

When users repeatedly upload the same rule groups too frequently, they can hit GCS's per-object mutation rate limit (1 create/update/delete second per object). Currently, these 429 errors from GCS are being mapped to 500 (Internal Server Error) when returned to users. This is actually a user problem, not a server issue. The logged error is `googleapi: Error 429: The object ... exceeded the rate limit for object mutation operations (create, update, and delete). Please reduce your request rate. See https://cloud.google.com/storage/docs/gcs429.`

This PR detects this specific rate limit error and returning 429. We don't want to return 429s for all object storage rate limit errors, since we can't guarantee all of them are user problems.

A more advanced solution would be to actually implement rate limiting for the ruler API before GCS (or other stores) are called. Just doing this simple check for now as a quick fix for something we occasionally see in practice.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Block-builder-scheduler: Detect gaps in scheduled and completed jobs. #11867
 * [ENHANCEMENT] Distributor: Experimental support for Prometheus Remote-Write 2.0 protocol has been updated. Created timestamps are now supported. This feature includes some limitations. If samples in a write request aren't ordered by time, the created timestamp might be dropped. Additionally, per-series metadata is automatically merged on the metric family level. Ingestion might fail if the client sends ProtoBuf fields out-of-order. The label `version` is added to the metric `cortex_distributor_requests_in_total` with a value of either `1.0` or `2.0`, depending on the detected remote-write protocol. #11977
 * [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. The optimizer can be enabled per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076
+* [ENHANCEMENT]: Ruler: Propagate gcs object mutation rate limit for rule group uploads. #12086
```
```suggestion
* [ENHANCEMENT]: Ruler: Propagate GCS object mutation rate limit for rule group uploads. #12086
```

tacole02 (approved):

pkg/ruler/api.go
```diff
@@ -815,3 +828,11 @@ func alertStateDescToPrometheusAlert(d *AlertStateDesc) *Alert {
 
 	return a
 }
+
+// isGCSObjectMutationRateLimitError checks whether the error message is from GCS and is an object mutation rate limit.
+// This is a per-object limit and is limited to 1 mutation per second (https://cloud.google.com/storage/quotas).
+func isGCSObjectMutationRateLimitError(err error) bool {
+	errStr := err.Error()
+	return strings.Contains(errStr, "googleapi: Error 429") &&
+		strings.Contains(errStr, "exceeded the rate limit for object mutation operations")
```
alexweav: Nit: For extra assuredness, would it be worth trying to cast the error into a GCS-specific type? e.g. it seems to be either [this](https://github.com/googleapis/gax-go/blob/cb080069bb4bc31f6598655e94555d9f495f8c1e/v2/apierror/apierror.go#L219) or [this](https://github.com/googleapis/gax-go/blob/cb080069bb4bc31f6598655e94555d9f495f8c1e/v2/apierror/apierror.go#L219), both of which have a way to get more strongly typed details

Completely optional!

fionaliao: Updated in https://github.com/grafana/mimir/pull/12086/commits/38acd60d5a5627521c1b57d3d67e67864a83365d to cast to `googleapi.Error` as per https://cloud.google.com/go/docs/reference/cloud.google.com/go/storage/latest#hdr-Errors. I don't think the gax error is used in the case, but if so, [it would wrap the `googleapi.Error`](https://github.com/grafana/mimir/blob/392588b98e386c0fd7cdd549a54e5f06d3155bc2/vendor/github.com/googleapis/gax-go/v2/apierror/apierror.go#L258-L261), so that would still work with `errors.As(err, &gErr)`

alexweav (approved):

PR #12087, opened 2025-07-14
Alertmanager to store fingerprint of configuration instead of structure

yuri-tceretian:
#### What this PR does
This PR improves multitenant Alertmanager to store a fingerprint of user configuration instead of a full structure. The structure uses this field to compare current configuration with one loaded from the storage and skip applying the configuration if it's not changed.
I decided to implement this function on my own and add necessary tests to assert the correctness of the fingerprint and protect for unintended modifications in future. The method uses approach that has been well tested in [Grafana ](https://github.com/grafana/grafana/blob/7688089a5763cb5417d6ff61bc4d23dbb6d896d5/pkg/services/ngalert/schedule/registry.go#L218). 

This PR is built on top of the refactoring https://github.com/grafana/mimir/pull/12052


pkg/alertmanager/multitenant.go
```diff
@@ -332,9 +336,9 @@ type MultitenantAlertmanager struct {
 
 	alertmanagersMtx sync.Mutex
 	alertmanagers    map[string]*Alertmanager
-	// Stores the current set of configurations we're running in each tenant's Alertmanager.
+	// Stores the current set of configurations hashes we're running in each tenant's Alertmanager.
```
santihernandezc: Super minor nit.
```suggestion
	// Stores the current set of configuration hashes we're running in each tenant's Alertmanager.
```

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+		}
+	}
+
+	writeString(f.User)
+	writeString(f.RawConfig)
+	writeBool(f.UsingGrafanaConfig)
+	if f.TmplExternalURL != nil {
+		writeString(f.TmplExternalURL.String())
+	} else {
+		writeBytes([]byte{})
```
santihernandezc: On line 922, we're using `writeString(nil)`. After some testing, I see it's the hash changes in the same way whether we pass an empty slice or `nil`. However, I wasn't sure if this was intentional or not at first.

We could also pass `nil` here to be consistent and avoid confusion.

```suggestion
		writeBytes(nil)
```

Another option is to pass `writeBytes([]byte{})` on line 922, as I see we're also using an empty slice on line 959.

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+		UsingGrafanaConfig: false,
+	}
+}
+
+func (f amConfig) fingerprint() model.Fingerprint {
+	sum := fnv.New64a()
+
+	writeBytes := func(b []byte) {
+		_, _ = sum.Write(b)
+		_, _ = sum.Write([]byte{255}) // add separator between fields. It is impossible in strings so, it reduces collisions in strings
```
santihernandezc, line 918:
```suggestion
		_, _ = sum.Write([]byte{255}) // Add separator between fields. It is impossible in strings, so it reduces collisions.
```

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+	}
+	writeString := func(s string) {
+		if len(s) == 0 {
+			writeBytes(nil)
+			return
+		}
+		// #nosec G103
+		// avoid allocation when converting string to byte slice
```
santihernandezc, line 926:
```suggestion
		// Avoid allocation when converting string to byte slice.
```

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+	// so we do not need to care about the random order of the pairs in the map.
+	var mapFp uint64
+	for k, v := range f.EmailConfig.StaticHeaders {
+		sum.Reset()
+		writeString(k)
+		writeString(v)
+		mapFp ^= sum.Sum64()
+	}
+
+	// Ignore order in templates because they're usually built from the map
```
santihernandezc, line 970:
```suggestion
	// Ignore order in templates because they're usually built from the map.
```

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+	var templatesFp uint64
+	for _, template := range f.Templates {
+		sum.Reset()
+		writeString(template.Name)
+		writeString(template.Content)
+		writeString(string(template.Kind))
+		templatesFp ^= sum.Sum64()
+	}
+
+	// Ignore order in content types because it does not matter
```
santihernandezc, line 980:
```suggestion
	// Ignore order in content types because it does not matter.
```

pkg/alertmanager/multitenant.go
```diff
@@ -901,6 +900,105 @@ type amConfig struct {
...
+	var contentTypesFp uint64
+	for _, ct := range f.EmailConfig.ContentTypes {
+		sum.Reset()
+		writeString(ct)
+		contentTypesFp ^= sum.Sum64()
+	}
+
+	// combine all hashes, including ones for empty slices\map
```
santihernandezc, line 988:
```suggestion
	// Combine all hashes, including ones for empty slices/map.
```

pkg/alertmanager/multitenant_test.go
```diff
@@ -3607,207 +3626,160 @@ func TestComputeConfig(t *testing.T) {
...
-						Content: simpleTemplateOne,
-					},
-				},
+		// Calculate total fields across all structs
+		totalFields := 0
+		totalFields += getFieldCount(definition.PostableApiTemplate{})
+		totalFields += getFieldCount(alertingReceivers.EmailSenderConfig{})
+		totalFields += getFieldCount(amConfig{})
+
+		require.Equalf(t, expectedTotalFields, totalFields, "Total fields across structs is %d, expected %d; new fields may require updating fingerprint method", totalFields, expectedTotalFields)
```
santihernandezc, line 3647: Nice! :100:

santihernandezc (approved):
Nice work! :+1: 

I left some comments. These are non-blocking, you can choose whether to accept the suggestions or not.

PR #12089, opened 2025-07-14
Alertmanager: Support kind in Templates coming from Grafana

yuri-tceretian:
#### What this PR does
Updates GrafanaAlertmanagerConfig with a new field `templates` of type `[]definition.PostableApiTemplate`. This field will eventually replace map `template_files` but for the sake of backward compatibility it is still supported. Both fields are merged into single collection, which is then propagated downstream. 

This will let us control which fields are populated on Grafana side, and transition to the new field gracefully. 

This PR is built on top of the refactoring https://github.com/grafana/mimir/pull/12052

No Mimir Alertmanager related  APIs or models are affected.

### DIFF
--- a/pkg/alertmanager/config_test.go
+++ b/pkg/alertmanager/config_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/grafana/alerting/definition"
 	"github.com/grafana/alerting/receivers"
+	"github.com/prometheus/alertmanager/config"
 	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/mimir/pkg/alertmanager/alertspb"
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/config.go
```diff
@@ -75,7 +75,7 @@ func createUsableGrafanaConfig(logger log.Logger, gCfg alertspb.GrafanaAlertConf
 	}
 	amCfg.AlertmanagerConfig.Receivers = rcvs
 
-	rawCfg, err := json.Marshal(amCfg.AlertmanagerConfig)
+	rawCfg, err := definition.MarshalJSONWithSecrets(amCfg.AlertmanagerConfig)
```
JacobsonMT, line 78: All tests currently pass with either `json.Marshal` or `definition.MarshalJSONWithSecrets` , let's add a Secret to one of the tests in `config_test`.

JacobsonMT (approved):
LGTM

PR #12090, opened 2025-07-14
ingester: Stagger head compaction intervals when running ingesters in multiple zones

### LINKED ISSUE #12081: Zone-Aware Staggered Head Compactions
Currently, Mimir ingesters perform head compactions at configured intervals regardless of which zone they are in. Because they all use the same interval dictated by `head_compaction_interval`, there could be cases where head compactions that cross multiple zones are executed at the same time.

This poses a problem as in Sigyn where strong consistency queries end up timing out because there's a delay in the consumption of the last produced offset while the head compaction is on-going. 

**Proposed Solution**

Implement zone-aware staggered compactions that distribute head compaction operations across time based on availability zones.

Example with 3 zones and 15-minute intervals:
•  Zone A: Compacts at 0:05, 0:20, 0:35, 0:50...
•  Zone B: Compacts at 0:10, 0:25, 0:40, 0:55...
•  Zone C: Compacts at 0:15, 0:30, 0:45, 1:00...

There's one caveat here: if the compaction takes longer than the interval, we might end up with non-staggered compactions again but my understand of the go tick is that it drops ticks and this shouldn't happen - I need to double check that.

gotjosh:
#### What this PR does

Attempts to calculate staggered intervals for ingesters in each zone to avoid ingesters across all zones to do head compaction at the same time, which in turn would make strong consistency queries fail. 

Fixes #12081

TODO:

 - [x] integration test
 - [x] fix unit test
 - [x] determine whenever we should take into account if compactions go over the interval duration and we have a risk of aligning compactions again

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -3243,9 +3244,9 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
 	// effectively spreading the compactions over the configured interval.
 	firstInterval, standardInterval := i.compactionServiceInterval(time.Now(), i.lifecycler.Zones())
 
-	// After the first interval, we want the compaction to run at a specified interval for the partial zone if we have multiple zones,
+	// After the first interval, we want the compaction to run at a specified interval for the zone if we have multiple zones,
 	// before we switch to running the compaction at the standard configured `HeadCompactionInterval`.
-	// If the criteria to have staggered compactions is not met, standardInterval and i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval are the same.
+	// If the criteria to have staggered compactions are not met, standardInterval and i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval are the same.
 	stopTicker, tickerChan := util.NewVariableTicker(firstInterval, standardInterval)
 	defer func() {
 		// We call stopTicker() from an anonymous function because the stopTicker()
@@ -3326,12 +3327,11 @@ func (i *Ingester) compactionServiceInterval(now time.Time, zones []string) (fir
 	// With jittering enabled, we want to apply a positive jitter to the staggered interval.
 	// We estimate that roughly 50% of the interval window is a good enough heuristic than when the 50% jitter is applied,
 	// we get enough variability to not get overlap between zones and ingesters.
-	var offset time.Duration
+	var jitter time.Duration
 	if len(zones) > 0 {
-		offset = (interval / time.Duration(len(zones))) / 2
+		jitter = time.Duration(rand.Int63n((interval.Nanoseconds() / int64(len(zones))) / 2))
 	}
 
-	jitter := util.DurationWithNegativeJitter(offset, 0.50)
 	return zoneAwareInterval + jitter, interval
 }
 


colega:
Can we do this without involving ring-consensus by just setting a different flag value on each zone's container?

pkg/ingester/ingester.go
```diff
@@ -3289,7 +3293,7 @@ func (i *Ingester) compactionServiceInterval() (firstInterval, standardInterval
 		// if ingest storage is enabled.
 		standardInterval = min(i.cfg.BlocksStorageConfig.TSDB.HeadCompactionIntervalWhileStarting, i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval)
 	} else {
-		standardInterval = i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval
+		standardInterval = i.calculateHeadCompactionInterval()
 	}
 
 	if i.cfg.BlocksStorageConfig.TSDB.HeadCompactionIntervalJitterEnabled {
```
pracucci: I think we have to rework the jitter logic. We want the jitter being applied after the per-zone time slicing. Let's say we have two zones and 20m interval. What we want is:
- zone-a: from time 0m to time 10m, with a jitter within that interval
- zone-b: from time 10m to time 20m, with a jitter within that interval

Currently we apply a 100% negative jitter, but that can also cause the overlapping between zones. I think we want a smaller jitter. The way I see it is:
- zone-a: compaction should run from time 0m to time 10m, but only start from time 0m to time 5m (that is a 50% positive jitter)
- zone-b: compaction should run from time 10m to time 20m, but only start from time 10m to time 15m (that is a 50% positive jitter)

pkg/ingester/ingester.go
```diff
@@ -3301,6 +3305,42 @@ func (i *Ingester) compactionServiceInterval() (firstInterval, standardInterval
 	return
 }
 
+// calculateHeadCompactionInterval calculates the head compaction interval based on the number of zones
+// and the configured head compaction interval to ensure we can stagger the compaction across zones.
+// If zone awareness is not enabled, it returns the configured interval as-is.
+func (i *Ingester) calculateHeadCompactionInterval() time.Duration {
```
pracucci: I can't understand how this work without using `time.Now()` too. We want to synchronize the compaction between different ingesters. We need to take `time.Now()` in account too, to know how long we have to delay the 1st compaction.

pkg/ingester/ingester.go
```diff
@@ -3242,7 +3242,11 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
 	// helps to have different ingesters running the compaction at a different time,
 	// effectively spreading the compactions over the configured interval.
 	firstInterval, standardInterval := i.compactionServiceInterval()
-	stopTicker, tickerChan := util.NewVariableTicker(firstInterval, standardInterval)
+
+	// After the first interval, we want the compaction to run at a specified interval for the partial zone if we have multiple zones,
+	// before we switch to running the compaction at the standard configured `HeadCompactionInterval`.
+	// If the criteria to have staggered compactions is not met, standardInterval and i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval are the same.
+	stopTicker, tickerChan := util.NewVariableTicker(firstInterval, standardInterval, i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval)
```
pracucci: To my understanding `standardInterval` is now used to delay the recurring compactions (that is every `i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval`). If so, I don't think this is the right approach, because when we apply the "delay" we don't also want to trigger a compaction once that delay has passed.

pracucci:
> Can we do this without involving ring-consensus by just setting a different flag value on each zone's container?

@colega We need to know how many total zones are and what are the actual values to make it deterministic. I think reading it from the ring is just easier. Don't see any real problem doing it.

colega:
> We need to know how many total zones are and what are the actual values to make it deterministic. 
We know that in jsonnet, right?

> 
Can we do this without involving ring-consensus by just setting a different flag value on each zone's container?

@colega We need to know how many total zones are and what are the actual values to make it deterministic. I think reading it from the ring is just easier. Don't see any real problem doing it.

I don't think there's any problem in reading it from the ring, but IMO it's easier to reason when it's set statically on each zone.

CHANGELOG.md
```diff
@@ -152,6 +152,7 @@
 * [ENHANCEMENT] Store-gateway: Change level 0 blocks to be reported as 'unknown/old_block' in metrics instead of '0' to improve clarity. Level 0 indicates blocks with metadata from before compaction level tracking was added to the bucket index. #11891
 * [ENHANCEMENT] Compactor, distributor, ruler, scheduler and store-gateway: Makes `-<component-ring-config>.auto-forget-unhealthy-periods` configurable for each component. Deprecates the `-store-gateway.sharding-ring.auto-forget-enabled` flag. #11923
 * [ENHANCEMENT] otlp: Stick to OTLP vocabulary on invalid label value length error. #11889
+* [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail #12090
```
```suggestion
* [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail. #12090
```

tacole02 (approved):

CHANGELOG.md
pracucci:
```suggestion
* [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail with experimental ingest storage is enabled. #12090
```

pkg/ingester/ingester.go
```diff
@@ -4320,3 +4397,38 @@ func createManagerThenStartAndAwaitHealthy(ctx context.Context, srvs ...services
...
+//
+// This creates a predictable, clock-aligned schedule where each zone compacts at fixed times
+// (e.g., zone 'a' at :00, :15, :30, :45 and zone 'b' at :07, :22, :37, :52 for a 15-minute interval).
+//
+// Returns the interval until the next scheduled compaction for the zone as a time.Duration.
+func nextCompactionInterval(now time.Time, compactionInterval, zoneOffset time.Duration) time.Duration {
+	// Calculate how much time has elapsed since the start of the current hour.
+	elapsed := time.Duration(now.Minute())*time.Minute +
+		time.Duration(now.Second())*time.Second +
+		time.Duration(now.Nanosecond())*time.Nanosecond
```
pracucci: [nit] Can be simplified:

```suggestion
	elapsed := now.Sub(now.Truncate(time.Hour))
```

pkg/ingester/ingester.go
```diff
@@ -4320,3 +4397,38 @@ func createManagerThenStartAndAwaitHealthy(ctx context.Context, srvs ...services
...
+
+	// Find the start of the current compaction interval by rounding down elapsed time
+	// to the nearest interval boundary.
+	currentIntervalStart := (elapsed / compactionInterval) * compactionInterval
```
pracucci: Note to reviewers: we have a validation on `HeadCompactionInterval` that is guaranteed to be up to 15m. This logic would be wrong if `HeadCompactionInterval` is > 1h, but it's not possible.

Note to Josh: I would clarify it in the comment that `HeadCompactionInterval` is guaranteed to be less than 1h by config validation.

pkg/ingester/ingester.go
```diff
@@ -4320,3 +4397,38 @@ func createManagerThenStartAndAwaitHealthy(ctx context.Context, srvs ...services
...
 }
+
+// nextCompactionInterval calculates the precise time until the next compaction for a specific zone
+// based on the current time, the configured compaction interval, and the computed zone's offset.
+//
+// This creates a predictable, clock-aligned schedule where each zone compacts at fixed times
+// (e.g., zone 'a' at :00, :15, :30, :45 and zone 'b' at :07, :22, :37, :52 for a 15-minute interval).
+//
+// Returns the interval until the next scheduled compaction for the zone as a time.Duration.
+func nextCompactionInterval(now time.Time, compactionInterval, zoneOffset time.Duration) time.Duration {
```
pracucci: I suggest to rename this function. An "interval" is a time at this a thing happens **periodically**. This is not an interval. This is the time before the next compaction.

pkg/ingester/ingester.go
```diff
@@ -4320,3 +4397,38 @@ func createManagerThenStartAndAwaitHealthy(ctx context.Context, srvs ...services
...
+	if currentCompactionTime > elapsed {
+		return currentCompactionTime - elapsed
+	}
+
+	// The compaction time has already passed in the current interval,
+	// so schedule it for the next interval at the same offset.
+	nextCompactionTime := currentIntervalStart + compactionInterval + zoneOffset
+
+	// Return the interval duration until the next scheduled compaction time.
+	return nextCompactionTime - elapsed
```
pracucci: This entire function looks excessively verbose to me. Can't we simply reverse the problem and look how long ago the **last** compaction happened? Then the `<interval> - <time since last> = <how to long to wait before next one>`:

```suggestion
      // Calculate how long elapsed since the last time compaction should have run for this zone.
      timeSinceLastCompaction := (elapsed - zoneOffset) % compactionInterval
      if timeSinceLastCompaction < 0 {
          timeSinceLastCompaction += compactionInterval
      }

      return compactionInterval - timeSinceLastCompaction
```

pkg/ingester/ingester.go
```diff
@@ -3265,12 +3269,12 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
 			// Check if the desired interval has changed. We only compare the standard interval
 			// before the first interval may be random due to jittering.
-			if newFirstInterval, newStandardInterval := i.compactionServiceInterval(); standardInterval != newStandardInterval {
+			if newFirstInterval, newStandardInterval := i.compactionServiceInterval(time.Now(), i.lifecycler.Zones()); standardInterval != newStandardInterval {
 				// Stop the previous ticker before creating a new one.
 				stopTicker()
 
 				standardInterval = newStandardInterval
-				stopTicker, tickerChan = util.NewVariableTicker(newFirstInterval, newStandardInterval)
+				stopTicker, tickerChan = util.NewVariableTicker(newFirstInterval, newStandardInterval, i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval)
```
pracucci: I don't understand this. Why we pass `i.cfg.BlocksStorageConfig.TSDB.HeadCompactionInterval` as 3rd param?

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
+
+	// Find the index of the current zone in the zones list to determine its offset
+	current := i.cfg.IngesterRing.InstanceZone
+	zoneIndex := -1
+	for idx, zone := range zones {
+		if zone == current {
+			zoneIndex = idx
+			break
+		}
+	}
```
pracucci: Can be simplified:

```suggestion
	zoneIndex := slices.Index(zones, i.cfg.IngesterRing.InstanceZone)
```

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
+
+	// Zone not found in the list - this shouldn't happen but handle gracefully
+	if zoneIndex == -1 {
+		level.Warn(i.logger).Log("msg", "current zone not found in zones list, using default interval", "current_zone", current, "available_zones", strings.Join(zones, ","))
```
pracucci: Clarify this message. It's not clear we're talking about the TSDB head compaction interval.

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
+	if len(zones) > 0 {
+		offset = (interval / time.Duration(len(zones))) / 2
+	}
+
+	offsetWithJitter := util.DurationWithNegativeJitter(offset, 0.50)
+
+	return staggeredInterval + offsetWithJitter, interval
+}
+
+// calculateStaggeredCompactionInterval calculates when the next compaction should occur,
```
pracucci: Please clarify that `zones` **must** be sorted. Alternatively, we could just sort here again to keep it safe, given this function is not called frequently.

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
+		level.Warn(i.logger).Log("msg", "current zone not found in zones list, using default interval", "current_zone", current, "available_zones", strings.Join(zones, ","))
+		return interval
+	}
+
+	// Calculate this zone's offset and the next compaction interval.
+	offsetStep := interval / time.Duration(len(zones))
+	zoneOffset := time.Duration(zoneIndex) * offsetStep
+	result := nextCompactionInterval(now, interval, zoneOffset)
+
+	level.Debug(i.logger).Log("msg", "calculated zone-aware compaction interval",
```
pracucci: [nit]

```suggestion
	level.Debug(i.logger).Log("msg", "calculated zone-aware TSDB Head compaction interval",
```

pracucci: But it's not really an interval. We should change this terminology.

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
+}
+
+// calculateStaggeredCompactionInterval calculates when the next compaction should occur,
+// staggering compactions across zones to distribute load over time.
+//
+// With zone awareness enabled, each zone gets a unique offset within the compaction interval.
+// Example with 15min interval and 2 zones:
+// - Zone 'a' compacts at 0:00, 0:15, 0:30, 0:45
+// - Zone 'b' compacts at 0:07, 0:22, 0:37, 0:52
+func (i *Ingester) calculateStaggeredCompactionInterval(now time.Time, zones []string) time.Duration {
```
pracucci: Same comment here about the "interval". It's not an interval. It's the time before the next compaction, and then the standard interval will be applied.

Also, in a log below you used "zone-aware compaction" terminology. We could change "staggered" in "zone-aware": I think it makes it more clear what we're doing.

pkg/ingester/ingester.go
```diff
@@ -3294,22 +3298,95 @@ func (i *Ingester) compactionServiceRunning(ctx context.Context) error {
...
-	return
+	// With jittering enabled, we want to apply a positive jitter to the staggered interval.
+	// We estimate that roughly 50% of the interval window is a good enough heuristic than when the 50% jitter is applied,
+	// we get enough variability to not get overlap between zones and ingesters.
+	var offset time.Duration
+	if len(zones) > 0 {
+		offset = (interval / time.Duration(len(zones))) / 2
+	}
+
+	offsetWithJitter := util.DurationWithNegativeJitter(offset, 0.50)
```
pracucci: The jigger should be positive only here. The calculated `staggeredInterval` is how long to wait before the start of compactions in this zone. From that time, we want to eventually delay it, but not start in the previous zone interval.

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/0494729fdb23ac331d772dde8ef49bb7e5b42531

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/0494729fdb23ac331d772dde8ef49bb7e5b42531

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/0494729fdb23ac331d772dde8ef49bb7e5b42531

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: It was a commit mistake - I had it remove in the local changes I had with the next set of changes.

I fixed it as part of https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: I could have sworn I noted this down - I also have a test that ensures the validation behaviour.

Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

CHANGELOG.md
gotjosh: Done in https://github.com/grafana/mimir/pull/12090/commits/a727241d191f6308fd0fb4f01d111138b80f3c39

pkg/ingester/ingester.go
gotjosh: I kept zone-aware based on your suggestion for `timeToNextZoneAwareCompaction` - do you want me to use the terminology staggered just in the logs or should I also change the function name? 

For example, `timeToNextStaggeredCompaction`, `computed timing for the next TSDB head staggered compaction for this zone` and `could not compute when the next staggered TSDB head compaction should occur, current zone not found in zones list, using the configured interval`

pkg/ingester/ingester.go
gotjosh: We discussed this offline - marco was right, with the implementation above we were only the baseline jitter was half of half of the half of the offset.

Done in https://github.com/grafana/mimir/pull/12090/commits/74df14a9aff53ce4a1881608200add4ff2f32c77

pracucci (approved):
LGTM, thanks!

PR #12091, opened 2025-07-14
parquet store-gateways: open labels file from bucket with flag

francoposa:
#### What this PR does
```
-blocks-storage.bucket-store.parquet-load-index-to-disk
    	True to download the Parquet labels file to disk before opening it. False to open it directly from the bucket. (default true)
```

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


npazosmendez:
Looks like the failing test is already failing in `parquet-main`. I'm gonna merge this and fix the test separately

PR #12094, opened 2025-07-15
querier: add native histogram definition to cortex_bucket_index_load_duration_seconds

juliusmh:
#### What this PR does

- Adds native histogram definition to `cortex_bucket_index_load_duration_seconds`, uses standard parameters:
  - BucketFactor: 1.1
  - MaxBuckets: 100
  - MinResetDuration: 1h
- adds new `promtest` package with helpers to check if
  -  a metric is native histogram
  - a histogram has correct amount of samples

Part of: https://github.com/grafana/mimir-squad/issues/3077

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/util/promtest/histogram.go
+++ b/pkg/util/promtest/histogram.go
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package promtest
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+// HasNativeHistogram checks if a gathered metric has a native histogram.
+//
+// A histogram uses native format if it uses negative/positive span or zero_count > 0.
+// If zero_count is 0, positive span always contains a no-op span.
+func HasNativeHistogram(g prometheus.Gatherer, metricName string) error {
+	family, err := gatherMetricFamily(g, metricName)
+	if err != nil {
+		return err
+	}
+	histograms, err := getHistograms(family)
+	if err != nil {
+		return err
+	}
+	if len(histograms) == 0 {
+		return fmt.Errorf("metric family %s has no histograms", metricName)
+	}
+	for i, h := range histograms {
+		switch {
+		case h.ZeroCount != nil && *h.ZeroCount > 0:
+		case h.ZeroCountFloat != nil && *h.ZeroCountFloat > 0:
+		case len(h.PositiveSpan) > 0:
+		case len(h.NegativeSpan) > 0:
+		default:
+			return fmt.Errorf("metric %d is not a native histogram", i)
+		}
+	}
+	return nil
+}
+
+// HasSampleCount checks if a gathered metric has a specific sample count.
+func HasSampleCount(g prometheus.Gatherer, metricName string, wantCount float64) error {
+	family, err := gatherMetricFamily(g, metricName)
+	if err != nil {
+		return err
+	}
+	histograms, err := getHistograms(family)
+	if err != nil {
+		return err
+	}
+	for i, h := range histograms {
+		var sampleCount float64
+		switch {
+		case h.SampleCount != nil:
+			sampleCount = float64(*h.SampleCount)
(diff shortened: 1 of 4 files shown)


krajorama (approved):
LGTM, note to other reviewers: no tests due to https://github.com/prometheus/client_golang/issues/1824

pkg/util/promtest/histogram.go
```diff
@@ -0,0 +1,85 @@
...
+	family, err := gatherMetricFamily(g, metricName)
+	if err != nil {
+		return err
+	}
+	histograms, err := getHistograms(family)
+	if err != nil {
+		return err
+	}
+	for i, h := range histograms {
+		if len(h.PositiveSpan)+len(h.NegativeSpan) == 0 {
```
krajorama: I think  https://github.com/prometheus/client_golang/blob/b157309b723f0b8588ea604bb78dbbba196803f2/prometheus/histogram.go#L846 means you need to also check for `&& h.ZeroThreshold==0 && h.ZeroCount==0` and the float version as well if it has one.

Because that referenced code allows for the zero bucket to be used, but no spans. https://prometheus.io/docs/specs/native_histograms/#zero-bucket

pkg/util/promtest/histogram.go
```diff
@@ -0,0 +1,85 @@
...
+			return fmt.Errorf("metric %d has no negative or positive spans", i)
```
krajorama: And here just say that the metric is not a native histogram.

pkg/util/promtest/histogram.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+	for i, h := range histograms {
+		if h.SampleCountFloat != nil && *h.SampleCountFloat != count {
+			return fmt.Errorf("histogram %d should have %f samples but got %f", i, count, *h.SampleCountFloat)
+		}
+		if h.SampleCount != nil && *h.SampleCount != uint64(count) {
+			return fmt.Errorf("histogram %d should have %f samples but got %d", i, count, *h.SampleCount)
+		}
+	}
```
krajorama: It shouldn't happen but technically this code returns no error if there was no count at all, so let's do something super explicit like
```
switch {
  case h.SampleCountFloat != nil:
     if *h.SampleCountFloat != count {
       // error
     }
  case h.SampleCount != nil:
    if *h.SampleCount != uint64(count) {
      // error
    } 
  default:
     // error
}
```

pkg/util/promtest/histogram.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+// IsNativeHistogram checks if a gathered metric has positive or negative spans.
+//
+// This is possible because native histograms without any observations and zero_threshold of 0
+// use a no-op span (offset 0, length 0).
+func IsNativeHistogram(g prometheus.Gatherer, metricName string) error {
```
krajorama: nit: technically the DTO can have classic buckets as well, so maybe call this HasNativeHistogram ? No strong feelings on this.

krajorama (commented):
Thank you. I'm not sure this is the most user friendly API, compared to the existing text based comparison that can do multiple metrics at the same time.  I've made some comments about the implementation.  Let's iterate on the API a little.

tacole02 (approved):

juliusmh:
> Thank you. I'm not sure this is the most user friendly API, compared to the existing text based comparison that can do multiple metrics at the same time. I've made some comments about the implementation. Let's iterate on the API a little.

I agree that the API isn't ideal, but it allows us to skip the exposition format in `GatherAndCompare` which is painful to work with for native histograms. 

To make this more compact, we could add some "syntactic sugar" along those lines:

```go
promtest.GatherAndTest(
  t, reg, 
  promtest.Histogram("a").IsNative().HasSampleCount(1),
  promtest.Counter("b").HasValue(100),
)
```

juliusmh:
Holding this back, as it will introduce a lot of new series and we're not sure if worth it.

PR #12101, opened 2025-07-15
Refactoring: remove querymiddleware.Codec interface

pracucci:
#### What this PR does

`querymiddleware.Codec` is an interface, but there's no need for an interface nowadays. We introduced the `Codec` interface years ago in Cortex because, at that time, Loki was vendoring query-frontend from Cortex and they needed a different implementation. Nowadays the only codec is `prometheusCodec` (I also checked GEM).

In this PR I propose to remove the `Codec` interface and directly expose the codec implemention instead. This removes an unnecessary indirection and makes navigating the code easier.

If this PR is accepted, I will take care of replacing the `Codec` usage in GEM.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/frontend/querymiddleware/codec.go
```diff
@@ -238,7 +213,8 @@ func newPrometheusCodecMetrics(registerer prometheus.Registerer) *prometheusCode
 	}
 }
 
-type prometheusCodec struct {
+// PrometheusCodec is used to encode/decode query requests and responses so they can be passed down to middlewares.
+type PrometheusCodec struct {
```
tcard: How about we just name this `Codec`? The `Prometheus` prefix makes me think this is an implementation of some interface, which it isn't anymore.

tcard (approved):

pkg/frontend/querymiddleware/codec.go
pracucci: Good point. I agree `Codec` is just fine. I will do it.

pracucci: Done in https://github.com/grafana/mimir/pull/12101/commits/c694fe0de4a1ecc44aa9ac82d8648393caa0e1ca

tcard (approved):

PR #12102, opened 2025-07-15
Standardise non-regex patterns in query blocking upon loading of config

zenador:
#### What this PR does

Make query blocking easier to use by not needing users to run `mimirtool promql format` themselves, instead applying a similar transformation to the patterns upon loading.

This is skipped for regex patterns as they may not compile into proper PromQL expressions.

Should be fairly efficient as it is only parsed upon loading and not every time it is compared.

Currently only applies to query blocking, not sure if we also want this for query limiting or others.

To keep things working similarly to before, we don't return any error if the PromQL query is invalid, and we just keep the string for comparison, but it will never match anything. If we would rather change this, we could move it to the validation step and return an error if a pattern is invalid. Or we could silently remove the useless string from being checked without returning an error.

Tested to work locally.

N/A

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.


docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,10 @@ The order of patterns is preserved, so the first matching pattern will be used.
 
 ## Format queries to block
 
-Use Mimirtool's `mimirtool promql format <query>` command to apply the Prometheus formatter to a query
-for use in a blocked query `pattern`.
-
-Queries received by Mimir are parsed into PromQL expressions before blocking is applied.
-The `pattern` from the blocked queries is compared against the formatted representation of the parsed query,
-in order to allow consistent query blocking behavior regardless of formatting differences in the submitted queries.
+Queries received by Mimir are parsed into PromQL expressions before blocking is applied. When `regex` is
```
fionaliao: I think we should add a warning that in the regex case, they need to check the regex will match against formatted queries. It might be worth still keep the  `mimirtool promql format <query>` tip, for the regex case - point out that they can use this command with a few example queries that the regex should match.

pkg/util/validation/limits.go
```diff
@@ -578,6 +584,19 @@ func (l *Limits) validate() error {
...
 
+func (l *Limits) canonicalizeQueries() {
+	for i, q := range l.BlockedQueries {
+		if q.Regex {
+			continue
+		}
+		expr, err := parser.ParseExpr(q.Pattern)
+		if err != nil {
+			continue
+		}
```
fionaliao, line 596: Can the failing expressions be logged?

pkg/util/validation/limits.go
```diff
@@ -578,6 +584,19 @@ func (l *Limits) validate() error {
...
+		l.BlockedQueries[i].Pattern = expr.String()
```
fionaliao: Can we log the original and formatted queries here? Thinking people might get confused when they see the pattern used in the blocked query isn't the same as what they input, having a log here that says it's been formatted could be useful

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,10 @@ The order of patterns is preserved, so the first matching pattern will be used.
...
+not used, the `pattern` from the blocked queries is also similarly parsed before being compared against
+the formatted representation of the parsed query, in order to allow consistent query blocking behavior
```
tacole02:
```suggestion
the formatted representation of the parsed query. This process allows for consistent query blocking behavior
```

tacole02: I think we can break this sentence up for greater readability.

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,10 @@ The order of patterns is preserved, so the first matching pattern will be used.
...
 ## Format queries to block
 
-Use Mimirtool's `mimirtool promql format <query>` command to apply the Prometheus formatter to a query
-for use in a blocked query `pattern`.
-
-Queries received by Mimir are parsed into PromQL expressions before blocking is applied.
-The `pattern` from the blocked queries is compared against the formatted representation of the parsed query,
-in order to allow consistent query blocking behavior regardless of formatting differences in the submitted queries.
+Queries received by Mimir are parsed into PromQL expressions before blocking is applied. When `regex` is
+not used, the `pattern` from the blocked queries is also similarly parsed before being compared against
```
tacole02: Is there a reason `pattern` is code formatted here? If this is just a generic pattern, we can remove the formatting.

zenador: pattern and regex are code formatted because those are the names of fields in blocked queries, and they were already formatted that way before this change. I've removed the formatting for now but please let me know if they should be restored!

docs/sources/mimir/configure/configure-blocked-queries.md
zenador: Good point, added that back.

pkg/util/validation/limits.go
zenador: I'm having problems figuring out how to pass a proper logger in here though, there are no logs anywhere else in this file, and the limits get initialised through unmarshalling json or yaml.  Do you know how?

fionaliao: I think the cleanest way to do the logging after the unmarshalling is complete. [ValidateLimits()](https://github.com/grafana/mimir/blob/d21d5cf02c45a759efb1bd6f6504a22d29e06f4c/pkg/mimir/mimir.go#L357-L363) would be a good candidate as it's run for the default limits plus each tenant's limits. That would still need some updates to pass a logger in but seems more feasible.

However, that seems like too much complexity for the value it provides, so let's just skip the logging. The runtime_config endpoint would show the canonicalized queries anyway, if someone wanted to see them.

fionaliao:
Could you add a test for this?

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,17 @@ The order of patterns is preserved, so the first matching pattern will be used.
 
 ## Format queries to block
 
+You can ignore this section if you are not using regex. If you are, you'll need to ensure that the regex
+will match against formatted queries as follows.
```
tacole02:
```suggestion
You can ignore this section if you're not using a regular expression. If you are, you need to ensure that it matches against formatted queries as follows.
```

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,17 @@ The order of patterns is preserved, so the first matching pattern will be used.
...
+
 Use Mimirtool's `mimirtool promql format <query>` command to apply the Prometheus formatter to a query
-for use in a blocked query `pattern`.
+for use in a blocked query pattern.
 
-Queries received by Mimir are parsed into PromQL expressions before blocking is applied.
-The `pattern` from the blocked queries is compared against the formatted representation of the parsed query,
-in order to allow consistent query blocking behavior regardless of formatting differences in the submitted queries.
+Queries received by Mimir are parsed into PromQL expressions before blocking is applied. When regex is
```
tacole02:
```suggestion
Mimir parses queries into PromQL expressions before blocking is applied. When a regular expression is
```

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,17 @@ The order of patterns is preserved, so the first matching pattern will be used.
...
+not used, the pattern from the blocked queries is also similarly parsed before being compared against
+the formatted representation of the parsed query. When regex is used, the pattern is used as-is. This
```
tacole02:
```suggestion
the formatted representation of the parsed query. When a regular expression is used, the pattern is used as-is. This
```

tacole02 (approved):

docs/sources/mimir/configure/configure-blocked-queries.md
```diff
@@ -38,12 +38,17 @@ The order of patterns is preserved, so the first matching pattern will be used.
 
 ## Format queries to block
 
+You can ignore this section if you are not using regex. If you are, you'll need to ensure that the regex
+will match against formatted queries as follows.
+
 Use Mimirtool's `mimirtool promql format <query>` command to apply the Prometheus formatter to a query
-for use in a blocked query `pattern`.
+for use in a blocked query pattern.
```
fionaliao: Suggested update in wording, to make it clearer what query we're formatting:

> Use Mimirtool's `mimirtool promql format <query>` command to apply the Prometheus formatter to a query that is expected to be blocked by the regex pattern, and then check the formatted query will match the regex.

fionaliao (approved):

PR #12114, opened 2025-07-16
Add a registry value to Helm Chart

### LINKED ISSUE #5449: Add a global setting for image repository in our Mimir Helm chart

It would be useful to have a setting in the global section of our Helm chart in order to point to a different image repo instead of having to modify all the charts manually.

Just add a variable "imageRepo" in the global section of "values.yaml" that will override existing values that are pointing to different external repo.

#### Additional context

Requested by a customer who is using an internal artifact repo and had to update the Helm chart manually

jcoelho93:
#### What this PR does

Adds the `mimir.image.registry` to the Helm chart so that we can use a different registry other than docker.io

Fixes #5449

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [BUGFIX] Fix indentation in the templates that resolve `extraVolumes` values. #11202
 * [BUGFIX] Added extraVolumes to provisioner to support mounting TLS certificates. #11400
 * [ENHANCEMENT] Add values for setting annotations and labels for the rollout-operator. #6733 #11924
+* [ENHANCEMENT] Add value for setting the Mimir image registry. #5449
```
dimitarvdimitrov:
```suggestion
* [ENHANCEMENT] Add value for setting the Mimir image registry. #12114
```

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -31,6 +31,8 @@ nameOverride: null
 fullnameOverride: null
 
 image:
+  # -- Grafana Mimir container image registry - shared between Grafana Mimir and Grafana Enterprise Metrics
+  registry: docker.io
```
dimitarvdimitrov, line 35: actually, can we default to empty string so that we don't introduce unnecessary changes. I think today it's technically possible to have the `repository` as `docker.io/grafana/mimir`. After this change it would be rendered as `docker.io/docker.io/grafana/mimir`.

This made me realize that the original issue talked about doing this for all images so that users don't have to modify the image repository of each image https://github.com/grafana/mimir/issues/5449#issuecomment-1657881285 I still think doing that change will be much more useful

jcoelho93: > I still think doing that change will be much more useful

I agree, but that requires a much larger effort to implement.

dimitarvdimitrov: 💯 but i also don't think the diff as it is now changes much. Operators can just overwrite the `image.repository` and achieve the same

zenador:

PR #12121, opened 2025-07-17
10638 - Fuzz tests for MQE

### LINKED ISSUE #10638: Fuzz testing for MQE
We've already extensively tested MQE, including:

- running the upstream test cases from Prometheus against MQE
- running our own test cases against MQE
- running Prometheus' engine and MQE in parallel in some staging environments within Grafana Labs

However, PromQL is complex and there are many edge cases we may have missed. Fuzz testing MQE may help us uncover cases we've missed.

@colega created a [PoC](https://github.com/grafana/mimir/pull/10575) that already uncovered [one issue](https://github.com/prometheus/prometheus/issues/15969). We could build on this.

https://github.com/cortexproject/promqlsmith might also be helpful here: it can generate a variety of random PromQL statements.

tcp13equals2:
#### What this PR does

This PR contains additional unit tests and Go Fuzz tests which attempt to validate the Mimir Query Engine.

The tests seek to use a range of query strings to confirm that the Mimir Query Engine responds in the same manner as the Prometheus Query Engine. This includes validating both a result, result annotation and/or errors.

There are some expected / accepted scenarios where an exact match does not occur. These are detailed and accommodated in the test code.

The Fuzz tests include a range of seed / corpus entries which enables the test suite to be run as a normal unit test. Alternatively, the specific Fuzz testing function can be invoked using the go test -fuzz=<func>. 

NOTE - the unit tests for this PR is awaiting https://github.com/grafana/mimir/pull/12408

https://github.com/grafana/mimir/issues/10638

Fixes #10638

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [ ] `CHANGELOG.md` updated


pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
+# Query Fuzz Testing
+
+This file contains notes related to the addition of Fuzz tests for the Mimir query engine.
+
+Fuzz tests have been added to the `engine_fuzz_test.go` file. 
+
+These will run as a normal unit test, or they can be invoked with the Go Fuzz engine with `go test -fuzz=FuzzQuery`
+
+In either case, the test function loads a set of sample data from `seed-data.test` and then runs a range of sample queries from `seed-queries.test`.
```
charleskorn: I would avoid documenting this and instead rely on the code to explain this (otherwise it'll get out-of-sync too easily).

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+func DoComparisonQuery(t *testing.T, asInstantQuery bool, query string, prometheusEngine *promql.Engine, mimirEngine *Engine, testStorage *teststorage.TestStorage) {
+
+	var prom, mimir *promql.Result
+	var promErr, mimiErr error
+	var promQry, mimiQry promql.Query
+
+	if asInstantQuery {
+
+		ts := time.Now()
```
charleskorn: I think we'll have to use a timestamp close to the Unix epoch for these tests - `promqltest.LoadedStorage` uses the Unix epoch as the base time for all loaded data.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	f.Fuzz(func(t *testing.T, query string) {
+		DoComparisonQuery(t, true, query, prometheusEngine, mimirEngine, testStorage)
+		DoComparisonQuery(t, false, query, prometheusEngine, mimirEngine, testStorage)
+	})
+}
+
+// Test different combinations of start, end times in a range query
+// Note that we do not compare to the Prometheus engine, as it validates range inputs in it's http api handler
```
charleskorn:
```suggestion
// Note that we do not compare to the Prometheus engine, as it validates range inputs in its http api handler
```

pkg/streamingpromql/testdata/fuzz/data/seed-data.test
```diff
@@ -0,0 +1,26 @@
...
+some_metric_often_stale _ stale _ stale _
+some_metric_with_stale_marker 0 1 stale 3
+some_metric_all_gaps _ _ _ _ _
+incr_histogram{env="1"}	{{schema:0 sum:4 count:4 buckets:[1 2 1]}}+{{sum:2 count:1 buckets:[1] offset:1}}x4
+incr_histogram{env="2"}	{{schema:0 sum:4 count:4 buckets:[1 2 1]}}+{{sum:4 count:2 buckets:[1 2] offset:1}}x4
+histogram_with_gaps	{{sum:1 count:1 buckets:[1]}} {{sum:2 count:2 buckets:[1 1]}} _ {{sum:3 count:3 buckets:[1 1 1]}}
+histogram_with_stale_marker	{{sum:1 count:1 buckets:[1]}} {{sum:2 count:2 buckets:[1 1]}} stale {{sum:4 count:4 buckets:[1 1 1 1]}}
+mixed_metric {{schema:0 sum:4 count:4 buckets:[1 2 1]}} 1 2 {{schema:0 sum:3 count:3 buckets:[1 2 1]}}
+mixed_metric_histogram_first {{schema:0 sum:4 count:4 buckets:[1 2 1]}} 1
+mixed_metric_float_first 1 {{schema:0 sum:4 count:4 buckets:[1 2 1]}}
```
charleskorn: [nit] I would indent all the series beneath each `load` instruction.

pkg/streamingpromql/testutils/utils.go
```diff
@@ -87,37 +98,60 @@ func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Res
...
-					requireSpansMatch(t, h1.PositiveSpans, h2.PositiveSpans)
-					requireFloatBucketsMatch(t, h1.PositiveBuckets, h2.PositiveBuckets)
+					requireHistogramMatch(t, expectedPoint.H, actualPoint.H)
 				}
 			}
 		}
 	case parser.ValueTypeString:
 		require.Equal(t, expected.String(), actual.String())
+	case parser.ValueTypeScalar:
+		requireInEpsilonIfNotZeroOrInf(t, expected.Value.(promql.Scalar).V, actual.Value.(promql.Scalar).V)
```
charleskorn, line 112: Should we also check the timestamp on the point?

pkg/streamingpromql/testutils/utils.go
```diff
@@ -20,6 +20,17 @@ import (
...
+	if expected == nil && actual == nil {
+		return
+	} else if expected == nil {
+		require.Fail(t, `expected nil but got type: %v`, actual)
+		return
+	} else if actual == nil {
+		require.Fail(t, "expected a value but got nil")
+		return
+	}
+
```
charleskorn: Might be able to simplify this to:

```suggestion
	if expected == nil {
		require.Nil(t, actual, "expected nil result")
		return
	}
	
	require.NotNil(t, actual, "expected a result but got nil")
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	const expression = "vector(0)"
+	var startTime = time.Now().Add(-time.Hour)
+	var endTime = time.Now()
+
+	opts := NewTestEngineOpts()
+	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), NewQueryPlanner(opts), log.NewNopLogger())
+	require.NoError(f, err)
+
+	testStepSecs := []int64{-1, 0, 60, math.MaxInt64, int64(math.MaxInt64 / time.Second), int64(math.MaxInt64/time.Second) - 1}
```
charleskorn: What if we try steps that are around the length of the query (eg. just under an hour, exactly an hour, just over an hour)?

And perhaps something well over the length of the query (eg. multiple hours)?

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+	f.Fuzz(func(t *testing.T, stepSecs int64) {
+
+		step := time.Duration(stepSecs) * time.Second
+
+		_, mimErr := mimirEngine.NewRangeQuery(context.Background(), nil, nil, expression, startTime, endTime, step)
+
+		if stepSecs <= 0 || stepSecs > int64(math.MaxInt64/time.Second) {
+			require.NotNil(t, mimErr, "Expected error for StepSecs=%d", stepSecs)
+		} else {
+			require.Nil(t, mimErr, "Expected success for StepSecs=%d", stepSecs)
```
charleskorn: Can we make a stronger assertion here (eg. that the results are the same as Prometheus)?

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+			require.Nil(t, mimErr, "Expected success for start=%s, end=%s", start, end)
+		} else {
+			require.NotNil(t, mimErr, "Expected failure for start=%s, end=%s", start, end)
+		}
+	})
+}
+
+// Seed the fuzz with sample durations in a range query
+// These tests we only check the Mimir engine, as the step size is validated in the Prometheus API handler (api.go)
+// Note also that the max resolution size (end-start/step) is validated in codec.go so we don't test that here
```
charleskorn: Might be worth clarifying if the purpose of this is to check the correctness of queries with different steps, or just the correctness of validation of the step and query range. (The answer to this might invalidate some of my other comments for this test.)

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/prometheus/prometheus/promql/promqltest"
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	types.EnableManglingReturnedSlices = true
+	parser.ExperimentalDurationExpr = true
+	parser.EnableExperimentalFunctions = true
+}
```
charleskorn: This shouldn't be necessary - the `init` function in `engine_test.go` will run whenever this package runs, and it already does these things.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+// we ignore these as they include the string index of where the invalid value is in the query
+// seems like prometheus does a trim() somewhere but MQE does not, so we end up with different char index's referenced in the warning
```
charleskorn:
```suggestion
// seems like prometheus does a trim() somewhere but MQE does not, so we end up with different char indices referenced in the warning
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+func fixUpEmptyLabels(r *promql.Result) {
+	if r == nil || r.Value == nil {
+		return
+	}
+	switch r.Value.Type() {
+	case parser.ValueTypeMatrix:
+		matrix, _ := r.Matrix()
+		for i := 0; i < len(matrix); i++ {
+			if matrix[i].Metric == nil {
+				matrix[i].Metric = labels.Labels{}
```
charleskorn: [nit] `labels.EmptyLabels()` might be a clearer option over `labels.Labels{}`

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+		}
+	}
+}
+
+func DoComparisonQuery(t *testing.T, asInstantQuery bool, query string, prometheusEngine *promql.Engine, mimirEngine *Engine, testStorage *teststorage.TestStorage) {
+
+	var prom, mimir *promql.Result
+	var promErr, mimiErr error
```
charleskorn: [nit] We'd usually spell out "Mimir" in full when using it as a prefix like this (same applies to other variable names below)

```suggestion
	var promErr, mimirErr error
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+		defer promQry.Close()
+		prom = promQry.Exec(context.Background())
+	}
+
+	if mimiErr == nil {
+		defer mimiQry.Close()
+		mimir = mimiQry.Exec(context.Background())
+	}
+
+	if promErr == nil {
```
charleskorn: Would it be possible to simplify some of these conditions by starting by checking `if promErr != nil`? Then, if that is not the case, then we can `require.NoError(t, mimirErr)` and eliminate most of the `if`s

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+			assert.Equal(t, "unexpected all postings", mimir.Err.Error())
+		} else {
+			testutils.RequireEqualResults(t, query, prom, mimir, skipAnnotationComparison)
+		}
+	} else {
+		require.NotNil(t, mimiErr)
+	}
+}
+
+// This is a utility test which can be useful testing individual query strings
```
charleskorn: Is this needed? It should be possible to run the fuzz test with a filter to specify a single expression to run (or does that not work with fuzz tests?)

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	})
+}
+
+// Seed the fuzz with sample durations in a range query
+// These tests we only check the Mimir engine, as the step size is validated in the Prometheus API handler (api.go)
+// Note also that the max resolution size (end-start/step) is validated in codec.go so we don't test that here
+func FuzzRangeTestStep(f *testing.F) {
+
+	const expression = "vector(0)"
```
charleskorn: Should we test other kinds of expressions?

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
...
+
+The `seed-queries.test` file has a number of randomly generated queries related to the seed data metrics.
+
+Many of these have been generated using the [https://github.com/cortexproject/promqlsmith](https://github.com/cortexproject/promqlsmith) utility.
+
+The `demo/main.go` can be modified to not connect to a running Prometheus instance, and instead we hard code our test metrics and labels and then have the utility generate it's queries.
+
+An example modification to the `run()` is as follows;
+
+```go
```
charleskorn: Rather than including the code in a readme like this, what if we added a tool that does this to the root `tools` directory?

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
...
+
+These will run as a normal unit test, or they can be invoked with the Go Fuzz engine with `go test -fuzz=FuzzQuery`
+
+In either case, the test function loads a set of sample data from `seed-data.test` and then runs a range of sample queries from `seed-queries.test`.
+
+The test assertions are that the Prometheus PromQL Engine returns the same result as the Mimir Query Engine. Considered in this test are the result errors, result values and result annotations.
+
+There are some caveats to this comparison;
+
+* there are cases where query result annotations (error/warning) do not match. The Mimir Query Engine has AST optimisation where a right hand side expression will not be evaluated if the left side has not results
```
charleskorn: Might be worth explaining how this is handled in the tests

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
...
+* there are cases where query result annotations do not match on quantile percentile being out of bounds. The warning annotation includes a position in the original query of the invalid percentile and there is a difference in whitespace trimming of the query expressions between engines which results in a difference
+* there are some expressions not yet supported by the Mimir Query Engine, such as info, mad_over_time, sort_by_label and sort_by_label_desc
```
charleskorn: Can we remove these functions from the test corpus, so that they're not generated?

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
...
+* there are some expressions which do not have a deterministic result - ie topk(), bottomk() are undefined if there are multiple series with the same value
```
charleskorn: Can we remove these from the test corpus, or somehow ensure that they run on expressions that have unique values?

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Fuzz Usage
+
+`go test -fuzz=FuzzQuery`
```
charleskorn: [nit] I would format this (and the command on line 32) like this, given the expression isn't appearing inline in a sentence:
````suggestion
```
go test -fuzz=FuzzQuery
```
````

File: pkg/streamingpromql/engine_fuzz_test.go (hunk 13 above)
Comment (tcp13equals2) [reply pkg/streamingpromql/engine_fuzz_test.go]: Ha - you have just experienced one of my OCD quirks! Much to the angst of my fellow colleagues, I like to line up my variables so they have the same number of chars - so it looks easier to read.

But not problem - I'll adjust :)

File: pkg/streamingpromql/engine_fuzz_test.go [hunk 22]
```diff
@@ -0,0 +1,270 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/prometheus/prometheus/tree/main/promql/engine_test.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Prometheus Authors
```
lamida: I believe we don't need this header here.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/prometheus/prometheus/promql/promqltest"
+	"github.com/prometheus/prometheus/util/teststorage"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+)
+
+// A construct to hold our query engines and storage. Also holds rules for how we process certain queries
```
lamida: Nits. [Godoc guidance for type](https://go.dev/doc/comment#type) normally started with `A <type> ...`. This applicable to other instance of godoc comment that started with "A concruct..." in other type in this file.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	mimir          EngineQueryPreparationResult
+	isInstantQuery func() bool
+}
+
+// A utility to abstract the actual query preparation steps
+type InvokeQueryEnginePreparation func(input TestQueryInput) MultiEnginePrepareResult
+
+// A function to add a query to the Fuzz corpus (ie t.Add(...))
+// This is abstracted to allow for custom corpus seeding of time ranges and step size
+type SeedFuzzFunc func(t *testing.F, query string)
```
lamida: Nits. I think it is better to use `f` for all argument name of `testing.F` to avoid it getting mixed with `testing.T` in standard test function.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+// Seed queries from the given file into the Fuzz test corpus
+// Stores a map in the testContext which details queries which should have annotation comparison disabled
+func seedQueryCorpus(f *testing.F, queryFile string, seedFuzzFunc SeedFuzzFunc, testContext *TestQueryContext) {
+	queries, err := os.ReadFile(queryFile)
+	require.NoError(f, err)
+
+	ignoreAnnotations := make(map[string]bool, 500)
+
+	for _, query := range strings.Split(string(queries), "\n") {
```
lamida: I understand the file is small. But since we want to iterate line by line anyway, why don't we just use [bufio.Scanner](https://pkg.go.dev/bufio#Scanner)?

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// This is a utility function to generate PromQL queries from a set of label sets
+// The label sets can be provided in a file or retrieved from a Prometheus server
+// The queries are generated using the promqlsmith library
+// The queries are written to stdout
```
lamida: Many tools in Mimir is not having godoc for command. If we want to add it here, I think it is better to follow the guidance [here](https://go.dev/doc/comment#cmd).

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+
+	// Flag for the Prometheus server address
+	prometheusAddressFlag = "prometheus.address"
+
+	// Flag for the Prometheus matcher
+	matcherFlag = "prometheus.matcher"
+
+	// Maximum number of queries to generate
+	maxCount = 10000
+)
```
lamida: Nits. There is common pattern of config object and flag in many places in mimir codebase when we need to parse plenty of command line flag.

Usually we will create a struct config like 
```
type config struct {
  seriesSource
  promqlCount
...
}
```

And then we will have registerFlags function like
```
func (cfg *config) registerFlags(f *flag.FlagSet) {
  f.StringVar(...)
  f.IntVar(...)
...
}
```

Maybe it is better to follow that for consistency?

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+	} else if *labelsSource == "prometheus" && len(*prometheusAddress) > 0 && len(*matcher) > 0 {
+		series, err = getLabelSetsFromPrometheus(*prometheusAddress, *matcher)
+		if err != nil {
+			printUsageAndExit("error:failed to get label set from Prometheus: %s\n", err)
+		}
+
+	} else {
+		printUsageAndExit("")
+	}
```
lamida: By having `config` struct as I suggested in previous comment, we can put this validation logic in `config.validate()` method.

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+	opts := []promqlsmith.Option{
+		promqlsmith.WithEnableOffset(true),
+		promqlsmith.WithEnableAtModifier(true),
+		promqlsmith.WithEnabledFunctions(getAvailableFunctions()),
+		promqlsmith.WithEnabledBinOps(enabledBinops),
+		promqlsmith.WithEnableVectorMatching(true),
+	}
+	ps := promqlsmith.New(rnd, modelLabelSetToLabels(labelSets), opts...)
+
+	for i := 0; i < count; i++ {
```
lamida:
```suggestion
	for range count {
```

I think we can do this.

pkg/streamingpromql/compat_fixup_labels_slice.go
```diff
@@ -0,0 +1,25 @@
...
+		return
+	}
+	switch r.Value.Type() {
+	case parser.ValueTypeMatrix:
+		matrix, _ := r.Matrix()
+		for i := 0; i < len(matrix); i++ {
+			if matrix[i].Metric == nil {
+				matrix[i].Metric = labels.Labels{}
+			}
+		}
```
lamida:
```suggestion
		for i, series := range matrix {
			if series.Metric == nil {
				matrix[i].Metric = labels.Labels{}
			}
		}
```
I think this is slightly more idiomatic.

pkg/streamingpromql/compat_fixup_labels_slice.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+func FixUpEmptyLabels(r *promql.Result) {
+	if r == nil || r.Value == nil {
+		return
+	}
+	switch r.Value.Type() {
+	case parser.ValueTypeMatrix:
+		matrix, _ := r.Matrix()
```
lamida: Should we check the error here?

pkg/streamingpromql/compat_fixup_labels_stringlabels.go
```diff
@@ -0,0 +1,12 @@
...
+//go:build stringlabels
+
+package streamingpromql
+
+import (
+	"github.com/prometheus/prometheus/promql"
+)
+
+func FixUpEmptyLabels(r *promql.Result) {
+	return
```
lamida: Why do we need return here? Or is this in progress?

pkg/streamingpromql/testutils/utils.go
lamida: Has there been offline conversation that lead to resolve this comment?

pkg/streamingpromql/testutils/utils.go
```diff
@@ -87,37 +98,61 @@ func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Res
...
+	requireSpansMatch(t, expected.PositiveSpans, expected.PositiveSpans)
+	requireFloatBucketsMatch(t, expected.PositiveBuckets, expected.PositiveBuckets)
+}
+
 func requireInEpsilonIfNotZeroOrInf(t testing.TB, expected, actual float64, msgAndArgs ...interface{}) {
-	if expected == 0 || math.IsInf(expected, +1) || math.IsInf(expected, -1) {
+	if math.IsNaN(expected) && math.IsNaN(actual) {
+		// ok
+	} else if expected == 0 || math.IsInf(expected, +1) || math.IsInf(expected, -1) {
```
lamida: Actually adding this check, breaking the function name `requireInEpsilonIfNotZeroOrInf`. I think it might be better to check NaN first before calling `requireInEpsilonIfNotZeroOrInf` and just leave this function as it original state.

tools/promql-generator/README.md
```diff
@@ -0,0 +1,22 @@
+# PromQL Generator
+
+This is a utility to assist in generating random PromQL queries.
+
+It is based on the demo available in the [https://github.com/cortexproject/promqlsmith](https://github.com/cortexproject/promqlsmith) project.
+
+This utility can either;
```
lamida:
```suggestion
This utility can either:
```

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+			builder.Set(string(k), string(v))
+		}
+		out[i] = builder.Labels()
+		builder.Reset(bufLabels)
+	}
+	return out
+}
+
+// Get available functions - add entries to unsupportedFunctions map to exclude them
+func getAvailableFunctions() []*parser.Function {
```
lamida:
```suggestion
func availableFunctions() []*parser.Function {
```
Usually function in go is not using `get` prefix.

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+	res := make([]*parser.Function, 0)
```
lamida:
```suggestion
	var results []*parser.Function
```
Nits. Unless we want to return this as non nil json or object, I think just initialise this as nil slice might be better.

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	labels := make([]model.LabelSet, len(result))
+
+	for _, item := range result {
+		labelSet := make(model.LabelSet)
+		for k, v := range item {
+			labelSet[model.LabelName(k)] = model.LabelValue(v)
+		}
+		labels = append(labels, labelSet)
```
lamida: This will append `labelSet` at the end of `labels` slice initialised in line 170 above. Is this intentional? Or did you mean to `labels[i] = labelSet`, with `i` the index of result slice iteration?

tools/promql-generator/main.go
```diff
@@ -0,0 +1,242 @@
...
+
+	if len(labels) == 0 {
+		return nil, errors.New("no matching label sets found")
+	}
+
+	return labels, nil
+}
+
+// Construct a []model.LabelSet from a json file defining the label sets
+func getLabelSetsFromFile(seriesFile string) ([]model.LabelSet, error) {
```
lamida:
```suggestion
func labelSetsFromFile(seriesFile string) ([]model.LabelSet, error) {
```
Like another comment, usually go function is not using `get` or `set` prefix.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/prometheus/prometheus/util/teststorage"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+)
+
+// A construct to hold our query engines and storage. Also holds rules for how we process certain queries
+type TestQueryContext struct {
```
lamida: Do we need this to be exported?

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	prometheusEngine  *promql.Engine
+	mimirEngine       *Engine
+	testStorage       *teststorage.TestStorage
+	ignoreAnnotations map[string]bool
+}
+
+// A construct to hold our input params to each test
+type TestQueryInput struct {
```
lamida: Do we need this to be exported?

lamida: I think the same comment applies to all other new struct in this file.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+)
+
+// A construct to hold our query engines and storage. Also holds rules for how we process certain queries
+type TestQueryContext struct {
+	prometheusEngine  *promql.Engine
+	mimirEngine       *Engine
```
lamida: Can't we just use `promql.QueryEngine` interface for these two fields? I don't see we need the actual implementation of those objects in this fuzz code.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+	testStorage       *teststorage.TestStorage
```
lamida: Likewise, can't we just use `storage.Queryable` interface here, because from what I can see, ultimately we just want to pass this to `NewInstantQuery` and `NewRangeQuery`.

pkg/streamingpromql/engine_fuzz_test.go
lamida: After reading the code, instead of `TestQueryContext`, I am thinking that this object might be more fit to be named as `QueryEngines`. Also naming this with `Context` suffix keep making me thinking about Go context where we pass information along in request scope, which is not the case for this.

pkg/streamingpromql/engine_fuzz_test.go
lamida: This struct also might be more appropriately named as something like `FuzzingArg`. To make it clear that this struct fields are being fuzzed for this fuzzing.

lamida:
@tcp13equals2 thanks for the PR. This is such a great work and I learn many things about fuzzing from this PR!

I want to note these general comments:

* I see few indirection from several structs in the engine_fuzz_test.go (`TestQueryContext`, `TestQueryInput`, `EngineQueryPreparationResult`, `MultiEnginePrepareResult` and so on which can be renamed better or simplified. 
* We have `TestQueryContext` with two engines and `MultiEnginePrepareResult` also with Query object from the two engines. Can we structure this better? Maybe we can combine the struct? I am also still thinking the ideal structure but I still feel we can improve `TestQueryContext` and `MultiEnginePrepareResult`.
* Most if not all of those structs seem need not to be exported, hence can just be private.

I have also left more detail comments in several places. Most of them just nits comment.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,270 @@
...
+		// explore the bounds after we have multiple samples
+		{epoch, epoch.Add(time.Minute*5 - time.Second)},
+		{epoch, epoch.Add(time.Minute * 5)},
+		{epoch, epoch.Add(time.Minute*5 + time.Second)},
+		{epoch.Add(time.Second * 4), epoch.Add(time.Second * 61)},
+		{epoch.Add(time.Minute * 2), epoch.Add(time.Second * 61)},
+	}
+
+	// For each query string we will test for each range & step combination
+	// The fuzzer will then be able then randomise on the query, time range and/or step size
```
lamida: Maybe something like this?
```suggestion
	// Generate Fuzzing arguments with different ranges and steps combination for the passed query.
```

pkg/streamingpromql/engine_fuzz_test.go
tcp13equals2: The mimir Engine type does not fill all the same definition as promql.Engine.

ie promql.Engine has metrics *engineMetrics but this is not in the mimir Engine definition.

pkg/streamingpromql/compat_fixup_labels_slice.go
```diff
@@ -0,0 +1,29 @@
...
+func FixUpEmptyLabels(r *promql.Result) error {
+	if r == nil || r.Value == nil {
+		return nil
+	}
+	switch r.Value.Type() {
+	case parser.ValueTypeMatrix:
+		matrix, error := r.Matrix()
+		if error != nil {
+			return error
+		}
```
lamida:
```suggestion
	case parser.ValueTypeMatrix:
		matrix, err := r.Matrix()
		if err != nil {
			return err
		}
```
For errror return type usually we just use `err` local variable

pkg/streamingpromql/compat_fixup_labels_slice.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/promql/parser"
+)
+
+func FixUpEmptyLabels(r *promql.Result) error {
+	if r == nil || r.Value == nil {
+		return nil
+	}
+	switch r.Value.Type() {
```
lamida: Do we need to check for `case parser.ValueTypeVector`? It has labels too in that result.

pkg/streamingpromql/compat_fixup_labels_stringlabels.go
lamida, line 1: `FixUpEmptyLabels` is used only in *_test.go, therefore, we should just have the function in _test.go file because the function is not used in the non-test code. Simple quick fix just append compat_fixup_labels_slice and compat_fixup_labels_stringlabels with _test name suffix.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+type engineQueryPreparationResult struct {
+	query promql.Query
+	error error
+}
+
+// A multiEnginePrepareResult holds the combined engineQueryPreparationResults for both engines.
+type multiEnginePrepareResult struct {
+	prometheus     engineQueryPreparationResult
+	mimir          engineQueryPreparationResult
+	isInstantQuery func() bool
```
lamida: After seeing the usage, I don't think we need to be `func() bool` type. Just bool type should suffice.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+)
+
+// A queryEngines holds our Prometheus and Mimir QueryEngines and a Queryable (storage).
+// A queryEngines also holds rules for how we process certain queries.
+type queryEngines struct {
+	prometheus        promql.QueryEngine
+	mimir             promql.QueryEngine
+	testStorage       storage.Queryable
```
lamida:
```suggestion
	queryable       storage.Queryable
```
In my opinion let's just name this queryable too.

pkg/streamingpromql/testutils/utils.go
```diff
@@ -87,35 +98,68 @@ func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Res
...
+		return
+	}
+
+	if actual == nil {
+		require.Fail(t, "expected non-nil, got nil")
+		return
+	}
+
+	require.Equal(t, expected.Schema, actual.Schema, "histogram schemas match")
+	requireInEpsilonIfNotZeroOrInfOrNan(t, expected.Count, expected.Count, "histogram counts match")
```
lamida: Both checking expected? Do you mean to check between expected and actual? Applioes also to other instances below.

pkg/streamingpromql/testutils/utils.go
```diff
@@ -87,35 +98,68 @@ func RequireEqualResults(t testing.TB, expr string, expected, actual *promql.Res
...
+	requireFloatBucketsMatch(t, expected.PositiveBuckets, expected.PositiveBuckets)
+}
+
+// requireInEpsilonIfNotZeroOrInfOrNan will be a success if expected and actual are both NaN, 0, Inf or InEpsilon.
+func requireInEpsilonIfNotZeroOrInfOrNan(t testing.TB, expected, actual float64, msgAndArgs ...interface{}) {
+	if math.IsNaN(expected) && math.IsNaN(actual) {
+		// ok
+	} else {
+		requireInEpsilonIfNotZeroOrInf(t, expected, actual, msgAndArgs...)
+	}
```
lamida: I am still unsure on the need to exclude NaN comparison. But even if we still want to do it, better to invert the logic to avoid having empty branch.
```suggestion
	if !math.IsNaN(expected) || !math.IsNaN(actual) {
		requireInEpsilonIfNotZeroOrInf(t, expected, actual, msgAndArgs...)
	}
```

But the next question would be, is it ok to have one side NaN only?

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+	require.NoError(f, err)
+	defer file.Close()
+
+	ignoreAnnotations := make(map[string]bool, 500)
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		query := scanner.Text()
+
+		if len(query) == 0 || strings.HasPrefix(query, "#") {
```
lamida: Suggest to extract `#` to constant.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+			continue
+		}
+
+		query, found := strings.CutPrefix(query, "[ignore_annotations] ")
```
lamida: Suggest to extract `[ignore_annotations] ` to constant.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+type fuzzingArg struct {
+	// promQL query
+	query string
+	// range query start or instant query time
+	start time.Time
+	// range query end time
+	end time.Time
+	// range query step size
+	step time.Duration
+}
```
lamida: After seeing this second time, would you consider to reuse `types.QueryTimeRange` for this? It has `IsInstant` query field too, hence we can remove `isInstantQuery` from `multiEnginePrepareResult`.

pkg/streamingpromql/engine_fuzz_test.go
lamida: On my other [comment](https://github.com/grafana/mimir/pull/12121#discussion_r2228250862), I am suggesting to consider to reuse `types.QueryTimeRange` instead of fuzzArg so that we can just use IsInstant field. That comment precede the comment above, if you consider to just remove the field.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	f.Fuzz(func(t *testing.T, query string, startEpoch int64, endEpoch int64, durationMilliSecs int64) {
+		input := fuzzingArg{
+			query: query,
+			start: time.UnixMilli(startEpoch),
+			end:   time.UnixMilli(endEpoch),
+			step:  time.Duration(durationMilliSecs) * time.Millisecond,
+		}
+		doComparisonQuery(t, input, *engines, instantQueryPreparationHandler)
+		doComparisonQuery(t, input, *engines, rangeQueryPreparationHandler)
```
lamida: In my previous review I mention on simplifying the struct and object needed to setup the fuzzing. I am thinking we can just do something like this here, instead of building intermediate struct on func alias.
```
		instantQueryProm, err := engines.prometheus.NewInstantQuery(...)
		instantQueryMimir, err := engines.mimir.NewInstantQuery(...)
                 // check or compare err
		doComparisonQuery(t, input, instantQueryProm, instantQueryMimir)
                 // also do for range query
```
With that we can remove the need to use `invokeQueryEnginePreparation`, `multiEnginePrepareResult`, `engineQueryPreparationResult`. Which will make this more straighforward.

pkg/streamingpromql/testutils/utils.go
tcp13equals2: If only one side is NaN then the requireInEpsilonIfNotZeroOrInf() will fail the assertion - which is correct. 

This code is dealing with the case where both values are NaN - Go numeric equality considers NaN != NaN, but for the purposes of these tests we want to treat NaN == NaN

pkg/streamingpromql/testutils/utils.go
tcp13equals2: That's embarrassing! Well spotted!

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+)
+
+// commentLinePrefix is the prefix we use in the query seeding file to indicate a comment
+const commentLinePrefix = "#"
+
+// ignoreAnnotationDirective is the prefix we use in the query seeding to flag that we will not match on query result annotations - trailing space is intentional
+const ignoreAnnotationDirective = "[ignore_annotation] "
```
lamida: Nits. We can combine multiple `const` or `var` declaration into something like
```
const (
  foo = ...
  bar = ...
)
```

lamida (approved):
Left one last nits comment but other than that LGTM.

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -4,22 +4,38 @@ This file contains notes related to the addition of Fuzz tests for the Mimir que
...
+* [annotation] There can be a difference in annotation messages which include the query string offset as to where the error occurs. There is a difference in query string whitespace trimming which results in the reported string offsets being different. For instance `PromQL warning: quantile value should be between 0 and 1, got 2 (1:25)` - the `(1:25)` may be different.
+* [result] There can be a difference in the returned series when the `topk` and `bottomk` queries are used over a dataset which matches the same values. In this scenario the returned series is indeterminate, and the returned series from Mimir may differ from that of Prometheus 
+* [result] There are some expressions not yet supported by the Mimir Query Engine, such as info, mad_over_time, sort_by_label and sort_by_label_desc
+
+Further information on the differences can be found [here](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/references/architecture/mimir-query-engine.md#known-differences-compared-to-prometheus-engine)
```
charleskorn: [nit] Would use a relative path here given this is a reference to the same repository (eg. `../../docs/sources/mimir/references/architecture/mimir-query-engine.md#known-differences-compared-to-prometheus-engine`)

tools/promql-generator/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+* read a list of label sets from a json file
+* read a list of label sets from a Prometheus server endpoint, restricted to a given matcher
```
charleskorn: Given the main purpose of this tool is to generate input for the fuzz tests, should we drop the support for a remote Prometheus endpoint?

tools/promql-generator/main.go
```diff
@@ -0,0 +1,239 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/cortexproject/promqlsmith"
+)
+
+var (
+	// add any functions to this map to exclude them from the generated queries
+	unsupportedFunctions = map[string]struct{}{}
+
+	enabledBinops = []parser.ItemType{
```
charleskorn: Similar to the above, should we invert the logic here to list unsupported binops?

tools/promql-generator/main.go
```diff
@@ -0,0 +1,239 @@
...
+	promqlCountFlag = "promql.count"
+
+	// Flag for the file containing the label sets json data
+	labelsFileFlag = "labels.file"
+
+	// Flag for the Prometheus server address
+	prometheusAddressFlag = "prometheus.address"
+
+	// Flag for the Prometheus matcher
+	matcherFlag = "prometheus.matcher"
```
charleskorn: Given these are only used in one place, I don't think having variables for this is necessary - we can have these strings inline in the `flag.Xxx` calls below.

tools/promql-generator/main.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Maximum number of queries to generate
+	maxCount = 10000
```
charleskorn: [nit] I would make this a constant, rather than a variable.

pkg/alertmanager/multitenant_test.go
```diff
@@ -2527,7 +2527,7 @@ receivers:
 }
 
 func TestMultitenantAlertmanager_computeFallbackConfig(t *testing.T) {
-	// If no fallback configration is set, it returns a valid empty configuration.
+	// If no fallback configuration is set, it returns a valid empty configuration.
```
charleskorn: Is this strictly necessary in this PR? (Why isn't it affecting other PRs? Could this be fixed in its own PR?)

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
+# Query Fuzz Testing
+
+This file contains notes related to the addition of Fuzz tests for the Mimir query engine.
+
+Fuzz tests have been added to the `engine_fuzz_test.go` file.
```
charleskorn: [nit] Rather than document what you did, document what the state of the world is:

```suggestion
Fuzz tests are in the `engine_fuzz_test.go` file.
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+The test assertions are that the Prometheus PromQL Engine returns the same result as the Mimir Query Engine.
```
charleskorn: [nit] We generally use "Prometheus' engine" and "MQE" when referring to them:

```suggestion
The test asserts that Prometheus' engine returns the same result as MQE.
```

(same feedback applies elsewhere)

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+Considered in this test are the result errors, result values and result annotations.
+
+There are some expected cases where the Prometheus and Mimir query engines will return different responses.
+
+It is important to be aware of these differences if you use the Fuzz tool to generate new queries, as these differences will appear as test failures.
```
charleskorn: [nit] "fuzz" isn't a proper noun:

```suggestion
It is important to be aware of these differences if you use the fuzz tool to generate new queries, as these differences will appear as test failures.
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+The scenarios where a difference can occur include;
+
+- [annotation] Mimir may not emit warning annotations on binary operations that produce no series - where an expression has a left and right hand side, the right hand side expression will not generate a warning if the left hand side would evaluate to having no results. For instance, in Mimir engine the query `false / quantile_over_time(1.95, metric[5m])` would not issue a quantile value being out of range warning
```
charleskorn: We should be clearer that this is MQE-specific, not Mimir-specific (as Mimir can run with Prometheus' engine and not exhibit this behaviour)

```suggestion
- [annotation] MQE may not emit warning annotations on binary operations that produce no series - where an expression has a left and right hand side, the right hand side expression will not generate a warning if the left hand side would evaluate to having no results. For instance, in Mimir engine the query `false / quantile_over_time(1.95, metric[5m])` would not issue a quantile value being out of range warning
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+- [annotation] There can be a difference in annotation messages which include the query string offset as to where the error occurs. There is a difference in query string whitespace trimming which results in the reported string offsets being different. For instance `PromQL warning: quantile value should be between 0 and 1, got 2 (1:25)` - the `(1:25)` may be different.
+- [result] There can be a difference in the returned series when the `topk` and `bottomk` queries are used over a dataset which matches the same values. In this scenario the returned series is indeterminate, and the returned series from Mimir may differ from that of Prometheus
```
charleskorn:
```suggestion
- [result] There can be a difference in the returned series when the `topk` and `bottomk` aggregations are used over multiple series with the same values. In this scenario the returned series are indeterminate, and the returned series from MQE may differ from that of Prometheus' engine
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+- [result] There are some expressions not yet supported by the Mimir Query Engine, such as info, mad_over_time, sort_by_label and sort_by_label_desc
+
+Further information on the differences can be found [here](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/references/architecture/mimir-query-engine.md#known-differences-compared-to-prometheus-engine)
+
+If you need to exclude a test query from comparing annotations, the query can be prefixed with `[ignore_annotations]` in the data seeding file.
```
charleskorn: [nit] Would use the file name here to make this unambiguous:

```suggestion
If you need to exclude a test query from comparing annotations, the query can be prefixed with `[ignore_annotations]` in `seed-queries.test`, for example:
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+(({metric}[2y]@012))
+
+# This query will ignore comparing returned annotations
+[ignore_annotations] quAntile without () (  +(2),(A))
+```
+
+## Fuzz Usage
+
+```
+go test -fuzz=FuzzQuery
```
charleskorn: Does this need a path on the end, like this?

```suggestion
go test -fuzz=FuzzQuery .
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+When invoked the seed queries are all tested first, and then the Go Fuzzer will start generating new string queries based off this seed corpus.
+
+If a generated string fails the test assertions, it's query is written to a new file in the `testdata/fuzz` directory.
```
charleskorn:
```suggestion
If a generated string fails the test assertions, its query is written to a new file in the `testdata/fuzz` directory.
```

pkg/streamingpromql/README_FUZZ_TESTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+These failures are always re-tested when the test cases are either run as unit tests or fuzz tests.
+
+By default, the fuzz tests will continue until a failure is found. A time bound can be added to the fuzz run as follows;
+
+```
+go test -fuzz=FuzzQuery -fuzztime 30s
```
charleskorn: Same comment as above - does this need a path on the end?

```suggestion
go test -fuzz=FuzzQuery -fuzztime 30s .
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+// testInstantQueries executes the given query against both engines and asserts that the results match
+func testInstantQueries(t *testing.T, startT time.Time, query string, engines *queryEngines) {
+
+	prometheusQuery, prometheusError := engines.prometheus.NewInstantQuery(context.Background(), engines.queryable, nil, query, startT)
+	mimirQuery, mimirError := engines.mimir.NewInstantQuery(context.Background(), engines.queryable, nil, query, startT)
+
+	// if the Prometheus engine can not prepare the query then we expect the Mimir engine to also fail
+	if prometheusError != nil {
+		require.NotNil(t, mimirError)
```
charleskorn: Might be useful to include more context in the error message:
```suggestion
		require.NotNilf(t, mimirError, "Prometheus' engine returned error %v, but MQE returned no error", prometheusError)
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+	require.NoError(f, err)
+
+	engines := queryEngines{
+		prometheus: promql.NewEngine(opts.CommonOpts),
+		mimir:      mimir,
+		queryable:  buildStorage(f, dataFile),
+	}
+
+	seedQueryCorpus(f, queryFile, seedFuzzFunc, &engines)
+	return &engines
```
charleskorn: Might be simpler to make `engines` a pointer:

```suggestion
	engines := &queryEngines{
		prometheus: promql.NewEngine(opts.CommonOpts),
		mimir:      mimir,
		queryable:  buildStorage(f, dataFile),
	}

	seedQueryCorpus(f, queryFile, seedFuzzFunc, engines)
	return engines
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// fixUpAndAssertPostingErrors addresses an engine mismatch in a postings error message.
+// For the relevant Prometheus error, this function will manually assert for the expected Mimir error string and modify the Mimir error so it will pass a later RequireEqualResults() test.
+func fixUpAndAssertPostingErrors(t *testing.T, prom, mimir *promql.Result) {
+	if mimir != nil && prom != nil && prom.Err != nil && prom.Err.Error() == "expanding series: unexpected all postings" {
+		assert.Equal(t, "unexpected all postings", mimir.Err.Error())
```
charleskorn: Why `assert.Equal` and not `require.Equal` here?

pkg/streamingpromql/testutils/utils.go
charleskorn: I think this comment still stands.

tools/promql-generator/sample_series.js
charleskorn, line 1: [nit] I would name this file `sample_series.json`, given that's the extension we use elsewhere to denote a file containing JSON.

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
charleskorn, line 1: Given this file is based on the example from https://github.com/cortexproject/promqlsmith, we should include a provenance comment here as well.

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
...
+
+// A configOptionInt is an extension of configOption and adds in a maxValue attribute
+type configOptionInt struct {
+	config   configOption[int]
+	maxValue int
+}
+
+// Our default and max PronQL query counts which a user can select
+const maxQueryCount int = 10000
+const defQueryCount int = 1000
```
charleskorn: Given this value is only used in one place, do we need a constant for this?

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
...
+	flag *T
+}
+
+// A configOptionInt is an extension of configOption and adds in a maxValue attribute
+type configOptionInt struct {
+	config   configOption[int]
+	maxValue int
+}
+
+// Our default and max PronQL query counts which a user can select
```
charleskorn:
```suggestion
// Our default and max PromQL query counts which a user can select
```

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
...
+	defaultValue T
+	// The value user input once parsed
+	flag *T
+}
+
+// A configOptionInt is an extension of configOption and adds in a maxValue attribute
+type configOptionInt struct {
+	config   configOption[int]
+	maxValue int
+}
```
charleskorn: This is quite different to how we handle configuration flags elsewhere - is there a particularly strong reason to diverge from the existing pattern rather than defining a type for config flags like this?

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
...
+		query := expr.Pretty(0)
+
+		fmt.Printf("%s\n", strings.ReplaceAll(query, "\n", ""))
+
+	}
+	return nil
+}
+
+// modelLabelSetToLabels converts []model.LabelSet to []labels.Labels
+func modelLabelSetToLabels(labelSets []model.LabelSet) []labels.Labels {
```
charleskorn: Given we never use the `model.LabelSet` directly, should we parse the labels from the remote server or JSON file directly into `labels.Labels`? Then we wouldn't need to do this conversion.

tools/promql-generator/promql-generator.go
```diff
@@ -0,0 +1,354 @@
...
+// getAvailableFunctions returns a []*parser.Function excluding functions listed in the unsupportedFunctions map
+func availableFunctions() []*parser.Function {
+	var results []*parser.Function
+	for _, f := range parser.Functions {
+		if f.Variadic != 0 {
+			continue
+		}
+		if slices.Contains(f.ArgTypes, parser.ValueTypeString) {
+			continue
+		}
```
charleskorn, line 199: Might be worth adding a comment explaining why we skip functions in these two cases.

tools/promql-generator/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+promql-generator -series.source=prometheus -prometheus.address=<url> -prometheus.matcher=<matcher> [-promql.count=<int>]
+```
+
+The flag variables are;
+
+- labels.file=<labels_file.js> (required for -series.source=file, default=sample_series.js) - a file which contains a json representation of 1 or more label sets. An example file is provided.
+- prometheus.address=<url> (required for -series.source=prometheus, default=http://127.0.0.1:9009/prometheus) - a fully qualified URL to a Prometheus server.
+- prometheus.matcher=<matcher> (required for -series.source=prometheus, default={job="prometheus"}) - a matcher to limit the label sets discovered from the Prometheus server.
+- promql.count=<int> (optional default=100) - the number of queries generated. 0 < count < 10000
```
charleskorn: As much as possible, I would avoid repeating the CLI flag documentation here and instead point readers towards running `promql-generator -help` to see the available CLI flags.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+func executeQueriesAndCompareResults(t *testing.T, engines queryEngines, mimirQuery promql.Query, prometheusQuery promql.Query, query string) {
+	// Execute our queries
+	defer prometheusQuery.Close()
+	prom := prometheusQuery.Exec(context.Background())
+
+	defer mimirQuery.Close()
+	mimir := mimirQuery.Exec(context.Background())
+
+	// Replace series Metric which are nil with an empty Labels{}
+	err := FixUpEmptyLabels(prom)
```
charleskorn: Is this something that should move to `RequireEqualResults` perhaps? Why is this only a problem during fuzz testing?

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+const (
+	// commentLinePrefix is the prefix we use in the query seeding file to indicate a comment
+	commentLinePrefix = "#"
+	// ignoreAnnotationDirective is the prefix we use in the query seeding to flag that we will not match on query result annotations - trailing space is intentional
+	ignoreAnnotationDirective = "[ignore_annotation] "
+)
+
+// A queryEngines holds our Prometheus and Mimir QueryEngines and a Queryable (storage).
+// A queryEngines also holds rules for how we process certain queries.
+type queryEngines struct {
```
charleskorn: [nit] Perhaps `fuzzConfig` or `fuzzTestEnvironment` would be clearer names given this struct holds more than just some engines? Then we could change the name of `buildQueryEngines` to match, given it does more than just create query engines.

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+		for _, step := range steps {
+			for _, rang := range ranges {
+				f.Add(query, rang[0].UnixMilli(), rang[1].UnixMilli(), step.Milliseconds())
+			}
+		}
+	}
+
+	engines := buildQueryEngines(f, "testdata/fuzz/data/seed-data.test", "testdata/fuzz/data/seed-queries.test", seedFuzzFunc)
+
+	f.Fuzz(func(t *testing.T, query string, startEpoch int64, endEpoch int64, durationMilliSecs int64) {
```
charleskorn, line 225: Is there a reason why the time and step parameters are integers rather than `time.Time` and `time.Duration` values?

tcp13equals2: Yes, the Go Fuzz interface only allows for simple types - ints, bools, strings etc. I'll add a comment to explain this

pkg/alertmanager/multitenant_test.go
tcp13equals2: It was stopping me from being able to run the full linter checks locally. I can make a separate PR for this.

pkg/streamingpromql/README_FUZZ_TESTING.md
tcp13equals2: Seems to run with or without a path on the end. But will add the .

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	seedQueryCorpus(f, queryFile, seedFuzzFunc, environment)
+	return environment
+}
+
+// fixUpAndAssertPostingErrors addresses an engine mismatch in a postings error message.
+// For the relevant Prometheus error, this function will manually assert for the expected MQE error string and modify the MQE error so it will pass a later RequireEqualResults() test.
+func fixUpAndAssertPostingErrors(t *testing.T, prom, mqe *promql.Result) {
+	if mqe != nil && prom != nil && prom.Err != nil && prom.Err.Error() == "expanding series: unexpected all postings" {
+		require.Equal(t, "unexpected all postings", mqe.Err.Error())
```
charleskorn: [nit] `require.EqualError` might simplify this slightly:
```suggestion
		require.EqualError(t, mqe.Err, "unexpected all postings")
```

pkg/streamingpromql/engine_fuzz_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+// testInstantQueries executes the given query against both engines and asserts that the results match
+func testInstantQueries(t *testing.T, startT time.Time, query string, testEnvironment *fuzzTestEnvironment) {
+
+	prometheusQuery, prometheusError := testEnvironment.prometheus.NewInstantQuery(context.Background(), testEnvironment.queryable, nil, query, startT)
+	mqeQuery, mqeError := testEnvironment.mqe.NewInstantQuery(context.Background(), testEnvironment.queryable, nil, query, startT)
+
+	// if the Prometheus engine can not prepare the query then we expect the MQE to also fail
+	if prometheusError != nil {
+		require.NotNilf(t, mqeError, "Prometheus' engine returned error %v, but MQE returned no error", prometheusError)
```
charleskorn: [nit] `require.Errorf` will give a slightly clearer assertion failure message:
```suggestion
		require.Errorf(t, mqeError, "Prometheus' engine returned error %v, but MQE returned no error", prometheusError)
```

charleskorn (approved):
LGTM modulo nits:

PR #12122, opened 2025-07-17
MQE: refactor responsibilities of `Engine` and `Query`, introduce `Evaluator` and `Materializer`

charleskorn:
#### What this PR does

This PR refactors the responsibilities of some of the types in MQE, and introduces some new types. The goal is to make implementing remote execution in queriers of query plans created by query-frontends easier.

Main changes:
* The `Materializer` type has been introduced: it owns the responsibility of converting a query plan into operators (which was previously the split betwen `Engine` and `Query`)
* The `Evaluator` type has been introduced: it owns the responsibility of evaluating an operator and passing the data to the evaluation observer. In the future, it'll support evaluating multiple related operators from the same query plan.
* The `Query` type now exists primarily to provide an implementation of the `promql.Query` interface, bridging the interface that Prometheus and its querying API expect with how MQE operates

There should be no user-visible behaviour changes.

Once we have remote execution of query plans, I expect the interactions between types to look something like this:
* query-frontends create a query plan by calling `Engine.NewInstantQuery` or `Engine.NewRangeQuery`, and pass some or all of the query plan to queriers
* queriers materialize the appropriate part of the query plan with `Materializer`, pass this to an `Evaluator`, and stream results back to query-frontends as they are produced by the `Evaluator`

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/query.go
+++ b/pkg/streamingpromql/query.go
@@ -44,7 +44,7 @@ type Query struct {
 	annotations    *annotations.Annotations
 	stats          *types.QueryStats
 
-	resultIsVector bool
+	resultIsVector bool // This is necessary as we need to know what kind of result to return (vector or matrix) if the result is empty.
 
 	succeeded bool
 }


pkg/streamingpromql/engine_test.go
```diff
@@ -1482,8 +1482,8 @@ func TestMemoryConsumptionLimit_SingleQueries(t *testing.T) {
 			rangeQueryExpectedPeak: 8*types.HistogramPointerSize + 8*types.HPointSize + types.SeriesMetadataSize,
 			rangeQueryLimit:        8*types.HistogramPointerSize + 8*types.HPointSize + types.SeriesMetadataSize,
 
-			instantQueryExpectedPeak: types.HistogramPointerSize + types.HPointSize + types.VectorSampleSize + types.SeriesMetadataSize,
-			instantQueryLimit:        types.HistogramPointerSize + types.HPointSize + types.VectorSampleSize + types.SeriesMetadataSize,
+			instantQueryExpectedPeak: types.HPointSize + types.VectorSampleSize + types.SeriesMetadataSize,
+			instantQueryLimit:        types.HPointSize + types.VectorSampleSize + types.SeriesMetadataSize,
```
charleskorn, line 1486: Note to reviewers: this change (and the change below) are necessary because the order of allocations has changed, so the peak memory consumption is slightly lower.

pkg/streamingpromql/materialize.go
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package streamingpromql
+
+import (
+	"github.com/grafana/mimir/pkg/streamingpromql/planning"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+type Materializer struct {
```
56quarters, line 12: Can you add a note here that indicates this type isn't concurrency safe?

pkg/streamingpromql/query.go
```diff
@@ -180,270 +78,238 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
...
+	q.succeeded = true
+	return result
 }
 
-func (q *Query) populateStringFromStringOperator(o types.StringOperator) promql.String {
-	return promql.String{
-		T: q.topLevelQueryTimeRange.StartT,
-		V: o.GetValue(),
-	}
+func (q *Query) SeriesMetadataEvaluated(evaluator *Evaluator, series []types.SeriesMetadata) error {
```
56quarters, line 129: WDYT about indicating that this for the `EvaluationObserver` interface?

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if err := observer.EvaluationCompleted(e, e.annotations, e.stats); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (e *Evaluator) evaluateInstantVectorOperator(ctx context.Context, o types.InstantVectorOperator, observer EvaluationObserver) error {
```
56quarters: It feels a bit silly but WDYT about adding unit tests for the `Evaluator` to make sure it calls the relevant methods on the observer?

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+
+		if err := observer.InstantVectorSeriesDataEvaluated(e, seriesIdx, d); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (e *Evaluator) evaluateRangeVectorOperator(ctx context.Context, o types.RangeVectorOperator, observer EvaluationObserver) error {
```
56quarters: nit: since we have an operator type and observer type, WDYT about using `op` instead of `o` for operators so it's more obvious what it is?

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if e.stats != nil {
+		e.stats.Close()
+	}
+}
+
+type EvaluationObserver interface {
+	// SeriesMetadataEvaluated notifies this observer when series metadata has been evaluated.
+	// Implementations of this method are responsible for returning the series slice to the pool when it is no longer needed.
```
56quarters, line 222: Are series metadata or instant vector series data allowed to be mutated?

pkg/streamingpromql/query.go
```diff
@@ -7,151 +7,49 @@ package streamingpromql
...
-func (e *Engine) newQuery(ctx context.Context, queryable storage.Queryable, opts promql.QueryOpts, timeRange types.QueryTimeRange, originalExpression string) (*Query, error) {
-	if opts == nil {
-		opts = promql.NewPrometheusQueryOpts(false, 0)
-	}
-
-	lookbackDelta := opts.LookbackDelta()
-	if lookbackDelta == 0 {
-		lookbackDelta = e.lookbackDelta
-	}
+	resultIsVector bool
```
56quarters: Can you add a comment explaining why this is required? From what I can see this is now we tell the difference between a result that returns a vector and a result that _should_ return a vector but doesn't have any results?

56quarters (commented):
Overall this LGTM with a few small suggestions.

pkg/streamingpromql/materialize.go
```diff
@@ -0,0 +1,48 @@
...
+		return f.Produce()
+	}
+
+	childTimeRange := node.ChildrenTimeRange(timeRange)
+	childrenNodes := node.Children()
+	childrenOperators := make([]types.Operator, 0, len(childrenNodes))
+	for _, child := range childrenNodes {
+		o, err := m.ConvertNodeToOperator(child, childTimeRange)
+		if err != nil {
+			return nil, err
```
lamida, line 36: If the err is non-nil, will operator.Close() going to be ever called? I see in the code, root.Close() will only be called in Evaluator.Evaluate(), which we will never reach to, if we got error here. If operator.Close() consist of slice memory pooling operation, missed call to Close() due to this error handling can result with inconsistent memory pooling state.

pkg/streamingpromql/materialize.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		childrenOperators = append(childrenOperators, o)
+	}
+
+	f, err := node.OperatorFactory(childrenOperators, timeRange, m.operatorParams)
+	if err != nil {
+		return nil, err
```
lamida, line 44: Same comment like before. If the err is non-nil, will operator.Close() going to be ever called? I see in the code, root.Close() will only be called in Evaluator.Evaluate(), which we will never reach to, if we got error here.

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/prometheus/prometheus/blob/main/promql/engine.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Prometheus Authors
```
lamida: Nits. Do we need prometheus attribution here?

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+	annotations              *annotations.Annotations
+	stats                    *types.QueryStats
+	cancel                   context.CancelCauseFunc
+}
+
+func NewEvaluator(root types.Operator, params *planning.OperatorParameters, timeRange types.QueryTimeRange, engine *Engine, opts promql.QueryOpts, activityTrackerDescription string) (*Evaluator, error) {
+	stats, err := types.NewQueryStats(timeRange, engine.enablePerStepStats && opts.EnablePerStepStats(), params.MemoryConsumptionTracker)
+	if err != nil {
+		return nil, err
+	}
```
lamida, line 38: Nits. What if we initialise stats outside of NewEvaluator and passing it in the NewEvaluator parameter? With that we don't need to pass timeRange and opts anymore.

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+	d, err := o.GetValues(ctx)
+	if err != nil {
+		return err
+	}
+
+	if err := observer.ScalarEvaluated(e, d); err != nil {
+		return err
+	}
+
+	return nil
```
lamida: Can't we just do this?

```suggestion
	return observer.ScalarEvaluated(e, d)
```

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (e *Evaluator) evaluateStringOperator(o types.StringOperator, observer EvaluationObserver) error {
+	v := o.GetValue()
+	if err := observer.StringEvaluated(e, v); err != nil {
+		return err
+	}
+
+	return nil
```
lamida: Can't we just do this?

```suggestion
	return observer.StringEvaluated(e, v)
```

pkg/streamingpromql/evaluator.go
```diff
@@ -0,0 +1,256 @@
...
+	// To make comparing to Prometheus' engine easier, only return the annotations if there are some, otherwise, return nil.
+	if len(*e.annotations) == 0 {
+		e.annotations = nil
+	}
+
+	if err := observer.EvaluationCompleted(e, e.annotations, e.stats); err != nil {
+		return err
+	}
+
+	return nil
```
lamida: Can we simplify to this?
```suggestion
	return observer.EvaluationCompleted(e, e.annotations, e.stats)
```

pkg/streamingpromql/engine.go
```diff
@@ -126,18 +131,76 @@ func (e *Engine) NewRangeQuery(ctx context.Context, q storage.Queryable, opts pr
...
+	q := &Query{
+		evaluator:                evaluator,
+		engine:                   e,
+		statement:                statement,
+		memoryConsumptionTracker: memoryConsumptionTracker,
+		originalExpression:       plan.OriginalExpression,
+		topLevelQueryTimeRange:   plan.TimeRange,
+	}
+
+	return q, nil
```
lamida: Nits. What if we just return directly without assigning to q first?
```suggestion
	return &Query{
		evaluator:                evaluator,
		engine:                   e,
		statement:                statement,
		memoryConsumptionTracker: memoryConsumptionTracker,
		originalExpression:       plan.OriginalExpression,
		topLevelQueryTimeRange:   plan.TimeRange,
	}, nil
```

pkg/streamingpromql/engine.go
```diff
@@ -126,18 +131,76 @@ func (e *Engine) NewRangeQuery(ctx context.Context, q storage.Queryable, opts pr
...
+	statement := &parser.EvalStmt{
+		Expr:          nil, // Nothing seems to use this, and we don't have a good expression to use here anyway, so don't bother setting this.
+		Start:         timestamp.Time(plan.TimeRange.StartT),
+		End:           timestamp.Time(plan.TimeRange.EndT),
+		Interval:      time.Duration(plan.TimeRange.IntervalMilliseconds) * time.Millisecond,
+		LookbackDelta: lookbackDelta,
+	}
+
+	if plan.TimeRange.IsInstant {
+		statement.Interval = 0 // MQE uses an interval of 1ms in instant queries, but the Prometheus API contract expects this to be 0 for instant queries.
```
lamida: Nits. Simplify the comment. The second clause should clear that we refer to instant queries too.
```suggestion
		statement.Interval = 0 // MQE uses an interval of 1ms in instant queries, but the Prometheus API contract expects this to be 0.
```

lamida (commented):
General refactoring looks quite obvious. I have left some comments. Most of them are non critical feedbacks.

pkg/streamingpromql/evaluator.go
charleskorn: Good point, addressed in aa5a421

pkg/streamingpromql/query.go
charleskorn: Added in aa5a421.

One thing I was thinking about was introducing a separate `queryObserver` type or something like that for `Query` to use, but I'm not sure that's necessary - what do you think?

pkg/streamingpromql/materialize.go
charleskorn: Good point, addressed in aa5a421

pkg/streamingpromql/evaluator.go
charleskorn: The observer effectively receives ownership of the slices, so it can do whatever it wants - I've made this clearer in 4554a12.

pkg/streamingpromql/materialize.go
charleskorn: This matches the existing behaviour before this PR. Worst case scenario, we'll end up not returning some slices to the pool, which is safe.

pkg/streamingpromql/materialize.go
charleskorn: See above - I don't think this is an issue.

pkg/streamingpromql/evaluator.go
charleskorn: Good point, removed in 07fb3fc.

pkg/streamingpromql/evaluator.go
charleskorn: Good catch, fixed in 7f63aa7.

pkg/streamingpromql/evaluator.go
charleskorn: Good catch, fixed in 7f63aa7.

pkg/streamingpromql/evaluator.go
charleskorn: Good catch, fixed in cd35bc2.

pkg/streamingpromql/engine.go
charleskorn: Changed in 7d527b6.

pkg/streamingpromql/engine.go
charleskorn: Changed in 7d527b6.

pkg/streamingpromql/evaluator.go
charleskorn: This is true, but then anywhere we need an evaluator needs to know how to correctly initialise the `QueryStats`, which doesn't feel like the right division of responsibilities to me.

pkg/streamingpromql/query.go
charleskorn: How's what I've added in 198bc7f?

pkg/streamingpromql/query.go
56quarters: > One thing I was thinking about was introducing a separate queryObserver type or something like that for Query to use, but I'm not sure that's necessary - what do you think?

I actually had a similar thought. I don't think it's necessary as part of this PR but maybe if we find that the observer related methods in `Query` get more involved or are changed more often than the rest of the struct.

PR #12123, opened 2025-07-17
chore(localdev): Bump Prometheus in local development docker to 3.5.0 LTS

krajorama:
Has NHCB fix https://github.com/prometheus/prometheus/pull/16720

Use the new global variable for always_scrape_classic_histograms option to not have to repeat it.

Let prometheusRW2 profile run in parallel with the prometheus profile. Alter scraped_by label and external port.
For testing
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41089

### DIFF
--- a/development/mimir-monolithic-mode/config/prometheusRW2.yaml
+++ b/development/mimir-monolithic-mode/config/prometheusRW2.yaml
@@ -10,7 +10,6 @@ scrape_configs:
       - targets: ['mimir-1:8001']
         labels:
           container: 'mimir-1'
-
   - job_name: mimir-2
     static_configs:
       - targets: ['mimir-2:8002']


development/mimir-monolithic-mode/config/prometheusRW2.yaml
```diff
@@ -1,21 +1,21 @@
...
+  always_scrape_classic_histograms: true
 
 scrape_configs:
   - job_name: mimir-1
     static_configs:
       - targets: ['mimir-1:8001']
         labels:
           container: 'mimir-1'
-    always_scrape_classic_histograms: true
+
```
krajorama:
```suggestion
```

PR #12124, opened 2025-07-17
Helm: add Keda addition triggers and make default ones optional

QuantumEnigmaa:
#### What this PR does

This PR updates the Keda `ScaledObjects` resources templates in 2 ways:

- It makes the current default triggers optional (the option is set to "true" by default in the values to avoid breaking changes).
- It adds the possibility to easily add triggers from the values.

No issues linked to this PR.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

QuantumEnigmaa:
Hey @tacole02, thanks for reviewing ! What's blocking before merging this ? Is there anything I can do to unblock the situation ?

tacole02:
Hi @QuantumEnigmaa thanks for your contribution. A code owner needs to review this PR before it can be merged. I've added the relevant people as reviewers. Thanks for your patience!

zenador:

QuantumEnigmaa:
@tacole02 any news concerning this PR ?

dimitarvdimitrov (commented):
hey, can you provide some example use case for how these options are used? what other triggers do you have? I don't see how much you gain from this change compared to implementing your own ScaledObject for each scaled component instead

QuantumEnigmaa:
> I don't see how much you gain from this change compared to implementing your own ScaledObject for each scaled component instead

That's the point: being able to only define triggers from the values instead of having to define an entire ScaledObject resource.

> can you provide some example use case for how these options are used? 

For example, in addition to the default triggers, one could want to define an additional trigger based on Loki that would scale the target according to specific logs. One could want to scale down to the minimum the mimir-queriers by default if no one connected to grafana recently and only scale those up if the grafana logs show that someone just connected.

PR #12128, opened 2025-07-17
fix(ingester): Keep read only when any number of compactions in progress

### LINKED ISSUE #11850: Bug: Ingesters can exit read-only mode during TSDB compactions

periklis:
#### What this PR does

This pull request is a refinement of the previous attempt to keep read-only state when any kind of compaction in progress (See #11890).  In contrast the previous PR removed accidentaly the `cortex_ingester_tsdb_forced_compactions_in_progress` metric that is used in downstream alerts.

Fixes #11850

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


tacole02 (approved):

pkg/ingester/downscale.go
```diff
@@ -52,8 +52,8 @@ func (i *Ingester) PrepareInstanceRingDownscaleHandler(w http.ResponseWriter, r
 		i.circuitBreaker.push.deactivate()
 
 	case http.MethodDelete:
-		// Don't leave read-only mode if there's a forced compaction pending or in progress.
-		if len(i.forceCompactTrigger) > 0 || i.forcedCompactionInProgress.Load() {
+		// Don't leave read-only mode if there is any compaction pending or in progress.
+		if i.numCompactionsInProgress.Load() != 0 {
 			msg := "cannot clear read-only mode while forced compaction is in progress"
```
pr00se:
```suggestion
			msg := "cannot clear read-only mode while a compaction is pending or in progress"
```

pkg/ingester/ingester.go
```diff
@@ -3301,16 +3311,29 @@ func (i *Ingester) compactionServiceInterval() (firstInterval, standardInterval
 	return
 }
 
+// prepareCompaction is incrementing the atomic counter of the number of compactions in progress.
+// It also exposes a metric tracking the number of compactions in progress.
+// It returns a callback that should be called when the compaction is finished, e.g. in defer statement.
+// This callback is decrementing the atomic counter of the number of compactions in progress.
+func (i *Ingester) prepareCompaction() func() {
```
pr00se: since we've removed the other metric, do we still need this function? it might be clearer to put the counter increment/decrement directly in the two call sites, rather than have to look up what `prepareCompaction` does.

periklis: Yes you are absolutely right. It is easier to reason inline with a atomic inc/dec rather than masking this in a separate function with a bad name :)

pkg/ingester/downscale.go
periklis: ack!

PR #12136, opened 2025-07-18
parquet: optimize query sharding by leveraging materialized series hooks 

npazosmendez:
Leverage this feature https://github.com/prometheus-community/parquet-common/pull/83 to optimize query sharding. Previously we were materializing all chunks and post-filtering them before sending through the wire. Now we can push down this filter to the materializer and avoid loading them entirely.


npazosmendez:
> Not sure how hard would be to test this

There's already a test for sharding, this is just changing the implementation. I do admit the existing tests could be better, but then again so do all our Parquet tests for now

Issue #4707, opened 2023-04-11
Docs: should list all required ports in one place


bboreham:
People who want to configure their network to run Mimir want to know what ports are used.
Example: https://community.grafana.com/t/mimir-port-9095/82272/2

#### Describe the solution that you’d like or the expected outcome

A page, or part of a page, listing all the ports, their function, and perhaps which component (ingester, distributor, etc.) they are used on.

osg-grafana:
@bboreham, which SME would you like to delegate this to? I can work with that person to move this forward.

tacole02:
@zhehao-grafana could you please comment on whether this issue still requires documentation, and if so, provide a relative priority? Thanks!

PR #12141, opened 2025-07-20
Add docs on Mimir ports

tacole02:
#### What this PR does

This PR adds a page overviewing all of the network ports Mimir uses.

Fixes https://github.com/grafana/mimir/issues/4707

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.

### DIFF
--- a/docs/sources/mimir/references/architecture/ports.md
+++ b/docs/sources/mimir/references/architecture/ports.md
@@ -1,0 +1,36 @@
+---
+description: Learn about the network ports that Grafana Mimir uses
+menuTitle: Netowrk ports
+title: About Grafana Mimir network ports
+weight: 25
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# About Grafana Mimir network ports
+
+Grafana Mimir uses various network ports to facilitate communication between its internal components, external services like Prometheus and Grafana, and for overall cluster operation. Proper port configuration is crucial for setting up your Mimir cluster, configuring firewalls, and ensuring secure communication between Mimir components and integrated tools.
+
+The ports required to run Grafana Mimir can vary slightly depending on your deployment mode and whether you're using additional components like Grafana or a load balancer.
+
+## Core Mimir ports
+
+The following table shows the core Mimir ports.These ports are fundamental to operating Mimir, whether in a monolithic or distributed setup.
+
+| Port | Function                  | Related components | Description                                                                                                                                                                                                 |
+| :--- | :------------------------ | :-------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 9009 | HTTP API / remote write   | Distributor, Querier        | This is the main entry point for Prometheus to remote-write metrics to Mimir through the Distributor and for Grafana and Prometheus to query data through the Querier or Query-frontend. This is the primary user-facing port. |
+| 9095 | Internal gRPC communication | All Mimir components      | Used for high-performance communication between different Mimir components, such as Distributor to Ingester, or Querier to Ingester. This communication is essential for distributed deployments.                                |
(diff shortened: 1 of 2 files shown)


docs/sources/mimir/references/architecture/ports.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Core Mimir ports
+
+The following table shows the core Mimir ports.These ports are fundamental to operating Mimir, whether in a monolithic or distributed setup.
+
+| Port | Function                    | Related components   | Description                                                                                                                                                                                                                    |
+| :--- | :-------------------------- | :------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 9009 | HTTP API / remote write     | Distributor, Querier | This is the main entry point for Prometheus to remote-write metrics to Mimir through the Distributor and for Grafana and Prometheus to query data through the Querier or Query-frontend. This is the primary user-facing port. |
```
56quarters: This isn't correct. This particular port is only used in the Docker-compose setup which picks different ports for various services so they can all run on the same (developer) machine. That's not how people will be operating Mimir in production. Default for the HTTP server is `8080`

```
# HTTP server listen port.
# CLI flag: -server.http-listen-port
[http_listen_port: <int> | default = 8080]
```

This also isn't usually exposed since Mimir is behind either and Nginx proxy, the GEM gateway, or Kubernetes ingress (all of which would expose `80` or `443`).

docs/sources/mimir/references/architecture/ports.md
```diff
@@ -0,0 +1,36 @@
...
+| 9095 | Internal gRPC communication | All Mimir components | Used for high-performance communication between different Mimir components, such as Distributor to Ingester, or Querier to Ingester. This communication is essential for distributed deployments.                              |
+| 7946 | Memberlist / Gossip         | All Mimir components | Used for service discovery and maintaining the consistent hash ring that allows Mimir components to find and communicate with each other. This process is critical for high availability and scaling.                          |
+
+## Additional ports
```
56quarters: I don't think we should maintain a list like this here. Maybe we could link to documentation for Grafana, Alloy, or popular exporters. I don't think we should be trying to include everything in Mimir documentation when it's highly variable for different environments and documented elsewhere.

docs/sources/mimir/references/architecture/ports.md
```diff
@@ -0,0 +1,36 @@
...
+weight: 25
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# About Grafana Mimir network ports
+
+Grafana Mimir uses various network ports to facilitate communication between its internal components, external services like Prometheus and Grafana, and for overall cluster operation. Proper port configuration is crucial for setting up your Mimir cluster, configuring firewalls, and ensuring secure communication between Mimir components and integrated tools.
+
+The ports required to run Grafana Mimir can vary slightly depending on your deployment mode and whether you're using additional components like Grafana or a load balancer.
```
56quarters, line 14: I'd also note that the values documented below are defaults and can be changed by Mimir configuration.

docs/sources/mimir/references/architecture/ports.md
56quarters: Also, this is applicable to all components. They all expose `/metrics` at least over HTTP and usually some administrative API.

docs/sources/mimir/references/architecture/ports.md
tacole02: I think this is a good point. I'm going to remove the list, I don't think it belongs in the Mimir docs.

docs/sources/mimir/references/architecture/ports.md
tacole02: Thanks for the clarification here, @56quarters . I've updated this for accuracy, and hopefully to provide some helpful context for users. Would you mind taking a look?

PR #12151, opened 2025-07-21
feat(parquet): Update parquet bucket stores to cache metadata of label files

jesusvazquez:
This PR:
* Extends metadata cache to also cache the metadata of parquet label files. (We could add the metadata of parquet chunk files here in the future as well)
* Updates the constructor of parquet bucket stores to initialize a caching bucket that its then passed down. In theory when the lazy reader invokes `ParquetBucketOpener.Open() -> bkt.Attributes()`, `CachingBucket.Attributes()` should check cache for `*.labels.parquet` files.


pkg/storegateway/parquet_bucket_stores.go
```diff
@@ -82,6 +83,16 @@ func NewParquetBucketStores(
 	reg prometheus.Registerer,
 ) (*ParquetBucketStores, error) {
 
+	chunksCacheClient, err := cache.CreateClient("parquet-chunks-cache", cfg.BucketStore.ChunksCache.BackendConfig, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
+	if err != nil {
+		return nil, errors.Wrapf(err, "parquet-chunks-cache")
+	}
+
+	cachingBucket, err := tsdb.CreateCachingBucket(chunksCacheClient, cfg.BucketStore.ChunksCache, cfg.BucketStore.MetadataCache, bkt, logger, reg)
```
jesusvazquez, line 91: These lines initialize the caching bucket the same way its currently done in the bucket stores. Check [this reference](https://github.com/grafana/mimir/blob/699a122aba83da75ccef884d63af468da4f5127e/pkg/storegateway/bucket_stores.go#L93-L100)

For us right now we're not caching the chunks and cfg.BucketStore.ChunksCache can be empty. But just by properly setting the config in cfg.BucketStore.MetadataCache we should get the attribute calls cached.

pkg/storegateway/parquet_bucket_stores.go
```diff
@@ -125,24 +136,24 @@ func NewParquetBucketStores(
...
 		Help: "Number of tenants discovered in the bucket.",
 	})
 	stores.tenantsSynced = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
-		Name: "cortex_bucket_stores_tenants_synced",
+		Name: "cortex_parquet_bucket_stores_tenants_synced",
 		Help: "Number of tenants synced.",
 	})
 	stores.blocksLoaded = prometheus.NewDesc(
-		"cortex_bucket_store_blocks_loaded",
+		"cortex_parquet_bucket_store_blocks_loaded",
```
jesusvazquez, line 156: Renaming these metrics is just a nit i commented some time ago in some other PR.

PR #12152, opened 2025-07-21
POC, DO NOT PIC: OTLP to Appender

krajorama:
Adopt https://github.com/prometheus/prometheus/pull/16855 .

Remove copy of Prometheus OTLP translator code from Mimir as it is now storage independent in (mimir-)Prometheus.

Add new CombinedAppender implementation that uses the mimirpb.PreAllocTimeseries as target.


pkg/distributor/otel.go
```diff
@@ -516,51 +459,56 @@ type conversionOptions struct {
...
-	mimirTS := converter.ToTimeseries(ctx, md, settings, logger)
+	converter.appender.WithOptions(otlpappender.CombinedAppenderOptions{
+		EnableCreatedTimestampZeroIngestion: opts.enableCTZeroIngestion,
+	})
+	timeseries, metadata := converter.ToSeriesAndMetadata(ctx, md, settings, logger)
 
 	dropped := converter.DroppedTotal()
-	if len(mimirTS) == 0 && dropped > 0 {
-		return nil, dropped, converter.Err()
+	if timeseries == nil && dropped > 0 {
```
aknuds1, line 527: Is there any benefit to changing this check?

pkg/distributor/otel.go
```diff
@@ -516,51 +459,56 @@ type conversionOptions struct {
...
 type otlpMimirConverter struct {
-	converter *otlp.MimirConverter
+	appender  *otlpappender.CombinedAppender
+	converter *otlptranslator.PrometheusConverter
 	// err holds OTLP parse errors
 	err error
 }
 
-func newOTLPMimirConverter() *otlpMimirConverter {
+func newOTLPMimirConverter(appender *otlpappender.CombinedAppender) *otlpMimirConverter {
```
aknuds1, line 540: [Nit] Is there any benefit to take the `otlpappender.CombinedAppender` argument? Can't it just be instantiated in this function, instead of leaving it to the caller?

aknuds1 (commented):
Left some comments.

krajorama:
outdated, new PR: #12652

pkg/distributor/otel.go
krajorama: fixed in next PR

pkg/distributor/otel.go
krajorama: Maybe, I haven't spent a lot of time on it, not sure why it's so complicated currently.

PR #12157, opened 2025-07-22
allow mimir alertmanager fallback config from an existing Kubernetes Secret

### LINKED ISSUE #12153: Bug: slack_api_url_file not respected by mimir alertmanager
### What is the bug?

Hoi Grafana team!

Grafana Mimir's pod in Kubernetes throws the following in the logs:

> ts=2025-07-21T13:24:33.9987957Z caller=multitenant.go:725 level=warn component=MultiTenantAlertmanager msg="error applying config" err="unable to load fallback configuration for anonymous: no global Slack API URL set" user=anonymous

Even though I have the `global.slack_api_url_file` parameter set in my fallback config. Slack alerts are not working because of this, even though the individual receivers do have their slack api url files also set.

Please let me know if you need any further info and if this isn't intended to be supported, what I should do to hide secret data instead?

Thank you for any help you can provide 🙏 

### How to reproduce it?

Via the latest helm chart values, I use this fallback config:

```yaml
alertmanager:
  enabled: true

  # these are secrets with slackURLs
  extraVolumes:
    - name: alertmanager-api-urls
      secret:
        secretName: alertmanager-api-urls

  # these are secrets URLs
  extraVolumeMounts:
    - name: alertmanager-api-urls
      mountPath: /etc/secrets/alertmanager
      readOnly: true

  fallbackConfig: |
    global:
      slack_api_url_file: '/etc/secrets/alertmanager/slack_url'

# let me know if you need my whole fallbackconfig but I've added the most crucial part
```

the secret that has the url looks like this (but I've redacted the actual secret):

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-api-urls
  namespace: monitoring
type: Opaque
data:
  slack_url: redacted
```

It should respect the `global.slack_api_url_file` parameter as this is respected in Prometheus alertmanager (which I have running in the same cluster working fine). The reason it matters is that this parameter contains sensitive information. I expect this based on the [docs](https://grafana.com/docs/mimir/latest/references/architecture/components/alertmanager/#tenant-configurations) that say:
> Each tenant has an Alertmanager configuration that defines notifications receivers and alerting routes. The Mimir Alertmanager uses the same [configuration file](https://prometheus.io/docs/alerting/latest/configuration/#configuration-file) that the Prometheus Alertmanager uses.

kubernetes: `v1.30.13-eks-5d4a308`
Helm chart version: `5.8.0-weekly.347`

Let me know if you need anything else!

_No response_

jessebot:
#### What this PR does

Adds a `alertmanager.fallbackConfigExistingSecret` helm parameter so users can pass in their alertmanager config from an existing Kubernetes Secret (this is important as the alertmanager config can contain API tokens that should be secret and not checked directly into git repos).

Fixes #12153 kind of. It creates a better workaround for now.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## main / unreleased
 
+* [CHANGE] allow mimir alertmanager fallback config from an existing Kubernetes Secret #12157
```
tacole02:
```suggestion
* [CHANGE] Allow the Mimir Alertmanager to use a fallback configuration from an existing Kubernetes secret. #12157
```

zenador:

tacole02:
Hi @jessebot ! My approval is only for the docs. We still need a maintainer to review the code. I've updated the reviewers so that an engineer will see it. Thanks for your patience!

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -615,6 +615,10 @@ alertmanager:
     route:
         receiver: default-receiver
 
+  # -- use an existing Kubernetes secret for the fallback config for alertmanager.
+  # if set, ignores alertmanager.fallbackConfig
+  fallbackConfigExistingSecret: ""
```
dimitarvdimitrov, line 620: we already have `externalConfigSecretName`. Can we use the same pattern -  `externalFallbackConfigSecretName`?

dimitarvdimitrov: is there a requirement on the key of the secret? I think the key is used to mount a file and then the file name is referenced in the AM config. Can you clarify that the name of the key need to be `alertmanager_fallback_config.yaml` 

like it is in https://github.com/grafana/mimir/blob/e32b5e2531d40677f9793f8c72ed18aa79fcf50b/operations/helm/charts/mimir-distributed/templates/alertmanager/alertmanager-config.yaml#L13

PR #8077, opened 2024-05-07
Split a per-partition WriteRequest into multiple Kafka records if bigger than max allowed size

pracucci:
#### What this PR does

_To be written..._

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12166, opened 2025-07-22
ingest-storage: Split very large RW2.0 requests into multiple Kafka records

alexweav:
#### What this PR does

This is the record format V2 equivalent of https://github.com/grafana/mimir/pull/8077. Kafka has a limit in payload size, here we break down extremely large requests into multiple records to work around it.

RW2 carries some complexity due to the symbols table. When splitting a request, we have to split the symbols as well. It's not reasonable to use all the symbols on each request, because symbols still dominate the overall request size.

This PR uses a similar bin-packing algorithm to RW1, but it also re-symbolizes each Timeseries as we move along, using a separate symbols table per sub-request. That means each Timeseries can re-use symbols from other timeseries in the same request as they're added, up until the point where the request is close to the max.

To determine when to split a request, we calculate an upper bound of the growth that the next timeseries will cause without actually processing any symbols yet. If it's possible that this timeseries can push things over the limit, we split it (preferring splitting early vs breaching the limit).

Addresses this comment: https://github.com/grafana/mimir/pull/12060#discussion_r2221466520

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/custom.go
+++ b/pkg/mimirpb/custom.go
@@ -228,6 +228,33 @@ func (m *WriteRequest) TimeseriesSize() int {
 	return n
 }
 
+// TimeseriesSize is like Size() but returns only the marshalled size of TimeseriesRW2 field.
+func (m *WriteRequest) TimeseriesRW2Size() int {
+	var n, l int
+
+	for _, e := range m.TimeseriesRW2 {
+		l = e.Size()
+		n += 1 + l + sovMimir(uint64(l))
+	}
+
+	return n
+}
+
+// TimeseriesSize is like Size() but returns only the marshalled size of SymbolsRW2 field.
+func (m *WriteRequest) SymbolsRW2Size() int {
+	if len(m.SymbolsRW2) == 0 {
+		return 0
+	}
+
+	var n, l int
+	for _, s := range m.SymbolsRW2 {
+		l = len(s)
+		n += 1 + l + sovMimir(uint64(l))
+	}
+
+	return n
+}
+
 func (h Histogram) IsFloatHistogram() bool {
 	_, ok := h.GetCount().(*Histogram_CountFloat)
 	return ok
--- a/pkg/mimirpb/custom.go
+++ b/pkg/mimirpb/custom.go
@@ -228,7 +228,7 @@ func (m *WriteRequest) TimeseriesSize() int {
 	return n
 }
 
-// TimeseriesSize is like Size() but returns only the marshalled size of TimeseriesRW2 field.
+// TimeseriesRW2Size is like Size() but returns only the marshalled size of TimeseriesRW2 field.
 func (m *WriteRequest) TimeseriesRW2Size() int {
 	var n, l int
 
@@ -240,12 +240,8 @@ func (m *WriteRequest) TimeseriesRW2Size() int {
 	return n
 }
 
-// TimeseriesSize is like Size() but returns only the marshalled size of SymbolsRW2 field.
+// SymbolsRW2Size is like Size() but returns only the marshalled size of SymbolsRW2 field.
 func (m *WriteRequest) SymbolsRW2Size() int {
-	if len(m.SymbolsRW2) == 0 {
-		return 0
-	}
-
 	var n, l int
 	for _, s := range m.SymbolsRW2 {
 		l = len(s)
--- a/pkg/mimirpb/prealloc_rw2.go
+++ b/pkg/mimirpb/prealloc_rw2.go
@@ -50,9 +50,10 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	// Represent metadata as extra, empty timeseries rather than attaching it to existing series.
 	// We never replicate metadata if there are multiple series with the same name, and it removes the requirement to match up the metadata to the right series.
 	for _, meta := range rw1.Metadata {
+		labelsRefs := []uint32{symbols.Symbolize("__name__"), symbols.Symbolize(meta.MetricFamilyName)}
 		rw2meta := FromMetricMetadataToMetadataRW2(meta, symbols)
 		rw2Timeseries = append(rw2Timeseries, TimeSeriesRW2{
-			LabelsRefs: []uint32{symbols.Symbolize("__name__"), symbols.Symbolize(meta.MetricFamilyName)},
+			LabelsRefs: labelsRefs,
 			Metadata:   rw2meta,
 		})
 	}
--- a/pkg/mimirpb/split.go
+++ b/pkg/mimirpb/split.go
@@ -51,7 +51,7 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int) 
 //
 // The request will split the RW2 symbols among the various sub-requests. The original symbols table will no longer be valid for the individual timeseries.
 // Timeseries are re-symbolized in place, so this function mutates the input.
-func SplitWriteRequestByMaxMarshalSizeRW2(req *WriteRequest, reqSize, maxSize int) []*WriteRequest {
+func SplitWriteRequestByMaxMarshalSizeRW2(req *WriteRequest, reqSize, maxSize int, offset uint32, commonSymbols []string) []*WriteRequest {
 	if reqSize <= maxSize {
 		return []*WriteRequest{req}
 	}
@@ -75,7 +75,7 @@ func SplitWriteRequestByMaxMarshalSizeRW2(req *WriteRequest, reqSize, maxSize in
 
 	// Split timeseries into partial write requests, and resymbolize each batch.
 	nextReqSymbols := symbolsTableFromPool()
-	// TODO: Common Symbols...
+	nextReqSymbols.ConfigureCommonSymbols(offset, commonSymbols)
 	defer reuseSymbolsTable(nextReqSymbols)
 	nextReq, nextReqSize := newPartialReq()
 	nextReqTimeseriesStart := 0
--- a/pkg/mimirpb/split.go
+++ b/pkg/mimirpb/split.go
@@ -19,10 +19,17 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int) 
 		return []*WriteRequest{req}
 	}
 
+	if len(req.TimeseriesRW2) > 0 || len(req.SymbolsRW2) > 0 {
+		return splitWriteRequestByMaxMarshalSizeRW2(req, reqSize, maxSize)
+	}
+	return splitWriteRequestByMaxMarshalSizeRW1(req, reqSize, maxSize)
+}
+
+func splitWriteRequestByMaxMarshalSizeRW1(req *WriteRequest, reqSize, maxSize int) []*WriteRequest {
 	partialReqsWithTimeseries := splitTimeseriesByMaxMarshalSize(req, reqSize, maxSize)
 	partialReqsWithMetadata := splitMetadataByMaxMarshalSize(req, reqSize, maxSize)
 
-	// Most of the time, a write request only have either Timeseries or Metadata.
+	// Most of the time, a write request only have either Timeseries OR TimeseriesRW2 OR Metadata.
 	if len(partialReqsWithMetadata) == 0 {
 		return partialReqsWithTimeseries
 	}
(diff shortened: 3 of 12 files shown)


pkg/mimirpb/split.go
```diff
@@ -37,6 +44,80 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+			SkipLabelValidation: req.SkipLabelValidation,
+		}
+
+		return r, r.Size()
+	}
+
+	symbolSize := req.SymbolsRW2Size()
+	timeSeriesSize := req.TimeseriesRW2Size()
+
+	// Naive implementation: send all symbols on all sub-requests.
```
alexweav: I _strongly_ suspect we can't actually do this. Symbols is an exceedingly high proportion of the request, we need to shard it or we can fail to split effectively. Also the duplication cuts into our Kafka throughput.

alexweav:
The remaining test failures at this point are largely ephemeral. E.g., the request is correct but the symbols come back in a different order - they resolve to the same request but are not bit-for-bit equal.

Longer term, I'd like to make all RW1->2->1 conversions completely stable/order-preserving, it can help with comparing requests massively.

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -50,9 +50,10 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	// Represent metadata as extra, empty timeseries rather than attaching it to existing series.
 	// We never replicate metadata if there are multiple series with the same name, and it removes the requirement to match up the metadata to the right series.
 	for _, meta := range rw1.Metadata {
+		labelsRefs := []uint32{symbols.Symbolize("__name__"), symbols.Symbolize(meta.MetricFamilyName)}
```
alexweav, line 53: Just changing the order that we symbolize fields, for consistency. It makes comparing RW2 requests easier in tests. It cannot change the actual meaning of a request.

pkg/storage/ingest/writer_test.go
```diff
@@ -920,10 +933,92 @@ func TestMarshalWriteRequestToRecords(t *testing.T) {
...
+			UnmarshalFromRW2: true,
+			RW2SymbolOffset:  V2RecordSymbolOffset,
+			RW2CommonSymbols: V2CommonSymbols,
+		}
+		require.NoError(t, actual.Unmarshal(records[0].Value))
+
+		actual.ClearTimeseriesUnmarshalData()
+		assert.Equal(t, mimirpb.RULE, actual.Source)
+		assert.Equal(t, true, actual.SkipLabelValidation)
+		expMetadata := []*mimirpb.MetricMetadata{
```
alexweav, line 952: The test comparison throughout are really verbose, because of the order not being deterministic, and also because of our extra timeseries that only carry metadata.

I have in the tasklist to handle these, after which it makes asserting on RW2 requests much easier, as we can just compare the whole thing by marshalling back and forth.

pkg/storage/ingest/writer_test.go
```diff
@@ -963,10 +1058,127 @@ func TestMarshalWriteRequestToRecords(t *testing.T) {
...
+
+			actual.ClearTimeseriesUnmarshalData()
+			partials = append(partials, &actual.WriteRequest)
+		}
+
+		assert.Equal(t, []*mimirpb.WriteRequest{
+			{
+				Source:              mimirpb.RULE,
+				SkipLabelValidation: true,
+				Timeseries: []mimirpb.PreallocTimeseries{
```
alexweav, line 1092: This test is IMO an interesting demonstration of the algorithm.

The first 3 requests can all bin-pack into 100 bytes, because they don't have any long strings and "__name__" is a common symbol, so it's excluded. The metadata series contain the longer help text plus series names, which pushes each one into its own request.

pkg/mimirpb/custom.go
```diff
@@ -228,6 +228,33 @@ func (m *WriteRequest) TimeseriesSize() int {
 	return n
 }
 
+// TimeseriesSize is like Size() but returns only the marshalled size of TimeseriesRW2 field.
```
pracucci: [nit]
```suggestion
// TimeseriesRW2Size is like Size() but returns only the marshalled size of TimeseriesRW2 field.
```

pkg/mimirpb/custom.go
```diff
@@ -228,6 +228,33 @@ func (m *WriteRequest) TimeseriesSize() int {
...
+
+	for _, e := range m.TimeseriesRW2 {
+		l = e.Size()
+		n += 1 + l + sovMimir(uint64(l))
+	}
+
+	return n
+}
+
+// TimeseriesSize is like Size() but returns only the marshalled size of SymbolsRW2 field.
```
pracucci: [nit]

```suggestion
// SymbolsRW2Size is like Size() but returns only the marshalled size of SymbolsRW2 field.
```

pkg/mimirpb/custom.go
```diff
@@ -228,6 +228,33 @@ func (m *WriteRequest) TimeseriesSize() int {
...
+func (m *WriteRequest) SymbolsRW2Size() int {
+	if len(m.SymbolsRW2) == 0 {
+		return 0
+	}
```
pracucci: [nit] Is there any real benefit having this `if` branch?

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
 	return merged
 }
 
+// SplitWriteRequestByMaxMarshalSizeRW2 splits RW2 write-requests into multiple ones, where each partial WriteRequest marshalled size
+// is at most maxSize. The input reqSize must be the value returned by WriteRequest.Size(); it's passed because the called
```
pracucci: [nit] The typo is in the original function above too. Do you mind fixing it there too?

```suggestion
// is at most maxSize. The input reqSize must be the value returned by WriteRequest.Size(); it's passed because the caller
```

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+// The returned requests may still retain references to fields in the original WriteRequest, i.e. they are tied to its lifecycle.
+//
+// The request will split the RW2 symbols among the various sub-requests. The original symbols table will no longer be valid for the individual timeseries.
+// Timeseries are re-symbolized in place, so this function mutates the input.
+func SplitWriteRequestByMaxMarshalSizeRW2(req *WriteRequest, reqSize, maxSize int, offset uint32, commonSymbols []string) []*WriteRequest {
+	if reqSize <= maxSize {
+		return []*WriteRequest{req}
+	}
+	if len(req.TimeseriesRW2) == 0 {
+		return []*WriteRequest{}
```
pracucci, line 59: This looks dangerous on edge cases. Isn't safer if we just return the input request?

```suggestion
		return []*WriteRequest{req}
```

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+	nextReqSymbols := symbolsTableFromPool()
+	nextReqSymbols.ConfigureCommonSymbols(offset, commonSymbols)
+	defer reuseSymbolsTable(nextReqSymbols)
+	nextReq, nextReqSize := newPartialReq()
+	nextReqTimeseriesStart := 0
+	nextReqTimeseriesLength := 0
+	nextReqSymbolsSize := 0
+
+	for i := 0; i < len(req.TimeseriesRW2); i++ {
+		// Both are upper bounds. In particular symbolsSize does have knowledge of whether symbols can be re-used.
```
pracucci: [nit] Am I right understanding that `maxRW2SeriesSizeAfterResymbolization()` does **not** have knowledge whether symbols can be reused?

```suggestion
		// Both are upper bounds. In particular symbolsSize doesn't have knowledge of whether symbols can be re-used.
```

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+			nextReqTimeseriesLength = 0
+			nextReqSymbolsSize = 0
+			nextReqSymbols.Reset()
+			nextReqSymbols.ConfigureCommonSymbols(offset, commonSymbols)
+		}
+
+		// Add the current series to next partial request.
+		deltaSize := resymbolizeTimeSeriesRW2(&req.TimeseriesRW2[i], req.SymbolsRW2, nextReqSymbols)
+		nextReqSize += deltaSize + 1 + sovMimir(uint64(seriesSize)) // Math copied from Size().
+		nextReqSize += nextReqSymbols.SymbolsSizeProto() - nextReqSymbolsSize
```
pracucci: I don't understand. Isn't `nextReqSymbolsSize` always 0?

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+			nextReqTimeseriesStart = i
+			nextReqTimeseriesLength = 0
+			nextReqSymbolsSize = 0
+			nextReqSymbols.Reset()
+			nextReqSymbols.ConfigureCommonSymbols(offset, commonSymbols)
+		}
+
+		// Add the current series to next partial request.
+		deltaSize := resymbolizeTimeSeriesRW2(&req.TimeseriesRW2[i], req.SymbolsRW2, nextReqSymbols)
+		nextReqSize += deltaSize + 1 + sovMimir(uint64(seriesSize)) // Math copied from Size().
```
pracucci: Can you explain me why you add the `deltaSize` that could also be negative?

pracucci: Is it missing an initialization to `nextReqSize`? I struggle to understand this math 🤔

pkg/mimirpb/split.go
```diff
@@ -163,3 +248,86 @@ func splitMetadataByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int) []*W
...
+	symbolsSize += resolvedSymbolSize(ts.Metadata.UnitRef, symbols, symbolOffset)
+	return
+}
+
+func resolvedSymbolSize(ref uint32, symbols []string, offset uint32) int {
+	if ref == 0 {
+		return 1 + sovMimir(uint64(0))
+	}
+	if ref > 0 && ref < offset {
+		return 0
```
pracucci, line 293: [nit] Can you add a comment to explain why? ... because it's part of the common symbols table and so won't be part of the per-request symbols.

pkg/mimirpb/split.go
```diff
@@ -163,3 +248,86 @@ func splitMetadataByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int) []*W
...
+	seriesSize += (symbolSizeUpperBound - sovMimir(uint64(ts.Metadata.HelpRef)))
+	symbolsSize += resolvedSymbolSize(ts.Metadata.HelpRef, symbols, symbolOffset)
+	seriesSize += (symbolSizeUpperBound - sovMimir(uint64(ts.Metadata.UnitRef)))
+	symbolsSize += resolvedSymbolSize(ts.Metadata.UnitRef, symbols, symbolOffset)
+	return
+}
+
+func resolvedSymbolSize(ref uint32, symbols []string, offset uint32) int {
+	if ref == 0 {
+		return 1 + sovMimir(uint64(0))
```
pracucci, line 289: [nit] Can you add a comment to explain why? ... to my understanding ref == 0 is the empty string, but due to the specs it must be encoded in the symbols table anyway. Am I right  or wrong?

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,91 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+// For this reason, this function is a best-effort: if a single Timeseries or Metadata marshalled size is bigger
+// than maxSize, then the returned partial WriteRequest marshalled size will be bigger than maxSize too.
+//
+// The returned partial WriteRequests are NOT a deep copy of the input one; they contain references to slices
+// and data from the original WriteRequest.
+// The returned requests may still retain references to fields in the original WriteRequest, i.e. they are tied to its lifecycle.
+//
+// The request will split the RW2 symbols among the various sub-requests. The original symbols table will no longer be valid for the individual timeseries.
+// Timeseries are re-symbolized in place, so this function mutates the input.
+func SplitWriteRequestByMaxMarshalSizeRW2(req *WriteRequest, reqSize, maxSize int, offset uint32, commonSymbols []string) []*WriteRequest {
```
pracucci, line 54: The most risky thing done by this is mutating the input `req` in place. I do understand this was done as an optimization, but could be quite risky.  The `SplitWriteRequestByMaxMarshalSize()` did kept some references, but didn't mutate the input request (and it was done on purpose to keep it safer). 

What if we pay some extra performance cost, and modify `SplitWriteRequestByMaxMarshalSizeRW2()` to not mutate the input request? Have you considered it?

pkg/mimirpb/split.go
alexweav: You're right, it's vestigial. Originally I used nextReqSymbolsSize to track the total symbols size of the request we are building, but later on I pivoted away from that and added `SymbolsSizeProto()` on the table, but didn't fully update this.

This is a bug because we're adding the full size of the symbols table to `nextReqSize` each pass, meaning symbols get double-counted.

It passes the tests because this results in splitting the next request a bit too early, which is still technically safe. We break it into too many requests, so we still meet the limit when we expect.

pkg/mimirpb/split.go
alexweav: Yes, that's exactly right.

pkg/mimirpb/split.go
alexweav: The function needs the ability to return multiple requests, here we do just return the original request but wrapped in a slice. It's different from a timeseries slice that may or may not get pooled. These slices are never pooled, even in the RW1 implementation.

The RW1 implementation does the same: https://github.com/grafana/mimir/blob/78a322f09c4ff0a473fef1eb8582c981e3519ff1/pkg/mimirpb/split.go#L18-L20

Also I think in practice this can never be hit outside of tests, in both paths. `marshalWriteRequestToRecords` has an earlier guard: https://github.com/grafana/mimir/blob/78a322f09c4ff0a473fef1eb8582c981e3519ff1/pkg/storage/ingest/writer.go#L248

pkg/mimirpb/split.go
alexweav: Yes. I got annoyed with this when writing tests and _nearly_ did it here. It's why we have to keep reconstructing the requests in every test case.

The idea with working in-place was to reduce allocations, but tbh think the rarity of this entire path beats out that argument, and we can afford the extra allocations for safety. (Plus, we can always optimize things over time.)

pkg/mimirpb/split.go
alexweav: It's hard to understand because it's wrong :sweat_smile: I refactored this right at the end and broke it.

Ironically, it got hidden by symbols size being over-estimated. So, we were over-calculating the size of symbols in `nextReq` and under-calculating the size of timeSeries, which countered each other a bit.

alexweav: I've pushed a fix, and ideally, this logic is written in a bit more readable way. Let me know what you think!

This did change the results of one test, and made it pack requests in less aggressively. It still technically passed before, because the total size of each sub-request was still less than the limit, but now it's by a larger margin.

pkg/storage/ingest/writer_test.go
alexweav: Note: The latest commits adjusted the bin-packing slightly, now we are less aggressive and only the first 2 fit together. Similar idea holds, though

pkg/mimirpb/split.go
pracucci: I don't understand this answer. My comment is about:

```
	if len(req.TimeseriesRW2) == 0 {
		return []*WriteRequest{}
        }
```

And I think we should do instead:


```
	if len(req.TimeseriesRW2) == 0 {
		return []*WriteRequest{req}
        }
```

pkg/mimirpb/split.go
```diff
@@ -37,6 +37,97 @@ func SplitWriteRequestByMaxMarshalSize(req *WriteRequest, reqSize, maxSize int)
...
+	if reqSize <= maxSize {
+		return []*WriteRequest{req}
+	}
+	if len(req.TimeseriesRW2) == 0 {
+		return []*WriteRequest{}
+	}
```
pracucci, line 60: I know that in RW2 we expect metadata as part of timeseries, but should we also check if the input request has `req.Metadata`? If yes, what should we do? The split requests will not have it.

pkg/mimirpb/split_test.go
```diff
@@ -38,71 +38,302 @@ func TestSplitWriteRequestByMaxMarshalSize(t *testing.T) {
...
+			},
+		}, partials)
+
+		for _, partial := range partials {
+			assert.LessOrEqual(t, partial.Size(), limit)
+		}
+	})
+
+	t.Run("should split the input WriteRequest into multiple requests, honoring the size limit - RW2", func(t *testing.T) {
+		// 150 is small enough to force each WriteRequest into its own sub-request
```
pracucci: [nit] Did you mean timeseries here?

```suggestion
		// 150 is small enough to force each timeseries into its own sub-request
```

pkg/mimirpb/split_test.go
```diff
@@ -182,46 +440,88 @@ func TestSplitWriteRequestByMaxMarshalSize_WriteRequestHasChanged(t *testing.T)
...
+					unmarshalFromRW2: true,
+				}
+				if err := unmarshalledReq.Unmarshal(marshalledReq); err != nil {
 					b.Fatal(err)
-				} else if len(data) >= maxSize {
-					b.Fatalf("the marshalled partial request (%d bytes) is larger than max size (%d bytes)", len(data), maxSize)
+				}
+
+				// Split the request.
+				partialReqs := SplitWriteRequestByMaxMarshalSizeRW2(req, req.Size(), maxSize, 0, nil)
```
pracucci: This looks a mistake we did originally in the benchmark. Can you fix the RW1 too, please?


```suggestion
				partialReqs := SplitWriteRequestByMaxMarshalSizeRW2(unmarshalledReq, unmarshalledReq.Size(), maxSize, 0, nil)
```

pkg/mimirpb/split.go
alexweav: Sorry, I got confused by the github UI. It was not clear if the comment was on line 56 or 59:
2025-07-29-112142_575x146_scrot

I'll do that - it feels safer on principle. Although, it's a no-op - you can't ever have TimeseriesRW2 and Metadata. If a request has both it's rejected ([validation](https://github.com/grafana/mimir/blob/7b9c5f2165a7c5a240c1c86171b3d6e5f2596c5e/pkg/mimirpb/mimir.pb.go#L7412-L7413), [error](https://github.com/grafana/mimir/blob/7b9c5f2165a7c5a240c1c86171b3d6e5f2596c5e/pkg/mimirpb/compat_rw2.go#L14)). So, an RW2 message with no timeseries does not carry anything to store.

pkg/mimirpb/split.go
alexweav: This is strictly invalid, any request mixing RW1 metadata with RW2 timeseries is broken and will straight up fail to unmarshal: https://github.com/grafana/mimir/blob/7b9c5f2165a7c5a240c1c86171b3d6e5f2596c5e/pkg/mimirpb/mimir.pb.go#L7412-L7413

alexweav:
> that's something we can address in a follow up PR

Agreed, it'll be easier to review on its own. Added to the task list.

PR #12169, opened 2025-07-23
MQE: Count samples saved on the CSE

Konstantinov-Innokentii:
#### What this PR does

When detecting a duplicate expression, set the sampleCountMultiplicator on the Vector/Matrix selector in the duplicated path. It's needed to preserve "original" number of samples processed, as there were no CSE applied. The multiplication is set on a planning stage - [code](https://github.com/grafana/mimir/pull/12169/files#diff-50436966a2717813bee0d99c1ed79b7d695ca83a9ca4e88c6bbeff8d61e0adf3R272)
Note that during this work, I found an edge case, when MQE returns 0 samples processed when CSE is applied - plan is to tackle it separately.  Here is a skipped [test](https://github.com/grafana/mimir/pull/12169/files#diff-df3817badf68588818b73b443096f0edd778d0019a2bc135e6a7f6eeb218946cR3602) covering it.
#### Checklist

- [x] Tests updated.
- [N/A] Documentation added.


pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -42,6 +42,20 @@ type Selector struct {
 	series    *seriesList
 
 	seriesIdx int
+
+	// Sample count factor for stats counting when applying CSE.
```
56quarters, line 46:
```suggestion
	// Sample count factor for stats counting at each step when applying common sub-expression elimination.
```

pkg/streamingpromql/engine_test.go
```diff
@@ -3473,14 +3461,232 @@ func TestQueryStats(t *testing.T) {
...
+			if testCase.skip != "" {
+				t.Skip(testCase.skip)
 			}
+
+			prometheusSamplesStats := runQueryAndGetSamplesStats(t, prometheusEngine, testCase.expr, testCase.isInstantQuery)
+
+			if testCase.skipCompareWithPrometheus == "" {
+				require.Equal(t, testCase.expectedTotalSamples, prometheusSamplesStats.TotalSamples, "invalid test case: expected total samples does not match value from Prometheus' engine")
+				require.Equal(t, testCase.expectedTotalSamplesPerStep, prometheusSamplesStats.TotalSamplesPerStep, "invalid test case: expected per stepsamples does not match value from Prometheus' engine")
+			}
```
56quarters, line 3685: Should we log the reason for skipping here?

```go
} else {
    t.Log(testCase.skipCompareWithPrometheus)
}
```

56quarters (approved):
LGTM

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go
```diff
@@ -375,3 +377,23 @@ func (p *path) String() string {
 	b.WriteRune(']')
 	return b.String()
 }
+
+// setSampleCountFactor sets SampleCountFactor on the VectorSelector and MatrixSelector nodes.
+// It sets it only if multipilcator was not set before to avoid overwriting it for nested duplicate expressions.
```
lamida, line 382: Nits. I think `multiplier` is a more common word for this.
```suggestion
// It sets it only if multiplier was not set before to avoid overwriting it for nested duplicate expressions.
```

pkg/streamingpromql/planning/core/matrix_selector.go
```diff
@@ -79,7 +79,9 @@ func (m *MatrixSelector) OperatorFactory(_ []types.Operator, timeRange types.Que
 		ExpressionPosition:       m.ExpressionPosition.ToPrometheusType(),
 		MemoryConsumptionTracker: params.MemoryConsumptionTracker,
 	}
-
+	if m.SampleCountFactor != 0 {
+		selector.SetSampleCountFactor(m.SampleCountFactor)
+	}
```
lamida, line 84: What if we just push the non zero check `!= 0` inside `SetSampleCountFactor` method itself? With that we can remove this if branch and also the one in vector_selector.go.

pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -42,6 +42,20 @@ type Selector struct {
...
+	sampleCountFactor uint32
+}
+
+func (s *Selector) SetSampleCountFactor(factor uint32) {
+	s.sampleCountFactor = factor
+}
```
lamida, line 52: As I mentioned in another comment, since we have setter method here, just add logic to set only for non-zero factor. Hence we can remove non-zero check in the caller.
```suggestion
func (s *Selector) SetSampleCountFactor(factor uint32) {
	if factor != 0 {
		s.sampleCountFactor = factor
	}
}
```

pkg/streamingpromql/engine_test.go
```diff
@@ -3473,14 +3461,232 @@ func TestQueryStats(t *testing.T) {
...
+
+		defer q.Close()
+
+		res := q.Exec(context.Background())
+		require.NoError(t, res.Err)
+
+		return q.Stats().Samples
+	}
+
+	testCases := map[string]struct {
```
lamida, line 3514: All the testCases below are for instant query. Should we have testCase for range query too? So that we can see the case where totalSamples is not equal with totalSamplesPerStep.

pkg/streamingpromql/operators/selectors/range_vector_selector.go
```diff
@@ -103,7 +103,7 @@ func (m *RangeVectorSelector) NextStepSamples() (*types.RangeVectorStepData, err
 	m.stepData.RangeStart = rangeStart
 	m.stepData.RangeEnd = rangeEnd
 
-	m.Stats.IncrementSamplesAtTimestamp(m.stepData.StepT, int64(m.stepData.Floats.Count())+m.stepData.Histograms.EquivalentFloatSampleCount())
+	m.Stats.IncrementSamplesAtTimestamp(m.stepData.StepT, (int64(m.stepData.Floats.Count())+m.stepData.Histograms.EquivalentFloatSampleCount())*int64(m.Selector.GetSampleCountFactor()))
```
lamida, line 106: With some castings, it takes sometime to see what is being computed. What if we extract some of the operation to local variables to make it clearer?
```suggestion
	floatCount := int64(m.stepData.Floats.Count())
	histogramCount := m.stepData.Histograms.EquivalentFloatSampleCount()
	sampleFactor := int64(m.Selector.GetSampleCountFactor())
	totalSamples := (floatCount + histogramCount) * sampleFactor
	
	m.Stats.IncrementSamplesAtTimestamp(m.stepData.StepT, totalSamples)```

File: pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/optimization_pass.go [hunk 8]
```diff
@@ -375,3 +377,23 @@ func (p *path) String() string {
...
+		return
+	case *core.MatrixSelector:
+		if selector.SampleCountFactor == 0 {
+			selector.SampleCountFactor = factor
+		}
+		return
+	}
+	for _, child := range node.Children() {
+		setSampleCountFactor(child, factor)
+	}
```
lamida, line 398: Instead of having two empty `return`s, what if we just add default clause to catch all?
```suggestion
	switch selector := node.(type) {
	case *core.VectorSelector:
		if selector.SampleCountFactor == 0 {
			selector.SampleCountFactor = factor
		}
	case *core.MatrixSelector:
		if selector.SampleCountFactor == 0 {
			selector.SampleCountFactor = factor
		}
	default:
		for _, child := range node.Children() {
			setSampleCountFactor(child, factor)
		}
	}
```

lamida (commented):
Overall almost looks good to me. But I left some suggestions anyway.

PR #12175, opened 2025-07-23
chore: update to latest dskit for DNS changes

### LINKED ISSUE #12160: Use more reliable cache service discovery
After the resolver introduced in https://github.com/grafana/dskit/pull/660 becomes the default in dskit, switch to it. This is part of Mimir 3.0 since the new resolver doesn't support search domains. However, this shouldn't affect users in most cases since we don't use them in the Jsonnet or Helm chart.

56quarters:
#### What this PR does

Update to the latest commit of dskit to pull in improvements to DNS-based service discovery for caches.

Fixes #12160

Pulls in https://github.com/grafana/dskit/pull/718

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Grafana Mimir
 
 * [CHANGE] Remove support for Redis as a cache backend. #12163
+* [CHANGE] Memcached: Remove experimental `-<prefix>.memcached.addresses-provider` flag to use alternate DNS service discovery backends. The more reliable backend introduced in #10895 is now the default. As a result of this change, DNS-based cache service discovery no longer supports search domains. #12175
```
dimitarvdimitrov: nit
```suggestion
* [CHANGE] Memcached: Remove experimental `-<prefix>.memcached.addresses-provider` flag to use alternate DNS service discovery backends. The more reliable backend introduced in 2.16.0 (#10895) is now the default. As a result of this change, DNS-based cache service discovery no longer supports search domains. #12175
```

dimitarvdimitrov (approved):

PR #12178, opened 2025-07-23
Test potential fix for HA tracker failovers on r349

zenador:
Not to be merged, just running tests in CI, using it as a base for vendoring and getting feedback.

Based on https://github.com/grafana/mimir/pull/12082, then updated to simplify by removing logging we no longer need, and also make the improvement considered there by making the new flag a duration config instead of a bool.


docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3351,6 +3351,14 @@ The `limits` block configures default and per-tenant limits imposed by component
 # CLI flag: -distributor.ha-tracker.failover-timeout
 [ha_tracker_failover_timeout: <duration> | default = 30s]
 
+# (advanced) Additional timeout to use for failover that uses earliest sample
```
```suggestion
# (advanced) Additional timeout to use for failover that uses the earliest sample
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3351,6 +3351,14 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# time instead of the current time. Defaults to 0 which is disabled. This is
```
```suggestion
# time instead of the current time. Defaults to 0, which is disabled. This is
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3351,6 +3351,14 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# useful to prevent samples being too close together during failover when write
```
```suggestion
# useful to prevent samples from being too close together during failover when write
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -3351,6 +3351,14 @@ The `limits` block configures default and per-tenant limits imposed by component
...
+# requests are delayed such that the sample time is earlier than the current
```
```suggestion
# requests are delayed so that the sample time is earlier than the current
```

tacole02 (approved):

zenador:
@tacole02 Thank you for the review, applied suggestions! For future reference, it's easier if you update the help text in the .go files directly, e.g. for this one it's [pkg/util/validation/limits.go](https://github.com/grafana/mimir/pull/12178/files#diff-1e0a07c0c1ea87ab5a582d79c587a73a9b8e0680d89c61d5c94be106823a4a4e), as the other files with the same texts are auto-generated based on the go code, so any changes there are more likely to be overwritten inadvertently.

tacole02:
That's really good to know! Thanks, @zenador !

PR #12184, opened 2025-07-24
Optimize activeseries.Matchers.Matches()

colega:
#### What this PR does

Optimizes the performance of `activeseries.Matchers.Matches()` and adds real-life benchmarks.

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/ingester/activeseries/model
cpu: Apple M3 Pro
                                          │  main.mem   │            optimized.mem            │
MatchesSeriesRealTrackers/match_series-12   80.20m ± 2%   22.68m ± 3%  -71.73% (p=0.000 n=10)

                                          │   main.mem   │            optimized.mem             │
                                          │     B/op     │     B/op      vs base                │
MatchesSeriesRealTrackers/match_series-12   18.47Ki ± 0%   16.39Ki ± 0%  -11.25% (p=0.000 n=10)

                                          │  main.mem   │          optimized.mem          │
                                          │  allocs/op  │  allocs/op   vs base            │
MatchesSeriesRealTrackers/match_series-12   2.000k ± 0%   2.000k ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

When creating the Matchers, we'll now split them in three groups:
- First, separate trackers that match on a single label
- Second, separate trackers that have an exact value matcher: either it's an equality or it's a set-matcher (`foo|bar`).
- The rest are called unoptimized matchers.

If we have matchers in the first two groups, we iterate the labels of the series and check for each label whether there's something matching: this way we avoid the lookup of the label value for each one of those matchers.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/activeseries/model/matchers.go
+++ b/pkg/ingester/activeseries/model/matchers.go
@@ -65,10 +65,12 @@ func NewMatchers(matchersConfig CustomTrackersConfig) *Matchers {
 
 	// Get label names and sort them by popularity (desc).
 	// We select labels that have the most different values matched first to make the decision tree higher and lower.
-	labelNames := slices.Collect(maps.Keys(valuesMatchedPerLabelName))
-	slices.SortFunc(labelNames, func(a, b string) int {
-		return cmp.Compare(len(valuesMatchedPerLabelName[b]), len(valuesMatchedPerLabelName[a]))
-	})
+	labelNames := slices.SortedFunc(
+		maps.Keys(valuesMatchedPerLabelName),
+		func(a, b string) int {
+			return cmp.Compare(len(valuesMatchedPerLabelName[b]), len(valuesMatchedPerLabelName[a]))
+		},
+	)
 
 	// valueMatchers is a map of label names to a map of values to indexedMatchers.
 	valueMatchers := make(map[string]map[string][]indexedMatchers, len(labelNames))
@@ -88,8 +90,7 @@ func NewMatchers(matchersConfig CustomTrackersConfig) *Matchers {
 				if m.Type == labels.MatchEqual && m.Value != "" {
 					// This is not a set matcher, but a simple equality matcher.
 					// We'll treat it as a set matcher with a single value.
-					stack := [1]string{m.Value}
-					sm = stack[:]
+					sm = []string{m.Value}
 				} else if sm = m.SetMatches(); len(sm) == 0 {
 					continue
 				}


Makefile
```diff
@@ -411,13 +411,6 @@ lint: check-makefiles
 		github.com/prometheus/client_golang/prometheus.{NewCounter,NewCounterVec,NewCounterFunc,NewGauge,NewGaugeVec,NewGaugeFunc,NewSummary,NewSummaryVec,NewHistogram,NewHistogramVec}=github.com/prometheus/client_golang/prometheus/promauto.With" \
 		./pkg/...
 
-	# Use the faster slices.Sort where we can.
-	# Note that we don't automatically suggest replacing sort.Float64s() with slices.Sort() as the documentation for slices.Sort()
-	# at the time of writing warns that slices.Sort() may not correctly handle NaN values.
-	faillint -paths \
-		"sort.{Strings,Ints}=slices.Sort" \
-		./pkg/... ./cmd/... ./tools/... ./integration/...
-
```
colega, line 424: `sort.{Strings,Ints}` actually calls `slices.Sort` as of go1.22, so changing the first one for the second one is a noop in terms of generated code.

So this faillint is: 
- spending useless time
- making the code less readable (IMO `sort.Strings()` is more readable than `slices.Sort()` call.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. You can enable the optimizer per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076 #12080
 * [ENHANCEMENT] Query-frontend: Standardise non-regex patterns in query blocking upon loading of config. #12102
 * [ENHANCEMENT] Ruler: Propagate GCS object mutation rate limit for rule group uploads. #12086
+* [ENHANCEMENT] Ingester: Improved the performance of active series custom trackers matchers. #12184
```
```suggestion
* [ENHANCEMENT] Ingester: Improve the performance of active series custom trackers matchers. #12184
```

zenador:

colega:
Merged `main` and fixed the CHANGELOG.md

pkg/ingester/activeseries/model/matchers.go
```diff
@@ -3,28 +3,147 @@
...
+			}
+		}
+		if !optimized {
+			unoptimized = append(unoptimized, indexedMatchers{labelsMatchers: ms, index: index})
+		}
+	}
+
+	// Get label names and sort them by popularity (desc).
+	// We select labels that have the most different values matched first to make the decision tree higher and lower.
+	labelNames := slices.Collect(maps.Keys(valuesMatchedPerLabelName))
```
tcard: Those two statements could be unified:

```suggestion
	labelNames := slices.SortedFunc(
		maps.Keys(valuesMatchedPerLabelName),
		func(a, b string) int {
			return cmp.Compare(len(valuesMatchedPerLabelName[b]), len(valuesMatchedPerLabelName[a]))
		},
	)
```

pkg/ingester/activeseries/model/matchers.go
```diff
@@ -3,28 +3,147 @@
...
+			for lmi, m := range ms.labelsMatchers {
+				if m.Name != name {
+					continue
+				}
+				var sm []string
+				if m.Type == labels.MatchEqual && m.Value != "" {
+					// This is not a set matcher, but a simple equality matcher.
+					// We'll treat it as a set matcher with a single value.
+					stack := [1]string{m.Value}
+					sm = stack[:]
```
tcard: I don't think there's an actual difference between this and just `sm := []string{m.Value}`. In both cases escape analysis is going to determine whether the underlying array should be allocated in the stack or escapes to the heap.

tcard (approved):

pkg/ingester/activeseries/model/matchers.go
colega: Done in https://github.com/grafana/mimir/pull/12184/commits/fbfc94c6bd276dfb6ce65ec6dc933ecbfccd4759

pkg/ingester/activeseries/model/matchers.go
colega: Yep, you're right, done in https://github.com/grafana/mimir/pull/12184/commits/fbfc94c6bd276dfb6ce65ec6dc933ecbfccd4759
```
pkg/ingester/activeseries/model/matchers.go:93:19: []string{...} does not escape
```

PR #12185, opened 2025-07-24
PromQL: Fix invalid source label name in label_join error message

aknuds1:
#### What this PR does

Fix invalid source label name in `label_join` error message, so it refers to the source label rather than the destination label. Add corresponding test cases.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


aknuds1:
@charleskorn I added a test suite. WDYT?

pkg/streamingpromql/operators/functions/label_test.go
charleskorn, line 1: Would it be possible to add these test cases to the existing ones in https://github.com/grafana/mimir/blob/39706d05f1e9a88ef74b6c4f4ab2a73b79802720/pkg/streamingpromql/testdata/ours/functions.test#L263?

aknuds1: I will check.

aknuds1: I moved test cases for invalid source/destination labels to functions.test (and removed labels_test.go in favour of the former). PTAL @charleskorn

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -302,6 +302,22 @@ eval range from 0 to 10m step 5m label_join(series, "(export)", "-", "label", "o
   series{"(export)"="b-2", label="b", opt="2", this="there"} 2.9 {{sum:4 count:23 buckets:[1 2 4]}}
   series{"(export)"="c-3", label="c", opt="3"} NaN _
 
+# Empty destination label name
+eval instant at 0m label_join(series, "", "-", "label")
+  expect fail regex:^invalid destination label name in label_join\(\): $
```
charleskorn, line 307: This might be slightly clearer with `expect fail msg:` instead - then there would be no need to escape the special characters.

(same applies to other tests)

aknuds1: @charleskorn Try for yourself :) `expect fail msg` doesn't work. It seems to strip the final space.

aknuds1: I did some debugging of the promqltest code, and found that the final space is stripped from the reference string. Didn't try to establish why, as I could solve it through the `regex` form.

charleskorn: 🤮 That's annoying - this is fine then.

PR #12193, opened 2025-07-25
test(otlp): make benchmark more realistic

krajorama:
Change the test to exercise labels much more.

Use 2000 series instead of 1.
Use 1 sample per series instead of 1000.
Use more realistic number of labels per series.
Make 10% of series use exponential histograms and let those have 10 exemplars per series instead of 1 per series.
Add exemplars to the first sample for each series.

Related to #12152

### DIFF
--- a/pkg/distributor/otel_test.go
+++ b/pkg/distributor/otel_test.go
@@ -706,7 +706,7 @@ func TestOTelDeltaIngestion(t *testing.T) {
 }
 
 // Extra labels to make a more realistic workload - taken from Kubernetes' embedded cAdvisor metrics.
-var extraLabels []labels.Label = []labels.Label{
+var extraLabels = []labels.Label{
 	{Name: "kubernetes_io_arch", Value: "amd64"},
 	{Name: "kubernetes_io_instance_type", Value: "c3.somesize"},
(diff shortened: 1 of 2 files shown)


aknuds1 (approved):

pkg/distributor/otel_test.go
```diff
@@ -705,33 +705,93 @@ func TestOTelDeltaIngestion(t *testing.T) {
 	}
 }
 
+// Extra labels to make a more realistic workload - taken from Kubernetes' embedded cAdvisor metrics.
+var extraLabels = []labels.Label{
```
bboreham, line 709: Very similar to here: https://github.com/grafana/mimir/blob/6178f379b9a5c70998d2bf1bbe6075bfc6443108/pkg/ingester/ingester_test.go#L10044-L10046

bboreham (approved):
Seems legit.

pkg/distributor/otel_test.go
krajorama: Intentional. I'd rather we don't sync two benchmarks on the test data. Could be skewing results accidentally because we don't run these as regression tests.

Issue #11919, opened 2025-06-30
index planning: Implement basic index lookup planning


dimitarvdimitrov:
Use existing cost estimation logic and introduce checks to abort planning when matcher counts exceed safe thresholds.

two parallelizable parts
* statistics collection:  
  * may need to add to MemPostings in mimir-prometheus or expose postings from the head(?)  
    * other stuff can probably happen in ingester  
    * \+ manage how we collect (on startup, periodically)  
    * probably don't want to collect stats during head compaction (or early head compaction)  
      * can we just skip/stop if early head compaction starts?  
    * there's a lock during MemPostings (iterates over all label names and all of their values) → takes \~400ms and could block writes during that time  
      * run in dev, look for metrics on mutex waiting (may need to enable smth) or look at profiles  
      * could lock/unlock on per-label-name basis 
  * plan generation \- [see plan.go](http://plan.go)

### Open PRs as part of this issue (updated)

- [x] https://github.com/grafana/mimir-prometheus/pull/940
- [x] https://github.com/grafana/mimir/pull/12197
- [x] https://github.com/grafana/mimir/pull/12199
- [x] https://github.com/grafana/mimir/pull/12248
- [x] https://github.com/grafana/mimir-prometheus/pull/962
- [x] https://github.com/grafana/mimir/pull/12407
- [x] https://github.com/grafana/mimir/pull/12530

PR #12197, opened 2025-07-25
index planning: implement cost model predicate

dimitarvdimitrov:
#### What this PR does

 is merged and vendored

This PR introduces the plan predicate. The predicate is a label matcher. The predicate has a selectivity, cost for applying it to a single value, and cardinality. All of these are estimations. The estimations use the changes from https://github.com/grafana/mimir-prometheus/pull/940

### Testing
This PR also introduces testing for the 3 properties of the predicate. 

The way I decided to approach testing was inspired by how CockroachDB approach testing their query cost model. They have query plans' estimates costs hard-coded (for example [unit tests for `SELECT` queries](https://github.com/cockroachdb/cockroach/blob/b8c2405930718b735bf009c802798b4918b66631/pkg/sql/opt/xform/testdata/coster/select); benchmarking for TPCH merge-join [plan](https://github.com/cockroachdb/cockroach/blob/b8c2405930718b735bf009c802798b4918b66631/pkg/sql/opt/opbench/opbench_test.go#L359-L407) and its [cost in CSV](https://github.com/cockroachdb/cockroach/blob/b8c2405930718b735bf009c802798b4918b66631/pkg/sql/opt/opbench/testdata/tpch-merge-join.csv)).

1. Hard-coded mock statistics

   This was easy to generate, it's easy to reason about and to write assertions against. An alternative is to have a generated statistics based on some cardinality code. Another alternative is to dump the cardinality from a real TSDB and use that. 
   
   The latter would be most realistic, but I'm unsure about confidentiality and about navigating this cardinality dump (i.e. writing new test cases against it).
3. CSV file instead of typical golang table tests.

   The reason that they're CSV files is so that when we improve the model, we can toggle a boolean in the test and then rewrite the files with the new estimates (inspired by cockroach's benchmarks). For this, I've also added a CSV mapper to help with mapping to and from CSV files.
   
   An alternative I tried is to only assert on the relative costs/cardinalities of different matchers. This was hard to reason about when there are more than a few matchers together in the same test case. If the test cases were few, then there were just a lot of test cases.
   
   

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/ingester/lookupplan/predicate_test.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+		},
+		func(tc testCase) []string {
+			return []string{tc.matcher.String(), fmt.Sprintf("%d", tc.seriesCount), fmt.Sprintf("%.1f", tc.expectedCost)}
+		},
+	)
+
+	testCases := mapper.ParseTestCases(t)
+
+	const writeOutNewResults = false
```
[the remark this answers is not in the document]
dimitarvdimitrov: i'm not sure if an env var (or a CLI flag) will be more readable or discoverable. The test is already failing, whoever is writing code is probably already looking at the test code.

pkg/ingester/lookupplan/csvmapper_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"testing"
+
+	"github.com/prometheus/alertmanager/matchers/parse"
+	amlabels "github.com/prometheus/alertmanager/pkg/labels"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// csvMapper provides generic CSV parsing and writing functionality for test data
```
chencs: I would elaborate here -- expected data can be read from `filePath` for tests, and test results can be written back to `filePath`. Any unit tests are expected to handle assertions on results against data from the CSV themselves. 

It took me a little longer than I would like to understand that that is what was happening in `predicate_test.go`.

chencs (approved):
This largely makes sense to me, just left a comment about clarity for the test documentation

pkg/ingester/lookupplan/csvmapper_test.go
dimitarvdimitrov: any ideas for a better name of the CVSMapper?

dimitarvdimitrov: i renamed it to `csvTestData` in https://github.com/grafana/mimir/pull/12197/commits/eec354b168a38134c02bacd40a242982d0704215

chencs: Sounds good. I think the updated comments are also much more clear, thanks!

PR #12199, opened 2025-07-25
index planning: implement cost model plan

dimitarvdimitrov:
> In draft until https://github.com/grafana/mimir-prometheus/pull/940 is merged and vendored
> This PR also depends on https://github.com/grafana/mimir/pull/12197

This PR introduces the plan. The plan is a collection of predicates. Some of which are used when reading the index and the rest are used for a sequential scan. 

### Cost

The plan has a cost model. The model consist of

- **index lookup cost**: performing index lookups for all predicates that use the index
- **intersection cost**: cost of intersecting posting lists from multiple index predicates
- **filter cost**: sequential scan i.e. cost of applying scan predicates to the fetched series)

### Testing
The tests here use the same pattern as the ones in https://github.com/grafana/mimir/pull/12197. In short:

1. Hard-coded mock statistics
3. CSV file instead of typical golang table tests.
   

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/storage/tsdb/config.go
dimitarvdimitrov, line 1: i thought i needed this constant and fixed up the comments around it. sorry for the noise

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+	predicates []planPredicate
+	// indexPredicate tracks which predicates should use index lookups vs sequential scans.
+	// indexPredicate[i] == true means predicates[i] should use the index
+	indexPredicate []bool
+	// totalSeries is the total number of series in the index.
+	totalSeries uint64
+}
+
+// newScanOnlyPlan returns a plan in which all predicates would be used to scan and none to reach from the index.
+func newScanOnlyPlan(ctx context.Context, matchers []*labels.Matcher, stats Statistics) (plan, error) {
```
chencs, line 29: Why is this desirable?

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+			return plan{}, fmt.Errorf("error converting matcher to plan predicate: %w", err)
+		}
+		p.predicates = append(p.predicates, pred)
+		p.indexPredicate = append(p.indexPredicate, false)
+	}
+
+	return p, nil
+}
+
+func (p plan) IndexMatchers() []*labels.Matcher {
```
chencs, line 47: Are you intentionally preferring value receivers?

dimitarvdimitrov: yes, i tried to keep these just as data representations and keep them immutable. if you need a new plan, then you must create one

pkg/ingester/lookupplan/plan.go
dimitarvdimitrov: it's just a starting plan so that the recursion can work more nicely. realistically it wouldn't ever be selected

dimitarvdimitrov: you can see how the recursion works in the other PR 

https://github.com/grafana/mimir/blob/dimitar/index-planning/basic-cost-model-planner/pkg/ingester/lookupplan/planner.go#L73-L87

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// useIndexFor returns a copy of this plan where predicate predicateIdx is used on the index.
+func (p plan) useIndexFor(predicateIdx int) plan {
+	p.indexPredicate = slices.Clone(p.indexPredicate)
+	p.indexPredicate[predicateIdx] = true
+	return p
+}
+
+// useScanFor returns a copy of this plan where predicate predicateIdx is used during sequential scanning.
```
chencs:
```suggestion
// useScanFor returns a copy of this plan where the predicate at predicateIdx is used during sequential scanning.
```

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+		// For example, if {protocol=~.*} matches all values, it doesn't mean it won't reduce the result set after intersection.
+		//
+		// We also assume independence between the predicates. This is a simplification.
+		// For example, the selectivity of {pod=~prometheus.*} doesn't depend on if we have already applied {statefulset=prometheus}.
+		finalSelectivity *= float64(pred.cardinality) / float64(p.totalSeries)
+	}
+	return uint64(finalSelectivity * float64(p.totalSeries))
+}
+
+// cardinality returns the total number of series that this plan would return.
```
chencs:
```suggestion
// cardinality returns an estimate of the total number of series that this plan would return.
```

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+	for i, pred := range p.predicates {
+		if !p.indexPredicate[i] {
+			continue
+		}
+
+		// We use the selectivity across all series instead of the selectivity across label values.
+		// For example, if {protocol=~.*} matches all values, it doesn't mean it won't reduce the result set after intersection.
+		//
+		// We also assume independence between the predicates. This is a simplification.
+		// For example, the selectivity of {pod=~prometheus.*} doesn't depend on if we have already applied {statefulset=prometheus}.
```
chencs, line 138: I would include a comment somewhere about what this means for the selectivity estimation. The resulting selectivity is neither an upperbound or a lowerbound, but assuming independence lets us make relative statements across different plans without needing to inspect predicate relationships.

pkg/ingester/lookupplan/plan.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return uint64(finalSelectivity * float64(p.totalSeries))
+}
+
+// cardinality returns the total number of series that this plan would return.
+func (p plan) cardinality() uint64 {
+	finalSelectivity := 1.0
+	for _, pred := range p.predicates {
+		// We use the selectivity across all series instead of the selectivity across label values.
+		// For example, if {protocol=~.*} matches all values, it doesn't mean it won't reduce the result set after intersection.
```
chencs: Double negatives are more confusing to read
```suggestion
		// For example, if {protocol=~.*} matches all values, it could still reduce the result set after intersection.
```

pkg/ingester/lookupplan/plan.go
dimitarvdimitrov:
```suggestion
		// For example, the selectivity of {pod=~prometheus.*} doesn't depend on if we have already applied {statefulset=prometheus}.
		// While finalSelectivity is neither an upper bound nor a lower bound, assuming independence allows us to come up with cost estimates comparable between plans.
```

pkg/ingester/lookupplan/plan.go
dimitarvdimitrov: thanks for calling it out

dimitarvdimitrov:
PTAL @chencs 🙏

pkg/ingester/lookupplan/plan_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	testCases := data.ParseTestCases(t)
+
+	const writeOutNewCost = false
+	if writeOutNewCost {
+		t.Cleanup(func() { data.WriteTestCases(t, testCases) })
+	}
+
+	for tcIdx, tc := range testCases {
+		testName := fmt.Sprintf("%s_%s", tc.indexMatchers, tc.scanMatchers)
```
chencs: Sorry, last bit of feedback. I think everything else looks good, but I do wish that these tests included names describing what the purpose of the test is rather than just the parameters, or at least have the intent for each test written down somewhere (maybe a description header in the CSV), e.g., "0 index matchers should incur large filter cost"

dimitarvdimitrov: yeah fair, a lot of them was just me fuzz-adding test cases, but i can try to describe them. i'll do the same for the planner.go PR as well

dimitarvdimitrov: ✔️ https://github.com/grafana/mimir/pull/12199/commits/25bfad7995def022daf16f30bf676cd5758eebd0

some of them aren't super helpful because they're there just to have more diverse test cases

dimitarvdimitrov:
i rebased on main, there shouldn't be any differences

Issue #11884, opened 2025-06-27
Require query-scheduler component


armandgrillet:
This is a breaking change. This allows us to remove the v1 frontend (the scheduler's request queue embedded into the frontend). It would also allow us to delete a bunch of type casts in the scheduler/queue code.

56quarters:
The plan for this is:
* [x] Remove option to disable query-scheduler in jsonnet. [PR](https://github.com/grafana/mimir/pull/12187)
* [x] Remove option to disable query-scheduler in helm. [PR](https://github.com/grafana/mimir/pull/12188)
* [x] Remove code that supports the v1 query-scheduler (embedded in the query-frontend). [PR](https://github.com/grafana/mimir/pull/12200)
* [x] ~Write migration docs for the helm chart~ I'm not sure what I was imagining for this. Using the query-scheduler has been the default in the helm chart for years.
* [x] Update docker-compose setup. [PR](https://github.com/grafana/mimir/pull/12445)
* [x] Update general docs to indicate query-scheduler is required. [PR](https://github.com/grafana/mimir/pull/12609)

PR #12200, opened 2025-07-25
refactor: make query-scheduler a required component

56quarters:
#### What this PR does

This change removes the ability to use the v1 query-frontend which used an embedded version of the query-scheduler. Instead, v2 of the query-frontend is now required which requires a stand-alone deployment of the query-scheduler. Note that this change does _not_ do any clean up of the query-frontend, it only removes v1.

Part of https://github.com/grafana/mimir/issues/11884

#### Notes to reviewers

I haven't updated documentation in this PR to indicate the scheduler is required. I need to track down the sources for some images we include in docs and update them (which I anticipate taking a little while).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/mimir/modules.go
```diff
@@ -845,10 +845,18 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
 	t.Cfg.Frontend.FrontendV2.LookBackDelta = t.Cfg.Querier.EngineConfig.LookbackDelta
 	t.Cfg.Frontend.FrontendV2.QueryStoreAfter = t.Cfg.Querier.QueryStoreAfter
 
-	roundTripper, frontendV1, frontendV2, err := frontend.InitFrontend(
+	// If the query-frontend is running in the same process as the query-scheduler and
+	// the query-scheduler hasn't been explicitly configured, Mimir will default to trying
+	// to connect to a scheduler on localhost on its own gRPC listening port.
+	if t.Cfg.isAnyModuleEnabled(QueryScheduler, Read, All) && !t.Cfg.Frontend.FrontendV2.IsSchedulerConfigured() {
```
56quarters, line 851: I'm actually not sure about this list of modules. I _think_ `Read` is wrong because the scheduler is part of `Backend` when running in read-write mode so trying `127.0.0.1:$port` wouldn't work.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * [CHANGE] Memcached: Remove experimental `-<prefix>.memcached.addresses-provider` flag to use alternate DNS service discovery backends. The more reliable backend introduced in 2.16.0 (#10895) is now the default. As a result of this change, DNS-based cache service discovery no longer supports search domains. #12175
 * [CHANGE] Query-frontend: Remove the CLI flag `-query-frontend.downstream-url` and corresponding YAML configuration and the ability to use the query-frontend to proxy arbitrary Prometheus backends. #12191
 * [CHANGE] Query-frontend: Remove experimental instant query splitting feature. #12267
+* [CHANGE] Query-frontend: Remove the CLI flags `-querier.max-outstanding-requests-per-tenant` and `-query-frontend.querier-forget-delay` and corresponding YAML configuration. This is part of a change that makes the query-scheduler a required component. A result of this removes the ability to run the query-frontend with an embedded query-scheduler. Instead, a dedicated query-scheduler component must be run. #12200
```
```suggestion
* [CHANGE] Query-frontend: Remove the CLI flags `-querier.max-outstanding-requests-per-tenant` and `-query-frontend.querier-forget-delay` and corresponding YAML configurations. This is part of a change that makes the query-scheduler a required component. This removes the ability to run the query-frontend with an embedded query-scheduler. Instead, you must run a dedicated query-scheduler component. #12200
```

tacole02 (approved):
Docs look good, a few minor nits in the changelog.

pkg/mimir/modules.go
```diff
@@ -659,7 +659,7 @@ func (t *Mimir) initQuerier() (serv services.Service, err error) {
 	}
 
 	// If neither query-frontend nor query-scheduler is in use, then no worker is needed.
```
charleskorn: Might not be for this PR, but does it make sense to run a querier without a query-scheduler?

charleskorn:
```suggestion
	// If no query-scheduler is in use, then no worker is needed.
```

pkg/mimir/modules.go
```diff
@@ -845,10 +845,18 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
 	t.Cfg.Frontend.FrontendV2.LookBackDelta = t.Cfg.Querier.EngineConfig.LookbackDelta
 	t.Cfg.Frontend.FrontendV2.QueryStoreAfter = t.Cfg.Querier.QueryStoreAfter
 
-	roundTripper, frontendV1, frontendV2, err := frontend.InitFrontend(
+	// If the query-frontend is running in the same process as the query-scheduler and
+	// the query-scheduler hasn't been explicitly configured, Mimir will default to trying
+	// to connect to a scheduler on localhost on its own gRPC listening port.
+	if t.Cfg.isAnyModuleEnabled(QueryScheduler, Read, All) && !t.Cfg.Frontend.FrontendV2.IsSchedulerConfigured() {
+		address := fmt.Sprintf("127.0.0.1:%d", t.Cfg.Server.GRPCListenPort)
+		level.Info(util_log.Logger).Log("msg", "The querier-frontend has not been configured with the query-scheduler address. Because Mimir is running in monolithic mode, it's attempting an automatic frontend configuration. If queries are unresponsive, consider explicitly configuring the query-scheduler address for querier-frontend.", "address", address)
```
charleskorn:
```suggestion
		level.Info(util_log.Logger).Log("msg", "The query-frontend has not been configured with the query-scheduler address. Because Mimir is running in monolithic mode, it's attempting an automatic frontend configuration. If queries are unresponsive, consider explicitly configuring the query-scheduler address for querier-frontend.", "address", address)
```

pkg/mimir/modules.go
charleskorn: Is there a way to ask if the query-scheduler is enabled, either directly or indirectly (eg. due to monolithic mode)? Then we wouldn't have to specify all the possible ways the query-scheduler is enabled here.

pkg/querier/worker/worker.go
```diff
@@ -150,20 +150,8 @@ func NewQuerierWorker(cfg Config, handler RequestHandler, log log.Logger, reg pr
...
-			return servicediscovery.NewDNS(log, cfg.FrontendAddress, cfg.DNSLookupPeriod, receiver)
-		}
-
-		grpcCfg = cfg.QueryFrontendGRPCClientConfig
-		workerClient = "query-frontend-worker"
-		processor = newFrontendProcessor(cfg, handler, log)
-
 	default:
-		return nil, errors.New("no query-scheduler or query-frontend address")
+		return nil, errors.New("no query-scheduler address")
```
charleskorn: Should we turn this into something like `if !cfg.IsSchedulerConfigured() { ... }` before the `switch`? Then we could drop the `switch` altogether.

CHANGELOG.md
56quarters: Fixed in fda0fda357884f4213f25572d2740087ecca1cb6

pkg/mimir/modules.go
56quarters: Fixed in fda0fda357884f4213f25572d2740087ecca1cb6

pkg/mimir/modules.go
56quarters: Fixed in fda0fda357884f4213f25572d2740087ecca1cb6

pkg/mimir/modules.go
56quarters: I don't think it does but I'll save this particular bit of cleanup for a follow-up PR.

pkg/mimir/modules.go
56quarters: > Is there a way to ask if the query-scheduler is enabled, either directly or indirectly (eg. due to monolithic mode)? Then we wouldn't have to specify all the possible ways the query-scheduler is enabled here.

If I understand correctly, that's kind of what we are doing here. I don't think there are any other ways we can tell the query-scheduler is enabled besides the logic here:

* The direct way to enable it is setting `scheduler-address` for the query-frontend component (`IsSchedulerConfigured()` here). In this case we don't need to configure any address since it's already set.
* The indirect way to tell if the query-scheduler is enabled is checking if query-frontend is being run in the same process as the query-scheduler due to monolithic mode. In this case, we set the `scheduler-address` to `localhost:9095` which _should_ work but we log a warning letting people know that it might not.

I can't think of any other way to determine the scheduler address at this level. We already set it to the scheduler headless service in the Helm chart and Jsonnet by default.

pkg/querier/worker/worker.go
56quarters: Fixed in fb590f732ba25e0eb606c937c09a64434221a604

pkg/mimir/modules.go
charleskorn: `isAnyModuleEnabled` checks if any of the provided module names were listed in the `-target` CLI flag (the name of this method is confusing to me - perhaps we need to rename it to something like `targetModuleIsAnyOf`?)

What I'm imagining is we use a similar method (`isEnabled`?) here that returns true if the given module is either given in the `-target` CLI flag, or if any of the modules in the `-target` CLI flag transitively depends on the given module. 

Then the logic here could be simplified to:

```go
if t.Cfg.isEnabled(QueryScheduler) && !t.Cfg.Frontend.FrontendV2.IsSchedulerConfigured() {
```

...and this would cover all the cases where the query-scheduler is running in this process but the scheduler address hasn't been explicitly set.

56quarters: Ah, I see. We could do something like this to check if the query-scheduler is the target module or a dependency of any of the target modules:

```go
	deps := make(map[string]struct{})
	for _, tgt := range t.Cfg.Target {
		for _, dep := range t.ModuleManager.DependenciesForModule(tgt) {
			deps[dep] = struct{}{}
		}
	}

	_, schedulerEnabled := deps[QueryScheduler]
	schedulerEnabled = schedulerEnabled || t.Cfg.isModuleEnabled(QueryScheduler)
```

However, I don't really like inspecting the module manager like this from the init method and I don't want to make the existing `Config` struct aware of it either. I'd prefer to leave the (admittedly hacky) existing check since it mirrors how other modules work.

pkg/querier/worker/worker.go
```diff
@@ -139,33 +134,20 @@ func NewQuerierWorker(cfg Config, handler RequestHandler, log log.Logger, reg pr
...
+	level.Info(log).Log("msg", "Starting querier worker connected to query-scheduler", "scheduler", cfg.SchedulerAddress)
 
-	default:
-		return nil, errors.New("no query-scheduler or query-frontend address")
+	factory = func(receiver servicediscovery.Notifications) (services.Service, error) {
+		return schedulerdiscovery.New(cfg.QuerySchedulerDiscovery, cfg.SchedulerAddress, cfg.DNSLookupPeriod, "querier", receiver, log, reg)
 	}
 
+	grpcCfg = cfg.QuerySchedulerGRPCClientConfig
+	workerClient = "query-scheduler-worker"
```
charleskorn: Are these variables required now that they're always set to the same thing?

56quarters: Good point, will fix.

56quarters: Fixed in 1bea872afe5304bcffcb33bfa0e4f4ea60db0628

PR #12222, opened 2025-07-28
<WIP> Block-builder concurrent fetcher

seizethedave:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -165,10 +165,7 @@ func (b *BlockBuilder) stopping(_ error) error {
 
 // running learns about the jobs from a block-builder-scheduler, and consumes one job at a time.
 func (b *BlockBuilder) running(ctx context.Context) error {
-	// We control our own context here. We'll stop processing jobs when the
-	// service context is cancelled, but let any in-flight jobs complete.
-
-	// Kick off the scheduler's run loop.
+	// Kick off the scheduler client's run loop.
 	go b.schedulerClient.Run(ctx)
 
 	for {
@@ -267,12 +264,42 @@ func (b *BlockBuilder) consumePartitionSection(
 	})
 	defer b.kafkaClient.RemoveConsumePartitions(map[string][]int32{b.cfg.Kafka.Topic: {partition}})
 
+	b.kafkaClient.ForceMetadataRefresh()
+
 	level.Info(logger).Log("msg", "start consuming", "partition", partition, "start_offset", startOffset, "end_offset", endOffset)
 
 	var (
 		firstRec *kgo.Record
 		lastRec  *kgo.Record
 	)
+
+	fetcher, ferr := ingest.NewConcurrentFetchers(
+		ctx,
+		b.kafkaClient,
+		logger,
+		b.cfg.Kafka.Topic,
+		partition,
+		startOffset,
+		b.cfg.ConcurrentFetcherConcurrency,
+		100_000_000,
+		false,
+		5*time.Second,
+		nil,
+		ingest.NewGenericOffsetReader[int64](func(context.Context) (int64, error) { return 0, nil }, 5*time.Second, logger), // Dummy.
+		backoff.Config{
+			MinBackoff: 100 * time.Millisecond,
+			MaxBackoff: 1 * time.Second,
+		},
+		nil)
+	if ferr != nil {
+		return lastConsumedOffset, fmt.Errorf("creating concurrent fetcher: %w", ferr)
+	}
+
+	fetcher.Start(ctx)
+	defer fetcher.Stop()
+
+	consumeMetric := b.blockBuilderMetrics.recordsConsumedTotal.WithLabelValues(fmt.Sprint(partition))
+	metricUpdatePending := 0
 
 consumerLoop:
 	for recOffset := int64(-1); recOffset < endOffset-1; {
@@ -284,7 +311,7 @@ consumerLoop:
 		// we cannot tell if the consumer has already reached the latest end of the partition, i.e. no more records to consume,
 		// or there is more data in the backlog, and we must retry the poll. That's why the consumer loop above has to guard
 		// the iterations against the endOffset, so it retries the polling up until the expected end of the partition is reached.
-		fetches := b.kafkaClient.PollFetches(ctx)
+		fetches, _ := fetcher.PollFetches(ctx)
 		fetches.EachError(func(_ string, _ int32, err error) {
 			if !errors.Is(err, context.Canceled) {
 				level.Error(logger).Log("msg", "failed to fetch records", "err", err)
@@ -293,6 +320,12 @@ consumerLoop:
 		})
 
 		for recIter := fetches.RecordIter(); !recIter.Done(); {
+
+			if metricUpdatePending > 100 {
+				consumeMetric.Add(float64(metricUpdatePending))
+				metricUpdatePending = 0
+			}
+
 			rec := recIter.Next()
 			recOffset = rec.Offset
 
@@ -305,6 +338,8 @@ consumerLoop:
 				break consumerLoop
 			}
 
+			metricUpdatePending++
+
 			err := builder.Process(ctx, rec)
 			if err != nil {
 				// All "non-terminal" errors are handled by the TSDBBuilder.
@@ -313,6 +348,9 @@ consumerLoop:
 			lastRec = rec
 		}
 	}
+
+	consumeMetric.Add(float64(metricUpdatePending))
+	metricUpdatePending = 0
 
 	// Nothing was consumed from Kafka at all.
 	if firstRec == nil {
--- a/pkg/storage/ingest/fetcher.go
+++ b/pkg/storage/ingest/fetcher.go
@@ -248,7 +248,7 @@ func newErrorFetchResult(ctx context.Context, partitionID int32, err error) fetc
 	}
 }
 
-type concurrentFetchers struct {
+type ConcurrentFetchers struct {
 	wg   sync.WaitGroup
 	done chan struct{}
 
@@ -270,7 +270,7 @@ type concurrentFetchers struct {
 	lastReturnedOffset int64
 	startOffset        int64
 	startConcurrency   int
-	startOffsetsReader *genericOffsetReader[int64]
+	startOffsetsReader *GenericOffsetReader[int64]
 
 	// fetchBackoffConfig is the config to use for the backoff in case of Fetch errors.
 	// We set it here so that tests can override it run faster.
@@ -285,8 +285,8 @@ type concurrentFetchers struct {
 	estimatedBytesPerRecord *atomic.Int64
 }
 
-// newConcurrentFetchers creates a new concurrentFetchers. startOffset can be kafkaOffsetStart, kafkaOffsetEnd or a specific offset.
-func newConcurrentFetchers(
+// NewConcurrentFetchers creates a new concurrentFetchers. startOffset can be kafkaOffsetStart, kafkaOffsetEnd or a specific offset.
+func NewConcurrentFetchers(
 	ctx context.Context,
 	client *kgo.Client,
 	logger log.Logger,
@@ -298,10 +298,10 @@ func newConcurrentFetchers(
 	trackCompressedBytes bool,
 	minBytesWaitTime time.Duration,
 	offsetReader *partitionOffsetClient,
-	startOffsetsReader *genericOffsetReader[int64],
+	startOffsetsReader *GenericOffsetReader[int64],
 	fetchBackoffConfig backoff.Config,
 	metrics *readerMetrics,
-) (*concurrentFetchers, error) {
+) (*ConcurrentFetchers, error) {
 	if fetchBackoffConfig.MaxBackoff == 0 {
 		// Ensure it's not the zero value, which means we haven't got the backoff config due to a bug.
 		return nil, errors.New("fetchBackoffConfig.MaxBackoff has not been set")
@@ -328,7 +328,7 @@ func newConcurrentFetchers(
 	if maxBufferedBytesLimit <= 0 {
 		maxBufferedBytesLimit = math.MaxInt32
 	}
-	f := &concurrentFetchers{
+	f := &ConcurrentFetchers{
 		bufferedFetchedRecords:  atomic.NewInt64(0),
 		bufferedFetchedBytes:    atomic.NewInt64(0),
 		estimatedBytesPerRecord: atomic.NewInt64(0),
@@ -366,27 +366,27 @@ func newConcurrentFetchers(
 }
 
 // BufferedRecords implements fetcher.
-func (r *concurrentFetchers) BufferedRecords() int64 {
+func (r *ConcurrentFetchers) BufferedRecords() int64 {
 	return r.bufferedFetchedRecords.Load()
 }
 
 // BufferedBytes implements fetcher.
-func (r *concurrentFetchers) BufferedBytes() int64 {
+func (r *ConcurrentFetchers) BufferedBytes() int64 {
 	return r.bufferedFetchedBytes.Load()
 }
 
-func (r *concurrentFetchers) EstimatedBytesPerRecord() int64 {
+func (r *ConcurrentFetchers) EstimatedBytesPerRecord() int64 {
 	return r.estimatedBytesPerRecord.Load()
 }
 
 // Start implements fetcher.
-func (r *concurrentFetchers) Start(ctx context.Context) {
+func (r *ConcurrentFetchers) Start(ctx context.Context) {
 	r.wg.Add(1)
 	go r.start(ctx, r.startOffset, r.startConcurrency)
 }
 
 // Stop implements fetcher.
-func (r *concurrentFetchers) Stop() {
+func (r *ConcurrentFetchers) Stop() {
 	// Ensure it's not already stopped.
 	select {
 	case _, ok := <-r.done:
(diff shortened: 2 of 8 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -49,11 +49,10 @@ type BlockBuilder struct {
 	schedulerClient schedulerpb.SchedulerClient
 	schedulerConn   *grpc.ClientConn
 
-	// the current job iteration number. For tests.
-	jobIteration atomic.Int64
```
seizethedave, line 53: No longer in use.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -235,6 +240,61 @@ func (b *BlockBuilder) consumeJob(ctx context.Context, key schedulerpb.JobKey, s
...
+			logger,
+			b.cfg.Kafka.Topic,
+			partition,
+			startOffset,
+			b.cfg.Kafka.FetchConcurrencyMax,
+			100_000_000,
+			false,
+			5*time.Second,
+			nil, // Don't need a reader since we've provided the start offset.
+			ingest.NewGenericOffsetReader(func(context.Context) (int64, error) { return 0, nil }, 5*time.Second, logger), // Dummy.
```
seizethedave: This is the thing that is still outstanding in this PR. If you read how it's used, it's used to figure out which offset to resume from _in the event_ that we encounter an `OffsetOutOfRange` error. We should not ever encounter that. If we did, we should probably abort the block-builder job. We do not want the reader to resume from "any old place" if there's an error. I'd rather influence the concurrentfetcher to return an error if this happened. I'll think about it, but open to discussion in the meantime.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -273,30 +333,47 @@ func (b *BlockBuilder) consumePartitionSection(
 			"last_consumed_offset", lastConsumedOffset, "num_blocks", len(blockMetas))
 	}(time.Now())
 
+	b.kafkaClient.ForceMetadataRefresh()
 	b.kafkaClient.AddConsumePartitions(map[string]map[int32]kgo.Offset{
 		b.cfg.Kafka.Topic: {
 			partition: kgo.NewOffset().At(startOffset),
 		},
 	})
 	defer b.kafkaClient.RemoveConsumePartitions(map[string][]int32{b.cfg.Kafka.Topic: {partition}})
```
seizethedave: Even with the concurrent fetcher, I encountered dodgy behavior/panics/errors when I didn't have these here.

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: i'm not sure if starting at 0 also won't trigger OffsetOutOfRange if 0 has been truncated already and lead to some infinite error loop. I think that was the idea of the start offset reader.

Is there a point in trying to compact an offset that's been truncated? Technically that data no longer exists, so we might as well admit data loss and sound some alarms.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -243,6 +250,61 @@ func (b *BlockBuilder) consumeJob(ctx context.Context, key schedulerpb.JobKey, s
...
+		f, ferr := ingest.NewConcurrentFetchers(
+			ctx,
+			b.kafkaClient,
+			logger,
+			b.cfg.Kafka.Topic,
+			partition,
+			startOffset,
+			b.cfg.Kafka.FetchConcurrencyMax,
+			100_000_000,
+			false,
```
dimitarvdimitrov: why hardcoded `false` here? there's `b.cfg.Kafka.UseCompressedBytesAsFetchMaxBytes`

there's also an option for the wait time (FetchMaxWait)

pkg/blockbuilder/blockbuilder.go
```diff
@@ -243,6 +250,61 @@ func (b *BlockBuilder) consumeJob(ctx context.Context, key schedulerpb.JobKey, s
...
+			backoff.Config{
+				MinBackoff: 100 * time.Millisecond,
+				MaxBackoff: 1 * time.Second,
+			},
+			b.readerMetrics)
+		if ferr == nil {
+			return f, nil
+		}
+		lastError = ferr
+		boff.Wait()
```
dimitarvdimitrov, line 326: nit to add some logging; otherwise this whole loop can take ~26s and the BB would look like it's doing nothing in this time

pkg/blockbuilder/blockbuilder.go
```diff
@@ -74,13 +73,25 @@ func newWithSchedulerClient(
 	limits *validation.Overrides,
 	schedulerClient schedulerpb.SchedulerClient,
 ) (*BlockBuilder, error) {
+	kpm := ingest.NewKafkaReaderClientMetrics(ingest.ReaderMetricsPrefix, "block-builder", reg)
+
+	var readerMetrics *ingest.ReaderMetrics
+	if cfg.Kafka.FetchConcurrencyMax > 0 {
+		// Reader metrics tracks some per-partition things, but block-builder
+		// will always be dealing with different partitions. So we pass -1.
+		m := ingest.NewReaderMetrics(-1, reg, &ingest.NoOpReaderMetricsSource{}, cfg.Kafka.Topic, kpm)
```
dimitarvdimitrov: i'm not sure why we declare the partition here anyways. we know the partition in `notifyLastConsumedOffset` so we can use that instead of passing it here. 

That way the metric will also give you more info for the BB. As it is right now it will reset itself as you move between partitions and there's the chance you lose some data as you set the gauge between scrape intervals

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: ForceMetadataRefresh is asynchronous, i don't know if you don't want to make sure you've observed an MD refresh before going further

pkg/storage/ingest/reader.go
```diff
@@ -969,7 +972,21 @@ type readerMetricsSource interface {
 	EstimatedBytesPerRecord() int64
 }
 
-func newReaderMetrics(partitionID int32, reg prometheus.Registerer, metricsSource readerMetricsSource, topic string) readerMetrics {
+type NoOpReaderMetricsSource struct{}
```
dimitarvdimitrov: nit/tip: concurrent fetchers implement these methods. You can wire them up to some `swappableMetricsSource` backed by ConcurrentFetchers which can be swapped out, so you still get these metrics exported.

pkg/storage/ingest/reader.go
```diff
@@ -969,7 +972,21 @@ type readerMetricsSource interface {
...
+
+func (n *NoOpReaderMetricsSource) BufferedRecords() int64 {
+	return 0
+}
+
+func (n *NoOpReaderMetricsSource) EstimatedBytesPerRecord() int64 {
+	return 0
+}
+
+func NewReaderMetrics(partitionID int32, reg prometheus.Registerer, metricsSource readerMetricsSource, topic string, kpromMetrics *kprom.Metrics) ReaderMetrics {
```
dimitarvdimitrov: should the `readerMetricsSource` interface be exported too?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -280,30 +342,47 @@ func (b *BlockBuilder) consumePartitionSection(
 			"last_consumed_offset", lastConsumedOffset, "num_blocks", len(blockMetas))
 	}(time.Now())
 
+	b.kafkaClient.ForceMetadataRefresh()
```
narqo: (nit) Could you add a code-comment, mentioning why the ForceMetadataRefresh is needed here

pkg/blockbuilder/blockbuilder.go
```diff
@@ -280,30 +342,47 @@ func (b *BlockBuilder) consumePartitionSection(
...
 		},
 	})
 	defer b.kafkaClient.RemoveConsumePartitions(map[string][]int32{b.cfg.Kafka.Topic: {partition}})
 
+	var fetchPoller fetchPoller = b.kafkaClient
+
+	if b.cfg.Kafka.FetchConcurrencyMax > 0 {
+		if b.readerMetrics == nil {
+			panic("readerMetrics should be non-nil when concurrent fetchers are used")
+		}
```
narqo: (nit) Can we move this panic inside `newFetchers`. Otherwise, it looks unrelated to what is happening inside this if

pkg/blockbuilder/blockbuilder.go
seizethedave: Haha. Now I'm going to have to figure out why this is needed. 😢

pkg/blockbuilder/blockbuilder.go
seizethedave: 3b0be8a2d6

pkg/storage/ingest/reader.go
seizethedave: fd35eb32cc

pkg/blockbuilder/blockbuilder.go
seizethedave: `b.cfg.Kafka.MaxBufferedBytes`, too.

pkg/blockbuilder/blockbuilder.go
seizethedave: hmm. `notifyLastConsumedOffset` is a method of `partitionReader` which I am not actually using in block-builder. 🤔

seizethedave: I've eliminated the partitionID parameter anyway. 410da9f2f9

seizethedave: Should I be using PartitionReader? ❓

pkg/storage/ingest/reader.go
seizethedave: Cool. dc0aecb187

pkg/blockbuilder/blockbuilder.go
seizethedave: I think it would make sense to tell `NewConcurrentFetchers` that my policy is the entire PollFetches should abort and fail if we ever saw an `OffsetOutOfRange`. It looks like I'd be blazing a new trail in there, though - there's no precedent for having a fetch error that aborts consumption.

Looking more at `ConcurrentFetchers.run` to see how I could add such a thing...

pkg/blockbuilder/blockbuilder.go
seizethedave: I'm going to test not calling this method at all. Would that be unsound?

pkg/blockbuilder/blockbuilder.go
seizethedave: Now I recall, I liked not using PartitionReader because that type also manipulates stored offsets. I'll assume sticking with the current regime for now.

pkg/blockbuilder/blockbuilder.go
seizethedave: I've found that I don't need to call it. Because `AddConsumePartitions` does the same, and I call that right after.

pkg/blockbuilder/blockbuilder.go
seizethedave: I'm not going to call it. AddConsumePartitions already calls the same helper called by ForceMetadataRefresh. Of course, I'll test it.

pkg/blockbuilder/blockbuilder.go
seizethedave: @dimitarvdimitrov I've added a new `OnRangeErrorAbort` policy that one can specify that, in case OutOfRange is encountered, the fetches will cease, and the results obtained by `PollFetches` will have an `Err` in it.
I still need to write a test, but let me know what you think of this system.

pkg/storage/ingest/fetcher.go
```diff
@@ -927,30 +966,37 @@ func handleKafkaFetchErr(err error, fw fetchWant, longBackoff waiter, partitionS
...
+				// We set a MaxWaitMillis on fetch requests, but even then there may be no records for some time.
+				// Wait for a short time to allow the broker to catch up or for new records to be produced.
+				level.Debug(logger).Log("msg", "offset out of range; waiting for new records to be produced")
+			}
 		} else {
-			// If the broker is behind or if we are requesting offsets which have not yet been produced, we end up here.
-			// We set a MaxWaitMillis on fetch requests, but even then there may be no records for some time.
-			// Wait for a short time to allow the broker to catch up or for new records to be produced.
-			level.Debug(logger).Log("msg", "offset out of range; waiting for new records to be produced")
+			panic(fmt.Sprintf("unexpected range error policy: %d", rangeErrorPolicy))
```
dimitarvdimitrov, line 1003: can you move this to the constructor and maybe return an error there? detecting this earlier is better i think although in reality we shouln't hit this

pkg/storage/ingest/fetcher.go
```diff
@@ -685,7 +701,29 @@ func (r *concurrentFetchers) run(ctx context.Context, wants chan fetchWant, logg
...
+				w, err = handleKafkaFetchErr(res.Err, w, errBackoff, r.rangeErrorPolicy, r.startOffsetsReader, r.client, attemptSpan)
+				if err != nil {
+					// An error returned from handleKafkaFetchErr is a fatal
+					// error. We produce an error result and abort further
+					// processing.
+
+					select {
+					case <-r.done:
+					case <-ctx.Done():
+					case w.result <- fetchResult{
```
dimitarvdimitrov: can't w.result can be nil? same below when we close it

pkg/storage/ingest/fetcher.go
```diff
@@ -910,7 +948,8 @@ type metadataRefresher interface {
 // handleKafkaFetchErr handles all the errors listed in the franz-go documentation as possible errors when fetching records.
 // For most of them we just apply a backoff. They are listed here so we can be explicit in what we're handling and how.
 // It may also return an adjusted fetchWant in case the error indicated, we were consuming not yet produced records or records already deleted due to retention.
-func handleKafkaFetchErr(err error, fw fetchWant, longBackoff waiter, partitionStartOffset *genericOffsetReader[int64], refresher metadataRefresher, logger log.Logger) fetchWant {
+func handleKafkaFetchErr(err error, fw fetchWant, longBackoff waiter, rangeErrorPolicy RangeErrorPolicy,
```
dimitarvdimitrov, line 955: can you update the godoc here for the error?

pkg/storage/ingest/fetcher.go
```diff
@@ -685,7 +701,29 @@ func (r *concurrentFetchers) run(ctx context.Context, wants chan fetchWant, logg
...
+							Partition: r.partitionID,
+							Err:       err,
+						},
+					}:
+					}
+
+					wantSpan.Finish()
+					close(w.result)
+					level.Warn(logger).Log("msg", "failed to handle Kafka fetch error", "error", err)
+					return
```
dimitarvdimitrov: This will leave a gap in records and will reduce our fetch concurrency because nothing will start this worker back. wdyt about returning an error on the w.result, but still continuing to execute as it was before? That way the concurrent fetchers are still usable after these errors, and it's up to the client to decide what they want to do.

this is also in a similar fashion to what kgo does with `KeepRetryableFetchErrors()`.


i think this doesn't make a ton of difference to how the block builder uses this, but i find it a bit more intuitive behaviour and less surprising and easier to document/explain

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: PartitionReader should be a thing which wires together the fetchers and the consumers.  it's more geared towards the ingester use-case with what it does at startup for example

pkg/storage/ingest/fetcher.go
dimitarvdimitrov: which also makes me think that these selects should be somehow abstracted away - there are already 4 slightly different versions of that select in this function. idk if you have time to think about this or i should. either way - probably something for a follow-up PR

pkg/storage/ingest/fetcher.go
seizethedave: Added a validation to the constructor { 9ac13c6545 }. But I'll leave this here, as it's more about "is this code handling all known values of this enum?"

pkg/storage/ingest/fetcher.go
seizethedave: Cool, that makes sense to me. The decision to abort is in the application domain. Kind of out of place in a general-purpose Kafka consumer type.

seizethedave: 593536ea9f

pkg/storage/ingest/fetcher.go
seizethedave: The `w.result==nil` thing I've rectified. But, yeah separate PR for the latter SG2M.

pkg/blockbuilder/blockbuilder.go
```diff
@@ -287,29 +372,50 @@ func (b *BlockBuilder) consumePartitionSection(
...
+	}
+
 	level.Info(logger).Log("msg", "start consuming", "partition", partition, "start_offset", startOffset, "end_offset", endOffset)
 
-	var (
-		firstRecOffset = int64(-1)
-		lastRecOffset  = int64(-1)
-	)
+	firstRecOffset := int64(-1)
+	lastRecOffset := int64(-1)
```
narqo: (nit) the `lastRecOffset` should be replaced with the `lastConsumedOffset`, that is defined at the top. The latter is what the top-most defer logs to track the over-consumption.

This chunk of the log is nice-to-have, thus nit. We can fix it separately.

narqo (approved):
My comments are addressed. I've left one more note, but the changes overall, LGTM 🔥

tacole02 (approved):
Changelog looks good!

pkg/blockbuilder/blockbuilder.go
seizethedave: Nice spot. I always get tripped by references to named return values.

🕶️  e0b145a65b362b9b592c2c654c45b4c217205204

PR #12232, opened 2025-07-29
build: Fix git worktree warnings in Docker containers

dimitarvdimitrov:
## Summary
- Fixes git repository errors when running make commands in worktrees
- Eliminates `fatal: not a git repository` warnings in Docker containers

## What was happening
When using git worktrees, make commands like `make generate-otlp` would show git warnings because the Docker container couldn't access the git metadata. In worktrees, `.git` is a file pointing to split metadata directories that weren't being mounted.

## The fix
Added conditional detection and mounting of both required git directories:
- Worktree-specific metadata (HEAD, index, refs)  
- Common git metadata (objects, shared refs, config)

Only activates in worktrees, maintains compatibility with regular repos.


56quarters (approved):
Approved but this seems like a lot for a makefile. Maybe we can move the parsing and detection of worktrees to a shell script in `tools/`?

dimitarvdimitrov:
fair, i did that in https://github.com/grafana/mimir/pull/12232/commits/17f6209519e25e38359086744a34f1e9e0ca8dcc

tools/git-worktree-volumes.sh
```diff
@@ -0,0 +1,35 @@
...
+# Find the common git directory by reading commondir
+if [[ -f "$worktree_gitdir/commondir" ]]; then
+    common_gitdir=$(cd "$worktree_gitdir" && cd "$(cat commondir)" && pwd)
+    
+    # Output volume mounts for both directories (without mount options, those are added by make)
+    echo "-v $worktree_gitdir:$worktree_gitdir"
+    if [[ -d "$common_gitdir" ]]; then
+        echo "-v $common_gitdir:$common_gitdir"
+    fi
+fi
```
dimitarvdimitrov:
```suggestion
fi

```

56quarters (approved):
Thanks!

PR #12238, opened 2025-07-30
Pass ctx to Map and MapExpr for ASTMapper

zenador:
#### What this PR does

Pass ctx to `Map` and `MapExpr` for `ASTMapper`, instead of storing it in the structs, so that we don't have to recreate the mapper when we only want to change ctx, and handle context cancellation in the `ASTMapper` so implementations don't have to. Suggested by @charleskorn in [here](https://github.com/grafana/mimir/pull/11971#discussion_r2196320853).

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/frontend/querymiddleware/prom2_range_compat.go
```diff
@@ -81,8 +81,8 @@ func (c *prom2RangeCompatHandler) rewrite(ctx context.Context, query string) (pa
 		return nil, apierror.New(apierror.TypeBadData, DecorateWithParamName(err, "query").Error())
 	}
 
-	mapper := astmapper.NewProm2RangeCompat(ctx)
-	rewritten, err := mapper.Map(expr)
+	mapper := astmapper.NewProm2RangeCompat()
```
charleskorn: Could we create the mapper once and reuse it for all `rewrite` calls now?

pkg/frontend/querymiddleware/prune.go
```diff
@@ -63,8 +63,8 @@ func (p *pruneMiddleware) pruneQuery(ctx context.Context, query string) (string,
 	}
 	origQueryString := expr.String()
 
-	mapper := astmapper.NewQueryPruner(ctx, p.logger)
-	prunedQuery, err := mapper.Map(expr)
+	mapper := astmapper.NewQueryPruner(p.logger)
```
charleskorn: Can we create this mapper once and reuse it for all requests now?

pkg/frontend/querymiddleware/astmapper/subtree_folder.go
```diff
@@ -21,7 +23,11 @@ func newSubtreeFolder() ASTMapper {
 }
 
 // MapExpr implements ExprMapper.
-func (f *subtreeFolder) MapExpr(expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
+func (f *subtreeFolder) MapExpr(ctx context.Context, expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
+	if err := ctx.Err(); err != nil {
+		return nil, false, err
+	}
+
```
charleskorn: Why is this change needed?

zenador:
Do we need a changelog for this since it's more like a refactoring?

charleskorn:
> Do we need a changelog for this since it's more like a refactoring?

I would say there's no need for a changelog entry.

PR #12244, opened 2025-07-30
refactor: use reflection for exported metrics validation

dimitarvdimitrov:
Replace hardcoded metric validation with reflection-based approach that automatically discovers available metrics from validation.Limits struct yaml tags.

## Why

The current system requires maintaining a hardcoded list of allowed metric names and manually adding if statements for each metric. This creates maintenance overhead when adding new exportable metrics.

## What

- Remove hardcoded `allowedMetricNames` slice
- Add reflection-based field discovery using yaml tags  
- Support all relevant types: int, int64, uint, uint64, float64, bool, and model.Duration

## How

The new system uses reflection to:
1. Discover all available metrics by scanning yaml tags on validation.Limits struct fields
2. Validate enabled metrics against actual struct fields at startup
3. Generate getter functions dynamically for each enabled metric
4. Convert field values to float64 using Go's built-in type conversion

Adding new exportable metrics now only requires adding a yaml tag to the struct field.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail when experimental ingest storage is enabled. #12090
 * [ENHANCEMENT] Querier: Add native histogram definition to `cortex_bucket_index_load_duration_seconds`. #12094
 * [ENHANCEMENT] Query-frontend: Allow users to set the `query-frontend.extra-propagated-headers` flag to specify the extra headers allowed to pass through to the rest of the query path. #12174
+* [ENHANCEMENT] Overrides-exporter: The overrides-exporter can now export arbitrary fields from the limits configuration. Metric names are automatically discovered from YAML tags in the limits struct, eliminating the need to maintain hardcoded lists when adding new exportable metrics. #12345
```
```suggestion
* [ENHANCEMENT] Overrides-exporter: The overrides-exporter can now export arbitrary fields from the limits configuration. Metric names are automatically discovered from YAML tags in the limits struct, eliminating the need to maintain hardcoded lists when adding new exportable metrics. #12244
```

```suggestion
* [ENHANCEMENT] Overrides-exporter: The overrides-exporter can now export arbitrary fields from the limits configuration. Metric names are automatically discovered from YAML tags in the limits struct, eliminating the need to maintain hardcoded lists when adding new exportable metrics. #12244
```

pkg/util/validation/exporter/exporter.go
```diff
@@ -105,17 +86,30 @@ func (c *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
...
 	}
+
+	// Use reflection to validate that enabled metrics correspond to actual struct fields
+	limitsType := reflect.TypeOf(validation.Limits{})
+	defaultLimits := &validation.Limits{} // Create a zero-value instance for type checking
+
 	for _, metricName := range c.EnabledMetrics {
-		if !util.StringsContain(allowedMetricNames, metricName) {
-			return fmt.Errorf("enabled-metrics: unknown metric name '%s'", metricName)
+		field, found := getFieldByYamlTag(limitsType, metricName)
```
56quarters: We're looping over every enabled metric/limit and then looping over each field in the limits struct looking for the right field. Could we build a mapping once instead of having to loop each time?

pkg/util/validation/exporter/exporter.go
```diff
@@ -180,79 +174,99 @@ func NewOverridesExporter(
...
+				// Create a getter function for this field using a closure that captures fieldName
+				exportedMetrics = append(exportedMetrics, ExportedMetric{
+					Name: metricName,
+					Get: func(fName string) func(*validation.Limits) float64 {
+						return func(limits *validation.Limits) float64 {
+							limitsValue := reflect.ValueOf(limits).Elem()
+							fieldValue := limitsValue.FieldByName(fName)
+							value, err := convertToFloat64(fieldValue)
+							if err != nil {
+								// Log error and return 0.0 as fallback to avoid breaking metrics collection
```
56quarters: Doesn't look like this is actually being logged.

pkg/util/validation/exporter/exporter.go
```diff
@@ -180,79 +174,99 @@ func NewOverridesExporter(
...
+}
+
+func setupExportedMetrics(enabledMetrics *util.AllowList, extraMetrics []ExportedMetric) []ExportedMetric {
+	var exportedMetrics []ExportedMetric
+	limitsType := reflect.TypeOf(validation.Limits{})
+
+	// Get all possible metric names from struct yaml tags and check if each is enabled
+	allMetricNames := getAllowedMetricNames()
+	for _, metricName := range allMetricNames {
+		if enabledMetrics.IsAllowed(metricName) {
```
56quarters, line 266: It seems like this is doing a lot of the same work the `Config.Validate()` method is doing. Could we put logic along with the validation into its own struct to avoid some duplicate code and logic?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail when experimental ingest storage is enabled. #12090
 * [ENHANCEMENT] Querier: Add native histogram definition to `cortex_bucket_index_load_duration_seconds`. #12094
 * [ENHANCEMENT] Query-frontend: Allow users to set the `query-frontend.extra-propagated-headers` flag to specify the extra headers allowed to pass through to the rest of the query path. #12174
+* [ENHANCEMENT] Overrides-exporter: The overrides-exporter can now export arbitrary fields from the limits configuration. Metric names are automatically discovered from YAML tags in the limits struct, eliminating the need to maintain hardcoded lists when adding new exportable metrics. #12244
```
```suggestion
* [ENHANCEMENT] Overrides-exporter: The overrides-exporter can now export arbitrary fields from the limits configuration. Metric names are automatically discovered from YAML tags in the limits structure, eliminating the need to maintain hardcoded lists when adding new exportable metrics. #12244
```

tacole02 (approved):
Docs look good! One tiny nit in the changelog.

pkg/util/validation/exporter/exporter.go
dimitarvdimitrov: i addressed this along with the other comment for using  map in https://github.com/grafana/mimir/pull/12244/commits/f8438093fa29ee2707ea0dd260ab0e03e6a925ca. WDYT?

pkg/util/validation/exporter/exporter.go
```diff
@@ -180,79 +162,132 @@ func NewOverridesExporter(
...
+		exportedMetrics []ExportedMetric
+		fieldRegistry   = newLimitsFieldRegistry()
+	)
+
+	// Get all possible metric names from global registry and check if each is enabled
+	for _, metricName := range fieldRegistry.GetAllowedMetricNames() {
+		if enabledMetrics.IsAllowed(metricName) {
+			field, found := fieldRegistry.GetField(metricName)
+			if !found {
+				panic("couldn't find fields that fields registry returned, this shouldn't happen")
```
56quarters: Can we include the field name that triggered the error here?

56quarters (approved):
LGTM, thanks for making those changes!

pkg/util/validation/exporter/exporter.go
dimitarvdimitrov:
```suggestion
				panic(fmt.Sprintf("couldn't find fields that the fields registry returned, this shouldn't happen (metric name: %s)", metricName))
```

PR #12245, opened 2025-07-30
ingester: provide noop lookup planner

dimitarvdimitrov:
#### What this PR does

i'm not 100% convinced that the naive default planner will always do a good job. That was added as a suggestion upstream

https://github.com/grafana/mimir/blob/b8396cd92300d053a58fd1e0085da98e5f92aee6/vendor/github.com/prometheus/prometheus/tsdb/index/lookup.go#L74-L117

this configurability was added today in https://github.com/grafana/mimir/pull/12237

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -58,6 +58,7 @@ import (
 	"github.com/grafana/mimir/pkg/ingester/activeseries"
 	asmodel "github.com/grafana/mimir/pkg/ingester/activeseries/model"
 	"github.com/grafana/mimir/pkg/ingester/client"
+	"github.com/grafana/mimir/pkg/ingester/lookupplan"
 	"github.com/grafana/mimir/pkg/mimirpb"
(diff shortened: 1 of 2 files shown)


pkg/ingester/ingester.go
```diff
@@ -2727,6 +2728,17 @@ func (i *Ingester) getTSDB(userID string) *userTSDB {
 	return db
 }
 
+// getIndexLookupPlanner returns the appropriate index lookup planner based on configuration.
+// When index lookup planning is enabled, it uses the upstream ScanEmptyMatchersLookupPlanner
+// which collects statistics and can defer some vector selector matchers to sequential scans.
```
dimitarvdimitrov:
```suggestion
// which can defer some vector selector matchers to sequential scans. Later we will replace with our own planner.
```

chencs (approved):

PR #12248, opened 2025-07-30
index planning: implement cost model planner 

dimitarvdimitrov:
This is a follow-up of https://github.com/grafana/mimir/pull/12199

This PR implements the cost-based planner that uses the plan infrastructure introduced in the previous PR. The planner generates all possible combinations of index vs scan predicate assignments and selects the plan with the lowest estimated cost.

### How it works

The planner works by generating all possible plans and picking the one with the lowest cost. For each set of matchers, it:

  1. Converts matchers to plan predicates with statistics
  2. Generates all 2^n combinations of index/scan assignments
  3. Calculates total cost for each plan (index lookup + intersection + filter costs)
  4. Returns the cheapest plan

The planner aborts early if there are more than 10 matchers to avoid exponential explosion.

### Testing

The tests follow the same CSV-based pattern as https://github.com/grafana/mimir/pull/12197 and https://github.com/grafana/mimir/pull/12199. In this PR I moved the statistics from `predicate_test.go` into their own file (`mock_statistics_test.go`) so the test files don't grow so fast.

### Which issue(s) this PR fixes or relates to


pkg/ingester/lookupplan/mock_statistics_test.go
```diff
@@ -0,0 +1,301 @@
...
+	total := uint64(0)
+	for _, value := range values {
+		if seriesCount, exists := labelValues[value]; exists {
+			total += seriesCount
+		}
+	}
+	return total, nil
+}
+
+// newHighCardinalityMockStatistics creates a mock statistics with higher cardinality
```
chencs:
```suggestion
// newHighCardinalityMockStatistics creates a mockStatistics with higher cardinality
```

pkg/ingester/lookupplan/mock_statistics_test.go
```diff
@@ -0,0 +1,301 @@
...
+	hostTypes := []string{"web", "api", "worker", "db", "cache", "monitor"}
+
+	seriesPerInstance := uint64(50) // Each instance has ~50 series
+	instanceCount := 0
+
+	for _, region := range regions {
+		for _, az := range azs {
+			for _, hostType := range hostTypes {
+				for i := 1; i <= 200; i++ { // 200 instances per host type per AZ
+					instanceName := fmt.Sprintf("%s-%s-%s%s-%03d", hostType, region, az, az, i)
```
chencs: Why two `az`s?

pkg/ingester/lookupplan/mock_statistics_test.go
```diff
@@ -0,0 +1,301 @@
...
+		if i >= len(deployments) {
+			deployment = fmt.Sprintf("%s-%d", deployment, i/len(deployments))
+		}
+
+		// Each deployment has 1-5 replicas
+		replicas := (i % 5) + 1
+		for j := 0; j < replicas; j++ {
+			// Generate random-looking pod suffix (simulating Kubernetes pod names)
+			podName := fmt.Sprintf("%s-%s-%s", deployment,
+				generateRandomString(10), generateRandomString(5))
```
chencs: Perhaps a nit, but I think `generateRandomString(10)` here would normally be the same for all pods running the same version, right? Should it be the same across multiple pods of the same `deployment`?

pkg/ingester/lookupplan/planner.go
```diff
@@ -0,0 +1,124 @@
...
+	noopPlan, err := newScanOnlyPlan(ctx, matchers, p.stats)
+	if err != nil {
+		return nil, fmt.Errorf("error generating index lookup plan: %w", err)
+	}
+	allPlans := make([]plan, 0, 1<<uint(len(matchers)))
+
+	return generatePredicateCombinations(allPlans, noopPlan, 0), nil
+}
+
+func generatePredicateCombinations(plans []plan, currentPlan plan, decidedPredicates int) []plan {
```
chencs, line 90: I would definitely add some documentation here describing what `generatePredicateCombinations` does. I also think we assume that any plan with at least _an_ index matcher is enough for us to return a correct result set, so we should probably say it here, where we generate all possible plans.

pkg/ingester/lookupplan/planner_test.go
```diff
@@ -0,0 +1,224 @@
...
+			allActualMatchers = append(allActualMatchers, matchersStrings(result.IndexMatchers())...)
+			allActualMatchers = append(allActualMatchers, matchersStrings(result.ScanMatchers())...)
+			assert.ElementsMatch(t, matchersStrings(tc.inputMatchers), allActualMatchers)
+
+			testCases[tcIdx].expectedIndexMatchers = result.IndexMatchers()
+			testCases[tcIdx].expectedScanMatchers = result.ScanMatchers()
+		})
+	}
+
+	t.Run("error_statistics", func(t *testing.T) {
```
chencs, line 88: I think testing that plan generation will error on mocked `Statistics` errors is somewhat nonsensical, but for now it makes sense since we don't have any real `Statistics`. This is just a note for me that once we have real `Statistics` implementations, we should properly test this with input data that would error, rather than mocking it out.

As it is currently mocked, I don't think we really _need_ to check it against every `testCase`, since we know every single one will result in the exact same behavior from `errorMockStatistics`.

pkg/ingester/lookupplan/planner_test.go
```diff
@@ -0,0 +1,224 @@
...
+	metrics := NewMetrics(nil)
+	planner := NewCostBasedPlanner(metrics, stats)
+
+	const writeOutNewResults = false
+	if writeOutNewResults {
+		t.Cleanup(func() { data.WriteTestCases(t, testCases) })
+	}
+
+	for tcIdx, tc := range testCases {
+		t.Run(fmt.Sprintf("%s", tc.inputMatchers), func(t *testing.T) {
```
chencs: Same feedback as in https://github.com/grafana/mimir/pull/12199#discussion_r2271111244, I think each test should clearly describe intent rather than having the reader interpret intent.

pkg/ingester/lookupplan/mock_statistics_test.go
dimitarvdimitrov: go figure. fixed in https://github.com/grafana/mimir/pull/12248/commits/fc24759c0f0b2472ccacdb48d8a8b711ad2a8347

pkg/ingester/lookupplan/mock_statistics_test.go
dimitarvdimitrov: yeah sure 👍 that'd be more realistic. fixed in 3c7dcc85a1082fccc73ea698bee66b998da4c511

pkg/ingester/lookupplan/planner.go
dimitarvdimitrov: now that I think about it, technically all plans are valid. you can treat the TSDB as an unindexed document store and just retrieve all series and sequentially scan them.

But PostingsForMatchers wouldn't like that - it would return no postings and we won't have anything to filter out. Just to be on the safe side, I'll add some safeguards in the planner against this. So that we pick the cheapest plan with at least one index matcher.

pkg/ingester/lookupplan/planner_test.go
dimitarvdimitrov: i think there's basically no error cases in this code so far. so any error would come from the underlying statistics.

but it will probably be better if we test a bit more around the possible places which can give an error. Right now we're only covering the very basic case of only the LabelValues count. This is the test coverage of the `error_statistics` test case

 
Screenshot 2025-08-13 at 18 16 33


would be nice to add some more test cases once we know where we're likely to get errors. Also maybe not error on the first predicate exactly?

pkg/ingester/lookupplan/planner_test.go
dimitarvdimitrov: ✔️  in https://github.com/grafana/mimir/pull/12248/commits/78ee9c43432a0a38f3bd2835264696b7e76cbf74

pkg/ingester/lookupplan/planner.go
dimitarvdimitrov: Actually `PostingsForMatchers` doesn't support passing it no index matchers. When it gets no matchers it would just return no postings and we would return an incorrect result. I've added a guard against this in this commit 6c40d93e26d382fab402f815be80de966ef297fb

PR #12255, opened 2025-07-31
Alertmanager: Use the current configuration to start per-tenant Alermanagers after a request

santihernandezc:
When strict initialization is enabled, per-tenant Alertmanagers can be skipped if they have default configurations. We initialize these Alertmanagers using the fallback config in case we receive a request for a tenant.

Some requests might require the Alertmanager to use the actual configuration we have stored for the tenant. E.g., a tenant using a default configuration might want to try an email contact point before saving it. In this case, we need any custom SMTP configs to be loaded before we send the test notification.

This PR loads the configurations (Mimir, Grafana) from the database and uses them to start the Alertmanager for a tenant sending requests.

#### Note for reviewers

I'm using `fmt.Errorf` instead of `errors.Wrap`. The former is a function from the standard library; the latter is part of a 3rd party package ([pkg/errors](https://github.com/pkg/errors)) that was archived in 2021. We should stop using it.


pkg/alertmanager/config.go
```diff
@@ -80,8 +80,37 @@ func createUsableGrafanaConfig(logger log.Logger, gCfg alertspb.GrafanaAlertConf
...
+	for _, desc := range t {
+		result = append(result, definition.PostableApiTemplate{
+			Name:    desc.Filename,
+			Content: desc.Body,
+			Kind:    kind,
+		})
+	}
+	return result
+}
+
```
santihernandezc: None of this was changed. I only extracted the logic to create an `EmailSenderConfig` into a separate `genEmailConfig` function.

PR #12257, opened 2025-07-31
Elide SeriesChunksStreamReader..StartBuffering span on queries

bboreham:
#### What this PR does

To save tracing bandwidth and screen real-estate when viewing traces, move the start and end of buffering (and any log messages) up to events on the parent `cortex.Ingester/QueryStream` span.

To illustrate, this picture would go from 12 spans to 9:

image

It could be argued that the parent span belongs to otelgrpc and it's confusing to add streamreader stuff to it.
I think the benefits of fitting more on the screen outweigh that.  Indeed I'd quite like to remove the `cortex.Ingester/QueryStream` client-side span because it doesn't add anything to its parent `Distributor.queryIngesterStream`.

#### Checklist


bboreham:
> Can we log also when we stop buffering? 

There is already an event: https://github.com/grafana/mimir/blob/d7dde5ab5c47ed75e6aa28f8c5732d93d9b5e1eb/pkg/ingester/client/streaming.go#L165

image


That's logged on every successful return from `readStream`, which is the last thing that happens, or an error is logged by `StartBuffering`.

PR #12261, opened 2025-07-31
querier: check size of whole frontend response before sending it

56quarters:
#### What this PR does

Check the size of the entire response being sent to the query-frontend before sending it, not just the body. This fixes edge cases where a response body wasn't larger than the grpc limit but the entire request was. This resulted in timeouts in the query-frontend since the scheduler processor was never able to sent a response to the query-frontend.

This is the same issue as #10154 but in the querier worker that pulls from the query-scheduler instead of the query-frontend.

Related #10154

Required for #12200

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 * [BUGFIX] Querier: Samples with the same timestamp are merged deterministically. Previously, this could lead to flapping query results when an out-of-order sample is ingested that conflicts with a previously ingested in-order sample's value. #8673
 * [BUGFIX] Store-gateway: Fix potential goroutine leak by passing the scoped context in LabelValues. #12048
 * [BUGFIX] Distributor: Fix pooled memory reuse bug that can cause corrupt data to appear in the err-mimir-label-value-too-long error message. #12048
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * [BUGFIX] Querier: Samples with the same timestamp are merged deterministically. Previously, this could lead to flapping query results when an out-of-order sample is ingested that conflicts with a previously ingested in-order sample's value. #8673
 * [BUGFIX] Store-gateway: Fix potential goroutine leak by passing the scoped context in LabelValues. #12048
 * [BUGFIX] Distributor: Fix pooled memory reuse bug that can cause corrupt data to appear in the err-mimir-label-value-too-long error message. #12048
+* [BUGFIX] Querier: fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #12261
```
```suggestion
* [BUGFIX] Querier: Fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #12261
```

tacole02 (approved):
Changelog looks good!

Issue #12253, opened 2025-07-31
Remove instant query splitting feature


armandgrillet:
https://github.com/grafana/mimir/pull/2469 was added 3 years ago but causes correctness issues such as:

- https://github.com/grafana/mimir/issues/2590
- https://github.com/grafana/mimir/issues/2599

We are working on replacing this by a new implementation in the new query engine, and as such can remove this experimental feature.

PR #2634, opened 2022-08-02
Add instant split interval to HTTP header options

ssncferreira:
#### What this PR does

Add instant split interval to HTTP header options.

Conflicts with PR https://github.com/grafana/mimir/pull/2633

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2633, opened 2022-08-02
Move split instant queries by interval to per-tenant config

ssncferreira:
#### What this PR does

Move split instant queries by interval from query-frontend config to a per-tenant config

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #2469, opened 2022-07-19
Add instant query split by range basic structure POC

ssncferreira:
#### What this PR does

Add split by range on instant queries.

Note: this is a WIP and is not production ready.

#### Which issue(s) this PR fixes or relates to

Fixes https://github.com/grafana/mimir-squad/issues/520

#### Checklist

- [ ] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #12267, opened 2025-08-01
Remove instant query splitting feature

charleskorn:
#### What this PR does

This PR removes the experimental instant query splitting feature. As discussed in #12253, there are a couple of correctness issues with the existing implementation and we're working on a new implementation in MQE.

Resolves https://github.com/grafana/mimir/issues/12253

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


lamida (commented):
The original PR for this was #2469. But on top of that there are two more PRs #2633 and  #2634.

This PR seems almost already cover most of things, but It I think we might want to revert also some stuffs from #2634 which is not already in this PR such on codec.go and model.proto.

charleskorn:
> The original PR for instant query splitting was #2469. But on top of that there are two more PRs #2633 and #2634.
> 
> This PR seems almost already cover most of things to revert #2469, but It I think we might want to revert also some stuffs from #2634 which is not already in this PR such on codec.go, model.proto, etc.

Good catch, I missed that - how does this look now?

lamida (approved):
image

🚀

dimitarvdimitrov (commented):
LGTM

PR #6191, opened 2023-10-02
Distributor write path: put httpgrpc.Errorf() calls at the topmost level

duricanikolic:
#### What this PR does
This PR introduces the following error types, reppresenting the problems on the distributor’s write path that need some special handling: 
-  `ReplicasNotMatch`, 
- `TooManyClusters`, 
- `ValidationError`, 
- `IngestionRateLimited`
- `RequestRateLimited`.
They all belong to the newly introduced package `distributorerror`.

Moreover, this PR refactors the distributor's write path in such a way that all the calls to `httpgrpc.Errorf()` have been moved to the topmost possible level,i.e., to:
- `push.handler()` and
- `distributor.Push()`.
The semantics of the distributor's write path has not been changed, i.e., the newly introduced errors are mapped to the same HTTP status codes their corresponding problems used to be assigned before this PR.

This refactoring is just a preliminary change needed for a bigger refactoring and improving of intester's and distributor's write path error handling.

#### Which issue(s) this PR fixes or relates to

Related to https://github.com/grafana/mimir/issues/1900, https://github.com/grafana/mimir/issues/6008

#### Checklist

- [x] Tests updated
- [ ] Documentation added
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

PR #12269, opened 2025-08-01
Update comments about retaining label strings

bboreham:
#6191 changed the expectations around errors and string re-use, but left some comments in place which said exactly the wrong thing.

Relates to #12266

### DIFF
--- a/pkg/distributor/distributor.go
+++ b/pkg/distributor/distributor.go
@@ -807,7 +807,7 @@ func (d *Distributor) checkSample(ctx context.Context, userID, cluster, replica 
 // validateSamples validates samples of a single timeseries and removes the ones with duplicated timestamps.
 // Returns an error explaining the first validation finding.
 // May alter timeseries data in-place.
-// The returned error may retain the series labels.
+// The returned error MUST NOT retain label strings - they point into a gRPC buffer which is re-used.
 func (d *Distributor) validateSamples(now model.Time, ts *mimirpb.PreallocTimeseries, userID, group string) error {
 	if len(ts.Samples) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


duricanikolic:
There are different occurrences of the following comment in [pkg/distributor/validate.go](https://github.com/grafana/mimir/blob/main/pkg/distributor/validate.go):
```
// The returned error may retain the provided series labels.
```
We need to update those comments too.
I would also put another comment stating that given that we use `fmt.Errorf()` the previous requirement is satisfied.

duricanikolic:
Related to [my previous comment](https://github.com/grafana/mimir/pull/12269#issuecomment-3143841184), and given that we actually use `extract.UnsafeMetricNameFromLabelAdapters()` only and exclusively in the places where we actually do **NOT** want an "unsafe" name, I am wondering if it maybe makes sense to replace  `extract.UnsafeMetricNameFromLabelAdapters()` with `extract.SsafeMetricNameFromLabelAdapters()`, and use that one:
```
// SafeMetricNameFromLabelAdapters extracts the metric name from a list of LabelPairs.
// The returned metric name string is a copy of the label value, and does not retain the series labels.
func UnsafeMetricNameFromLabelAdapters(labels []mimirpb.LabelAdapter) (string, error) {
	for _, label := range labels {
		if label.Name == model.MetricNameLabel {
			return fmt.Sprintf("%s", label.Value), nil
		}
	}
	return "", errNoMetricNameLabel
}
```

bboreham:
> We need to update those comments too.

Fixed, I think.

> I would also put another comment stating that given that we use fmt.Errorf() the previous requirement is satisfied.

I don't want to do that.  How far do you go to teach people a complex subject in source comments?

> we actually use extract.UnsafeMetricNameFromLabelAdapters() only and exclusively in the places where we actually do NOT want an "unsafe" name

This is not true; [example1](https://github.com/grafana/mimir/blob/aaae72d65d14af5f57be572d55603558017ed189/pkg/distributor/distributor.go#L1297), [example2](https://github.com/grafana/mimir/blob/aaae72d65d14af5f57be572d55603558017ed189/pkg/distributor/validate.go#L428).

Less important: we use `Unsafe` to mark the exceptions; a "safe" variant doesn't need special marking.

duricanikolic:
> > We need to update those comments too.
>
> Fixed, I think.

@bboreham no, you haven't :) The only modified source is `pkg/distributor/distributor.go`, while additional changes should be done in `pkg/distributor/validate.go`.

duricanikolic (approved):
LGTM
Thank you

PR #12270, opened 2025-08-01
feat(converter): monitor discovery and processing goroutines

jesusvazquez:
Before this change we were launching the goroutines without checking if they died
```
	go c.runBlockDiscovery(ctx)
	go c.runBlockProcessing(ctx)
```

I've modified the code slightly to monitor them so we can see panics in case they were failing silently before.

Also brings new parquet-common which has some performance improvements.

### DIFF
--- a/pkg/parquetconverter/parquet_converter.go
+++ b/pkg/parquetconverter/parquet_converter.go
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/thanos-io/objstore"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/grafana/mimir/pkg/storage/bucket"
 	mimir_tsdb "github.com/grafana/mimir/pkg/storage/tsdb"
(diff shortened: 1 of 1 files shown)


pkg/parquetconverter/parquet_converter.go
```diff
@@ -286,59 +287,90 @@ func newRingAndLifecycler(cfg RingConfig, logger log.Logger, reg prometheus.Regi
...
+
+		discoveryTicker := time.NewTicker(c.Cfg.DiscoveryInterval)
+		defer discoveryTicker.Stop()
+
+		for {
+			select {
+			case <-gCtx.Done():
+				return nil
+			case <-discoveryTicker.C:
+				func() {
```
npazosmendez, line 307: What's the purpose of this `func() {` ?

npazosmendez (approved):

pkg/parquetconverter/parquet_converter.go
jesusvazquez: The func() bit is essential because if not the underlying defer() check is attached to the outer function.

By having this func + defer we can catch and recover in the infinite select loop if discoverAndEnqueueBlocks() panics. Then outside of the infinite select there is another catch in case something else panics but the process would not restart.

jesusvazquez:
Closing in favor of https://github.com/grafana/mimir/pull/12367

PR #12271, opened 2025-08-01
Warn about unsafe mutable labels in type names and doc strings

tcard:
#### What this PR does

* Creates aliases for `string` (`unsafeMutableString`) and `mimirpb.LabelAdapter` (`UnsafeMutableLabel`) to try and make it more obvious when it's not safe to shallow-copy then, as it usually is with strings/structs of strings.
* Add notices in the doc comments of `PreallocTimeseries` and descendents about avoiding shallow copies.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/mimirpb/timeseries.go
```diff
@@ -309,7 +315,9 @@ func (p *PreallocTimeseries) MarshalToSizedBuffer(buf []byte) (int, error) {
 }
 
 // LabelAdapter is a labels.Label that can be marshalled to/from protos.
-type LabelAdapter labels.Label
+type LabelAdapter struct {
```
tcard, line 318: `LabelAdapter` is reused in too many places (and, for what I could find, safely so) to just rename it to `UnsafeMutableLabel`, hence the alias.

pkg/mimirpb/timeseries.go
```diff
@@ -503,6 +511,25 @@ func (bs *LabelAdapter) Compare(other LabelAdapter) int {
 	return strings.Compare(bs.Value, other.Value)
 }
 
+// UnsafeMutableLabel is an alias of LabelAdapter meant to highlight its unsafety.
+//
+// # DO NOT SHALLOW-COPY
+//
+// When an UnsafeMutableLabel is referenced from a PreallocTimeseries, the data it
+// references may change once the timeseries is returned to the shared pool.
+type UnsafeMutableLabel = LabelAdapter
```
colega, line 520: Do we really want to define this as a type alias? Why not defining a new type `type UnsafeMutableLabel LabelAdapter` so an explicit conversion would be needed?

tcard: Explained in [this misplaced comment](https://github.com/grafana/mimir/pull/12271#discussion_r2248213989):

> `LabelAdapter` is reused in too many places (and, for what I could find, safely so) to just rename it to `UnsafeMutableLabel`, hence the alias.

tcard: Actually that explains why not rename it, not why not create a new type.

I'll try the new type approach.

tcard: Discussed off-PR but will repeat here. After giving it a try I've abandonded this attempt. Usage of yoloStrings is way too entrenched and a separate type for them doesn't really significantly solve the actual issue, as a type doesn't prevent you from retaining references any more than an alias does.

This is a social rather than technical remediation in the hopes that devs read "unsafeMutable" and take notice, and for that purpose, an alias is just as good as a new type.

PR #12272, opened 2025-08-01
Alertmanager: Revisit computeConfig

santihernandezc:
This PR refactors the `computeConfig` method on the `MultitenantAlertmanager` struct and gives default Grafana configurations higher precedence than default Mimir configurations.

Reasoning: Each tenant gets a default Mimir configuration when it's created, but a tenant can't have a default Grafana configuration unless there's a Grafana instance sending a configuration (remote Alertmanager feature). Grafana configurations can also be deleted in case a tenant doesn't want to use them, but there's no way of deleting a default Mimir config and using the default Grafana one.

In other words, having a default Grafana config is always intentional, so it should have higher precedence.

This PR was broken down into:
- https://github.com/grafana/mimir/pull/12293
- https://github.com/grafana/mimir/pull/12289

### DIFF
--- a/pkg/alertmanager/multitenant.go
+++ b/pkg/alertmanager/multitenant.go
@@ -781,7 +781,7 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 	// Unpromoted/empty Grafana configurations are always ignored.
 	if !cfgs.Grafana.Promoted || cfgs.Grafana.RawConfig == "" {
 		// Only return the default Mimir config (lowest precedence) if the tenant is receiving requests.
-		if am.isReceivingRequests(cfgs.Mimir.User) {
+		if am.isTenantActive(cfgs.Mimir.User) {
 			return amConfigFromMimirConfig(cfgs.Mimir, am.cfg.ExternalURL.URL), true, nil
 		}
 		return amConfig{}, false, nil
@@ -794,8 +794,8 @@ func (am *MultitenantAlertmanager) computeConfig(cfgs alertspb.AlertConfigDescs)
 		return cfg, true, err
 	}
 
-	// We have no custom configs. Check the last request time to determine whether to start the AM.
-	if !am.isReceivingRequests(cfgs.Mimir.User) {
+	// We have no custom configs. Check the last activity time to determine whether to start the AM.
+	if !am.isTenantActive(cfgs.Mimir.User) {
 		return amConfig{}, false, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/multitenant.go
```diff
@@ -760,62 +760,82 @@ func (am *MultitenantAlertmanager) syncConfigs(ctx context.Context, cfgMap map[s
...
 	}
+}
 
-	if !isMimirConfigCustom {
-		level.Debug(am.logger).Log("msg", "using grafana config with the default globals", "user", userID)
-		cfg, err := am.amConfigFromGrafanaConfig(cfgs.Grafana)
-		return cfg, true, err
+// isReceivingRequests checks the last request time for a tenant, returning 'true' if the grace period has not yet expired.
+// If strict initialization is not enabled, it always returns true, as we don't keep track of request times.
+func (am *MultitenantAlertmanager) isReceivingRequests(userID string) bool {
```
yuri-tceretian: This name does not seem to describe well its implementation. Let's pick more generic. What about `isTenantActive`?

Or better reverse the result and rename it to `isTenantInactive`. This will improve readability of usages

pkg/alertmanager/multitenant.go
```diff
@@ -760,62 +760,82 @@ func (am *MultitenantAlertmanager) syncConfigs(ctx context.Context, cfgMap map[s
...
+	if !am.cfg.StrictInitializationEnabled {
+		return true
 	}
 
-	level.Warn(am.logger).Log("msg", "merging configurations not implemented, using mimir config", "user", userID)
-	return cfg, true, nil
-}
+	createdAt, loaded := am.lastRequestTime.LoadOrStore(userID, time.Time{}.Unix())
```
yuri-tceretian: var name should be lastRequestTime

pkg/alertmanager/multitenant.go
```diff
@@ -760,62 +760,82 @@ func (am *MultitenantAlertmanager) syncConfigs(ctx context.Context, cfgMap map[s
...
+	// If the value stored is not what we have in memory, the tenant received a request since the last read.
+	gracePeriodExpired := time.Since(time.Unix(createdAt.(int64), 0)) >= am.cfg.GrafanaAlertmanagerIdleGracePeriod
+	if gracePeriodExpired && am.lastRequestTime.CompareAndSwap(userID, createdAt, time.Time{}.Unix()) {
+		return false
+	}
 
-// isGrafanaConfigUsable returns true if the Grafana configuration is promoted, non-default, and not empty.
-func isGrafanaConfigUsable(cfg alertspb.GrafanaAlertConfigDesc) bool {
-	return cfg.Promoted && !cfg.Default && cfg.RawConfig != ""
+	level.Debug(am.logger).Log("msg", "user has no usable config but is receiving requests, keeping Alertmanager active", "user", userID)
```
yuri-tceretian, line 837: I wonder if we need to see this log message every time we read config from the store and try to apply it?

PR #12284, opened 2025-08-04
Distributor: Support disabling of escaped metric/label names in OTLP endpoint

aknuds1:
#### What this PR does

Add flag -distributor.otel-unescaped-names, to support disabling of escaped metric/label names in the distributor's OTLP endpoint.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


integration/otlp_ingestion_test.go
```diff
@@ -94,7 +116,12 @@ func testOTLPIngestion(t *testing.T, enableSuffixes bool) {
 		labels.MustNewMatcher(labels.MatchEqual, "user", "user-1"))))
 
 	// Query the series.
-	result, err := c.Query(fmt.Sprintf("series_1%s", sfx), now)
+	metricQuery := convertedMetricName
```
juliusmh: [nit] I think the conditional is not required here, we can always just query using the _new_ syntax.

```suggestion
	metricQuery := fmt.Sprintf(`{%q}`, convertedMetricName)
```

Internally, promql parses `test{}` and `{"test"}` equally as `{"__name__"="test"}`.

juliusmh:
I'm wondering if we should check that validationScheme is set to utf8 when enabling the "no-escape" mode. A warning during `limits.validate` if flags are incompatible would point users in the right direction when enabling no-escape mode (who forgot to set the validation scheme).

integration/otlp_ingestion_test.go
aknuds1: I made the query definitions unconditional.

aknuds1:
@juliusmh I addressed your feedback. PTAL.

PR #12288, opened 2025-08-04
Native histogram definitions for cortex_query_{frontend|scheduler}_queue_duration_seconds

juliusmh:
#### What this PR does
Add native histogram definitions to
- `cortex_query_scheduler_queue_duration_seconds`
- `cortex_query_frontend_queue_duration_seconds`

with the following parameters:
```go
	NativeHistogramBucketFactor:     1.1,
	NativeHistogramMinResetDuration: 1 * time.Hour,
	NativeHistogramMaxBucketNumber:  100,
```

Part of: https://github.com/grafana/mimir-squad/issues/3077

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [CHANGE] Query-frontend: Remove experimental instant query splitting feature. #12267
 * [FEATURE] Distributor, ruler: Add experimental `-validation.name-validation-scheme` flag to specify the validation scheme for metric and label names. #12215
 * [FEATURE] Distributor: Add experimental `-distributor.otel-translation-strategy` flag to support configuring the metric and label name translation strategy in the OTLP endpoint. #12284 #12306
+* [ENHANCEMENT] Query-scheduler/query-frontend: Add native histogram definitions to `cortex_query_{scheduler|frontend}_queue_duration_seconds`. #12288
 * [ENHANCEMENT] Compactor: Add `-compactor.update-blocks-concurrency` flag to control concurrency for updating block metadata during bucket index updates, separate from deletion marker concurrency. #12117
 * [ENHANCEMENT] Stagger head compaction intervals across zones to prevent compactions from aligning simultaneously, which could otherwise cause strong consistency queries to fail when experimental ingest storage is enabled. #12090
 * [ENHANCEMENT] Querier: Add native histogram definition to `cortex_bucket_index_load_duration_seconds`. #12094
--- a/pkg/frontend/v1/frontend_test.go
+++ b/pkg/frontend/v1/frontend_test.go
@@ -223,6 +223,37 @@ func TestFrontendMetrics(t *testing.T) {
 
 		assert.Equal(t, "Hello World", string(body))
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * [CHANGE] Query-frontend: Remove the CLI flag `-query-frontend.downstream-url` and corresponding YAML configuration and the ability to use the query-frontend to proxy arbitrary Prometheus backends. #12191
 * [CHANGE] Query-frontend: Remove experimental instant query splitting feature. #12267
 * [FEATURE] Distributor, ruler: Add experimental `-validation.name-validation-scheme` option to specify the validation scheme for metric and label names. #12215
+* [ENHANCEMENT] Query-scheduler/query-frontend: Add native histogram definition to `cortex_query_{scheduler|frontend}_queue_duration_seconds`. #12288
```
aknuds1: [Nit]
```suggestion
* [ENHANCEMENT] Query-scheduler/query-frontend: Add native histogram definitions to `cortex_query_{scheduler|frontend}_queue_duration_seconds`. #12288
```

pkg/frontend/v1/frontend_test.go
```diff
@@ -198,7 +199,7 @@ func TestFrontendCancel(t *testing.T) {
 	testFrontend(t, defaultFrontendConfig(), handler, test, nil, nil)
 }
 
-func TestFrontendMetricsCleanup(t *testing.T) {
+func TestFrontendMetrics(t *testing.T) {
```
aknuds1, line 202: Why is the test name changing?

aknuds1 (commented):
Looks good AFAICT, except for a changelog nit, and I'm not sure you should be changing the test name since I think its topic is cleanup.

@krajorama should review the native histogram definitions.

pkg/frontend/v1/frontend_test.go
juliusmh: Because this test is not just testing the `Cleanup` but also other parts of metrics.

aknuds1: I can see though that the test is for the inactive user metrics cleanup scenario (`fr.cleanupInactiveUserMetrics("1")`), and verifying its effect. Therefore, the original test name seems intentional to me. I don't think we should change it. Maybe you should add another test function instead of modifying the intended scenario of this one?

krajorama (approved):
Native histogram definitions ok. Follows the recommended settings and histogram don't lose classic buckets due to explicitly set boundaries as well.

pkg/frontend/v1/frontend_test.go
juliusmh: I pushed and github doesn't show this properly, I left the `Cleanup` test intact and created a new case just to verify the histogram metrics. 
- TestFrontendMetrics (new)
- TestFrontendMetricsCleanup

aknuds1: Thanks @juliusmh!

tacole02 (approved):
Changelog looks good!

PR #12292, opened 2025-08-04
Remove unsafe from mimirpb.CopyLabels

tcard:
#### What this PR does

Part of a bigger effort to contain the unsafety within mimirpb.

`CopyLabels` can be implemented in a safe manner with the same memory optimization by using a StringBuilder rather than a `[]byte` and `yoloString`.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/compat_slice.go
+++ b/pkg/mimirpb/compat_slice.go
@@ -8,6 +8,7 @@
 package mimirpb
 
 import (
+	"strings"
 	"unsafe"
 
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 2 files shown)


pkg/mimirpb/compat_slice.go
```diff
@@ -46,27 +47,20 @@ func CopyLabels(input []labels.Label) labels.Labels {
...
+	allStrings := b.String()
 
-// Copies string to buffer (which must be big enough), and converts buffer slice containing
-// the string copy into new string.
-func copyStringToBuffer(in string, buf []byte) (string, []byte) {
-	l := len(in)
-	c := copy(buf, in)
-	if c != l {
-		panic("not copied full string")
+	for i, l := range input {
```
colega: Now we're looping twice though, however, we also don't run this code anymore (this is `slicelabels` implementation, and we use `stringlabels` tag).

tcard: Actually it can be done in a single loop, pretty much in the same way as before: [6cc1eea](https://github.com/grafana/mimir/pull/12292/commits/6cc1eea59a08a2feead024d550ce76e2933d11e0)

> (this is slicelabels implementation, and we use stringlabels tag).

[TIL](https://prometheus.io/blog/2023/03/21/stringlabel/), oh well.

PR #12294, opened 2025-08-04
test(otlp): test start time handling end 2 end

krajorama:
Prepare for the rewrite of the OTLP translator upstream in https://github.com/prometheus/prometheus/pull/16951 by adding an end-to-end test to check for regressions.

Related to https://github.com/grafana/mimir/pull/12152.

Also switch from raw JSON in the test to building the payload in code and marshaling to protobuf.


krajorama:
folding into #12652

PR #12295, opened 2025-08-04
feat(parquet querier): Introduce querier mode for querying parquet blocks

jesusvazquez:
This PR reintroduces the approach we had at the beginning for  querying the blocks from the queriers in https://github.com/grafana/mimir/tree/parquet-mimir but now using the upstream library and with a lot of inspiration from the work our peers at cortex did for managing the query fallback to store-gateways if some blocks have not been converted yet.

The idea is to test this approach and compare it to the results we're getting from the work we're doing in the store-gateways.


dimitarvdimitrov:

PR #12302, opened 2025-08-05
Remote execution of query plans in queriers

charleskorn:
#### What this PR does

This PR adds support for remote execution of query plans in queriers.

Query-frontends can now send a Protobuf message (instead of a smuggled HTTP request) to queriers via query-schedulers, and queriers will respond with a stream of results. This has been implemented in a way where we can add other kinds of query requests and responses in the future, with the goal of eventually moving away from smuggling HTTP requests in Protobuf messages.

In the case of queries, query-frontends can now send a query plan and request evaluation of a particular node in the plan, and queriers will respond with results for that node.

I've also added a HTTP endpoint to queriers for debugging purposes (`/api/v1/evaluate`), and a CLI tool to exercise this (`tools/evaluate-query-plan`). I plan to remove this in a subsequent PR, once I've added the functionality to query-frontends to send these requests to queriers.

Not included in this PR:
* batching of responses to query-frontends (each series' data is sent as its own Protobuf message for now)
* actually sending requests from query-frontends
* metrics and tracing for requests to queriers

There are a number of other items I'll handle in follow up PRs (eg. propagating strong consistency information and dealing with query stats) - my goal with this PR is to get the general structure in place and start getting feedback on the approach before I invest too much more time in this.

I've chosen not to add a changelog entry given this code is not actively used at present, but open to other opinions on this.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/scheduler/schedulerpb/scheduler.proto
```diff
@@ -38,7 +39,11 @@ message SchedulerToQuerier {
   // Query ID as reported by frontend. When querier sends the response back to frontend (using frontendAddress),
   // it identifies the query by using this ID.
   uint64 queryID = 1;
-  httpgrpc.HTTPRequest httpRequest = 2;
+
+  oneof payload {
```
charleskorn, line 43: Note to reviewers: I've confirmed that this change is backwards and forwards compatible (ie. something with the old definition without the `oneof` can read messages encoded by something with the new definition with the `oneof`, and vice versa). 

This is also confirmed in [the Protobuf docs](https://protobuf.dev/programming-guides/proto3/#wire-safe-changes): 

> Changing a single explicit presence field or extension into a member of a **new** `oneof` is safe.

pkg/scheduler/schedulerpb/scheduler.proto
```diff
@@ -83,7 +93,10 @@ message FrontendToScheduler {
 
   // Following are used by ENQUEUE only.
   string userID = 4;
-  httpgrpc.HTTPRequest httpRequest = 5;
+  oneof payload {
```
charleskorn, line 96: Note to reviewers: [same comment as above](https://github.com/grafana/mimir/pull/12302/files#r2255977246) - I've confirmed this is backwards and forwards compatible.

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Prometheus Authors.
```
lamida, line 4: Do we need this header?

tools/evaluate-query-plan/main.go
```diff
@@ -0,0 +1,199 @@
...
+	expr         string
+	tenantID     string
+	instantQuery bool
+
+	// Instant query:
+	at flagext.Time
+
+	// Range query:
+	from flagext.Time
+	to   flagext.Time
```
lamida: Nits, understand this tool will be removed as mentioned in the PR, but can we still have a similar nomenclature like what [prometheus API](https://prometheus.io/docs/prometheus/latest/querying/api/) do?

at --> time
from --> start
to --> end

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// ServeHTTP responds to requests made to the evaluation HTTP endpoint.
+// This is primarily used for debugging: most requests will arrive from query-frontends via
+// the query-scheduler over gRPC and therefore be handled by HandleProtobuf.
+//
+// Why do we call this method "ServeHTTP", and the other "HandleProtobuf"?
+// We want to satisfy the http.Handler interface, which requires a method called ServeHTTP,
+// and the querier/worker.RequestHandler interface that uses the verb "handle"
+// in its method name, so we've made the querier/worker.ProtobufRequestHandler interface use "handle" as well.
```
lamida: Nits, let's put comment relevant to this method only. The detail of of HandleProtobuf comment should be put above that method. I am suggesting the HandleProtobuf comment to be added above that method instead. See my another comment.
```suggestion
// ServeHTTP responds to requests made to the evaluation HTTP endpoint.
// This method satisfy the http.Handler interface.
//
// This is primarily used for debugging: most requests will arrive from query-frontends via
// the query-scheduler over gRPC and therefore be handled by HandleProtobuf.
```

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+			stream: stream,
+			logger: d.logger,
+		}
+		d.evaluateQuery(r.Context(), body, writer)
+	default:
+		http.Error(w, "unknown request type", http.StatusUnsupportedMediaType)
+	}
+}
+
+func (d *Dispatcher) HandleProtobuf(ctx context.Context, req *prototypes.Any, stream frontendv2pb.QueryResultStream) {
```
lamida: Nits, suggesting of adding the comment here instead of above ServeHTTP method above.
```suggestion

// HandleProtobuf responds to requests made to the evaluation protobuf endpoint.
// This method satisfy worker.ProtobufRequestHandler interface.
//
// The interface method is named HandleProtobuf to align with what worker.RequestHandler interface 
// method named Handle.
func (d *Dispatcher) HandleProtobuf(ctx context.Context, req *prototypes.Any, stream frontendv2pb.QueryResultStream) {
```

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+		writer := &queryResponseWriter{
+			stream: stream,
+			logger: d.logger,
+		}
+
+		d.evaluateQuery(ctx, req.Value, writer)
+
+	default:
+		writeErrorToStream(ctx, stream, mimirpb.QUERY_ERROR_TYPE_BAD_DATA, fmt.Sprintf("unknown query request type %q", req.TypeUrl), d.logger)
+		return
```
lamida: This return is redundant and can be removed.

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	return mimirpb.QUERY_ERROR_TYPE_EXECUTION
+}
+
+type httpResultStream struct {
+	w http.ResponseWriter
+}
+
+func (w *httpResultStream) Write(ctx context.Context, m *frontendv2pb.QueryResultStreamRequest) error {
```
lamida: Nits. Suggest to rename receiver 
w --> h
Suggest to rename second argument
m --> r
```suggestion
func (h *httpResultStream) Write(ctx context.Context, r *frontendv2pb.QueryResultStreamRequest) error {
```

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+
+	return nil
+}
+
+type queryResponseWriter struct {
+	stream frontendv2pb.QueryResultStream
+	logger log.Logger
+}
+
+func (w *queryResponseWriter) Write(ctx context.Context, m querierpb.EvaluateQueryResponse) error {
```
lamida:
```suggestion
func (w *queryResponseWriter) Write(ctx context.Context, r querierpb.EvaluateQueryResponse) error {
```

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+				RangeStart:  stepData.RangeStart,
+				RangeEnd:    stepData.RangeEnd,
+				Floats:      mimirpb.FromFPointsToSamples(floats),
+				Histograms:  mimirpb.FromHPointsToHistograms(histograms),
+			},
+		},
+	})
+}
+
+func combineSlices[T any](head, tail []T, pool *types.LimitingBucketedPool[[]T, T], memoryConsumptionTracker *limiter.MemoryConsumptionTracker) ([]T, bool, error) {
```
lamida: Returning this kind of bool flag is a bit strange and not obvious. What if we return some kind of cleanup function instead?


```suggestion
func combineSlices[T any](head, tail []T, pool *types.LimitingBucketedPool[[]T, T], memoryConsumptionTracker *limiter.MemoryConsumptionTracker) ([]T, func(), error) {
```

If no need to return thing to pool we can just do 
```
return tail, func(){}, nil // empty function cleanup
```

But if we reach the end of function where we have to return thing to pool we can do something like
```
return combined, func() {
		pool.Put(&combined, memoryConsumptionTracker)
	}, nil
```

Later in the caller we just do the defer of cleanup function
```
floats, cleanup, err := combineSlices(...)
if err != nil {
    return err
}
defer cleanup()
```

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+
+	protoType := params["proto"]
+	if protoType == "" {
+		http.Error(w, "missing proto parameter in Content-Type header", http.StatusBadRequest)
+		return
+	}
+
+	switch protoType {
+	case evaluateQueryRequestMessageName:
+		body, err := io.ReadAll(r.Body)
```
lamida, line 74: Should we limit the request size? Otherwise rogue user can send a very big request body?

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+
+	// TODO: batch up series / steps to return, rather than sending each step one at a time?
+
+	floatsHead, floatsTail := stepData.Floats.UnsafePoints()
+	floats, needToReturnFloats, err := combineSlices(floatsHead, floatsTail, types.FPointSlicePool, evaluator.MemoryConsumptionTracker)
+	if err != nil {
+		return err
+	}
+
+	if needToReturnFloats {
```
lamida: I was suggesting to use cleanup func(){} approach instead of return flag like this. The suggestion applies also for histogram pool below.

pkg/mimir/modules.go
```diff
@@ -614,11 +614,18 @@ func (t *Mimir) initQuerier() (serv services.Service, err error) {
 	t.Cfg.Worker.MaxConcurrentRequests = t.Cfg.Querier.EngineConfig.MaxConcurrent
 	t.Cfg.Worker.QuerySchedulerDiscovery = t.Cfg.QueryScheduler.ServiceDiscovery
 
+	var dispatcher *querier.Dispatcher
+
+	if t.Cfg.Querier.QueryEngine == querier.MimirEngine {
```
lamida, line 619: Do we want to have flag to allow user not enabling remote execution query plan, even if we use MQE?

pkg/querier/querier.go
```diff
@@ -182,15 +182,17 @@ func New(cfg Config, limits *validation.Overrides, distributor Distributor, quer
...
+	var streamingEngine *streamingpromql.Engine
 
 	switch cfg.QueryEngine {
 	case PrometheusEngine:
 		eng = promql.NewEngine(opts)
 	case MimirEngine:
 		limitsProvider := NewTenantQueryLimitsProvider(limits)
-		streamingEngine, err := streamingpromql.NewEngine(mqeOpts, limitsProvider, queryMetrics, planner, logger)
+		var err error
+		streamingEngine, err = streamingpromql.NewEngine(mqeOpts, limitsProvider, queryMetrics, planner, logger)
```
lamida, line 193: Why do we change the short variable declaration with var?

lamida (commented):
I leave some initial comments. Haven't deep dive the test part. But in general the change looks OK.

pkg/querier/dispatcher.go
charleskorn: Yes, the `errorTypeForError` function is derived from some code in the linked file.

pkg/querier/dispatcher.go
charleskorn: Addressed in 1976d5b.

pkg/querier/dispatcher.go
charleskorn: Addressed in 1976d5b.

pkg/querier/dispatcher.go
charleskorn: Removed in 3bc6e44.

pkg/querier/dispatcher.go
charleskorn: Changed in 3a19cc0.

pkg/querier/dispatcher.go
charleskorn: Changed in 3a19cc0.

tools/evaluate-query-plan/main.go
charleskorn: Good idea, changed in f0f2cc6.

pkg/querier/querier.go
charleskorn: Because we want to use the `streamingEngine` variable that has been declared on line 185.

pkg/mimir/modules.go
charleskorn: Yes, but I'm imagining this will be done in the query-frontend. If the flag is turned off, then the query-frontend will send requests to the querier via the same means as before.

pkg/querier/dispatcher.go
charleskorn: Given this is a temporary debugging endpoint that would never be exposed outside the cluster, I'm not too worried about this.

pkg/querier/dispatcher.go
charleskorn: Good idea, changed in fa01766.

pkg/querier/dispatcher.go
charleskorn: Good idea, changed in fa01766.

pkg/api/handlers.go
```diff
@@ -329,19 +329,23 @@ func NewQuerierHandler(
...
+	router.Path(path.Join(promPrefix, "/metadata")).Methods("GET").Handler(metadataQueryStats.Wrap(querier.NewMetadataHandler(metadataSupplier)))
+	router.Path(path.Join(promPrefix, "/cardinality/label_names")).Methods("GET", "POST").Handler(cardinalityQueryStats.Wrap(querier.LabelNamesCardinalityHandler(distributor, limits)))
+	router.Path(path.Join(promPrefix, "/cardinality/label_values")).Methods("GET", "POST").Handler(cardinalityQueryStats.Wrap(querier.LabelValuesCardinalityHandler(distributor, limits)))
+	router.Path(path.Join(promPrefix, "/cardinality/active_series")).Methods("GET", "POST").Handler(cardinalityQueryStats.Wrap(querier.ActiveSeriesCardinalityHandler(distributor, limits)))
+	router.Path(path.Join(promPrefix, "/cardinality/active_native_histogram_metrics")).Methods("GET", "POST").Handler(cardinalityQueryStats.Wrap(querier.ActiveNativeHistogramMetricsHandler(distributor, limits)))
+	router.Path(path.Join(promPrefix, "/format_query")).Methods("GET", "POST").Handler(formattingQueryStats.Wrap(promRouter))
+
+	if dispatcher != nil {
+		router.Path("/api/v1/evaluate").Methods("POST").Handler(dispatcher)
```
56quarters, line 347: Is the point here to _not_ prefix this with `/prometheus`? I'd rather prefix it for consistency even if it's temporary and not part of the Prometheus API.

tools/evaluate-query-plan/main.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if err := a.Run(); err != nil {
+		fmt.Printf("Application failed with an error: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func (a *app) ParseFlags() error {
+	flag.StringVar(&a.rootUrl, "root-url", "http://localhost:8005", "root URL of querier")
+	flag.StringVar(&a.expr, "expr", "", "expression to evaluate")
```
56quarters: Can you clarify if this is a PromQL expression or something else?

pkg/querier/dispatcher.go
```diff
@@ -0,0 +1,380 @@
...
+func (o *evaluationObserver) InstantVectorSeriesDataEvaluated(ctx context.Context, evaluator *streamingpromql.Evaluator, seriesIndex int, seriesData types.InstantVectorSeriesData) error {
+	defer types.PutInstantVectorSeriesData(seriesData, evaluator.MemoryConsumptionTracker)
+
+	// TODO: batch up series to return, rather than sending each series one at a time?
+
+	return o.w.Write(ctx, querierpb.EvaluateQueryResponse{
+		Message: &querierpb.EvaluateQueryResponse_InstantVectorSeriesData{
+			InstantVectorSeriesData: &querierpb.EvaluateQueryResponseInstantVectorSeriesData{
+				NodeIndex:  o.nodeIndex,
+				Floats:     mimirpb.FromFPointsToSamples(seriesData.Floats),
```
56quarters, line 271: These conversion methods do unsafe casts. I think this use is safe because the message is immediately serialized and written before the series data is returned to the pool, right?

56quarters (approved):
A few small comments but this looks really good! Besides the obvious implications, I'm looking forward to untangling our use of `httpgrpc`

tools/evaluate-query-plan/main.go
charleskorn: Done in 6fbd45a.

pkg/querier/dispatcher.go
charleskorn: Yep, that's right. I've added some comments to make this clear in 66dc838.

pkg/api/handlers.go
charleskorn: > Is the point here to _not_ prefix this with `/prometheus`?

Yep, exactly.

> I'd rather prefix it for consistency even if it's temporary and not part of the Prometheus API.

We seem to be a bit inconsistent about this - the general pattern seems to be that we prefix only Prometheus-compatible endpoints with `/prometheus`, and everything else doesn't get the prefix. But then we have some exceptions like the `/api/v1/cardinality/...` endpoints which are prefixed. 

My opinion is that we should try to keep `/prometheus` for Prometheus-compatible endpoints so we don't run into issues if Prometheus should ever introduce another endpoint we need to support. And I imagine that we'll start to retire some of these endpoints on queriers (eg. for instant and range queries) in favour of their non-httpgrpc equivalents from query-frontends.

lamida (approved):

PR #12303, opened 2025-08-05
Move pruning middleware into prune toggles AST optimisation pass

zenador:
#### What this PR does

Move pruning middleware into prune toggles AST optimisation pass.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/frontend/querymiddleware/ast_optimization_pass_test.go
charleskorn, line 1: Is there a reason to keep these tests here rather than move them to the `streamingpromql/optimize/ast` package?

pkg/frontend/querymiddleware/ast_optimization_pass_test.go
```diff
@@ -0,0 +1,124 @@
...
+	downstream := &downstreamHandler{
+		engine:    engine,
+		queryable: queryable,
+	}
+
+	for input, expected := range testCases {
+		if input == expected {
+			continue
+		}
+		t.Run(input, func(t *testing.T) {
```
charleskorn: Is there a reason why you're running these tests through the query-frontend? Given the optimisation pass runs entirely in the engine, could we instead instantiate a MQE `Engine` and evaluate queries with that?

tacole02 (approved):
Docs look good. Thank you!

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -0,0 +1,94 @@
...
+			require.NoError(t, resRewritten.Err)
+			matrixRewritten, ok := resRewritten.Value.(promql.Matrix)
+			require.True(t, ok)
+
+			approximatelyEqualsMatrix(t, matrixInput, matrixRewritten)
+		})
+	}
+}
+
+func approximatelyEqualsMatrix(t *testing.T, a, b promql.Matrix) {
```
charleskorn: The `testutils.RequireEqualResults` method does something similar - can we use that here?

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -0,0 +1,94 @@
...
+func testASTOptimizationPassWithData(t *testing.T, loadTemplate string, testCases map[string]string) {
+	numSamples := 100
+	replacer := strings.NewReplacer("<num samples>", fmt.Sprintf("%d", numSamples))
+	data := replacer.Replace(loadTemplate)
+
+	const step = 20 * time.Second
+
+	queryable := promqltest.LoadedStorage(t, data)
+
+	engine := promql.NewEngine(promql.EngineOpts{
```
charleskorn: Is there a reason why you're using a Prometheus engine instance here rather than MQE? If so, it'd be good to add a comment explaining why.

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -0,0 +1,94 @@
...
+		Logger:               promslog.NewNopLogger(),
+		Reg:                  nil,
+		MaxSamples:           100000000,
+		Timeout:              time.Minute,
+		EnableNegativeOffset: true,
+		EnableAtModifier:     true,
```
charleskorn: Can we use `streamingpromql.NewTestEngineOpts().CommonOpts` instead?

pkg/streamingpromql/optimize/ast/util_test.go
zenador: Didn't know about that one, thanks!

pkg/streamingpromql/optimize/ast/util_test.go
zenador: Needed to do https://github.com/grafana/mimir/pull/12326 first to avoid an import cycle

charleskorn: What if we change the `ast` tests to use the `ast_test` package, similar to what's done for `pkg/streamingpromql/optimize/plan/skip_histogram_decoding_test.go`?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * [CHANGE] Memcached: Remove experimental `-<prefix>.memcached.addresses-provider` flag to use alternate DNS service discovery backends. The more reliable backend introduced in 2.16.0 (#10895) is now the default. As a result of this change, DNS-based cache service discovery no longer supports search domains. #12175
 * [CHANGE] Query-frontend: Remove the CLI flag `-query-frontend.downstream-url` and corresponding YAML configuration and the ability to use the query-frontend to proxy arbitrary Prometheus backends. #12191
 * [CHANGE] Query-frontend: Remove experimental instant query splitting feature. #12267
+* [CHANGE] Query-frontend, querier: Replace `query-frontend.prune-queries` flag with `querier.mimir-query-engine.enable-prune-toggles` as pruning middleware has been moved into MQE AST optimization passes. #12303
```
charleskorn: [nit] "AST optimization passes" likely doesn't mean much to most users:

```suggestion
* [CHANGE] Query-frontend, querier: Replace `query-frontend.prune-queries` flag with `querier.mimir-query-engine.enable-prune-toggles` as pruning middleware has been moved into MQE. #12303
```

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+			qInput, err := engine.NewRangeQuery(ctx, queryable, nil, input, startTime, endTime, step)
+			require.NoError(t, err)
+			resInput := qInput.Exec(context.Background())
+			require.NoError(t, resInput.Err)
+
+			qRewritten, err := engine.NewRangeQuery(ctx, queryable, nil, expected, startTime, endTime, step)
+			require.NoError(t, err)
+			resRewritten := qRewritten.Exec(context.Background())
+			require.NoError(t, resRewritten.Err)
```
charleskorn: [nit] We should call `Close` on `qInput` and `qRewritten` once we're done with them.

PR #12304, opened 2025-08-05
Add matcher propagation across binary operations

zenador:
#### What this PR does

Add propagate matchers as an AST optimisation pass but run in query frontend rewrite middleware (so that it can be run before query sharding).

Fixes https://github.com/grafana/mimir-squad/issues/2991

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


cmd/mimir/help-all.txt.tmpl
```diff
@@ -2349,6 +2349,8 @@ Usage of ./cmd/mimir/mimir:
     	Override the default minimum TLS version. Allowed values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13
   -query-frontend.results-cache.memcached.tls-server-name string
     	Override the expected name on the server certificate.
+  -query-frontend.rewrite-propagate-matchers
+    	[experimental] True to enable rewriting queries to propagate label matchers across binary expressions.
```
tacole02:
```suggestion
    	[experimental] Set to true to enable rewriting queries to propagate label matchers across binary expressions.
```

tacole02 (approved):
Docs look good. I added a suggestion to clarify the description in the config parameter.

pkg/streamingpromql/planning.go
```diff
@@ -54,6 +54,10 @@ func NewQueryPlanner(opts EngineOpts) *QueryPlanner {
 		planner.RegisterASTOptimizationPass(ast.NewPruneToggles()) // Do this next to ensure that toggled off expressions are removed before the other optimization passes are applied.
 	}
 	planner.RegisterASTOptimizationPass(&ast.SortLabelsAndMatchers{}) // This is a prerequisite for other optimization passes such as common subexpression elimination.
+	// The following is applied in query pruning middleware for now as we need to run them before query sharding, but we leave most of the code here as we plan to move them here later. Note that these do not track the changed state properly.
```
charleskorn: > `Note that these do not track the changed state properly.`

Could you please elaborate on this comment? I'm not quite sure I understand it.

charleskorn: Rather than adding this commented-out code here, I would:

* leave a similar comment in `roundtrip.go` above the code where we add the rewriting middleware
* remove the commented-out code here
* add a comment here as a reminder that we want to move the optimisation pass here eventually

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,139 @@
...
+	`max by (foo, baz) (avg without (foo) (up)) / down{foo="bar", baz="fob"}`:                              `max by (foo, baz) (avg without (foo) (up{baz="fob"})) / down{foo="bar", baz="fob"}`,
+	`max without (baz) (avg by (foo, baz) (up)) / down{foo="bar", baz="fob"}`:                              `max without (baz) (avg by (foo, baz) (up{foo="bar"})) / down{foo="bar", baz="fob"}`,
+}
+
+func TestPropagateMatchers(t *testing.T) {
+	ctx := context.Background()
+
+	for input, expected := range testCasesPropagateMatchers {
+		t.Run(input, func(t *testing.T) {
+			expectedExpr, err := parser.ParseExpr(expected)
```
charleskorn, line 136: I assume this is necessary so we don't have to worry about formatting in the test cases when we compare them with `String` below - if so, might be good to add a comment explaining this

pkg/frontend/querymiddleware/rewrite.go
```diff
@@ -0,0 +1,110 @@
...
+) MetricsQueryMiddleware {
+	metrics := rewriteMetrics{
+		propagateMatchersAttempts: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Name: "cortex_frontend_rewrites_propagate_matchers_attempted_total",
+			Help: "Total number of queries the query-frontend attempted to rewrite by propagating matchers across binary operations.",
+		}),
+		propagateMatchersSuccesses: promauto.With(registerer).NewCounter(prometheus.CounterOpts{
+			Name: "cortex_frontend_rewrites_propagate_matchers_succeeded_total",
+			Help: "Total number of queries the query-frontend has successfully rewritten by propagating matchers across binary operations.",
+		}),
```
charleskorn: Similar comment as for https://github.com/grafana/mimir/pull/12375#discussion_r2271813352 - I think "success" might be a confusing term to use here.

pkg/frontend/querymiddleware/rewrite.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (m *rewriteMiddleware) Do(ctx context.Context, r MetricsQueryRequest) (Response, error) {
+	log := spanlogger.FromContext(ctx, m.logger)
+
+	rewrittenQuery, success, err := m.rewriteQuery(ctx, r.GetQuery())
+	if err != nil {
+		level.Warn(log).Log("msg", "failed to rewrite the input query, falling back to the original query", "query", r.GetQuery(), "err", err)
+
+		return m.next.Do(ctx, r)
+	}
```
charleskorn: Under what circumstances does rewriting fail? Is it only if the expression is malformed? If so, I think it'd be OK to just return the error immediately rather than continuing.

pkg/frontend/querymiddleware/rewrite.go
```diff
@@ -0,0 +1,110 @@
...
+func (m *rewriteMiddleware) rewriteQuery(ctx context.Context, query string) (string, bool, error) {
+	// Parse the query.
+	expr, err := parser.ParseExpr(query)
+	if err != nil {
+		return "", false, apierror.New(apierror.TypeBadData, DecorateWithParamName(err, "query").Error())
+	}
+	rewrittenQuery := expr
+	changed := false
+
+	if m.cfg.RewriteQueriesPropagateMatchers {
```
charleskorn: Should we move this check earlier and immediately return if `RewriteQueriesPropagateMatchers` is false?

pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -130,6 +132,10 @@ func (cfg *Config) cardinalityBasedShardingEnabled() bool {
 	return cfg.TargetSeriesPerShard > 0
 }
 
+func (cfg *Config) isPruningQueriesEnabled() bool {
```
charleskorn: Given this always returns `cfg.RewriteQueriesPropagateMatchers`, I'm not sure the extra method adds much - any reason not to use `cfg.RewriteQueriesPropagateMatchers` directly below?

pkg/streamingpromql/optimize/ast/astmapper_with_state.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/prometheus/prometheus/promql/parser"
+
+	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
+)
+
+// ExprMapperWithState wraps ExprMapper to include a method for checking if the mapper has changed the expression.
+type ExprMapperWithState interface {
```
charleskorn: [nit] This seems like a better fit for the `astmapper` package.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	return "Matcher propagation across binary operations"
+}
+
+func (p *PropagateMatchers) Apply(ctx context.Context, expr parser.Expr) (parser.Expr, error) {
+	return p.mapper.Map(ctx, expr)
+}
+
+func (p *PropagateMatchers) HasChanged() bool {
+	return p.mapper.HasChanged()
+}
```
charleskorn: Given we're not using this as an optimisation pass yet, I wouldn't bother adding this yet.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	containsAgg bool
+}
+
+func (mapper *propagateMatchers) propagateMatchersInBinaryExpr(e *parser.BinaryExpr) ([]*enrichedVectorSelector, []*labels.Matcher) {
+	if e.Op == parser.LOR {
+		return nil, nil
+	}
+
+	vssL, matchersL := mapper.extractVectorSelectors(e.LHS)
+	okL := len(vssL) > 0
```
charleskorn: [nit] Maybe `anyVectorSelectorsOnLeft` (or something like that) would be a clearer name?

Same applies for `okR` below.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+		return nil, nil
+	case *parser.AggregateExpr:
+		return mapper.extractVectorSelectorsFromAggregateExpr(e)
+	case *parser.BinaryExpr:
+		return mapper.propagateMatchersInBinaryExpr(e)
+	// Explicitly define what is not handled to avoid confusion.
+	case *parser.StepInvariantExpr:
+		// Used only for optimizations and not produced directly by parser, and should not contain
+		// any vector selectors anyway.
```
charleskorn: This is not true - `my_metric @ 123` is a step-invariant expression, and it contains a vector selector.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		return nil, nil
+	case *parser.SubqueryExpr:
+		// We do not support subqueries for now due to complexity.
+		return nil, nil
```
charleskorn: Could you please add some tests that cover this case?

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/optimize/ast"
+)
+
+var testCasesPropagateMatchers = map[string]string{
```
charleskorn, line 15: Could you please add some test cases where the expressions below are wrapped in `scalar`? eg. `scalar(up{foo="bar"} * down)`

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	for key := range set1 {
+		if _, exists := set2[key]; !exists {
+			diff[key] = struct{}{}
+		}
+	}
+	return diff
+}
+
+// TODO: Consider more functions, and add tests for these.
+func getIndexForEligibleFunction(funcName string) int {
```
charleskorn: Perhaps `vectorSelectorArgumentIndex` would be a clearer name?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	diff := make(map[string]struct{}, len(set1))
+	for key := range set1 {
+		if _, exists := set2[key]; !exists {
+			diff[key] = struct{}{}
+		}
+	}
+	return diff
+}
+
+// TODO: Consider more functions, and add tests for these.
```
charleskorn: You could add a test that iterates over all of the known functions (in `parser.Functions`) and confirms that they're all represented here.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	groupingSet := makeStringSet(e.Grouping)
+	for _, vs := range vss {
+		updateVecSelWithAggInfo(vs, groupingSet, include)
+	}
+	newMatchers := mapper.getMatchersToPropagate(labelMatchers, groupingSet, include)
+	return vss, newMatchers
+}
+
+func updateVecSelWithAggInfo(vs *enrichedVectorSelector, groupingSet map[string]struct{}, include bool) {
```
charleskorn: [nit] Let's use the full names:

```suggestion
func updateVectorSelectorWithAggregationInfo(vs *enrichedVectorSelector, groupingSet map[string]struct{}, include bool) {
```

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	if !ok {
+		return expr, false, nil
+	}
+
+	mapper.propagateMatchersInBinaryExpr(e)
+	return e, true, nil
+}
+
+// enrichedVectorSelector is a struct to hold additional information about vector selectors,
+// mainly the set of matchers that are allowed to propagate inwards and outwards (same set used
```
charleskorn:
```suggestion
// mainly the set of labels that are allowed to propagate inwards and outwards (same set used
```

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		return nil, nil
+	case !okL:
+		return vssR, matchersR
+	case !okR:
+		return vssL, matchersL
+	}
+
+	if e.VectorMatching == nil {
+		return nil, nil
+	}
```
charleskorn: Under what circumstances would `VectorMatching` be `nil` but there are vector selectors on both sides of the expression?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+
+	matchingLabelsSet := makeStringSet(e.VectorMatching.MatchingLabels)
+	var newMatchersL []*labels.Matcher
+	if e.Op == parser.LUNLESS {
+		// For LUNLESS, we do not propagate matchers from the right-hand side to the left-hand side.
```
charleskorn: Might be worth elaborating on this comment with a simple example.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+	}
+	vss := append(vssL, vssR...)
+	matchers, _ := combineMatchers(newMatchersL, newMatchersR, map[string]struct{}{}, false)
+	return vss, matchers
+}
+
+// extractVectorSelectors returns the vector selectors found in the expression (wrapped with additional
+// info, see enrichedVectorSelector), along with the label matchers associated with them, and a boolean
+// indicating whether any vector selectors were found.
```
charleskorn:
```suggestion
// extractVectorSelectors returns the vector selectors found in the expression (wrapped with additional
// info, see enrichedVectorSelector), along with the label matchers associated with them.
```

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+	}
+	return newMatchers, changed
+}
+
+func isMetricNameMatcher(m *labels.Matcher) bool {
+	return m.Name == labels.MetricName
+}
+
+func makeStringSet(list []string) map[string]struct{} {
```
charleskorn: [nit] I would call this function `newStringSet`, and have it return a dedicated `stringSet` type with `Contains`, `Add` etc. methods

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	return matchersToAdd
+}
+
+func combineMatchers(matchers, matchersToAdd []*labels.Matcher, labelsSet map[string]struct{}, include bool) ([]*labels.Matcher, bool) {
+	matchersMap := makeMatchersMap(matchers)
+	newMatchers := make([]*labels.Matcher, 0, len(matchers)+len(matchersToAdd))
+	newMatchers = append(newMatchers, matchers...)
+	changed := false
+	for _, m := range matchersToAdd {
+		if _, ok := matchersMap[m.String()]; !ok {
```
charleskorn: To help reduce the nesting here, I would invert the condition here, so we `continue` if `ok` is `true`.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		_, exists := labelsSet[m.Name]
+		if include {
+			if !exists {
+				continue
+			}
+		} else {
+			if exists {
+				continue
+			}
+		}
```
charleskorn: Can this be simplified to `if include != exists { continue }`?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+	// Differences between samples (using matrix selectors)
+	case "rate", "delta", "increase", "idelta", "irate":
+		return 0
+	default:
+		return -1
+	}
+}
+
+func (mapper *propagateMatchers) getMatchersToPropagate(matchersSrc []*labels.Matcher, labelsSet map[string]struct{}, include bool) []*labels.Matcher {
+	matchersToAdd := make([]*labels.Matcher, 0, len(labelsSet))
```
charleskorn: I'm not sure using `len(labelsSet)` makes sense here:

* if `include` is true, then we'll return at most `min(len(labelsSet), len(matchersSrc))` matchers
* if `include` is false, then we'll return at most `len(matchersSrc)` matchers, regardless of the size of `labelsSet`

Am I missing something?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+		matchersToAdd = append(matchersToAdd, m)
+	}
+
+	return matchersToAdd
+}
+
+func combineMatchers(matchers, matchersToAdd []*labels.Matcher, labelsSet map[string]struct{}, include bool) ([]*labels.Matcher, bool) {
+	matchersMap := makeMatchersMap(matchers)
+	newMatchers := make([]*labels.Matcher, 0, len(matchers)+len(matchersToAdd))
+	newMatchers = append(newMatchers, matchers...)
```
charleskorn: Should we only bother doing this once we've found a matcher to add? Or can we `append` directly to `matchers` and return that rather than creating a new slice?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,305 @@
...
+			_, exists := labelsSet[m.Name]
+			if include {
+				if !exists {
+					continue
+				}
+			} else {
+				if exists {
+					continue
+				}
+			}
```
charleskorn: Same comment as above, this looks like it could be simplified to `if include != exists { continue }`

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,139 @@
...
+	`up * on(foo) sum by (foo) (down{foo="bar", baz="fob"})`:                                               `up{foo="bar"} * on(foo) sum by (foo) (down{foo="bar", baz="fob"})`,
+	`up * on(foo) sum by (baz) (down{foo="bar", baz="fob"})`:                                               `up * on(foo) sum by (baz) (down{foo="bar", baz="fob"})`,
+	`up * ignoring(foo) sum by (baz) (down{foo="bar", baz="fob"})`:                                         `up{baz="fob"} * ignoring(foo) sum by (baz) (down{foo="bar", baz="fob"})`,
+	`up * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`:                                    `up * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`,
+	`up{soo="sar"} * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`:                         `up{soo="sar"} * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob", soo="sar"})`,
+	`topk by (foo) (5, up{foo="bar"}) / bottomk by (region) (5, down{baz="fob"})`:                          `topk by (foo) (5, up{foo="bar"}) / bottomk by (region) (5, down{baz="fob"})`,
```
charleskorn: What about something like `topk by (foo) (5, up{foo="bar"}) / bottomk by (foo) (5, down{baz="fob"})`?

cmd/mimir/config-descriptor.json
```diff
@@ -7389,6 +7389,17 @@
           "fieldFlag": "query-frontend.parallelize-shardable-queries",
           "fieldType": "boolean"
         },
+        {
+          "kind": "field",
+          "name": "rewrite_propagate_matchers",
```
56quarters: I don't think we need to include "rewrite" as part of the setting name here, _how_ the matchers are propagated doesn't really matter - just that they are.

pkg/streamingpromql/planning.go
zenador: (squashed some commits so it'd be easier to move some code over to the 2nd PR for reordering histogram aggregation)

That only applied to the commented out code in that state. When we move them from the query middleware to here in future, we can fix it to track `changed` properly by re-creating the mapper appropriately. Anyway, these comments are now removed along with the unused code for this.

pkg/frontend/querymiddleware/rewrite.go
zenador: Currently it's only if the expression is malformed, but there could be other kinds of errors in future. Would it be safer to just leave it so this middleware will never break anything if there's an error? Malformed queries wouldn't really make it to this step right (unless the middleware before this breaks)?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Removed the bit about vector selectors from the comment. It looks like the "used only for optimizations and not produced directly by parser" is still true, so I left the rest as-is. Added a unit test for this and it still works, so it's not being treated as a step-invariant expression.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: I don't think this actually happens currently, but it's a just-in-case check to avoid accessing stuff on a nil VectorMatching in case something breaks somewhere.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Good catch, I must have got them mixed up. Thank you!

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: I've used your second suggestion. It feels slightly unsafe to mutate it, but looking at the current code it shouldn't really affect anything and it still passes tests, and doing this is more efficient, so I just added a comment to be careful.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
zenador: Added that test case, is that how you'd expect it to behave?

cmd/mimir/config-descriptor.json
zenador: We plan to add a 2nd kind of optimization pass in this rewrite middleware after merging this PR, so I thought it'd be clearer to specify that both of them are part of the rewrite middleware but focus on different areas. If you still feel that both of them should omit the "rewrite_" part I'm open to changing this.

pkg/frontend/querymiddleware/rewrite.go
charleskorn: I think if this is failing, we should just return an error - it keeps things simple, and also means we'll catch any problems sooner.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
charleskorn: I would remove this check then - that way we'll catch the unexpected state rather than silently ignore it.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,292 @@
...
+
+	if e.VectorMatching == nil {
+		return nil, nil
+	}
+
+	matchingLabelsSet := newStringSet(e.VectorMatching.MatchingLabels)
+	var newMatchersL []*labels.Matcher
+	if e.Op == parser.LUNLESS {
+		// For LUNLESS, we do not propagate matchers from the right-hand side to the left-hand side.
+		// e.g. `up unless down{foo="bar"}` would remain unchanged but `up{foo="bar"} unless down` would become `up{foo="bar"} unless down{foo="bar"}`.
```
charleskorn:
```suggestion
		// For LUNLESS, we cannot propagate matchers from the right-hand side to the left-hand side for correctness reasons.
		// e.g. `up unless down{foo="bar"}` must remain unchanged, but `up{foo="bar"} unless down` can become `up{foo="bar"} unless down{foo="bar"}`.
```

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -212,69 +184,105 @@ func vectorSelectorArgumentIndex(funcName string) int {
...
+
+func getDifference(set1, set2 stringSet) stringSet {
+	diff := make(stringSet, len(set1))
+	for key := range set1 {
+		if !set2.Contains(key) {
+			diff.Add(key)
+		}
+	}
+	return diff
+}
```
charleskorn: [nit] These could become methods with a `stringSet` receiver.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -93,7 +98,8 @@ var testCasesPropagateMatchers = map[string]string{
 	`up * ignoring(foo) sum by (baz) (down{foo="bar", baz="fob"})`:                                         `up{baz="fob"} * ignoring(foo) sum by (baz) (down{foo="bar", baz="fob"})`,
 	`up * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`:                                    `up * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`,
 	`up{soo="sar"} * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob"})`:                         `up{soo="sar"} * ignoring(foo) sum without (baz) (down{foo="bar", baz="fob", soo="sar"})`,
-	`topk by (foo) (5, up{foo="bar"}) / bottomk by (region) (5, down{baz="fob"})`:                          `topk by (foo) (5, up{foo="bar"}) / bottomk by (region) (5, down{baz="fob"})`,
+	`topk by (foo) (5, up{foo="bar"}) / bottomk by (baz) (5, down{baz="fob"})`:                             `topk by (foo) (5, up{foo="bar"}) / bottomk by (baz) (5, down{baz="fob"})`,
+	`topk by (foo) (5, up{foo="bar"}) / bottomk by (foo) (5, down{baz="fob"})`:                             `topk by (foo) (5, up{foo="bar"}) / bottomk by (foo) (5, down{baz="fob"})`,
```
charleskorn: Can't we propagate the `foo` matcher in this case?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -212,69 +184,105 @@ func vectorSelectorArgumentIndex(funcName string) int {
 	// Differences between samples (using matrix selectors)
 	case "rate", "delta", "increase", "idelta", "irate":
 		return 0
+	// Aggregation over time
+	case "avg_over_time", "min_over_time", "max_over_time", "sum_over_time", "count_over_time", "stddev_over_time", "stdvar_over_time", "last_over_time", "present_over_time":
+		return 0
+	case "quantile_over_time":
+		return 1
+	// Explicitly not supported
```
charleskorn: Might be worth clarifying that `scalar()` isn't supported because it's not valid to propagate matchers from within it to outside it.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
charleskorn: I can see you've changed this to `len(matchersSrc)` in the most recent version, but I think this should be `min(len(labelsSet), len(matchersSrc))` if `include` is true.

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -67,6 +67,11 @@ var testCasesPropagateMatchers = map[string]string{
 	`up{foo="bar"} / round(down{baz="fob"}, 1)`:                                `up{foo="bar", baz="fob"} / round(down{foo="bar", baz="fob"}, 1)`,
 	`rate(up{foo="bar"}[2m]) + delta(down{baz="fob"}[4m])`:                     `rate(up{foo="bar", baz="fob"}[2m]) + delta(down{foo="bar", baz="fob"}[4m])`,
 	`up * -down{foo="bar"}`:                                                    `up{foo="bar"} * -down{foo="bar"}`,
+	`up{foo="bar"} @ 1 + down @ 2`:                                             `up{foo="bar"} @ 1 + down{foo="bar"} @ 2`,
+	`min_over_time(rate(up{foo="bar"}[5m])[30m:1m]) + min_over_time(rate(down[5m])[30m:1m])`: `min_over_time(rate(up{foo="bar"}[5m])[30m:1m]) + min_over_time(rate(down[5m])[30m:1m])`,
```
charleskorn: Might be worth adding a comment here that we don't support subqueries at present.

(could you please also create a follow-up issue so we know to come back and look at subquery support at some point in the future?)

zenador: Tried to add basic support for subqueries and it passes tests, should we just leave it for now? Otherwise, I can revert the commit and create an issue.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,312 @@
...
+	case "histogram_fraction":
+		return 2, nil
+	// Aggregation over time
+	case "avg_over_time", "min_over_time", "max_over_time", "sum_over_time", "count_over_time", "stddev_over_time", "stdvar_over_time", "last_over_time", "present_over_time", "absent_over_time", "mad_over_time", "ts_of_min_over_time", "ts_of_max_over_time", "ts_of_last_over_time":
+		return 0, nil
+	case "quantile_over_time":
+		return 1, nil
+	// Time
+	case "minute", "hour", "day_of_week", "day_of_month", "day_of_year", "days_in_month", "month", "year", "timestamp":
+		return 0, nil
```
charleskorn: For these functions, the argument is optional - we might need to add a check in `extractVectorSelectors` to make sure we don't try to read an argument that's not there (and some tests to cover these cases 🙂)

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,312 @@
...
+	// No vector/matrix selectors
+	case "pi", "time", "vector":
+		return -1, nil
+	// Explicitly not supported because it's not valid to propagate matchers across these functions.
+	case "scalar":
+		return -1, nil
+	// Explicitly not supported because we want to avoid unexpected interactions with labels or ordering.
+	case "label_join", "label_replace", "info", "sort", "sort_desc", "sort_by_label", "sort_by_label_desc":
+		return -1, nil
```
charleskorn, line 218: Could you please add some tests for these cases to make sure we don't inadvertently break this in the future?

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -162,31 +163,52 @@ func updateVectorSelectorWithAggregationInfo(vs *enrichedVectorSelector, groupin
...
-// TODO: Consider more functions, and add tests for these.
-func vectorSelectorArgumentIndex(funcName string) int {
+func VectorSelectorArgumentIndex(funcName string) (int, error) {
 	// If the function is eligible for matcher propagation, return the index of the argument
 	// that contains the vector selector (may be embedded in other kinds of expressions).
 	// Otherwise, return -1.
 	switch funcName {
+	// Simple
+	case "absent", "changes", "resets":
+		return 0, nil
```
charleskorn: Could you please add some tests specifically for `absent` and `absent_over_time`? 

I suspect we'll have to exclude them from this optimisation, as adding labels to the selector inside an `absent`/`absent_over_time` call changes the labels returned by the function.

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -33,28 +33,45 @@ func testASTOptimizationPassWithData(t *testing.T, loadTemplate string, testCase
...
-
-			testutils.RequireEqualResults(t, "", resInput, resRewritten, true)
+
+		t.Run(input+" range", func(t *testing.T) {
+			runAndCompare(t, input, expected, makeRangeQuery)
+		})
+
+		t.Run(input+" instant", func(t *testing.T) {
+			runAndCompare(t, input, expected, makeInstantQuery)
 		})
```
charleskorn: [nit] This might help organise the test output slightly more and make filtering easier:

```suggestion
		t.Run(input, func(t *testing.T) {
			t.Run("range query", func(t *testing.T) {
				runAndCompare(t, input, expected, makeRangeQuery)
			})

			t.Run("instant query", func(t *testing.T) {
				runAndCompare(t, input, expected, makeInstantQuery)
			})
		})
```

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Good catch, thanks!

pkg/streamingpromql/optimize/ast/propagate_matchers.go
zenador: Good catch, thanks!

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,183 @@
...
+	`max by (foo, baz) (avg without (foo) (up)) / down{foo="bar", baz="fob"}`:                              `max by (foo, baz) (avg without (foo) (up{baz="fob"})) / down{foo="bar", baz="fob"}`,
+	`max without (baz) (avg by (foo, baz) (up)) / down{foo="bar", baz="fob"}`:                              `max without (baz) (avg by (foo, baz) (up{foo="bar"})) / down{foo="bar", baz="fob"}`,
+
+	// Not supported
+	`scalar(up{foo="bar"} * down)`:                        `scalar(up{foo="bar"} * down{foo="bar"})`,
+	`scalar(up{foo="bar"}) * down`:                        `scalar(up{foo="bar"}) * down`,
+	`scalar(up) * down{foo="bar"}`:                        `scalar(up) * down{foo="bar"}`,
+	`absent(up{foo="bar99"}) + down{baz="fob"}`:           `absent(up{foo="bar99"}) + down{baz="fob"}`,
+	`absent(up{foo="bar99"}) + absent(down{baz="fob99"})`: `absent(up{foo="bar99"}) + absent(down{baz="fob99"})`,
+	`absent_over_time(rate(up{foo="bar99"}[5m])[30m:1m]) + absent_over_time(rate(down[5m])[30m:1m])`:                        `absent_over_time(rate(up{foo="bar99"}[5m])[30m:1m]) + absent_over_time(rate(down[5m])[30m:1m])`,
```
charleskorn: [nit] I think we can simplify this example - the subquery obscures the intent of this test, which is that matchers from within `absent_over_time` can't be propagated:
```suggestion
	`absent_over_time(up{foo="bar99"}[5m]) + absent_over_time(down[5m])`,
```

pkg/streamingpromql/optimize/ast/propagate_matchers_test.go
```diff
@@ -0,0 +1,183 @@
...
+	`label_join(up{foo="bar"}, "joint", "-", "boo", "faf") + label_join(down{baz="fob"}, "joint", "-", "boo", "faf")`:       `label_join(up{foo="bar"}, "joint", "-", "boo", "faf") + label_join(down{baz="fob"}, "joint", "-", "boo", "faf")`,
+	`label_replace(up{foo="bar"}, "zoo", "$1", "faf", "(.*)") + label_replace(down{baz="fob"}, "zoo", "$1", "faf", "(.*)")`: `label_replace(up{foo="bar"}, "zoo", "$1", "faf", "(.*)") + label_replace(down{baz="fob"}, "zoo", "$1", "faf", "(.*)")`,
+	`info(up{foo="bar"}) + info(down{baz="fob"})`:                                                                           `info(up{foo="bar"}) + info(down{baz="fob"})`,
+	`sort(up{foo="bar"}) + sort(down{baz="fob"})`:                                                                           `sort(up{foo="bar"}) + sort(down{baz="fob"})`,
+	`sort_by_label(up{foo="bar"}, "boo") + sort_by_label(down{baz="fob"}, "faf")`:                                           `sort_by_label(up{foo="bar"}, "boo") + sort_by_label(down{baz="fob"}, "faf")`,
+}
+
+func TestPropagateMatchers(t *testing.T) {
+	parser.EnableExperimentalFunctions = true
```
charleskorn: In order to not affect other tests, we should restore this back to its original value after the test ends. 

Same comment applies to `TestPropagateMatchersWithData` too.

pkg/streamingpromql/optimize/ast/propagate_matchers.go
```diff
@@ -0,0 +1,323 @@
...
+		return mapper.extractVectorSelectors(e.Expr)
+	case *parser.SubqueryExpr:
+		return mapper.extractVectorSelectors(e.Expr)
+	case *parser.Call:
+		if i, optional, _ := VectorSelectorArgumentIndex(e.Func.Name); i >= 0 {
+			if optional && len(e.Args) <= i {
+				return nil, nil
+			}
+			return mapper.extractVectorSelectors(e.Args[i])
+		}
```
charleskorn: Do we need `optional` here? Could we do this instead?

```suggestion
		if i, _ := VectorSelectorArgumentIndex(e.Func.Name); i >= 0 {
			if len(e.Args) <= i {
				return nil, nil
			}
			return mapper.extractVectorSelectors(e.Args[i])
		}
```

zenador:
@charleskorn is this good to merge now?

charleskorn:
> @charleskorn is this good to merge now?

Sorry, I reviewed this and forgot to submit my review 🤦

charleskorn (approved):

PR #12305, opened 2025-08-05
Add reordering of histogram aggregations

zenador:
#### What this PR does

Add reordering of histogram aggregations as an AST optimisation pass but run in query frontend rewrite middleware (so that it can be run before query sharding).

Fixes https://github.com/grafana/mimir-squad/issues/3027

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation.go
+++ b/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation.go
@@ -1,0 +1,118 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package ast
+
+import (
+	"context"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql/parser"
+
+	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
+)
+
+// NewReorderHistogramAggregationMapper optimizes queries by reordering histogram functions and aggregations
+// for more efficient execution.
+func NewReorderHistogramAggregationMapper() *astmapper.ASTExprMapperWithState {
+	mapper := &reorderHistogramAggregation{}
+	return astmapper.NewASTExprMapperWithState(mapper)
+}
+
+type reorderHistogramAggregation struct {
+	changed bool
+}
+
+func (mapper *reorderHistogramAggregation) HasChanged() bool {
+	return mapper.changed
+}
+
+func (mapper *reorderHistogramAggregation) MapExpr(ctx context.Context, expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
+	call, ok := expr.(*parser.Call)
+	if !ok || !mapper.isSwitchableCall(call.Func) {
+		return expr, false, nil
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/streamingpromql/optimize/ast/testdata/reorder_histogram_aggregation.go
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package testdata
+
+var TestCasesReorderHistogramAggregation = map[string]string{
```
charleskorn: There's two other edge cases we should consider:

* `histogram_sum(sum by (__name__) ({...}))` - we can't move the histogram function inside the aggregation as the function drops the metric name, which will cause incorrect aggregations
* `histogram_sum(sum({__name__=~"foo.*"}))` - we can't move the histogram function inside the aggregation as the function drops the metric name and the selector possibly selects metrics with different names, and the input to the aggregation could have multiple series with the same labels (which will cause an `vector cannot contain metrics with the same labelset` error)

pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/optimize/ast"
+)
+
+var testCasesReorderHistogramAggregation = map[string]string{
```
charleskorn, line 15: Could you please add some tests for cases with unsupported aggregations (eg. `max`)?

pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return newExpr, false, nil
+}
+
+func (*reorderHistogramAggregation) isSwitchableCall(callFunc *parser.Function) bool {
+	return callFunc.Name == "histogram_sum" || callFunc.Name == "histogram_count"
+}
+
+func (*reorderHistogramAggregation) isSwitchableAgg(op parser.ItemType) bool {
+	return op == parser.SUM || op == parser.AVG
```
charleskorn, line 79: Would `count` and `group` be valid too?

tacole02 (approved):
Docs look good! Thanks, @zenador !

pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation.go
zenador: Unlike the supported sum and avg which result in histograms with histogram input, count and group on histograms result in floats, so it is not interchangeable, e.g. `count(histogram_count(histo))` has a result but `histogram_count(count(histo))` is empty as you cannot use histogram functions on floats.

charleskorn: Good point, ignore me 🙂

PR #12306, opened 2025-08-05
Distributor: Introduce OTel translation strategy

aknuds1:
#### What this PR does

Replace experimental `-distributor.otel-enable-unescaped-names` flag with `-distributor.otel-translation-strategy`. Follow-up to #12284, as @ywwg requested we expose the functionality via the same set of strategies as Prometheus' [`otlp.translation_strategy` option](https://github.com/prometheus/prometheus/blob/74610b7c89bdc14ebddb1879ebc3c6113e517528/docs/configuration/configuration.md?plain=1#L215) instead.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/validation/limits.go
```diff
@@ -346,7 +346,7 @@ func (l *Limits) RegisterFlags(f *flag.FlagSet) {
...
-	f.BoolVar(&l.OTelEnableUnescapedNames, "distributor.otel-enable-unescaped-names", false, "Whether to disable escaping of metric and label names to the classical Prometheus format when ingesting OTLP metrics. If -distributor.otel-metric-suffixes-enabled is true, type and unit suffixes are still appended as required. When this is enabled, -validation.name-validation-scheme has to be configured as 'utf8'.")
+	f.Var(&l.OTelTranslationStrategy, "distributor.otel-translation-strategy", fmt.Sprintf("Translation strategy to apply in OTLP endpoint for metric and label names. Supported values: %s.", strings.Join([]string{`""`, string(otlptranslator.UnderscoreEscapingWithSuffixes), string(otlptranslator.UnderscoreEscapingWithoutSuffixes), string(otlptranslator.NoUTF8EscapingWithSuffixes), string(otlptranslator.NoTranslation)}, ", ")))
```
ywwg: is there related mimir documentation that we will also need to update?

ywwg (approved):
looks perfect!

pkg/util/validation/limits.go
aknuds1: There's just a page documenting experimental features. I'm updating that in the PR.

juliusmh: [nit] Perhaps we could add some more explanation on the empty case, e.g. "If left empty, translation strategy is derived from --validation.name-validation-scheme and --distributor.otel-metric-suffixes-enabled"

aknuds1: That's a good point @juliusmh, thanks! I can add it.

aknuds1: Added. Could you take another look please @juliusmh?

juliusmh (approved):

PR #12314, opened 2025-08-05
Block-Builder: Add Graceful Shutdown

dmwilson-grafana:
#### What this PR does

Block-Builder will attempt to complete a job it even if the service context is cancelled. In some cases, when the service processes a job right before shutdown, the metrics emitted for that job are never scraped. This PR adds `block-builder.shutdown-grace-period`, which blocks the `Service.stopping()`. This option should be configured long enough so that any pull-based metric collector can observe data related to the recently-completed job.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


narqo:
TBH, I feel this solution is a bit too complicated, comparing to the original problem. If we're only concerned about the false-positive alert, I propose we improve the alert https://github.com/grafana/mimir/pull/12319

To delay the restart, hoping that an external monitoring will come at the last minute, and scrap the metrics, feels a bit unreliable.

dmwilson-grafana:
Closed in favor of https://github.com/grafana/mimir/pull/12319

PR #12315, opened 2025-08-06
Support for zone aware PDB in jsonnet

tcp13equals2:
#### What this PR does

Allows for jsonnet to generate manifests related to rollout-operator zone aware PDB.

Relates to https://github.com/grafana/rollout-operator/pull/253

Fixes #194

Relates to https://github.com/grafana/rollout-operator/pull/253

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir/rollout-operator.libsonnet
charleskorn, line 1: Not something for this PR, but would be a good follow-up action: we probably should move some of this to a general rollout-operator Jsonnet module shared by all users of the rollout-operator, a good chunk of this is not specific to Mimir at all.

operations/mimir/multi-zone.libsonnet
```diff
@@ -15,6 +15,8 @@
     multi_zone_ingester_replication_read_path_enabled: true,
     multi_zone_ingester_replicas: 0,
     multi_zone_ingester_max_unavailable: 50,
+    // use the PodDisruptionZoneBudget - requires the rollout operator webhook to also be enabled
```
charleskorn:
```suggestion
    // use the ZoneAwarePodDisruptionBudget - requires the rollout operator webhook to also be enabled
```

operations/mimir/multi-zone.libsonnet
```diff
@@ -23,6 +25,8 @@
     // When store-gateway lazy loading is disabled, store-gateway may take a long time to startup.
     // To speed up rollouts, we increase the max unavailable to rollout all store-gateways in a zone in a single batch.
     multi_zone_store_gateway_max_unavailable: if $._config.store_gateway_lazy_loading_enabled then 50 else 1000,
+    // use the PodDisruptionZoneBudget - requires the rollout operator webhook to also be enabled
```
charleskorn:
```suggestion
    // use the ZoneAwarePodDisruptionBudget - requires the rollout operator webhook to also be enabled
```

operations/mimir/rollout-operator.libsonnet
```diff
@@ -166,6 +190,80 @@
       namespace: $._config.namespace,
     }),
 
+  zpdb_validation_webhook: if !zpdb_enabled then null else
```
charleskorn: I think we should always enable this webhook provided the rollout-operator supports ZPDBs - otherwise, there's a race between creating this webhook and creating the first ZPDB, which might allow us to create an invalid ZPDB.

operations/mimir/rollout-operator.libsonnet
```diff
@@ -108,6 +125,13 @@
           policyRule.withVerbs(['get', 'patch']),
         ] else []
       )
+      + (
+        if zpdb_enabled then [
```
charleskorn: I think we should always enable these rules provided the rollout-operator supports ZPDBs - otherwise, won't the rollout-operator fail to start because it won't have permission to watch ZPDBs?

operations/mimir/zone-aware-pod-disruption-budget.libsonnet
```diff
@@ -0,0 +1,24 @@
+{
+  // zpdbTemplate creates new ZoneAwarePodDisruptionBudget resource.
+  // name is the rollout-group
+  zpdbTemplate(name, rolloutGroup, maxUnavailable, ingestStorageEnabled):: {
```
charleskorn: [nit] I'd call this something like `newZPDB` or `zpdb`: we're a bit inconsistent with the naming convention here, but we seem to use either of those two patterns.

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@
 
 * [CHANGE] Distributor: Reduce calculated `GOMAXPROCS` to be closer to the requested number of CPUs. #12150
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12187
+* [CHANGE] Use `irate()` when calculating CPU scaling metric. Using `irate()` prevents underestimating CPU utilization when scraping fails. #12384
+* [CHANGE] Remove `vector(0)` when calculating memory scaling metric to prevent underestimating memory usage when scraping fails. #12386
```
charleskorn: These looks out of place here - perhaps these became duplicated during one of the merges?

operations/helm/charts/mimir-distributed/templates/crds/zoneawarepoddisruptionbudgets.rollout-operator.grafana.com-CustomResourceDefinition.yml
charleskorn, line 1: Rather than adding this here, I would add it to the rollout-operator Helm chart instead.

operations/mimir/zone-aware-pod-disruption-budget.libsonnet
```diff
@@ -0,0 +1,24 @@
...
+      },
+    },
+    spec: {
+      maxUnavailable: maxUnavailable,
+      selector: {
+        matchLabels: {
+          'rollout-group': rolloutGroup,
+        },
+      },
+      podNamePartitionRegex: if (ingestStorageEnabled) then '[a-z\\-]+-zone-[a-z]-([0-9]+)' else '',
```
charleskorn: I would have the caller of `newZPDB` pass this regex - then this method knows nothing about ingesters or how we label pods in different zones etc.

operations/mimir/rollout-operator.libsonnet
```diff
@@ -166,6 +185,80 @@
...
+            name: 'rollout-operator',
+            namespace: $._config.namespace,
+            path: '/admission/zpdb-validation',
+            port: 443,
+          },
+        },
+      },
+    ]),
+
+  pod_eviction_webhook: if !zpdb_enabled then null else
```
charleskorn: Should we make this the same condition as the other webhooks? (`!rollout_operator_enabled || !$._config.enable_rollout_operator_webhook`)

operations/mimir/multi-zone.libsonnet
```diff
@@ -155,10 +159,13 @@
     $.newIngesterZoneService($.ingester_zone_c_statefulset),
 
   ingester_rollout_pdb: if !$._config.multi_zone_ingester_enabled then null else
-    podDisruptionBudget.new('ingester-rollout') +
-    podDisruptionBudget.mixin.metadata.withLabels({ name: 'ingester-rollout' }) +
-    podDisruptionBudget.mixin.spec.selector.withMatchLabels({ 'rollout-group': 'ingester' }) +
-    podDisruptionBudget.mixin.spec.withMaxUnavailable(1),
+    if $._config.multi_zone_ingester_zpdb_enabled then
+      $.newZPDB('ingester-rollout', 'ingester', 1, $._config.ingest_storage_enabled)
```
charleskorn: It'd be good to make the `maxUnavailable` value configurable via a value on `$._config`

(same applies for store-gateways below)

operations/mimir/rollout-operator.libsonnet
tcp13equals2: The reason I did the specific zpdb_enabled check is that if the zpdb is not being used, the pod eviction webhook becomes inline to every eviction. 

ie every eviction will go through this webhook and be allowed because no zpdb config is found. 

Do you see any value in having this setup this way if there are no zpdb deployments?

operations/helm/charts/mimir-distributed/templates/crds/zoneawarepoddisruptionbudgets.rollout-operator.grafana.com-CustomResourceDefinition.yml
tcp13equals2: Would it be ok to leave this here for the purpose of this PR, and given my next task is to work on the rollout-operator helm chart I can move this across as part of that work?

CHANGELOG.md
```diff
@@ -41,7 +41,10 @@
 
 * [CHANGE] Distributor: Reduce calculated `GOMAXPROCS` to be closer to the requested number of CPUs. #12150
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12187
+* [CHANGE] Use `irate()` when calculating CPU scaling metric. Using `irate()` prevents underestimating CPU utilization when scraping fails. #12384
```
tcp13equals2: I don't know why these are still showing in this PR?

I've merged main into this branch ??

operations/mimir/multi-zone.libsonnet
```diff
@@ -159,8 +164,10 @@
     $.newIngesterZoneService($.ingester_zone_c_statefulset),
 
   ingester_rollout_pdb: if !$._config.multi_zone_ingester_enabled then null else
-    if $._config.multi_zone_ingester_zpdb_enabled then
-      $.newZPDB('ingester-rollout', 'ingester', 1, $._config.ingest_storage_enabled)
+    if $._config.multi_zone_ingester_zpdb_enabled && $._config.ingest_storage_enabled && $._config.rollout_operator_enabled &&  $._config.enable_rollout_operator_webhook then
+      $.newZPDB('ingester-rollout', 'ingester', $._config.multi_zone_ingester_zpdb_max_unavailable, $._config.multi_zone_ingester_zpdb_partition_name_regex, $._config.multi_zone_ingester_zpdb_partition_name_regex_group)
+    else if $._config.multi_zone_ingester_zpdb_enabled && $._config.rollout_operator_enabled &&  $._config.enable_rollout_operator_webhook then
+      $.newZPDB('ingester-rollout', 'ingester', $._config.multi_zone_ingester_zpdb_max_unavailable)
```
charleskorn: Might be worth introducing a local for `$._config.multi_zone_ingester_zpdb_enabled && $._config.rollout_operator_enabled &&  $._config.enable_rollout_operator_webhook` to help make these two conditions a bit clearer

CHANGELOG.md
charleskorn: It may be that these lines have been removed from `main` (I think the PRs in question were reverted)

tcp13equals2: urg ... ok I'll go check and delete from here

operations/mimir/rollout-operator.libsonnet
charleskorn: I think it's valuable to be able to check that the rollout-operator doesn't interfere with anything if no ZPDBs have been defined.

operations/helm/charts/mimir-distributed/templates/crds/zoneawarepoddisruptionbudgets.rollout-operator.grafana.com-CustomResourceDefinition.yml
charleskorn: Yep, that's fine

operations/mimir/rollout-operator.libsonnet
tcp13equals2: how about I make it a config variable which defaults to true ... so it could be turned off if need be?

charleskorn: Sounds great to me.

operations/mimir/rollout-operator.libsonnet
```diff
@@ -37,8 +37,32 @@
                               $._config.cortex_compactor_concurrent_rollout_enabled ||
                               $._config.ingest_storage_ingester_autoscaling_enabled ||
                               $._config.enable_rollout_operator_webhook,
+
+    // do not deploy the pod eviction webhook if there is no zone pdb configured.
+    // the eviction webhook can be deployed even if there are no zone pdb configurations.
+    // pod evictions not managed by a zpdb will be allowed and contine to be handled by any other pdb configurations.
+    disable_rollout_operator_eviction_webhook_when_no_zpdb: false,
```
charleskorn: I would simplify this a bit: I would have an option like `enabled_rollout_operator_eviction_webhook` that defaults to the value of `$._config.enable_rollout_operator_webhook`.

operations/mimir/rollout-operator.libsonnet
tcp13equals2: https://github.com/grafana/mimir-squad/issues/3179

operations/mimir/rollout-operator.libsonnet
```diff
@@ -26,6 +26,7 @@
     // block other operations that would block the service creation.
     ignore_rollout_operator_no_downscale_webhook_failures: false,
     ignore_rollout_operator_prepare_downscale_webhook_failures: false,
+    ignore_webhook_failures: false,
```
tcp13equals2: A flag to set all the rollout-operator webhooks to Ignore

operations/mimir/rollout-operator.libsonnet
```diff
@@ -49,6 +74,10 @@
 
   rollout_operator_node_affinity_matchers:: [],
 
+  // Create custom resource template for PodDisruptionZoneBudget
+  zpdb_crd:: std.parseYaml(importstr 'zone-aware-pod-disruption-budget-crd.yaml'),
+  zpdb_custom_resource: if !zpdb_enabled then null else $.zpdb_crd,
```
tcp13equals2, line 79: I have moved this behind the flag to avoid getting the zpdb crd in every jsonnet unit test.

operations/mimir/rollout-operator.libsonnet
```diff
@@ -101,6 +130,13 @@
         policyRule.withResources(['configmaps']) +
         policyRule.withVerbs(['get', 'update', 'create']),
       ] + (
+        if zpdb_enabled then [
```
tcp13equals2: This now matches the way the replica template rules are done.

tcp13equals2:
Note also - this code will be rendered obsolete with the active work being completed for https://github.com/grafana/mimir-squad/issues/3179 ... the rollout-operator related jsonnet will be moved into the rollout-operator repo and removed from mimir repo.

operations/mimir/rollout-operator.libsonnet
charleskorn: Don't we always need these permissions and the CRD enabled if the rollout-operator is running a version that supports ZPDBs? I'd expect it to fail to start if these permissions or the CRD wasn't in place, because the ZPDB observer will fail to start.

tcp13equals2: hmmm ... yes that is true. I was trying to avoid having to pull in those files to all the jsonnet mimir unit tests.

I'll update this to be if the webhooks are enabled as we only start the pod_eviction_controller if the TLS server is enabled.

operations/mimir/rollout-operator.libsonnet
```diff
@@ -37,8 +38,37 @@
...
+    $._config.multi_zone_ingester_zpdb_enabled &&
+    $._config.enable_rollout_operator_webhook,
+
+  local store_gateway_zpdb_enabled =
+    $._config.rollout_operator_enabled &&
+    $._config.multi_zone_store_gateway_enabled &&
+    $._config.multi_zone_store_gateway_zpdb_enabled &&
+    $._config.enable_rollout_operator_webhook,
+
+  local zpdb_enabled = ingester_zpdb_enabled || store_gateway_zpdb_enabled,
```
charleskorn, line 66: [nit] Maybe `any_zpdb_enabled` would be a clearer name?

operations/mimir/rollout-operator.libsonnet
```diff
@@ -49,6 +79,10 @@
 
   rollout_operator_node_affinity_matchers:: [],
 
+  // Create custom resource template for PodDisruptionZoneBudget
+  zpdb_crd:: std.parseYaml(importstr 'zone-aware-pod-disruption-budget-crd.yaml'),
+  zpdb_custom_resource: if !zpdb_resources_enabled then null else $.zpdb_crd,
```
charleskorn: [nit] Might be able to simplify this a little:

```suggestion
  zpdb_custom_resource: if !zpdb_resources_enabled then null else std.parseYaml(importstr 'zone-aware-pod-disruption-budget-crd.yaml'),
```

operations/mimir/rollout-operator.libsonnet
```diff
@@ -213,7 +328,7 @@
     mutatingWebhookConfiguration.withWebhooksMixin([
       mutatingWebhook.withName('prepare-downscale-%s.grafana.com' % $._config.namespace)
       + mutatingWebhook.withAdmissionReviewVersions(['v1'])
-      + mutatingWebhook.withFailurePolicy(if $._config.ignore_rollout_operator_prepare_downscale_webhook_failures then 'Ignore' else 'Fail')
+      + mutatingWebhook.withFailurePolicy(if $._config.ignore_rollout_operator_prepare_downscale_webhook_failures || $._config.ignore_webhook_failures then 'Ignore' else 'Fail')
```
charleskorn: Should we introduce separate `ignore_rollout_operator_zpdb_eviction_webhook_failures` and `ignore_rollout_operator_zpdb_validation_webhook_failures` options rather than introducing an overlapping `ignore_webhook_failures` option here?

operations/mimir/rollout-operator.libsonnet
```diff
@@ -165,6 +206,80 @@
...
+            name: 'rollout-operator',
+            namespace: $._config.namespace,
+            path: '/admission/zpdb-validation',
+            port: 443,
+          },
+        },
+      },
+    ]),
+
+  pod_eviction_webhook: if disable_eviction_webhook && !zpdb_enabled then null else
```
charleskorn: Do we need to disable this if we're not deploying any ZPDBs?

```suggestion
  pod_eviction_webhook: if disable_eviction_webhook then null else
```

operations/mimir/rollout-operator.libsonnet
tcp13equals2: The reason I did the single variable that applies to all the webhooks is that changing this failure policy is only needed in the scenario of enabling the webhooks for the first time or enabling the webhooks + upgrading the rollout operator. ie a validating webhook is registered but the rollout operator does not yet support that endpoint.

It's less likely that this failure policy needs to be toggled on an individual webhook.

If the failure mode is toggled, it needs to be re-deployed anyway to set back to Fail - so toggling one variable on/off seems more practical. 

But happy to change to having all individual flags if you think it is better this way.

operations/mimir/rollout-operator.libsonnet
tcp13equals2: I just did this to match the way the replica template CRD was done. I'll adjust both to your suggestion as I pull this jsonnet into the rollout operator repo.

operations/mimir/rollout-operator.libsonnet
tcp13equals2: The intent of this criteria was to allow for the pod eviction webhook to be disabled IFF there was no zpdb required. 

With this suggested change a user could break the zpdb and be denied all evictions.  ie perhaps they were running with the no pod eviction webhook and then they enabled a zpdb. This configuration would not be valid if we only check for the disable_eviction_webhook variable.

operations/mimir/rollout-operator.libsonnet
charleskorn: I think individual flags is safer. For example, if the rollout-operator is already deployed with the no-downscale webhook running, even temporarily switching the no-downscale webhook to ignore failures could lead to workloads being downscaled too soon if these are happening around the time of the rollout-operator rollout, so I'd want whoever was setting these flags to explicitly opt-in to setting all of them to ignore failures.

operations/mimir/rollout-operator.libsonnet
charleskorn: I think I'm missing your point sorry - if there are no ZPDBs deployed, then won't the eviction webhook allow all evictions?

tcp13equals2: If there are no zpdbs deployed then yes, the eviction webhook will allow the eviction request and it will fall through to any other standard pdb configuration (if any).

The purpose of disable_eviction_webhook is to allow a user to fully remove the pod eviction webhook if there is no zpdb in place. Although it is safe to have the eviction webhook running when there are no zpdbs, this option allows a user to remove the eviction webhook.

charleskorn: Right, in that case I would change the condition here to `if disable_eviction_webhook then null else ...` and add [an assertion](https://jsonnet.org/learning/tutorial.html#errors) that `disable_eviction_webhook` is not true if `zpdb_enabled` is true - this way, rendering the manifests will fail and the user can choose what to do.

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
 * [CHANGE] Distributor: Reduce calculated `GOMAXPROCS` to be closer to the requested number of CPUs. #12150
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12187
 * [CHANGE] Rollout-operator: Add `watch` permission to the rollout-operators's cluster role. #12360. See [rollout-operator#262](https://github.com/grafana/rollout-operator/pull/262)
+* [CHANGE] Rollout-operator: Support for zone and partition aware pod disruption budgets, enabling finer control over pod eviction policies. #12315. See [rollout-operator#253](https://github.com/grafana/rollout-operator/pull/253)
```
```suggestion
* [CHANGE] Rollout-operator: Support for zone and partition aware pod disruption budgets, enabling finer control over pod eviction policies. #12315. Refer to [rollout-operator#253](https://github.com/grafana/rollout-operator/pull/253).
```

tacole02 (approved):
Changelog looks good... left a minor suggestion.

tcp13equals2:
Do not merge until we have rolled out new rollout-operator and zpdb dependencies via deployment tools ...

tcp13equals2:
Need to disable the crd generation since this will be done at the cluster level - https://github.com/grafana/deployment_tools/pull/333367

tcp13equals2:
This PR is replaced by https://github.com/grafana/mimir/pull/12688

PR #11889, opened 2025-06-27
otlp: Stick to OTLP vocabulary on invalid label value length error

tcard:
#### What this PR does

Use specific error message for the "label value too long" error when from the OTLP endpoint, translating vocabulary to OTel and replacing Mimir config specifics with a link to OTLP translation documentation.

**Old message**

> <code>received a series whose label value length exceeds the limit, label: 'too_long', value: 'huge value' (truncated) series: 'foo{too_long="huge value"}' (err-mimir-label-value-too-long). To adjust the related per-tenant limit, configure -validation.max-length-label-value, or contact your service administrator.</code>

**New message**

> <code>received a metric whose attribute value length exceeds the limit of 9, attribute: 'too_long', value: 'huge value' (truncated) metric: 'foo{too_long="huge value"}'. See: https://grafana.com/docs/grafana-cloud/send-data/otlp/otlp-format-considerations/#metrics-ingestion-limits</code>

#### Which issue(s) this PR fixes or relates to

See backreferences.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12320, opened 2025-08-06
Trigger request buffer reuse in distributor validation tests

tcard:
#### What this PR does

The [recent incident](https://grafana.com/blog/2025/08/03/grafana-cloud-metrics-memory-corruption-issue-resolved/) introduced by https://github.com/grafana/mimir/pull/11889 would have been caught with a sufficiently exhaustive test suite.

* Change the distributor validation tests so that they submit actual gRPC write requests, which use the request buffer pool.
* Repeat each test a number of times to increase the likelihood of triggering use-after-release issues.
* Taint the buffers before returning them to the pool, so that even if the buffer isn't reused, lingering references still point to unexpected contents to make tests fail.

To demonstrate that these changes are enough to catch the issue, the original PR has been resubmitted (as a draft) on top of this branch, and [tests do fail](https://github.com/grafana/mimir/actions/runs/16779690455/job/47514746172?pr=12273).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/costattribution/testutils/test_utils.go
+++ b/pkg/costattribution/testutils/test_utils.go
@@ -25,10 +25,10 @@ func NewMockCostAttributionOverrides(limits validation.Limits, overrides map[str
 		// user6 has opted to rename team to eng_team.
 		"user6": {MaxCostAttributionCardinality: 5, CostAttributionLabelsStructured: []costattributionmodel.Label{{Input: "team", Output: "eng_team"}}},
 	}
-	for _, ul := range userLabels {
-		baseLimits[ul[0]] = &validation.Limits{
+	for _, uls := range userLabels {
+		baseLimits[uls[0]] = &validation.Limits{
 			MaxCostAttributionCardinality: 10,
-			CostAttributionLabels:         ul[1:],
+			CostAttributionLabels:         uls[1:],
 		}
 	}
 	for userID, l := range baseLimits {
--- a/pkg/costattribution/testutils/test_utils.go
+++ b/pkg/costattribution/testutils/test_utils.go
@@ -8,7 +8,14 @@ import (
 	"github.com/grafana/mimir/pkg/util/validation"
 )
 
-func NewMockCostAttributionLimits(idx int, lvs ...string) *validation.Overrides {
+func NewMockCostAttributionLimits(idx int, userLabels ...[]string) *validation.Overrides {
+	return NewMockCostAttributionOverrides(validation.Limits{}, nil, idx, userLabels...)
+}
+
+func NewMockCostAttributionOverrides(limits validation.Limits, overrides map[string]*validation.Limits, idx int, userLabels ...[]string) *validation.Overrides {
+	if overrides == nil {
+		overrides = map[string]*validation.Limits{}
+	}
 	baseLimits := map[string]*validation.Limits{
 		"user1": {MaxCostAttributionCardinality: 5, CostAttributionLabels: []string{"team"}},
 		"user2": {MaxCostAttributionCardinality: 2, CostAttributionLabels: []string{}},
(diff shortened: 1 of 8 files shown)


pkg/distributor/validate_test.go
```diff
@@ -292,92 +353,179 @@ func TestValidateLabels(t *testing.T) {
...
+								assert.Equal(t, 200, resp.Code, resp.Body.String())
+							}
+						}()
+					}
+
+					wg.Wait()
+				})
+			}
+		})
+	}
```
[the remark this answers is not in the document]
tcard: The comment above `finalChecks` explains why this isn't possible.

```
// We want to check those after all subtests are done, but parent tests
// cannot wait for subtests. So we're going to run this in the last subtest
// that finishes instead.
```

pkg/costattribution/testutils/test_utils.go
```diff
@@ -18,26 +25,35 @@ func NewMockCostAttributionLimits(idx int, lvs ...string) *validation.Overrides
 		// user6 has opted to rename team to eng_team.
 		"user6": {MaxCostAttributionCardinality: 5, CostAttributionLabelsStructured: []costattributionmodel.Label{{Input: "team", Output: "eng_team"}}},
 	}
-	if len(lvs) > 0 {
-		baseLimits[lvs[0]] = &validation.Limits{
+	for _, ul := range userLabels {
```
aknuds1: [Nit] Could this be named so it's obvious that the variable is a list, rather than a string?
```suggestion
	for _, uls := range userLabels {
```

pkg/distributor/validate_test.go
```diff
@@ -76,22 +81,54 @@ func (vm validateMetadataCfg) MaxMetadataLength(_ string) int {
...
-	reg := prometheus.NewPedanticRegistry()
-	s := newSampleValidationMetrics(reg)
+	t.Parallel()
 
-	var cfg validateLabelsCfg
-	userID := "testUser"
+	const repeatsPerCase = 100
+
+	defaultUserID := "testUserDefault"
+	utf8UserID := "testUserUTF8"
```
aknuds1: [Nit] Helpful to the reader to be able to see they won't change:
```suggestion
	const defaultUserID = "testUserDefault"
	const utf8UserID = "testUserUTF8"
```

pkg/distributor/push.go
```diff
@@ -165,7 +165,12 @@ func handler(
 			http.Error(w, err.Error(), http.StatusBadRequest)
 		}
 		supplier := func() (*mimirpb.WriteRequest, func(), error) {
-			rb := util.NewRequestBuffers(requestBufferPool)
+			var rb *util.RequestBuffers
+			if newRequestBuffers != nil {
```
aknuds1, line 169: Is there any current case where `newRequestBuffers` is nil?

pkg/distributor/validate_test.go
```diff
@@ -292,92 +353,179 @@ func TestValidateLabels(t *testing.T) {
...
+						if strings.Contains(err.Error(), string(id)) {
+							if discardedSamplesValues[id.LabelValue()] == nil {
+								discardedSamplesValues[id.LabelValue()] = map[string]map[string]int{}
+							}
+							var userID string
+							switch scheme {
+							case model.LegacyValidation:
+								userID = defaultUserID
+							case model.UTF8Validation:
+								userID = utf8UserID
```
aknuds1, line 387: For robustness, could you add `default` cases to this scheme switches?
```suggestion
								userID = utf8UserID
							default:
								panic(fmt.Errorf("unhandled name validation scheme: %s", scheme))
```

pkg/distributor/push.go
tcard: Yes, in some other tests, we were previously passing a nil `requestBufferPool` and now a nil `newRequestBuffers`. So after this change the behavior stays the same, ie. we call `util.NewRequestBuffers(nil)`.

pkg/costattribution/testutils/test_utils.go
tcard: Done at f749ac143c.

pkg/distributor/validate_test.go
tcard: Done at f749ac143c.

pkg/distributor/validate_test.go
tcard: Done at f749ac143c.

pkg/distributor/push.go
aknuds1: Right, I must've missed that.

PR #12323, opened 2025-08-06
helm: Support revisionhistorylimit

DrFaust92:

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/operations/helm/charts/mimir-distributed/ci/offline/test-revisionhistorylimit-values.yaml
+++ b/operations/helm/charts/mimir-distributed/ci/offline/test-revisionhistorylimit-values.yaml
@@ -1,0 +1,26 @@
+distributor:
+  revisionHistoryLimit: 3
+
+overrides_exporter:
+  revisionHistoryLimit: 3
+
+querier:
+  revisionHistoryLimit: 3
+
+query_frontend:
+  revisionHistoryLimit: 3
+
+query_scheduler:
+  revisionHistoryLimit: 3
+
+ruler_querier:
+  revisionHistoryLimit: 3
+
+ruler_query_frontend:
+  revisionHistoryLimit: 3
+
+ruler_query_scheduler:
+  revisionHistoryLimit: 3
+
+ruler:
+  revisionHistoryLimit: 3
--- a/operations/helm/tests/test-revisionhistorylimit-values-generated/mimir-distributed/charts/minio/templates/configmap.yaml
+++ b/operations/helm/tests/test-revisionhistorylimit-values-generated/mimir-distributed/charts/minio/templates/configmap.yaml
@@ -1,0 +1,405 @@
+---
+# Source: mimir-distributed/charts/minio/templates/configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test-revisionhistorylimit-values-minio
+  labels:
+    app: minio
+    chart: minio-5.4.0
+    release: test-revisionhistorylimit-values
+    heritage: Helm
+data:
+  initialize: |-
+    #!/bin/sh
+    set -e # Have script exit in the event of a failed command.
+    MC_CONFIG_DIR="/tmp/minio/mc/"
+    MC="/usr/bin/mc --insecure --config-dir ${MC_CONFIG_DIR}"
+    
+    # connectToMinio
+    # Use a check-sleep-check loop to wait for MinIO service to be available
+    connectToMinio() {
+    	SCHEME=$1
+    	ATTEMPTS=0
+    	LIMIT=29 # Allow 30 attempts
+    	set -e   # fail if we can't read the keys.
+    	ACCESS=$(cat /config/rootUser)
+    	SECRET=$(cat /config/rootPassword)
+    	set +e # The connections to minio are allowed to fail.
+    	echo "Connecting to MinIO server: $SCHEME://$MINIO_ENDPOINT:$MINIO_PORT"
+    	MC_COMMAND="${MC} alias set myminio $SCHEME://$MINIO_ENDPOINT:$MINIO_PORT $ACCESS $SECRET"
+    	$MC_COMMAND
+    	STATUS=$?
+    	until [ $STATUS = 0 ]; do
+    		ATTEMPTS=$(expr $ATTEMPTS + 1)
+    		echo \"Failed attempts: $ATTEMPTS\"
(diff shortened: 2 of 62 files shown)


operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12188
 * [ENHANCEMENT] Gateway ingress: Support labels for gateway ingress. #11964
 * [ENHANCEMENT] Store-gateway: Configure options for emptyDir. #11951
+* [ENHANCEMENT] Add support for `revisionhistorylimit` for all Deployments. #12307
```
tacole02:
```suggestion
* [ENHANCEMENT] Add support for `revisionhistorylimit` for all deployments. #12307
```

narqo:
Thank you for the contribution. I don't mind expanding the chart, but I'd like to know the motivation and the problem that you want to solve with that. Specifically,

* why do we set `revisionHistoryLimit: 10` to all workloads — this is the current default in Kubernetes, iirc?
* why would a user want to change the default?

DrFaust92:
> Thank you for the contribution. I don't mind expanding the chart, but I'd like to know the motivation and the problem that you want to solve with that. Specifically,
> 
> * why do we set `revisionHistoryLimit: 10` to all workloads — this is the current default in Kubernetes, iirc?
> * why would a user want to change the default?

* Yes, the default is 10.
* Some cluster operators wish (especially with large clusters) to reduce load on etcd (and argocd tracking objects) and reduce this number .You can see this I a common parameter in charts for things like Argo, Prometheus, and bitnami chart.
So it's not something exotic either

operations/helm/charts/mimir-distributed/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12188
 * [ENHANCEMENT] Gateway ingress: Support labels for gateway ingress. #11964
 * [ENHANCEMENT] Store-gateway: Configure options for emptyDir. #11951
+* [ENHANCEMENT] Add support for `revisionhistorylimit` for all deployments. #12307
```
narqo: The reference is expected to be the PR, that introduces the changes

```suggestion
* [ENHANCEMENT] Add support for `revisionhistorylimit` for all deployments. #12323
```

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -566,6 +566,7 @@ alertmanager:
   # Otherwise each zone starts `ceil(replicas / number_of_zones)` number of pods.
   #   E.g. if 'replicas' is set to 4 and there are 3 zones, then 4/3=1.33 and after rounding up it means 2 pods per zone are started.
   replicas: 1
+  revisionHistoryLimit: 10
```
narqo: Could we avoid having the magic number, that the most users _are likely_ don't care, in the default values?

That is, I'd rather define the `revisionHistoryLimit: null` in the defaults (for documentation purposes). And then updated the resources' templates to override their `.spec.revisionHistoryLimit` only if it was explicitly passed by the user. For the reference, see how the `priorityClassName` is managed by the chart.

DrFaust92: ive tested locally if null gives same effect as default and it does. ive changed to null

operations/helm/charts/mimir-distributed/templates/gateway/gateway-dep.yaml
```diff
@@ -14,6 +14,8 @@ spec:
 {{- if not .autoscaling.enabled }}
   replicas: {{ .replicas }}
 {{- end }}
+  revisionHistoryLimit: {{ .revisionHistoryLimit }}
+  {{- if not $isGEMGateway }}
```
narqo: I'm not sure this `if` must be here for these changes. It looks like a typo to me.

operations/helm/charts/mimir-distributed/templates/nginx/nginx-dep.yaml
```diff
@@ -13,6 +13,8 @@ spec:
 {{- if not .Values.nginx.autoscaling.enabled }}
   replicas: {{ .Values.nginx.replicas }}
 {{- end }}
+  revisionHistoryLimit: {{ .Values.nginx.revisionHistoryLimit }}
+  {{- end }}
```
narqo: Also a typo? ✂️

operations/helm/charts/mimir-distributed/templates/gateway/gateway-dep.yaml
DrFaust92: fixed

operations/helm/charts/mimir-distributed/templates/nginx/nginx-dep.yaml
DrFaust92: fixed

DrFaust92:
> operations/helm/charts/mimir-distributed/ci/offline

Ive made a change to use `with` otherwise the `make build-helm-tests` command causes a huge diff in manifests.
Ive added an example of override, creating a dedicated values files with all options seems very excessive

operations/helm/charts/mimir-distributed/small.yaml
```diff
@@ -25,6 +25,7 @@
 # object storage service such as S3 or GCS.
 
 alertmanager:
+  revisionHistoryLimit: 5
```
narqo: `small.yaml` **is not** a test file. It is a preset, [the documentation][1] suggests for users, whose clusters target ingestion of ~1M series. 

Please, revert the changes here, and add the value to a test, as I suggested previously.

[1]: https://grafana.com/docs/helm-charts/mimir-distributed/latest/run-production-environment-with-helm/#plan-capacity

DrFaust92: @narqo done, please LMK if this look right to you

narqo (approved):
Looks good to me now. Thank you 🔥

PR #12324, opened 2025-08-06
docs: draft topic for mimir query best practices

tacole02:
#### What this PR does

This PR creates a new topic under "Query metric data from Mimir" called "Best practices for querying Grafana Mimir". It overviews some general best practices for writing database queries, both in general and for Mimir specifically. 

The goal of this topic is to provide Mimir users opinionated guidances on how they can structure their queries to optimize performance and results.

Fixes https://github.com/grafana/mimir-squad/issues/3122

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.


tacole02:
@eamonryan let me know your thoughts!

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Best practices for querying Grafana Mimir
+
+The way you write queries for Grafana Mimir affects the speed and quality of your results. Follow these best practices to ensure optimal performance and reliability for queries.
+
+## Use precise label selectors
+
+To reduce the initial dataset size, start with the most specific label selectors. For example, if you have labels like `namespace` and `app_name` where `app_name` is more specific, lead with `app_name="myapp"` rather than starting with the broader `namespace` selector.
```
charleskorn: The order of matchers does not really matter (and @dimitarvdimitrov and @chencs are working on some changes where we'll automatically pick the best order to evaluate matchers in).

charleskorn: Terminology nit: `foo{namespace="bar", app="baz"}` is a selector, while the individual terms within it (eg. `namespace="bar"`) are matchers.

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+
+To further reduce the number of series that require processing, include as many label selectors as possible for your query.
```
charleskorn: This is true up to a point: if extra selectors don't help filter the amount of data, then they can be counter-productive.

eg. if `metric{cluster="prod", region="us"}` returns the same results as `metric{cluster="prod", region="us", provider="aws"}`, then the form with fewer selectors is slightly more efficient.

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+
+Limit your query to a specific time period to reduce the number of metrics Mimir needs to process. As a best practice, query the shortest feasible time range to narrow down your results. Larger time ranges require more computing resources and increase query latency.
+
+Follow these guidelines for setting a time range based on your query type:
+
+- Dashboards: For dashboards, use a time range that matches the typical viewing window. While users can zoom out to view more history, the default view should be efficient and focused on recent, relevant data.
+- Alerting rules: Alerting rules should almost always operate on short, recent time windows to detect current states or recent trends, rather than long-term historical patterns. Timely alerts depend on the quick evaluations of fresh data.
+- Ad-hoc analysis: When exploring data, start with small time ranges and gradually expand if more historical context is required for your investigation. This iterative approach prevents unnecessary strain on the system during exploratory phases.
+
+If you don't specify a time range in your query, consider using the `-store.max-labels-query-length` CLI flag to limit the maximum time range.
```
charleskorn: Specifying a time range is always required for range queries, and defaults to the current time for instant queries, which seem to be the types of queries we're talking about here. The `-querier.max-query-lookback` and `-querier.max-partial-query-length` CLI flags limit how far back and how long a query's time range can be.

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+
+Additionally, choose a step interval that matches your requirements for running the query. Smaller step intervals increase a query's cost but provide higher resolution.
+
+## Match your query time range to the evaluation frequency
+
+Align the time range of your query with how frequently the query is evaluated or how current the data needs to be. For instance, if a query runs every minute, such as for an alerting rule or a frequently refreshing dashboard panel, there's no need for it to look back for a long time period, such as 30 days.
+
+Queries with very long lookback periods can be computationally expensive. They require Mimir to process a large volume of historical data, even if only the latest value is of interest.
```
charleskorn: It's not clear to me what this section is referring to - is this talking about expressions like `rate(metric[30d])`?

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+- Improved query speed: Queries that are pre-filtered are generally faster to run and return results.
+
+Use these strategies to implement early filtering:
+
+- Use label selectors: Leverage precise label selectors within your PromQL query to narrow down the dataset from the start. For more information, refer to [Use Precise Label Selectors](#use-precise-label-selectors).
+- Apply functions and aggregations: If you need to transform or aggregate data, do so within the query. For example, `sum by (job) (metric_name)` is more efficient than retrieving all `metric_name` series and then summing them in Grafana.
+
+## Avoid high cardinality operations
+
+Avoid high cardinality operations to avoid performance issues or latency with your queries. High cardinality operations typically involve processing exceptionally large datasets.
```
charleskorn: This seems like a duplicate of the "filter early in the query" and "use precise label selectors" sections.

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+High cardinality typically results from the following circumstances:
+
+- Using a large number of labels
+- Attaching a large number of values to a label.
+
+To reduce cardinality in your queries, limit operations that involve many labels, such as the `group_by` function.
+
+## Use recording rules
+
+If you need to run a particularly expensive or complex query, consider creating a recording rule to minimize its load. Recording rules in Prometheus involve running queries at a predetermined time and then precomputing the query results to save for faster retrieval later.
```
charleskorn:
```suggestion
If you need to run a particularly expensive or complex query, consider creating a recording rule to minimize its load. Recording rules in Mimir involve precomputing queries at a predetermined time and then saving the results for faster retrieval later.
```

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,84 @@
...
+
+Recording rules are useful for the following types of queries:
+
+- Dashboard queries that run frequently
+- Complex aggregations across high-cardinality metrics
+- Queries spanning long time ranges
```
charleskorn, line 67: These are both valid use cases for recording rules, but there's a trade-off here: if the recording rule results are only used infrequently, then they'll be paying the cost of evaluating and storing the recording rule results when they're not used. Perhaps we need to clarify this to emphasise that the best candidates for recording rules are queries that are both expensive and evaluated relatively frequently?

docs/sources/mimir/query/query-best-practices.md
tacole02: Yes. This is basically saying, make sure the frequency you run a query makes sense with the look back period. If you are evaluating data from the past 30 days, like in the above example, it makes no sense to run that query at a super frequent interval, such as every minute. You are running an expensive query that isn't brining meaningful or insightful results.

tacole02:
Hi @charleskorn , thanks so much for taking a look! I've incorporated your feedback. Would you mind looking this over one more time?

docs/sources/mimir/query/query-best-practices.md
charleskorn: OK, makes sense. Perhaps using the word "match" in the heading and "align" in the first sentence confused me - to me, they suggest the two things should be roughly equal. 

Maybe using a different word like "consider" would be less confusing (eg. "consider your query time range and evaluation frequency")?

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Best practices for querying Grafana Mimir
+
+The way you write queries for Grafana Mimir affects the speed and quality of your results. Follow these best practices to ensure optimal performance and reliability for queries.
+
+## Use label selectors
+
+To reduce the initial size of your dataset, apply label selectors to your query. Label selectors help you to filter your data based on its associated labels. For example, you if you have a label called `app_name`, you can append `app_name="myapp"` to your query.
```
charleskorn:
```suggestion
## Use label matchers

To reduce the initial size of your dataset, apply label matchers to your query. Label matchers help you to filter your data based on its associated labels. For example, you if you have a label called `app_name`, you can append `app_name="myapp"` to your query.
```

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+
+Whenever possible, use an exact label match, such as `label="value"`, instead of a regular expression, such as `label=~"pattern"`. Regular expressions are more computationally expensive.
+
+For more information about how to query metric labels, refer to [Query metric labels](https://grafana.com/docs/mimir/<MIMIR_VERSION>/query/query-metric-labels/).
+
+## Narrow down your time range
+
+Limit your query to a specific time period to reduce the number of metrics Mimir needs to process. As a best practice, query the shortest feasible time range to narrow down your results. Larger time ranges require more computing resources and increase query latency.
```
charleskorn:
```suggestion
Limit your query to a specific time period to reduce the number of metrics and samples Mimir needs to process. As a best practice, query the shortest feasible time range to narrow down your results. Larger time ranges require more computing resources and increase query latency.
```

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+- Ad-hoc analysis: When exploring data, start with small time ranges and gradually expand if more historical context is required for your investigation. This iterative approach prevents unnecessary strain on the system during exploratory phases.
+
+Additionally, choose a step interval that matches your requirements for running the query. Smaller step intervals increase a query's cost but provide higher resolution.
+
+## Match your query time range to the evaluation frequency
+
+Align the time range of your query with how frequently the query is evaluated or how current the data needs to be. For instance, if a query runs every minute, such as for an alerting rule or a frequently refreshing dashboard panel, there's no need for it to look back over a long time period, such as 30 days.
+
+Queries with very long look-back periods can be computationally expensive. They require Mimir to process a large volume of historical data, even if only the latest value is of interest.
```
charleskorn: Might be good to add an example here:

```suggestion
Queries with very long look-back periods, such as `rate(my_metric[30d])`, can be computationally expensive. They require Mimir to process a large volume of historical data, even if only the latest value is of interest.
```

docs/sources/mimir/query/query-best-practices.md
```diff
@@ -0,0 +1,71 @@
...
+
+Filtering data in the query provides the following benefits:
+
+- Reduced data transfer: Filtering at the query level means Mimir sends only the data you need to Grafana, which reduces network overhead.
+- Lower processing load: Mimir can apply filters at the storage layer, allowing it to discard irrelevant time series and data points before they are retrieved, processed, and aggregated. This significantly reduces the computational load on the Mimir cluster.
+- Improved query speed: Queries that are pre-filtered are generally faster to run and return results.
+
+Use these strategies to implement early filtering:
+
+- Use label selectors: Leverage precise label selectors within your PromQL query to narrow down the dataset from the start. For more information, refer to [Use Precise Label Selectors](#use-precise-label-selectors).
```
charleskorn: I think this heading has changed now:

```suggestion
- Use label selectors: Leverage precise label selectors within your PromQL query to narrow down the dataset from the start. For more information, refer to [Use Label Selectors](#use-label-selectors).
```

docs/sources/mimir/query/query-best-practices.md
tacole02: That totally makes sense! I'll adjust.

tacole02:
Thanks so much, @charleskorn ! I've incorporated your latest round of feedback. Would you mind taking one more look?

PR #12325, opened 2025-08-07
golangci-lint: Re-enable staticcheck defaults

aknuds1:
#### What this PR does

golangci-lint was inadvertently configured to disable _all_ staticcheck checks except for `ST1016` in #11691. I'm re-enabling the default staticcheck checks, _plus_ ST1016 in this PR.

I noticed due to the fact that deprecated symbols weren't reported.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/mimirpb/symbols.go
```diff
@@ -68,6 +68,7 @@ func reuseSymbolsSlice(s []string) {
 	for i := range s {
 		s[i] = ""
 	}
+	//nolint:staticcheck
```
charleskorn: Could you please add more explanation about why you're disabling the linter here? And is it possible to scope this just to a particular rule, rather than staticcheck as a whole?

charleskorn (commented):
Sorry about this - the changes generally look good to me modulo one suggestion:

pkg/mimirpb/symbols.go
aknuds1: I'll try.

aknuds1: AFAICT, it's not possible to inline disable a specific staticcheck rule. Grepping the code base backs this up. I added an explanation.

charleskorn (approved):

PR #12326, opened 2025-08-07
Move streamingpromql engine opts into its own package

zenador:
#### What this PR does

Move streamingpromql engine opts into its own package so it can be imported into other packages like `ast` without causing a cycle. Not sure if this is the best approach or if there's a better package name, please let me know.

Will unblock https://github.com/grafana/mimir/pull/12303

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

Not a user-facing change, more like a refactor.


juliusmh:
I'm not a fan of `opts`.. It's a very common abbreviation for `options` and also often used as such in our and prometheus source code. Also this should not be confused with query options, and simply 'opts' is ambiguous.

What about calling the package 
- config, as this is configuring the engine
- options, less likely to collide
- engineopts, my favorite, it's short and descriptive

wdyt?

PR #12333, opened 2025-08-08
Add postings cache sharing and invalidation to ingesters

jhalterman:
Adds postings cache sharing and invalidation support to ingesters, following https://github.com/grafana/mimir-prometheus/pull/944. When sharing is enabled, cache entries are distinguished by tenant id.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


cmd/mimir/config-descriptor.json
```diff
@@ -10382,6 +10382,39 @@
...
+              "fieldDefaultValue": false,
+              "fieldFlag": "blocks-storage.tsdb.head-postings-for-matchers-cache-invalidation",
+              "fieldType": "boolean",
+              "fieldCategory": "experimental"
+            },
+            {
+              "kind": "field",
+              "name": "head_postings_for_matchers_cache_versions",
+              "required": false,
+              "desc": "The number of metric versions to store in each cache when invalidation is enabled.",
```
dimitarvdimitrov: i think we talked about in the dskit PR that this should be comparable to the number of different metric names. i would clarify here that it should be comparable to the number of metric names across all tenants in an ingester.

cmd/mimir/config-descriptor.json
```diff
@@ -10382,6 +10382,39 @@
...
+              "fieldDefaultValue": false,
+              "fieldFlag": "blocks-storage.tsdb.shared-postings-for-matchers-cache",
+              "fieldType": "boolean",
+              "fieldCategory": "experimental"
+            },
+            {
+              "kind": "field",
+              "name": "head_postings_for_matchers_cache_invalidation",
+              "required": false,
+              "desc": "Whether head block postings should be tracked and invalidated when they change.",
```
dimitarvdimitrov: can you clarify what happens if they aren't invalidated? or perhaps mention that if invalidation is enabled, then the TTL of records can be much higher

cmd/mimir/config-descriptor.json
```diff
@@ -10382,6 +10382,39 @@
               "fieldType": "int",
               "fieldCategory": "experimental"
             },
+            {
+              "kind": "field",
+              "name": "shared_postings_for_matchers_cache",
+              "required": false,
+              "desc": "Whether postings for matchers cache should be shared across blocks, as opposed to instantiated per block. With a shared cache, one cache is created for head blocks, and one for compacted blocks.",
```
dimitarvdimitrov, line 10389:
```suggestion
              "desc": "Whether postings for matchers cache should be shared across blocks and tenants, as opposed to instantiated per block. With a shared cache, one cache is created for head blocks, and one for compacted blocks for each tenant.",
```

not sure mine is a better wording, but just pointing out that we should clarify whether it's per-tenant or per-ingester

pkg/ingester/user_tsdb.go
```diff
@@ -351,6 +358,9 @@ func (u *userTSDB) PostDeletion(metrics map[chunks.HeadSeriesRef]labels.Labels)
 			continue
 		}
 		u.seriesInMetric.decreaseSeriesForMetric(metricName)
+		if u.postingsCache != nil {
+			u.postingsCache.InvalidateMetric(u.userID, metricName)
+		}
```
dimitarvdimitrov: the code which works with postings is already made to work with series IDs which do not exist. We should be able to get away without invalidating on delete


https://github.com/grafana/mimir/blob/b8396cd92300d053a58fd1e0085da98e5f92aee6/vendor/github.com/prometheus/prometheus/tsdb/querier.go#L635-L637

dimitarvdimitrov: here's the same for labelvalues 

https://github.com/grafana/mimir/blob/b8396cd92300d053a58fd1e0085da98e5f92aee6/vendor/github.com/prometheus/prometheus/tsdb/querier.go#L522-L524


and for active series 

https://github.com/grafana/mimir/blob/699a122aba83da75ccef884d63af468da4f5127e/pkg/ingester/active_series.go#L74-L77

cmd/mimir/config-descriptor.json
jhalterman: Good suggestion. Added per ingester, but also clarified this to mention it's the size of the metrics cache.

pkg/ingester/user_tsdb.go
jhalterman: Good suggestion.

PR #12338, opened 2025-08-08
Document yoloStrings and buffer pooling in contributing/README.md

tcard:
#### What this PR does

Introduces a section in our contributing guidelines in the hopes of raise awareness of yoloStrings and its perils.

This also updates CLAUDE.md, which is a symlink.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/internal/contributing/README.md
+++ b/docs/internal/contributing/README.md
@@ -105,6 +105,22 @@ You have to commit the changes to `go.mod` and `go.sum` before submitting the pu
 
 Please see the dedicated "[Design patterns and Code conventions](design-patterns-and-conventions.md)" page.
 
+## ⚠️ Unsafe memory tricks
+
+For performance, buffers used for decompressing gRPC requests into are kept in a pool shared between requests.
+
+Also, when unmarshaling those byte buffers into structured Go values, we construct strings that aren't actually references to immutable memory, but to those same shared buffers. The codebase refers to this trick as `yoloString`, and in some places (but not exhaustively), such strings are denoted as type `unsafeMutableString`, which is an alias of `string`.
+
+This means that innocent-looking string values that outlive the window between taking a request buffer from the pool and releasing it back may actually **refer to another request's buffer**. Given Mimir is multi-tenant, it's likely to result in **cross-tenant data leakage**.
+
+Take into account that copying `string`s around actually moves _references_ around. Make sure strings that must outlive the handling window (e. g. for error messages) are deep-copied, e. g. with `strings.Clone` or `strings.Builder`. Note that, if those strings are within a slice, it is **not enough** to clone the slice with `slices.Clone` or similar, as only the references will be cloned, not the underlying bytes in memory. The same applies if there are inside a struct or array.
+
+There are currently no guardrails around this, beyond tests.
+
+Some (but maybe not all) specific critical areas are:
+
+- In the distributor, such strings must not outlive the `PushFunc` passed to `Handler`.
+
 ## Documentation
 
 The Grafana Mimir documentation and the Helm chart _documentation_ for Mimir and GEM are compiled and published to [https://grafana.com/docs/mimir/latest/](https://grafana.com/docs/mimir/latest/) and [https://grafana.com/docs/helm-charts/mimir-distributed/latest/](https://grafana.com/docs/helm-charts/mimir-distributed/latest/). Run `make docs` to build and serve the documentation locally.
--- a/docs/internal/contributing/README.md
+++ b/docs/internal/contributing/README.md
@@ -119,7 +119,8 @@ There are currently no guardrails around this, beyond tests.
 
 Some (but maybe not all) specific critical areas are:
 
-- In the distributor, such strings must not outlive the `PushFunc` passed to `Handler`.
+- In the distributor, those strings must not outlive the `PushFunc` passed to `Handler`.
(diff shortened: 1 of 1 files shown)


tcard:
I asked Claude Sonnet 4 to review [the offending PR](https://github.com/grafana/mimir/pull/11889) in the hopes this guide would be enough for it to detect the issue. Alas, it was very happy with my `slices.Clone`. Oh well.

Screenshot 2025-08-08 at 13 56 28

docs/internal/contributing/README.md
```diff
@@ -105,6 +105,22 @@ You have to commit the changes to `go.mod` and `go.sum` before submitting the pu
 
 Please see the dedicated "[Design patterns and Code conventions](design-patterns-and-conventions.md)" page.
 
+## ⚠️ Unsafe memory tricks
+
+For performance, buffers used for decompressing gRPC requests into are kept in a pool shared between requests.
```
aknuds1, line 110: Not sure if this sentence makes sense? We don't have any choice in the matter, the buffers are pooled by gRPC itself, right. The sentence makes it sound as if we choose to pool the buffers for performance.

aknuds1 (commented):
Please see comment.

docs/internal/contributing/README.md
tcard: That's what we initially thought, but investigating deeper, it was our own pooling (not gRPC's) from [RequestBuffers](https://github.com/grafana/mimir/blob/8aeca96620cab68143176b1454deb4c6fdd934f1/pkg/util/requestbuffers.go) which causes the issue.

aknuds1: You're right @tcard. I misread decompressing as deserializing.

docs/internal/contributing/README.md
```diff
@@ -105,6 +105,22 @@ You have to commit the changes to `go.mod` and `go.sum` before submitting the pu
...
+
+Also, when unmarshaling those byte buffers into structured Go values, we construct strings that aren't actually references to immutable memory, but to those same shared buffers. The codebase refers to this trick as `yoloString`, and in some places (but not exhaustively), such strings are denoted as type `unsafeMutableString`, which is an alias of `string`.
```
aknuds1, line 112: Can you point to an example where we construct a `yoloString` into a _decompression_ buffer? I just know that we have `yoloString`s into _deserialization_ buffers.

docs/internal/contributing/README.md
```diff
@@ -105,6 +105,22 @@ You have to commit the changes to `go.mod` and `go.sum` before submitting the pu
...
+
+This means that innocent-looking string values that outlive the window between taking a request buffer from the pool and releasing it back may actually **refer to another request's buffer**. Given Mimir is multi-tenant, it's likely to result in **cross-tenant data leakage**.
+
+Take into account that copying `string`s around actually moves _references_ around. Make sure strings that must outlive the handling window (e. g. for error messages) are deep-copied, e. g. with `strings.Clone` or `strings.Builder`. Note that, if those strings are within a slice, it is **not enough** to clone the slice with `slices.Clone` or similar, as only the references will be cloned, not the underlying bytes in memory. The same applies if there are inside a struct or array.
+
+There are currently no guardrails around this, beyond tests.
```
korniltsev, line 118: > There are currently no guardrails around this

random idea not related to the PR, but to the whole situation(so feel free to ignore): 

Can we add guardrails somehow? 

One way to do this could be by  using mmaped memory instead of a pooled buffer.
(Maybe for tests only and with a custom go build tag, or even in production, but highly sampled (once in N milllions). This would crash the test with SIGSEGV if someone preserves a non-cloned string pointer into a buffer that was protected with PROT_NONE.

docs/internal/contributing/README.md
tcard: Ultimately, those uncompressed buffers are passed to [LabelAdapter.Unmarshal](https://github.com/grafana/mimir/blob/8aeca96620cab68143176b1454deb4c6fdd934f1/pkg/mimirpb/timeseries.go#L362), where we call `yoloString` to get strings pointing into the source buffers.

aknuds1: `yoloString` in `LabelAdapter.Unmarshal` references the `dAtA` argument to `LabelAdapter.Unmarshal`. Isn't `dAtA` a gRPC pooled buffer?

tcard: `LabelAdapter` is wrapped by `PreallocWriteRequest`.

`PreallocWriteRequest` is passed to `util.ProtoReader` as a destination parameter [here](https://github.com/grafana/mimir/blob/ec4af8e88b6b0898276f5a413ec8b037bf1f8421/pkg/distributor/push.go#L100).

Because `PreallocWriteRequest` implements `proto.Unmarshaler`, `util.ProtoReader` wll (after having called `decompressRequest` into a buffer from `RequestBuffers`) call `PreallocWriteRequest.Unmarshal` directly, [here](https://github.com/grafana/mimir/blob/ec4af8e88b6b0898276f5a413ec8b037bf1f8421/pkg/util/http.go#L161-L165).

I think by doing this we're bypassing gRPC's internal pooling entirely.

aknuds1: Ah that explains it, thanks @tcard. I don't even remember seeing that trick. Could you maybe mention it in your documentation here?

tcard: I've expanded a bit into [34f5c3a](https://github.com/grafana/mimir/pull/12338/commits/34f5c3ab31a98a962e99a762afd22859c5ca16b0), although I don't think this is the place to go too deep into it.

PR #11983, opened 2025-07-04
Helm: Allow users to override automatically calculated GOMAXPROCS

narqo:
## Summary

As raised in https://github.com/grafana/mimir/issues/6467 right now users have no way to override the default env variables some components define. This PR improves that.

• Add reusable `mimir.lib.containerEnv` helper to merge environment variables with proper precedence
• Enable GOMAXPROCS override for distributor, ingester, querier, ruler-querier, and store-gateway components  
• Maintain backward compatibility with existing automatic `GOMAXPROCS` calculations
• Add comprehensive test coverage for override functionality

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #10860, opened 2025-03-11
Idea: Automatically set GOMEMLIMIT for Mimir components


dewjam:
### What is the problem you are trying to solve?

Right now, the mimir-distributed Helm chart doesn’t set a default GOMEMLIMIT value for all Mimir components. Additionally, manually setting GOMEMLIMIT via Helm can be cumbersome in complex deployments where multiple stacks have different resource allocations.

Ideally, GOMEMLIMIT should be automatically set based on a predefined ratio of a pod’s memory limit, with an option to override that ratio if needed. This would help ensure better memory management while reducing the risk of OOMKills.

### Which solution do you envision (roughly)?

Similar to how [Alloy](https://grafana.com/docs/alloy/latest/reference/cli/environment-variables/#automatically-set-gomemlimit) handles memory limits, I’d suggest that Mimir use a package like [automemlimit](https://github.com/KimMachineGun/automemlimit). This would automatically set GOMEMLIMIT to 90% of the container’s memory limit unless explicitly overridden via an environment variable.

Example code:
```
memLimit, err := memlimit.SetGoMemLimitWithOpts()
if err != nil {
	level.Error(util_log.Logger).Log("msg", "Failed to set GOMEMLIMIT", "err", err)
} else {
	level.Info(util_log.Logger).Log("msg", "Setting GOMEMLIMIT using automemlimit", "memory_limit_bytes", memLimit)
}
```

### Have you considered any alternatives?

**Pass in ENV via Helm**
Manually setting GOMEMLIMIT via Helm can be cumbersome in complex deployments where multiple stacks have different resource allocations.  Improving the mimir-distributed Helm chart to set sane values for GOMEMLIMIT could address this issue as well.

**Mutating Webhook**
Use a Kyverno policy to mutate pods and inject the GOMEMLIMIT env.  

### How long do you think this would take to be developed?

Small (<= 1 month dev)

narqo:
In a simple case, where one only wants to derive the `GOMEMLIMIT` from the container's resource limit, one can define the env as in the snippet below. This can be simpler then bringing a dependency to the application, that reads the cgroups from inside the container.

```yaml
env:
- name: GOMEMLIMIT
  valueFrom:
    resourceFieldRef:
      resource: limits.memory
      divisor: 1
```

[...]

PR #12341, opened 2025-08-08
feat(helm): support goMemLimitPercentage for distributor, ingester, querier, query-frontend, store-gateway, and compactor

shlomitubul:
#### What this PR does
Adds support for configuring GOMEMLIMIT via goMemLimitPercentage for all Mimir components, with per-component overrides and global fallback.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


tacole02 (approved):
Changelog looks good. A maintainer will still need to do code review.

dimitarvdimitrov:
thanks for the contribution

i think an easier approach which doesn't require knowledge of the go runtime and experience operating go programs is to set GOMEMLIMIT to some percentage, similar to how it's done for the store-gateway. This approach feels like it's adding too much complexity to the chart (granted, Helm doesn't make it easy) for the potential gain and flexibility.

shlomitubul:
> thanks for the contribution
> 
> i think an easier approach which doesn't require knowledge of the go runtime and experience operating go programs is to set GOMEMLIMIT to some percentage, similar to how it's done for the store-gateway. This approach feels like it's adding too much complexity to the chart (granted, Helm doesn't make it easy) for the potential gain and flexibility.

@dimitarvdimitrov While I agree that simplicity should be the default, I don’t see how we can simplify this without hardcoding knobs that, in my view, ought to remain tunable.
also, the current implementation in store-gateway is why I opened this PR in first place as setting GOMEMLIMIT to the pod’s granted memory value didn’t work in my case

dimitarvdimitrov:
fair enough. i wouldn't fight too much for this, it's a reasonable knob to tune. I've only got two remarks
1. Can we put the template as a named template somewhere in _helpers.tpl so that it's not repeated for each component?
2. Can you come up with some reasonable default? That's a bit harder. I'd like to see some results with the value + a description of the workload of your Mimir cluster to help understand if your use case is a reasonable average workload that others will also use.
	a. An alternative to this is to just have the `goMemLimitPercentage` default to unset for all components. It's easier but fewer people would benefit from it.

dimitarvdimitrov:

narqo:
I want to push back on the proposed changes. Specifically, this part

>  the current implementation in store-gateway is why I opened this PR in first place as setting GOMEMLIMIT to the pod’s granted memory value didn’t work in my case

Since https://github.com/grafana/mimir/pull/11983 we allow users to override the default environment variables, including the default values for `GOMEMLIMIT` or `GOMAXPROCS` (see [example in the tests][1]). I expect this to allow solving the mentioned problem.

[1]: https://github.com/grafana/mimir/blob/433e9b04cf9f7feb7970963e714c74ef26af4470/operations/helm/charts/mimir-distributed/ci/offline/test-gomaxprocs-override-values.yaml#L33-L42

frittentheke:
This PR could be linked to issue https://github.com/grafana/mimir/issues/10860 which is exactly about this topic.

As I said there and also in my feature request for Grafana Loki (https://github.com/grafana/loki/issues/19586), I believe the setting of `GOMEMLIMIT`  should be done via https://github.com/KimMachineGun/automemlimit and not some complex templating and calculations via Helm.

DrFaust92:
BTW, this was implemented in both tempo and Loki helm charts already

frittentheke:
> BTW, this was implemented in both tempo and Loki helm charts already

@DrFaust92 ... do you have a link or other pointers to where this was implemented - just for people finding this issue here.

DrFaust92:
@frittentheke 
https://github.com/grafana-community/helm-charts/pull/327
https://github.com/grafana-community/helm-charts/pull/230

PR #12329, opened 2025-08-07
distributor: Optimize RW1 -> RW2 translation

alexweav:
#### What this PR does

This PR optimizes RW1 to RW2 translation primarily for use in ingest-storage.
It speeds up symbolization, adds pooling for the two highest sources of allocation, and adds a benchmark.

<details>

<summary>benchmark results</summary>

```
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/storage/ingest
cpu: AMD Ryzen 7 PRO 5850U with Radeon Graphics
                                             │  before.txt  │              after.txt               │
                                             │    sec/op    │    sec/op     vs base                │
RecordSerializer/v2_serialize_(full_flow)-16   404.8µ ± ∞ ¹   407.7µ ± ∞ ¹       ~ (p=1.000 n=1) ²
RecordSerializer/v1_serialize_(full_flow)-16   172.5µ ± ∞ ¹   172.1µ ± ∞ ¹       ~ (p=1.000 n=1) ²
RecordSerializer/v1_->_v2_convert-16           280.1µ ± ∞ ¹   280.6µ ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                        269.5µ         270.0µ        +0.21%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                             │   before.txt   │               after.txt                │
                                             │      B/op      │     B/op       vs base                 │
RecordSerializer/v2_serialize_(full_flow)-16   335.66Ki ± ∞ ¹   82.37Ki ± ∞ ¹        ~ (p=1.000 n=1) ²
RecordSerializer/v1_serialize_(full_flow)-16    80.23Ki ± ∞ ¹   80.23Ki ± ∞ ¹        ~ (p=1.000 n=1) ³
RecordSerializer/v1_->_v2_convert-16           263516.0 ± ∞ ¹     258.0 ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                         190.7Ki         11.85Ki        -93.78%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
³ all samples are equal

                                             │   before.txt   │              after.txt               │
                                             │   allocs/op    │  allocs/op   vs base                 │
RecordSerializer/v2_serialize_(full_flow)-16    2010.00 ± ∞ ¹   11.00 ± ∞ ¹        ~ (p=1.000 n=1) ²
[...]

PR #12342, opened 2025-08-08
mimirpb: Associate metadata to series when possible during RW2 conversion

alexweav:
#### What this PR does

RW2 attaches metric metadata to series, where RW1 carries metadata separately.
Previously, we took a fast path, where we always generate new series just to carry each metadata.
This PR changes the conversion to metadata into the corresponding series, if both are present in the same request.

The main benefit is on requests that are already RW2 (and tend to carry a mixture of series and metadata). We squeeze some additional bytes out of the new format by not repeating the label name references two times per metric family, plus dropping extra `repeated TimeSeriesRW2` separators in protobuf. It does not change data-only or metadata-only pushes.

We now do extra work in the distributor to match these elements up. Previously we theorized this wasn't worth it, but now after analyzing typical distributor profile data, I expect the extra work here to not be as impactful on perf or TCO as once thought. The upside is we keep the record size absolutely minimized, as clients shift over to send RW2 push requests.

contrib: https://github.com/grafana/mimir-squad/issues/2253

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/prealloc_rw2.go
+++ b/pkg/mimirpb/prealloc_rw2.go
@@ -35,7 +35,7 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	symbols.ConfigureCommonSymbols(offset, commonSymbols)
 
 	var metadataMap map[string]*MetricMetadata
-	if len(rw1.Metadata) > 0 {
+	if len(rw1.Metadata) > 0 && len(rw1.Timeseries) > 0 {
 		metadataMap = make(map[string]*MetricMetadata, len(rw1.Metadata))
 		for _, meta := range rw1.Metadata {
 			metadataMap[meta.MetricFamilyName] = meta
@@ -72,11 +72,14 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	}
 
 	// If there are extra metadata not associated with any timeseries, we fabricate an empty timeseries to carry it.
-	// We never replicate metadata if there are multiple series with the same name.
 	for _, meta := range rw1.Metadata {
-		if _, ok := metadataMap[meta.MetricFamilyName]; !ok {
-			continue
+		// If we are matching metadata to series, and we already matched this metadata, we can skip it.
+		if metadataMap != nil {
+			if _, ok := metadataMap[meta.MetricFamilyName]; !ok {
+				continue
+			}
 		}
+
 		labelsRefs := []uint32{symbols.Symbolize(labels.MetricName), symbols.Symbolize(meta.MetricFamilyName)}
 		rw2meta := FromMetricMetadataToMetadataRW2(meta, symbols)
 		rw2Timeseries = append(rw2Timeseries, TimeSeriesRW2{


pkg/mimirpb/prealloc_rw2.go
```diff
@@ -30,27 +34,50 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	defer reuseSymbolsTable(symbols)
 	symbols.ConfigureCommonSymbols(offset, commonSymbols)
 
+	var metadataMap map[string]*MetricMetadata
+	if len(rw1.Metadata) > 0 {
```
alexweav, line 38: Throughout we take extra care to never allocate or interact with the map if the request has no metadata.
This keeps the impact on regular series pushes low.

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -30,27 +34,50 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
...
+		metadataMap = make(map[string]*MetricMetadata, len(rw1.Metadata))
```
alexweav, line 39: We have an opportunity to pool this map.
The entire RW2 conversion flow is not deeply optimized. I would rather append the pooling to https://github.com/grafana/mimir/pull/12329 rather than this PR.

I also don't think pooling would have high impact after measuring/profiling the above changes, and pooling adds an element of memory unsafety, so I've left it out for now. Requests with no metadata don't allocate any extra anyway.

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -30,27 +34,50 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
...
+		metricName := ""
 		for i := range ts.Labels {
+			if ts.Labels[i].Name == labels.MetricName {
+				metricName = ts.Labels[i].Value
+			}
 			refs = append(refs, symbols.Symbolize(ts.Labels[i].Name), symbols.Symbolize(ts.Labels[i].Value))
 		}
 
+		var metadata MetadataRW2
+		if len(rw1.Metadata) > 0 {
```
pracucci: [nit] Isn't more accurate checking `metadataMap` instead?
```suggestion
		if len(metadataMap) > 0 {
```

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -30,27 +34,50 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
...
+			if rw1Metadata, ok := metadataMap[metricName]; ok {
+				metadata = FromMetricMetadataToMetadataRW2(rw1Metadata, symbols)
+			}
+			delete(metadataMap, metricName)
```
pracucci: [nit] I think we can just call `delete()` if there was a match:

```suggestion
			    delete(metadataMap, metricName)
			}
```

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -30,27 +34,50 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
...
 			Histograms:       ts.Histograms,
 			Exemplars:        FromExemplarsToExemplarsRW2(ts.Exemplars, symbols),
-			Metadata:         MetadataRW2{},
+			Metadata:         metadata,
 			CreatedTimestamp: ts.CreatedTimestamp,
 		})
 	}
 
 	// Represent metadata as extra, empty timeseries rather than attaching it to existing series.
 	// We never replicate metadata if there are multiple series with the same name, and it removes the requirement to match up the metadata to the right series.
```
pracucci: [nit] Can you update this comment, to explain that here we append the remaining metadata belonging to series that didn't have any sample?

pkg/mimirpb/prealloc_rw2.go
pracucci: Have you considered building `metadataMap` only if the RW1 contains some samples? If it's just metadata (as in a real RW1 request) then we would have no performance penalty by building this map. On the other side, always building the map would always exercise this code, so I think it's a risk-performance trade-off.

pracucci (approved):

pkg/mimirpb/prealloc_rw2.go
alexweav: Fair enough, anything that guards when the map doesn't exist. Checking the map itself is better if we change the conditions for initializing it

pkg/mimirpb/prealloc_rw2.go
alexweav: Done - specifically we only build it if there's Timeseries + Metadata in the same request.

PR #7986, opened 2024-04-26
ingester: ActiveSeriesRequest can return native histograms information

krajorama:
#### What this PR does

ingester: ActiveSeriesRequest can return native histograms information

The interface  `ActiveSeries(ActiveSeriesRequest)`, gets a new `Type` field that requests active series by default (0) but can request native histograms (1). In case of histograms the new  bucket_count array is filled with bucket counts in the response. Both these changes should be backwards compatible, enabling a smooth upgrade.

Internally I flipped the order of postings iterators from sharding over active series to filtering active native histogram series over sharding to be able to get the bucket counts from the postings iterator directly. Also removes the use of the intermediate postings list to be able to directly stream the results.

Two alternative were considered:
1) add a new RPC interface. Lot's of duplicate code.
2) encode bucket count in a special label. Due to tag=stringlabels this is not trivial and also would involve extra processing to put the new label into its properly sorted position in the labels. Overall sounds messy.

#### Which issue(s) this PR fixes or relates to

Related to #7981 

#### Checklist

- [X] Tests updated.
- N/A Documentation added.
- N/A deferred `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- N/A [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12357, opened 2025-08-11
ingester: Move test-only code to list_series_test.go

dimitarvdimitrov:
i noticed this while reviewing #12333 because it doesn't handle `storage.ErrNotFound` errors

this used to be used in prod ocde before #7986


flxbk (approved):
LGTM but the linter wants to have its way.

PR #12358, opened 2025-08-11
Usage tracker service (pre-kafka tenant active series limits)

colega:
#### What this PR does

This PR creates a new service called usage-tracker.

The purpose of the usage tracker is to enforce active series limits before ingesting the data into Kafka in the new ingest storage architecture.

This service is basically a distributed, coordinated and replicated in-memory hashmap storing series hashes for each one of the tenant.

This PR is huge, this has been a while of development, so I've come up with some instructions for review:

- First of all, please read the doc in `docs/internal/usage-tracker.md`, they describe how the service works, the expectations and the decisions taken.
- After that, I would recommend proceeding with review from inside out:
- `pkg/usagetracker/tenantshard` package contains the implementation of the in-memory swiss-like hashmap: this is an implementation of `map[hash]timestamp` optimized for our needs:
  - Our data is small, we store timestamps as "minutes since the last 2h multiple"
    - We don't care about values higher than 1h anyway: that's the maximum active series TTL this service allows to be set. We also don't care about higher precision: we do the cleanup once in a minute, so sub-minute precision is useless here.  
    - The code is optimized to perform the update logic in the map itself, so we don't have to retrieve, compare, and put back the value for each series hash.
      - This is almost everything this service does, so it's worth optimizing it.
    - The code can peform garbage collection by just iterating the values (not the keys), exploiting cache locality even more.
- `pkg/usagetracker/clock` just contains a type-wrapper for our minutes-timestamp.    
- `pkg/usagetracker/tracker_store.go` is the core business logic of the usage-tracker. It handles the track series requests, creating the tenantshard instances, and also processes the incoming "created series" events.
    - `pkg/usagetracker/tracker_store_snapshot.go` extends that with methods necessary to genearte & load snapshots.
    - `pkg/usagetracker/tracker_store_collector.go` implements a prometheus collector, so each tracker store can be registered in the prometheus registry directly.
- `pkg/usagetracker/partition_handler.go` is the `service.Service` handling the lifecycle of one partition in the usage-tracker service replica. It holds one tracker store, and adds layers of event processing, snapshot coordination, etc. 
- `pkg/usagetracker/tracker.go` is the main usage-tracker service file: it coordinates the creation and removal of partition handlers, as well as the general instance read-only state and shutdown process.
- Finally, `pkg/usagetracker/tracer_instance_ring.go` and `pkg/usagetracker/tracer_partition_ring.go` are the usual boilerplate for both instance and partition ring used by usage-tracker.

Finally, for the distributor/client part, check out the code in `pkg/usagetracker/usagetrackerclient` and `prePushMaxSeriesLimitMiddleware` in the distributor.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


pkg/distributor/validate.go
```diff
@@ -67,6 +67,9 @@ var (
 	// reasonTooManyHAClusters is one of the reasons for discarding samples.
 	reasonTooManyHAClusters = "too_many_ha_clusters"
 
+	// TODO in the ingester the reason is "per_user_series_limit". Here we use a different value to not conflict with ingesters when running in read-write or monolithic mode.
+	reasonPerUserSeriesLimit = "pre_ingestion_per_user_series_limit"
```
colega: Note for myself: solve this TODO.

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+The collision of series hashes are ignored as they have negligible impact on the system precision.
+
+## Distributor interaction
+
+Distributors interact with usage tracker service through the `UsageTrackerClient` in one of the middlewares of the push chain.
+
+A quick overview of this interaction from the Distributor's perspective:
+
+- Series ready to be ingested are hashed into uint64 values, these hashes are what Usage tracker service tracks.
+- The hashes are sharded across a fixed number of partitions.
```
tcard, line 43: To provide more explicit context to the rest, I'd maybe add here:

```suggestion
- The hashes are sharded across a fixed number of partitions. Each partition is owned by at least one usage tracker service instance, and a single instance doesn't own more than one partition.
```

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+- Distributor filters the `WriteRequest` to remove rejected series hashes.
+- If series are rejected, the distributor returns an error to the client with a `429 Too Many Requests` status code.
+
+The middleware is positioned last in the chain to enforce limits after all filtering, relabeling, and transformations.
+
+## Partitions and Shards
+
+### Partitions
+
+Usage tracker service uses a fixed number of **partitions** (default: 64) this number is configurable but cannot be changed when the system is running without undesirable side effects.
```
tcard:
```suggestion
Usage tracker service uses a fixed number of **partitions** (default: 64). This number is configurable but cannot be changed when the system is running without undesirable side effects.
```

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Distributor interaction
+
+Distributors interact with usage tracker service through the `UsageTrackerClient` in one of the middlewares of the push chain.
+
+A quick overview of this interaction from the Distributor's perspective:
+
+- Series ready to be ingested are hashed into uint64 values, these hashes are what Usage tracker service tracks.
+- The hashes are sharded across a fixed number of partitions.
+- The usage tracker partition ring is checked to find the partition owner for each partition.
```
tcard: As there may be no single owner for a partition:

```suggestion
- The usage tracker partition ring is checked to find a partition owner for each partition.
```

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Partitions
+
+Usage tracker service uses a fixed number of **partitions** (default: 64) this number is configurable but cannot be changed when the system is running without undesirable side effects.
+
+- Each partition is usually owned by at least one replica in each zone.
+- During scaling, partitions are reassigned to new replicas and multiple replicas can own the same partition in the same zone.
+- It's distributor's responsibility to choose the correct partition where series hashes should be sent to, usage-tracker does not verify this.
+
+### Shards
```
tcard, line 62: I think it may be confusing to have "partitions" and "shards" as separate concepts, as they're often used as synonymous, or as sharding as a specific case of partitioning.

From Wikipedia:

> A database shard, or simply a shard, is a [horizontal partition](https://en.wikipedia.org/wiki/Partition_(database)#Horizontal_partitioning) of data in a [database](https://en.wikipedia.org/wiki/DBMS) or [search engine](https://en.wikipedia.org/wiki/Search_engine).

I would maybe call this "bucket", "group", or just "subpartition".

Anyway, not a huge deal.

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+- Start a Kafka consumer to read events from the corresponding Kafka partition at the last offset produced when the snapshot was created.
+
+The partition service will subscribe to the events topic, which may receive two kinds of events:
+
+- **SeriesCreatedEvent**: contains tenant ID, series hashes, and timestamps. Used to track new series.
+  - This event is produced by the replicas handling this same partition. Own events are ignored (by checking a record header)
+  - This event is used to achieve eventual consistency across replicas.
+- **SnapshotEvent**: contains the information about a single snapshot file produced by some replica (own events are ignored).
+  - This file might contain one or multiple shards.
+  - Continuous snapshot loading is done to ensure that a replica that has not received updates for a given series hash won't garbage-collect it if a different replica updated it in the meanatime.
```
tcard:
```suggestion
  - Continuous snapshot loading is done to ensure that a replica that has not received updates for a given series hash won't garbage-collect it if a different replica updated it in the meantime.
```

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+
+- These snapshots are written into files, one shard at a time, producing a new file each time a file size limit is reached (100MB by default).
+- Once a file is produced, it's written to object storage and a `SnapshotEvent` is produced to the Kafka topic.
+- Once all shards are written, a `SnapshotRecord` record is produced to the snapshots metadata topic.
+
+The partition will also periodically run a garbage-collection process to remove the series hashes that have not received updates for a configured amount of time (default: 20 minutes).
+
+## Horizontal Scaling
+
+### Partition distribution across usage-tracker replicas.
```
tcard: For consistency:

```suggestion
### Partition distribution across usage tracker replicas
```

docs/internal/usage-tracker.md
colega: A single instance does own multiple partitions.

tcard: Ah sure, sorry, I had the ingester partitions model in mind at this point I guess.

I think maybe just moving "Distributor interaction" below "Partitions and shards" may provide the necessary context for it.

BTW I'm now noticing here we use "sharded" in a different sense than how "shard" is used below. We could use "distributed" or "spread" maybe.

pkg/usagetracker/tenantshard/map.go
```diff
@@ -0,0 +1,249 @@
...
+	return pooled
+}
+
+func (m *Map) Iterator() func(LengthCallback, IteratorCallback) {
+	keysClone := pooledClone(m.keys, keysPool)
+	dataClone := pooledClone(m.data, dataPool)
+	count := m.count()
+
+	return func(length LengthCallback, iterSeries IteratorCallback) {
+		length(count)
```
tcard: Why get a `LengthCallback` if the length is known by the time `Iterator` returns? It could just be returned as a number.

It also seems to me that getting the length is independent from iterating. Why not move it to its own method?

Also, it would be nice if this returned a standard `iter.Seq2[uint64, clock.Minutes]` instead.

CHANGELOG.md
```diff
@@ -117,6 +119,7 @@
 * [FEATURE] You can configure Mimir to export traces in OTLP exposition format through the standard `OTEL_` environment variables. #11618
 * [FEATURE] distributor: Allow configuring tenant-specific HA tracker failover timeouts. #11774
 * [FEATURE] OTLP: Add experimental support for promoting OTel scope metadata (name, version, schema URL, attributes) to metric labels, prefixed with `otel_scope_`. Enable via the `-distributor.otel-promote-scope-metadata` flag. #11795
+* [FEATURE] Usage-tracker: new component that tracks active series before their ingestion to Kafka in the ingest architecture. #11816
```
tcard: Why is this both in "unreleased" and here?

pkg/usagetracker/partition_handler.go
```diff
@@ -0,0 +1,972 @@
...
+			}
+			if i == 0 {
+				level.Info(p.logger).Log("msg", "loaded the first snapshot file", "filename", snapshot.filename, "shards", len(snapshot.data), "load_time", time.Since(t0))
+			}
+			i++
+		}
+		errs <- nil
+	}()
+
+	go func() {
```
tcard, line 321: Mostly for code simplicity, instead of launching two goroutines (consumer and producer) and have the parent goroutine waiting, we can launch a single producer goroutine and consume in the parent goroutine.

pkg/usagetracker/tracker.go
```diff
@@ -0,0 +1,929 @@
...
+	// We'll take the write mutex when we want to modify t.partitions.
+	t.partitionsMtx.RLock()
+	defer t.partitionsMtx.RUnlock()
+	// locked will run f while write-locked
+	locked := func(f func()) {
+		t.partitionsMtx.RUnlock()
+		t.partitionsMtx.Lock()
+		f()
+		t.partitionsMtx.Unlock()
+		t.partitionsMtx.RLock()
```
tcard, line 377: Is it OK if another goroutine read-locks in between line 376 and 377? If yes, then why take the read-lock at all? As the comment says, no other goroutine is going to write to `t.partitions`, so we can safely read from it in this function/goroutine.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
 * [CHANGE] Query-frontend: Use the Mimir Query Engine (MQE) by default. #12361
 * [FEATURE] Distributor, ruler: Add experimental `-validation.name-validation-scheme` flag to specify the validation scheme for metric and label names. #12215
 * [FEATURE] Distributor: Add experimental `-distributor.otel-translation-strategy` flag to support configuring the metric and label name translation strategy in the OTLP endpoint. #12284 #12306 #12369
+* [FEATURE] Distributor: Add experimental `-distributor.otel-translation-strategy` flag to support configuring the metric and label name translation strategy in the OTLP endpoint. #12284 #12306
+* [FEATURE] Usage-tracker: Introduced a new experimental service to enforce active series limits before Kafka ingestion. #12358
```
```suggestion
* [FEATURE] Usage-tracker: Introduce a new experimental service to enforce active series limits before Kafka ingestion. #12358
```

cmd/mimir/help.txt.tmpl
```diff
@@ -891,6 +895,380 @@ Usage of ./cmd/mimir/mimir:
...
+  -usage-tracker.created-series-events-batch-ttl duration
+    	Time after which a batch of created series events is published even if it's not full. (default 250ms)
+  -usage-tracker.created-series-events-max-batch-size-bytes int
+    	Maximum size of a batch of created series events to be published. (default 1048576)
+  -usage-tracker.created-series-events-max-pending int
+    	Maximum number of pending created series events waiting to be published. (default 10000)
+  -usage-tracker.created-series-events-publish-concurrency int
+    	Number of concurrent workers publishing created series events. (default 10)
+  -usage-tracker.do-not-apply-series-limits
+    	If true, the usage-tracker service will track all series and will not apply series limits.
```
tacole02:
```suggestion
    	If true, the usage-tracker service tracks all series and doesn't apply series limits.
```

cmd/mimir/help.txt.tmpl
```diff
@@ -891,6 +895,380 @@ Usage of ./cmd/mimir/mimir:
...
+  -usage-tracker.snapshots-storage.swift.user-domain-name string
+    	OpenStack Swift user's domain name.
+  -usage-tracker.snapshots-storage.swift.user-id string
+    	OpenStack Swift user ID.
+  -usage-tracker.snapshots-storage.swift.username string
+    	OpenStack Swift username.
+  -usage-tracker.target-snapshot-file-size-bytes int
+    	Target size of a snapshot file in bytes. This is used to determine when to create a new snapshot file. It should be greater than 0. (default 104857600)
+  -usage-tracker.use-global-series-limits
+    	If true, the usage-tracker service will use global in-memory series limits instead of the active series limits. This is useful for testing purposes only.
```
tacole02:
```suggestion
    	If true, the usage-tracker service uses global in-memory series limits instead of the active series limits. This is useful for testing purposes only.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2409,6 +2409,24 @@ How to **fix** it:
...
 - Consider scaling out the ingesters.
 
+### err-mimir-max-active-series
+
+This error occurs when the number of active series for a given tenant exceeds the configured limit.
+
+This limit is applied before the series are ingested into Kafka.
+
+The limit is used to protect ingesters from overloading in case a tenant writes a high number of active series, as well as to protect the whole system’s stability from potential abuse or mistakes.
+To configure the limit on a per-tenant basis, use the `-distributor.max-active-series-per-user` option (or `max_active_series_per_user` in the runtime configuration).
```
tacole02:
```suggestion
To configure the limit on a per-tenant basis, use the `-distributor.max-active-series-per-user` option or `max_active_series_per_user` in the runtime configuration.
```

docs/sources/mimir/manage/mimir-runbooks/_index.md
```diff
@@ -2409,6 +2409,24 @@ How to **fix** it:
...
+
+How to **fix** it:
+
+- Ensure the actual number of active series written by the affected tenant is legit.
+- Consider increasing the per-tenant limit by using the `-distributor.max-active-series-per-user` option (or `max_active_series_per_user` in the runtime configuration).
```
tacole02:
```suggestion
- Consider increasing the per-tenant limit by using the `-distributor.max-active-series-per-user` option or `max_active_series_per_user` in the runtime configuration.
```

tacole02 (approved):
Docs look great! Thanks, @colega

tacole02:
Does this service only apply to ingest storage architecture? If so, we'll want to explicitly note that in the docs for 3.0.

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
+# Usage tracker service
+
+## Mission, design & decisions
+
+### Mission
+
+In Mimir's ingest architecture, the max series limit enforcement creates three critical challenges when performed by the ingesters (like in classic architecture):
+
+1. **Ingestion cost** - Series over-the-limit are ingested into Kafka/Warpstream before being rejected, generating cost for data that cannot be used.
+2. **Block-builder inconsistency** - Block-builder must independently enforce limits, potentially making different decisions than ingesters.
```
seizethedave: making different decisions -> them each storing an inconsistent view of the data.

docs/internal/usage-tracker.md
```diff
@@ -0,0 +1,232 @@
...
+Usage tracker is designed to be a highly available, horizontally scalable service that tracks series hashes and enforces series limits across multiple availability zones.
+
+The synchronization between multiple availability-zones is eventually consistent, and it leverages a Kafka-compatible storage to both synchronize replicas in runtime and to persist the state of the service upon restarts.
+
+### Consistency over precision
+
+This service is not designed to be meant as a cost-control mechanism but a system protection.
+
+The consistency of decisions is considered more important than the precision of the limiting algorithm,
+so in some scenarios the system will accept more series than the configured limit, but it will reject series that were already accepted and are still active.
```
seizethedave: This last bit doesn't make sense to me. It may accept more than the limit, but it will reject series that were already accepted?

seizethedave (commented):
I'll find time to read more.

docs/internal/usage-tracker.md
colega: As you say, they're both same concept but applied in different places, I tried to use different words to make that separation clearer.

pkg/usagetracker/tenantshard/map.go
colega: This iterator is only used by the snapshot writer, so it's tailored to its needs.
- We need the length beforehand because that's how encoding works (to allow slice pre-allocation)
- This iterator ensures that the length provided by the callback is consistent with the elments provided by it, if it's a separate method we can't ensure that without exposing the mutex details to the caller, which I tried to avoid.

CHANGELOG.md
colega: This was a bad merge, but for some reason Github doesn't show it in the diff. 
Fixed in https://github.com/grafana/mimir/pull/12358/commits/5a7f646e02a6a897cff5ad1ad66a82b029793a28

pkg/usagetracker/tenantshard/map.go
tcard: It still seems to me that `LengthCallback` is unnecessarily convoluted, as it could be replaced by just returning the length:

```diff
m.Lock()
-iter := m.Iterator()
+length, iter := m.Iterator()
m.Unlock()
// Once we have the iter we don't need to hold the mutex anymore as it works on a iter.
tenant.RUnlock()

+snapshot.PutUvarint64(uint64(length))

iter(
-	func(length int) {
-		snapshot.PutUvarint64(uint64(length))
-	},
	func(s uint64, ts clock.Minutes) {
		snapshot.PutBE64(s)
		snapshot.PutByte(byte(ts))
	},
)
```

Just a nitpick though.

colega:
> Does this service only apply to ingest storage architecture? If so, we'll want to explicitly note that in the docs for 3.0.

Yes, but it's experimental so far, I don't think we'll mark it as stable for 3.0.

docs/internal/usage-tracker.md
colega: It makes more sense with a `never` :D 
```suggestion
so in some scenarios the system will accept more series than the configured limit, but it will never reject series that were already accepted and are still active.
```

docs/internal/usage-tracker.md
colega: Sounds good to me:
```suggestion
2. **Block-builder inconsistency** - Block-builder must independently enforce limits, potentially making them each storing an inconsistent view of the data.
```

cmd/mimir/help.txt.tmpl
colega: Thanks, done in https://github.com/grafana/mimir/pull/12358/commits/115ac8f23d0b2e6179fe5867053d17c70f61fd63

cmd/mimir/help.txt.tmpl
colega: Thanks, done in https://github.com/grafana/mimir/pull/12358/commits/115ac8f23d0b2e6179fe5867053d17c70f61fd63

CHANGELOG.md
colega: Thanks, done in https://github.com/grafana/mimir/pull/12358/commits/115ac8f23d0b2e6179fe5867053d17c70f61fd63

pkg/usagetracker/tracker.go
colega: I guess it would be technically correct but also unexpected for the reader. I'd rather just keep the usual "I need to read so I'll take the read mutex" pattern.

pkg/usagetracker/tenantshard/map.go
colega: I see, I like it, implemented it in: https://github.com/grafana/mimir/pull/12358/commits/cbf0de56b641385c2b7ef2f507c0c9fc194f4e2d

pkg/usagetracker/tracker.go
tcard: I'm not so sure this is more unexpected for the reader than the alternative. To me, this code pattern (read-unlock-write-lock) smells of a typical check-then-act race condition.

Now, as the (code) comment above notes, in this case we can't actually have a check-then-act race condition, because there is no other thread that can write-lock. But the fact that this thread takes the read-lock suggests to me (as a reader) that it expects such possibility to exist (why would it read-lock otherwise?), and thus this read-unlock-write-lock may be a bug.

Eye of the beholder I guess.

PR #12361, opened 2025-08-12
Make MQE the default query engine in query-frontends

charleskorn:
#### What this PR does

We've been running query-frontends with MQE enabled in production at Grafana Labs for quite a while now, and have had no issues. 

So in this PR I'm making MQE the default. It's still possible to switch back to Prometheus' engine by setting `-query-frontend.query-engine=prometheus`.

Resolves https://github.com/grafana/mimir/issues/11882

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


charleskorn:
> But when we should bring MQE from experimental? I believe should be sometime before Mimir 3.0. Also what are "checklist" that we have to pass for MQE stable?

Agreed. Given MQE is now the default for both queriers and query-frontends, I would consider MQE no longer experimental. We'll likely keep all of the related CLI flags marked as experimental so that we can remove them later (and therefore make MQE the only option).

PR #12369, opened 2025-08-12
Distributor: Auto-deduce OTel translation strategy in OTLP endpoint if unset

aknuds1:
#### What this PR does

Modify validation code to not configure default OTel metric and label name translation strategy, but instead leave that to the distributor's OTLP endpoint logic. This is because we found a bug connected to automatically deducing the translation strategy within `Limits.Validate`, and it might as well be done at the call site (in the OTLP endpoint).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -292,6 +292,24 @@ func newOTLPParser(
 		promoteScopeMetadata := limits.OTelPromoteScopeMetadata(tenantID)
 		allowDeltaTemporality := limits.OTelNativeDeltaIngestion(tenantID)
 		translationStrategy := limits.OTelTranslationStrategy(tenantID)
+		if translationStrategy == "" {
+			// Generate translation strategy based on other settings.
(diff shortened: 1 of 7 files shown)


pkg/distributor/otel.go
```diff
@@ -292,6 +292,24 @@ func newOTLPParser(
...
+					translationStrategy = otlptranslator.UnderscoreEscapingWithoutSuffixes
+				}
+			case model.UTF8Validation:
+				if suffixesEnabled {
+					translationStrategy = otlptranslator.NoUTF8EscapingWithSuffixes
+				} else {
+					translationStrategy = otlptranslator.NoTranslation
+				}
+			}
+		}
```
colega: What if all this logic was part of `limits.OTelTranslationStrategy()`?

aknuds1: Sounds like it might work, let me try.

aknuds1: PTAL @colega.

PR #12375, opened 2025-08-12
Add metrics for prune toggles optimization pass

zenador:
#### What this PR does

Add metrics for prune toggles optimization pass

Follow up to https://github.com/grafana/mimir/pull/12303

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/optimize/ast/prune_toggles.go
+++ b/pkg/streamingpromql/optimize/ast/prune_toggles.go
@@ -16,13 +16,12 @@ import (
 // e.g. `(avg(rate(foo[1m]))) and on() (vector(1) == 1)` -> `(avg(rate(foo[1m])))`
 // e.g. `(avg(rate(foo[1m]))) and on() (vector(1) == -1)` -> `(vector(1) == -1)`
 type PruneToggles struct {
-	mapper astmapper.ASTMapper
+	mapper *astmapper.ASTExprMapperWithState
 }
 
 func NewPruneToggles() *PruneToggles {
-	mapper := astmapper.NewASTExprMapper(&pruneToggles{})
 	return &PruneToggles{
-		mapper: mapper,
+		mapper: NewPruneTogglesMapper(),
 	}
 }
 
@@ -34,7 +33,22 @@ func (p *PruneToggles) Apply(ctx context.Context, expr parser.Expr) (parser.Expr
 	return p.mapper.Map(ctx, expr)
 }
 
-type pruneToggles struct{}
+func (p *PruneToggles) HasChanged() bool {
+	return p.mapper.HasChanged()
+}
+
+func NewPruneTogglesMapper() *astmapper.ASTExprMapperWithState {
+	mapper := &pruneToggles{}
+	return astmapper.NewASTExprMapperWithState(mapper)
+}
+
+type pruneToggles struct {
+	changed bool
+}
+
+func (mapper *pruneToggles) HasChanged() bool {
+	return mapper.changed
+}
 
 func (mapper *pruneToggles) MapExpr(ctx context.Context, expr parser.Expr) (mapped parser.Expr, finished bool, err error) {
 	e, ok := expr.(*parser.BinaryExpr)
@@ -55,10 +69,12 @@ func (mapper *pruneToggles) MapExpr(ctx context.Context, expr parser.Expr) (mapp
 	if isEmpty {
 		// The right hand side is empty, so the whole expression is empty due to
 		// "and on()", return the right hand side.
+		mapper.changed = true
 		return e.RHS, false, nil
 	}
 	// The right hand side is const and not empty, so the whole expression is
 	// just the left side.
+	mapper.changed = true
 	return e.LHS, false, nil
 }
 
--- a/pkg/streamingpromql/optimize/ast/prune_toggles.go
+++ b/pkg/streamingpromql/optimize/ast/prune_toggles.go
@@ -20,9 +20,7 @@ type PruneToggles struct {
 }
 
 func NewPruneToggles() *PruneToggles {
-	return &PruneToggles{
-		mapper: NewPruneTogglesMapper(),
-	}
+	return &PruneToggles{}
 }
 
 func (p *PruneToggles) Name() string {
@@ -30,10 +28,14 @@ func (p *PruneToggles) Name() string {
 }
 
 func (p *PruneToggles) Apply(ctx context.Context, expr parser.Expr) (parser.Expr, error) {
+	p.mapper = NewPruneTogglesMapper()
 	return p.mapper.Map(ctx, expr)
 }
 
 func (p *PruneToggles) HasChanged() bool {
+	if p.mapper == nil {
+		return false
+	}
 	return p.mapper.HasChanged()
 }
 
--- a/pkg/streamingpromql/optimize/ast/prune_toggles.go
+++ b/pkg/streamingpromql/optimize/ast/prune_toggles.go
@@ -5,6 +5,8 @@ package ast
 import (
 	"context"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/promql/parser"
 
 	"github.com/grafana/mimir/pkg/frontend/querymiddleware/astmapper"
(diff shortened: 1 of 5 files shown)


pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -16,13 +16,29 @@ import (
 // e.g. `(avg(rate(foo[1m]))) and on() (vector(1) == 1)` -> `(avg(rate(foo[1m])))`
 // e.g. `(avg(rate(foo[1m]))) and on() (vector(1) == -1)` -> `(vector(1) == -1)`
 type PruneToggles struct {
-	mapper astmapper.ASTMapper
+	mapper *ASTExprMapperWithState
+
+	pruneTogglesMetrics
```
charleskorn: Is there a reason why this is embedded rather than an ordinary field?

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -16,13 +16,29 @@ import (
...
+}
+
+type pruneTogglesMetrics struct {
+	pruneTogglesAttempts  prometheus.Counter
+	pruneTogglesSuccesses prometheus.Counter
```
charleskorn: Given these fields are in a struct called `pruneTogglesMetrics`, the `pruneToggles` prefixes on both seem unnecessary to me.

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -16,13 +16,29 @@ import (
...
+func NewPruneToggles(reg prometheus.Registerer) *PruneToggles {
+	metrics := pruneTogglesMetrics{
+		pruneTogglesAttempts: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "cortex_mimir_query_engine_prune_toggles_attempted_total",
+			Help: "Total number of queries that the optimization pass has attempted to rewrite by pruning toggles.",
+		}),
+		pruneTogglesSuccesses: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "cortex_mimir_query_engine_prune_toggles_succeeded_total",
+			Help: "Total number of queries where the optimization pass has successfully rewritten by pruning toggles.",
+		}),
```
charleskorn: The term "success" here might be a bit misleading - doesn't this count the number of times the expression was changed? ("success" to me implies it didn't fail with an error, but doesn't mean that it changed the expression)

Perhaps "rewritten expressions" or "applied expressions" would be less ambiguous?

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -31,10 +47,34 @@ func (p *PruneToggles) Name() string {
...
+	if p.mapper.HasChanged() {
+		p.pruneTogglesSuccesses.Inc()
+	}
+	return newExpr, err
+}
+
+func (p *PruneToggles) HasChanged() bool {
+	if p.mapper == nil {
+		return false
+	}
```
charleskorn: When would `mapper` be `nil`?

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -66,6 +68,7 @@ func TestPruneToggles(t *testing.T) {
 			require.NoError(t, err)
 
 			require.Equal(t, expectedExpr.String(), outputExpr.String())
+			require.Equal(t, input != expected, optimizer.HasChanged())
```
charleskorn: [nit] Might be nice to check that the metric values are as expected as well.

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -31,10 +47,34 @@ func (p *PruneToggles) Name() string {
...
+	return p.mapper.HasChanged()
+}
```
charleskorn: Is this method needed?

pkg/streamingpromql/optimize/ast/prune_toggles.go
```diff
@@ -31,10 +47,34 @@ func (p *PruneToggles) Name() string {
 }
 
 func (p *PruneToggles) Apply(ctx context.Context, expr parser.Expr) (parser.Expr, error) {
-	return p.mapper.Map(ctx, expr)
+	p.mapper = NewPruneTogglesMapper()
```
charleskorn: `Apply` might be called by multiple queries simultaneously, so setting `p.mapper` like this isn't safe. 

Do we need to store the mapper on `PruneToggles`? It looks like it could be a local variable.

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: No particular reason, but I saw other files with this pattern and liked the way it was organised (metrics separate from the rest), but am open to changing it.

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: We can access them directly like `p.pruneTogglesAttempts.Inc()` without the struct `pruneTogglesMetrics`. If you think it's better should I change it to remove `pruneTogglesMetrics`?

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: Using "rewritten" instead.

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: If HasChanged() is ever somehow called before Apply() (since we make the mapper there and not on initialisation of the struct), not really likely to happen but just a safeguard. Since we're not storing mapper in the struct anymore or keeping this method this doesn't matter now.

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: It was only used in tests, but now we replace it by checking metrics.

pkg/streamingpromql/optimize/ast/prune_toggles.go
zenador: It was only used to check HasChanged() in unit tests, but now we solve that in other ways.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -54,25 +57,45 @@ var testCasesPruneToggles = map[string]string{
...
+%v %v
+`, metricNameTotal, metricNameTotal, metricNameTotal, expectedTotal)
+	require.NoError(t, testutil.GatherAndCompare(g, strings.NewReader(expectedMetricsTotal), metricNameTotal))
+
+	const metricNameChanged = "cortex_mimir_query_engine_prune_toggles_rewritten_total"
+	expectedMetricsChanged := fmt.Sprintf(`# HELP %v Total number of queries where the optimization pass has rewritten the query by pruning toggles.
+# TYPE %v counter
+%v %v
+`, metricNameChanged, metricNameChanged, metricNameChanged, expectedChanged)
+	require.NoError(t, testutil.GatherAndCompare(g, strings.NewReader(expectedMetricsChanged), metricNameChanged))
```
charleskorn: [nit] This could be combined into one `GatherAndCompare` call.

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -54,25 +57,45 @@ var testCasesPruneToggles = map[string]string{
...
 		})
 	}
 }
 
+func checkPruneTogglesMetrics(t *testing.T, g prometheus.Gatherer, expectedTotal, expectedChanged int) {
+	const metricNameTotal = "cortex_mimir_query_engine_prune_toggles_attempted_total"
+	expectedMetricsTotal := fmt.Sprintf(`# HELP %v Total number of queries that the optimization pass has attempted to rewrite by pruning toggles.
+# TYPE %v counter
+%v %v
+`, metricNameTotal, metricNameTotal, metricNameTotal, expectedTotal)
```
charleskorn: This could be simplified slightly with [explicit argument indexes](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes):

```suggestion
	expectedMetricsTotal := fmt.Sprintf(`# HELP %[1]v Total number of queries that the optimization pass has attempted to rewrite by pruning toggles.
# TYPE %[1]v counter
%[1]v %[2]v
`, metricNameTotal, expectedTotal)
```

pkg/streamingpromql/optimize/ast/prune_toggles.go
charleskorn: I'd err on the side of making it an ordinary field unless there's a compelling reason to make it embedded.

charleskorn (approved):
LGTM modulo suggestions below:

PR #12377, opened 2025-08-13
MQE: shift responsibilities of `Materializer`

charleskorn:
#### What this PR does

This PR changes the responsibilities of MQE's `Materializer` type.

Previously, `Materializer.ConvertNodeToOperator` would materialize all of the given node's children and pass it to `Node.OperatorFactory`, and `OperatorFactory` would have no access to the `Materializer`. With the existing set of query plan nodes, this was not an issue, but it is problematic for some scenarios, for example:

* for remote execution, there's no point materializing operators for the parts of the query plan that will be executed remotely
* for caching and splitting, there's also no point materializing operators for cached results, but we'll then need to materialize the child operator if there is a cache miss, or only materialize the child operator for a subset of the time range

This PR also moves `Materializer` to the `planning` package to prevent an import cycle (and seems like a more natural fit for there anyway).

Depending on which PR merges first, either this PR or https://github.com/grafana/mimir/pull/12302 will need changes to account for the changes in the other.

This PR has no user-visible impact and so I've chosen not to add a changelog entry.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/planning/plan.go
+++ b/pkg/streamingpromql/planning/plan.go
@@ -76,7 +76,9 @@ type Node interface {
 	ChildrenTimeRange(timeRange types.QueryTimeRange) types.QueryTimeRange
 
 	// OperatorFactory returns a factory that produces operators for this node.
-	OperatorFactory(children []types.Operator, timeRange types.QueryTimeRange, params *OperatorParameters) (OperatorFactory, error)
+	//
+	// Implementations may retain the provided Materializer for later use.
+	OperatorFactory(materializer *Materializer, timeRange types.QueryTimeRange, params *OperatorParameters) (OperatorFactory, error)
 
 	// ResultType returns the kind of result this node produces.
 	//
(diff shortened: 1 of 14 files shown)


pkg/streamingpromql/planning/plan.go
```diff
@@ -76,7 +76,9 @@ type Node interface {
 	ChildrenTimeRange(timeRange types.QueryTimeRange) types.QueryTimeRange
 
 	// OperatorFactory returns a factory that produces operators for this node.
-	OperatorFactory(children []types.Operator, timeRange types.QueryTimeRange, params *OperatorParameters) (OperatorFactory, error)
+	//
+	// Implementations may retain the provided Materializer for later use.
```
56quarters, line 80: Because the `Materializer` isn't thread safe, it seems like this makes `planning.Node` also potentially not thread safe. I don't have a suggestion for this PR but since a lot of the MQE is meant to be run in a single thread, it'd be nice to enforce this at the type level somehow (like how rust has `Send`/`Sync` traits)

PR #12378, opened 2025-08-13
fix(pkg): Incorrect conversion between integer types

kreeksec:
fix the problem, we need to ensure that any value parsed from the form field and converted to `uint32` is within the valid range for `uint32` (i.e., 0 to `math.MaxUint32`). The best way to do this is to add an explicit upper bound check after parsing the value with `strconv.Atoi`. If the value exceeds `math.MaxUint32`, we should clamp it to `math.MaxUint32` (or, depending on desired semantics, set it to a default value or return an error). This change should be made in `pkg/compactor/planned_jobs_http.go` where the value is parsed and assigned to `mergeShards` and `splitGroups`. We need to import the `math` package in this file to access `math.MaxUint32`.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

#### References
Wikipedia [Integer overflow](https://en.wikipedia.org/wiki/Integer_overflow)
Go language specification [Integer overflow](https://golang.org/ref/spec#Integer_overflow)
Documentation for [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi)
Documentation for [strconv.ParseInt](https://golang.org/pkg/strconv/#ParseInt)
Documentation for [strconv.ParseUint](https://golang.org/pkg/strconv/#ParseUint)

### DIFF
--- a/pkg/compactor/planned_jobs_http.go
+++ b/pkg/compactor/planned_jobs_http.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strconv"
 	"time"
+	"math"
 
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


colega (commented):
Thank you for your contribution!

Can you add a test for this, please, to illustrate what we're fixing?

PR #12380, opened 2025-08-13
alertmanager: fix flaky state test

titolins:
## Context
- fixes https://github.com/grafana/mimir/issues/12146

### DIFF
--- a/pkg/alertmanager/alertmanager_test.go
+++ b/pkg/alertmanager/alertmanager_test.go
@@ -798,12 +798,9 @@ func TestGetFullStateHandler(t *testing.T) {
 
 		var parsedState clusterpb.FullState
 		require.NoError(t, parsedState.Unmarshal([]byte(decodedState)))
-
 		require.Len(t, parsedState.Parts, 2)
-		var (
-			nflC int
-			silC int
-		)
+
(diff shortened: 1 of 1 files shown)


pkg/alertmanager/alertmanager_test.go
```diff
@@ -778,17 +779,43 @@ func TestGetFullStateHandler(t *testing.T) {
...
-			"status": "success"
+		var (
+			parsedBody struct {
+				Status string `json:"status"`
+				Data   struct {
+					State string `json:"state"`
+				} `json:"data"`
+			}
+			parsedState clusterpb.FullState
+		)
```
santihernandezc: This is a bit hard to read.
Wdyt about declaring just `parsedBody` here?
```suggestion
		var parsedBody struct {
			Status string `json:"status"`
			Data   struct {
				State string `json:"state"`
			} `json:"data"`
		}
```

We don't need `parsedState` until line 800.

santihernandezc (approved):
LGTM!

pkg/alertmanager/alertmanager_test.go
titolins: Suure, done now 👍

PR #12384, opened 2025-08-13
Multiple Components: Use irate in CPU HPA query

dmwilson-grafana:
#### What this PR does

At Grafana Labs, we have encountered cases where rate() extrapolates a metric over missing values, underestimates CPU utilization, and triggers an unnecessary downscaling. This PR switches the default CPU scaler metrics to use `irate()` rather than `rate()` (see:`autoscaling.libsonnet`) and affects the scaling behavior of the following components: 

- alertmanager
- query frontend
- ruler
- ruler querier 
- ruler query frontend 
- distributor

This change makes the scaling metric more reliable when scrapes fail, because irate() only uses the last two data points rather than extrapolating over the full window. For example, consider the following chart where the target prometheus instance begins failing intermittently around 21:17. The `rate()` based scaling metric (in green) drops while the `irate()` based metric (yellow) remains around its baseline levels.

Screenshot 2025-08-13 at 4 02 46 PM

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 
 * [CHANGE] Distributor: Reduce calculated `GOMAXPROCS` to be closer to the requested number of CPUs. #12150
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12187
-* [CHANGE] The query-scheduler is now a required component that is always used by queriers and query-frontends. #12384
+* [CHANGE] Use `irate()` when calculating CPU scaling metric. Using `irate()` prevents underestimating CPU utilization when scraping fails. #12384
 
 ### Documentation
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 
 * [CHANGE] Distributor: Reduce calculated `GOMAXPROCS` to be closer to the requested number of CPUs. #12150
 * [CHANGE] Query-scheduler: The query-scheduler is now a required component that is always used by queriers and query-frontends. #12187
+* [CHANGE] The query-scheduler is now a required component that is always used by queriers and query-frontends. #12384
```
aldernero: I'm guessing you meant to update this?

dimitarvdimitrov (commented):
hey 👋 same as on the vector(0) PR - can you make the same change to the helm chart

PR #12390, opened 2025-08-14
MQE: separate responsibilities of `Node`

charleskorn:
#### What this PR does

This PR refactors the responsibilities of the `planning.Node` type to remove responsibility for creating `OperatorFactory` instances.

Instead, this responsibility has been moved to the newly-introduced `NodeMaterializer` interface.

This will allow us to implement new kinds of nodes that can't be materialized with only values available from `OperatorParameters`, and where it doesn't make sense to extend `OperatorParameters` to include these values. For example, a caching operator might need cache connection information, and a remote execution node might need access to a client for remote executors, but these things are specific to only these kinds of nodes, so it doesn't make sense to have the engine itself depend on these things. 

From a code organisation perspective, it also means the `streamingpromql` package won't need to depend on a bunch of other packages that may introduce circular dependencies (eg. for remote execution, we would likely need to reference query-frontend packages, but the query-frontend also needs to reference the `streamingpromql` package).

I've not added a changelog entry given this should have no user-visible impact.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/engine.go
+++ b/pkg/streamingpromql/engine.go
@@ -133,6 +133,15 @@ type Engine struct {
 	nodeMaterializers map[planning.NodeType]planning.NodeMaterializer
 }
 
+func (e *Engine) RegisterNodeMaterializer(nodeType planning.NodeType, materializer planning.NodeMaterializer) error {
+	if _, exists := e.nodeMaterializers[nodeType]; exists {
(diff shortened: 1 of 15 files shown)


fionaliao (approved):
LGTM

PR #12394, opened 2025-08-14
MQE: query-frontend refactoring to support remote execution of query plans in queriers

charleskorn:
#### What this PR does

This PR refactors parts of the query-frontend to make the introduction of remote execution of query plans in queriers easier.

I've tried to make each commit self-contained, and would recommend reviewing each separately.

This PR does not need a changelog entry as it makes no user-visible changes.

Related to #12302

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/v2/frontend.go
+++ b/pkg/frontend/v2/frontend.go
@@ -279,15 +279,14 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, httpRequest *httpgrpc.HTTP
 	case <-ctx.Done():
 		spanLogger.DebugLog("msg", "request context cancelled after enqueuing request, aborting", "err", ctx.Err())
 
-		if cancelCh != nil {
-			select {
-			case cancelCh <- freq.queryID:
-				// cancellation sent.
-			default:
-				// failed to cancel, ignore.
-				level.Warn(spanLogger).Log("msg", "failed to send cancellation request to scheduler, queue full")
-			}
+		select {
+		case cancelCh <- freq.queryID:
+			// cancellation sent.
+		default:
+			// failed to cancel, ignore.
+			level.Warn(spanLogger).Log("msg", "failed to send cancellation request to scheduler, queue full")
 		}
+
 		return nil, nil, ctx.Err()
 
 	case resp := <-freq.response:
--- a/pkg/frontend/v2/frontend.go
+++ b/pkg/frontend/v2/frontend.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/gogo/protobuf/proto"
 	"github.com/grafana/dskit/cancellation"
 	"github.com/grafana/dskit/flagext"
 	"github.com/grafana/dskit/grpcclient"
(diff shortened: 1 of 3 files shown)


pkg/frontend/v2/frontend_scheduler_adapter.go
```diff
@@ -27,29 +32,65 @@ type frontendToSchedulerAdapter struct {
...
+		}
+
+		// Propagate trace context for this query in the HTTP payload.
+		// We can't use the trace headers from the gRPC request, as it is a long-running stream from the frontend to the scheduler
+		// that handles many queries.
+		otel.GetTextMapPropagator().Inject(req.ctx, (*httpgrpcutil.HttpgrpcHeadersCarrier)(req.httpRequest))
+
+		msg.Payload = &schedulerpb.FrontendToScheduler_HttpRequest{HttpRequest: req.httpRequest}
+	} else {
+		// TODO: set additional queue dimensions based on queried time range of request (taking into account lookback etc.)
```
charleskorn, line 65: This needs https://github.com/grafana/mimir/pull/12396, so I'll tackle this in a separate PR.

pkg/frontend/v2/frontend.go
```diff
@@ -127,10 +127,11 @@ type queryResultWithBody struct {
 }
 
 type frontendRequest struct {
-	queryID      uint64
-	request      *httpgrpc.HTTPRequest
-	userID       string
-	statsEnabled bool
+	queryID         uint64
+	httpRequest     *httpgrpc.HTTPRequest
+	protobufRequest proto.Message
```
charleskorn, line 132: Note to reviewers: this isn't actually set anywhere yet, so all the existing code paths should behave as they did before, and always use a HTTP request payload.

Konstantinov-Innokentii (approved):
LGTM! 
One question though, mostly for my own understanding - why we don't need to make changes in frontend/v1?

56quarters:
> LGTM! One question though, mostly for my own understanding - why we don't need to make changes in frontend/v1?

v1 doesn't exist anymore as of yesterday. v1 was the version that used an embedded query-scheduler queue within the query-frontend. For Mimir 3.0, we now require a stand-alone query-scheduler: https://github.com/grafana/mimir/issues/11884

pkg/frontend/v2/frontend.go
```diff
@@ -270,39 +270,9 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, req *httpgrpc.HTTPRequest)
...
-			if retries > 0 {
-				spanLogger.DebugLog("msg", "enqueuing request failed, will retry")
-				goto enqueueAgain
-			}
-		}
-
-		spanLogger.DebugLog("msg", "enqueuing request failed, retries are exhausted, aborting")
-
-		return nil, nil, httpgrpc.Errorf(http.StatusInternalServerError, "failed to enqueue request")
+	cancelCh, err := f.enqueueRequestWithRetries(ctx, freq, spanLogger)
```
56quarters, line 271: Because `cancelCh` is always non-nil if error _is_ nil, we might be able to remove the nil checks below.

charleskorn: Good point, fixed in 27628b6.

PR #12395, opened 2025-08-14
Use IterWithAttributes in the stalePartialBlockLastModifiedTime function if possible

### LINKED ISSUE #12359: Idea: Minimize object storage calls in blocks cleaner

The compactor's blocks cleaner uses object attributes calls to search for the last modified time of stale partial blocks ([link](https://github.com/grafana/mimir/blob/250fcf45a29638582e57c69214b9c703507f257f/pkg/compactor/blocks_cleaner.go#L694)). The current approach performs one attributes call per object. 

A more performant approach is to use IterWithAttributes which will fetch attributes as part of the iteration, saving on object storage calls. This approach does require that the object storage backend supports this operation, so implementing code will need to ensure that `slices.Contains(bkt.SupportedIterOptions(), objstore.UpdatedAt)` is true for the user bucket, else use the existing approach for fetching attributes.

_No response_

_No response_

Not sure

_No response_

### Proposer?

_No response_

olegy2008:
#### What this PR does
This PR updates stalePartialBlockLastModifiedTime function to use IterWithAttributes instead of Iter to reduce the amount of object attributes calls if bucket supports objstore.UpdatedAt IterOptionType.

Fixes #12359

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/compactor/blocks_cleaner.go
+++ b/pkg/compactor/blocks_cleaner.go
@@ -67,6 +67,9 @@ type BlocksCleaner struct {
 	// Keep track of the last owned users.
 	lastOwnedUsers []string
 
+	// Whether the bucket supports UpdatedAt iteration option.
+	supportsUpdatedAtIter bool
+
 	// Metrics.
 	runsStarted                         prometheus.Counter
 	runsCompleted                       prometheus.Counter
@@ -86,12 +89,13 @@ type BlocksCleaner struct {
 
 func NewBlocksCleaner(cfg BlocksCleanerConfig, bucketClient objstore.Bucket, ownUser func(userID string) (bool, error), cfgProvider ConfigProvider, logger log.Logger, reg prometheus.Registerer) *BlocksCleaner {
 	c := &BlocksCleaner{
-		cfg:          cfg,
-		bucketClient: bucketClient,
-		usersScanner: mimir_tsdb.NewUsersScanner(bucketClient, ownUser, logger),
-		cfgProvider:  cfgProvider,
-		singleFlight: concurrency.NewLimitedConcurrencySingleFlight(cfg.CleanupConcurrency),
-		logger:       log.With(logger, "component", "cleaner"),
+		cfg:                   cfg,
+		bucketClient:          bucketClient,
+		usersScanner:          mimir_tsdb.NewUsersScanner(bucketClient, ownUser, logger),
+		cfgProvider:           cfgProvider,
+		singleFlight:          concurrency.NewLimitedConcurrencySingleFlight(cfg.CleanupConcurrency),
+		logger:                log.With(logger, "component", "cleaner"),
+		supportsUpdatedAtIter: slices.Contains(bucketClient.SupportedIterOptions(), objstore.UpdatedAt),
 		runsStarted: promauto.With(reg).NewCounter(prometheus.CounterOpts{
 			Name: "cortex_compactor_block_cleanup_started_total",
 			Help: "Total number of blocks cleanup runs started.",
@@ -623,7 +627,7 @@ func (c *BlocksCleaner) cleanUserPartialBlocks(ctx context.Context, partials map
 	// Check if partial blocks are older than delay period, and mark for deletion
 	if !partialDeletionCutoffTime.IsZero() {
 		for _, blockID := range partialBlocksWithoutDeletionMarker {
-			lastModified, err := stalePartialBlockLastModifiedTime(ctx, blockID, userBucket, partialDeletionCutoffTime)
+			lastModified, err := c.stalePartialBlockLastModifiedTime(ctx, blockID, userBucket, partialDeletionCutoffTime)
 			if err != nil {
 				level.Warn(userLogger).Log("msg", "failed while determining if partial block should be marked for deletion", "block", blockID, "err", err)
 				continue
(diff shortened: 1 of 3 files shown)


olegy2008:
I didn't update any tests, since existing test TestStalePartialBlockLastModifiedTime uses filesystem bucket which supports UpdatedAt option, so the existing test covers the new code.

Please tell if I should add test or update existing one to cover the case when UpdatedAt option is not supported.

pkg/compactor/blocks_cleaner.go
```diff
@@ -685,24 +686,53 @@ var errStopIter = errors.New("stop iteration")
...
+			}
+			if attrib.LastModified.After(partialDeletionCutoffTime) {
+				return errStopIter
+			}
+			if attrib.LastModified.After(lastModified) {
+				lastModified = attrib.LastModified
+			}
+			return nil
+		}, objstore.WithRecursiveIter())
+	}
```
dimitarvdimitrov: can we move this to a function of its own? it looks like we just duplciated it witht he toher if branch

pkg/compactor/blocks_cleaner.go
```diff
@@ -685,24 +686,53 @@ var errStopIter = errors.New("stop iteration")
...
-		if attrib.LastModified.After(partialDeletionCutoffTime) {
-			return errStopIter
-		}
-		if attrib.LastModified.After(lastModified) {
-			lastModified = attrib.LastModified
-		}
-		return nil
-	}, objstore.WithRecursiveIter())
+		}, objstore.WithRecursiveIter(), objstore.WithUpdatedAt())
+	} else {
```
dimitarvdimitrov, line 719: it's a bit hard to navigate this with so many callbacks and wide if-else's. Can we have perhaps two smaller functions which are now the two branches of the `if`?

pkg/compactor/blocks_cleaner.go
```diff
@@ -685,24 +686,53 @@ var errStopIter = errors.New("stop iteration")
...
+		}).IterWithAttributes(ctx, blockID.String(), func(attrs objstore.IterObjectAttributes) error {
+			if strings.HasSuffix(attrs.Name, objstore.DirDelim) {
+				return nil
+			}
+
+			modified, ok := attrs.LastModified()
+			if !ok {
+				// If for some reason we don't get the last modified time, we can't check it, so we just skip it.
+				return nil
+			}
```
dimitarvdimitrov: what are the possible reasons? how often would this happen?

dimitarvdimitrov (commented):
nice, generally looks ok

> Please tell if I should add test or update existing one to cover the case when UpdatedAt option is not supported.

is there a way to wrap the buckt implmentation with our own which returns empty `SupportedIterOptions`?

pkg/compactor/blocks_cleaner.go
```diff
@@ -685,24 +686,53 @@ var errStopIter = errors.New("stop iteration")
...
-	err := userBucket.WithExpectedErrs(func(err error) bool {
-		return errors.Is(err, errStopIter) // sentinel error
-	}).Iter(ctx, blockID.String(), func(name string) error {
-		if strings.HasSuffix(name, objstore.DirDelim) {
+
+	var err error
+
+	// If bucket supports UpdatedAt IterOptionType, use IterWithAttributes
+	// to reduce the amount of object attributes calls.
+	if slices.Contains(userBucket.SupportedIterOptions(), objstore.UpdatedAt) {
```
andyasp: This can be evaluated once upon construction of the blocks cleaner instead of every call. That would also provide a way to unit test both paths even if the test bucket supports this type of listing.

pkg/compactor/blocks_cleaner.go
andyasp: The answer is common prefixes, which would be captured by the `strings.HasSuffix(attrs.Name, objstore.DirDelim)` check above. This is a false case. Even without that it wouldn't be necessary because the `modified.After` checks would effectively skip it anyway since it would be a zero value of `time.Time`.

pkg/compactor/blocks_cleaner.go
andyasp: There could be an inner function that accepts a last modified time that is called by both the `Iter` and `IterWithAttributes`. The `Iter` would have to supply the last modified time from the result of an `Attributes` call while the other is supplied by the listing.

andyasp (commented):
Thanks for the PR. Both branches should be unit tested.

pkg/compactor/blocks_cleaner.go
olegy2008: Removed this unnecessary check

pkg/compactor/blocks_cleaner.go
olegy2008: Added inner function checkModifiedTime which is called in both places. Please tell if you think that should be a separate function.

pkg/compactor/blocks_cleaner.go
olegy2008: Used inner function as suggested in another comment, but i can make it a separate function if you think that would be better.

pkg/compactor/blocks_cleaner.go
olegy2008: Could you elaborate?
Do you want to add a new bool field to a BlocksCleaner struct, something like `supportsUpdatedAt`, set it in the `NewBlocksCleaner` function, and then use in the `stalePartialBlockLastModifiedTime` function?

olegy2008:
Applied suggestions from the review, updated test to also use a filesystem bucket wrapper which doesn't support UpdatedAt option.

pkg/compactor/blocks_cleaner.go
andyasp: Yes, I meant setting something in the `BlocksCleaner` struct in `NewBlocksCleaner`. It could be a bool or you could decide between two functions to eliminate the if as well.  The idea is that the bucket won't change whether it supports the listing or not midway. Currently this is rechecking that unchanging property per block.

olegy2008: I've added a `supportsUpdatedAtIter` bool to the `BlocksCleaner` struct. I tried to use functions instead (created 2 separate functions, one with Iter, second with IterWithAttributes, and selected one of them in the `BlocksCleaner` struct, but it seemed too complicated, so I thought that having if statement with bool variable would be better and cleaner.

olegy2008: @andyasp, I would appreciate your thoughts on this. If the current implementation is ok, then I will update the description and finalize the PR.

olegy2008:
Hey @andyasp, @dimitarvdimitrov, @jhalterman could you review this PR once more? Also, I don't have permissions to add the `changelog-not-needed` label.

PR #6772, opened 2023-11-28
apply multidimensional query request queuing: supply queue dimensions from frontend & utilize in scheduler

francoposa:
Changelog to be done once we solidify the shape of this.

This is the second part of our multi-dimensional queuing project to mitigate the issue where a single slow query component (like store-gateway) can also slow down a tenant's requests that don't hit that component (ingesters-only queries), or vice versa.

The first pull request https://github.com/grafana/mimir/pull/6533 introduced the TreeQueue datastructures which can handle an arbirary number of queue dimensions, but kept it to the single queue dimension of `tenantID`.

This pull request adds code to query-frontend to assign the additional queue dimension query component for a query (ingesters, store-gateway, or both), and adds the handling of that into the scheduler as well.

Various design discussions went into this, but the most critical were the decision that the calculation of additional queue dimensions should take place in the query-frontend.
The idea is that the query-frontend is already Prometheus-aware, that is it understands the shape of Prometheus queries and can make decisions based on Prometheus query attributes. The scheduler needs to utilize some arbitrary information to create additional queue dimensions, but it should not need to know anything about the shape of the items in its queue, Prometheus queries or otherwise.

As noted in the comments below, as of PR submission, this is only applied to the v2 frontend - there were some conflicting ideas on whether the functionality should be added to the v1 frontend.

The multi-dimension queue implementation is robust to enqueuing and fairly dequeuing a tenant's request even if the requests are mixed between having an additional queue dimension or not.
This means the scheduler can tolerate failures to calculate additional queue dimensions as well as receiving requests from a frontend which has not implemented the additional queue dimension calculations at the same time as receiving requests from a frontend which has.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [ ] [`about-versioning.md`](/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12396, opened 2025-08-14
MQE: add the ability to compute the queried time range for a query plan node

charleskorn:
#### What this PR does

This PR extends the `planning.Node` interface to add a `QueriedTimeRange` method.

This is needed to allow computing the queried components (ingesters or store-gateways or both) when remotely evaluating a query plan, so that query-schedulers can prioritise requests appropriately when one is under load. See https://github.com/grafana/mimir/pull/6772 for more details of this.

This has no user-visible impact, so I have not added a changelog entry.

https://github.com/grafana/mimir/pull/6772

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -137,6 +121,28 @@ func (s *Selector) loadSeriesSet(ctx context.Context) error {
 	return nil
 }
 
+func ComputeQueriedTimeRange(timeRange types.QueryTimeRange, timestamp *int64, rng time.Duration, offset int64, lookbackDelta time.Duration) (int64, int64) {
```
56quarters: nit: `rng` makes me think of `Random Number Generator`

```suggestion
func ComputeQueriedTimeRange(timeRange types.QueryTimeRange, timestamp *int64, range time.Duration, offset int64, lookbackDelta time.Duration) (int64, int64) {
```

pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -137,6 +121,28 @@ func (s *Selector) loadSeriesSet(ctx context.Context) error {
...
+	if lookbackDelta != 0 && rng != 0 {
+		panic("both lookback delta and range are non-zero")
```
56quarters: Can we include the range and lookback delta in this message since this indicates a bug?

pkg/streamingpromql/planning/plan.go
```diff
@@ -85,9 +85,48 @@ type Node interface {
 	// May return an error if the kind of result cannot be determined (eg. because the node references an unknown function).
 	ResultType() (parser.ValueType, error)
 
+	// QueriedTimeRange returns the range of data queried from ingesters and store-gateways by this node
+	// and its children.
+	//
+	// If no data is queried by this node and its children, QueriedTimeRange returns false.
```
56quarters: > If no data is queried by this node and its children, QueriedTimeRange returns false

Is this meant to say "If no data is queried by this node and its children, QueriedTimeRange.AnyDataQueried will be false"?

pkg/streamingpromql/planning/plan_test.go
```diff
@@ -118,3 +120,62 @@ func (t *testNode) OperatorFactory(_ *Materializer, _ types.QueryTimeRange, _ *O
...
+}
+
+func TestQueriedTimeRange_Union(t *testing.T) {
+	testCases := map[string]struct {
+		first    QueriedTimeRange
+		second   QueriedTimeRange
+		expected QueriedTimeRange
+	}{
+		"neither queries any data": {
+			first:    QueriedTimeRange{AnyDataQueried: false},
```
56quarters: Is it valid for `QueriedTimeRange` to set `MinT` and `MaxT` at the same time as `AnyDataQueried = false`? If not, can we document that that's the case for the type and possibly enforce it by using specific constructors for the type.

pkg/streamingpromql/planning/plan_test.go
```diff
@@ -118,3 +120,62 @@ func (t *testNode) OperatorFactory(_ *Materializer, _ types.QueryTimeRange, _ *O
...
+			expected: QueriedTimeRange{MinT: timestamp.Time(1000), MaxT: timestamp.Time(5000), AnyDataQueried: true},
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			result := testCase.first.Union(testCase.second)
+			require.Equal(t, testCase.expected, result)
+
+			// Swapping the order should produce the same result.
```
56quarters, line 172: This might not be the case if if `MinT` and `MaxT` are set at the same time as `AnyDataQueried = false`

pkg/streamingpromql/operators/selectors/selector.go
charleskorn: Unfortunately `range` is a keyword so we can't use that as a variable name - maybe `selectorRange` would be clearer? I've changed it in f097961.

pkg/streamingpromql/operators/selectors/selector.go
charleskorn: Yep, good idea, added in f097961.

pkg/streamingpromql/planning/plan.go
charleskorn: Good catch, fixed in 6450aae. This was a leftover from an earlier iteration of this where `QueriedTimeRange` returned `int64, int64, bool` rather than the struct.

pkg/streamingpromql/planning/plan_test.go
charleskorn: Good idea, added in a4d4f37.

pkg/streamingpromql/planning/plan_test.go
charleskorn: Good point, added a comment in 4f31fb8.

PR #12397, opened 2025-08-14
ruler: Add support for labels at the rule group level

### LINKED ISSUE #12379: Bug: Missing GroupLabels in alerts
### What is the bug?

I'm using Mimir and Grafana. When sending a prometheus spec [rulefmt.Rulegroup](https://github.com/prometheus/prometheus/blob/main/model/rulefmt/rulefmt.go#L162-L170) i expected grafana and mimir to support those as well. I couldn't find any reference in the documentation or code that suggested otherwise. When looking at the [HTTP handler in mimir](https://github.com/grafana/mimir/blob/main/pkg/ruler/api.go#L622) it's parsing the rulefmt.RuleGroup from the input to the proto serialized version [here](https://github.com/grafana/mimir/blob/main/pkg/ruler/api.go#L713) which uses the [toProto func](https://github.com/grafana/mimir/blob/main/pkg/ruler/rulespb/compat.go#L20). The [RuleGroupDesc](https://github.com/grafana/mimir/blob/main/pkg/ruler/rulespb/rules.proto#L20) does not support any form of group labels. meaning that they get lost here. 

Is this by design, ifso, could you elaborate on why this is.

Ifnot, would it be possible to support grouplabels? One option could be to merge the grouplabels in the the rule labels when parsing, as a first easy step

### How to reproduce it?

send the below payload to `POST /<prometheus-http-prefix>/config/v1/rules/{namespace}`

```yaml
    name: alert-group-1
    interval: 10s
    labels:
      foo: bar
    rules:
      - alert: alertname
        expr: '1>0'
        labels:
           baz: bar
```

i expect my alerts to show 2 labels (`foo: bar` and `baz: bar`) in grafana and/or alertmanager. Unfortunately it only shows the label from the actual alertrule (`baz: bar`)

running LGTM stack with helm on local mac with kind clusters, also on openshift. if you want i can share versions of mimir/grafana but looking at the code it seems irrelevant

_No response_

tliefheid:
#### What this PR does

Add GroupLabel support to Rules

Fixes #12379

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/rulespb/compat_test.go
+++ b/pkg/ruler/rulespb/compat_test.go
@@ -12,7 +12,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 
-	"github.com/grafana/mimir/pkg/mimirpb"
+	"github.com/grafana/mimir/pkg/mimirpb" //lint:ignore faillint allowed to import other protobuf
 )
 
 func TestProtoConversionShouldBeIdempotent(t *testing.T) {


tliefheid:
should an example be added to the documentation?

pkg/ruler/rulespb/compat_test.go
```diff
@@ -178,6 +179,70 @@ func TestToProto(t *testing.T) {
 				Rules:           []*RuleDesc{},
 			},
 		},
+		"check grouplabels are properly propagated": {
+			input: rulefmt.RuleGroup{
+				Name: "group",
+				Labels: map[string]string{
+					"groupKey": "groupValue",
```
56quarters: Can you add another key-value pair here to make sure the labels are ordered in the `RuleGroupDesc` object?

CHANGELOG.md
```diff
@@ -391,11 +392,11 @@
...
-* [BUGFIX] Querier: fix issue where `label_join` could incorrectly return multiple series with the same labels rather than failing with `vector cannot contain metrics with the same labelset`. https://github.com/prometheus/prometheus/pull/15975 #10826
-* [BUGFIX] Querier: fix issue where counter resets on native histograms could be incorrectly under- or over-counted when using subqueries. https://github.com/prometheus/prometheus/pull/15987 #10871
-* [BUGFIX] Querier: fix incorrect annotation emitted when `quantile_over_time` is evaluated over a range with both histograms and floats. https://github.com/prometheus/prometheus/pull/16018 #10884
-* [BUGFIX] Querier: fix duplicated double quotes in invalid label name error from `count_values`. https://github.com/prometheus/prometheus/pull/16054 #10884
+* [BUGFIX] TSDB: fix unknown series errors and possible lost data during WAL replay when series are removed from the head due to inactivity and reappear before the next WAL checkpoint. <https://github.com/prometheus/prometheus/pull/16060> <https://github.com/prometheus/prometheus/pull/16231> #10824 #10955
```
56quarters: Could you please revert the unrelated CHANGELOG changes?

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 * [BUGFIX] Store-gateway: Fix potential goroutine leak by passing the scoped context in LabelValues. #12048
 * [BUGFIX] Distributor: Fix pooled memory reuse bug that can cause corrupt data to appear in the err-mimir-label-value-too-long error message. #12048
 * [BUGFIX] Querier: Fix timeout responding to query-frontend when response size is very close to `-querier.frontend-client.grpc-max-send-msg-size`. #12261
+* [BUGFIX] Ruler: Support GroupLabels. #12397
```
56quarters:
```suggestion
* [BUGFIX] Ruler: Support labels at the rule group level. These were previously ignored even when set via the API. #12397
```

CHANGELOG.md
tliefheid: whoops probably a linter that went rogue. i'll have a look

tliefheid:
updated according to comments

56quarters (approved):
LGTM, thanks!

pkg/ruler/rulespb/compat_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/mimir/pkg/mimirpb"
```
56quarters: `github.com/grafana/mimir` imports need to be in their own group beneath 3rd party imports

tacole02 (approved):
Changelog looks good. Thank you!

pkg/ruler/rulespb/compat_test.go
tliefheid: yeah i noticed indeed, lemme update accordingly

tliefheid:
i'm not familiar with `faillint` and why this is an error, any suggestions on how to resolve this would be appreciated

56quarters:
> i'm not familiar with `faillint` and why this is an error, any suggestions on how to resolve this would be appreciated

I'm not sure why it's failing for the test file but not for the regular file. I'll take a look.

56quarters:
> > i'm not familiar with `faillint` and why this is an error, any suggestions on how to resolve this would be appreciated
> 
> I'm not sure why it's failing for the test file but not for the regular file. I'll take a look.

Oh, we specifically ignore this lint for `compat.go`. Adding the same ignore to the import in `compat_test.go` should fix it.

pkg/ruler/rulespb/compat_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
+
+	"github.com/grafana/mimir/pkg/mimirpb"
```
56quarters:
```suggestion
	"github.com/grafana/mimir/pkg/mimirpb" //lint:ignore faillint allowed to import other protobuf
```

PR #8421, opened 2024-06-19
Add profiling labels to ingester's read path

colega:
#### What this PR does

Adds labels to profiling that identify the read requests.

We're already adding the trace ID when the request is sampled, this would also add the method, userID and matchers of that read request.

This should help debugging issues when ingesters are busy running heavy
queries.

#### Checklist

- [na] Tests updated.
- [na] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`.
- [na] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12404, opened 2025-08-14
ingester: Add configurable tenant ID in profile labels

dimitarvdimitrov:
Adds ability to include tenant ID in pprof labels for ingester operations. This helps debug performance issues when specific tenants are causing high CPU usage or slow queries.

Based on previous work in #8421, but simplified to just include tenant ID and made configurable via CLI flag.

The flag `-include-tenant-id-in-profile-labels` is disabled by default to avoid any overhead when not needed.

Uses a `ProfilingWrapper` pattern similar to `ActivityTrackerWrapper` to cleanly wrap all ingester methods.

### DIFF
--- a/pkg/ingester/api.go
+++ b/pkg/ingester/api.go
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+package ingester
+
+import (
+	"net/http"
+
+	"github.com/grafana/mimir/pkg/ingester/client"
+)
+
+type IngesterAPI interface {
+	client.IngesterServer
+	FlushHandler(http.ResponseWriter, *http.Request)
+	ShutdownHandler(http.ResponseWriter, *http.Request)
+	PrepareShutdownHandler(http.ResponseWriter, *http.Request)
+	PreparePartitionDownscaleHandler(http.ResponseWriter, *http.Request)
+	PrepareUnregisterHandler(w http.ResponseWriter, r *http.Request)
+	UserRegistryHandler(http.ResponseWriter, *http.Request)
+	TenantsHandler(http.ResponseWriter, *http.Request)
+	TenantTSDBHandler(http.ResponseWriter, *http.Request)
+	PrepareInstanceRingDownscaleHandler(http.ResponseWriter, *http.Request)
+}
--- a/pkg/ingester/api.go
+++ b/pkg/ingester/api.go
@@ -2,9 +2,11 @@
 package ingester
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/grafana/mimir/pkg/ingester/client"
+	"github.com/grafana/mimir/pkg/mimirpb"
 )
 
 type IngesterAPI interface {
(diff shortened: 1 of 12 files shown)


pkg/ingester/ingester_profiling.go
```diff
@@ -0,0 +1,345 @@
...
+	return trace.SpanFromContext(ctx).SpanContext().IsSampled()
+}
+
+func (i *ProfilingWrapper) Push(ctx context.Context, request *mimirpb.WriteRequest) (*mimirpb.WriteResponse, error) {
+	if isTraceSampled(ctx) {
+		userID, _ := tenant.TenantID(ctx)
+		labels := pprof.Labels("method", "Ingester.Push", "userID", userID)
+		ctx = pprof.WithLabels(ctx, labels)
+		pprof.SetGoroutineLabels(ctx)
+		defer pprof.SetGoroutineLabels(ctx)
```
narqo: The `defer` call here, and in the methods below, looks off to me. If we want to recover the original context of the goroutine we should schedule the `defer` call **before** we override the local `ctx`, right?

dimitarvdimitrov: oh you're right. i didn't realize the point of that defer

dimitarvdimitrov: fixed in https://github.com/grafana/mimir/pull/12404/commits/5bc79956b14b22023b904599b27738e70995c4e4

narqo (commented):
(nit) I'm not sure if that the part of the plan, but this _likely_ won't work in ingest storage mode. There, the ingester skip `IngesterAPI.Push`, but uses the internal pusher.

pkg/ingester/ingester_profiling.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	return i.ing.Push(ctx, request)
+}
+
+func (i *ProfilingWrapper) QueryStream(request *client.QueryRequest, server client.Ingester_QueryStreamServer) error {
+	ctx := server.Context()
+	if isTraceSampled(ctx) {
+		userID, _ := tenant.TenantID(ctx)
+		labels := pprof.Labels("method", "Ingester.QueryStream", "userID", userID)
```
narqo: (nit) Tell me if I'm wrong, but I suspect the method's name should already be in the profile labels, because [Span Profiler][1] injects it to the root span, under the `span_name` name. Maybe we don't need to both?

[1]: https://github.com/grafana/dskit/tree/5313fe3063cd521d66d8c6e64645daf4d160ced4/spanprofiler

tacole02 (approved):
Docs look good!

pkg/ingester/ingester_profiling.go
dimitarvdimitrov: you're absolutely right! i didn't know this. i'll get rid of the method label

Screenshot 2025-08-18 at 20 22 17

dimitarvdimitrov:
> (nit) I'm not sure if that the part of the plan, but this _likely_ won't work in ingest storage mode. There, the ingester skip `IngesterAPI.Push`, but uses the internal pusher.

good point. I think it's still valuable if we have this labelled as well. See this diff https://github.com/grafana/mimir/pull/12404/commits/fe1c6260656bbe45dbc70f05c13a5a9bbe811a2a

pkg/ingester/ingester_profiling.go
dimitarvdimitrov: removed in https://github.com/grafana/mimir/pull/12404/commits/58de7895625f4b181413beb3f92454be1dc14613

narqo (approved):
Works for me

pkg/ingester/api.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/grafana/mimir/pkg/ingester/client"
+	"github.com/grafana/mimir/pkg/mimirpb"
+)
+
+type IngesterAPI interface {
```
colega: This should be just `API` since the package is already is `ingester` so the full name would be `ingester.API` instead of `ingester.IngesterAPI` (I'm surprised that no linter complained).

dimitarvdimitrov: > I'm surprised that no linter complained

me too

dimitarvdimitrov: fixed in e78a2a1b10f74b8274d16dbb1608812fabbc8efa

PR #12406, opened 2025-08-14
Multiple Components: Updates to CPU and Memory HPA query

dmwilson-grafana:
#### What this PR does

At Grafana Labs, we have encountered cases where rate() extrapolates a metric over missing values, underestimates CPU utilization, and triggers an unnecessary downscaling. This PR switches the default CPU scaler metrics to use `irate()` rather than `rate()` (see:autoscaling.libsonnet) and affects the scaling behavior of the following components:

- alertmanager
- query frontend
- ruler
- ruler querier
- ruler query frontend
- distributor

Because `irate()` can be more volatile, using the maximum of irate can cause _over-scaling_. We mitigate this by using the p75 of the `irate()` distribution. Mimir operators can opt-in to using `irate()` by setting a new config option - `autoscaling_cpu_hpa_use_irate`.

We also remove `or vector(0)` from a leg of the memory metric. In the first leg of the query, we remove `or vector(0)` so that if metrics are unavailable, we part returns no data instead of zeroes to prevents underscaling when metrics are unavailable. We keep `or vector(0)` on the second leg because this part should always contribute a value to the calculation. Mimir operators can opt-in to this change by setting a new config option `autoscaling_memory_hpa_require_metrics`.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


operations/mimir/autoscaling.libsonnet
```diff
@@ -270,36 +276,62 @@
...
-          )[15m:]
-        ) * 1000
-      |||
+      if $._config.autoscaling_cpu_hpa_use_irate then
+        // We use irate() instead of rate() because it is more reliable when scrapes fail. It calculates
+        // the metric from the last two data points in the range, whereas rate() extrapolates missing
+        // values over the full window, which can underestimate CPU utilization and trigger unnecessary
+        // downscaling. To prevent overscaling, we use a lower percentile in irate() compared to rate().
+        |||
+          quantile_over_time(0.65,
```
aldernero, line 291: is this meant to be p65?

dmwilson-grafana: yes. Idea is that b.c. irate is more volatile than rate to avoid over-scaling we just look at a lower percentile. Idea is that (very roughly) p95 of rate ~ p65 irate. Just choosing the last 12h from a ruler-querier as an example, p95 of rate is  ~5, p95 of irate would be ~10, p65 of irate ~ p95 rate. I'm open to alternatives, but from a few spot checks these line up pretty well.

Screenshot 2025-08-19 at 10 27 10 AM

aldernero (approved):
LGTM

operations/mimir/autoscaling.libsonnet
```diff
@@ -270,36 +276,62 @@
...
+          quantile_over_time(0.95,
+            sum(
+              sum by (pod) (rate(container_cpu_usage_seconds_total{container="%(container)s",namespace="%(namespace)s"%(extra_matchers)s}[5m]))
+              and
+              max by (pod) (min_over_time(kube_pod_status_ready{namespace="%(namespace)s",condition="true"%(extra_matchers)s}[1m])) > 0
+            )[15m:]
+          ) * 1000
+        |||
     else
       // To scale out relatively quickly, but scale in slower, we look at the average CPU utilization
```
pracucci: We should update this comment because the way it works is different between rate() and irate(). E.g. in the new query it's **not** the "we look at the average CPU utilization per replica over 5m (rolling window)".

operations/mimir/autoscaling.libsonnet
```diff
@@ -270,36 +276,62 @@
   local cpuHPAQuery(with_ready_trigger) = (
     if with_ready_trigger then
       // To scale out relatively quickly, but scale in slower, we look at the average CPU utilization
```
pracucci: We should update this comment because the way it works is different between rate() and irate(). E.g. in the new query it's **not** the "we look at the average CPU utilization per replica over 5m (rolling window)".

pracucci (commented):
As discussed offline, backtesting the changes in Grafana Cloud I noticed that are some workloads (mainly "distributor") where the **new** CPU-based metric scaling query would run a significantly lower number of replicas, and other workloads (mainly "ruler-querier" and "ruler-query-frontend") where the **new** CPU-based metric scaling query run a significantly higher number of replicas. I suggest you to further investigate it.

In other news, the change to the memory scaling metric is non controversial, so you could just extract it on a separate PR so we merge it.

pracucci (approved):
After more backtesting and analysis I think these queries are a step forward the current ones, in particular to not skrew up too much in case of gaps in the source metrics used by the scaling queries.

PR #12407, opened 2025-08-15
Bring in head stats collection, config, and metrics

chencs:
#### What this PR does

This PR brings in changes from mimir-prometheus for collecting statistics about TSDB head postings, and introduces `HeadStatisticsCollectionFrequency`, a configuration option that will be passed to the TSDB to determine how frequently to collect statistics on the TSDB head.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/ingester/ingester.go
```diff
@@ -2846,6 +2846,8 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
 		StripeSize:                            i.cfg.BlocksStorageConfig.TSDB.StripeSize,
 		HeadChunksWriteBufferSize:             i.cfg.BlocksStorageConfig.TSDB.HeadChunksWriteBufferSize,
 		HeadChunksEndTimeVariance:             i.cfg.BlocksStorageConfig.TSDB.HeadChunksEndTimeVariance,
+		CollectHeadStatistics:                 i.cfg.BlocksStorageConfig.TSDB.CollectHeadStatistics,
```
dimitarvdimitrov: would it be easier to configure if we use `i.cfg.BlocksStorageConfig.TSDB.IndexLookupPlanningEnabled` instead? Planning without stats is invalid and stats without planning is kind of pointless.

(Also the docs on `IndexLookupPlanningEnabled` mention stats collection too.

i think we can also validate this config in `TSDBConfig.Validate`. Also that the collection time is >0 when stats/planning is enabled. Can you do that?

dimitarvdimitrov (commented):
this looks ok, maybe we can make do with fewer config options 👀 

Did you also plan to bridge the stats and the planner? if not, I had a [draft](https://github.com/grafana/mimir/commit/117f4f99eb9eff90f8da5ba28366e22204123feb) with that and can have a go at it tomorrow.

pkg/ingester/ingester.go
chencs: I ended up removing `CollectHeadStatistics` and just treating `0` as disabling stats collection in https://github.com/grafana/mimir-prometheus/pull/962, so I'm updating here to reflect. I am adding a check in `Validate()` for non-negative values, though.

chencs:
> Did you also plan to bridge the stats and the planner? 

I'm not sure what you mean by this. I did remove `interfaces.go` as part of pulling in the mimir-prometheus changes.

dimitarvdimitrov:
> > Did you also plan to bridge the stats and the planner?
> 
> I'm not sure what you mean by this. I did remove `interfaces.go` as part of pulling in the mimir-prometheus changes.

so sorry for missing this comment. What i meant by this is that we don't have anything which passes stats to a planner instance for that planner to do its planning.

PR #12416, opened 2025-08-17
compactor: include sparse index headers in meta.json

### LINKED ISSUE #12037: compactor: include sprase index header in meta.json
There is a comment in the readme of undelete-blocks which says that it will only restore the files that are present in the meta.json file. The meta.json is supposed to contain all files that pertain to the block. 

Currently we don’t store the [sparse header](https://github.com/grafana/mimir/blob/069080f2ea75505f20924112c3e63053b0fd0f3a/pkg/compactor/bucket_compactor.go#L437-L459) in the meta.json ([code](https://github.com/grafana/mimir/blob/5b8a44775c8d7f17f5deebcd6637f8fd10d4a4be/pkg/storage/tsdb/block/block.go#L340-L378) where it's supposed to happen). We should include it there just to avoid confusion in the future.

pranshu-raj-211:
#### What this PR does
Tries to read sparse-index-header (best effort), includes in meta.json if it exists. Skips the file silently if not found, fails for other errors.

Fixes #12037 compactor: include sparse index header in meta.json

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/tsdb/block/block.go
+++ b/pkg/storage/tsdb/block/block.go
@@ -159,28 +159,27 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blockDi
 		return nil
 	})
 
-	if err := eg.Wait(); err != nil {
-		return cleanUp(logger, bkt, id, err)
-	}
-
-	src := filepath.Join(blockDir, SparseIndexHeaderFilename)
-	dst := filepath.Join(id.String(), SparseIndexHeaderFilename)
 	hasSparseHeaderInfo := false
-    for _, f := range meta.Thanos.Files {
-        if f.RelPath == SparseIndexHeaderFilename {
-            hasSparseHeaderInfo = true
-            break
-        }
-    }
-
-	// upload sparse header info if GatherFileStats returned an entry for it, otherwise skip
-	if hasSparseHeaderInfo{
-		if err:=objstore.UploadFile(ctx, logger, bkt, src, dst); err != nil {
-			return cleanUp(logger, bkt, id, &UploadError{err, FileTypeSparseIndexHeader})
+	for _, f := range meta.Thanos.Files {
+		if f.RelPath == SparseIndexHeaderFilename {
+			hasSparseHeaderInfo = true
+			break
 		}
 	}
-	else{
-		level.Debug(logger).Log("msg","sparse index header entry not found, skipping upload", "block", id.String())
+
+	if hasSparseHeaderInfo {
+		eg.Go(func() (err error) {
+			if err := objstore.UploadFile(uctx, logger, bkt, filepath.Join(blockDir, SparseIndexHeaderFilename), path.Join(id.String(), SparseIndexHeaderFilename)); err != nil {
+				return &UploadError{err, FileTypeSparseIndexHeader}
+			}
+			return nil
+		})
+	} else {
+		level.Debug(logger).Log("msg", "sparse index header entry not found, skipping upload", "block", id.String())
+	}
+
+	if err := eg.Wait(); err != nil {
+		return cleanUp(logger, bkt, id, err)
 	}
 
 	// Meta.json always need to be uploaded as a last item. This will allow to assume block directories without meta file to be pending uploads.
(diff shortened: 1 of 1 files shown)


pkg/storage/tsdb/block/block.go
```diff
@@ -365,6 +365,16 @@ func GatherFileStats(blockDir string) (res []File, _ error) {
 	}
 	res = append(res, mf)
 
+	// sparse index headers are best effort, skip if file does not exist
+	sparseInfo, err := os.Stat(filepath.Join(blockDir, SparseIndexHeaderFilename))
```
dimitarvdimitrov: nitpick to name this `sparseHeaderInfo` (info about the sparse header, not info which is sparse)

pkg/storage/tsdb/block/block.go
```diff
@@ -365,6 +365,16 @@ func GatherFileStats(blockDir string) (res []File, _ error) {
...
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "stat %v", filepath.Join(blockDir, SparseIndexHeaderFilename))
+		}
+	} else {
+		res = append(res, File{RelPath: sparseInfo.Name(), SizeBytes: sparseInfo.Size()})
+	}
```
dimitarvdimitrov: doing a soft-failure in `Upload` means that the meta.json would contain this file, but the file might fail to upload. I think at this point we've had enough experience with the sparse headers to be able to make them a required part of the block. 

The challenge is that Upload is called by the ingester who doesn't compute sparse headers and by the compactor who does compute them. So I can think of a few changes

1. Let's make the upload of the sparse header required only if `GatherFileStats` returned a sparse header entry.
2. Can you also update the comment here to clarify why the sparse header might not exist 

I think in a follow-up PR I am going to make sparse header uploading in the compactor always enabled.

dimitarvdimitrov (commented):
thanks for working on this! I left a couple of comments wrt how this works together with the rest of the compaction job, can you take a look?

pkg/storage/tsdb/block/block.go
pranshu-raj-211: Done

pkg/storage/tsdb/block/block.go
pranshu-raj-211: Updated comment, upload of sparse header in latest commit.

One thing I'm not sure about - do we call cleanup when upload of the sparse header fails or just return an error later like the previous version which had an oerr var, which I removed.

pranshu-raj-211: @dimitarvdimitrov Could you take a look and suggest changes to be made, specifically need help on the cleanup part on line 179. Thanks!

pkg/storage/tsdb/block/block.go
```diff
@@ -163,16 +163,25 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blockDi
...
+
+	// upload sparse header info if GatherFileStats returned an entry for it, otherwise skip
+	if hasSparseHeaderInfo{
+		if err:=objstore.UploadFile(ctx, logger, bkt, src, dst); err != nil {
+			return cleanUp(logger, bkt, id, &UploadError{err, FileTypeSparseIndexHeader})
 		}
 	}
+	else{
+		level.Debug(logger).Log("msg","sparse index header entry not found, skipping upload", "block", id.String())
+	}
```
dimitarvdimitrov: can we do this in a goroutine like the other files are done above? that way we can make use of the cleanUp invocation that already exists

pkg/storage/tsdb/block/block.go
dimitarvdimitrov: let's call cleanup if sparse headers fail to upload too (again only if the sparse header was there returned from GatherFileStats)

pranshu-raj-211:
Updated objectstore uploads to use goroutine to match earlier uploads (and make use of cleanUp invocation, which has been shifted to after sparseIndexHeader uploads).

Fixed a couple errors (name change to sparseHeaderInfo was not being used in variable names, would have resulted in errors).

PR #12425, opened 2025-08-18
Alertmanager: Make state read timeout configurable

santihernandezc:
We're currently using a hard-coded 15s timeout for reading state from object storage during the initial sync. This PR adds an arg to make it configurable, keeping the 15s default.


alexander-akhmetov (approved):
LGTM

cmd/mimir/help-all.txt.tmpl
```diff
@@ -361,6 +361,8 @@ Usage of ./cmd/mimir/mimir:
     	Backend storage to use for the ring. Supported values are: consul, etcd, inmemory, memberlist, multi. (default "memberlist")
   -alertmanager.sharding-ring.zone-awareness-enabled
     	True to enable zone-awareness and replicate alerts across different availability zones.
+  -alertmanager.state-read-timeout duration
+    	[experimental] Timeout for reading the state from object storage during the initial sync. 0 = no timeout. (default 15s)
```
56quarters: Would this make more sense under the `alertmanager_store` block/config?

56quarters (commented):
Code looks fine, just a comment about the name / location of the option.

cmd/mimir/help-all.txt.tmpl
tacole02:
```suggestion
    	[experimental] Timeout for reading the state from object storage during the initial sync. Set to `0` for no timeout. (default 15s)
```

tacole02 (approved):
Docs look good!

cmd/mimir/help-all.txt.tmpl
santihernandezc: I'm failing to realize how to add this option to the storage section and having the deadline only affect this specific operation (reading state only in the initial sync).

On another note, updating the storage section results in the following args being added after running `make doc`:
- `-alertmanager-storage.state_read_timeout`
- `-common.storage.state_read_timeout`
- `-ruler-storage.state_read_timeout`
- `-blocks-storage.state_read_timeout`

PR #12435, opened 2025-08-18
Add Compactor Scheduler Skeleton

dmwilson-grafana:

This PR introduces experimental support for a "compactor scheduler" mode that allows the compactor to request jobs from an external scheduler service. This PR makes a few changes to the `MultitenantCompactor` but filters most of the logic for the new mode through a new concept, a `CompactionExecutor`. 

The `CompactionExecutor` currently has two implementations, `StandaloneExecutor` and  `SchedulerExecutor`.  The former is a copy of the traditional compactor's `running` method, the latter implements the experimental execution pattern.

To support the new mode, this PR adds the following configuration options : 
  * `-compactor.mode` Sets operation mode (one of `"standalone"`, `"scheduler"`)
  * `-compactor.scheduler-endpoint` - Scheduler service address
  * `-compactor.scheduler-update-interval` - How often does the compactor send an update to extend its lease on a job with the scheduler endpoint.

 
As of writing there are a few caveats: 
- This PR assumes the scheduler is always up - or the penalty for it going down is trivial. I don't think this is a reasonable long-term assumption.
    - We may consider a job history snapshotting mechanism  (see: `BlockBuilderScheduler, snapOffsets` in `pkg/blockbuilder`) or retries on specific RPCs (e.g. `SchedulerExecutor.updateJobStatus`)

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/compactor/scheduler/spawner.go
```diff
@@ -43,7 +43,7 @@ func NewSpawner(
 	logger log.Logger) *Spawner {
 	s := &Spawner{
 		allowedTenants:       allowList,
-		rotator: rotator,
+		rotator:              rotator,
```
seizethedave, line 46: 🔥

pkg/compactor/compactor.go
```diff
@@ -132,6 +151,13 @@ type Config struct {
 	UploadSparseIndexHeaders       bool               `yaml:"upload_sparse_index_headers" category:"experimental"`
 	SparseIndexHeadersSamplingRate int                `yaml:"-"`
 	SparseIndexHeadersConfig       indexheader.Config `yaml:"-"`
+
+	// Scheduler mode options
+	SchedulingMode          string            `yaml:"scheduling_mode" category:"advanced"`
```
seizethedave: this one is probably `experimental` since it would be deleted in the future. (wouldn't it?)

pkg/compactor/compactor.go
```diff
@@ -486,6 +574,13 @@ func (c *MultitenantCompactor) starting(ctx context.Context) error {
 		return err
 	}
 
+	// Initialize scheduler client.
+	if c.compactorCfg.SchedulingMode == planningModeScheduler {
+		if c.schedulerClient, c.schedulerConn, err = c.makeSchedulerClient(); err != nil {
+			return fmt.Errorf("failed to create scheduler client: %w", err)
```
seizethedave: I'm a `%w` supporter myself, but the rest of this function uses `errors.Wrap` so it kinda sticks out.

pkg/compactor/compactor.go
```diff
@@ -615,6 +725,194 @@ func (c *MultitenantCompactor) running(ctx context.Context) error {
...
+	}
+
+	// clear current job only if it matches the expected job
+	if c.activeJob.key.Id == currentId {
+		c.activeJob = nil
+		return nil
+	}
+
+	return errCannotWriteActiveJob
+}
```
seizethedave: (nit)
if this conditional is inverted, then the entire function follows the pattern of detecting errors first, then terminating with the "happy path" and returning a `nil` error.

i.e.:
```
if id != currentId {
   return err...
}

c.activeJob = nil
return nil
```

pkg/compactor/compactor.go
```diff
@@ -205,6 +235,22 @@ func (cfg *Config) Validate(logger log.Logger) error {
 		return errInvalidCompactionOrder
 	}
 
+	if !util.StringsContain(CompactionPlanningModes, cfg.SchedulingMode) {
```
seizethedave:
```suggestion
	if !slices.Contains(CompactionPlanningModes, cfg.SchedulingMode) {
```

Inspired by this PR, I have deleted that utility function over here: 6550005722cc992354283661ad0b93af6ddc03ee

pkg/compactor/compactor.go
```diff
@@ -132,6 +150,13 @@ type Config struct {
 	UploadSparseIndexHeaders       bool               `yaml:"upload_sparse_index_headers" category:"experimental"`
 	SparseIndexHeadersSamplingRate int                `yaml:"-"`
 	SparseIndexHeadersConfig       indexheader.Config `yaml:"-"`
+
+	// Scheduler mode options
```
andyasp, line 151: All of these configuration options should be marked with `category:"experimental"` since this is all very experimental.

pkg/compactor/compactor.go
```diff
@@ -50,13 +55,26 @@ const (
...
+var CompactionPlanningModes = []string{planningModeScheduler, planningModeStandalone}
+
 var (
 	errInvalidBlockRanges                         = "compactor block range periods should be divisible by the previous one, but %s is not divisible by %s"
 	errInvalidCompactionOrder                     = fmt.Errorf("unsupported compaction order (supported values: %s)", strings.Join(CompactionOrders, ", "))
 	errInvalidMaxOpeningBlocksConcurrency         = fmt.Errorf("invalid max-opening-blocks-concurrency value, must be positive")
 	errInvalidMaxClosingBlocksConcurrency         = fmt.Errorf("invalid max-closing-blocks-concurrency value, must be positive")
 	errInvalidSymbolFlushersConcurrency           = fmt.Errorf("invalid symbols-flushers-concurrency value, must be positive")
 	errInvalidMaxBlockUploadValidationConcurrency = fmt.Errorf("invalid max-block-upload-validation-concurrency value, can't be negative")
+	errInvalidPlanningMode                        = fmt.Errorf("invalid compactor mode, supported values: %s", strings.Join(CompactionPlanningModes, ", "))
```
andyasp: I think these errors would be more helpful if they referenced the name of the flag directly like above. For instance, this could be `"invalid scheduling-mode, supported values: ..."`

pkg/compactor/compactor.go
```diff
@@ -615,6 +721,201 @@ func (c *MultitenantCompactor) running(ctx context.Context) error {
...
+		WorkerId: workerID,
+	}
+
+	resp, err := c.schedulerClient.LeaseJob(ctx, req)
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "failed to lease job from scheduler", "err", err)
+		return false, err
+	}
+
+	if resp.Key == nil || resp.Spec == nil {
```
andyasp: This feels like an imaginary case. When would this be true? An error is returned when there is no job to lease.

pkg/compactor/compactor.go
```diff
@@ -615,6 +721,201 @@ func (c *MultitenantCompactor) running(ctx context.Context) error {
...
+}
+
+func (c *MultitenantCompactor) executeCompactionJob(ctx context.Context, spec *schedulerpb.JobSpec) (schedulerpb.UpdateType, error) {
+	return schedulerpb.COMPLETE, nil
+}
+
+// startJobUpdates sends periodic keep-alive updates to the scheduler and removes jobs that have exceeded
+// max job
+func (c *MultitenantCompactor) startJobStatusUpdates(ctx context.Context) {
+	if c.compactorCfg.SchedulingMode != planningModeScheduler {
```
andyasp: I think these kinds of guard checks are a sign that the code is in the wrong place.

pkg/compactor/compactor.go
```diff
@@ -280,6 +334,14 @@ type MultitenantCompactor struct {
 	// Client used to run operations on the bucket storing blocks.
 	bucketClient objstore.Bucket
 
+	// Scheduler client for requesting jobs when in scheduler mode.
```
andyasp: I don't think these belong in `MultitenantCompactor`. I would expect something higher level here that contained them.

pkg/compactor/compactor.go
```diff
@@ -615,6 +721,201 @@ func (c *MultitenantCompactor) running(ctx context.Context) error {
...
+	b := backoff.New(ctx, backoff.Config{
+		MinBackoff: 100 * time.Millisecond,
+		MaxBackoff: c.compactorCfg.CompactionInterval,
+	})
+
+	for {
+
+		work, err := c.leaseAndExecuteJob(ctx, workerID)
+		if err != nil {
+			level.Warn(c.logger).Log("msg", "failed to lease or execute job", "err", err)
```
andyasp: `leaseAndExecuteJob` is logging errors as well so this is often double logging

pkg/compactor/executor.go
```diff
@@ -0,0 +1,287 @@
...
+	return nil
+}
+
+func (e *StandaloneExecutor) CreateShardingStrategy(enabledTenants, disabledTenants []string, ring *ring.Ring, ringLifecycler *ring.BasicLifecycler, cfgProvider ConfigProvider) shardingStrategy {
+	allowedTenants := util.NewAllowList(enabledTenants, disabledTenants)
+	return newSplitAndMergeShardingStrategy(allowedTenants, ring, ringLifecycler, cfgProvider)
+}
+
+// SchedulerExecutor requests compaction jobs from an external scheduler.
+type SchedulerExecutor struct {
```
seizethedave: Anything in this file that is only used in the `compactor` package can (should) be unexported.

pkg/compactor/scheduler/schedulerpb/scheduler.proto
```diff
@@ -14,9 +14,15 @@ message JobKey {
   int64 epoch = 2;
 }
 
+enum JobType {
```
andyasp, line 17: An empty blockIDs was the intent to signal a plan job, was that not sufficient? I could possibly see the argument for a JobType if cleanup was added later, but I'm not following why this is needed currently.

docs/sources/mimir/configure/about-versioning.md
```diff
@@ -72,6 +72,10 @@ The following features are currently experimental:
     - `-compactor.max-lookback`
   - Enable the compactor to upload sparse index headers to object storage during compaction cycles.
     - `-compactor.upload-sparse-index-headers`
+  - Scheduler mode for requesting compaction jobs from an external scheduler service instead of local planning
+    - `-compactor.mode`
```
andyasp:
```suggestion
    - `-compactor.planning-mode`
```

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+		if work {
+			b.Reset()
+		}
+
+		select {
+		case <-time.After(b.NextDelay()):
+			continue
+		case <-ctx.Done():
+			return nil
+		case err := <-c.ringSubservicesWatcher.Chan():
```
andyasp: Why would we care if the ring failed if we aren't using it in this executor? Are we?

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+		return false, err
+	}
+
+	jobID := resp.Key.Id
+	jobTenant := resp.Spec.Tenant
+	jobType := resp.Spec.JobType
+
+	// TODO: Consider retry logic for status update failures. We currently only log a warning if the scheduler is unreachable.
+	// First and intermediate status updates will be retried when ticker.C fires, what if scheduler is down for final status update?
+	if err := e.updateJobStatus(ctx, resp.Key, resp.Spec, schedulerpb.IN_PROGRESS); err != nil {
```
andyasp: Immediately marking the job as in progress is unnecessary

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+		case status := <-statusChan:
+			if err := e.updateJobStatus(ctx, key, spec, status); err != nil {
+				level.Warn(e.logger).Log("msg", "failed to send final status update", "job_id", jobId, "tenant", jobTenant, "status", status, "err", err)
+			}
+			return
+		}
+	}
+}
+
+func (e *schedulerExecutor) createShardingStrategy(enabledTenants, disabledTenants []string, ring *ring.Ring, ringLifecycler *ring.BasicLifecycler, cfgProvider ConfigProvider) shardingStrategy {
```
andyasp: I'm still not following why `createShardingStrategy` exists. Once we have a job and want to execute it aren't we past the stage of where the compactor uses the ring for compaction jobs? Why would some other part of the code want to know, from the executor, this information?

pkg/compactor/compactor.go
```diff
@@ -154,6 +175,12 @@ func (cfg *Config) RegisterFlags(f *flag.FlagSet, logger log.Logger) {
...
 	f.IntVar(&cfg.CleanupConcurrency, "compactor.cleanup-concurrency", 20, "Max number of tenants for which blocks cleanup and maintenance should run concurrently.")
 	f.StringVar(&cfg.CompactionJobsOrder, "compactor.compaction-jobs-order", CompactionOrderOldestFirst, fmt.Sprintf("The sorting to use when deciding which compaction jobs should run first for a given tenant. Supported values are: %s.", strings.Join(CompactionOrders, ", ")))
+	f.StringVar(&cfg.PlanningMode, "compactor.planning-mode", planningModeStandalone, fmt.Sprintf("Compactor operation mode. Supported values are: %s (plan and execute compactions), %s (request jobs from a remote scheduler).", planningModeStandalone, planningModeScheduler))
+	f.StringVar(&cfg.SchedulerAddress, "compactor.scheduler-endpoint", "", "Compactor scheduler endpoint. Required when compactor mode is 'scheduler'.")
+	f.DurationVar(&cfg.SchedulerUpdateInterval, "compactor.scheduler-update-interval", 15*time.Second, "Interval between scheduler job lease updates.")
+	f.DurationVar(&cfg.SchedulerMinBackoff, "compactor.scheduler-min-backoff", 100*time.Millisecond, "Minimum backoff time between scheduler job lease requests.")
```
andyasp: `compactor.scheduler-min-backoff` (and the max equivalent) doesn't make it clear that this is for the leasing specifically, maybe `compactor.scheduler-lease-min-backoff`?

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+	// start async status updater, handles renewing lease every cfg.SchedulerUpdateInterval
+	go e.startJobStatusUpdater(statusCtx, resp.Key, resp.Spec, statusChan)
+
+	var status schedulerpb.UpdateType
+
+	if jobType == schedulerpb.COMPACTION {
+		status, err = e.executeCompactionJob(ctx, c, resp.Spec)
+		if err != nil {
+			level.Warn(e.logger).Log("msg", "failed to execute job", "job_id", jobID, "tenant", jobTenant, "job_type", jobType, "err", err)
+			statusChan <- schedulerpb.ABANDON
```
andyasp: The intent of ABANDON was if we knew a job would never succeed due to some condition (like the blocks don't even exist anymore). It's a signal for the scheduler to drop it. If that's unclear at this stage defer to using REASSIGN.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+		plannedJobs, planErr := e.executePlanningJob(ctx, c, resp.Spec)
+		if planErr != nil {
+			level.Warn(e.logger).Log("msg", "failed to execute planning job", "job_id", jobID, "tenant", jobTenant, "job_type", jobType, "err", planErr)
+			statusChan <- schedulerpb.ABANDON
+			return true, planErr
+		}
+
+		if err := e.sendPlannedJobs(ctx, resp.Spec, plannedJobs); err != nil {
+			level.Warn(e.logger).Log("msg", "failed to send planned jobs", "job_id", jobID, "tenant", jobTenant, "num_jobs", len(plannedJobs), "err", err)
+			statusChan <- schedulerpb.ABANDON
```
andyasp: This is an incorrect flow. Execution should look like:

Loop:
1. Ask for work
2. Received work, work on it and send IN_PROGRESS updates during
3. Send final result

Currently the code will:

- Still possibly be sending concurrent IN_PROGRESS updates while at 3.
- If 3 failed it follows with an ABANDON update. That's not helpful. The only message that needs to be communicated at this stage is that the job is done. If we can't communicate that then we're waiting for the scheduler to tell us it doesn't care anymore.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+
+// startJobStatusUpdater starts a goroutine that handles periodic keep-alive updates and final status updates
+func (e *schedulerExecutor) startJobStatusUpdater(ctx context.Context, key *schedulerpb.JobKey, spec *schedulerpb.JobSpec, statusChan <-chan schedulerpb.UpdateType) {
+	ticker := time.NewTicker(e.cfg.SchedulerUpdateInterval)
+	defer ticker.Stop()
+
+	jobId := key.Id
+	jobTenant := spec.Tenant
+
+	for {
```
andyasp, line 150: There should be a path for when we want to stop sending status updates because the job has been completed. I think mixing in the non in-progress updates is obscuring the intended flow because that mixes two states.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+
+	return []*schedulerpb.PlannedCompactionJob{plannedJob}, nil
+}
+
+// sendPlannedJobs sends the planned compaction jobs back to the scheduler.
+func (e *schedulerExecutor) sendPlannedJobs(ctx context.Context, spec *schedulerpb.JobSpec, plannedJobs []*schedulerpb.PlannedCompactionJob) error {
+	req := &schedulerpb.PlannedJobsRequest{
+		Key: &schedulerpb.JobKey{
```
andyasp: Use the `JobKey` from the `LeaseJobResponse`. A dummy key means the scheduler has no connection to what job this is.

pkg/compactor/scheduler/schedulerpb/scheduler.proto
dmwilson-grafana: It's just easier to check a field's value instead of `len(resp.spec.job.blockIds)`

andyasp: I don't think we should add the field then since the information it contains be determined without it.

pkg/compactor/executor.go
dmwilson-grafana: `e.createShardingStrategy` is only called when we initialize the `MultitenantCompactor`. I added this so we can have the executor be the only object that's concerned with the ownership checks in `shardingStrategy`. Because ownership checks depend on the planning-mode, seems we should have the executor  set them. The alternative would be to create a `splitAndMergeShardingStrategy` and selectively use or not use `blocksCleanerOwnsUser`, `compactorOwnsUser`, and `ownJob` based on the mode.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			level.Warn(e.logger).Log("msg", "failed to execute job", "job_id", jobID, "tenant", jobTenant, "job_type", jobType, "err", err)
+			e.sendFinalJobStatus(ctx, resp.Key, resp.Spec, schedulerpb.REASSIGN)
+			return true, err
+		}
+		e.sendFinalJobStatus(ctx, resp.Key, resp.Spec, status)
+		return true, nil
+	}
+
+	if jobType == schedulerpb.PLANNING {
```
seizethedave:
```go
if t == compaction {
   ...
} else if t == planning {
   ...
} else {
   panic("unexpected job type")
}
```

common defensive pattern.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,324 @@
...
+	return true, nil
+}
+
+func (e *schedulerExecutor) updateJobStatus(ctx context.Context, key *schedulerpb.JobKey, spec *schedulerpb.JobSpec, updType schedulerpb.UpdateType) error {
+	switch spec.JobType {
+	case schedulerpb.COMPACTION:
+		req := &schedulerpb.UpdateCompactionJobRequest{Key: key, Tenant: spec.Tenant, Update: updType}
+		_, err := e.schedulerClient.UpdateCompactionJob(ctx, req)
+		return err
+	default:
```
seizethedave: ditto... better to list the cases and default to an error or panic.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,325 @@
...
+				return false
+			}
+		}).
+		WithBackoffFactor(1*time.Second, 32*time.Second, 2.0).
+		WithJitter(500 * time.Millisecond).
+		Build()
+}
+
+var (
+	plannedJobsRetryPolicy = createSchedulerRetryPolicy[*schedulerpb.PlannedJobsResponse]()
```
jhalterman: Retry policies are stateless, so we can share the same policy for multiple jobs if the policy is the same. We can also use `any` as the result type, since the result isn't used by any of our Failsafe calls or the retry policy. So the policy could be something like:

```go
var (
	retryPolicy = retrypolicy.Builder[any]().
		HandleIf(func(_ any, err error) bool {
			errStatus, ok := grpcutil.ErrorToStatus(err)
			if !ok {
				return false
			}
			switch code := errStatus.Code(); code {
			case codes.Unavailable, codes.DeadlineExceeded, codes.ResourceExhausted:
				return true
			default:
				return false
			}
		}).
		WithBackoffFactor(1*time.Second, 32*time.Second, 2.0).
		WithJitter(500 * time.Millisecond).
		Build()
)
```

pkg/compactor/executor.go
```diff
@@ -0,0 +1,325 @@
...
+// Compaction jobs send final statuses on completion, planning jobs only on failure for reassignment.
+func (e *schedulerExecutor) sendFinalJobStatus(ctx context.Context, key *schedulerpb.JobKey, spec *schedulerpb.JobSpec, status schedulerpb.UpdateType) {
+	jobId := key.Id
+	jobTenant := spec.Tenant
+
+	var err error
+	switch spec.JobType {
+	case schedulerpb.COMPACTION:
+		req := &schedulerpb.UpdateCompactionJobRequest{Key: key, Tenant: spec.Tenant, Update: status}
+		_, err = failsafe.Get(func() (*schedulerpb.UpdateJobResponse, error) {
```
jhalterman: Since we're not returning a result from the Failsafe call, we can use `.Run` instead of `.Get`:

```go
	err := failsafe.Run(func() error {
		_, err := e.schedulerClient.PlannedJobs(ctx, req)
		return err
	}, retryPolicy)
```

And similar for the other calls below

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+	"google.golang.org/grpc/codes"
+
+	"github.com/grafana/mimir/pkg/compactor/scheduler/schedulerpb"
+	"github.com/grafana/mimir/pkg/util"
+)
+
+var (
+	// retryPolicy retries indefinitely on transient scheduler errors. This is used for to avoid discarding
+	// completed work due to temporary scheduler unavailability.
+	retryPolicy = retrypolicy.Builder[any]().
```
andyasp: The default max retries of a `retrypolicy` is 2. You can pass -1 for max attempts to not have a limit ([documentation](https://failsafe-go.dev/retry/#max-attempts)).

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+		HandleIf(func(_ any, err error) bool {
+			errStatus, ok := grpcutil.ErrorToStatus(err)
+			if !ok {
+				return false
+			}
+			switch code := errStatus.Code(); code {
+			case codes.Unavailable, codes.DeadlineExceeded, codes.ResourceExhausted:
```
andyasp: Where are these status codes coming from? The scheduler code isn't sending them.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+				return true
+			default:
+				return false
+			}
+		}).
+		WithBackoffFactor(1*time.Second, 32*time.Second, 2.0).
```
andyasp: This is hardcoded and because you've consolidated into a single retry policy and it's a global the structure isn't in place to make it configurable like lease backoffs are (there's also now two different backoff methods used, but that isn't a huge deal). Can it be configurable?

pkg/compactor/executor.go
andyasp: > I added this so we can have the executor be the only object that's concerned with the ownership checks in shardingStrategy

The scheduler executor isn't concerned with the ownership checks though, is it? It's acting as a factory for an object it's not using. Couldn't the same strategy be used between both modes? The blocks cleaner should be the same.

pkg/compactor/executor.go
dmwilson-grafana: I skimmed this [list](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) and added the ones that _might_ be returned. I think the only case that matters for now is `UNAVAILABLE`, which is generated on the client side when the connection can't be established. Going to just use `UNAVAILABLE` for now.

pkg/compactor/executor.go
dmwilson-grafana: sure, adding this.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return executor, nil
+}
+
+// retryPolicy retries indefinitely on transient scheduler errors. This is used to avoid discarding
+// completed work due to temporary scheduler unavailability.
+func (e *schedulerExecutor) retryPolicy() failsafe.Policy[any] {
```
jhalterman: We should be able to create just a single instance of the retrypolicy as a package var. Since these are stateless, no need to create multiple instances.

pkg/compactor/executor.go
dmwilson-grafana: > The scheduler executor isn't concerned with the ownership checks though, is it? It's acting as a factory for an
  object it's not using. Couldn't the same strategy be used between both modes? The blocks cleaner should be the same.

Not directly, but the scheduler executor is concerned with ensuring ownership checks don't short-circuit scheduler-assigned jobs. In a future PR we'll implement `executeCompactionJob`, and likely  will need to call `compactUser`, `filterOwnJobs`, or `BucketCompactor.Compact` these currently do ring-based checks and the strategy from the scheduler executor needs to either bypass or override those checks since job ownership should be independent of the ring.

pkg/compactor/executor.go
andyasp: The configuration values for the backoff wouldn't be set at initialization time.

pkg/compactor/executor.go
andyasp: > In a future PR we'll implement executeCompactionJob, and likely will need to call compactUser, filterOwnJobs, or BucketCompactor.Compact these currently do ring-based checks and the strategy from the scheduler executor needs to either bypass or override those checks since job ownership should be independent of the ring.

`compactUser` compacts an entire user. `filterOwnJobs` to filter the one job? `BucketCompactor.Compact` includes discovering jobs. Is your plan to try to nullify all of that along the way? That sounds very complex.

 I haven't attempted to scaffold it, but I figured we'd be targeting a more direct pathway to `BucketCompactor.runCompactionJob`.

andyasp: For the above I mean for compaction jobs, not planning jobs to be clear.

pkg/compactor/executor.go
jhalterman: Right - when the executor is constructed would work.

pkg/compactor/executor.go
dmwilson-grafana: > I haven't attempted to scaffold it, but I figured we'd be targeting a more direct pathway to BucketCompactor.runCompactionJob.

That is a good idea. I hadn't thought this bit all the way through either. I just anticipate that we may need to be able to use a code path that calls `ownJob`.

pkg/compactor/scheduler/schedulerpb/scheduler.proto
jhalterman: FWIW, I like being explicit about the job type, even with just 2, since this is part of a spec.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	executor.retryPol = retrypolicy.Builder[any]().
+		HandleIf(func(_ any, err error) bool {
+			errStatus, ok := grpcutil.ErrorToStatus(err)
+			if !ok {
+				return false
+			}
+			switch code := errStatus.Code(); code {
+			case codes.Unavailable:
```
jhalterman: Do we want to retry on DeadlineExceeded and ResourceExhausted as well? Those are traditionally considered retryable, but we could be careful about retrying the latter. FWIW, the `failsafegrpc` package has a policy that handles these by default ([doc](https://failsafe-go.dev/grpc/#retrying-grpc-failures), [code](https://github.com/failsafe-go/failsafe-go/blob/main/failsafegrpc/policy.go#L21)) if you want to retry them all.

cmd/mimir/help-all.txt.tmpl
```diff
@@ -1039,6 +1039,12 @@ Usage of ./cmd/mimir/mimir:
     	Comma separated list of tenants that cannot be compacted by the compactor. If specified, and the compactor would normally pick a given tenant for compaction (via -compactor.enabled-tenants or sharding), it will be ignored instead.
   -compactor.enabled-tenants comma-separated-list-of-strings
     	Comma separated list of tenants that can be compacted. If specified, only these tenants will be compacted by the compactor, otherwise all tenants can be compacted. Subject to sharding.
+  -compactor.executor-max-retry-backoff duration
+    	[experimental] Maximum backoff time for compaction executor retries when sending scheduler status updates. (default 32s)
+  -compactor.executor-min-retry-backoff duration
+    	[experimental] Minimum backoff time for compaction executor retries when sending scheduler status updates. (default 1s)
+  -compactor.executor-retry-backoff-factor float
```
jhalterman: IMO, the backoff factor of 2 is a fine setting for almost any situation, and doesn't need to be configurable. Elsewhere in Mimir we just have the min and max backoffs be configurable.

pkg/compactor/executor_test.go
```diff
@@ -0,0 +1,438 @@
...
+
+package compactor
+
+import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	//"github.com/failsafe-go/failsafe-go/retrypolicy"
```
jhalterman: Can we clean these comments

pkg/compactor/executor_test.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+func (m *mockCompactorSchedulerClient) GetUpdateJobCallCount() int {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.updateJobCallCount
+}
+
+func TestSchedulerExecutor_JobStatusUpdates(t *testing.T) {
+
```
jhalterman: Total nit - but I notice a blank line at the beginning of functions here and there in this PR. Is this intended?

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+// Compaction jobs send final statuses on completion, planning jobs only on failure for reassignment.
+func (e *schedulerExecutor) sendFinalJobStatus(ctx context.Context, key *schedulerpb.JobKey, spec *schedulerpb.JobSpec, status schedulerpb.UpdateType) {
+	jobId := key.Id
+	jobTenant := spec.Tenant
+
+	var err error
+	switch spec.JobType {
+	case schedulerpb.COMPACTION:
+		req := &schedulerpb.UpdateCompactionJobRequest{Key: key, Tenant: spec.Tenant, Update: status}
+		err = failsafe.Run(func() error {
```
jhalterman: Just realized we should pass the context to the failsafe executor, so retries stop when the service does:

```go
err = failsafe.NewExecutor(e.retryPol).WithContext(ctx).Run(func() error {
	_, err := e.schedulerClient.UpdateCompactionJob(ctx, req)
	return err
})
```

Same for the other failsafe.Run calls

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+	"google.golang.org/grpc/codes"
+
+	"github.com/grafana/mimir/pkg/compactor/scheduler/schedulerpb"
+	"github.com/grafana/mimir/pkg/util"
+)
+
+// compactionExecutor defines how compaction work is executed.
+type compactionExecutor interface {
+	run(ctx context.Context, compactor *MultitenantCompactor) error
+	createShardingStrategy(enabledTenants, disabledTenants []string, ring *ring.Ring, ringLifecycler *ring.BasicLifecycler, cfgProvider ConfigProvider) shardingStrategy
```
jhalterman: Instead of accepting enabled and disabled tenants, could it just accept an AllowList?

pkg/compactor/executor.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+}
+
+func (e *standaloneExecutor) stop() error {
+	return nil
+}
+
+func (e *standaloneExecutor) createShardingStrategy(enabledTenants, disabledTenants []string, ring *ring.Ring, ringLifecycler *ring.BasicLifecycler, cfgProvider ConfigProvider) shardingStrategy {
+	allowedTenants := util.NewAllowList(enabledTenants, disabledTenants)
+	return newSplitAndMergeShardingStrategy(allowedTenants, ring, ringLifecycler, cfgProvider)
```
jhalterman: The contents of the `newSplitAndMergeShardingStrategy` could probably move here now and that method could be removed if it won't be used anywhere else? This would also be symmetrical with the schedulerExecutor.createShardingStrategy

pkg/compactor/executor.go
jhalterman: Is the idea here that the executor owns strategy construction because the strategy is 1:1 with the executor, and presumably there would be no mixing/matching strategies/executors? If that's the case, having it where it is seems fine.

pkg/compactor/scheduler/schedulerpb/scheduler.proto
andyasp: I'm fine with dropping this. If we want to add it then it has to be added to the messages that are sent from the scheduler code.

pkg/compactor/executor.go
andyasp: > I just anticipate that we may need to be able to use a code path that calls ownJob.

Can we do this when such a code path is introduced then?

> Is the idea here that the executor owns strategy construction because the strategy is 1:1 with the executor, and presumably there would be no mixing/matching strategies/executors? If that's the case, having it where it is seems fine.

My main question is why there are different strategies being built in the first place. No code in the scheduler executor is doing anything that depends on the strategy (it's stubbed). Doing nothing and keeping the same existing strategy for both works here.

pkg/compactor/executor_test.go
dmwilson-grafana: not intentional, I can remove them.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	jobID := resp.Key.Id
+	jobTenant := resp.Spec.Tenant
+	jobType := resp.Spec.JobType
+
+	// Start async keep-alive updater for periodic IN_PROGRESS messages
+	go e.startJobStatusUpdater(context.WithoutCancel(ctx), resp.Key, resp.Spec)
```
andyasp: This goroutine is passed `context.WithoutCancel(ctx)`, but that means that its Done channel is `nil` ([docs](https://pkg.go.dev/context#WithoutCancel)). The only way it stops though is if it receives from the Done channel (line 157):
```go
    for {
		select {
		case <-ticker.C:
			if err := e.updateJobStatus(ctx, key, spec, schedulerpb.IN_PROGRESS); err != nil {
				level.Warn(e.logger).Log("msg", "failed to send keep-alive update", "job_id", jobId, "tenant", jobTenant, "err", err)
			}
		case <-ctx.Done():
			return
		}
	}
```

So these updater goroutines leak and will never stop sending the in progress updates.

andyasp: The in-progress updates also shouldn't be unconditional with only a warning. The heartbeat responses are the opportunity for the scheduler to notify the worker if the job has been canceled from the scheduler side. When that happens it returns the not found status (meaning the lease was not found).

dmwilson-grafana: Good catch. Just updated.

pkg/compactor/executor.go
```diff
@@ -0,0 +1,313 @@
...
+		return false, err
+	}
+
+	jobID := resp.Key.Id
+	jobTenant := resp.Spec.Tenant
+	jobType := resp.Spec.JobType
+
+	// Create a cancellable context for this job that can be canceled if the scheduler cancels the job
+	jobCtx, cancelJob := context.WithCancelCause(ctx)
+	defer cancelJob(nil)
```
andyasp: `defer` is too late since it could be sending in progress updates all the way until after the final status is already sent. The right place would be after `executeCompactionJob` or `executePlanningJob` returns (regardless of whether it's an error or not).

andyasp (commented):
Looking good, one last thing I can see

andyasp (approved):
Thanks! LGTM

PR #12443, opened 2025-08-19
blockbuilder: reuse ingester-storage pusher consumer

narqo:
#### What this PR does

This PR updates the `blockbuilder` to reuse the `PusherConsumer` from the `storage/ingest`. This is so the block-builder could benefit from the performance improvements of the latter.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/blockbuilder/tsdb_test.go
```diff
@@ -405,111 +399,42 @@ func mockUploaderFunc(t *testing.T, destDir string) blockUploader {
...
-	err = builder.Process(context.Background(), &rec)
+	req2 := mimirpb.WriteRequest{}
+	req2.Timeseries = make([]mimirpb.PreallocTimeseries, 0)
+	err = builder.PushToStorageAndReleaseRequest(ctx, &req2)
 	require.NoError(t, err)
 
 	require.NoError(t, builder.tsdbs[tsdbTenant{0, userID}].Close())
 }
 
-func TestTSDBBuilder_KafkaRecordVersion(t *testing.T) {
```
narqo, line 441: Note to a reviewer: I had to rework this test completely because, with the changes in this PR, the parts about `KafkaRecordVersion` are hidden from the block-builder.

pkg/storage/ingest/pusher.go
```diff
@@ -34,39 +34,41 @@ type PusherCloser interface {
...
 
-	pusher Pusher
+	deserializer DeserializerFunc
+	pusher       Pusher
 }
 
-// newPusherConsumer creates a new pusherConsumer instance.
-func newPusherConsumer(pusher Pusher, kafkaCfg KafkaConfig, metrics *pusherConsumerMetrics, logger log.Logger) *pusherConsumer {
+// NewPusherConsumer creates a new PusherConsumer instance.
+func NewPusherConsumer(deserializer DeserializerFunc, pusher Pusher, kafkaCfg KafkaConfig, metrics *PusherConsumerMetrics, logger log.Logger) *PusherConsumer {
```
dimitarvdimitrov: can you make the deserializer an interface? It's super hard to trace down which are the different deserializers if it's a func passed down many layers

pkg/storage/ingest/pusher.go
```diff
@@ -34,39 +34,41 @@ type PusherCloser interface {
...
 // It'll use a separate goroutine to unmarshal the next record while we push the current record to storage.
-func (c pusherConsumer) Consume(ctx context.Context, records []record) (returnErr error) {
+func (c PusherConsumer) Consume(ctx context.Context, records []Record) (returnErr error) {
 	defer func(processingStart time.Time) {
 		c.metrics.processingTimeSeconds.Observe(time.Since(processingStart).Seconds())
 	}(time.Now())
 
 	type parsedRecord struct {
-		*mimirpb.PreallocWriteRequest
+		mimirpb.PreallocWriteRequest
```
dimitarvdimitrov: why did you make this change? I couldn't figure out why it's needed

pkg/storage/ingest/reader.go
```diff
@@ -45,28 +45,39 @@ var (
 	errUnknownPartitionLeader                   = fmt.Errorf("unknown partition leader")
 )
 
-type record struct {
-	// Context holds the tracing (and potentially other) info, that the record was enriched with on fetch from Kafka.
+type Record struct {
```
dimitarvdimitrov: what's the point of exposing the Record if all the fields are unexported? I also don't see it used outside of this package

pkg/storage/ingest/reader.go
```diff
@@ -45,28 +45,39 @@ var (
...
+		version:  ParseRecordVersion(rec),
+	}
+}
+
+type RecordConsumer interface {
 	// Consume consumes the given records in the order they are provided. We need this as samples that will be ingested,
 	// are also needed to be in order to avoid ingesting samples out of order.
 	// The function is expected to be idempotent and incremental, meaning that it can be called multiple times with the same records, and it won't respond to context cancellation.
-	Consume(context.Context, []record) error
+	Consume(context.Context, []Record) error
```
dimitarvdimitrov: WDYT about adding a new method which just takes `[]*kgo.Fetches` instead? right now we're exporting a type which is practically unusable by other packages only for other packages to use a function we export (`RecordFromKafkaRecord`) to pass this type to us. We can do the conversion ourselves, no?

pkg/blockbuilder/blockbuilder.go
```diff
@@ -292,50 +303,43 @@ consumerLoop:
 			}
 		})
 
-		for recIter := fetches.RecordIter(); !recIter.Done(); {
-			rec := recIter.Next()
+		records := make([]ingest.Record, 0, fetches.NumRecords())
```
dimitarvdimitrov: nit to reuse this slice between different `Consume`s

pkg/blockbuilder/blockbuilder.go
```diff
@@ -292,50 +303,43 @@ consumerLoop:
...
 		}
 	}
 
 	// Nothing was consumed from Kafka at all.
-	if firstRec == nil {
+	if firstRecOffset == -1 {
 		level.Info(logger).Log("msg", "no records were consumed")
 		return
 	}
-
```
dimitarvdimitrov, line 322: i actually don't see the return value from this function being used anywhere. Can you also drop `lastRecOffset` too? (maybe in a separate PR); also i'm not sure if we need `firstRecOffset` either since `builder.CompactAndUpload` already short-cuts on `len(b.tsdbs) == 0`

pkg/blockbuilder/tsdb_test.go
dimitarvdimitrov: to me this means that the record parsing and validation are no longer a responsibility of the block-builder and you can skip testing this in `pkg/blockbuilder`. Your call

pkg/storage/ingest/reader.go
narqo: There is the `RecordFromKafkaRecord` constructor function for this type — a user-package must use the constructor to make an instance of the data type, and it can't mutate it after. I don't understand why it's a wrong pattern, it's a "sealed data type" ;)

I need to expose the type because I don't want to also refactor the guts in `pkg/storage/ingest.PartitionReader`. See [`PartitionReader.consumeFetches`](https://github.com/grafana/mimir/blob/4e3f469544bd7a336048206b3701cb7256236f9e/pkg/storage/ingest/reader.go#L545). Can we maybe do it separately?

pkg/storage/ingest/reader.go
narqo: Do you mean `[]kgo.Fetches` or `[]kgo.Record`? In both cases, this requires to rethink the internals of the `PartitionReader`. I'm not a fan of touching this code. See comment above https://github.com/grafana/mimir/pull/12443/files/622d1765737e82a4805ad359b9234d20859d68a6#r2287103164

pkg/blockbuilder/blockbuilder.go
narqo: I think you're right, but I'd do that in a separate PR. Before the scheduler, there was a reason why the block-builder needed these first/last markers here. I'll need to understand what has changed since then.

pkg/blockbuilder/tsdb_test.go
narqo: I've removed the test completely 👍

pkg/storage/ingest/pusher.go
narqo: I've reverted this change. I think it was an unnecessary cosmetic change

pkg/storage/ingest/pusher.go
narqo: I decided I don't want to expose the deserializer at all, and just use the `PusherConsumer` in the block-builder as is. It makes the PR smaller and is good enough for now

pkg/blockbuilder/blockbuilder.go
narqo: Done in ~~https://github.com/grafana/mimir/pull/12443/commits/dbb8a0462f941abf3ec9af285a35b3e560fedf5b~~ https://github.com/grafana/mimir/pull/12443/commits/070459bf79b282a3480cb0138a6fdaf2a1235ea5

pkg/storage/ingest/reader.go
dimitarvdimitrov: i meant `[]*kgo.Record`. Why is it a big change? you just run the conversion loop which in the current version of the pr the block builder needs to run itself and then feed that into `Consume()`


```go
func (c) ConsumeKgo(ctx, records []*kgo.Record) error {
  internalRecs := make([]record, 0, len(records))
  for _, r := range records {
    internalRecs = append(internalRecs, RecordFromKafkaRecord(r)
  }
  return  c.Consume(ctx, internalRecs)
}

dimitarvdimitrov: it avoids exporting an empty type, keeps the package surface smaller and easier to use

pkg/blockbuilder/blockbuilder.go
dimitarvdimitrov: yeah for sure another pr is better

pkg/storage/ingest/reader.go
narqo: As we discussed in a DM: rather than adding a new `ConsumeKgo` method, I'd rather refactored the `Consume` to take an `iter.Seq[*kgo.Record]`. I'll do that in a separate PR, though

**Update** see https://github.com/grafana/mimir/pull/12482

PR #12448, opened 2025-08-20
Skip load shedding when an ingester is the only available replica

jhalterman:
#### What this PR does

This change propagates some gRPC meta from queriers to ingesters indicating when the ingester is the only replica being called. A server middleware passes along this info in a context, which the ingester can use to skip the reactive limiter and circuit breaker.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/ingester/ingester.go
```diff
@@ -1757,24 +1757,31 @@ func (i *Ingester) StartReadRequest(ctx context.Context) (resultCtx context.Cont
...
-	if err != nil {
-		return nil, err
+	// When this is not the only replica handling a request, limit reads for overloads
+	if !client.IsOnlyReplicaContext(ctx) {
+		// Check that a permit can be later acquired
+		if i.reactiveLimiter.read != nil && !i.reactiveLimiter.read.CanAcquirePermit() {
+			i.metrics.rejected.WithLabelValues(reasonIngesterMaxInflightReadRequests).Inc()
+			return nil, newReactiveLimiterExceededError(reactivelimiter.ErrExceeded)
+		}
+		finish, err := i.circuitBreaker.tryAcquireReadPermit()
```
dimitarvdimitrov, line 1772: do we still want to acquire the CB permit even if we don't use the reactive limiter? I think it will probably need to be something like `mustAcquireReadPermit` instead of `tryAcquireReadPermit` because we still want to run the request even if the CB is open.

But my point is that this request may still provide some info the CB so it opens/closes?

pkg/ingester/client/only_replica.go
```diff
@@ -0,0 +1,84 @@
...
+// context, in preparation for use by any tap handle.
+func OnlyReplicaTapHandle(ctx context.Context, _ *tap.Info) (context.Context, error) {
+	return ctxFromMetadata(ctx), nil
+}
+
+// OnlyReplicaClientStreamInterceptor propagates an only-replica marker from the context to the outgoing gRPC request
+// metadata.
+func OnlyReplicaClientStreamInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	if replicas, ok := ring.GetAvailableReplicas(ctx); replicas == 1 && ok {
+		ctx = metadata.AppendToOutgoingContext(ctx, onlyReplicaGrpcMdKey, "true")
```
dimitarvdimitrov, line 45: on that call way back I mentioned generalizing this to a quality of service level, so that later we can apply the same to different queries (Marco suggested alert evaluations) . WDYT about that?

jhalterman: We should definitely do that - I think there's a separate issue about this - but I think propagating request priority information is separate from propagating this "only replica" information.

pkg/ingester/ingester.go
jhalterman: I wondered about this too: should we record execution results, successes or failures, even when we're essentially disregarding the circuit breaker's current state and forcefully acquiring a permit? I don't think we should - since this could cause the breaker to get into a weird situation where it's recording lots of failures, but isn't doing anything about them since we're forcefully acquiring. This would have a surprising effect when the breaker start's being used normally again. I think it's simpler and reasonable enough to just not use the breaker when we're the only replica.

dimitarvdimitrov: ok that makes sense. The example that I was thinking of is when the breaker is open but the forced request succeeds so maybe the breaker can be closed quicker. But I guess that's a nice to have and it has the weird implications you mentioned. So the current approach sounds good.

pkg/ingester/client/only_replica.go
dimitarvdimitrov: I thought that number of replicas are priority are interchangeable right now. single replica <=> high priority and multiple replicas <=> low priority. We don't have the have the whole chain of priority passing for us to pass priority here. 

not insisting on it because it's an easy code change. But I also don't see a reason not to model it that way from the start

jhalterman: > I thought that number of replicas are priority are interchangeable right now. single replica <=> high priority and multiple replicas <=> low priority

I don't think that's the case. Even with multiple replicas, an individual replica might be available but overloaded and might want to shed lower priority requests. So I think the priority would be based on where the request came from, ex: ruler requests could be high priority. And this would be separate from the number of replicas.

dimitarvdimitrov: > So I think the priority would be based on where the request came from, ex: ruler requests could be high priority. And this would be separate from the number of replicas.


💯 agree. That just refines the definition of priority. Today the "priority" is chosen very crudely - based on the number of instances in the ring (replica set)

pkg/ingester/client/only_replica_test.go
```diff
@@ -0,0 +1,97 @@
...
+		ctx := ContextWithOnlyReplica(context.Background())
+		require.True(t, IsOnlyReplicaContext(ctx))
+	})
+
+	t.Run("when is only replica does not exist in a context", func(t *testing.T) {
+		require.False(t, IsOnlyReplicaContext(context.Background()))
+	})
+}
+
+type serverStreamMock struct {
```
dimitarvdimitrov: can't we use the `ctxStream` from the prod code instead?

tacole02 (approved):
Changelog looks good!

PR #12449, opened 2025-08-20
MQE: separate query planner for querier and query-frontends

charleskorn:
#### What this PR does

This PR introduces a separate query planner for queriers and another for query-frontends.

Remote execution of query plans is implemented as an optimisation pass that identifies subtrees of the query plan that should be executed remotely. This optimisation pass must only run in query-frontends: queriers should not remotely execute parts of the query plan themselves.

Therefore, in order to behave correctly when running in monolithic or read/write mode, we need to have separate query planners for queriers and query-frontends.

I have not added a changelog entry given this is a refactoring.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/mimir/mimir_test.go
+++ b/pkg/mimir/mimir_test.go
@@ -45,6 +45,7 @@ import (
 	"github.com/grafana/mimir/pkg/compactor"
 	"github.com/grafana/mimir/pkg/distributor"
 	"github.com/grafana/mimir/pkg/frontend"
+	v2 "github.com/grafana/mimir/pkg/frontend/v2"
 	"github.com/grafana/mimir/pkg/ingester"
 	"github.com/grafana/mimir/pkg/querier"
 	"github.com/grafana/mimir/pkg/ruler"
@@ -168,6 +169,9 @@ func TestMimir(t *testing.T) {
 		},
 		Frontend: frontend.CombinedFrontendConfig{
 			QueryEngine: "prometheus",
+			FrontendV2: v2.Config{
+				SchedulerAddress: "localhost",
+			},
 		},
 		MemberlistKV: memberlist.KVConfig{
 			WatchPrefixBufferSize: 128,
--- a/pkg/mimir/modules.go
+++ b/pkg/mimir/modules.go
@@ -919,8 +919,8 @@ func (t *Mimir) initQuerierQueryPlanner() (services.Service, error) {
 	_, mqeOpts := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, t.ActivityTracker, util_log.Logger, reg)
 	t.QuerierQueryPlanner = streamingpromql.NewQueryPlanner(mqeOpts)
 
-	// If the query-frontend is running in this process, we want to expose its planner through the analysis endpoint, not
-	// the querier's planner.
+	// Only expose the querier's planner through the analysis endpoint if the query-frontend isn't running in this process.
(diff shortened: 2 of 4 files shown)


pkg/mimir/modules.go
```diff
@@ -860,7 +861,7 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
 	// If the query-frontend is running in the same process as the query-scheduler and
 	// the query-scheduler hasn't been explicitly configured, Mimir will default to trying
 	// to connect to a scheduler on localhost on its own gRPC listening port.
-	if t.Cfg.isAnyModuleEnabled(QueryScheduler, Read, All) && !t.Cfg.Frontend.FrontendV2.IsSchedulerConfigured() {
+	if t.Cfg.isQuerySchedulerEnabled() && !t.Cfg.Frontend.FrontendV2.IsSchedulerConfigured() {
```
charleskorn, line 868: Note to reviewers: this logic seemed incorrect to me: the old version was checking if the `read` target was enabled, but the query-scheduler runs in the `backend` target in read/write mode. There was further discussion of this in https://github.com/grafana/mimir/pull/12200#discussion_r2248882311.

This is also the reason for the [change](https://github.com/grafana/mimir/pull/12449/commits/aae7f07edec87661d937992d0d276e9d23341d36) to `TestMimir` in `mimir_test.go`.

pkg/mimir/mimir_test.go
```diff
@@ -168,6 +169,9 @@ func TestMimir(t *testing.T) {
 		},
 		Frontend: frontend.CombinedFrontendConfig{
 			QueryEngine: "prometheus",
+			FrontendV2: v2.Config{
+				SchedulerAddress: "localhost",
+			},
```
charleskorn, line 174: See [comment below](https://github.com/grafana/mimir/pull/12449/files#r2287010968) for explanation of this - previously this test passed due to the incorrect behaviour explained there.

pkg/mimir/modules.go
```diff
@@ -913,19 +914,35 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
...
-	t.QueryPlanner = streamingpromql.NewQueryPlanner(mqeOpts)
+func (t *Mimir) initQuerierQueryPlanner() (services.Service, error) {
+	reg := prometheus.WrapRegistererWith(prometheus.Labels{"component": "querier"}, t.Registerer)
+	_, mqeOpts := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, t.ActivityTracker, util_log.Logger, reg)
+	t.QuerierQueryPlanner = streamingpromql.NewQueryPlanner(mqeOpts)
 
-	analysisHandler := streamingpromql.AnalysisHandler(t.QueryPlanner)
+	// If the query-frontend is running in this process, we want to expose its planner through the analysis endpoint, not
+	// the querier's planner.
+	if !t.Cfg.isQueryFrontendEnabled() {
```
56quarters, line 924: The conditional in the comment is technically correct but the opposite of what the code is doing. Maybe "Only expose the querier's planner when the query-frontend is not running in this process"

56quarters (approved):
LGTM

pkg/mimir/modules.go
charleskorn: Good point, addressed in 8b8a42d

PR #12453, opened 2025-08-20
Docs: overview of ingest storage architecture

tacole02:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.


docs/sources/mimir/get-started/about-grafana-mimir-architecture/index.md
```diff
@@ -11,16 +11,43 @@ weight: 5
...
+{{< admonition type="note" >}}
+Classic architecture is still supported in Grafana Mimir version 3.0. However, this architecture is set to be deprecated in a future release. As a best practice, use ingest storage architecture when setting up a new deployment of Grafana Mimir.{{< /admonition >}}
+
+### How ingest storage architecture works
+
+Ingest storage architecture fundamentally changes how Grafana Mimir handles data. In classic architecture, ingester nodes are highly stateful. They combine in-memory data with local write-ahead logs (WALs) and participate in both writes and reads. This can lead to heavy queries disrupting live writes.
+
+Ingest storage architecture mitigates this issue through decoupling the read and write paths and significantly reducing the statefulness of ingesters. It instead uses object storage as a central data transfer pipeline between the write and read paths.
+
+Here's an overview of how ingest storage architecture works (WIP):
```
tacole02: @dimitarvdimitrov could you please help provide some quick bullet points about how the process works? I'm not sure we want to get into our use of Warpstream here, since that's just our specific implementation of it. What do you think?

tacole02:
Hi @dimitarvdimitrov , do you think you could help with the updates needed for the "write path" section? Thanks!

docs/sources/mimir/get-started/about-grafana-mimir-architecture/index.md
```diff
@@ -11,16 +11,43 @@ weight: 5
...
+Ingest storage architecture is designed to enhance reliability, support growth, and enable use cases at an even larger scale. It enables you to run more robust and cost-effective Mimir deployments.
+
+{{< admonition type="note" >}}
+Classic architecture is still supported in Grafana Mimir version 3.0. However, this architecture is set to be deprecated in a future release. As a best practice, use ingest storage architecture when setting up a new deployment of Grafana Mimir.{{< /admonition >}}
+
+### How ingest storage architecture works
+
+Compared with classic architecture, ingest storage architecture fundamentally changes how Grafana Mimir handles data. In classic architecture, ingester nodes are highly stateful. They combine in-memory data with local write-ahead logs (WALs) and participate in both writes and reads. This can lead to heavy queries disrupting live writes.
+
+Ingest storage architecture mitigates this issue through decoupling the read and write paths and significantly reducing the statefulness of ingesters. It instead uses object storage as a central data transfer pipeline between the write and read paths.
```
dimitarvdimitrov:
```suggestion
Ingest storage architecture mitigates this issue through decoupling the read and write paths and significantly reducing the statefulness of ingesters. It instead uses Kafka as a central data transfer pipeline between the write and read paths.
```

warpstream chooses to implement the kakfa protocol on object store, but oss kafka runs with good old disks. In practice mimir lets the kafka implementation deal with the storage medium.

docs/sources/mimir/get-started/about-grafana-mimir-architecture/index.md
dimitarvdimitrov: so here's how the flow looks like 

- distributors are the point of entry for write requests. the is the same as classic architecture. distributor's responsibilities do not change significantly - they're still responsible for validating requests, applying transformations if any and forwarding the samples to the storage layer which will persiste them
- distributors are connected to a single kafka topic; they shard each write request to multiple partitions in the same kafka topic. Sharding to partitions can also use shuffle sharding.
- The distributors briefly buffers sharded write requests to the same partition before sending them to the kafka broker who is the leader for the partition. 
- The kafka broker persists the write requests (aka "records" if we use kafka terminology) durably and optionally replicates it to other brokers before acknowledging the write (optionally if the kafka cluster has been configured with replication). 
- The distributor acknowledges the write requests.

_the rest bleeds into the read path_

- ingesters continuous consume from kafka
- each ingester consumes from a single partition. As a result there need to be at least as many partitions as ingesters. Extra partitions will not be used until there is an ingester with the same ID as the partition. ingesters know what partition to consume from by looking at their hostname or optionally the `ingester.ring.instance-id` configuration flag)
- to provide high availability on the read path there can be multiple zones and sibling ingesters of each zone will consume the same partition (for example `ingester-zone-a-1` and `ingester-zone-b-1` both consume form partition 1)
- after fetching write requests (aka records) from the kafka partition each ingester adds them to their in-memory state and persists them to disk
- from this point onwards the samples will be included in query results





I don't know if you want to include all of this detail here. Maybe some of it can be left out

dimitarvdimitrov (commented):
hey i pushed my suggested for the read/write path and kafka to 5f0b0c276ca8d8ebc71cea7b5cdc835aabbe60a9, check them out. it's still a WIP, but it's moving in some direction :smile:

docs/sources/mimir/get-started/about-grafana-mimir-architecture/index.md
tacole02: Thank so much, @dimitarvdimitrov! I think I got all that summarized. Please let me know how it looks!

tacole02:
Hi @dimitarvdimitrov ! What you have for write/read path updates looks great so far! Let me know when it's ready for review 😄

dimitarvdimitrov:
https://github.com/grafana/mimir/pull/12515 thiis is what I arrived at after another iteration. Can you review it @tacole02 ? i expect it will need a lot of touching-up but it's a starting point 😅 

@56quarters can you take a look at this and at https://github.com/grafana/mimir/pull/12515 when you have some time? I'm looking for feedback on the level of detail and the completeness of the docs.

dimitarvdimitrov (approved):
SGTM, did you mean to commit the .cursor and .docs files? we already have some CLAUDE.md files and i wonder if we can't unify the agent config somehow. Do you want to figure that out in a separate PR instead of blocking this one?

Issue #12177, opened 2025-07-23
Create design document for corresponding helm chart for Mimir 3.0


56quarters:
Mimir 3.0 introduces the ingest storage architecture but also retains classic architecture. We need to create a design document for supporting these in the helm chart that weighs the pros and cons of each option and decide which one we will pursue.

Options as I understand them:

*  Don't make any changes to the helm chart for ingest storage. Keep classic architecture as the default and only supported way to run Mimir.
* Change the chart to make ingest storage the only supported way to run Mimir.
* Add support for both ingest storage _and_ classic architecture in the helm chart.

Any other changes to the helm chart are out of scope for this issue.

There is an existing [issue for the next major helm chart release](https://github.com/grafana/mimir/issues/4560).

narqo:
Recoding here what we've agreed internally.

For the next major release of the Helm chart

1. We will make ingest storage the default
2. We will provide a "classic mode" preset (a set of chart values), that will run Mimir in the "classic" (pre-Kafka) architecture
3. We will embed a demo Kafka cluster (enabled by default), with the same "don't use it in production" notice as Minio

With that, new users will have an option to quick start a full self-contained Mimir cluster.

For **production use cases**, we will expect users to disable our embedded Kafka, and to feed the chart with credentials of an external Kafka-compatible backend, that the users manage themselves. This is similar to how users are expected to disable Minio, and feed the chart with credentials of an object storage bucket, managed by the users.

PR #12459, opened 2025-08-20
helm: embed demo kafka cluster for ingest storage

narqo:
#### What this PR does

The PR adds a demo Kafka cluster (disable by default, **for now**), that will be used when the chart will deploy Mimir in the ingest storage architecture.

This is the first step for what we've agreed in https://github.com/grafana/mimir/issues/12177

> We will embed a demo Kafka cluster (enabled by default), with the same "don't use it in production" notice as Minio

The PR also adds a new test file `test-ingest-storage-values.yaml`, that configures Mimir to run in the ingest storage mode, and use the embedded Kafka.

I've tested it in a local k8s cluster:

```
helm upgrade mimir . --install --create-namespace --namespace test-mimir -f ci/offline/test-ingest-storage-values.yaml
```

The above created a new Mimir cluster, and connected the write- and read-components through the embedded single-node Kafka cluster. Both read and writes to the Mimir cluster worked as I expected.

Relates to https://github.com/grafana/mimir/issues/12177  https://github.com/grafana/mimir/issues/4560

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


monoxane (approved):
Chart updates correctly on top of existing deployment and does not touch resources with Kafka disabled by default.

Enabling Kafka and updating the structured configuration to use the ingest storage write path works as expected and I am now successfully ingesting and querying metrics in my cluster.

Once the ingest storage architecture is properly implemented in helm it might be nice to use [Kafka's proposed first-party chart](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1149%3A+Helm+Chart+for+Apache+Kafka) instead of our own templates so we can offload the maintenance work to upstream.

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+          imagePullPolicy: {{ .image.pullPolicy }}
+          ports:
+            - containerPort: 9092
+              name: kafka
+              protocol: TCP
+            - containerPort: 9093
+              name: controller
+              protocol: TCP
+          env:
+            # Kubernetes expands env variables in lexical order. So "_POD_NAME" (sic) here is so the name was expanded before KAFKA_ADVERTISED_LISTENERS is.
```
dimitarvdimitrov, line 64: what does sic mean? 🤔

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+            - name: KAFKA_NODE_ID
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.labels['apps.kubernetes.io/pod-index']
+            - name: KAFKA_PROCESS_ROLES
+              value: "broker,controller"
+            - name: KAFKA_LISTENERS
+              value: "PLAINTEXT://0.0.0.0:9092,CONTROLLER://0.0.0.0:9093"
+            - name: KAFKA_ADVERTISED_LISTENERS
+              value: PLAINTEXT://$(_POD_NAME).{{ include "mimir.resourceName" (dict "ctx" $ "component" "kafka") }}-headless.{{ $.Release.Namespace }}.svc:9092
```
dimitarvdimitrov: you can use `.values.global.clusterDomain` here for an fqdn

dimitarvdimitrov: and if you do that can we drop the `KAFKA_CONTROLLER_QUORUM_VOTERS`

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+            failureThreshold: 3
+      volumes:
+        - name: kafka-config
+          emptyDir: {}
+        - name: tmp
+          emptyDir: {}
+        {{- with .extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+  {{- if .persistence.enabled }}
```
dimitarvdimitrov, line 140: any reason not to have disks? I mean will kafka even work if there's nowhere to write segments?

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+            - name: KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR
+              value: "{{ min 3 $replicas }}"
+            - name: KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR
+              value: "{{ min 3 $replicas }}"
+            - name: KAFKA_TRANSACTION_STATE_LOG_MIN_ISR
+              value: "{{ min 2 $replicas }}"
+            - name: KAFKA_NUM_PARTITIONS
+              value: "100"
+            - name: KAFKA_AUTO_CREATE_TOPICS_ENABLE
+              value: "true"
```
dimitarvdimitrov: shouldn't we disable this? this way it's configured in a few places and it's may be confusing which actually takes precedence. 

the mimir one takes precedence since we do explicit metadata requests to check the topic and create it

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+              value: "{{ range $i := until (int $replicas) }}{{ $i }}@{{ include "mimir.resourceName" (dict "ctx" $ "component" "kafka") }}-{{ $i }}.{{ include "mimir.resourceName" (dict "ctx" $ "component" "kafka") }}-headless.{{ $.Release.Namespace }}.svc.{{ trimSuffix "." $.Values.global.clusterDomain }}:9093{{ if lt $i (sub (int $replicas) 1) }},{{ end }}{{ end }}"
+            - name: KAFKA_CONTROLLER_LISTENER_NAMES
+              value: "CONTROLLER"
+            - name: KAFKA_INTER_BROKER_LISTENER_NAME
+              value: "PLAINTEXT"
+            - name: KAFKA_LOG_DIRS
+              value: "/var/lib/kafka/data"
+              # The "REPLICATION_FACTOR" must be defined so the broker created offsets topic. Otherwise, the Mimir ingesters won't be able to start.
+            - name: KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR
+              value: "{{ min 3 $replicas }}"
```
dimitarvdimitrov, line 92: isn't 2 more common? it's enough for higher availability

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -3047,6 +3047,77 @@ minio:
...
+    # -- Service type
+    type: ClusterIP
+    # -- Service annotations
+    annotations: {}
+    # -- Service labels
+    labels: {}
+  # -- Kafka persistence configuration
+  persistence:
+    enabled: true
+    size: 5Gi
```
dimitarvdimitrov, line 3088: is the kafka running without any retention settings? can we set the retention to like 24h? I'm also not sure if 5GB won't fill up quickly for almost all use cases. can we use something like 10GB - 5x the ingester disk size (this is a gut feeling, sooo if you disagree, we can stick with 5GB)

dimitarvdimitrov: (i'm also inclined to make the retention configurable)

dimitarvdimitrov (commented):
a few comments, but this looks simpler than what i expected it would be.

what's missing is configuring mimir to use ingest storage. Did we agree to have kafka as the default basically?

operations/helm/charts/mimir-distributed/values.yaml
narqo: I don't have any data at hands. But I wonder, how is this important, comparing to default 5GB, that the Helm chart uses for Minio, and 2GB that is set for ingesters and compactors?

I don't mind lowering the default retention for topic, though.

**Update**: set the topic retention to 1d in https://github.com/grafana/mimir/pull/12459/commits/5f0b75b09c4b3ad8102917578ae8ff3f878b3f5e

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: I don't understand this one, sorry. 

* with full FQDN in `advertised.listeners` the ingesters fail to connect to the broker (I didn't understand why; it might be that it didn't relate to DNS, but I already spend a whole day battling with that 😵‍💫)
* `controller.quorum.voters` is what Kafka controller requests in the KRaft mode ([docs][1]).

[1]: https://docs.confluent.io/platform/current/kafka-metadata/config-kraft.html

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: > what does sic mean? 🤔

It means that `_POD_NAME` (with the underscore in front) is not a typo, but indeed how it's expected here. I.e. if you rename it to `POD_NAME` the `KAFKA_ADVERTISED_LISTENERS` won't be resolved to a pod name, and Kafka will fail to start with 

```
Exception in thread "main" java.lang.IllegalArgumentException: Error creating broker listeners from
  'PLAINTEXT://$(POD_NAME).mimir-kafka-headless.test-mimir.svc.cluster.local.:9092': Unable to parse PLAINTEXT://$(POD_NAME).mimir-kafka-headless.test-mimir.svc.cluster.local.:9092
```

narqo:
> what's missing is configuring mimir to use ingest storage. Did we agree to have kafka as the default basically?

I will do that in a separate PR, yes

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: The [default value is 3](https://kafka.apache.org/documentation/#brokerconfigs_offsets.topic.replication.factor), and this assumes the default `min.insync.replicas=2`. I didn't proofread this, but that's what all examples and docs are using, so that's what claude generated.

operations/helm/charts/mimir-distributed/ci/offline/test-ingest-storage-values.yaml
```diff
@@ -0,0 +1,28 @@
...
+kafka:
+  enabled: true
+
+mimir:
+  structuredConfig:
+    ingest_storage:
+      enabled: true
+      kafka:
+        # Address of Kafka broker to bootstrap the connection.
+        address: '{{ template "mimir.fullname" . }}-kafka.{{ .Release.Namespace }}.svc:9092'
```
56quarters: nit: I know we're inconsistent with this but can we use the full hostname here: `mimir-kafka.namespace.svc.cluster.local.` ?

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+kind: StatefulSet
+metadata:
+  name: {{ include "mimir.resourceName" (dict "ctx" $ "component" "kafka") }}
+  labels:
+    {{- include "mimir.labels" (dict "ctx" $ "component" "kafka") | nindent 4 }}
+  annotations:
+    {{- toYaml .podAnnotations | nindent 4 }}
+  namespace: {{ $.Release.Namespace | quote }}
+spec:
+  podManagementPolicy: Parallel
```
56quarters: Do we actually want this to be parallel? Since we don't have multiple zones it seems like we'd want one pod being updated at a time.

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+              mountPath: /var/lib/kafka
+            {{- end }}
+            - name: kafka-config
+              mountPath: /opt/kafka/config
+            - name: tmp
+              mountPath: /tmp
+            {{- with .extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          livenessProbe:
```
56quarters: `livenessProbe` is a niche use case. I don't think we should set it here unless there are Kafka docs that recommend it.

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: Done in https://github.com/grafana/mimir/pull/12459/commits/5f0b75b09c4b3ad8102917578ae8ff3f878b3f5e

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: > you can use .values.global.clusterDomain here for an fqdn

It works now. I don't understand why it didn't yesterday. Done in https://github.com/grafana/mimir/pull/12459/commits/fb2b6ee0532ace62637ac20304b5534373fd3949

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -3047,6 +3047,77 @@ minio:
...
+    storageClassName: ""
+  # -- Resource requests and limits for Kafka
+  resources:
+    requests:
+      cpu: 100m
```
56quarters: This CPU and memory seems too low for a Java process. I'd err on the side of giving it too many resources to avoid having people run into weird issues where it doesn't start. Maybe 2 CPUs and a few gigs of memory?

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: Good call. I missed this part, claude generated. This must've been fall back to an emptyDir. Fixed in https://github.com/grafana/mimir/pull/12459/commits/93355bf1b899d2bd63267500c3e935878ff09f98

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
```diff
@@ -0,0 +1,178 @@
+{{- if and .Values.kafka.enabled (not .Values.federation_frontend.disableOtherComponents) -}}
+{{- with .Values.kafka -}}
+{{- $replicas := .replicas | default 1 -}}
```
56quarters, line 3: Do we want to hard code the number of replicas as `1` and only support a single broker? I'm not sure if that would be a bad user experience since they'd be stuck with a very non-production setup but it seems like it would make the setup with Kafka simpler.

operations/helm/charts/mimir-distributed/templates/kafka/kafka-svc.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if and .Values.kafka.enabled (not .Values.federation_frontend.disableOtherComponents) -}}
```
56quarters, line 1: Do we actually want a non-headless service?

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -3047,6 +3047,80 @@ minio:
...
 
+kafka:
+  # -- Enable Kafka for ingest-storage architecture (demo purposes only)
+  enabled: false
+
+  image:
+    # -- The container image registry
+    registry: docker.io
+    # -- The container image repository
+    repository: apache/kafka-native
```
56quarters, line 3068: TIL, neat.

56quarters (commented):
This looks great overall, thanks for taking this on.

operations/helm/charts/mimir-distributed/values.yaml
narqo: I actually run a single-node java-based (non-native binary) Kafka in a homelab VM for a year. It peaks at 1GB memory 👍 

```
root@kafka-1:~# systemctl status kafka.service

   Main PID: 268 (java)
      Tasks: 94 (limit: 34661)
     Memory: 1.0G (peak: 1.1G)
        CPU: 1h 11min 23.069s
     CGroup: /system.slice/kafka.service
```

I've updated the resources in the default values to 1 CPU / 1Gi

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: I'd try to not overthink it and waited until real users came with their requests or questions.

operations/helm/charts/mimir-distributed/templates/kafka/kafka-svc.yaml
narqo: We use none-headless (cluster IP) service to discover the broker. It looks nicer in the config, than when one hardcodes the name of the first pod:

```diff
# mimir.yaml

ingest_storage:
  enabled: true
  kafka:
-    address: kafka-0.mimir-kafka-headless.mimir-demo.svc.cluster.local.:9092
+    address: mimir-kafka.mimir-demo.svc.cluster.local.:9092
```

narqo:
@dimitarvdimitrov @56quarters I think, I've addresses/answered all questions. PHAL

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
dimitarvdimitrov: til

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
dimitarvdimitrov: 🤷 thanks for trying it again

dimitarvdimitrov: and do we still need `KAFKA_CONTROLLER_QUORUM_VOTERS` then?

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
dimitarvdimitrov: can you also drop `KAFKA_NUM_PARTITIONS`?

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
dimitarvdimitrov: sgtm

operations/helm/charts/mimir-distributed/templates/kafka/kafka-statefulset.yaml
narqo: > do we still need KAFKA_CONTROLLER_QUORUM_VOTERS then?

Yes. Otherwise it dies on startup,

```
===> Configuring ...
Running in KRaft mode...
===> Launching ...
Exception in thread "main" org.apache.kafka.common.config.ConfigException: If using process.roles, either controller.quorum.bootstrap.servers must contain the set of bootstrap controllers or controller.quorum.voters must contain a parseable set of controllers.
```

PR #12457, opened 2025-08-20
lookupplan: Add ContextWithDisabledPlanning functionality

dimitarvdimitrov:
## Summary

Add the ability to disable cost-based planning using context values. When planning is disabled via `ContextWithDisabledPlanning()`, the `CostBasedPlanner` returns the input plan unchanged.

This allows callers to bypass the planning optimization when needed while maintaining compatibility with the existing planner interface.

related to https://github.com/grafana/mimir/issues/12085

Issue #11916, opened 2025-06-30
TSDB index lookup planning


dimitarvdimitrov:
This epic tracks the work for the index lookup planning project in Mimir.

dimitarvdimitrov:
some results after deploying this to clusters at GL. Most small&medium clusters (<100M series) don’t see a difference, but a few see massive changes
* mimir cluster A has ~45% reducing in ingester CPU
* mimir cluster B has 18-29% lower ingester CPU
* mimir cluster C has 20-44% lower CPU
* mimir cluster D has 36% lower average CPU (but the same peak CPU, so we won’t be able to drop usage here)
* mimir cluster E has 46% lower p99 latency

PR #12460, opened 2025-08-20
lookupplan: Mirror Select calls with and without planner

dimitarvdimitrov:
### What this PR does

Implements a mirroring chunk querier that validates index lookup planning correctness by comparing results with and without planning enabled.

This PR isn't perfect, but it should help us find bugs. I expect we'll get few false-positives and they'll be only for the HEAD block.

This doesn't have as much test coverage as ideal, but the idea is for us to delete this code once https://github.com/grafana/mimir/issues/11916 is done.

### How

The querier is injected at block creation time and controlled by `IndexLookupPlanningComparisonPortion` configuration to avoid significant overhead. It tries to filter out newly created series

Based on changes from PR #12457.


docs/sources/mimir/configure/about-versioning.md
```diff
@@ -212,6 +212,8 @@ The following features are currently experimental:
     - `-blocks-storage.tsdb.shared-postings-for-matchers-cache`
     - `-blocks-storage.tsdb.head-postings-for-matchers-cache-invalidation`
     - `-blocks-storage.tsdb.head-postings-for-matchers-cache-versions`
+  - Index lookup planning comparison to validate planning correctness
+    - `-blocks-storage.tsdb.index-lookup-planning-comparison-portion`
```
dimitarvdimitrov, line 217:
```suggestion
    - `-blocks-storage.tsdb.index-lookup-planning-enabled`
    - `-blocks-storage.tsdb.index-lookup-planning-comparison-portion`
```

tacole02 (approved):
Docs look good. Thanks, @dimitarvdimitrov !

pkg/ingester/ingester.go
```diff
@@ -2949,6 +2950,28 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
 	return userDB, nil
 }
 
+// createBlockChunkQuerier creates a BlockChunkQuerierFunc that optionally wraps the default querier with mirroring
```
chencs:
```suggestion
// createBlockChunkQuerier creates a BlockChunkQuerier that optionally wraps the default querier with mirroring
```
?

pkg/ingester/mirror_querier.go
```diff
@@ -0,0 +1,267 @@
...
+		return
+	}
+
+	// Create a fake secondaryLabels slice for the logging function
+	q.recordComparisonOutcome(extraSeries, missingSeries, primaryLabels, secondarySeriesCount)
+}
+
+func (q *mirroredChunkQuerier) recordComparisonOutcome(extraSeries []string, missingSeries []string, primaryLabels []labels.Labels, secondaryLabels int) {
+	switch {
+	case len(extraSeries) > 0 && len(missingSeries) > 0:
```
chencs, line 178: Can we add a `user` label to these? Even if we're only running mirrored queries on a proportion of queries, it could be useful for debugging if it turns out one user is consistently getting inconsistent results on non-mirrored queries as well.

pkg/ingester/mirror_querier_test.go
```diff
@@ -0,0 +1,353 @@
...
+			comparisonOutcomes := promauto.With(prometheus.NewPedanticRegistry()).NewCounterVec(
+				prometheus.CounterOpts{Name: "test_comparison_outcomes_" + tc.name},
+				[]string{"outcome", "user"},
+			)
+
+			querier := &mirroredChunkQuerier{
+				logger:             logger,
+				comparisonOutcomes: comparisonOutcomes,
+				delegate:           &mockChunkQuerier{},
+				userID:             "test-user",
```
chencs: I think this should be removed.

pkg/ingester/mirror_querier.go
```diff
@@ -0,0 +1,267 @@
...
+	return q.delegate.LabelValues(ctx, name, hints, matchers...)
+}
+
+func (q *mirroredChunkQuerier) LabelNames(ctx context.Context, hints *storage.LabelHints, matchers ...*labels.Matcher) ([]string, annotations.Annotations, error) {
+	level.Warn(q.logger).Log("msg", "LabelNames is not implemented in mirroredChunkQuerier, not comparing results")
+	return q.delegate.LabelNames(ctx, hints, matchers...)
+}
+
+func (q *mirroredChunkQuerier) Close() error {
+	if retunedErr := q.returnedSeries.Err(); errors.Is(retunedErr, context.Canceled) {
```
chencs:
```suggestion
	if returnedErr := q.returnedSeries.Err(); errors.Is(returnedErr, context.Canceled) {
```
?

pkg/ingester/mirror_querier.go
```diff
@@ -0,0 +1,267 @@
...
+		level.Warn(q.logger).Log("msg", "Select was canceled, skipping comparison")
+		q.comparisonOutcomes.WithLabelValues("context_cancelled").Inc()
+		return q.delegate.Close()
+	} else if retunedErr != nil {
```
chencs:
```suggestion
	} else if returnedErr != nil {
```
?

chencs:
Heads up, I ran into inconsistent ingester panics locally with `index_lookup_planning_comparison_portion: 1`, on `*retainingChunkSeriesSet` 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x34a2974]

goroutine 177 [running]:
github.com/grafana/mimir/pkg/ingester.(*retainingChunkSeriesSet).Err(0x0)
	/Users/casie/mimir/pkg/ingester/mirror_querier.go:258 +0x24
github.com/grafana/mimir/pkg/ingester.(*mirroredChunkQuerier).Close(0x40000268c0)
	/Users/casie/mimir/pkg/ingester/mirror_querier.go:65 +0x3c
github.com/prometheus/prometheus/tsdb.(*HeadAndOOOChunkQuerier).Close(0x4001f38b90)
	/Users/casie/mimir/vendor/github.com/prometheus/prometheus/tsdb/ooo_head_read.go:638 +0x6c
github.com/prometheus/prometheus/storage.(*mergeGenericQuerier).Close(0x4001eb84b0)
	/Users/casie/mimir/vendor/github.com/prometheus/prometheus/storage/merge.go:277 +0xb0
github.com/grafana/mimir/pkg/ingester.(*Ingester).executeStreamingQuery(0x40019b7508, {0x4b93e98, 0x400469fce0}, 0x40009da280, 0x198eda5302b, 0x198edb77500, {0x4004b20d40, 0x4, 0x4}, 0x4004fbd480, ...)
...
```

I did this by running Mimir locally in microservices mode, and just opening up the `Mimir / Reads` or `Mimir / Reads resources` dashboards over the last 1h and clicking around, viewing panels, etc (again, the panic isn't happening consistently). I dug around a little and didn't immediately see any issues with how that pointer gets set, but this does seem like a blocking issue that we should figure out (and maybe add a test for) before merging.

dimitarvdimitrov:
> Heads up, I ran into inconsistent ingester panics locally with `index_lookup_planning_comparison_portion: 1`

i also ran it today and saw the same. It's fixed by this commit https://github.com/grafana/mimir/pull/12460/commits/fe555daf5b46305ab2741904fa30a94894f41cda

I took the opportunity to test the histograms for planning duration and fixed another panic here https://github.com/grafana/mimir/pull/12460/commits/0aad1f2c15572f00e7676511ee438bdfb0beed14 and added a test in 3cae31d043

i think once we have this working end-to-end these kind of panics will surface very quickly

PR #12461, opened 2025-08-20
rw2: Do not silently drop metric metadata of type Unknown

alexweav:
#### What this PR does

Fixes a bug where we silently drop a metric metadata, if the type is `UNKNOWN`. Such metadata might still contain help/unit text that we want to store. This affects both the Remote Write 2.0 push API and ingest-storage V2 record format.

This happens because we normalize the metric name at the moment we parse the type. Since `UNKNOWN` is the default metric type, Proto often chooses to save a few bytes and never send the field at all. That means we also never parse the `type` field, thus we never normalize the name, thus the value of `normalizeMetricName` is always empty string. The parser now thinks the metric name is empty and drops the metadata as it maps to no series.

The fix simply normalizes the metric name a little later, outside of the `metricType` parsing branch.

Contrib https://github.com/grafana/mimir-squad/issues/2253

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/prealloc_rw2_test.go
+++ b/pkg/mimirpb/prealloc_rw2_test.go
@@ -365,6 +365,53 @@ func TestWriteRequestRW2Conversion(t *testing.T) {
 		require.Equal(t, expSymbols, rw2.SymbolsRW2)
 		require.Equal(t, expTimeseries, rw2.TimeseriesRW2)
 	})
+
+	t.Run("metadata with UNKNOWN type", func(t *testing.T) {
+		req := &WriteRequest{
+			Timeseries: nil,
+			Metadata: []*MetricMetadata{
+				{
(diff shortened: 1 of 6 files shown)


pkg/mimirpb/prealloc_rw2_test.go
```diff
@@ -365,6 +365,53 @@ func TestWriteRequestRW2Conversion(t *testing.T) {
 		require.Equal(t, expSymbols, rw2.SymbolsRW2)
 		require.Equal(t, expTimeseries, rw2.TimeseriesRW2)
 	})
+
+	t.Run("metadata with UNKNOWN type", func(t *testing.T) {
```
alexweav, line 369: This test is for posterity, it passes without the fix.

tcard (approved):

PR #12466, opened 2025-08-21
Parquet streaming poc

npazosmendez:


pkg/storage/parquet/queryable_shard.go
```diff
@@ -68,39 +80,66 @@ func (b queryableShard) Query(ctx context.Context, sorted bool, sp *prom_storage
...
+				// TODO: closeable
+				chunksIter, err := b.m.MaterializeChunks(ctx, rgi, mint, maxt, groupRR)
+				if err != nil {
+					return errors.Wrapf(err, "error materializing chunks for group %d", groupIdx)
+				}
+
+				// TODO: we are returning the chunks iterator without knowing if it has any values at all.
+				// And, more importantly, labels for the series. In Prometheus world that violates the querier's
+				// interface. For us, whether or not that is problematic depends on how Queriers handle responses
+				// from Store Gateways that have labels for which there are no series. We'll see...
```
francoposa, line 190: Is this still TODO? Haven't we observed behavior in dev?

pkg/storage/parquet/queryable_shard.go
```diff
@@ -68,39 +80,66 @@ func (b queryableShard) Query(ctx context.Context, sorted bool, sp *prom_storage
...
+
+			// MaterializeChunks is designed to materialize rows from top to bottom of the Parquet file. That
+			// is not necessarily ordered by labels set, as rows are ordered by a particular subset of labels.
+			// If we got them in that order, we'd have to load them all inmemory to then sort (because we need them sorted).
+			// Instead we leverage the fact that rows are sorted by label set _after_ sorted by that particular
+			// subset of columns (known as the 'sorting columns'). That means that if we were to materialize a set of rows
+			// with equal sorting column values, they would be sorted by label set.
+			// So the trick is: group our label sets in groups such that all series within a group have the same values
+			// for the sorting columns. We can then materialize each of the groups separately and then merge them sorted
+			// on the fly as we iterate, given each of them is sorted.
```
francoposa, line 169: We could probably push most of this detail into the `groupBySortingColumns`, but it is nice to have in general.

pkg/storage/parquet/queryable_shard.go
npazosmendez: We have observed dev "apparently working", but we have not done proper research about whether this could be problematic in certain (edge) cases. We are YOLO-ing it basically, which is fine for our POC I'd say.

PR #12470, opened 2025-08-22
 MQE: Move DeduplicateAndMerge to planning stage 

Konstantinov-Innokentii:
#### What this PR does
This is the first batch of work for the Skipping DeduplicateAndMerge operator when possible.
The PR introduces the DeduplicateAndMerge node and inserts it into the planning stage so it can later be pruned with the optimization pass. The optimization itself will be implemented in a next PR.

#### Implementation details

- A new DeduplicationAndMerge node type has been introduced. It simply wraps the inner node - [code](https://github.com/grafana/mimir/pull/12470/files#diff-a7515378fd6575121f3f2dd05dbdd8d0ad395afa1e8b19d9f8482dc5ba868670)
- During the planning stage, functions that drop name label, unary negations, binary operations, and vector-scalar operations are wrapped in DeduplicationAndMerge –  - [code](https://github.com/grafana/mimir/pull/12470/files#diff-8251d65134441870ca8a04b92479b72d640d81e53f86844fd87f18eb3aef5a13)
- Previously, the DeduplicationAndMerge operator was inserted in various places (factories, operator constructors, etc.). This has been cleaned up, and the operator is now created when the plan is materialized.

#### Checklist

- [x] Tests updated.
- [ ] ~Documentation added~.
- [ ] ~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.~


pkg/streamingpromql/operators/binops/vector_scalar_binary_operation.go
```diff
@@ -48,7 +48,7 @@ func NewVectorScalarBinaryOperation(
 	memoryConsumptionTracker *limiter.MemoryConsumptionTracker,
 	annotations *annotations.Annotations,
 	expressionPosition posrange.PositionRange,
-) (*VectorScalarBinaryOperation, error) {
+) (types.InstantVectorOperator, error) {
```
charleskorn: Why has this changed?

pkg/streamingpromql/planning/core/deduplicate_and_merge.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (d *DeduplicateAndMerge) EquivalentTo(other planning.Node) bool {
+	otherDedup, ok := other.(*DeduplicateAndMerge)
+
+	return ok && d.Inner.EquivalentTo(otherDedup.Inner)
+}
+
+func (d *DeduplicateAndMerge) Describe() string {
+	return "deduplicate and merge"
```
charleskorn: This could be an empty string, the name of the node is enough to explain what this node does.

(contrast this with an aggregation node for example, where it's useful to know if it's a sum or an average etc.)

pkg/streamingpromql/planning.go
```diff
@@ -440,6 +477,51 @@ func findFunction(name string) (functions.Function, bool) {
...
+		functions.FUNCTION_ROUND, functions.FUNCTION_PREDICT_LINEAR, functions.FUNCTION_TIMESTAMP,
+		functions.FUNCTION_DOUBLE_EXPONENTIAL_SMOOTHING,
+		// Histogram functions
+		functions.FUNCTION_HISTOGRAM_AVG,
+		functions.FUNCTION_HISTOGRAM_COUNT, functions.FUNCTION_HISTOGRAM_FRACTION,
+		functions.FUNCTION_HISTOGRAM_QUANTILE,
+		functions.FUNCTION_HISTOGRAM_STDDEV, functions.FUNCTION_HISTOGRAM_STDVAR,
+		functions.FUNCTION_HISTOGRAM_SUM,
+		// Label manipulation functions
+		functions.FUNCTION_LABEL_JOIN, functions.FUNCTION_LABEL_REPLACE:
```
charleskorn: [nit] I would format each of these on their own line, rather than having some on their own line and some with multiple on the same line

pkg/streamingpromql/planning.go
```diff
@@ -440,6 +477,51 @@ func findFunction(name string) (functions.Function, bool) {
 	return functions.FUNCTION_UNKNOWN, false
 }
 
+// functionNeedsDeduplication checks if a function needs deduplication and merge operator.
+// This is determined by whether the function drops the __name__ label or otherwise manupulates it.
+func functionNeedsDeduplication(fnc functions.Function) bool {
```
charleskorn, line 485: It'd be good to have some way to ensure we don't forget to add a newly-added function here.

Perhaps this `switch` should be expanded to cover all `functions.Function`s and should panic if `fnc` is not listed? Then we could have a unit test that checks all known functions don't cause `functionNeedsDeduplication` to panic. 

Or maybe use a map?

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -46,13 +46,7 @@ func SingleInputVectorFunctionOperatorFactory(name string, f FunctionOverInstant
 			return nil, fmt.Errorf("expected an instant vector argument for %s, got %T", name, args[0])
 		}
 
-		var o types.InstantVectorOperator = NewFunctionOverInstantVector(inner, nil, memoryConsumptionTracker, f, expressionPosition, timeRange)
-
-		if f.SeriesMetadataFunction.NeedsSeriesDeduplication {
```
charleskorn, line 51: We should be able to remove the `NeedsSeriesDeduplication` field now.

pkg/streamingpromql/planning.go
Konstantinov-Innokentii: Thanks, makes sense

pkg/streamingpromql/operators/binops/vector_scalar_binary_operation.go
Konstantinov-Innokentii: Thanks for noticing, it's a leftover from the initial implementation, I'll clean this up.

pkg/streamingpromql/operators/functions/common.go
```diff
@@ -137,10 +136,4 @@ type SeriesMetadataFunctionDefinition struct {
 	//
 	// If Func is nil, the input series are used as-is.
 	Func SeriesMetadataFunction
-
-	// NeedsSeriesDeduplication enables deduplication and merging of output series with the same labels.
-	//
-	// This should be set to true if Func modifies the input series labels in such a way that duplicates may be
-	// present in the output series labels (eg. dropping a label).
-	NeedsSeriesDeduplication bool
```
charleskorn, line 145: In the interests of keeping this PR small, this can be done as a follow-up PR, but it likely makes sense to drop the `SeriesMetadataFunctionDefinition` struct and just use `SeriesMetadataFunction` everywhere instead.

pkg/streamingpromql/planning.go
```diff
@@ -440,6 +477,106 @@ func findFunction(name string) (functions.Function, bool) {
 	return functions.FUNCTION_UNKNOWN, false
 }
 
+// functionNeedsDeduplication checks if a function needs deduplication and merge operator.
+// This is determined by whether the function drops the __name__ label or otherwise manipulates it.
+func functionNeedsDeduplication(fnc functions.Function) bool {
+	switch fnc {
+	// Functions that need deduplication (drop __name__ label)
```
charleskorn: [nit] In the case of `label_replace` or `label_join`, they don't necessarily drop the `__name__` label
```suggestion
	// Functions that need deduplication (manipulate labels)
```

pkg/streamingpromql/planning.go
```diff
@@ -440,6 +477,106 @@ func findFunction(name string) (functions.Function, bool) {
...
+		functions.FUNCTION_HISTOGRAM_QUANTILE,
+		functions.FUNCTION_HISTOGRAM_STDDEV,
+		functions.FUNCTION_HISTOGRAM_STDVAR,
+		functions.FUNCTION_HISTOGRAM_SUM,
+		// Label manipulation functions
+		functions.FUNCTION_LABEL_JOIN,
+		functions.FUNCTION_LABEL_REPLACE:
+		return true
+
+	// Functions that do NOT need deduplication (preserve __name__ label)
```
```suggestion
	// Functions that do NOT need deduplication (don't manipulate labels or are guaranteed to return only one series)
```

pkg/streamingpromql/planning.go
```diff
@@ -312,7 +312,27 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
 				ReturnBool:         expr.ReturnBool,
 				ExpressionPosition: core.PositionRangeFrom(expr.PositionRange()),
 			},
-		}, nil
+		}
+		if expr.Op == parser.LOR {
```
charleskorn, line 319: Might be worth adding a note here that only `or` and scalar/vector expressions needs deduplication - all other variations either can't produce duplicate series (eg. `and` and `unless`) or are handled by the operator (eg. vector/vector expressions).

pkg/streamingpromql/planning_test.go
```diff
@@ -1073,6 +1089,378 @@ func TestPlanCreationEncodingAndDecoding(t *testing.T) {
 				},
 			},
 		},
+		"unary negation - should deduplicate and merge": {
```
charleskorn: It might be clearer to test these cases with a string-formatted plan like in the optimisation pass tests - these tests here primarily exist to exercise the encoding and decoding of query plans.

pkg/streamingpromql/operators/functions/common.go
Konstantinov-Innokentii: OK, I'll do it in a follow-up, thanks for the idea

charleskorn (approved):

PR #12476, opened 2025-08-22
Add limit to Writes Resources > In-memory series panel

tcard:
#### What this PR does

Adds the `max_series` limit in the Writes Resources > Ingester > In-memory series panel.

Screenshot 2025-08-22 at 13 08 13

—

#### Checklist

- [ ] Tests updated.
- [x] Documentation added.


operations/mimir-mixin-tools/screenshots/go.mod
```diff
@@ -3,19 +3,18 @@ module screenshots-go
 go 1.23
 
 require (
-	github.com/chromedp/cdproto v0.0.0-20250803210736-d308e07a266d
+	github.com/chromedp/cdproto v0.0.0-20250221214400-5cdcd25b55f6
```
tcard, line 6: I had to downgrade this, d308e07a266d isn't compatible with `chromedp@v0.10.0`. I wonder how it was working before...

juliusmh:
Dashboard looks good, but the chromedp change bothers me.. We should not have to touch it at all for the mixin.

tcard:
@juliusmh It's not for the mixin exactly, it's for generating the screenshot that got updated to reflect the change in the mixin for the docs.

juliusmh (approved):
We could also upgrade chromedp instead of downgrading the dependency. Nevertheless, makes sense!

PR #12489, opened 2025-08-25
MQE: simplify `RangeVectorOperator` interface

charleskorn:
#### What this PR does

This PR removes the `StepCount` and `Range` methods from the `RangeVectorOperator` interface.

`StepCount` wasn't used anywhere, and `Range` was only used for a select few range vector functions. 

Despite now needing to compute the range on every step, removing `Range` seems to have had a neutral (or possibly even slightly positive) impact on latency in our benchmarks for range vector functions like `rate`.

I've opted not to add a changelog entry given this is an internal refactoring with no user-visible impact.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/operators/functions/common.go
+++ b/pkg/streamingpromql/operators/functions/common.go
@@ -80,7 +80,6 @@ func PassthroughData(seriesData types.InstantVectorSeriesData, _ []types.ScalarD
 //   - err error.
 type RangeVectorStepFunction func(
 	step *types.RangeVectorStepData,
-	rangeSeconds float64,
 	scalarArgsData []types.ScalarData,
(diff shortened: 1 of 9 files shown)


lamida (approved):
LGTM. 

Just curious how come this the latency impact is neutral/positive despite need to calculate in every step now.

charleskorn:
>Just curious how come this the latency impact is neutral/positive despite need to calculate in every step now.

I haven't looked into it in detail, but I'm guessing we save a bit of time from not needing to pass the range as an argument when calling `RangeVectorStepFunction`, which offsets the time spent doing the calculation (and I assume that's very cheap to compute).

PR #12492, opened 2025-08-25
MQE: introduce `Finalize` method to operators

charleskorn:
#### What this PR does

This PR introduces a new method to the `Operator` interface: `Finalize`.

`Finalize` is expected to be used to complete any outstanding work such as emitting annotations or updating query stats before the result is returned to the caller. The particular use case that inspired this is remote execution: we need to apply any annotations and query stats from the remote executor to the local query.

It differs from `Close` in that it is valid to call `Close` at any time, even if the query has failed or if the result has already been sent to the caller. In contrast, `Finalize` will only be called if the query has succeeded, and before results are sent to the caller.

I haven't added a changelog entry given this is an internal refactoring.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/evaluator.go
+++ b/pkg/streamingpromql/evaluator.go
@@ -121,7 +121,7 @@ func (e *Evaluator) Evaluate(ctx context.Context, observer EvaluationObserver) e
 	}
 
 	if e.engine.pedantic {
-		// Finalize the root operator a second time to ensure all operators behave correctly if Close is called multiple times.
+		// Finalize the root operator a second time to ensure all operators behave correctly if Finalize is called multiple times.
 		if err := e.root.Finalize(ctx); err != nil {
 			return fmt.Errorf("pedantic mode: failed to finalize operator a second time after successfully finalizing the first time: %w", err)
 		}
--- a/pkg/streamingpromql/evaluator.go
+++ b/pkg/streamingpromql/evaluator.go
@@ -87,8 +87,7 @@ func (e *Evaluator) Evaluate(ctx context.Context, observer EvaluationObserver) e
 
 	defer e.engine.activeQueryTracker.Delete(queryID)
 
-	err = e.root.Prepare(ctx, &types.PrepareParams{QueryStats: e.stats})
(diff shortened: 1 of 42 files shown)


pkg/streamingpromql/operators/aggregations/quantile.go
```diff
@@ -86,14 +86,21 @@ func (q *QuantileAggregation) NextSeries(ctx context.Context) (types.InstantVect
...
 
 	return q.Param.Prepare(ctx, params)
 }
 
+func (q *QuantileAggregation) Finalize(ctx context.Context) error {
+	if err := q.Aggregation.Finalize(ctx); err != nil {
+		return err
+	}
+
+	return q.Param.Finalize(ctx)
```
lamida, line 101: Is it possible param to be nil? If yes, we should have nil check here.

pkg/streamingpromql/evaluator.go
```diff
@@ -117,6 +116,17 @@ func (e *Evaluator) Evaluate(ctx context.Context, observer EvaluationObserver) e
 		return fmt.Errorf("operator type %T produces unknown result type", root)
 	}
 
+	if err := e.root.Finalize(ctx); err != nil {
+		return err
+	}
+
+	if e.engine.pedantic {
+		// Finalize the root operator a second time to ensure all operators behave correctly if Close is called multiple times.
```
lamida:
```suggestion
		// Finalize the root operator a second time to ensure all operators behave correctly if Finalize is called multiple times.
```
Do you mean this?

pkg/streamingpromql/operators/aggregations/quantile.go
charleskorn: No, it's not valid for `quantile` to have no parameter.

pkg/streamingpromql/evaluator.go
charleskorn: Yes I do, good catch

lamida (approved):

PR #12498, opened 2025-08-25
helm: replace kubectl image with alpine equivalent

narqo:
#### What this PR does

Starting 28/08/2025 Bitnami deprecates their free images. In order to mitigate any interruptions we update the default values, to use alternative Docker images.

In this PR we update the Docker image with kubectl to `alpine/kubectl`. This image comes with a shell, which is required by the provisioning job, that uses the image (i.e. that is why we don't use `registry.k8s.io/kubectl`).

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


operations/helm/charts/mimir-distributed/templates/provisioner/provisioner-job.yaml
```diff
@@ -80,14 +80,14 @@ spec:
           image: {{ include "mimir.kubectlImage" . }}
           imagePullPolicy: {{ .Values.kubectlImage.pullPolicy | default "IfNotPresent" }}
           command:
-            - /bin/bash
+            - /bin/sh
```
narqo, line 83: Note for the reviewers: the `alpine/kubectl` doesn't have bash. But, from what I can tell, the default `sh` should work fine for what this job does.

tacole02 (approved):
Changelog looks good!

dimitarvdimitrov (approved):

PR #12499, opened 2025-08-25
ingest-storage: Disable normalization of metric names in metadata in record format V2

alexweav:
#### What this PR does

The RW2 deserialization contains some app logic - it mutates metric names in metadata for certain types. For example, given a series that's part of a classic histogram called `my_histogram_seconds_bucket`, the unmarshaller will drop the `_bucket` suffix. This also means it dedupes metadata for corresponding `_count` and `_sum` series.

For ingest storage, it's more predictable to have the record format be as transparent as possible and not edit the data in unexpected ways. That way, a RW1 request is still guaranteed to retain its original behavior regardless of record version.

This PR adds an internal option to skip the normalization, and enables it for the ingest storage record path.

I don't think the deduping is necessarily bad in general; removes elements of the request, and all members of a histogram tend to have the same metadata. But, if we decide to normalize for V1, we should do it in the distributor, prior to sending the data through warpstream, not in the ingester unmarshal. That way it's not implicit, and it has a chance of shrinking the record size.

Requests sent from the RW2 ingestion API will already be deduped in the distributor and this does not affect them.
No changelog until we mark the V2 format as stable, as usual.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/mimirpb/mimir.pb.go.expdiff
+++ b/pkg/mimirpb/mimir.pb.go.expdiff
@@ -1,5 +1,5 @@
 diff --git a/pkg/mimirpb/mimir.pb.go b/pkg/mimirpb/mimir.pb.go
-index 98742def9a..24ebc65858 100644
+index dd58c2f3fb..24ebc65858 100644
 --- a/pkg/mimirpb/mimir.pb.go
 +++ b/pkg/mimirpb/mimir.pb.go
 @@ -14,7 +14,6 @@ import (
@@ -27,8 +27,8 @@ index 98742def9a..24ebc65858 100644
 -
 -	// Skip unmarshaling of exemplars.
 -	skipUnmarshalingExemplars bool
--	// Skip normalization of metadata metric names
--	skipNormalizeMetricName bool
+-	// Skip normalization of metadata metric names when unmarshalling the request.
+-	skipNormalizeMetadataMetricName bool
 -	// Unmarshal from Remote Write 2.0. if rw2symbols is not nil.
 -	unmarshalFromRW2 bool
 -	rw2symbols       rw2PagedSymbols
@@ -144,7 +144,7 @@ index 98742def9a..24ebc65858 100644
  			}
  			m.Timeseries = append(m.Timeseries, PreallocTimeseries{})
 -			m.Timeseries[len(m.Timeseries)-1].skipUnmarshalingExemplars = m.skipUnmarshalingExemplars
--			if err := m.Timeseries[len(m.Timeseries)-1].Unmarshal(dAtA[iNdEx:postIndex], nil, nil, m.skipNormalizeMetricName); err != nil {
+-			if err := m.Timeseries[len(m.Timeseries)-1].Unmarshal(dAtA[iNdEx:postIndex], nil, nil, m.skipNormalizeMetadataMetricName); err != nil {
 +			if err := m.Timeseries[len(m.Timeseries)-1].Unmarshal(dAtA[iNdEx:postIndex]); err != nil {
  				return err
  			}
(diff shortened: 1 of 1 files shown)


pkg/mimirpb/timeseries.go
```diff
@@ -271,15 +275,15 @@ var TimeseriesUnmarshalCachingEnabled = true
 //   - in case 3 the exemplars don't get unmarshaled if
 //     p.skipUnmarshalingExemplars is false,
 //   - is symbols is not nil, we unmarshal from Remote Write 2.0 format.
-func (p *PreallocTimeseries) Unmarshal(dAtA []byte, symbols *rw2PagedSymbols, metadata map[string]*orderAwareMetricMetadata) error {
+func (p *PreallocTimeseries) Unmarshal(dAtA []byte, symbols *rw2PagedSymbols, metadata map[string]*orderAwareMetricMetadata, skipNormalizeMetricName bool) error {
```
alexweav, line 278: I don't love the bool arg, but it's less dangerous than storing the bool on the Timeseries, which gets pooled, and re-used, and risks storing the option somewhere we don't expect. It's safer to keep it only on the WriteRequest.

A better approach might be to wrap every single option (there are now 5) into an `UnmarshalOptions` struct. It might reduce how often we have to touch the generated file. Doing this would significantly increase the diff, so I'd rather not do it in this PR, but I could take it on in a follow-up PR if reviewers agree?

pkg/mimirpb/mimir.pb.go.expdiff
```diff
@@ -20,20 +20,22 @@ index 50ea1bc359..2013d3ee8c 100644
...
  	Metadata   []*MetricMetadata       `protobuf:"bytes,3,rep,name=metadata,proto3" json:"metadata,omitempty"`
-@@ -290,12 +286,6 @@ type WriteRequest struct {
+@@ -293,14 +289,6 @@ type WriteRequest struct {
  	SkipLabelValidation bool `protobuf:"varint,1000,opt,name=skip_label_validation,json=skipLabelValidation,proto3" json:"skip_label_validation,omitempty"`
  	// Skip label count validation.
  	SkipLabelCountValidation bool `protobuf:"varint,1001,opt,name=skip_label_count_validation,json=skipLabelCountValidation,proto3" json:"skip_label_count_validation,omitempty"`
 -
 -	// Skip unmarshaling of exemplars.
 -	skipUnmarshalingExemplars bool
+-	// Skip normalization of metadata metric names
```
pracucci: [nit] I would clarify it's done at unmarshalling time.

```suggestion
-	// Skip normalization of metadata metric names when unmarshalling the request.
```

pkg/mimirpb/prealloc_rw2.go
```diff
@@ -28,6 +28,7 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 		SkipLabelValidation:       rw1.SkipLabelValidation,
 		SkipLabelCountValidation:  rw1.SkipLabelCountValidation,
 		skipUnmarshalingExemplars: rw1.skipUnmarshalingExemplars,
+		skipNormalizeMetricName:   rw1.skipNormalizeMetricName,
```
pracucci: I don't feel strong, but I'm wondering if, to de risk from future logic changes, we should set `skipNormalizeMetricName` in each place where we create `WriteRequest` from another request (there are 2 places in `split.go`: `splitMetadataByMaxMarshalSize()`, `SplitWriteRequestByMaxMarshalSizeRW2()`.

pkg/mimirpb/mimir.pb.go
```diff
@@ -11759,7 +11761,11 @@ func MetricMetadataUnmarshalRW2(dAtA []byte, symbols *rw2PagedSymbols, metadata
 	if iNdEx > l {
 		return io.ErrUnexpectedEOF
 	}
-	normalizeMetricName, _ = getMetricName(metricName, metricType)
+	if skipNormalizeMetricName {
+		normalizeMetricName = metricName
+	} else {
+		normalizeMetricName, _ = getMetricName(metricName, metricType)
```
pracucci: [nit] Have you considered renaming `getMetricName()` to `normaliseMetricName()` for clarity?

pkg/mimirpb/mimir.pb.go.expdiff
```diff
@@ -20,20 +20,22 @@ index 50ea1bc359..2013d3ee8c 100644
...
+-	skipNormalizeMetricName bool
```
pracucci: Don't hate me, please. Have you considered naming it `skipNormalizeMetadataMetricName` to clarify it just mangles the metric name in the metadata, and not in the series?

alexweav: Oh yes, that's a better name, I'll switch to that :+1:

pkg/mimirpb/prealloc_rw2.go
alexweav: Done, I also did the same for `skipUnmarshalingExemplars` which was not present in either spot.
In both cases the split requests we only use for marshalling, and the field does not persist in the proto data to the unmarshalled struct, but I agree it can help reduce surprise.

PR #12500, opened 2025-08-25
continuous-test: Add new test for verifying ingest-storage record correctness

alexweav:
#### What this PR does

Adds a new, off-by-default, continuous test that connects to the ingest-storage Kafka and verifies ingest-storage records. It intercepts a random subset of live write requests to evaluate each time the continuous tester runs.

Its primary purpose is to test the new V2 record format. If the cell is using the V1 record format, this test marshals records to V2 and asserts that both record formats map to the same actual request. It also emits statistics about the difference between V1 and V2, for the purpose of observing how V2 behaves without putting the write path at risk.

Once the V2 format reaches stability, it becomes less relevant -  but it can be a nice utility for cells wishing to migrate to V2, and also the general idea of the test might be applicable to other areas and could be extended or modified going forward.

The test does not alert when it fails, at least for now. It simply logs its output and continues.

It is designed to only apply a bounded and configurable amount of pressure to Kafka. It consumes up to a fixed amount of records per run, regardless of the throughput of the cell, and manually commits offsets to "seek" forward in time when limits are exhausted. Most records are never fetched, and the number of fetches and commits performed is a function of the limit.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/grafana/mimir/pkg/storage/ingest"
+)
+
+type IngestStorageRecordTestConfig struct {
+	Enabled                  bool               `yaml:"enabled"`
+	Kafka                    ingest.KafkaConfig `yaml:"-"`
+	ConsumerGroup            string             `yaml:"consumer_group"`
+	MaxJumpLimitPerPartition int                `yaml:"max_jump_size"`
+	MaxRecordsPerRun         int                `yaml:"max_records_per_run"`
+	RecordsProcessedPercent  int                `yaml:"records_processed_percent"`
```
pracucci: We don't support YAML config for the other test config. I would just remove the `yaml` tags. WDYT?

cmd/mimir-continuous-test/main.go
```diff
@@ -32,6 +32,7 @@ func main() {
 	)
 	flag.CommandLine.IntVar(&serverMetricsPort, "server.metrics-port", 9900, "The port where metrics are exposed.")
 	cfg.RegisterFlags(flag.CommandLine)
+	cfg.IngestStorageRecordTest.Kafka.RegisterFlagsWithPrefix("ingest-storage.kafka.", flag.CommandLine)
```
pracucci, line 35: Why not registering the Kafka flags inside `IngestStorageRecordTestConfig.RegisterFlags()`?

pracucci, line 35: Have you considered using the prefix `tests.ingest-storage-record.kafka.` to keep consistency with `IngestStorageRecordTestConfig` flags?

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+
+func (cfg *IngestStorageRecordTestConfig) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, "tests.ingest-storage-record.enabled", false, "Whether the test for ingest-storage record correctness is enabled.")
+	f.StringVar(&cfg.ConsumerGroup, "tests.ingest-storage-record.consumer-group", "ingest-storage-record", "The Kafka consumer group used for getting/setting commmitted offsets.")
+	f.IntVar(&cfg.MaxJumpLimitPerPartition, "tests.ingest-storage-record.max-jump-size", 100000000, "If a partition increases by this many offsets in a run, we skip processing it, to protect against downloading unexpectedly huge batches.")
+	f.IntVar(&cfg.MaxRecordsPerRun, "tests.ingest-storage-record.max-records-per-run", 200000, "Limit on the number of total records to be processed in a run, to keep memory bounded in large cells. This limit is approximate, we might go over by a partial fetch.")
+	f.IntVar(&cfg.RecordsProcessedPercent, "tests.ingest-storage-record.records-processed-percent", 5, "The approximate percent of records to actually fetch and compare.")
+}
+
+type IngestStorageRecordTestMetrics struct {
```
pracucci, line 45: I suggest to track a metric when the test fails, similarly to `queryResultChecksFailedTotal` in the other existing test. It would be much easier to have a quick a look at failures, rather than having to query logs.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+		return nil
+	}
+
+	level.Info(t.logger).Log("msg", "starting kafka client")
+
+	kc, err := ingest.NewKafkaReaderClient(
+		t.cfg.Kafka,
+		ingest.NewKafkaReaderClientMetrics(ingest.ReaderMetricsPrefix, "record-continuous-test", t.reg),
+		t.logger,
+		kgo.ConsumeTopics("ingest"),
```
pracucci: The name of the topic should be read by the config. You already have it.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/twmb/franz-go/pkg/kgo"
+
+	"github.com/grafana/mimir/pkg/mimirpb"
+	"github.com/grafana/mimir/pkg/storage/ingest"
+)
+
+type IngestStorageRecordTestConfig struct {
+	Enabled                  bool               `yaml:"enabled"`
+	Kafka                    ingest.KafkaConfig `yaml:"-"`
+	ConsumerGroup            string             `yaml:"consumer_group"`
```
pracucci: There's also ConsumerGroup in `ingest.KafkaConfig`. I'm wondering if an option is to use it, and just change the default.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+	}
+
+	// Find where the last run stopped at, if any.
+	topics, err := t.adminClient.ListTopics(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to ping kafka: %w", err)
+	}
+	for _, to := range topics {
+		level.Info(t.logger).Log("msg", "detected topic", "topic", to.Topic)
+	}
```
pracucci: What is this used for?

pkg/continuoustest/ingest_storage_record.go
pracucci: Same applies to other places below.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+		startOffsets["ingest"][partition] = kgo.NewOffset().At(committedOffset.At)
+	}
+
+	recordsRemainingInBatch := (totalOffsetDiff / 100) * int64(t.cfg.RecordsProcessedPercent)
+	if recordsRemainingInBatch > int64(t.cfg.MaxRecordsPerRun) {
+		level.Info(t.logger).Log("msg", "the expected number of records is larger than the limit", "expectedRecordsToProcess", recordsRemainingInBatch, "limit", t.cfg.MaxRecordsPerRun)
+		recordsRemainingInBatch = int64(t.cfg.MaxRecordsPerRun)
+	}
+
+	t.client.AddConsumePartitions(startOffsets)
```
pracucci, line 220: Don't you have to call RemoveConsumePartitions() once this function has done? If we don't do it, the kafka client will keep buffering records under the hood.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+	// Spin up workers to evaluate records in parallel.
+	const numWorkers = 4
+	jobs := make(chan kgo.Fetches)
+	var wg sync.WaitGroup
+	for range numWorkers {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for job := range jobs {
+				err = t.testBatch(job)
```
pracucci: There's a race on `err`. It's written by multiple goroutines.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,560 @@
...
+			return false
+		}
+	}
+	if len(this.Samples) != len(that1.Samples) {
+		return false
+	}
+	for i := range this.Samples {
+		/*if !this.Samples[i].Equal(&that1.Samples[i]) {
+			return false
+		}*/
```
pracucci: Should be removed?

pkg/mimir/modules.go
```diff
@@ -1302,13 +1302,16 @@ func (t *Mimir) initBlockBuilderScheduler() (services.Service, error) {
 }
 
 func (t *Mimir) initContinuousTest() (services.Service, error) {
+	t.Cfg.ContinuousTest.IngestStorageRecordTest.Kafka = t.Cfg.IngestStorage.KafkaConfig
```
pracucci, line 1305: Uhm... in a previous comment I suggest to reuse the ConsumerGroup from `t.Cfg.IngestStorage.KafkaConfig` but then if we do this, it would be risky because we could override ingesters.

pkg/continuoustest/ingest_storage_record.go
alexweav: Debugging way early on, and nothing anymore. I will delete it.

cmd/mimir-continuous-test/main.go
alexweav: I avoided it because it's really operationally convenient to exactly re-use the ingest-storage flags that are already shared among distributors/ingesters/blockbuilders/rulers. We can just consider the continuous tester as yet another consumer, meaning it's just a 1 liner in jsonnet to configure it, and it keeps in sync with the other services really well.

This bit was inspired by block builder. Block builder has several `block-builder.*` flags, but the Kafka configuration remains just `ingest-storage.kafka.*` without changing prefix.

alexweav: This `main.go` seems defunct since continuous-test was migrated to be a Mimir module. I'm not sure it's used anywhere anymore, the dockerfile runs `/bin/mimir -target=continuous-test`

When this runs as a module, the flags are already registered externally as part of the base Mimir config. That means we get a double registration panic if we register inside the test config's `RegisterFlags()`. This becomes a non-issue if we change the flag prefix because it doesn't collide anymore, but I also feel there's a valid reason to not change the prefix - if we decide to change the prefix, I'll go this route.

pkg/continuoustest/ingest_storage_record.go
alexweav: Agreed, let's drop them

pkg/mimir/modules.go
alexweav: Can joining the consumer group of another service interfere with it in any way? It seems block builder also uses its own consumer group, and also ignores the consumer group inside KafkaConfig, I assumed for similar reasons :thinking:

pkg/continuoustest/ingest_storage_record.go
alexweav: Removed it, it's not necessary. (was to demonstrate what code changed from the vendored one)

pkg/mimir/modules.go
alexweav: Oh, I understand better now - you're saying to reuse the config option, not the actual consumer group. We could do that. For all this config wiring I exactly matched the pattern from block builder. If we decide not to match the base config options and wire up things in the same style as block-builder (per the other thread) let's drop it.

pkg/continuoustest/ingest_storage_record.go
alexweav: Very subtle catch, nice, I meant to shadow it. Changed to `:=`

pkg/continuoustest/ingest_storage_record.go
alexweav: Done :+1: Added `mimir_continuous_test_ingest_storage_batches_total` and `mimir_continuous_test_ingest_storage_batches_failed_total`

pkg/continuoustest/ingest_storage_record.go
alexweav: That makes sense, I thought PollFetches was synchronously fetching records, but that's wrong, there's an internal buffer and async loop.

Went back through and I don't think we need to change anything else beyond adding this.

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,570 @@
...
+	endOffsets := allPartitionEndOffsets[t.cfg.Kafka.Topic]
+
+	// Build a set of offsets to start reading from.
+	startPartitions := []int32{}
+	startOffsets := map[string]map[int32]kgo.Offset{t.cfg.Kafka.Topic: {}}
+	totalOffsetDiff := int64(0)
+	for partition, endOffset := range endOffsets {
+		// If we never committed an offset for this partition yet, we just ignore it on the current run.
+		// It likely has a ton of history that hasn't been processed yet. We'll still seek to the end later on and pick up on the next time around.
+		committedOffset, ok := committedOffsets[t.cfg.Kafka.Topic][partition]
```
alexweav: can be `.Lookup()`

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,570 @@
...
+	if recordsRemainingInBatch > int64(t.cfg.MaxRecordsPerRun) {
+		level.Info(t.logger).Log("msg", "the expected number of records is larger than the limit", "expectedRecordsToProcess", recordsRemainingInBatch, "limit", t.cfg.MaxRecordsPerRun)
+		recordsRemainingInBatch = int64(t.cfg.MaxRecordsPerRun)
+	}
+
+	t.client.AddConsumePartitions(startOffsets)
+
+	// Spin up workers to evaluate records in parallel.
+	const numWorkers = 4
+	jobs := make(chan kgo.Fetches)
```
alexweav, line 224: Channel is missing a bound here

pkg/continuoustest/ingest_storage_record.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+
+	t.client = kc
+	t.adminClient = kadm.NewClient(kc)
+
+	return nil
+}
+
+// Run implements Test.
+func (t *IngestStorageRecordTest) Run(ctx context.Context, now time.Time) error {
```
alexweav, line 156: Should listen for context.Done() where appropriate below (i.e. exit fetch loop or workers)

alexweav:
Reviewed synchronously on a call with @gotjosh. The comments posted by me above are notes from the call. Thanks!

pkg/continuoustest/ingest_storage_record.go
alexweav: We're ok on this one, couldn't remember in the call if we were or not.

It's not unbounded, the default bound is 0, meaning each fetch blocks until there's a worker ready to process it - that's what we want. The linter complains if I try to be more clear here, it wants to simplify `make(chan kgo.Fetches, 0)` to this.

pkg/continuoustest/ingest_storage_record.go
alexweav: If we kill the workers when ctx.Done(), the fetch loop won't exit, and potentially block infinitely on `jobs <- fetches`. If we close the channel it also might panic the fetch loop on the same line. So we must do this in the outer fetch loop.

alexweav: Done via checking `ctx.Err()` which returns a context cancellation error if the context is Done. If it gets cancelled we now stop fetching immediately wherever we are.

PR #12501, opened 2025-08-26
update deployment modes topic

tacole02:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/docs/sources/mimir/references/architecture/deployment-modes/index.md
+++ b/docs/sources/mimir/references/architecture/deployment-modes/index.md
@@ -11,17 +11,24 @@ weight: 20
 
 # Grafana Mimir deployment modes
 
-You can deploy Grafana Mimir in three modes:
+Grafana Mimir offers two deployment modes to accommodate different operational requirements and scale needs. Choose the deployment mode that best fits your use case:
 
-- Monolithic mode
-- Microservices mode
-- Read-Write mode (experimental)
+- Monolithic mode: Run all components in a single process for simple deployments
+- Microservices mode: Deploy components separately for maximum scalability and flexibility
 
-The deployment mode is determined by the `-target` parameter, which you can set via CLI flag or YAML configuration.
+Configure the deployment mode using the `-target` parameter, which you can set via CLI flag or YAML configuration.
 
-## Monolithic mode
+## About monolithic mode
 
-The monolithic mode runs all required components in a single process and is the default mode of operation, which you can set by specifying `-target=all`. Monolithic mode is the simplest way to deploy Grafana Mimir and is useful if you want to get started quickly or want to work with Grafana Mimir in a development environment. To see the list of components that run when `-target` is set to `all`, run Grafana Mimir with the `-modules` flag:
+Monolithic mode runs all required components in a single process and is the default deployment mode. You enable monolithic mode by setting `-target=all`. 
+
+This mode is ideal when you want to:
+- Get started quickly with Mimir
+- Run Mimir in a development environment  
+- Deploy with minimal operational complexity
+- Scale all components together
+
+To see the complete list of components that run in monolithic mode, use the `-modules` flag:
 
 ```bash
 ./mimir -modules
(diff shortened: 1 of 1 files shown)


tacole02:
hey @dimitarvdimitrov, I've updated the [Grafana Mimir deployment modes](https://grafana.com/docs/mimir/latest/references/architecture/deployment-modes/) topic, giving it a general clean up and noting that read-write mode isn't supported in 3.0.

I plan to update the diagram under [Microservices mode](https://grafana.com/docs/mimir/latest/references/architecture/deployment-modes/#microservices-mode) as soon as I get the new one from the content team. 

Are there are other updates required for ingest storage? Thanks!

dimitarvdimitrov (commented):
have you thought about just deleting the docs for it? The code for read-write mode will be gone in 3.0. Are we ok with people having to switch to a 2.x version of the docs to view the read-write docs?

as for other changes - looking at the content right now I don't see any third-party deps like etcd/consul/object store, so i think we can also skip mentioning kafka

tacole02:
I think you're right. If the code is gone in 3.0, and it was only ever experimental to begin with, I think we can safely delete it. 

I'm still waiting for the updated diagram from the content team. After I get that, this will be ready to merge 🥳

tacole02:
Hi @dimitarvdimitrov , I added an updated diagram for ingest storage architecture in microservices mode. Based on our discussion at our last sync, I added it alongside the existing diagram for classic. Let me know if there's anything else I need to adjust here. Thanks!

dimitarvdimitrov:
Screenshot 2025-09-04 at 11 51 46


the new graph looks a bit fried, do you have one with higher resolution?

tacole02:
This is ready, with the exception of the updated ingest storage overview diagram, which is captured in https://github.com/grafana/mimir-squad/issues/3128.

docs/sources/mimir/references/architecture/deployment-modes/index.md
```diff
@@ -11,27 +11,39 @@ weight: 20
...
+- Microservices mode: Deploy components separately for maximum scalability and flexibility.
 
-The deployment mode is determined by the `-target` parameter, which you can set via CLI flag or YAML configuration.
+Configure the deployment mode using the `-target` parameter, which you can set via CLI flag or YAML configuration.
 
-## Monolithic mode
+## About monolithic mode
 
-The monolithic mode runs all required components in a single process and is the default mode of operation, which you can set by specifying `-target=all`. Monolithic mode is the simplest way to deploy Grafana Mimir and is useful if you want to get started quickly or want to work with Grafana Mimir in a development environment. To see the list of components that run when `-target` is set to `all`, run Grafana Mimir with the `-modules` flag:
+Monolithic mode runs all required components in a single process and is the default deployment mode. You enable monolithic mode by setting `-target=all`.
```
56quarters: I don't really like mentioning that monolithic mode is the default here because:
* We don't want people to be running it in production but "default" makes it sound like it's the preferred way to run Mimir
* It's _not_ the default in the helm chart. It's only really the default when you run Mimir from the CLI and don't otherwise configure each component.

docs/sources/mimir/references/architecture/deployment-modes/index.md
```diff
@@ -41,46 +53,26 @@ Monolithic mode can be horizontally scaled out by deploying multiple Grafana Mim
 Because monolithic mode requires scaling all Grafana Mimir components together, this deployment mode isn't recommended for large-scale deployments.
 {{< /admonition >}}
 
-## Microservices mode
+## About microservices mode
 
-In microservices mode, components are deployed in distinct processes. Scaling is per component, which allows for greater flexibility in scaling and more granular failure domains. Microservices mode is the preferred method for a production deployment, but it is also the most complex.
+Microservices mode deploys each component in separate processes, enabling independent scaling and creating granular failure domains. This deployment mode is recommended for production environments that require high availability and flexible scaling.
```
56quarters: I'd rather not qualify when microservices mode is recommended for production, it's always recommended. We'd only expect people to _not_ run it in production in special circumstances that tend to be very unique to their environment or organization.

docs/sources/mimir/references/architecture/deployment-modes/index.md
```diff
@@ -11,27 +11,39 @@ weight: 20
...
+## About monolithic mode
 
-The monolithic mode runs all required components in a single process and is the default mode of operation, which you can set by specifying `-target=all`. Monolithic mode is the simplest way to deploy Grafana Mimir and is useful if you want to get started quickly or want to work with Grafana Mimir in a development environment. To see the list of components that run when `-target` is set to `all`, run Grafana Mimir with the `-modules` flag:
+Monolithic mode runs all required components in a single process and is the default deployment mode. You enable monolithic mode by setting `-target=all`.
+
+This mode is ideal when you want to:
+
+- Get started quickly with Mimir
+- Run Mimir in a development environment
+- Deploy with minimal operational complexity
```
56quarters: This makes microservices mode sound like it's an optimization when it's actually the right way to be running Mimir. I would remove this and the next point. We don't want people to try running monolithic mode anywhere outside of "getting started".

docs/sources/mimir/references/architecture/deployment-modes/index.md
tacole02: Good call out, I'll remove this.

tacole02:
Thanks for your feedback @56quarters . I've addressed all of your comments. Do you mind taking another look?

56quarters (approved):
LGTM, thanks!

PR #12506, opened 2025-08-26
Add debug logging to flaky store-gateway limits test

dimitarvdimitrov:
This test intermittently fails in CI. Adding instrumentation to help debug the issue when it occurs in future runs.

Related to #12485

### DIFF
--- a/integration/store_gateway_limits_hit_test.go
+++ b/integration/store_gateway_limits_hit_test.go
@@ -74,6 +74,7 @@ func Test_MaxSeriesAndChunksPerQueryLimitHit(t *testing.T) {
 	client, err := e2emimir.NewClient(distributor.HTTPEndpoint(), "", "", "", "test")
 
 	for i, ts := range []time.Time{timeStamp1, timeStamp2, timeStamp3} {
+		t.Logf("Writing block %d at timestamp %v", i+1, ts)
 		for j := 0; j < numSeriesPerBlock; j++ {
 			series, _, _ := generateAlternatingSeries(i)(fmt.Sprintf("series_%d", j), ts)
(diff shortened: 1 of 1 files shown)


integration/store_gateway_limits_hit_test.go
```diff
@@ -146,22 +149,54 @@ func Test_MaxSeriesAndChunksPerQueryLimitHit(t *testing.T) {
...
+			t.Logf("Components started and ready: querier, store-gateway-1, store-gateway-2")
 			t.Cleanup(func() {
 				require.NoError(t, scenario.Stop(querier, storeGateway1, storeGateway2, compactor))
 			})
 
 			client, err = e2emimir.NewClient(distributor.HTTPEndpoint(), querier.HTTPEndpoint(), "", "", "test")
 			require.NoError(t, err)
 
+			// Add a small delay to ensure store-gateway has loaded all blocks
+			time.Sleep(2 * time.Second)
```
tcard: **suggestion:** Instead of having a "big enough" wait and then a single check, what I usually do is checking multiple times with a smaller wait interval. This way the test ends faster if the condition is met sooner.

I usually have an `eventually(check, timeout, checkInterval)` helper function around, I don't know if we already have an equivalent.

tcard: Actually it's built-in in testify, so we can use `require.Eventually` or `require.EventuallyWithT`.

dimitarvdimitrov: 👍 done in https://github.com/grafana/mimir/pull/12506/commits/5f18b3dacb137ec8bba323690d4540b41a8697a5

PR #12507, opened 2025-08-26
streamingpromql: handle adaptive metrics reserved functions in `functionNeedsDeduplication`

ortuman:
#### What this PR does

This PR addresses a panic observed in `streamingpromql` during `TestAggregatedQueryCorrectness` caused by unhandled reserved enum values for Adaptive Metrics functions.

Fixes n/a

ref: https://raintank-corp.slack.com/archives/C07FBAEPENT/p1756205426612819


pkg/streamingpromql/planning.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/grafana/mimir/pkg/streamingpromql/types"
 )
 
+const (
+	ADAPTIVE_METRICS_RESERVED_FUNCTION_0 = 74
```
56quarters: We reserve the `74` and `75` ordinals already for Adaptive metrics in `pkg/streamingpromql/operators/functions/functions.proto`. Might as well just define them there instead of reserving them and then defining them here. I don't think there's any drawbacks to do that as long as we keep the same behavior here. WDYT @charleskorn ?

charleskorn: I'm happy either way - I originally didn't define the functions in `functions.proto` in the interests of keeping as much of the details of the closed-source Adaptive Metrics functionality out of open source Mimir, but I don't hold this opinion strongly.

ortuman: fine be me! 

_changed @ 1a3cce41092d96f05ded9379dff455b07a98633e_

ortuman:
@56quarters @charleskorn this is ready for review again

pkg/streamingpromql/planning.go
```diff
@@ -556,7 +556,10 @@ func functionNeedsDeduplication(fnc functions.Function) bool {
 		functions.FUNCTION_HISTOGRAM_SUM,
 		// Label manipulation functions
 		functions.FUNCTION_LABEL_JOIN,
-		functions.FUNCTION_LABEL_REPLACE:
+		functions.FUNCTION_LABEL_REPLACE,
+		// Adaptive metrics reserved functions
+		functions.FUNCTION_ADAPTIVE_METRICS_RESERVED_1,
```
Konstantinov-Innokentii, line 561: I'm wondering if these functions should be added to the case below. I'm not sure if they are manipulating labes to be honest. Do we actually pass these functions to the query engine, or adaptive metrics query rewriting happens earlier, before the querier?

56quarters: If we don't know in Mimir if they manipulate labels or not, is it safer to assume they do or assume they don't? Or are you saying they _couldn't_ manipulate labels?

charleskorn: If we're unsure, it's safer to assume that something does manipulate labels, as then we'll insert a deduplicate and merge operation that will be a no-op if there's nothing to do.

However, in the case of these two functions, they merely inspect the labels, so we don't need a deduplicate and merge operation for them.

PR #12509, opened 2025-08-26
MQE: implement delayed name removal

zenador:
#### What this PR does

Implement https://github.com/prometheus/prometheus/pull/14477 in MQE.

Fixes https://github.com/grafana/mimir-squad/issues/3065

#### Checklist

- [x] Tests updated.
- [x] Documentation added.

### DIFF
--- a/pkg/streamingpromql/testdata/ours/name_label_dropping.test
+++ b/pkg/streamingpromql/testdata/ours/name_label_dropping.test
@@ -95,10 +95,10 @@ eval instant at 10m sum by (__name__, foo) (down or rate(up[5m]))
   down{foo="bar2"} 40
   {foo="bar3"} 0.08333333333333334
 
-# Temporarily comment out flaky tests.
+# Temporarily comment out tests that are consistent in Mimir but flaky in Prometheus
+# until they are fixed, see https://github.com/prometheus/prometheus/pull/17161
 #eval instant at 10m sum by (__name__) (up{foo="bar"} or rate(up{foo="bar2"}[5m]))
 #  up 10.05
 
-# Temporarily comment out flaky tests.
 #eval instant at 10m sum by (__name__) (rate(up{foo="bar2"}[5m]) or up{foo="bar"})
 #  {} 10.05


pkg/streamingpromql/engine_test.go
```diff
@@ -4131,3 +4132,55 @@ type dummyMaterializer struct{}
 func (d dummyMaterializer) Materialize(n planning.Node, materializer *planning.Materializer, timeRange types.QueryTimeRange, params *planning.OperatorParameters) (planning.OperatorFactory, error) {
 	panic("not implemented")
 }
+
+func TestEnableDelayedNameRemoval(t *testing.T) {
```
charleskorn: Is there a reason to run these tests as Go tests rather than with a PromQL test script in `testdata/ours`?

pkg/streamingpromql/engine_test.go
```diff
@@ -4131,3 +4132,55 @@ type dummyMaterializer struct{}
...
+
+	ctx := context.Background()
+	startTime := timestamp.Time(0)
+	endTime := startTime.Add(time.Duration(numSamples) * time.Minute)
+
+	testCases := map[string]string{
+		"original":             `count_over_time({__name__!=""}[1m])`,
+		"fixed":                `label_replace(count_over_time({__name__!=""}[1m]), "__name__", "count_$1", "__name__", "(.+)")`,
+		"function after label": `abs(label_replace(count_over_time({__name__!=""}[1m]), "__name__", "count_$1", "__name__", "(.+)"))`,
+	}
```
charleskorn: Are these test cases different to those in `testdata/upstream/name_label_dropping.test`? It seems like at least the first two are already covered there.

pkg/streamingpromql/types/operator.go
```diff
@@ -50,7 +50,7 @@ type SeriesOperator interface {
 	// The returned []SeriesMetadata can be modified by the caller or returned to a pool.
 	// SeriesMetadata may return series in any order, but the same order must be used by both SeriesMetadata and NextSeries.
 	// SeriesMetadata should be called no more than once.
-	SeriesMetadata(ctx context.Context) ([]SeriesMetadata, error)
+	SeriesMetadata(ctx context.Context) (*SeriesMetadataSet, error)
```
charleskorn: For other operator methods, we haven't used pointers for simple return values like this (eg. `InstantVectorSeriesData`) - is there a reason to make this a pointer?

pkg/streamingpromql/query.go
```diff
@@ -81,6 +81,8 @@ func (q *Query) Exec(ctx context.Context) *promql.Result {
 	_, isInstantVectorOperator := q.evaluator.root.(types.InstantVectorOperator)
 	q.resultIsVector = q.topLevelQueryTimeRange.IsInstant && isInstantVectorOperator
 
+	ctx = context.WithValue(ctx, types.ContextEnableDelayedNameRemoval, q.engine.enableDelayedNameRemoval)
+
```
charleskorn: Rather than adding this to the context, I would add another field to `planning.OperatorParameters`, and allow operators that need this to read the value from there, similar to `EagerLoadSelectors`.

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/range_vector_operator.go
```diff
@@ -59,14 +59,17 @@ func (b *RangeVectorDuplicationBuffer) AddConsumer() *RangeVectorDuplicationCons
...
+	var dropName bool
 	if b.seriesMetadataCount == 0 {
 		// Haven't loaded series metadata yet, load it now.
 		var err error
-		b.seriesMetadata, err = b.Inner.SeriesMetadata(ctx)
+		seriesMetadata, err := b.Inner.SeriesMetadata(ctx)
 		if err != nil {
 			return nil, err
 		}
+		b.seriesMetadata = seriesMetadata.Metadata
```
charleskorn: What if we changed `b.seriesMetadata` to be a `types.SeriesMetadataSet`?

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/instant_vector_operator.go
```diff
@@ -52,14 +52,17 @@ func (b *InstantVectorDuplicationBuffer) AddConsumer() *InstantVectorDuplication
...
+	var dropName bool
 	if b.seriesMetadataCount == 0 {
 		// Haven't loaded series metadata yet, load it now.
 		var err error
-		b.seriesMetadata, err = b.Inner.SeriesMetadata(ctx)
+		seriesMetadata, err := b.Inner.SeriesMetadata(ctx)
 		if err != nil {
 			return nil, err
 		}
+		b.seriesMetadata = seriesMetadata.Metadata
```
charleskorn: What if we changed `b.seriesMetadata` to be a `SeriesMetadataSet`?

pkg/streamingpromql/operators/functions/common.go
```diff
@@ -10,21 +10,28 @@ import (
...
 			tracker.DecreaseMemoryConsumptionForLabels(seriesMetadata[i].Labels)
-			seriesMetadata[i].Labels = seriesMetadata[i].Labels.DropMetricName()
+			if !enableDelayedNameRemoval {
+				seriesMetadata[i].Labels = seriesMetadata[i].Labels.DropMetricName()
+			}
 			err := tracker.IncreaseMemoryConsumptionForLabels(seriesMetadata[i].Labels)
 			if err != nil {
 				return nil, err
 			}
 		}
```
charleskorn: If delayed name removal is enabled, then we can skip this whole loop. (or even just return early?)

pkg/streamingpromql/operators/functions/histogram_function.go
```diff
@@ -257,7 +257,7 @@ func (h *HistogramFunction) SeriesMetadata(ctx context.Context) ([]types.SeriesM
 	// as soon as possible when accumulating them.
 	sort.Sort(bucketGroupSorter{seriesMetadata, h.remainingGroups})
 
-	return seriesMetadata, nil
+	return &types.SeriesMetadataSet{Metadata: seriesMetadata, DropName: innerSeries.DropName}, nil
```
charleskorn: Could you please add some tests for this?

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -81,7 +84,7 @@ func (a *Absent) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, er
 		}
 		types.PutInstantVectorSeriesData(series, a.MemoryConsumptionTracker)
 	}
-	return metadata, nil
+	return &types.SeriesMetadataSet{Metadata: metadata, DropName: innerMetadata.DropName}, nil
```
charleskorn: Could you please add some tests for this and `absent_over_time`?

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -177,7 +177,7 @@ func (a *Aggregation) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
 
 	sort.Sort(groupSorter{seriesMetadata, a.remainingGroups})
 
-	return seriesMetadata, nil
+	return &types.SeriesMetadataSet{Metadata: seriesMetadata, DropName: innerSeries.DropName}, nil
```
charleskorn: Shouldn't `DropName` always be `false` in this case?

pkg/streamingpromql/operators/aggregations/count_values.go
```diff
@@ -149,7 +149,7 @@ func (c *CountValues) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
 		countValuesSeriesPool.Put(s)
 	}
 
-	return outputMetadata, nil
+	return &types.SeriesMetadataSet{Metadata: outputMetadata, DropName: innerMetadata.DropName}, nil
```
charleskorn: Same as above - shouldn't `DropName` always be false?

pkg/streamingpromql/operators/aggregations/topkbottomk/instant_query.go
```diff
@@ -161,7 +161,7 @@ func (t *InstantQuery) SeriesMetadata(ctx context.Context) ([]types.SeriesMetada
 		instantQuerySeriesSlicePool.Put(&g.series, t.MemoryConsumptionTracker)
 	}
 
-	return outputSeries, nil
+	return &types.SeriesMetadataSet{Metadata: outputSeries, DropName: innerSeries.DropName}, nil
```
charleskorn: Same as above - shouldn't `DropName` always be false?

pkg/streamingpromql/operators/deduplicate_and_merge.go
```diff
@@ -34,14 +34,36 @@ func NewDeduplicateAndMerge(inner types.InstantVectorOperator, memoryConsumption
...
+		return nil, types.ErrEnableDelayedNameRemovalNotFound
+	}
+
+	deduplicateAndMergeStackCount, ok := ctx.Value(types.ContextDeduplicateAndMergeStackCount).(int)
+	if !ok {
+		deduplicateAndMergeStackCount = 0
+	}
+
+	ctx = context.WithValue(ctx, types.ContextDeduplicateAndMergeStackCount, deduplicateAndMergeStackCount+1)
+
```
charleskorn: Rather than doing this, I think it would be more robust for there to be a flag on the `DeduplicateAndMerge` operator that tells it if it should do name removal.

We can then set this on the root-level `DeduplicateAndMerge` node in the query plan when the naive plan is first created [here](https://github.com/grafana/mimir/blob/bb3e69db2b03db43df6b4437c5d14474e024e1fb/pkg/streamingpromql/planning.go#L159), and pass it to the operator when creating the operator.

One complication is that the optimisation @Konstantinov-Innokentii is working on will need to not remove the `DeduplicateAndMerge` node if the name removal flag is set. (On this topic: we can also omit all the  `DeduplicateAndMerge` nodes for function calls, unary negation and vector/scalar binary operations if delayed name removal is enabled and that node isn't the root of the query. We'll still always need them around `or`, `label_join` and `label_replace`.)

pkg/streamingpromql/engine_test.go
zenador: It's easier for me to debug during development, but I can remove these when the PR is ready to merge.

pkg/streamingpromql/engine_test.go
zenador: Same as above.

pkg/streamingpromql/types/operator.go
zenador: It was so we could continue returning nil when needed, but I've changed it to returning an empty struct instead so we don't have to check for nil.

pkg/streamingpromql/query.go
zenador: Couldn't get this to work, what's the best way to pass the new field to places like [this line](https://github.com/grafana/mimir/blob/zenador/mqe-delayed-name-removal-remove-context/pkg/streamingpromql/operators/functions/function_over_range_vector.go#L95)?

zenador:
Also having some trouble getting some test cases in [TestUpstreamTestCases](https://github.com/grafana/mimir/actions/runs/17278822286/job/49042408053?pr=12509) to pass.

pkg/streamingpromql/types/operator.go
charleskorn: Couldn't we check the slice inside the struct for `nil` if we need to do that?

pkg/streamingpromql/query.go
charleskorn: `MaterializeFunctionCall` will need to pass it from the `planning.OperatorParameters` to the function's factory [here](https://github.com/grafana/mimir/blob/7b8de88d7e61f1fdbabe8f4889ed182f10355eeb/pkg/streamingpromql/planning/core/function_call.go#L109), and then that will be able to pass it when creating the `FunctionOverRangeVector` instance.

zenador: I've tried that, is this what you meant? Since we're changing the signature so much anyway, should we just pass in params as a whole to minimise future changes?

pkg/streamingpromql/types/operator.go
zenador: We don't have to add extra nil checks for the slice inside the struct as that was what we were originally using, but with the previous pointers, we would need to check if the struct is nil before we use the slice inside it. Now without the pointer, we create an empty struct instead of nil so no need for a new nil check.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * [FEATURE] Distributor: Add experimental `-distributor.otel-translation-strategy` flag to support configuring the metric and label name translation strategy in the OTLP endpoint. #12284 #12306 #12369
 * [FEATURE] Query-frontend: Add `query-frontend.rewrite-propagate-matchers` flag that enables a new MQE AST optimization pass that copies relevant label matchers across binary operations. #12304
 * [FEATURE] Query-frontend: Add `query-frontend.rewrite-histogram-queries` flag that enables a new MQE AST optimization pass that rewrites histogram queries for a more efficient order of execution. #12305
+* [FEATURE] Query-frontend: Support delayed name removal (Prometheus experimental function) in MQE. #12509
```
charleskorn:
```suggestion
* [FEATURE] Query-frontend: Support delayed name removal (Prometheus experimental feature) in MQE. #12509
```

pkg/streamingpromql/planning/core/deduplicate_and_merge.go
```diff
@@ -78,6 +78,7 @@ func MaterializeDeduplicateAndMerge(d *DeduplicateAndMerge, materializer *planni
 		return nil, err
 	}
 
-	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker)
+	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker, params.EnableDelayedNameRemoval, d.RunDelayedNameRemoval)
```
charleskorn: What if `DeduplicateAndMerge` had just a single parameter, like this?

```suggestion
	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker, params.EnableDelayedNameRemoval && d.RunDelayedNameRemoval)
```

pkg/streamingpromql/operators/functions/absent.go
```diff
@@ -81,7 +81,7 @@ func (a *Absent) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, er
 		}
 		types.PutInstantVectorSeriesData(series, a.MemoryConsumptionTracker)
 	}
-	return metadata, nil
+	return types.SeriesMetadataSet{Metadata: metadata, DropName: innerMetadata.DropName}, nil
```
charleskorn: Is this correct? I would expect `DropName` to always be `false`

pkg/streamingpromql/operators/functions/absent_over_time.go
```diff
@@ -46,47 +46,47 @@ func NewAbsentOverTime(
...
-				return nil, err
+				return types.NewEmptySeriesMetadataSet(), err
 			}
 			if step.Floats.Any() || step.Histograms.Any() {
 				a.presence[stepIdx] = true
 			}
 		}
 	}
-	return metadata, nil
+	return types.SeriesMetadataSet{Metadata: metadata, DropName: innerMetadata.DropName}, nil
```
charleskorn: Same as above - I would expect `DropName` to always be false.

pkg/streamingpromql/operators/binops/vector_scalar_binary_operation.go
```diff
@@ -94,26 +94,23 @@ func NewVectorScalarBinaryOperation(
...
 	}
 
 	if !v.Op.IsComparisonOperator() || v.ReturnBool {
 		// We don't need to do deduplication and merging of series in this operator: we expect that this operator
 		// is wrapped in a DeduplicateAndMerge.
-		metadata, err = functions.DropSeriesName.Func(metadata, v.MemoryConsumptionTracker)
-		if err != nil {
-			return nil, err
-		}
+		return functions.DropSeriesName.Func(metadata, v.MemoryConsumptionTracker, false)
```
charleskorn: Don't we need to pass `EnableDelayedNameRemoval` here? (rather than always passing false)

pkg/streamingpromql/planning/core/deduplicate_and_merge.go
charleskorn, line 1: Reminder from in-person discussion: we should update `Describe` to include the state of `runDelayedNameRemoval`

pkg/streamingpromql/query.go
zenador: Tried changing the signature as discussed. No import cycle so I just kept passing in `planning.OperatorParameters` for simplicity.

pkg/streamingpromql/operators/aggregations/aggregation.go
zenador: Ended up reverting this to pass tests

pkg/streamingpromql/operators/aggregations/count_values.go
zenador: Ended up reverting this to pass tests

pkg/streamingpromql/operators/aggregations/topkbottomk/instant_query.go
zenador: Ended up reverting this to pass tests

zenador:
This was kind of 99% there but there's just a few tests (e.g. TestUpstreamTestCases/upstream/operators.test/line_182) that don't pass now that makes me think we might need to set dropName per series instead of the entire batch because of OR operations. Which would be a huge change, so we can't split the PR into a refactor that just changes the signature first as we cannot confirm the needed signature yet. Hopefully I'm overcomplicating things or misunderstood something somewhere and there's a much more straightforward solution?

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -2705,6 +2705,11 @@ The `querier` block configures the querier.
 # CLI flag: -querier.lookback-delta
 [lookback_delta: <duration> | default = 5m]
 
+# (experimental) Enable the experimental Prometheus feature for delayed name
```
charleskorn, line 2708: Should we add this to the list of experimental features in `about-versioning.md` too?

pkg/streamingpromql/engine_test.go
```diff
@@ -206,13 +207,22 @@ func TestUpstreamTestCases(t *testing.T) {
...
+
 	opts := NewTestEngineOpts()
-	planner, err := NewQueryPlanner(opts)
-	require.NoError(t, err)
-	mimirEngine, err := NewEngine(opts, NewStaticQueryLimitsProvider(0), stats.NewQueryMetrics(nil), planner, log.NewNopLogger())
-	require.NoError(t, err)
+	mimirEngine, prometheusEngine := makeEngines(t, opts)
 
-	prometheusEngine := promql.NewEngine(opts.CommonOpts)
+	opts.CommonOpts.EnableDelayedNameRemoval = true
```
charleskorn: It might be safer to create a new set of options here rather than reusing `opts` - there's no guarantee that the engines created above don't read the options passed to them after they're created.

pkg/streamingpromql/operators/binops/and_unless_binary_operation.go
```diff
@@ -143,9 +143,9 @@ func (a *AndUnlessBinaryOperation) computeSeriesMetadata(ctx context.Context) ([
 	}
 
 	if a.IsUnless {
-		return a.computeUnlessSeriesMetadata(leftMetadata), nil
+		return a.computeUnlessSeriesMetadata(leftMetadata.Metadata), leftMetadata.DropName || rightMetadata.DropName, nil
 	}
-	return a.computeAndSeriesMetadata(leftMetadata), nil
+	return a.computeAndSeriesMetadata(leftMetadata.Metadata), leftMetadata.DropName || rightMetadata.DropName, nil
```
charleskorn: Could you please add some tests that cover the scenarios where neither, one or both sides have delayed name removal?

pkg/streamingpromql/operators/binops/or_binary_operation.go
```diff
@@ -57,73 +57,77 @@ func NewOrBinaryOperation(
...
 
-	o.computeGroups(leftMetadata, rightMetadata)
+	o.computeGroups(leftMetadata.Metadata, rightMetadata.Metadata)
 
-	return o.computeSeriesOutputOrder(leftMetadata, rightMetadata)
+	result, err := o.computeSeriesOutputOrder(leftMetadata.Metadata, rightMetadata.Metadata)
+	if err != nil {
+		return types.NewEmptySeriesMetadataSet(), err
+	}
+	return types.SeriesMetadataSet{Metadata: result, DropName: leftMetadata.DropName || rightMetadata.DropName}, nil
```
charleskorn: Same as above, it would be good to add some tests for the cases where neither, one or both sides need name removal.

pkg/streamingpromql/operators/deduplicate_and_merge_test.go
charleskorn, line 1: Could you please add some tests for the case where delayed name removal is enabled and the series metadata set has the `DropName` flag set or not set?

pkg/streamingpromql/planning.go
```diff
@@ -173,6 +175,34 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 			return nil, err
 		}
 
+		if p.enableDelayedNameRemoval {
```
charleskorn, line 179: [nit] What if we extracted this to a method? Then we could `return` early in a bunch of cases, which should help make this easier to follow.

pkg/streamingpromql/planning.go
```diff
@@ -173,6 +175,34 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
...
+				shouldWrapInDedupAndMerge := true
+				switch node := root.(type) {
+				case *core.NumberLiteral, *core.StringLiteral:
+					shouldWrapInDedupAndMerge = false
+				case *core.BinaryExpression:
+					if node.VectorMatching == nil {
+						shouldWrapInDedupAndMerge = false
+					}
+				case *core.FunctionCall:
+					if !functionHasVectorSelectors(node.Function) {
```
charleskorn: I don't think this condition is quite right - we should check the return value type of the function. If it's a scalar, then we don't need the deduplication and merge node, but otherwise, we do.

pkg/streamingpromql/planning/core/binary_expression.go
```diff
@@ -165,7 +165,7 @@ func MaterializeBinaryExpression(b *BinaryExpression, materializer *planning.Mat
 		vector = lhsVector
 	}
 
-	o, err := binops.NewVectorScalarBinaryOperation(scalar, vector, scalarIsLeftSide, op, b.ReturnBool, timeRange, params.MemoryConsumptionTracker, params.Annotations, b.ExpressionPosition())
+	o, err := binops.NewVectorScalarBinaryOperation(scalar, vector, scalarIsLeftSide, op, b.ReturnBool, timeRange, *params, b.ExpressionPosition())
```
charleskorn: Rather than deferencing the parameters, should we just pass the pointer? 

(same thing applies elsewhere)

pkg/streamingpromql/planning/core/deduplicate_and_merge.go
```diff
@@ -78,6 +81,7 @@ func MaterializeDeduplicateAndMerge(d *DeduplicateAndMerge, materializer *planni
 		return nil, err
 	}
 
-	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker)
+	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker, params.EnableDelayedNameRemoval && d.RunDelayedNameRemoval)
```
charleskorn: During materialization, we should respect whatever is in the node and ignore the engine options - otherwise different queriers could make different decisions based on their config, which will be problematic for remote execution of query plans, where we want to have all queriers behave consistently based on what the query-frontend tells them to do.
```suggestion
	o := operators.NewDeduplicateAndMerge(inner, params.MemoryConsumptionTracker, d.RunDelayedNameRemoval)
```

charleskorn:
> This was kind of 99% there but there's just a few tests (e.g. TestUpstreamTestCases/upstream/operators.test/line_182) that don't pass now that makes me think we might need to set dropName per series instead of the entire batch because of OR operations. Which would be a huge change, so we can't split the PR into a refactor that just changes the signature first as we cannot confirm the needed signature yet. Hopefully I'm overcomplicating things or misunderstood something somewhere and there's a much more straightforward solution?

🤮 Yep, you're correct, it looks like we need to do this per-series. 

On the upside, it likely means less refactoring is required: we don't need `SeriesMetadataSet` and can instead add a `DropName` flag to the `SeriesMetadata` struct, so all the `SeriesMetadata()` methods can continue returning `[]SeriesMetadata`.

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -168,7 +170,7 @@ func (a *Aggregation) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadat
 	a.remainingGroups = make([]*group, 0, len(groups))
 
 	for _, g := range groups {
-		seriesMetadata, err = types.AppendSeriesMetadata(a.MemoryConsumptionTracker, seriesMetadata, types.SeriesMetadata{Labels: g.labels})
+		seriesMetadata, err = types.AppendSeriesMetadata(a.MemoryConsumptionTracker, seriesMetadata, types.SeriesMetadata{Labels: g.labels, DropName: g.dropName})
```
charleskorn, line 176: Should `DropName` always be false in this case? Seems unexpected that it would be determined by the first series seen for the group.

pkg/streamingpromql/operators/deduplicate_and_merge.go
```diff
@@ -41,6 +42,20 @@ func (d *DeduplicateAndMerge) SeriesMetadata(ctx context.Context) ([]types.Serie
 		return nil, err
 	}
 
+	if d.RunDelayedNameRemoval {
+		for i := range innerMetadata {
+			if !innerMetadata[i].DropName {
+				continue
+			}
+			d.MemoryConsumptionTracker.DecreaseMemoryConsumptionForLabels(innerMetadata[i].Labels)
+			innerMetadata[i].Labels = innerMetadata[i].Labels.DropMetricName()
```
charleskorn, line 51: Should we also set `innerMetadata[i].DropName = false` here?

tacole02 (approved):
Docs LGTM!

pkg/streamingpromql/operators/deduplicate_and_merge.go
zenador: Don't think we actually need to since this should only happen on the outermost DeduplicateAndMerge and it won't be passed further, but it probably doesn't hurt so I added it anyway.

pkg/streamingpromql/operators/aggregations/aggregation.go
zenador: If it's always false it fails some tests in name_label_dropping in upstream test cases, so it has to be passed in from the series. I've updated it to consider the subsequent series as well (not just the first), so dropName will be true if if it's true for any of the series in the group.

charleskorn: >If it's always false it fails some tests in name_label_dropping in upstream test cases

Do you have an example?

>I've updated it to consider the subsequent series as well (not just the first), so dropName will be true if if it's true for any of the series in the group.

Could you please add a test that covers this scenario?

pkg/streamingpromql/operators/deduplicate_and_merge_test.go
```diff
@@ -157,37 +160,99 @@ func TestDeduplicateAndMerge(t *testing.T) {
...
-			inner := &TestOperator{Series: testCase.inputSeries, Data: testCase.inputData, MemoryConsumptionTracker: memoryConsumptionTracker}
-			o := NewDeduplicateAndMerge(inner, memoryConsumptionTracker)
+			testDeduplicateAndMerge(t, false, testCase)
+		})
+	}
+}
+
+func TestDeduplicateAndMergeWithDelayedNameRemoval(t *testing.T) {
+	testCases := map[string]testCase{
+		"with dropName set": {
```
charleskorn: What about the case where some series have `DropName` set, but others don't?

pkg/streamingpromql/operators/deduplicate_and_merge_test.go
```diff
@@ -157,37 +160,99 @@ func TestDeduplicateAndMerge(t *testing.T) {
...
-				if err != nil {
-					break
-				}
+		if err != nil {
+			break
+		}
 
-				outputData = append(outputData, nextSeries)
-			}
+		outputData = append(outputData, nextSeries)
```
charleskorn: Should we also assert that all output series don't have `DropName` set?

pkg/streamingpromql/planning.go
```diff
@@ -203,6 +220,23 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
...
 }
 
+func shouldWrapInDedupAndMerge(root planning.Node) bool {
+	switch node := root.(type) {
+	case *core.NumberLiteral, *core.StringLiteral:
+		return false
+	case *core.BinaryExpression:
+		if node.VectorMatching == nil {
+			return false
+		}
```
charleskorn: Is this because we expect vector/scalar binary operations to already be wrapped in a deduplication and merge node?

pkg/streamingpromql/planning.go
```diff
@@ -203,6 +220,23 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
...
+	case *core.FunctionCall:
+		fn, exists := parser.Functions[node.Function.PromQLName()]
+		if exists && fn.ReturnType == parser.ValueTypeScalar {
+			return false
+		}
```
charleskorn: `planning.Node` has a `ResultType()` function that might help simplify this a bit

pkg/streamingpromql/operators/aggregations/aggregation.go
zenador: Added a new test case in ours (based on the failing upstream test) that will fail if `DropName` there is either true or false without considering the individual series.

Actually can't really think of any test case where dropName will differ within the group. Can you think of one? Perhaps it's theoretically not possible, but I'm not sure. We could perhaps remove the extra code that considers subsequent series, but it might be safer to leave it just in case there's something we can't think of that makes it possible.

pkg/streamingpromql/operators/deduplicate_and_merge_test.go
zenador: We compare the whole struct (which includes DropName). Added a function which ensures that DropName is included in the expected data as well.

pkg/streamingpromql/planning.go
zenador: Yes, it shouldn't reach this function as the top level is already dedup and merge, so it's really only necessary to check for scalar/scalar. But just to be safe I updated the logic here to account for that too.

pkg/streamingpromql/operators/aggregations/aggregation.go
charleskorn: >Actually can't really think of any test case where dropName will differ within the group. Can you think of one? Perhaps it's theoretically not possible, but I'm not sure. We could perhaps remove the extra code that considers subsequent series, but it might be safer to leave it just in case there's something we can't think of that makes it possible.

Perhaps an `or` where series on both sides share the same grouping label values, but one side is constructed so `DropName` is true and the other is false?

pkg/streamingpromql/planning.go
```diff
@@ -203,6 +220,28 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
...
+		resR, err := node.RHS.ResultType()
+		if err == nil && resR == parser.ValueTypeScalar {
+			return false
+		}
+	case *core.FunctionCall:
+		res, err := root.ResultType()
+		if err == nil && res == parser.ValueTypeScalar {
+			return false
+		}
```
charleskorn: Should we give up and return an error if these `ResultType` calls return an error?

pkg/streamingpromql/testutils/utils.go
```diff
@@ -231,6 +231,19 @@ func LabelsToSeriesMetadata(lbls []labels.Labels) []types.SeriesMetadata {
 	return m
 }
 
+func LabelsToSeriesMetadataWithDropName(lbls []labels.Labels, dropName []bool) []types.SeriesMetadata {
+	m := LabelsToSeriesMetadata(lbls)
+
+	if dropName == nil {
```
charleskorn, line 237: If `dropName` is `nil`, can't we return `m` as-is, given all the `DropName` values will default to false?

pkg/streamingpromql/operators/deduplicate_and_merge_test.go
```diff
@@ -153,20 +156,82 @@ func TestDeduplicateAndMerge(t *testing.T) {
...
+			expectedOutputSeries: []labels.Labels{
+				labels.FromStrings("foo", "1"),
+				labels.FromStrings(labels.MetricName, "metric", "foo", "2"),
+				labels.FromStrings("foo", "3"),
+			},
+			expectedOutputData: []types.InstantVectorSeriesData{
+				{Floats: []promql.FPoint{{T: 0, F: 0}, {T: 1, F: 1}}},
+				{Floats: []promql.FPoint{{T: 0, F: 10}, {T: 1, F: 11}}},
+				{Floats: []promql.FPoint{{T: 0, F: 20}, {T: 1, F: 21}}},
+			},
```
charleskorn, line 219: Might be good to be explicit here and show that we expect `DropName` to be false for all series.

pkg/streamingpromql/operators/aggregations/aggregation.go
zenador: Added test cases according to what we discussed on the call. We can't make it based on if the group by labels include `__name__` because that's not how the Prometheus logic works, as if we do `rate(B) or A` instead of `A or rate(B)` (see new test cases), the final name changes (dropName is applied in one but not the other). It seems instead of taking the first dropName seen, the last dropName applies instead, so I changed that logic accordingly.

However, the order in Prometheus isn't deterministic. So if you keep running those new tests in Prometheus, it should mostly pass, but sometimes they will fail, so this would be a new flaky test.

charleskorn: >However, the order in Prometheus isn't deterministic. So if you keep running those new tests in Prometheus, it should mostly pass, but sometimes they will fail, so this would be a new flaky test.

This sounds like a bug in Prometheus (as does the order of the `or` operands changing the output) - could you please raise it with them to see what they think?

zenador: Raised in https://github.com/prometheus/prometheus/pull/17161

In the meantime, should we just proceed first? I've commented it out for now so we don't add a new flaky test.

I don't think the order of the or operands changing the output is a bug, the following is deterministic and seems correct:
```
load 1m
  up{foo="bar"} 0+1x10
  down{foo="bar"} 0+2x10
  up{foo="bar2"} 0+3x10
  down{foo="bar2"} 0+4x10
  up{foo="bar3"} 0+5x10

eval instant at 10m up or down
  up{foo="bar"} 10
  up{foo="bar2"} 30
  up{foo="bar3"} 50

eval instant at 10m down or up
  down{foo="bar"} 20
  down{foo="bar2"} 40
  up{foo="bar3"} 50
```

charleskorn: Thanks for raising that - I've left a comment there about my order question.

pkg/streamingpromql/testdata/ours/name_label_dropping.test
```diff
@@ -95,8 +95,10 @@ eval instant at 10m sum by (__name__, foo) (down or rate(up[5m]))
   down{foo="bar2"} 40
   {foo="bar3"} 0.08333333333333334
 
-eval instant at 10m sum by (__name__) (up{foo="bar"} or rate(up{foo="bar2"}[5m]))
-  up 10.05
+# Temporarily comment out flaky tests.
```
charleskorn: Could you please add a link to https://github.com/prometheus/prometheus/pull/17161 here?

pkg/streamingpromql/operators/aggregations/aggregation.go
zenador: Revised that PR to a bug fix.

PR #12514, opened 2025-08-27
Fix ruler remotequerier request body consumption on retries

fionaliao:
#### What this PR does

PR #11833 introduced a bug where retried ruler remote querier requests were sent with empty bodies, meaning retries would never succeed and end up with `unknown position: parse error: no expression found in input` errors.

This is because the same `http.Request` is being reused for the retries. This uses a `bytes.Buffer` for the body:
https://github.com/grafana/mimir/blob/9b1e82284bd07b51040e51bbec36609819c7fef6/pkg/ruler/remotequerier.go#L163

The first request will read the entire buffer and get to the end, meaning retries would then send empty requests as they would only start reading from the end of the buffer. 

Previously we were using `httpgrpc.HTTPRequest` with a `[]byte` body that can be read multiple times, which is why the retries used to work:
https://github.com/grafana/mimir/blob/42099b0ea598e7e93d32deef3b7405e750ca3c16/pkg/ruler/remotequerier.go#L181-L192

The bug can be seen in the initial test commit: https://github.com/grafana/mimir/commit/382f66400fe94cbb212ffe4a53db0c3ac2a98bad

Fixed by recreating request on every retry.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 * [BUGFIX] Block-builder-scheduler: Fix a caching bug in initial job probing causing excessive memory usage at startup. #12389
 * [BUGFIX] Ruler: Support labels at the rule group level. These were previously ignored even when set via the API. #12397
 * [BUGFIX] Distributor: Fix metric metadata of type Unknown being silently dropped from RW2 requests. #12461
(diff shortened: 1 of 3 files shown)


56quarters (approved):
Nice fix!

PR #12522, opened 2025-08-27
Handle failed jobs in block-builder worker/client.

seizethedave:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -194,6 +194,11 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 		// Once we've gotten a job, we attempt to complete it even if the context is cancelled.
 		if _, err := b.consumeJob(context.WithoutCancel(ctx), key, spec); err != nil {
 			level.Error(b.logger).Log("msg", "failed to consume job", "job_id", key.Id, "epoch", key.Epoch, "err", err)
+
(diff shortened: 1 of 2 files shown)


pkg/blockbuilder/blockbuilder.go
```diff
@@ -194,6 +194,11 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 		// Once we've gotten a job, we attempt to complete it even if the context is cancelled.
 		if _, err := b.consumeJob(context.WithoutCancel(ctx), key, spec); err != nil {
 			level.Error(b.logger).Log("msg", "failed to consume job", "job_id", key.Id, "epoch", key.Epoch, "err", err)
+
+			if err := b.schedulerClient.FailJob(key); err != nil {
+				level.Error(b.logger).Log("msg", "failed to fail job", "job_id", key.Id, "epoch", key.Epoch, "err", err)
```
narqo, line 199: Should the block-builder continue in case it couldn't fail the assigned job? Maybe it's safer to fail the process altogether in this case?

(note, I see that, in practice, the `SchedulerClient` cannot fail because it only "forgets" the job in its local registry; so this comment isn't a huge deal)

seizethedave: I had initially placed a `panic` here. I'll put it back- if FailJob fails, it's a good reason to exit the process. Good thinking. :)

PR #12527, opened 2025-08-28
ci: bump kubernetes version in helm-ci

narqo:
#### What this PR does

The PR updates the Kubernetes version, that CI uses for the Helm chart tests from v1.20. ~~I picked v1.31 because it's the oldest version that is still supported.~~ v1.20 was released three years ago, and went out of support long since.

**Update** The CI uses kind v0.20, whose latest pre-build Kubernetes version was 1.29. I think it's good enough.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


.github/workflows/helm-ci.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       ct_configfile: operations/helm/ct.yaml
       ct_check_version_increment: false
       helm_version: v3.17.2
-      kind_kubectl_version: v1.20.8
-      kind_node_image: kindest/node:v1.20.15
+      kind_kubectl_version: v1.29.5
+      kind_node_image: kindest/node:v1.29.0
```
narqo, line 21: **Note:** `v1.29.0` is the latest node version, available in kind version, that the downstream workflow uses

dimitarvdimitrov (commented):
generally lgtm, but can you also modify where we mention what version of k8s we support that's mentioned in operations/helm/charts/mimir-distributed/README.md:23 and docs/sources/helm-charts/mimir-distributed/get-started-helm-charts/_index.md:32

narqo:
> modify where we mention what version of k8s

I've updated the chart's `Chart.yaml` to reflect the compatible version, and updated the docs. Also, added a changelog entry. PHAL

dimitarvdimitrov (approved):
LGTM, thanks

PR #12528, opened 2025-08-28
MQE: Skip DeduplicateAndMerge if possible

Konstantinov-Innokentii:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
+++ b/pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
@@ -1,0 +1,523 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package plan_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/streamingpromql"
+	"github.com/grafana/mimir/pkg/streamingpromql/optimize/plan"
+	"github.com/grafana/mimir/pkg/streamingpromql/planning"
+	"github.com/grafana/mimir/pkg/streamingpromql/planning/core"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func TestEliminateDeduplicateAndMergeOptimizationPass(t *testing.T) {
+	testCases := map[string]struct {
+		expr            string
+		expectedPlan    string
+		nodesEliminated int // Number of DeduplicateAndMerge nodes expected to be eliminated
+	}{
+		"function where selector has exact name matcher - should eliminate DeduplicateAndMerge": {
+			expr: `rate(foo[5m])`,
+			expectedPlan: `
+				- FunctionCall: rate(...)
+					- MatrixSelector: {__name__="foo"}[5m0s]
+			`,
+			nodesEliminated: 1,
+		},
+		"function where selector has no name matcher - should keep DeduplicateAndMerge": {
+			expr: `rate({job="test"}[5m])`,
+			expectedPlan: `
+				- DeduplicateAndMerge
+					- FunctionCall: rate(...)
+						- MatrixSelector: {job="test"}[5m0s]
+			`,
+			nodesEliminated: 0,
+		},
+		"function where selector has regex name matcher - should keep DeduplicateAndMerge": {
+			expr: `rate({__name__=~"(foo|bar)"}[5m])`,
+			expectedPlan: `
+				- DeduplicateAndMerge
+					- FunctionCall: rate(...)
+						- MatrixSelector: {__name__=~"(foo|bar)"}[5m0s]
+			`,
+			nodesEliminated: 0,
+		},
+		"nested functions where selector has exact name matcher - should eliminate all DeduplicateAndMerge": {
+			expr: `abs(rate(foo[5m]))`,
+			expectedPlan: `	
+				- FunctionCall: abs(...)
+					- FunctionCall: rate(...)
+						- MatrixSelector: {__name__="foo"}[5m0s]
+			`,
+			nodesEliminated: 2,
+		},
+		"nested functions where selector has no name matcher - should keep only inner DeduplicateAndMerge": {
+			expr: `abs(rate({job="test"}[5m]))`,
+			expectedPlan: `
+				- FunctionCall: abs(...)
+					- DeduplicateAndMerge
+						- FunctionCall: rate(...)
+							- MatrixSelector: {job="test"}[5m0s]
+			`,
+			nodesEliminated: 1,
+		},
+		"nested functions where selector has regex name matcher - should keep only inner DeduplicateAndMerge": {
+			expr: `abs(rate({__name__=~"(foo|bar)"}[5m]))`,
+			expectedPlan: `
+				- FunctionCall: abs(...)
+					- DeduplicateAndMerge
+						- FunctionCall: rate(...)
+							- MatrixSelector: {__name__=~"(foo|bar)"}[5m0s]
+			`,
+			nodesEliminated: 1,
+		},
+		"deeply nested functions where selector has no name matcher - should keep only innermost DeduplicateAndMerge": {
+			expr: `abs(ceil(rate({job="test"}[5m])))`,
+			expectedPlan: `
+				- FunctionCall: abs(...)
+					- FunctionCall: ceil(...)
+						- DeduplicateAndMerge
+							- FunctionCall: rate(...)
+								- MatrixSelector: {job="test"}[5m0s]
+			`,
+			nodesEliminated: 2,
+		},
+		"vector-scalar operation where vector side has exact name matcher - should eliminate DeduplicateAndMerge": {
+			expr: `2 * foo`,
+			expectedPlan: `
+				- BinaryExpression: LHS * RHS
+					- LHS: NumberLiteral: 2
+					- RHS: VectorSelector: {__name__="foo"}
+			`,
+			nodesEliminated: 1,
+		},
+		"vector-scalar operation where vector side has no name matcher - should keep DeduplicateAndMerge": {
+			expr: `2 * {job="test"}`,
+			expectedPlan: `
+				- DeduplicateAndMerge
+					- BinaryExpression: LHS * RHS
+						- LHS: NumberLiteral: 2
+						- RHS: VectorSelector: {job="test"}
+			`,
+			nodesEliminated: 0,
+		},
+		"vector-scalar operation where vector side has regex name matcher - should keep DeduplicateAndMerge": {
+			expr: `2 * {__name__=~"(foo|bar)"}`,
+			expectedPlan: `
+				- DeduplicateAndMerge
+					- BinaryExpression: LHS * RHS
+						- LHS: NumberLiteral: 2
+						- RHS: VectorSelector: {__name__=~"(foo|bar)"}
+			`,
+			nodesEliminated: 0,
+		},
+		"vector-scalar operation where vector side is a function overs selector with exact name matcher - eliminates all DeduplicateAndMerge": {
+			expr: `2 * rate(foo[5m])`,
+			expectedPlan: `
+				- BinaryExpression: LHS * RHS
+					- LHS: NumberLiteral: 2
+					- RHS: FunctionCall: rate(...)
+						- MatrixSelector: {__name__="foo"}[5m0s]
+			`,
+			nodesEliminated: 2,
+		},
+		"vector-scalar operation where vector side is a function over selector with no name matcher - keeps only inner DeduplicateAndMerge": {
+			expr: `2 * rate({job="test"}[5m])`,
+			expectedPlan: `
+				- BinaryExpression: LHS * RHS
+					- LHS: NumberLiteral: 2
+					- RHS: DeduplicateAndMerge
+						- FunctionCall: rate(...)
+							- MatrixSelector: {job="test"}[5m0s]
+			`,
+			nodesEliminated: 1,
+		},
+		"nested vector-scalar operation where inner operation has selector with exact name matcher - eliminates all DeduplicateAndMerge": {
+			expr: `2 * (3 + foo)`,
+			expectedPlan: `
+				- BinaryExpression: LHS * RHS
+					- LHS: NumberLiteral: 2
+					- RHS: BinaryExpression: LHS + RHS
+						- LHS: NumberLiteral: 3
+						- RHS: VectorSelector: {__name__="foo"}
+			`,
+			nodesEliminated: 2,
+		},
+		"nested vector-scalar operation where inner operation has selector with no name matcher - should keep only inner DeduplicateAndMerge": {
+			expr: `2 * (3 + {job="test"})`,
+			expectedPlan: `
+				- BinaryExpression: LHS * RHS
+					- LHS: NumberLiteral: 2
+					- RHS: DeduplicateAndMerge
+						- BinaryExpression: LHS + RHS
+							- LHS: NumberLiteral: 3
+							- RHS: VectorSelector: {job="test"}
+			`,
+			nodesEliminated: 1,
+		},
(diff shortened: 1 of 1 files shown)


pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+	"github.com/grafana/mimir/pkg/streamingpromql"
+	"github.com/grafana/mimir/pkg/streamingpromql/optimize/plan"
+	"github.com/grafana/mimir/pkg/streamingpromql/planning"
+	"github.com/grafana/mimir/pkg/streamingpromql/planning/core"
+	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func TestEliminateDeduplicateAndMergeOptimizationPass(t *testing.T) {
```
charleskorn: Could you please add some tests for cases where we need to retain the `DeduplicateAndMerge` for cases with `or`, as well as vector/scalar comparisons with `bool` (eg. `metric == bool 6`)?

It would also be good to see tests for binary operations, to confirm that we're skipping them correctly.

pkg/streamingpromql/planning.go
```diff
@@ -78,6 +78,11 @@ func NewQueryPlanner(opts EngineOpts) (*QueryPlanner, error) {
 		planner.RegisterQueryPlanOptimizationPass(plan.NewNarrowSelectorsOptimizationPass(opts.Logger))
 	}
 
+	// EliminateDeduplicateAndMergeO doesn't support delayed name removal yet.
```
```suggestion
	// EliminateDeduplicateAndMerge doesn't support delayed name removal yet.
```

pkg/streamingpromql/planning.go
```diff
@@ -78,6 +78,11 @@ func NewQueryPlanner(opts EngineOpts) (*QueryPlanner, error) {
...
+	if !opts.CommonOpts.EnableDelayedNameRemoval {
+		planner.RegisterQueryPlanOptimizationPass(plan.NewEliminateDeduplicateAndMergeOptimizationPass())
```
charleskorn: Should we run this before CSE? Then we don't have to think about how to handle deduplicated subexpressions in this optimisation pass.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
```diff
@@ -0,0 +1,154 @@
...
+	if newRoot != nil {
+		plan.Root = newRoot
+	}
+	return plan, nil
+}
+
+func (e *EliminateDeduplicateAndMergeOptimizationPass) collectNodesToRemove(node planning.Node, parent planning.Node, childIndex int, nodesToRemove *[]dedupNodeInfo) {
+	// Binary operations are not supported yet.
+	// 1. It makes elimination logic much more complex - it's needed to track if name was dropped on both sides of the binary operation.
+	// 2. We likely can't do label projections when binary operation is involved, becaue of conflict error messages.
```
charleskorn: This isn't 100% correct: if the query contains aggregations within binary operations, for example `sum(rate(foo[5m])) / sum(rate(bar[5m]))`, then we can perform projections.

I think it's straightforward to remove `DeduplicateAndMerge` nodes provided they don't contain binary operations. For example, in `rate(foo[5m]) / rate(bar[5m])`, we can remove the `DeduplicateAndMerge` nodes around the `rate()` calls.

I also think we can remove `DeduplicateAndMerge` nodes that contain binary operations, but we'd need to inspect the binary operation to confirm if it can produce series with a `__name__` label, which sounds complex, as you say - would be good to write this down somewhere as a future improvement.

pkg/streamingpromql/planning.go
Konstantinov-Innokentii: What's special about deduplicated subexpressions in this pass? I mean, isn't CSE just replaces subexpression with DuplicateConsumer, and that shouldn't affect the Deduplication?

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
Konstantinov-Innokentii: makes sense

pkg/streamingpromql/planning.go
charleskorn: It may make the logic in `EliminateDeduplicateAndMergeOptimizationPass` more complex, as before CSE, the query plan is a tree, but afterwards, it may no longer be a tree due to multiple paths culminating in the same `Duplicate` node.

Konstantinov-Innokentii: Done, thanks for catching this.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
Konstantinov-Innokentii: Addded

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
Konstantinov-Innokentii: So, I propose to keep it as it is and add support for `rate(foo[5m]) / rate(bar[5m])` expressions in the follow-up PR, just to push this over the line.

pkg/streamingpromql/planning.go
```diff
@@ -65,6 +65,14 @@ func NewQueryPlanner(opts EngineOpts) (*QueryPlanner, error) {
 	planner.RegisterASTOptimizationPass(&ast.SortLabelsAndMatchers{}) // This is a prerequisite for other optimization passes such as common subexpression elimination.
 	// After query sharding is moved here, we want to move propagate matchers and reorder histogram aggregation here as well before query sharding.
 
+	// EliminateDeduplicateAndMerge doesn't support delayed name removal yet.
+	// This optimization pass is registered before CSE to keep the query plan as a simple tree structure.
+	// After CSE, the query plan may no longer be a tree due to multiple paths culminating in the same Duplicate node,
+	// which would make the elimination logic more complex.
+	if !opts.CommonOpts.EnableDelayedNameRemoval {
```
charleskorn: It'd be good to add a feature flag for this optimisation pass as well, so we can disable it easily if we discover any problems.

Konstantinov-Innokentii: Done

Konstantinov-Innokentii:
Please, do not merge it yet, I think I found some corner cases.

Konstantinov-Innokentii:
> Please, do not merge it yet, I think I found some corner cases.

So, my concern was that we are running upstream test cases with delayed name removal enabled, which lowers test coverage for the optimization, since it's runs only when delayed name removal is disabled :)
I addressed it in this commit, basically just adding more tests - https://github.com/grafana/mimir/pull/12528/commits/b1aa6fc6e15e418fc1243d9cd98a600c2c2f3a80.

cmd/mimir/config-descriptor.json
```diff
@@ -2758,6 +2758,17 @@
               "fieldFlag": "querier.mimir-query-engine.enable-narrow-binary-selectors",
               "fieldType": "boolean",
               "fieldCategory": "experimental"
+            },
+            {
+              "kind": "field",
+              "name": "enable_eliminate_deduplicate_and_merge",
+              "required": false,
+              "desc": "Enable eliminating redundant DeduplicateAndMerge nodes from the query plan when it can be proven that each input series will produce a unique output series.",
```
```suggestion
              "desc": "Enable eliminating redundant DeduplicateAndMerge nodes from the query plan when it can be proven that each input series produces a unique output series.",
```

operations/mimir/mimir-flags-defaults.json
charleskorn, line 1: Should `querier.mimir-query-engine.enable-eliminate-deduplicate-and-merge` be added here as well?

pkg/streamingpromql/planning.go
```diff
@@ -65,6 +65,14 @@ func NewQueryPlanner(opts EngineOpts) (*QueryPlanner, error) {
...
 	// After query sharding is moved here, we want to move propagate matchers and reorder histogram aggregation here as well before query sharding.
 
+	// EliminateDeduplicateAndMerge doesn't support delayed name removal yet.
+	// This optimization pass is registered before CSE to keep the query plan as a simple tree structure.
+	// After CSE, the query plan may no longer be a tree due to multiple paths culminating in the same Duplicate node,
+	// which would make the elimination logic more complex.
+	if !opts.CommonOpts.EnableDelayedNameRemoval && opts.EnableEliminateDeduplicateAndMerge {
+		planner.RegisterQueryPlanOptimizationPass(plan.NewEliminateDeduplicateAndMergeOptimizationPass())
+	}
+
```
charleskorn: Rather than silently ignore the invalid configuration, I would return an error:

```suggestion
	// This optimization pass is registered before CSE to keep the query plan as a simple tree structure.
	// After CSE, the query plan may no longer be a tree due to multiple paths culminating in the same Duplicate node,
	// which would make the elimination logic more complex.
	if opts.EnableEliminateDeduplicateAndMerge {
		if opts.CommonOpts.EnableDelayedNameRemoval {
			return nil, errors.New("eliminating deduplicate and merge nodes is not supported with delayed name removal")
		}
	
		planner.RegisterQueryPlanOptimizationPass(plan.NewEliminateDeduplicateAndMergeOptimizationPass())
	}
```

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -0,0 +1,543 @@
...
+	}{
+		"series with different labelset after name drop": {
+			data: `
+				load 1m
+					a{env="prod"} 0+1x10
+					a{env="test"} 0+2x10
+			`,
+			expr: `abs(rate(a[5m]))`,
+			expectedResult: `{env="prod"} => 0.016666666666666666 @[600000]
+{env="test"} => 0.03333333333333333 @[600000]`,
```
charleskorn: [nit] It might be easier to read `expectedResult` with slightly different formatting: (the extra whitespace can be removed with `testutils.TrimIndent` like we do for the query plan output in the tests above)

```suggestion
			expectedResult: `
				{env="prod"} => 0.016666666666666666 @[600000]
				{env="test"} => 0.03333333333333333 @[600000]
			`,
```

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -0,0 +1,543 @@
...
+	if _, ok := node.(*core.DeduplicateAndMerge); ok {
+		count = 1
+	}
+	for _, child := range node.Children() {
+		count += countDeduplicateAndMergeNodes(child)
+	}
+	return count
+}
+
+func TestEliminateDeduplicateAndMergeOptimizationPass(t *testing.T) {
```
charleskorn: [nit]

```suggestion
func TestEliminateDeduplicateAndMergeOptimizationPassCorrectness(t *testing.T) {
```

charleskorn: One other idea for this: you could run all of our existing test cases (including the upstream ones) with the optimisation pass enabled to confirm there are no cases where this introduces an issue.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
Konstantinov-Innokentii: Thanks!

operations/mimir/mimir-flags-defaults.json
Konstantinov-Innokentii: You mean in config-descriptor.json? It's already there

charleskorn: No, I mean in this file, `mimir-flags-defaults.json`.

Konstantinov-Innokentii: It seems only flag which are true by default are there for some reason. For example, there is no "enable-narrow-binary-selectors" there,

charleskorn: Right, makes sense. But why did `ruler.evaluation-consistency-max-delay` need to be added as part of this PR?

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
Konstantinov-Innokentii: Added TestEliminateDeduplicateAndMergeOptimizationDoesNotRegress, which runs upstream/ours test cases with delayed name removal disabled.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -0,0 +1,603 @@
...
+func TestEliminateDeduplicateAndMergeOptimizationDoesNotRegress(t *testing.T) {
+	runTestCasesWithDelayedNameRemovalDisabled(t, "upstream/*.test")
+	runTestCasesWithDelayedNameRemovalDisabled(t, "ours*/*.test")
+}
+
+func runTestCasesWithDelayedNameRemovalDisabled(t *testing.T, globPattern string) {
+
+	types.EnableManglingReturnedSlices = true
+	parser.ExperimentalDurationExpr = true
+	parser.EnableExperimentalFunctions = true
```
charleskorn, line 574: We should restore these to their previous values after the test finishes, to ensure this doesn't affect any other tests.

operations/mimir/mimir-flags-defaults.json
Konstantinov-Innokentii: It's generated automatically when you run `make reference-help` and it seems it was not added for some reason in the [PR](https://github.com/grafana/mimir/pull/12751) which added this setting

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
Konstantinov-Innokentii: Makes sense. But in the engine_test.go these are set in the init and not resetted later, should we fix it?
```
func init() {
	types.EnableManglingReturnedSlices = true
	parser.ExperimentalDurationExpr = true
	parser.EnableExperimentalFunctions = true

	// Set a tracer provider with in memory span exporter so we can check the spans later.
	otel.SetTracerProvider(
		tracesdk.NewTracerProvider(
			tracesdk.WithSpanProcessor(tracesdk.NewSimpleSpanProcessor(spanExporter)),
		),
	)
}
```

cmd/mimir/config-descriptor.json
Konstantinov-Innokentii: Thanks!

operations/mimir/mimir-flags-defaults.json
Konstantinov-Innokentii: Actually it was already added to main, so this diff is a github glitch - https://github.com/grafana/mimir/commit/23b6a82e7e1ad2baeadcff0cdb9324f395216794#diff-572ec83221bd126ed9d442b6f3b75989c37772084f29bde1134cf58b483e66d4R383

Konstantinov-Innokentii:
@charleskorn Is there anything you want me to adress before merging?

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
charleskorn: For `types.EnableManglingReturnedSlices = true`: we should probably set this in `init()` for each test package beneath `pkg/streamingpromql`.

For the other two: I think we should just enable and restore them in the places that need them (otherwise this might affect other tests).

Konstantinov-Innokentii: Ok, I'll address it in the sepate PR

charleskorn:
> @charleskorn Is there anything you want me to adress before merging?

LGTM, happy for the remaining comment thread to be resolved in a follow-up PR

PR #12529, opened 2025-08-28
Refactor query-frontend round-tripper creation

charleskorn:
#### What this PR does

This PR refactors how the query-frontend round-tripper is created.

This is to allow substituting in an alternative to `httpQueryRequestRoundTripperHandler` that performs query planning and evaluates the query with MQE. (This will be added in a subsequent PR.)

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/limits.go
+++ b/pkg/frontend/querymiddleware/limits.go
@@ -213,9 +213,9 @@ type limitedParallelismRoundTripper struct {
 }
 
 // NewLimitedParallelismRoundTripper creates a new roundtripper that enforces MaxQueryParallelism to the `next` roundtripper across `middlewares`.
-func NewLimitedParallelismRoundTripper(next http.RoundTripper, codec Codec, limits Limits, logger log.Logger, middlewares ...MetricsQueryMiddleware) http.RoundTripper {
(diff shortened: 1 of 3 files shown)


pkg/frontend/querymiddleware/limits.go
```diff
@@ -270,17 +267,25 @@ func (rt limitedParallelismRoundTripper) RoundTrip(r *http.Request) (*http.Respo
 	return rt.codec.EncodeMetricsQueryResponse(ctx, r, response)
 }
 
-// roundTripperHandler is an adapter that implements the MetricsQueryHandler interface using a http.RoundTripper to perform
+func NewHTTPQueryRequestRoundTripperHandler(next http.RoundTripper, codec Codec, logger log.Logger) MetricsQueryHandler {
```
aknuds1, line 270: [Nit] I don't see the need to export this function:
```suggestion
func newHTTPQueryRequestRoundTripperHandler(next http.RoundTripper, codec Codec, logger log.Logger) MetricsQueryHandler {
```

charleskorn: This function must be exported for use in our closed source code for GEM.

PR #12530, opened 2025-08-28
Pull statistics into planner

chencs:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -982,6 +982,19 @@ func (i *Ingester) updateLimitMetrics() {
 	}
 }
 
+func (i *Ingester) UserTSDBStatistics(ctx context.Context) (index.Statistics, error) {
+	user, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error extracting tenant ID from context: %w", err)
+	}
+	db := i.getTSDB(user)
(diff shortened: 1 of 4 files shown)


pkg/ingester/ingester.go
```diff
@@ -982,6 +982,19 @@ func (i *Ingester) updateLimitMetrics() {
 	}
 }
 
+func (i *Ingester) UserTSDBStatistics(ctx context.Context) (index.Statistics, error) {
+	user, err := tenant.TenantID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error extracting tenant ID from context: %w", err)
+	}
+	db := i.getTSDB(user)
+	// getTSDB may return nil if the userID isn't found in the ingester's tsdbs.
```
dimitarvdimitrov: i checked the code and we shouldn't actually hit this path. if we can't find a DB, then the ingester short-circuits and doesn't invoke any querier or planner 

https://github.com/grafana/mimir/blob/5be3127fa87dfb8ff81cd6c01be44df463b785b0/pkg/ingester/ingester.go#L2288-L2291

i think you can drop this comment or replace it with an explanation that we shouldn't be running lookup planning if there is no TSDB.

dimitarvdimitrov (approved):

PR #12531, opened 2025-08-28
development: Add service dependencies to NGINX in microservices mode

dimitarvdimitrov:
Make NGINX depend on all services it routes to (distributor-1, alertmanager-1, ruler-1, query-frontend, compactor, grafana) so that when these services restart, NGINX startup ordering is properly enforced.

Also fix COMPACTOR_HOST port from incorrect 8007 to correct 8006.


narqo (approved):
🔥

PR #12532, opened 2025-08-28
parquet: Recover benchmark work

jesusvazquez:
This PR recovers most of whats available in https://github.com/grafana/mimir/tree/njpm/parquet-benchmarks but updated to work with the current version of the code.

Benchmark parquet performance with 
```
go test ./pkg/storegateway/parquetbench -bench=. -benchmark-store=parquet -benchmark-sort-by="__name__,zone" -benchtime=2s
```

Note there are a few flags available:
* `-benchmark-store=` with options `tsdb` (for the existing store-gateways performance) or `parquet`
* `-benchmark-sort-by="__name__,zone"` which modifies converter behavior to sort by extra columns. If not provided it sorts by `__name__`.
* `-benchmark-compression=true` to play with converter conversion. By default its true

Run same benchmark against existing store-gateway bucket stores implementation with
```
go test ./pkg/storegateway/parquetbench -bench=. -benchmark-store=tsdb -benchtime=2s
```

And then compare results


pkg/storegateway/bucket.go
```diff
@@ -971,7 +971,19 @@ func (s *BucketStore) sendMessage(typ string, srv storegatewaypb.StoreGateway_Se
 	pmsg := &grpc.PreparedMsg{}
 	err := pmsg.Encode(srv, msg)
 	*encodeDuration += time.Since(encodeBegin)
+
+	// If PreparedMsg.Encode fails (e.g., in tests with mock servers that don't have gRPC rpcInfo),
+	// fall back to regular Send method
 	if err != nil {
+		if resp, ok := msg.(*storepb.SeriesResponse); ok {
```
jesusvazquez, line 978: This is needed for our benchmark because we mock the server

npazosmendez (approved):

PR #12541, opened 2025-08-28
mimirpb: Preserve inconsistent metric metadata in RW1 to 2 conversion

alexweav:
#### What this PR does

Occasionally RW1.0 requests will contain inconsistent metadata. That is, they might contain multiple metadatas for the same series that don't agree with each other, like a new and old description simultaneously. The prometheus behavior when accepting RW1.0 is to not choose a winner; both metadatas are stored and will appear in the metadata API until they both age out.

RW2.0 tends to prevent this by attaching metadata to series. However, for requests that were converted from 1.0 to 2.0 transparently, we want to preserve the original behavior and ensure both metadata make it through the system.

This PR adjusts the RW1.0 to 2.0 conversion to only deduplicate metadata whose fields are exactly equal during the timeseries matching process. The previous implementation would discard inconsistent metadata beyond the first seen one. If we see multiple metadata that exactly match, we save space and deduplicate it; if we see a metadata that is inconsistent, we fabricate an extra timeseries and send both versions.

No changelog as it only affects record format V2, which is not marked stable.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/prealloc_rw2.go
+++ b/pkg/mimirpb/prealloc_rw2.go
@@ -36,10 +36,25 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
 	symbols.ConfigureCommonSymbols(offset, commonSymbols)
 
 	var metadataMap map[string]*MetricMetadata
+	var extraMetadata []*MetricMetadata
 	if len(rw1.Metadata) > 0 && len(rw1.Timeseries) > 0 {
 		metadataMap = make(map[string]*MetricMetadata, len(rw1.Metadata))
+		// Very rarely, a request might send multiple conflicting metadatas for the same metric.
+		// For example, in Prometheus this can happen on the border of a metric churning,
+		// if new and old versions of the same metadata are scraped from different processes and end up in the same batch.
+		//
+		// The Prometheus behavior for RW1.0 is to store *both* metadatas and not take any preference.
+		// extraMetadata is used to carry duplicates of existing metadata entries that don't match, so we can propagate them at the end.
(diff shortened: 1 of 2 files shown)


pkg/mimirpb/prealloc_rw2.go
```diff
@@ -35,11 +35,28 @@ func FromWriteRequestToRW2Request(rw1 *WriteRequest, commonSymbols []string, off
...
-	var metadataMap map[string]*MetricMetadata
+	// If our request has both metadata and timeseries, we can try to match.
+	var metadataMap map[string]int // Metric name -> index of the first metadata, in the Metadata slice.
+	var matchedMetadataMask []bool
 	if len(rw1.Metadata) > 0 && len(rw1.Timeseries) > 0 {
-		metadataMap = make(map[string]*MetricMetadata, len(rw1.Metadata))
-		for _, meta := range rw1.Metadata {
-			metadataMap[meta.MetricFamilyName] = meta
+		metadataMap = make(map[string]int, len(rw1.Metadata))
+		matchedMetadataMask = make([]bool, len(rw1.Metadata))
```
alexweav, line 43: We can't use absence of a key in the map to track if something's been matched anymore. Now we might have multiple metadatas for the same key.

A one-time, bounded allocation of this mask seemed preferable to `O(n)` allocations if we, for instance, wrap the map values with something tracking indexes, or `O(log(n))` allocations if we use a map of slices. We also can get rid of the map `delete()` later on which saves a bit of CPU.

This slice can be trivially pooled as well if we want. Super safe, no pointers or usable data stored inside

PR #12551, opened 2025-08-29
MQE: perform query planning in query-frontends and send to queriers

charleskorn:
#### What this PR does

This PR introduces experimental support for performing query planning in query-frontends, building on the querier changes in https://github.com/grafana/mimir/pull/12302.

When enabled, query-frontends compute a query plan for each query and send it to queriers for evaluation.

The key parts of this are:

* When enabled, query-frontends use the newly-introduced `engineQueryRequestRoundTripperHandler` instead of the `httpQueryRequestRoundTripperHandler` to handle instant and range queries
* `engineQueryRequestRoundTripperHandler` is responsible for taking a query request and passing it to the query-frontend's query engine for evaluation
* The query-frontend's query planner is configured with the newly-introduced `remoteexec.OptimizationPass` optimization pass that injects a `RemoteExecution` node at the root of the query plan if the query contains selectors but does not contain sharded or spun-off expressions
  * Keep in mind that the query-frontend's engine instance is used in two contexts now: once while evaluating the overall sharded query in the sharding middleware (if the query is shardable) and then while evaluating the overall query (if the query isn't shardable) or each individual leg (if it is shardable) in the roundtripper.
  * This is why we don't insert `RemoteExecution` nodes in query plans containing sharded expressions, as we don't want remote execution to run during the evaluation initiated by the sharding middleware.
  * This is a temporary thing until sharding is running within MQE.
  * Similar logic applies to subquery spin-off as well.
* The `RemoteExecution` nodes are materialized as `remoteexec.InstantVectorOperator`, `remoteexec.RangeVectorOperator` or `remoteexec.ScalarOperator` operators.
* These operators use the frontend's `v2.RemoteExecutor` to submit requests for evaluation which are then forwarded to query-schedulers and then queriers for evaluation.

Subsequent PRs will introduce improvements or missing functionality, such as:
* propagating strong consistency information to queriers
* sending query component flags (ingester-only, store-gateway-only or both) to query-schedulers
* propagating query stats to query-frontends
* possibly batching up results from queriers

Out of scope for this PR, but will be introduced at some point in the future:
* performing query sharding as part of the query plan
* performing subquery spin-off as part of the query plan
* performing range query splitting and caching as part of the query plan

In each of the cases above, the individual sharded legs / cacheable extents are independently planned and evaluated.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/config.go
+++ b/pkg/frontend/config.go
@@ -57,6 +57,11 @@ func (cfg *CombinedFrontendConfig) Validate() error {
 	if err := cfg.QueryMiddleware.Validate(); err != nil {
 		return err
 	}
+
+	if cfg.EnableRemoteExecution && cfg.QueryEngine != querier.MimirEngine {
+		return errors.New("remote execution is only supported when the Mimir query engine is in use")
+	}
+
 	return nil
 }
 
--- a/pkg/frontend/config_test.go
+++ b/pkg/frontend/config_test.go
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package frontend
+
+import (
+	"testing"
+
+	"github.com/grafana/dskit/flagext"
+	"github.com/stretchr/testify/require"
+
+	"github.com/grafana/mimir/pkg/querier"
+)
+
+func TestConfig_Validate(t *testing.T) {
+	testCases := map[string]struct {
+		enableRemoteExecution bool
+		queryEngine           string
+		expectedError         string
+	}{
+		"remote execution disabled, MQE enabled": {
+			enableRemoteExecution: false,
+			queryEngine:           querier.MimirEngine,
+			expectedError:         "",
+		},
+		"remote execution disabled, MQE disabled": {
+			enableRemoteExecution: false,
+			queryEngine:           querier.PrometheusEngine,
+			expectedError:         "",
+		},
+		"remote execution enabled, MQE enabled": {
+			enableRemoteExecution: true,
+			queryEngine:           querier.MimirEngine,
+			expectedError:         "",
+		},
+		"remote execution enabled, MQE disabled": {
+			enableRemoteExecution: true,
+			queryEngine:           querier.PrometheusEngine,
+			expectedError:         "remote execution is only supported when the Mimir query engine is in use",
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			config := &CombinedFrontendConfig{}
+			flagext.DefaultValues(config)
+			config.EnableRemoteExecution = testCase.enableRemoteExecution
+			config.QueryEngine = testCase.queryEngine
+			err := config.Validate()
+
+			if testCase.expectedError == "" {
+				require.NoError(t, err)
+			} else {
+				require.EqualError(t, err, testCase.expectedError)
+			}
+		})
+	}
+}


pkg/frontend/querymiddleware/roundtrip.go
```diff
@@ -288,7 +293,13 @@ func newQueryTripperware(
 		// It means that the first roundtrippers defined in this function will be the last to be
 		// executed.
 
-		queryHandler := NewHTTPQueryRequestRoundTripperHandler(next, codec, log)
+		var queryHandler MetricsQueryHandler
+
+		if useRemoteExecution {
+			queryHandler = NewEngineQueryRequestRoundTripperHandler(streamingEngine, log)
```
56quarters, line 299: ~I don't quite follow this part. Even if remote execution is enabled, we still can't use it for queries that have been sharded. Don't we need to fallback to an existing roundtripper in that case?~

Nevermind, IIUC this will never be used when a query is sharded since the sharding middleware will use its own roundtripper.

56quarters (commented):
This seems like a good approach to me.

pkg/frontend/querymiddleware/roundtrip.go
charleskorn: This will still be used if the query is sharded. If the query is sharded, the individual sharded legs will be executed by this roundtripper.

integration/query_frontend_test.go
```diff
@@ -394,23 +437,47 @@ func runQueryFrontendTest(t *testing.T, cfg queryFrontendTestConfig) {
 	wg.Wait()
 
 	// Compute the expected number of queries.
-	expectedQueriesCount := float64(numUsers*numQueriesPerUser) + 4
+	simpleQueryCount := float64(numUsers * numQueriesPerUser)
```
56quarters, line 440: I know you didn't start this but I really don't like integration tests that test that some internal thing happened a certain number of times. This feels like its verifying the implementation of something rather than the results. If the implementation changes and the number of some type of query changes, debugging this requires being able to keep the whole read path in your head.

Is there a way we can assert that remote execution was used for a query without going this deep into the implementation?

pkg/api/error/error.go
```diff
@@ -31,6 +33,41 @@ const (
 	TypeNotAcceptable   Type = "not_acceptable"
 )
 
+// TypeForError returns the appropriate Type for err, or fallback if no appropriate Type
+// can be inferred.
+func TypeForError(err error, fallback Type) Type {
```
56quarters, line 38: This is pretty similar to the logic of `mapEngineError()` in `querysharding.go`. Is there anything we should combine here?

pkg/frontend/v2/frontend.go
```diff
@@ -223,37 +241,49 @@ func (f *Frontend) stopping(_ error) error {
...
-		response: make(chan queryResultWithBody, 1),
+		enqueue: make(chan enqueueResult, 1),
+
+		spanLogger: spanlogger.FromContext(ctx, f.log),
+	}
+
+	return freq, ctx, cancel, nil
+}
+
+// RoundTripGRPC round trips a proto (instead of an HTTP request).
```
56quarters: Not really your changes but could you clarify this comment because we're round tripping a protobuf message of an HTTP request compared to an normal HTTP request or normal protobuf message.

pkg/frontend/v2/frontend.go
```diff
@@ -309,16 +339,178 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, httpRequest *httpgrpc.HTTP
 	}
 }
 
-func (f *Frontend) enqueueRequestWithRetries(ctx context.Context, freq *frontendRequest, spanLogger *spanlogger.SpanLogger) (chan<- uint64, error) {
+// DoProtobufRequest initiates a Protobuf request to queriers.
+// If the returned error is nil, then callers must either Close the returned stream
+// or cancel ctx to ensure resources are not leaked.
+func (f *Frontend) DoProtobufRequest(ctx context.Context, req proto.Message) (*ProtobufResponseStream, error) {
+	logger, ctx := spanlogger.New(ctx, f.log, tracer, "DoProtobufRequest")
```
56quarters:
```suggestion
	logger, ctx := spanlogger.New(ctx, f.log, tracer, "frontend.DoProtobufRequest")
```

pkg/frontend/v2/frontend.go
```diff
@@ -127,16 +130,21 @@ type queryResultWithBody struct {
...
 
-	ctx context.Context
+	ctx        context.Context
+	spanLogger *spanlogger.SpanLogger
 
-	enqueue  chan enqueueResult
-	response chan queryResultWithBody
+	enqueue chan enqueueResult
+
+	httpRequest  *httpgrpc.HTTPRequest
```
56quarters, line 143: Can you comment which of these fields are mutually exclusive?

pkg/frontend/v2/frontend.go
```diff
@@ -309,16 +339,178 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, httpRequest *httpgrpc.HTTP
...
+	}()
+
+	return freq.protobufResponseStream, nil
+}
+
+type ProtobufResponseStream struct {
+	// Why do we have two channels here?
+	// Different goroutines write to each, and each needs to close its corresponding channel when finished.
+	// There's no guarantee which order the goroutines finish in, and one may never be called at all.
+	c            chan protobufResponseMessage
```
56quarters: Can we use a more descriptive name for this?

pkg/frontend/v2/frontend.go
```diff
@@ -404,10 +634,28 @@ func (f *Frontend) QueryResultStream(stream frontendv2pb.FrontendForQuerier_Quer
...
 	switch d := firstMessage.Data.(type) {
 	case *frontendv2pb.QueryResultStreamRequest_Metadata:
-		return f.receiveResultForHTTPRequest(userID, firstMessage, d, stream)
+		if req.httpResponse == nil {
+			return errUnexpectedHTTPResponse
+		}
+
+		return f.receiveResultForHTTPRequest(req, firstMessage, d, stream)
 	default:
+		if req.protobufResponseStream != nil {
```
56quarters: WDYT about swapping this to match the conditional above so they're both checking for `nil` and returning an error (or vice versa)?

integration/query_frontend_test.go
charleskorn: There are two parts to my changes here:

The [first part](https://github.com/grafana/mimir/blob/e05439884bafe0053809b8ae6f751934a8201448/integration/query_frontend_test.go#L442-L473) was the discovery that this test wasn't testing what it was meant to (it was counting requests to endpoints like `/metrics` and `/ready`). If you have any suggestions on how to improve this, I'd love to hear them, I don't love this either.

The [second part](https://github.com/grafana/mimir/blob/e05439884bafe0053809b8ae6f751934a8201448/integration/query_frontend_test.go#L475-L480) was the best way I could find to achieve what you're describing, which is ensuring that remote execution was used.

pkg/frontend/v2/frontend.go
charleskorn: Good idea, fixed in 20c7b24.

pkg/frontend/v2/frontend.go
charleskorn: Changed in 20c7b24.

pkg/frontend/v2/frontend.go
charleskorn: Good point, changed in 44fc2ab.

pkg/frontend/v2/frontend.go
charleskorn: Yes we can: 15e9632

pkg/frontend/v2/frontend.go
charleskorn: How's what I've added in 036d891?

pkg/api/error/error.go
charleskorn: Good catch, I've updated `mapEngineError` to use this new method in 96b1b63.

pkg/frontend/v2/frontend.go
56quarters: LGTM

integration/query_frontend_test.go
56quarters: I don't have a good solution to this besides making invasive changes to the response format just for integration testing. :+1: from me

56quarters (approved):
LGTM but I'd like someone else to take a look given the scope of the changes

integration/query_frontend_test.go
```diff
@@ -394,23 +437,47 @@ func runQueryFrontendTest(t *testing.T, cfg queryFrontendTestConfig) {
...
+	expectedQueryFrontendQueryCount := simpleQueryCount + 6
```
Konstantinov-Innokentii: Does it make sense to add a comment to the "+6", similar to the comments below?

pkg/frontend/querymiddleware/limits.go
```diff
@@ -302,6 +305,92 @@ func (rth httpQueryRequestRoundTripperHandler) Do(ctx context.Context, r Metrics
...
+}
+
+type unqueryableQueryable struct{}
+
+var errShouldNeverBeQueried = errors.New("this Queryable should never be queried as all selectors should be evaluated remotely: if you are seeing this, this is a bug")
+
+func (u unqueryableQueryable) Querier(mint, maxt int64) (storage.Querier, error) {
+	return nil, errShouldNeverBeQueried
+}
+
```
Konstantinov-Innokentii, line 393: nit: maybe it makes sense to drag `smallestPositiveNonZeroDuration` closer to limitsMiddleware.Do where it's actually used, rather to hang around under the `engineQueryRequestRoundTripperHandler` related code?

pkg/frontend/v2/frontend.go
Konstantinov-Innokentii: WDYT about adding the comment to frontendRequest, describing when and why cases it will be httpgrpc or protobuf?

pkg/frontend/v2/remoteexec.go
```diff
@@ -0,0 +1,369 @@
...
+	stream, err := r.startExecution(ctx, fullPlan, node, timeRange)
+	if err != nil {
+		return nil, err
+	}
+
+	return newRangeVectorExecutionResponse(stream, memoryConsumptionTracker), nil
+}
+
+func (r *RemoteExecutor) startExecution(ctx context.Context, fullPlan *planning.QueryPlan, node planning.Node, timeRange types.QueryTimeRange) (responseStream, error) {
+	subsetPlan := &planning.QueryPlan{
```
Konstantinov-Innokentii, line 59: I'm not sure I fully get it - why remote executor expects a node? Is it because it capable of executing whole plan and plan starting from the node?

pkg/mimir/modules.go
```diff
@@ -874,7 +883,7 @@ func (t *Mimir) initQueryFrontend() (serv services.Service, err error) {
 		t.Cfg.Frontend.FrontendV2.SchedulerAddress = address
 	}
 
-	roundTripper, frontendV2, err := frontend.InitFrontend(
+	roundTripper, frontend, err := frontend.InitFrontend(
```
Konstantinov-Innokentii, line 886: Did we got rid of frontendv1/v2 in mimir 3?

pkg/streamingpromql/optimize/plan/remoteexec/operators.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"github.com/prometheus/prometheus/util/annotations"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+)
+
+func finalise(ctx context.Context, resp RemoteExecutionResponse, annos *annotations.Annotations, queryStats *types.QueryStats) error {
```
Konstantinov-Innokentii: nit: should we stick with finalize or finalise everywhere?

pkg/streamingpromql/optimize/plan/remoteexec/range_vector_operator.go
```diff
@@ -0,0 +1,77 @@
...
+	stepsReadForCurrentSeries int
+}
+
+var _ types.RangeVectorOperator = &RangeVectorRemoteExec{}
+
+func (r *RangeVectorRemoteExec) Prepare(ctx context.Context, params *types.PrepareParams) error {
+	r.QueryStats = params.QueryStats
+
+	var err error
+	r.resp, err = r.RemoteExecutor.StartRangeVectorExecution(ctx, r.RootPlan, r.Node, r.TimeRange, r.MemoryConsumptionTracker)
```
Konstantinov-Innokentii, line 36: ~Why does remote execution start in prepare?~
After some thinking, I think it's actually a good use of Prepare.

pkg/streamingpromql/optimize/plan/remoteexec/optimization_pass.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (o *OptimizationPass) Apply(ctx context.Context, plan *planning.QueryPlan) (*planning.QueryPlan, error) {
+	containsSelectors, containsShardedOrSpunOffExpression := o.inspect(plan.Root)
+
+	if !containsSelectors || containsShardedOrSpunOffExpression {
+		return plan, nil
+	}
+
+	n := &RemoteExecution{}
```
Konstantinov-Innokentii: Maybe move wrapping to separate function, something like "MakeRemoteExecutable"?

pkg/frontend/config.go
```diff
@@ -46,6 +47,7 @@ func (cfg *CombinedFrontendConfig) RegisterFlags(f *flag.FlagSet, logger log.Log
 
 	f.StringVar(&cfg.QueryEngine, "query-frontend.query-engine", querier.MimirEngine, fmt.Sprintf("Query engine to use, either '%v' or '%v'", querier.PrometheusEngine, querier.MimirEngine))
 	f.BoolVar(&cfg.EnableQueryEngineFallback, "query-frontend.enable-query-engine-fallback", true, "If set to true and the Mimir query engine is in use, fall back to using the Prometheus query engine for any queries not supported by the Mimir query engine.")
+	f.BoolVar(&cfg.EnableRemoteExecution, "query-frontend.enable-remote-execution", false, "If set to true and the Mimir query engine is in use, use remote execution to evaluate queries in queriers.")
```
Konstantinov-Innokentii, line 50: It will do nothing if MQE is not enabled? Should we error in this case?

pkg/frontend/querymiddleware/limits.go
charleskorn: That method has been there for quite a while, and while I generally agree with you, I want to keep the amount of changes in this PR to a minimum if I can.

pkg/frontend/v2/remoteexec.go
charleskorn: Exactly - the remote executor can evaluate any node. In the future, we'll pass it a set of nodes to evaluate in one request.

pkg/mimir/modules.go
charleskorn: Yep, v1 was removed in https://github.com/grafana/mimir/pull/12200.

pkg/streamingpromql/optimize/plan/remoteexec/operators.go
charleskorn: Good catch, it should be finalize, as much as that pains me 🙂 I'll fix this.

pkg/frontend/config.go
charleskorn: Good idea, added in a1f07b7.

pkg/streamingpromql/optimize/plan/remoteexec/optimization_pass.go
charleskorn: Done in 96b30f5.

pkg/streamingpromql/optimize/plan/remoteexec/operators.go
charleskorn: Fixed in 5a8a51c

integration/query_frontend_test.go
charleskorn: Added in 8432b8c.

PR #12552, opened 2025-08-29
Introduce `FromSamplesToFPoints` and `FromHistogramsToHPoints`

charleskorn:
#### What this PR does

This PR introduces two new methods to the `mimirpb` package, to convert from the `mimirpb` types to their `promql` equivalents.

This has been extracted from a larger PR to try to make that PR smaller and easier to review.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


lamida (approved):
Since the converted fields are either from prometheus or in mimir protobuf which won't change its field order the I think this should be LGTM.

PR #12571, opened 2025-09-01
Expose query expression length metrics

monoxane:
#### What this PR does
This PR adds a native histogram representing the length of query expressions handled by the query frontend in bytes, and presents this data to users in the queries, tenants, and top-tenant mixin dashboards. 

This data allows Mimir administrators to better decide on the value to set for the `max-query-expression-size-bytes` limit option.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 * [FEATURE] Alertmanager: Add experimental `-alertmanager.storage.state-read-timeout` flag to configure the timeout for reading the Alertmanager state (notification log, silences) from object storage during the initial sync. #12425
 * [FEATURE] Ingester: Add experimental `-blocks-storage.tsdb.head-statistics-collection-frequency` flag to configure the periodic collection of statistics from the TSDB head. #12407
 * [FEATURE] Ingester: Add experimental `blocks-storage.tsdb.index-lookup-planning-enabled` flag to configure use of a cost-based index lookup planner. #12530
+* [FEATURE] Query-frontend: Add a native histogram presenting the length of of query expressions handled by the query-frontend #12571
```
```suggestion
* [FEATURE] Query-frontend: Add a native histogram presenting the length of query expressions handled by the query-frontend #12571
```

tacole02 (approved):
Minor suggestion to the changelog.

CHANGELOG.md
monoxane: I should have caught that 🤦

development/mimir-microservices-mode/docker-compose.jsonnet
charleskorn, line 1: [nit] I would separate this change out to a separate PR, they're not strictly related to this PR.

operations/mimir-mixin/dashboards/queries.libsonnet
```diff
@@ -178,6 +178,33 @@ local filename = 'mimir-queries.json';
...
+        $.queryPanel(
+          'histogram_quantile(0.99, cortex_query_frontend_queries_expression_bytes{namespace="$namespace"})',
+          '99th Percentile'
+        ) +
+        $.queryPanel(
+          'histogram_quantile(0.90, cortex_query_frontend_queries_expression_bytes{namespace="$namespace"})',
+          '90th Percentile'
+        ) +
+        $.queryPanel(
+          'histogram_avg(cortex_query_frontend_queries_expression_bytes{namespace="$namespace"})',
```
charleskorn: I don't think these queries are quite right.

I think the first should be something like `sum(rate(cortex_query_frontend_queries_expression_bytes[$__rate_interval]))` the `rate` ensures we get the values for this step (rather than all time since each query-frontend started), and the `sum` combines the values from each query-frontend.

And I think the percentile queries need a similar change as well - so they'd be something like `histogram_quantile(0.99, sum(rate(cortex_query_frontend_queries_expression_bytes{namespace="$namespace"}[$__rate_interval])))`.

operations/mimir-mixin/dashboards/tenants.libsonnet
charleskorn, line 1: Same comments apply to the queries in this dashboard too.

development/mimir-microservices-mode/docker-compose.jsonnet
monoxane: Extracted to https://github.com/grafana/mimir/pull/12600. Should I filter-branch it out of this branch and force-push or just don't do that next time?

charleskorn: Up to you, I don't think it's a big deal to leave things as-is in this PR.

monoxane: I'll leave it here for this one and make sure to do that sort of work separately going forward 👍

operations/mimir-mixin/dashboards/queries.libsonnet
monoxane: Yeap, you're right here. This looks fine with a single instance of the query-frontend but when running multiple it returns inaccurate results. I've fixed them all after testing locally with a multi-frontend mimir environment.

charleskorn (approved):

PR #12577, opened 2025-09-02
MQE: add selectors to binary RHS based on LHS results

56quarters:
#### What this PR does

This change introduces a query planning pass that finds fields that would be good to add selectors to reduce the data queried for binary operations on one side when we already know the series returned by the other side.

N/A

#### Checklist

- [X] Tests updated.
- [x] Documentation added.


pkg/streamingpromql/operators/binops/one_to_one_vector_vector_binary_operation.go
```diff
@@ -45,6 +44,7 @@ type OneToOneVectorVectorBinaryOperation struct {
 	expressionPosition posrange.PositionRange
 	annotations        *annotations.Annotations
 	timeRange          types.QueryTimeRange
+	hints              *types.QueryHints
```
charleskorn: [nit] Given these hints apply specifically to binary operations, I would move the `QueryHints` type to this package.

pkg/streamingpromql/types/operator.go
```diff
@@ -50,7 +51,9 @@ type SeriesOperator interface {
 	// The returned []SeriesMetadata can be modified by the caller or returned to a pool.
 	// SeriesMetadata may return series in any order, but the same order must be used by both SeriesMetadata and NextSeries.
 	// SeriesMetadata should be called no more than once.
-	SeriesMetadata(ctx context.Context) ([]SeriesMetadata, error)
+	// TODO: Docs
+	// TODO: Pointer or value type?
+	SeriesMetadata(ctx context.Context, selectors *SeriesSelectors) ([]SeriesMetadata, error)
```
charleskorn: [nit] Should we call the new parameter "matchers" or "selector"? A selector is a set of matchers.

pkg/streamingpromql/planning/core/core.proto
```diff
@@ -20,6 +20,10 @@ message PositionRange {
   int64 end = 2 [(gogoproto.casttype) = "github.com/prometheus/prometheus/promql/parser/posrange.Pos"];
 }
 
+message Hints {
```
charleskorn: Should this be called something like `BinaryExpressionHints`? I imagine we'll have other kinds of hints for other kinds of nodes in the future.

charleskorn (commented):
Overall approach and structure looks good to me, some suggestions on naming:

pkg/streamingpromql/operators/binops/binary_operation.go
```diff
@@ -746,3 +749,42 @@ var boolComparisonOperationFuncs = map[parser.ItemType]binaryOperationFunc{
...
+func BuildMatchers(metadata []types.SeriesMetadata, hints *QueryHints) types.Matchers {
+	var matchers []*labels.Matcher
+
+	for _, field := range hints.Include {
+		values := make(map[string]struct{})
+
+		for _, series := range metadata {
+			val := series.Labels.Get(field)
+			if val != "" {
+				values[regexp.QuoteMeta(val)] = struct{}{}
```
charleskorn: How expensive is `regexp.QuoteMeta`? I'm wondering if we should use `val` as-is for the key, and then only bother quoting it in the loop below. Otherwise we'll potentially waste cycles here quoting the same string over and over again.

pkg/streamingpromql/operators/binops/binary_operation.go
```diff
@@ -746,3 +749,42 @@ var boolComparisonOperationFuncs = map[parser.ItemType]binaryOperationFunc{
...
+		//  that since we're building a regex, we need to not include it at all if we exceed
+		//  the max number of values rather than include only a few matchers.
+
+		if len(values) != 0 {
+			ordered := make([]string, 0, len(values))
+			for k := range maps.Keys(values) {
+				ordered = append(ordered, k)
+			}
+
+			slices.Sort(ordered)
```
charleskorn, line 780: Why do we need to sort this slice?

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: It's looping through the string at least once so I'll change it to avoid quoting until building the matcher.

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: We don't _need_ to but I was imagining that we might need to deduplicate matchers and so I wanted the value in a predictable order. I can remove this until there's a concrete need.

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: Fixed in 567724cc6c7f73da13e334073ba5ddeec81bb128

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: Fixed in 567724cc6c7f73da13e334073ba5ddeec81bb128

56quarters: Actually, here's an example. This is not ideal but seems harmless from what I can tell.

Given the following (nonsense) query:
```
stddev by (container) (rate(cortex_request_duration_seconds_sum{cluster="docker-compose"}[$__rate_interval]))
/ 
(
  stddev by (container) (rate(cortex_request_duration_seconds_count{cluster="docker-compose"}[$__rate_interval])) + 
  stddev by (container) (rate(cortex_kv_request_duration_seconds_count{cluster="docker-compose"}[$__rate_interval]))
)
```

The selector for `cortex_kv_request_duration_seconds_count` is passed matchers that look like this:

```
container=~"compactor|continuous-test|querier|query-frontend|store-gateway|distributor|ingester|query-scheduler|ruler",
container=~"compactor|continuous-test|distributor|ingester|querier|query-frontend|query-scheduler|ruler|store-gateway",
__name__="cortex_kv_request_duration_seconds_count",
cluster="docker-compose"
```

...because it's on the RHS of both binary operations. In this case the matchers are identical except for the order of the matches in the regular expression.

charleskorn: >This is not ideal but seems harmless from what I can tell.

Do any of the regex optimisations in store-gateways or ingesters perform better with sorted matchers? Or is there any caching of the parsed regex or postings or anything like that that would benefit from multiple selectors having the alternatives in a consistent order? (Or is all caching at the level of the entire selector, rather than single matchers?)

56quarters: I've introduced a new type in b7deae7384554e080e1312ff21b43ffc34ea29c9 so that we can deduplicate matchers using a `map`. I don't know if there are benefits to ordered matchers in terms of caching but I've avoided the issue by deduplicating in `Selector`.

pkg/streamingpromql/operators/aggregations/aggregation.go
```diff
@@ -118,9 +118,9 @@ func (a *Aggregation) ExpressionPosition() posrange.PositionRange {
 	return a.expressionPosition
 }
 
-func (a *Aggregation) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
+func (a *Aggregation) SeriesMetadata(ctx context.Context, matchers types.Matchers) ([]types.SeriesMetadata, error) {
 	// Fetch the source series
-	innerSeries, err := a.Inner.SeriesMetadata(ctx)
+	innerSeries, err := a.Inner.SeriesMetadata(ctx, matchers)
```
charleskorn, line 139: Do we need to filter these matchers to only include matchers for the group labels produced by this aggregation?

eg. if the aggregation is `sum by (environment) (foo)`, and this method received matchers for `environment` and `architecture`, do we need to drop the `architecture` matcher?

Or are we guaranteed to only get matchers that are for the group labels?

pkg/streamingpromql/operators/binops/and_unless_binary_operation.go
```diff
@@ -82,8 +82,8 @@ func (a *AndUnlessBinaryOperation) SeriesMetadata(ctx context.Context) ([]types.
 	return series, nil
 }
 
-func (a *AndUnlessBinaryOperation) computeSeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
-	leftMetadata, err := a.Left.SeriesMetadata(ctx)
+func (a *AndUnlessBinaryOperation) computeSeriesMetadata(ctx context.Context, matchers types.Matchers) ([]types.SeriesMetadata, error) {
+	leftMetadata, err := a.Left.SeriesMetadata(ctx, matchers)
```
charleskorn, line 86: Not something for this PR, but I imagine we could apply a similar optimisation here, propagating matchers to the RHS based on the LHS.

pkg/streamingpromql/operators/binops/binary_operation.go
```diff
@@ -746,3 +749,64 @@ var boolComparisonOperationFuncs = map[parser.ItemType]binaryOperationFunc{
 		return 0, nil, true, true, nil
 	},
 }
+
+// QueryHints are hints that can be applied to binary operations to avoid doing
+// unnecessary work.
+type QueryHints struct {
```
charleskorn: [nit] Perhaps just `Hints` is enough? The `Query` seems unnecessary (and the hints aren't really for the whole query, just for a binary operation)

pkg/streamingpromql/operators/binops/binary_operation.go
```diff
@@ -746,3 +749,64 @@ var boolComparisonOperationFuncs = map[parser.ItemType]binaryOperationFunc{
...
+		// values that we'll include in a matcher. In this case, we can't use the
+		// values collected so far to build a matcher.
+		if len(values) >= maxValues {
+			return map[string]struct{}{}
+		}
+
+		val := series.Labels.Get(label)
+		if val != "" {
+			values[val] = struct{}{}
+		}
```
charleskorn, line 807: [non-blocking, needs a benchmark] I suspect this is going to result in a lot of string allocations, many of which will be the same value over and over again. 

Perhaps we could use a similar trick to what we do for aggregation grouping, using `series.Labels.BytesWithLabels` instead of `series.Labels.Get`?

pkg/streamingpromql/operators/binops/binary_operation.go
```diff
@@ -746,3 +749,64 @@ var boolComparisonOperationFuncs = map[parser.ItemType]binaryOperationFunc{
...
+				Value: strings.Join(ordered, "|"),
+			})
+		}
+	}
+
+	return matchers
+}
+
+func getUniqueLabelValues(metadata []types.SeriesMetadata, label string, maxValues int) map[string]struct{} {
+	values := make(map[string]struct{})
```
charleskorn, line 794: [non-blocking, needs a benchmark] I wonder if it would be worth pre-allocating this for `maxHintMatcherValues`, and possibly pooling it as well.

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/instant_vector_operator.go
```diff
@@ -234,8 +234,8 @@ type InstantVectorDuplicationConsumer struct {
 
 var _ types.InstantVectorOperator = &InstantVectorDuplicationConsumer{}
 
-func (d *InstantVectorDuplicationConsumer) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata, error) {
-	return d.Buffer.SeriesMetadata(ctx)
+func (d *InstantVectorDuplicationConsumer) SeriesMetadata(ctx context.Context, matchers types.Matchers) ([]types.SeriesMetadata, error) {
+	return d.Buffer.SeriesMetadata(ctx, matchers)
```
charleskorn, line 241: I think we have to discard the matchers in this case (and same for the range vector duplication operator) - as it stands, we'll only use the set of matchers passed by the first instance to use this duplicated result. That means that if the other instance that use this duplicated result compute different matchers, these will be ignored, and the first set of matchers might filter out series that the subsequent instances need.

pkg/streamingpromql/types/operator.go
```diff
@@ -50,7 +51,8 @@ type SeriesOperator interface {
 	// The returned []SeriesMetadata can be modified by the caller or returned to a pool.
 	// SeriesMetadata may return series in any order, but the same order must be used by both SeriesMetadata and NextSeries.
 	// SeriesMetadata should be called no more than once.
-	SeriesMetadata(ctx context.Context) ([]SeriesMetadata, error)
+	// TODO: Docs
```
charleskorn: Is this TODO still relevant?

pkg/streamingpromql/optimize/plan/narrow_selectors.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func NewNarrowSelectorsOptimizationPass(logger log.Logger) *NarrowSelectorsOptimizationPass {
+	return &NarrowSelectorsOptimizationPass{
+		logger: logger,
+	}
+}
+
+func (n *NarrowSelectorsOptimizationPass) Name() string {
+	return ""
```
charleskorn: This name is used by the analysis endpoint (and UI) - something like "narrow selectors" would make sense here

pkg/streamingpromql/optimize/plan/narrow_selectors.go
```diff
@@ -0,0 +1,104 @@
...
+	switch e := node.(type) {
+	case *core.BinaryExpression:
+		// Set hints for a binary expression based on the expression itself and any
+		// children from the left hand side of the expression. Note that this stops
+		// after finding the first node that allows us to generate hints for the query
+		// (binary expressions or aggregations).
+		e.Hints = n.hintsFromNode(ctx, e)
+		if e.Hints != nil {
+			sl := spanlogger.FromContext(ctx, n.logger)
+			sl.DebugLog("msg", "setting query hint on binary expression", "fields", e.Hints.GetInclude())
```
charleskorn: [nit] Would this be a better name?

```suggestion
			sl.DebugLog("msg", "setting query hint on binary expression", "labels", e.Hints.GetInclude())
```

pkg/streamingpromql/optimize/plan/narrow_selectors.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (n *NarrowSelectorsOptimizationPass) hintsFromNode(ctx context.Context, node planning.Node) *core.BinaryExpressionHints {
+	switch e := node.(type) {
+	case *core.BinaryExpression:
+		if e.VectorMatching != nil && e.VectorMatching.On && len(e.VectorMatching.MatchingLabels) > 0 {
```
charleskorn, line 77: Are you planning to do something similar for binary expressions with `without`? (doesn't have to be part of this PR)

pkg/streamingpromql/optimize/plan/narrow_selectors_test.go
```diff
@@ -0,0 +1,183 @@
...
+			expr: `sum by (region) (some_metric) / some_other_metric`,
+			expectedPlan: `
+				- BinaryExpression: LHS / RHS, hints (region)
+					- LHS: AggregateExpression: sum by (region)
+						- VectorSelector: {__name__="some_metric"}
+					- RHS: VectorSelector: {__name__="some_other_metric"}
+			`,
+		},
+		"binary expression aggregation LHS aggregation RHS": {
+			expr: `sum by (region) (some_metric) / sum(some_other_metric)`,
```
charleskorn, line 67: Could you please add a test case like `sum by (region) (some_metric) / sum by (region) (some_other_metric)`?

pkg/streamingpromql/optimize/plan/narrow_selectors_test.go
```diff
@@ -0,0 +1,183 @@
...
+			expectedPlan: `
+				- BinaryExpression: LHS / RHS, hints (region)
+					- LHS: AggregateExpression: sum by (region)
+						- VectorSelector: {__name__="some_metric"}
+					- RHS: AggregateExpression: sum
+						- VectorSelector: {__name__="some_other_metric"}
+			`,
+		},
+		"binary expression multiple aggregation LHS aggregation RHS": {
+			expr: `sum by (region, env) (some_metric) / sum(some_other_metric)`,
```
charleskorn, line 87: It'd be good to add some more variations on this:

* `sum by (region, env) (some_metric) / sum by (region, env) (some_other_metric)`
* `sum by (region, env) (some_metric) / sum by (region, cluster) (some_other_metric)`
* `sum by (region, env) (some_metric) / on(region) sum(some_other_metric)`

pkg/streamingpromql/operators/selectors/selector.go
```diff
@@ -89,7 +89,32 @@ func (s *Selector) SeriesMetadata(ctx context.Context) ([]types.SeriesMetadata,
 	return metadata, s.seriesSet.Err()
 }
 
-func (s *Selector) loadSeriesSet(ctx context.Context) error {
+func (s *Selector) mergeMatchers(m1, m2 types.Matchers) types.Matchers {
```
charleskorn, line 92: Could you please add some tests for this merging behaviour?

pkg/streamingpromql/optimize/plan/narrow_selectors_test.go
```diff
@@ -0,0 +1,183 @@
...
+		"binary expression raw vector selectors": {
+			expr: `some_metric + some_other_metric`,
+			expectedPlan: `
+				- BinaryExpression: LHS + RHS
+					- LHS: VectorSelector: {__name__="some_metric"}
+					- RHS: VectorSelector: {__name__="some_other_metric"}
+			`,
+		},
+		"binary expression on raw vector selectors": {
+			expr: `some_metric + on (cluster) some_other_metric`,
```
charleskorn, line 40: For a case like `some_metric{cluster="prod"} + on (cluster) some_other_metric{cluster="prod"}`, should we skip the hint for `cluster`? 

(this doesn't have to be part of this PR)

pkg/streamingpromql/operators/selectors/selector.go
charleskorn, line 92: Something else I'm wondering (not for this PR): would it make sense to try to merge individual matchers for the same label?

eg. imagine the original selector was `my_metric{cluster=~"prod-us-.*"}` and a binary operation results in an additional matcher like `cluster=~"prod-us-east|prod-us-west|prod-eu-south"`. The merging as it stands here would produce a selector like `my_metric{cluster=~"prod-us-.*", cluster=~"prod-us-east|prod-us-west|prod-eu-south"}`, but we could take the intersection of the two matchers to produce `my_metric{cluster=~"prod-us-east|prod-us-west"}`.

Or maybe this is something we should leave for ingesters and store-gateways to optimise? 🤷

pkg/streamingpromql/types/operator.go
56quarters: :man_facepalming: yep

pkg/streamingpromql/optimize/plan/narrow_selectors.go
56quarters: Yes, I just wanted to limit the scope for this PR and keep it simple while I got my head around it.

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/optimize/plan/commonsubexpressionelimination/instant_vector_operator.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/types/operator.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/optimize/plan/narrow_selectors.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/optimize/plan/narrow_selectors.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/optimize/plan/narrow_selectors_test.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/optimize/plan/narrow_selectors_test.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/operators/selectors/selector.go
56quarters: Fixed in a6ce953d4dccf79bac031c8708997e8fc3f65d67

pkg/streamingpromql/operators/aggregations/aggregation.go
56quarters: > Or are we guaranteed to only get matchers that are for the group labels?

We aren't. I'll update this to only retain matchers for the fields being grouped by.

pkg/streamingpromql/operators/binops/binary_operation.go
56quarters: I took a quick pass at this and the current naive implementation is the fastest with fewest allocations version I'm able to come up with. I've pushed the benchmark to this branch and the alternate implementation here: https://github.com/grafana/mimir/compare/56quarters/skip-querying-rhs-2..56quarters/skip-querying-rhs-2-alt-matcher-builder

pkg/streamingpromql/operators/aggregations/aggregation.go
56quarters: Actually, I'm trying to think of when this could come up. With the current optimization pass, we'd only pass selectors to the RHS if the binary operation is joining on the label or the LHS is aggregating on the label. If the RHS is aggregating by different labels the query wouldn't return any results.

I think we should update the operator so it prevents misuse (removing matchers for labels that aren't being grouped by) but I don't think this could happen in practice. Am I missing something?

charleskorn: I think you're right, but I also agree that it would be good to filter the matchers so we only apply those for grouping labels for safety.

pkg/streamingpromql/operators/binops/binary_operation.go
charleskorn: Interesting - what you've got in this PR looks fine to me, we can always optimise it later if need be.

pkg/streamingpromql/operators/aggregations/aggregation.go
56quarters: Fixed in c1041034600b66d7e43563ee7ec5e9a4f546f54f

PR #12578, opened 2025-09-02
MQE: return `APIError` instances for non-execution errors

charleskorn:
#### What this PR does

This PR modifies MQE to return `APIError` instances when appropriate, and modifies the querier's `Dispatcher` to translate these to the appropriate Protobuf constant for transmission back to query-frontends when remote execution is being used.

This ensures that the correct HTTP status code is used when returning an error response for a query request evaluated with remote execution.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/querier/dispatcher.go
+++ b/pkg/querier/dispatcher.go
@@ -196,6 +196,10 @@ func errorTypeForError(err error) mimirpb.QueryErrorType {
 	var apiError *apierror.APIError
 	if errors.As(err, &apiError) {
 		t, conversionErr := mimirpb.ErrorTypeFromAPIErrorType(apiError.Type)
+
+		// ErrorTypeFromAPIErrorType should never fail, as the APIError and QueryErrorType enums should remain
+		// in sync (and this is enforced with a test).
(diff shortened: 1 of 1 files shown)


pkg/querier/dispatcher.go
```diff
@@ -192,6 +193,14 @@ func errorTypeForError(err error) mimirpb.QueryErrorType {
 		return mimirpb.QUERY_ERROR_TYPE_INTERNAL
 	}
 
+	var apiError *apierror.APIError
+	if errors.As(err, &apiError) {
+		t, conversionErr := mimirpb.ErrorTypeFromAPIErrorType(apiError.Type)
+		if conversionErr == nil {
```
56quarters, line 203: Does a conversion error here indicate a bug in Mimir? Should we do something to indicate that here or is the idea that the tests that parse Prometheus source files will catch that?

charleskorn: The conversion tests should catch that, but I'll make this assumption more explicit here.

PR #12553, opened 2025-08-29
Add a test to check OTLP response status code on 'out of order exemplars'

pracucci:
#### What this PR does

TBD

#### Which issue(s) this PR fixes or relates to

N/A

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12579, opened 2025-09-02
OTLP: Return HTTP OK when partially rejecting e.g. due to OOO exemplars

aknuds1:
#### What this PR does

Modify the OTLP endpoint to return HTTP OK when partially rejecting due to OOO exemplars, to adhere with [spec](https://opentelemetry.io/docs/specs/otlp/#otlphttp-response).

I'm also modifying the ingester to include in gRPC error details whether an error is soft or not, so distributors can recognize so-called soft errors (corresponding to partial rejection).

I've manually tested the fix in a development deployment, and can confirm it works.

Supersedes #12553. Relates to #12626.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/distributor/otel.go
```diff
@@ -177,6 +183,19 @@ func OTLPHandler(
 	})
 }
 
+func handlePartialOTLPPush(pushErr error, w http.ResponseWriter, r *http.Request, req *Request, logger log.Logger) {
+	// Respond as per spec:
+	// https://opentelemetry.io/docs/specs/otlp/#otlphttp-response.
+	expResp := colmetricpb.ExportMetricsServiceResponse{
+		PartialSuccess: &colmetricpb.ExportMetricsPartialSuccess{
+			RejectedDataPoints: 0,
```
duricanikolic, line 191: Why do we set this to 0? Can't we use the exact number of rejected samples?

pkg/distributor/errors.go
```diff
@@ -408,13 +408,16 @@ func wrapDeadlineExceededPushError(err error) error {
...
-		return ingesterPushErr.Cause() == mimirpb.ERROR_CAUSE_BAD_DATA
+		cause = ingesterPushErr.Cause()
+	} else if errors.As(err, &partitionPushErr) {
+		cause = partitionPushErr.Cause()
 	}
 
-	var partitionPushErr partitionPushError
-	if errors.As(err, &partitionPushErr) {
-		return partitionPushErr.Cause() == mimirpb.ERROR_CAUSE_BAD_DATA
+	if cause == mimirpb.ERROR_CAUSE_BAD_DATA || cause == mimirpb.ERROR_CAUSE_SOFT_BAD_DATA {
```
duricanikolic: I think you've changed the semantics here.
Previously, if it was an `ingesterError`, or a `partitionPushError`, we will returning `true` or `false` depending on whether the cause was `mimirpb.ERROR_CAUSE_BAD_DATA`.
Only if it was neither `ingesterError` nor `partitionPushError` we were going further. Now we go further even if it is an `ingesterError` or `partitionPushError` with a cause `mimirpb.UNKNOWN` for example.

pkg/mimirpb/mimir.proto
```diff
@@ -59,6 +59,7 @@ enum ErrorCause {
   ERROR_CAUSE_METHOD_NOT_ALLOWED = 11;
   ERROR_CAUSE_TENANT_LIMIT = 12;
   ERROR_CAUSE_ACTIVE_SERIES_LIMITED = 13;
+  ERROR_CAUSE_SOFT_BAD_DATA = 14;
 }
 
 message ErrorDetails {
```
duricanikolic, line 64: I am not convinced that adding another cause is the right way to go:
- one reason is because, we will have to add one additional cause for each type of (soft) error we would like to process as partially successful on OTLP side
- another reason is because it will complicate the logic on distributor or OTLP side where we will have to check the cause.

What if, instead of adding `ERROR_CAUSE_SOFT_BAD_DATA` (I don't like this name so much), we extend the `ErrorDetails` message by doing the following:
```
message ErrorDetails {
  ErrorCause Cause = 1;
  bool Soft = 2;
}
```

Similarly, we will have to extend `ingester.ingesterError` and `distributor.Error` interfaces:
```
// belongs to the ingester package
type ingesterError interface {
  errorCause() mimirpb.ErrorCause
  isSoft() bool
}

// belongs to the distributor package
type Error interface {
  Cause() mimirpb.ErrorCause
  IsSoft() bool
}
```

This way, `ingesterError` (on ingester side) can propagate its `isSoft()` info via `mimirpb.ErrorDetails` to the distributor. Similarly, `Error` (on distributor side) can propagate its `IsSoft()` info to the OTLP handler, which can just check if `IsSoft()` is `true`, and in that case, can map the error to `httpstatus.OK`.
WDYT?

pkg/distributor/otel.go
aknuds1: No samples are rejected though, just exemplars. Doesn't data point mean sample? Also, how can we pass the number of rejected exemplars back to the distributor. Would it have to be via the error message?

pkg/distributor/errors.go
aknuds1: Will look into it, thanks.

pkg/mimirpb/mimir.proto
aknuds1: Thanks, I will look into it. I wasn't sure whether it was possible to add another field. Not really familiar with our gRPC error passing system yet.

pkg/distributor/errors.go
aknuds1: I dropped this change.

pkg/mimirpb/mimir.proto
aknuds1: @duricanikolic I implemented this scheme in my latest revision. Can you take a look please?

I did say originally I wanted to limit the scope of this PR to OOO exemplars, but when propagating whether ingester errors are soft, which is an existing concept, I don't think it's practical to avoid also translating other "soft" errors in the OTLP handler. AFAICT, the "soft" concept should correspond 1:1 to OTel's ["partial success" concept](https://opentelemetry.io/docs/specs/otlp/#partial-success-1) though. WDYT?

pkg/distributor/otel.go
```diff
@@ -385,16 +404,17 @@ func validateTranslationStrategy(translationStrategy otlptranslator.TranslationS
...
+		return codes.Internal, http.StatusServiceUnavailable, false
 	}
 
-	grpcStatusCode := errorCauseToGRPCStatusCode(distributorErr.Cause(), false)
-	httpStatusCode := errorCauseToHTTPStatusCode(distributorErr.Cause(), false)
-	otlpHTTPStatusCode := httpRetryableToOTLPRetryable(httpStatusCode)
-	return grpcStatusCode, otlpHTTPStatusCode
+	errCause := distributorErr.Cause()
+	isSoft := distributorErr.IsSoft()
+	grpcStatusCode := errorCauseToGRPCStatusCode(errCause, false)
```
duricanikolic: [nit] I would use `distributorErr.Cause()` and `distributor.IsSoft()` instead of declaring variables. But this is just my personal preference. Semantically they are identical.

pkg/distributor/otel.go
```diff
@@ -710,8 +730,8 @@ func TimeseriesToOTLPRequest(timeseries []prompb.TimeSeries, metadata []mimirpb.
 				if i == 0 {
 					for _, tsEx := range ts.Exemplars {
 						ex := datapoint.Exemplars().AppendEmpty()
-						ex.SetDoubleValue(histogram.Sum / 10.0) // Doesn't really matter, just a placeholder
-						ex.SetTimestamp(pcommon.Timestamp(histogram.Timestamp * time.Millisecond.Nanoseconds()))
+						ex.SetDoubleValue(tsEx.Value)
```
duricanikolic, line 731: I don't understand this change, but I trust you :)

pkg/distributor/otel.go
aknuds1: Good point, I think it was a change that unintentionally carried over from the previous revision. I dropped that change.

pkg/distributor/otel.go
aknuds1: It's a change from Marco's [PR](https://github.com/grafana/mimir/pull/12553), which adds the test.

pkg/distributor/otel_test.go
```diff
@@ -1510,154 +1510,184 @@ func TestHandler_toOtlpGRPCHTTPStatus(t *testing.T) {
...
 			err:                newIngesterPushError(createStatusWithDetails(t, codes.Unavailable, originalMsg, mimirpb.ERROR_CAUSE_CIRCUIT_BREAKER_OPEN), ingesterID),
 			expectedHTTPStatus: http.StatusServiceUnavailable,
 			expectedGRPCStatus: codes.Unavailable,
+			expectedSoft:       false,
 		},
 		"a wrapped ingesterPushError with CIRCUIT_BREAKER_OPEN cause gets translated into an Unavailable error with CIRCUIT_BREAKER_OPEN cause": {
 			err:                errors.Wrap(newIngesterPushError(createStatusWithDetails(t, codes.Unavailable, originalMsg, mimirpb.ERROR_CAUSE_CIRCUIT_BREAKER_OPEN), ingesterID), "wrapped"),
 			expectedHTTPStatus: http.StatusServiceUnavailable,
 			expectedGRPCStatus: codes.Unavailable,
+			expectedSoft:       false,
```
duricanikolic, line 1725: I would expect to see additional 2 test cases of type:
- an soft ingesterPushError gets translated into a soft error
- a wrapped soft ingesterPushError gets translated into a soft error

```
		"a soft ingesterPushError gets translated into soft error": {
			err:                ingesterPushError{message: originalMsg, cause: mimirpb.ERROR_CAUSE_BAD_DATA, soft: true},
			expectedHTTPStatus: http.StatusBadRequest,
			expectedGRPCStatus: codes.InvalidArgument,
			expectedSoft:       true,
		},
		"a wrapped soft ingesterPushError gets translated into soft error": {
			err:                errors.Wrap(ingesterPushError{message: originalMsg, cause: mimirpb.ERROR_CAUSE_BAD_DATA, soft: true}, "wrapped"),
			expectedHTTPStatus: http.StatusBadRequest,
			expectedGRPCStatus: codes.InvalidArgument,
			expectedSoft:       true,
		},
```

duricanikolic:
I would expect to see an extension of `TestHandlerOTLPPush()` in `otel_test.go` and to see test cases where:
- the push handler returns an `ingesterPushError` (resulting in an erroneous status code), and
- the push handler resturns a soft `ingesterPushError` (resulting in `http.STATUS_OK`)

Something like:
```
		{
			// this one works correctly
                         name:       "Unexpected ingesterPushError",
			maxMsgSize: 100000,
			series:     sampleSeries,
			metadata:   sampleMetadata,
			verifyFunc: func(*testing.T, context.Context, *Request, testCase) error {
				return ingesterPushError{message: "unexpected ingester error", cause: mimirpb.ERROR_CAUSE_BAD_DATA, soft: false}
			},
			responseCode:          http.StatusBadRequest,
			responseContentType:   pbContentType,
			responseContentLength: 29,
			errMessage:            "unexpected ingester error",
			expectedLogs:          []string{`level=warn user=test msg="detected an error while ingesting OTLP metrics request (the request may have been partially ingested)" httpCode=400 err="unexpected ingester error" insight=true`},
			expectedRetryHeader:   false,
		},
		{
			// this one needs a fix because the returned status has an unknown field
                         name:       "Unexpected soft ingesterPushError",
			maxMsgSize: 100000,
			series:     sampleSeries,
			metadata:   sampleMetadata,
			verifyFunc: func(*testing.T, context.Context, *Request, testCase) error {
				return ingesterPushError{message: "unexpected ingester error", cause: mimirpb.ERROR_CAUSE_BAD_DATA, soft: true}
			},
			responseCode:          http.StatusOK,
			responseContentType:   pbContentType,
			responseContentLength: 29,
			errMessage:            "unexpected ingester error",
			expectedRetryHeader: false,
		},
```

I tried to execute these tests myself. The first one works, but for the second one (the soft error) I don't know how it is expected to unmarshal the result in order to see that the response will contain a `PartialSuccess` field with the expected error message.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
 * [BUGFIX] Querier: Fix possible panic when evaluating a nested subquery where the parent has no steps. #12524
 * [BUGFIX] Ingester: Fix a bug ingesters would get stuck in read-only mode after compactions. #12538
 * [BUGFIX] Ingester: Fix a bug where prepare-instance-ring-downscale endpoint would return an error while compacting and not read-only. #12548
+* [BUGFIX] OTLP: Return HTTP OK for partially rejected requests, e.g. due to OOO exemplars. #12579
```
duricanikolic, line 66: Is this still up-to-date? Now you map all soft errors to HTTP OK.

aknuds1: I changed it, the current wording is supposed to be correct. Soft errors correspond to "partially rejected requests".

pkg/distributor/otel.go
duricanikolic: Doesn't your new implementation now reject samples too? Given that you are trying to map all soft errors to HTTP OK?
Anyway, if we opt for returning the number of rejected samples, we need to do it via error details too. How do we get them? I have no clue? :)

CHANGELOG.md
aknuds1: I could also reword it as "partially accepted" requests to align with the OTLP spec's wording, if that's better.

pkg/distributor/otel.go
aknuds1: Should we maybe try to fix that problem after this PR? I think returning HTTP 200 is definitely an improvement on its own. I'm not sure whether any users would actually be blocked by `RejectedDataPoints` not being correctly filled in. There is on the other hand at least one known user affected by us returning the wrong error code for partially accepted requests.

pkg/mimirpb/mimir.proto
duricanikolic: It is ok for me. The most important thing is to test the behaviour of this change in a dev cell, to see if the expectations are met.

duricanikolic:
The PR looks much better now and I really like it.
Let's additionally improve the tests by adding the ones for soft errors, and we can merge this.

pkg/distributor/otel_test.go
aknuds1: Done, PTAL.

aknuds1:
> but for the second one (the soft error) I don't know how it is expected to unmarshal the result in order to see that the response will contain a PartialSuccess field with the expected error message

Thanks @duricanikolic - I will try to make it work.

aknuds1:
> The first one works, but for the second one (the soft error) I don't know how it is expected to unmarshal the result in order to see that the response will contain a PartialSuccess field with the expected error message.

@duricanikolic I fixed also the second test case. PTAL.

duricanikolic (approved):
LGTM
Thank you for addressing my findings.

PR #12583, opened 2025-09-02
distributor: Add length to "label value too long" error

tcard:
#### What this PR does

* Show limit in label value length error for non-OTLP too.
* Show length of the offending label value in label value length error.

—

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/validation/limits.go
```diff
@@ -79,6 +79,19 @@ var (
 	errInvalidIngestStorageReadConsistency         = fmt.Errorf("invalid ingest storage read consistency (supported values: %s)", strings.Join(api.ReadConsistencies, ", "))
 	errInvalidMaxEstimatedChunksPerQueryMultiplier = errors.New("invalid value for -" + MaxEstimatedChunksPerQueryMultiplierFlag + ": must be 0 or greater than or equal to 1")
 	errNegativeUpdateTimeoutJitterMax              = errors.New("HA tracker max update timeout jitter shouldn't be negative")
+
```
56quarters: I would rather keep the existing error messages than have to rely on reflection to generate them.

tcard: How about I just copy the struct tags and then add tests to ensure they don't drift apart?

56quarters: Sure, that works

tcard: See 3bda9c1606

pkg/util/validation/limits.go
```diff
@@ -75,6 +75,22 @@ const (
 	MinCompactorPartialBlockDeletionDelay = 4 * time.Hour
 )
 
+// NOTE: Keep those in sync with Limits struct tags and TestParameterConstantsMatchStructTags.
```
56quarters:
```suggestion
// NOTE: Keep these in sync with Limits struct tags and TestParameterConstantsMatchStructTags.
```

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@
 * [ENHANCEMENT] Ingester: Skip read path load shedding when an ingester is the only available replica. #12448
 * [ENHANCEMENT] Querier: Include more information about inflight queries in the activity tracker. A querier logs this information after it restarts following a crash. #12526
 * [ENHANCEMENT] Ingester: Add experimental `-blocks-storage.tsdb.index-lookup-planning-comparison-portion` flag to enable mirrored chunk querier comparison between queries with and without index lookup planning. #12460
+* [ENHANCEMENT] Distributor: Show per-tenant configuration parameters instead of CLI flags in validation error messages where applicable. #12583
```
56quarters: I don't understand why this change is being made. Why is it better to show the per-tenant limit instead of the CLI flag or saying "contact the admin"? The problem with needing to keep tags in sync with struct members doesn't exist if we use CLI flags.

tcard: Because there is no such thing as a per-tenant CLI flag, so showing the config parameter instead makes the error more correct and actionable.

56quarters: > Because there is no such thing as a per-tenant CLI flag

Yes, I'm aware. The CLI flag is an option for people running their own Mimir instances

tcard:
I've reduced the scope of this PR to just adding the label value length to the error messages. PTAL

PR #12592, opened 2025-09-03
Query-frontend: more efficient expression-walking code

bboreham:
#### What this PR does

Now that Prometheus exposes an iterator we can skip a couple of layers of indirection and run just the code that we need.

This is more efficient in particular because `parser.Walk` would allocate a slice for the expression path, which was never used.

Maybe one of the tree-walking functions can go in a utility package?  I didn't see an obvious place.

I don't think it needs a CHANGELOG entry as it should be invisible to users.

#### Checklist

- [x] Tests updated.
- NA Documentation added.
- NA `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/subtree_folder.go
+++ b/pkg/frontend/querymiddleware/astmapper/subtree_folder.go
@@ -65,48 +65,28 @@ func isVectorSelector(n parser.Node) (bool, error) {
 }
 
 // anyNode is a helper which walks the input node and returns true if any node in the subtree
-// returns true for the specified predicate function.
+// returns true for the specified predicate function. If fn returns an error, return that.
 func anyNode(node parser.Node, fn predicate) (bool, error) {
-	v := &visitor{
-		fn: fn,
+	result, err := fn(node)
+	if result || err != nil {
+		return result, err
 	}
 
-	if err := parser.Walk(v, node, nil); err != nil {
-		return false, err
+	for node := range parser.ChildrenIter(node) {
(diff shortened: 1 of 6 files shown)


pkg/frontend/querymiddleware/astmapper/subtree_folder.go
```diff
@@ -65,48 +65,28 @@ func isVectorSelector(n parser.Node) (bool, error) {
...
+		if result || err != nil {
+			return result, err
+		}
 	}
-	return v.result, nil
+	return false, nil
 }
 
 // visitNode recursively traverse the node's subtree and call fn for each node encountered.
 func visitNode(node parser.Node, fn func(node parser.Node)) {
```
56quarters: Doesn't look like this is called in the `subtree_folder.go` file. Can we move it to the place it's called?

56quarters (approved):
Besides the same feeling that it would be nice to put these tree walking functions somewhere common, LGTM. Thanks!

pkg/frontend/querymiddleware/astmapper/subtree_folder.go
bboreham: After another look, I simplified `anyNode` and then `visitNode` was practially the same so I got rid of it.

PR #12603, opened 2025-09-04
ingester: Add IndexLookupPlannerFunc support

dimitarvdimitrov:
This PR vendors grafana/mimir-prometheus#973 which adds `IndexLookupPlannerFunc` support to TSDB.

## Summary

This change enables per-block index lookup planning by adding a function-based approach similar to `BlockChunkQuerierFunc`. Instead of using the same planner for all blocks, each block can now have its own planner instance. In this version all planner instances are effectively the same.

## Context

This is foundation work for implementing per-block statistics in Mimir ingester. Currently, we use head statistics for index lookup planning on all blocks (both head and disk blocks). This has a problem that when the head is empty we use empty stats for planning lookups to blocks, which results in very bad plans.

## Next Steps

After this vendoring is merged, the follow-up work will be:
- Implement per-block statistics provider
- Create background process to maintain block-specific statistics

related to https://github.com/grafana/mimir/issues/12604

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -330,7 +330,7 @@ require (
 	go.etcd.io/etcd/client/pkg/v3 v3.5.4 // indirect
 	go.etcd.io/etcd/client/v3 v3.5.4 // indirect
 	go.mongodb.org/mongo-driver v1.14.0 // indirect
-	go.opentelemetry.io/collector/semconv v0.128.0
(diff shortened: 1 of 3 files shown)


dimitarvdimitrov:
this is ready for a review, but i need https://github.com/grafana/mimir/pull/12670 to be merged, so we can also vendor the change from https://github.com/grafana/mimir-prometheus/pull/973

PR #12606, opened 2025-09-04
 query-frontend: enable labels-query-optimizer-enabled by default

narqo:
#### What this PR does

We've been running the `labels-query-optimizer-enabled` enabled in Grafana Labs production for a couple of months, and are confident now to make it the default. We still keep the feature-flag, just in case.

Note, I don't have much of context about the feature, and thus, I don't mind keeping the flag "experimental" for the next release (cc @pracucci).

Following #12080

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


CHANGELOG.md
```diff
@@ -194,7 +194,7 @@
...
-* [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. You can enable the optimizer per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076 #12080
+* [ENHANCEMENT] Query-frontend: Added labels query optimizer that automatically removes redundant `__name__!=""` matchers from label names and label values queries, improving query performance. You can enable the optimizer per-tenant with the `labels_query_optimizer_enabled` runtime configuration flag. #12054 #12066 #12076 #12080 #12606
```
56quarters: nit: I think making this the default should get its own changelog entry in the unreleased section

narqo: I didn't realize it was added in 2.17.0. Great spotting — updated the changelog

PR #12607, opened 2025-09-04
index planning: Implement per-block statistics

dimitarvdimitrov:
## Summary

Implements per-block statistics for IndexLookupPlanner instead of using the same head statistics for all blocks. This provides more accurate query planning for historical data by generating statistics specific to each block.

I was initially planning to keep this only for on-disk blocks, but this wasn't that easy (can explain if you're curious). So this PR replaces the stats generation in mimir-prometheus with stats generation in mimir. In a follow-up PR I will remove stats generation from mimir-prometheus.

This is built on top of https://github.com/grafana/mimir/pull/12603

## Using the index to build stats

I tried building statistics for a more sizeable block to see how fast it is

* 1.7M series
* 2h block
* generating stats took 2.2 seconds
* stats are 29MB

```
ts=2025-09-11T18:07:28.743546941Z caller=stats.go:49 level=info user=anonymous block=01K4WVRYG221YXPM277CQN56HY msg="generated statistics for block" duration=2.203544561s total_series=1740433 total_size_bytes=29027487
```

The size isn't great, especially given that the index size itself is 364MB; so stats are at ~10% of the index size. Some napkin math: ingesters have around 3M series and have 6 block such as this one. `29MB * 2 * 6 = 348MB`; while not perfect, it also shouldn't be detrimental 

## Telemetry

I added logs like the one above. I chose not to add metrics, because I'm not sure what would be a useful metric - a gauge with the last duration will get overwritten, a histogram will be mostly empty. I am hoping loki can be a good replacement for this information.

### DIFF
--- a/pkg/ingester/lookupplan/stats.go
+++ b/pkg/ingester/lookupplan/stats.go
@@ -40,14 +40,13 @@ func (g StatisticsGenerator) Stats(meta tsdb.BlockMeta, r tsdb.IndexReader) (ret
 	ctx := context.Background()
 
 	defer func(startTime time.Time) {
-		l := log.With(g.logger, "block", meta.ULID.String())
+		l := g.logger
 		if retErr != nil {
-			l = level.Error(l)
-			l = log.With(l, "err", retErr)
+			l = log.With(level.Error(l), "err", retErr)
 		} else {
 			l = level.Info(l)
 		}
-		l.Log("msg", "generated statistics for block", "duration", time.Since(startTime).String(), "total_series", retStats.TotalSeries(), "total_size_bytes", size.Of(retStats))
+		l.Log("msg", "generated statistics for block", "block", meta.ULID.String(), "duration", time.Since(startTime).String(), "total_series", retStats.TotalSeries(), "total_size_bytes", size.Of(retStats))
 	}(time.Now())
 
 	// Use the "all series" postings to count total series


pkg/ingester/lookupplan/stats_test.go
```diff
@@ -0,0 +1,431 @@
...
+	"github.com/prometheus/prometheus/tsdb"
+	"github.com/prometheus/prometheus/tsdb/chunks"
+	"github.com/prometheus/prometheus/tsdb/index"
+	"github.com/prometheus/prometheus/tsdb/tombstones"
+	"github.com/stretchr/testify/require"
+)
+
+const countMinEpsilon = 0.005
+
+// mockMemPostings is a simplified in-memory postings implementation for testing
```
chencs: Why not use/extend `NewMemPostings`, which already exists and is used for testing?

dimitarvdimitrov: i didn't realize it's so easy to use them. i still need some shim around it to provide an `IndexReader` interface, but now I'm using `index.NewMemPostings()` instead of running our own maps. see https://github.com/grafana/mimir/pull/12607/commits/e86786ba3778d1242d56592fa94995139ebdbb31

pkg/ingester/ingester.go
```diff
@@ -2630,16 +2617,29 @@ func (i *Ingester) getTSDB(userID string) *userTSDB {
...
+		return func(tsdb.BlockMeta, tsdb.IndexReader) index.LookupPlanner { return lookupplan.NoopPlanner{} }
+	}
+	metrics := lookupplan.NewMetrics(r)
+	statsGenerator := lookupplan.NewStatisticsGenerator(log.With(i.logger, "user", userID))
+
+	return func(meta tsdb.BlockMeta, reader tsdb.IndexReader) index.LookupPlanner {
+		stats, err := statsGenerator.Stats(meta, reader)
+		if err != nil {
+			level.Warn(i.logger).Log("msg", "failed to generate statistics; queries for this block won't use query planning", "user", userID, "block", meta.ULID.String(), "err", err)
+			stats = nil
```
chencs: Why not just return the noop planner early in the error block?

PR #12627, opened 2025-09-05
distributor: Add truncate and drop options for long label values

tcard:
#### What this PR does

At the moment, a label value over the length limit always causes an error. But most of the time, label values that long can just be truncated or dropped altogether.

This PR adds an option to decide a strategy to handle a label value too long error: truncate the label value, or drop the label altogether.

As a result, several timeseries that were before different may now be the same. To avoid duplicated samples, the hash of the (original) label value is set (at the end for "truncate", or fully replacing for "drop") as the label value.

See backreferences.

#### Checklist

- [x] Tests updated.
- [x] Documentation added.


aknuds1 (changes requested):
I think this PR doesn't correspond to the solution we agreed, because the strategy we want is instead to truncate too long label values, and then append a hashsum at the end of each truncated value.

I don't think the PR should _also_ add an alternate strategy, i.e. the "drop" strategy. It's better if we concentrate on the one strategy we agreed upon beforehand (see above).

aknuds1:
Taking another look.

pkg/distributor/distributor_test.go
```diff
@@ -6472,7 +6472,18 @@ func (i *mockIngester) Close() error {
 }
 
 func (i *mockIngester) Push(ctx context.Context, req *mimirpb.WriteRequest, _ ...grpc.CallOption) (*mimirpb.WriteResponse, error) {
-	i.trackCall("Push", ctx, req)
+	// Clone req by unmarhsalling and unmarshalling. Otherwise tests cannot
```
aknuds1:
```suggestion
	// Clone req by marshalling and unmarshalling. Otherwise tests cannot
```

pkg/distributor/validate.go
```diff
@@ -451,8 +455,11 @@ func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userI
 
 	maxLabelNameLength := cfg.MaxLabelNameLength(userID)
 	maxLabelValueLength := cfg.MaxLabelValueLength(userID)
+	labelValueLengthOverLimitStrategy := cfg.LabelValueLengthOverLimitStrategy(userID)
+
 	lastLabelName := ""
-	for _, l := range ls {
+	for i := range ls {
+		l := ls[i]
```
aknuds1: Unnecessary stylistic change:
```suggestion
	for i, l := range ls {
```

pkg/distributor/validate.go
```diff
@@ -466,12 +473,21 @@ func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userI
...
+			case validation.LabelValueLengthOverLimitStrategyError:
+				cat.IncrementDiscardedSamples(ls, 1, reasonLabelValueTooLong, ts)
+				m.labelValueTooLong.WithLabelValues(userID, group).Inc()
+				return labelValueTooLongError{
+					Label:  labels.Label{Name: strings.Clone(l.Name), Value: strings.Clone(l.Value)},
+					Series: mimirpb.FromLabelAdaptersToString(ls),
+					Limit:  maxLabelValueLength,
+				}
+			case validation.LabelValueLengthOverLimitStrategyTruncate:
+				_ = hashLabelValueInto(l.Value[len(l.Value)-validation.LabelValueHashLen:], l.Value)
```
aknuds1: This is a buggy implementation because the label value length is exceeded. The following implementation should work:
```suggestion
				if _, err := hashLabelValueInto(l.Value[maxLabelValueLength-validation.LabelValueHashLen:], l.Value); err != nil {
					return err
				}
				ls[i].Value = ls[i].Value[0:maxLabelValueLength]
```

pkg/distributor/validate.go
```diff
@@ -481,9 +497,32 @@ func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userI
...
+	hash := hashLabelValue(src)
+	buf := unsafeMutableStringToBytes(dst)
+	// Encode as hex inline instead of fmt.Sprintf to avoid allocations due to interface values.
+	copy(buf, "(hash:")
+	for i := range 16 {
+		buf[6+i] = "0123456789abcdef"[(hash>>(60-4*i))&0xf]
+	}
+	buf[22] = ')'
+	return dst[:validation.LabelValueHashLen]
+}
```
aknuds1: Let's simplify and not ignore errors:
```suggestion
func hashLabelValueInto(dst, src mimirpb.UnsafeMutableString) (mimirpb.UnsafeMutableString, error) {
	h := fnv.New64a()
	if _, err := h.Write(unsafeMutableStringToBytes(src)); err != nil {
		return "", fmt.Errorf("hash label value into: %w", err)
	}
	hash := h.Sum64()

	buf := unsafeMutableStringToBytes(dst)
	// Encode as hex inline instead of fmt.Sprintf to avoid allocations due to interface values.
	copy(buf, "(hash:")
	for i := range 16 {
		buf[6+i] = "0123456789abcdef"[(hash>>(60-4*i))&0xf]
	}
	buf[22] = ')'
	return dst, nil
}
```

pkg/distributor/validate_test.go
```diff
@@ -96,17 +107,25 @@ func TestValidateLabels(t *testing.T) {
...
-	for _, userID := range []string{defaultUserID, utf8UserID} {
+	for _, userID := range []string{defaultUserID, utf8UserID, truncatingUserID, droppingUserID} {
 		limits := *limits
 		perTenant[userID] = &limits
 	}
 	perTenant[defaultUserID].NameValidationScheme = model.LegacyValidation
 	perTenant[utf8UserID].NameValidationScheme = model.UTF8Validation
 
+	require.NoError(t, perTenant[truncatingUserID].LabelValueLengthOverLimitStrategy.Set("truncate"))
+	perTenant[truncatingUserID].MaxLabelNamesPerSeries = 5
```
aknuds1: I don't think you should set this?
```suggestion
```

pkg/distributor/validate_test.go
```diff
@@ -96,17 +107,25 @@ func TestValidateLabels(t *testing.T) {
...
+	perTenant[truncatingUserID].MaxLabelNameLength = 40
```
aknuds1: You must mean the following?
```suggestion
	perTenant[truncatingUserID].MaxLabelValueLength = 40
```

pkg/distributor/validate_test.go
```diff
@@ -96,17 +107,25 @@ func TestValidateLabels(t *testing.T) {
...
+
+	require.NoError(t, perTenant[droppingUserID].LabelValueLengthOverLimitStrategy.Set("drop"))
+	perTenant[droppingUserID].MaxLabelNameLength = 40
```
aknuds1: You must mean the following?
```suggestion
	perTenant[droppingUserID].MaxLabelValueLength = 40
```

pkg/distributor/validate_test.go
```diff
@@ -96,17 +107,25 @@ func TestValidateLabels(t *testing.T) {
...
+
 	overrides := func(limits *validation.Limits) *validation.Overrides {
 		return testutils.NewMockCostAttributionOverrides(*limits, perTenant, 0,
 			[]string{defaultUserID, "team"},
 			[]string{utf8UserID, "team"},
+			[]string{droppingUserID, "team"},
```
aknuds1:
```suggestion
			[]string{truncatingUserID, "team"},
			[]string{droppingUserID, "team"},
```

pkg/distributor/validate_test.go
```diff
@@ -157,7 +176,10 @@ func TestValidateLabels(t *testing.T) {
 		metric                   model.Metric
 		skipLabelNameValidation  bool
 		skipLabelCountValidation bool
-		wantErr                  func(model.ValidationScheme) error
+		customUserID             string
+
+		wantErr    func(model.ValidationScheme) error
+		wantLabels map[model.LabelName]model.LabelValue
```
aknuds1: Unnecessary stylistic change:
```suggestion
		wantErr    func(model.ValidationScheme) error
		wantLabels map[model.LabelName]model.LabelValue
```

pkg/distributor/validate_test.go
```diff
@@ -240,6 +262,32 @@ func TestValidateLabels(t *testing.T) {
...
+			name:                     "label value too long gets truncated",
+			metric:                   map[model.LabelName]model.LabelValue{model.MetricNameLabel: "badLabelValue", "much_shorter_name": "test_value_please_ignore_no_really_nothing_to_see_here", "team": "biz"},
+			skipLabelNameValidation:  false,
+			skipLabelCountValidation: false,
+			customUserID:             truncatingUserID,
+			wantLabels: map[model.LabelName]model.LabelValue{
+				model.MetricNameLabel: "badLabelValue",
+				"team":                "biz",
+				"group":               "custom label",
+				"much_shorter_name":   model.LabelValue(fmt.Sprintf("test_value_please_ignore_no_rea(hash:%016x)", hashLabelValue("test_value_please_ignore_no_really_nothing_to_see_here"))),
```
aknuds1: It's much better to use static test expectations rather than dynamic ones:
```suggestion
				"much_shorter_name":   "test_value_please(hash:c13fa87bc8c8c5bc)",
```

pkg/distributor/validate_test.go
```diff
@@ -240,6 +262,32 @@ func TestValidateLabels(t *testing.T) {
...
+			name:                     "label value too long gets dropped",
+			metric:                   map[model.LabelName]model.LabelValue{model.MetricNameLabel: "badLabelValue", "much_shorter_name": "test_value_please_ignore_no_really_nothing_to_see_here_2", "team": "biz"},
+			skipLabelNameValidation:  false,
+			skipLabelCountValidation: false,
+			customUserID:             droppingUserID,
+			wantLabels: map[model.LabelName]model.LabelValue{
+				model.MetricNameLabel: "badLabelValue",
+				"team":                "biz",
+				"group":               "custom label",
+				"much_shorter_name":   model.LabelValue(fmt.Sprintf("(hash:%016x)", hashLabelValue("test_value_please_ignore_no_really_nothing_to_see_here_2"))),
```
aknuds1:
```suggestion
				"much_shorter_name":   "(hash:0710946ebb65ac31)",
```

pkg/distributor/validate_test.go
```diff
@@ -353,14 +401,18 @@ func TestValidateLabels(t *testing.T) {
...
 		discardedSamplesValues := map[string]map[string]map[string]int{}
 		for _, c := range testCases {
 			if c.wantErr == nil {
 				continue
 			}
-			for _, scheme := range validationSchemes {
+			caseSchemes := validationSchemes
+			if c.customUserID != "" {
+				caseSchemes = []model.ValidationScheme{overrides(limits).NameValidationScheme(c.customUserID)}
+			}
```
aknuds1, line 411: Why change the case schemes for `customUserID != ""`?

pkg/distributor/validate_test.go
```diff
@@ -442,26 +496,30 @@ func TestValidateLabels(t *testing.T) {
...
-
-	// Run final checks after all subtests are done.
-	t.Cleanup(func() { finalChecks(t) })
+	})
 
 	for _, c := range testCases {
+		caseSchemes := validationSchemes
+		if c.customUserID != "" {
+			caseSchemes = []model.ValidationScheme{overrides(limits).NameValidationScheme(c.customUserID)}
+		}
```
aknuds1, line 502: Why change the case schemes for `customUserID != ""`?

pkg/distributor/validate_test.go
```diff
@@ -493,6 +551,20 @@ func TestValidateLabels(t *testing.T) {
 						go func() {
 							defer wg.Done()
 
+							expectedTraceID := trace.TraceID{}
+							_, _ = rand.Read(expectedTraceID[:])
```
aknuds1:
```suggestion
							_, err := rand.Read(expectedTraceID[:])
							require.NoError(t, err)
```

pkg/distributor/validate_test.go
```diff
@@ -513,6 +584,23 @@ func TestValidateLabels(t *testing.T) {
 							} else {
 								assert.Equal(t, 200, resp.Code, resp.Body.String())
 							}
+
+							if c.wantLabels != nil {
+								var gotReq *mimirpb.WriteRequest
+								ingesters[0].assertCalledFunc("Push", func(args ...any) {
+									ctx, req := args[0].(context.Context), args[1].(*mimirpb.WriteRequest)
+									traceID, _ := tracing.ExtractTraceID(ctx)
```
aknuds1:
```suggestion
									traceID, ok := tracing.ExtractTraceID(ctx)
									require.True(t, ok)
```

pkg/distributor/validate_test.go
```diff
@@ -1127,3 +1215,9 @@ func TestValidUTF8Message(t *testing.T) {
 		}
 	}
 }
+
+func TestLabelValueHashLen(t *testing.T) {
+	x := strings.Repeat("x", 100)
+	x = hashLabelValueInto(x, x)
+	require.Len(t, x, validation.LabelValueHashLen)
+}
```
aknuds1: The original test doesn't make much sense to me. BTW, why do the hexadecimal representations have to start with 0?
```suggestion
func TestLabelValueHashLen(t *testing.T) {
func TestHashLabelValueInto(t *testing.T) {
	x := strings.Repeat("x", 100)
	var err error
	x, err = hashLabelValueInto(x[:validation.LabelValueHashLen], x)
	require.NoError(t, err)
	require.Equal(t, "(hash:052c9e7cec411035", x)
}
```

pkg/mimirpb/timeseries.go
```diff
@@ -523,16 +523,16 @@ func (bs *LabelAdapter) Compare(other LabelAdapter) int {
 // references may change once the timeseries is returned to the shared pool.
 type UnsafeMutableLabel = LabelAdapter
 
-// A unsafeMutableString is a string that may violate the invariant that it's
-// immutable. Contrary to string, holding a value of unsafeMutableString may
+// A UnsafeMutableString is a string that may violate the invariant that it's
```
aknuds1:
```suggestion
// An UnsafeMutableString is a string that may violate the invariant that it's
```

pkg/util/validation/label_value_length_over_limit_strategy.go
```diff
@@ -0,0 +1,84 @@
...
+func (s LabelValueLengthOverLimitStrategy) String() string {
+	switch s {
+	case LabelValueLengthOverLimitStrategyError:
+		return "error"
+	case LabelValueLengthOverLimitStrategyTruncate:
+		return "truncate"
+	case LabelValueLengthOverLimitStrategyDrop:
+		return "drop"
+	default:
+		panic(fmt.Errorf("unrecognized LabelValueLengthOverLimitStrategy: %d", int(s)))
```
aknuds1: Unnecessary cast:
```suggestion
		panic(fmt.Errorf("unrecognized LabelValueLengthOverLimitStrategy: %d", s))
```

pkg/util/validation/label_value_length_over_limit_strategy_test.go
```diff
@@ -0,0 +1,79 @@
...
+		{
+			input: "drop",
+			want:  LabelValueLengthOverLimitStrategyDrop,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run("input="+tc.input, func(t *testing.T) {
+			for _, format := range []struct {
+				name string
+				set  func(*LabelValueLengthOverLimitStrategy, string) error
```
aknuds1:
```suggestion
				set  func(t *testing.T, *LabelValueLengthOverLimitStrategy, string) error
```

pkg/util/validation/label_value_length_over_limit_strategy_test.go
```diff
@@ -0,0 +1,79 @@
...
+			}{
+				{name: "json", set: func(v *LabelValueLengthOverLimitStrategy, input string) error {
+					b, _ := json.Marshal(input)
```
aknuds1: Errors shouldn't be ignored:
```suggestion
				{name: "json", set: func(t *testing.T, v *LabelValueLengthOverLimitStrategy, input string) error {
					t.Helper()
					b, err := json.Marshal(input)
					require.NoError(t, err)
```

pkg/util/validation/label_value_length_over_limit_strategy_test.go
```diff
@@ -0,0 +1,79 @@
...
+					return json.Unmarshal(b, &v)
+				}},
+				{name: "yaml", set: func(v *LabelValueLengthOverLimitStrategy, input string) error {
+					b, _ := yaml.Marshal(input)
```
aknuds1:
```suggestion
				{name: "yaml", set: func(t *testing.T, v *LabelValueLengthOverLimitStrategy, input string) error {
					t.Helper()
					b, err := yaml.Marshal(input)
					require.NoError(t, err)
```

pkg/util/validation/label_value_length_over_limit_strategy_test.go
```diff
@@ -0,0 +1,79 @@
...
+					return yaml.Unmarshal(b, &v)
+				}},
+				{name: "flag", set: func(v *LabelValueLengthOverLimitStrategy, input string) error {
+					return v.Set(input)
```
aknuds1:
```suggestion
				{name: "flag", set: func(_ *testing.T, v *LabelValueLengthOverLimitStrategy, input string) error {
					return v.Set(input)
```

pkg/util/validation/limits.go
```diff
@@ -81,7 +82,10 @@ var (
 	errNegativeUpdateTimeoutJitterMax              = errors.New("HA tracker max update timeout jitter shouldn't be negative")
 )
 
-const errInvalidFailoverTimeout = "HA Tracker failover timeout (%v) must be at least 1s greater than update timeout - max jitter (%v)"
+const (
+	errInvalidFailoverTimeout     = "HA Tracker failover timeout (%v) must be at least 1s greater than update timeout - max jitter (%v)"
+	errLabelValueHashExceedsLimit = "cannot set -" + LabelValueLengthOverLimitStrategyFlag + " to %q: label value hash would exceed max label value length of %d"
```
aknuds1:
```suggestion
	errLabelValueHashExceedsLimit = "cannot set -" + LabelValueLengthOverLimitStrategyFlag + " to %q: label value hash suffix would exceed max label value length of %d"
```

pkg/util/validation/limits.go
```diff
@@ -787,6 +806,12 @@ func (o *Overrides) MaxLabelValueLength(userID string) int {
 	return o.getOverridesForUser(userID).MaxLabelValueLength
 }
 
+// LabelValueLengthOverLimitStrategy returns maximum length a label value can be. This also is
+// the maximum length of a metric name.
```
aknuds1:
```suggestion
// LabelValueLengthOverLimitStrategy returns the strategy for when label values exceed the limit.
```

pkg/util/validation/limits_test.go
```diff
@@ -1777,6 +1777,46 @@ func TestLimits_Validate(t *testing.T) {
 			}(),
 			expectedErr: fmt.Errorf("OTLP translation strategy NoTranslation is not allowed unless metric suffixes are disabled"),
 		},
+		"should fail if label_value_length_over_limit_strategy=truncate and hash would be higher than max label value limit": {
```
aknuds1:
```suggestion
		"should fail if label_value_length_over_limit_strategy=truncate and hash suffix would be longer than max label value limit": {
```

pkg/util/validation/limits_test.go
```diff
@@ -1777,6 +1777,46 @@ func TestLimits_Validate(t *testing.T) {
...
+			cfg: func() Limits {
+				cfg := Limits{}
+				flagext.DefaultValues(&cfg)
+				cfg.LabelValueLengthOverLimitStrategy = LabelValueLengthOverLimitStrategyTruncate
+				cfg.MaxLabelValueLength = LabelValueHashLen - 1
+				return cfg
+			}(),
+			expectedErr: errors.New(`cannot set -validation.label-value-length-over-limit-strategy to "truncate": label value hash would exceed max label value length of 22`),
+		},
+		"should fail if label_value_length_over_limit_strategy=drop and hash would be higher than max label value limit": {
```
aknuds1:
```suggestion
		"should fail if label_value_length_over_limit_strategy=drop and hash suffix would be longer than max label value limit": {
```

pkg/util/validation/limits_test.go
```diff
@@ -1777,6 +1777,46 @@ func TestLimits_Validate(t *testing.T) {
...
+			cfg: func() Limits {
+				cfg := Limits{}
+				flagext.DefaultValues(&cfg)
+				cfg.LabelValueLengthOverLimitStrategy = LabelValueLengthOverLimitStrategyDrop
+				cfg.MaxLabelValueLength = LabelValueHashLen - 1
+				return cfg
+			}(),
+			expectedErr: errors.New(`cannot set -validation.label-value-length-over-limit-strategy to "drop": label value hash would exceed max label value length of 22`),
+		},
+		"should pass if label_value_length_over_limit_strategy=truncate and hash would be lower or equal than max label value limit": {
```
aknuds1:
```suggestion
		"should pass if label_value_length_over_limit_strategy=truncate and hash suffix would be shorter than or equal to max label value limit": {
```

pkg/util/validation/limits_test.go
```diff
@@ -1777,6 +1777,46 @@ func TestLimits_Validate(t *testing.T) {
...
+			cfg: func() Limits {
+				cfg := Limits{}
+				flagext.DefaultValues(&cfg)
+				cfg.LabelValueLengthOverLimitStrategy = LabelValueLengthOverLimitStrategyTruncate
+				cfg.MaxLabelValueLength = LabelValueHashLen
+				return cfg
+			}(),
+			expectedErr: nil,
+		},
+		"should pass if label_value_length_over_limit_strategy=drop and hash would be lower or equal than max label value limit": {
```
aknuds1:
```suggestion
		"should pass if label_value_length_over_limit_strategy=drop and hash suffix would be shorter than or equal to max label value limit": {
```

pkg/distributor/validate.go
tcard: This error is in `Hash.Write`'s signature just to comply with `io.Writer` signature, and [will always be nil](https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/hash/fnv/fnv.go;l=127). I think propagating it just adds noise.

I've also thought of having a specialized copy of `fnv.New64a` to avoid the allocation of creating a `hash.Hash` interface value. I think this interface as return type is just an unfortunate legacy design that has stuck for backwards-compatibility; the usual norm is to return concrete types rather than interfaces, and the stdlib usually abides by it, this case being an exception. 

At the same time, I could get rid of the error in the `Write` signature, as we don't really care about complying with `io.Writer`. WDYT?

pkg/distributor/validate_test.go
tcard: This is actually a leftover from the previous implementation. The default value of 25 is fine. I'll just remove it.

pkg/distributor/validate_test.go
tcard: This is actually a leftover from the previous implementation. The default value of 25 is fine. I'll just remove it.

pkg/distributor/validate_test.go
tcard: This actually doesn't have any effect, because we aren't discarding any samples from those two users in the test, so we have no `cortex_discarded_attributed_samples_total` from them. I'll remove `droppingUserID` too.

pkg/distributor/validate_test.go
tcard: Because the validation scheme used is defined by the user ID. Custom user IDs have a single validation scheme, whichever it may be. So test cases for each validation schemes wouldn't actually be using each validation scheme, as the same exact test would run twice, so the test case name would be lying.

pkg/distributor/validate.go
tcard: Fixed at [7e1e3b4](https://github.com/grafana/mimir/pull/12627/commits/7e1e3b4dab19cccd4a889035e38adcbc185738c4)

pkg/distributor/validate_test.go
tcard: My thinking was that we don't particularly care about the value of the hash, so a change in the hash function shouldn't cause the test to fail. But anyway, I've made it static.

pkg/distributor/validate_test.go
tcard: See https://github.com/grafana/mimir/pull/12627#discussion_r2351756204

pkg/distributor/validate_test.go
tcard: The value sput by `hashLabelValueInto` isn't really interesting, we only care that it has a predefined length. Hence why the test only checks the length.

tcard: I've added the value check anyway, plus checks that it actually writes into the same string, at a5ff955d7e.

pkg/util/validation/limits.go
```diff
@@ -131,23 +135,24 @@ type Limits struct {
...
+	DropLabels                                  flagext.StringSlice               `yaml:"drop_labels" json:"drop_labels" category:"advanced"`
+	MaxLabelNameLength                          int                               `yaml:"max_label_name_length" json:"max_label_name_length"`
+	MaxLabelValueLength                         int                               `yaml:"max_label_value_length" json:"max_label_value_length"`
+	LabelValueLengthOverLimitStrategy           LabelValueLengthOverLimitStrategy `yaml:"label_value_length_over_limit_strategy" json:"label_value_length_over_limit_strategy" category:"advanced" doc:"description=What to do for label values over the length limit. Options are: 'error', 'truncate', 'drop'. For 'truncate' and 'drop', the hash of the full value is added as a new label, named '<original label name><label_value_length_over_limit_hash_suffix>'."`
```
56quarters: Can you make this `experimental` for now since it's a new feature and we may decide to remove the option and make one particular behavior the default in the future?

pkg/distributor/validate.go
```diff
@@ -481,9 +497,29 @@ func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userI
 
 		lastLabelName = l.Name
 	}
+
 	return nil
 }
 
+func hashLabelValueInto(dst, src mimirpb.UnsafeMutableString) mimirpb.UnsafeMutableString {
```
56quarters, line 505: Can you please use a cryptographic hash for this (we use blake2b elsewhere)? There's an example including converting to a printable string [here](https://github.com/grafana/mimir/blob/5ef1d3711e61e491bcaeed77c30825f3ba9edd3e/pkg/storage/tsdb/bucketcache/caching_bucket.go#L799-L809). If performance is a concern, you could add a benchmark in the distributor to see how much it affects response time compared to everything else the distributor is doing.

aknuds1: Thanks for pointing this out @56quarters.

tcard: I've purposedly used a non-cryptographic hash because we only need it to keep the uniqueness from the input, not for any security feature. But anyway, I've used blake2b at [dd5718b](https://github.com/grafana/mimir/pull/12627/commits/dd5718b7b0d6c3f948e74eea2c4c5564c999fc4b).

I've kept the hex representation because I think it's more familiar for hashes, and the inline implementation to avoid the extra `string` allocation of calling `encoding/hex` (or `base64`). But no strong opinion about those.

pkg/distributor/validate.go
```diff
@@ -481,9 +497,29 @@ func validateLabels(m *sampleValidationMetrics, cfg labelValidationConfig, userI
...
+
+	buf := unsafeMutableStringToBytes(dst)
+	// Encode as hex inline instead of fmt.Sprintf to avoid allocations due to interface values.
+	copy(buf, "(hash:")
+	const hexChars = "0123456789abcdef"
+	for i, b := range h[:8] {
+		buf[6+i*2] = hexChars[b>>4]
+		buf[7+i*2] = hexChars[b&0x0f]
+	}
+	buf[22] = ')'
```
aknuds1: I think we should be using the _whole_ hash here. Given that the hash prefix is only 3.5% of the default label value length limit, it shouldn't be problem, right? Otherwise, if you want a shorter hash, I think you should be generating a shorter blake2b hash explicitly instead.

Also, do you think `hex.Encode` will be inefficient?
```suggestion
	_ = hex.Encode(buf[6:], h[:blake2b.Size256])
	buf[6+blake2b.Size256*2] = ')'
```

aknuds1 (changes requested):
I think the whole blake2b hash must be used.

pkg/distributor/validate.go
aknuds1: As I understand it @tcard, blake2b has much stronger collision resistance than fnv-1. That should be beneficial in this case, since we want identifying signatures.

pkg/distributor/validate.go
tcard: > Also, do you think `hex.Encode` will be inefficient?

Ah, I had somehow missed it!

Done both at [5de3bc7](https://github.com/grafana/mimir/pull/12627/commits/5de3bc7bc3e2de03cd7bc894cb7a1772cea1fa61)

pkg/util/validation/limits.go
```diff
@@ -648,9 +655,21 @@ func (l *Limits) Validate() error {
...
+		}
+	}
+
 	return nil
 }
 
+// LabelValueHashLen is the length of the hash portion that replaces part of all
+// of a label value when it exceeds [Limits.MaxLabelValueLength] and [Limits.LabelValueLengthOverLimitStrategy]
+// is [Limits.LabelValueLengthOverLimitStrategyTruncate] or  [Limits.LabelValueLengthOverLimitStrategyDrop].
+const LabelValueHashLen = len("(hash:)") + 256/8*2
```
aknuds1: [Nit] The following is more readable to me:
```suggestion
const LabelValueHashLen = len("(hash:)") + blake2b.Size256*2
```

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 * [FEATURE] Query-frontend: Add a native histogram presenting the length of query expressions handled by the query-frontend #12571
 * [FEATURE] Query-frontend and querier: Add experimental support for performing query planning in query-frontends and distributing portions of the plan to queriers for execution. #12302 #12551 #12665
 * [FEATURE] Alertmanager: add Microsoft Teams V2 as a supported integration. #12680
+* [FEATURE] Distributor: Add `label_value_length_over_limit_strategy` to configure how to handle label values over the length limit. #12627
```
aknuds1: I think we should be explicit about _what_ is added. I suggest we refer to the flag (the configuration parameter is implied, as the flag's "twin"), since that's the typical convention in the changelog:
```suggestion
* [FEATURE] Distributor: Add experimental flag `-validation.label-value-length-over-limit-strategy` to configure how to handle label values over the length limit. #12627
```

pkg/util/validation/limits.go
```diff
@@ -131,23 +135,24 @@ type Limits struct {
...
+	DropLabels                                  flagext.StringSlice               `yaml:"drop_labels" json:"drop_labels" category:"advanced"`
+	MaxLabelNameLength                          int                               `yaml:"max_label_name_length" json:"max_label_name_length"`
+	MaxLabelValueLength                         int                               `yaml:"max_label_value_length" json:"max_label_value_length"`
+	LabelValueLengthOverLimitStrategy           LabelValueLengthOverLimitStrategy `yaml:"label_value_length_over_limit_strategy" json:"label_value_length_over_limit_strategy" category:"experimental" doc:"description=What to do for label values over the length limit. Options are: 'error', 'truncate', 'drop'. For 'truncate' and 'drop', the hash of the full value is added as a new label, named '<original label name><label_value_length_over_limit_hash_suffix>'."`
```
aknuds1: The description is out of date:
```suggestion
	LabelValueLengthOverLimitStrategy           LabelValueLengthOverLimitStrategy `yaml:"label_value_length_over_limit_strategy" json:"label_value_length_over_limit_strategy" category:"experimental" doc:"description=What to do for label values over the length limit. Options are: 'error', 'truncate', 'drop'. For 'truncate', the hash of the full value replaces the end portion of the value. For 'drop', the hash fully replaces the value."`
```

aknuds1 (changes requested):
We're getting there, but please see suggestions.

PR #12628, opened 2025-09-05
ruler: Add native histogram version of cortex_ruler_sync_rules_duration_seconds

alexweav:
#### What this PR does

Exports `cortex_ruler_sync_rules_duration_seconds` as both classic and native histograms.
Improves observability on syncs; the bucketing scheme can make it difficult to see timing.

The native histogram configuration is fairly standard. This PR makes sure to align with other long-running background jobs with a wide bucketing scheme, e.g. [compactor histograms](https://github.com/grafana/mimir/blob/ad6f32f8076ce5532ffbb0409d33df2537ee4e98/pkg/compactor/bucket_compactor.go#L759-L782)

Contrib https://github.com/grafana/mimir-squad/issues/3077

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ruler/ruler_test.go
+++ b/pkg/ruler/ruler_test.go
@@ -2005,8 +2005,12 @@ func verifySyncRulesMetrics(t *testing.T, reg prometheus.Gatherer, initialCount,
 			cortex_ruler_sync_rules_total{reason="ring-change"} 0
 		`, initialCount, apiChangeCount)), "cortex_ruler_sync_rules_total")
 	})
-	require.NoError(t, promtest.HasNativeHistogram(reg, "cortex_ruler_sync_rules_duration_seconds"))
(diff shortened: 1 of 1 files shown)


pkg/ruler/ruler_test.go
```diff
@@ -2004,6 +2005,12 @@ func verifySyncRulesMetric(t *testing.T, reg prometheus.Gatherer, initialCount,
 			cortex_ruler_sync_rules_total{reason="ring-change"} 0
 		`, initialCount, apiChangeCount)), "cortex_ruler_sync_rules_total")
 	})
+	test.Poll(t, time.Second, nil, func() interface{} {
```
alexweav, line 2008: we have to poll because `sync_rules_total` gets updated at the beginning of the sync and `sync_rules_duration` gets updated at the end of the sync. If this assertion happens in the middle of a sync it'll see mismatched values

56quarters (approved):
LGTM

PR #12639, opened 2025-09-08
Ensure query-frontend safely handles the case where a response is received multiple times for the same query

charleskorn:
#### What this PR does

This PR builds on https://github.com/grafana/mimir/pull/12551.

This PR fixes an issue where the `QueryResultStream` method on frontends may panic or produce incorrect results if a response for the same query is received multiple times.

I've also added a test to confirm that it correctly handles the scenario where a response is received for a query ID that was used for a query from a different tenant.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/frontend/v2/frontend.go
```diff
@@ -592,23 +594,29 @@ func (f *Frontend) QueryResult(ctx context.Context, qrReq *frontendv2pb.QueryRes
...
+		return nil, errUnexpectedHTTPResponse
+	}
+
+	select {
+	case req.httpResponse <- queryResultWithBody{
+		queryResult: qrReq,
+	}:
+		// Should always be possible, unless QueryResult is called multiple times with the same queryID.
+	default:
+		level.Warn(f.log).Log("msg", "failed to write query result to the response channel", "queryID", qrReq.QueryID, "user", userID)
```
dimitarvdimitrov:
```suggestion
		level.Warn(f.log).Log("msg", "failed to receive write query result; the same result has likely been received already", "queryID", qrReq.QueryID, "user", userID)
```

talking about channels doesn't really tell you much unless you're reading the code

charleskorn: Good point, fixed in 39c2f1f

PR #12643, opened 2025-09-08
Block-builder: Don't exit the scheduler client's runloop during graceful shutdown.

seizethedave:
#### What this PR does

This fixes a bug where the scheduler client's runloop would be terminated during a graceful shutdown, thus terminating lease renewals. This means that most job completions inside a graceful shutdown would effectively lose the lease and have it start over reassigned to a different worker, followed by the completion request being rejected due to a bad assignment epoch.

#### Checklist

- (n/a) Tests updated.
- (n/a) Documentation added.


narqo (approved):
Thank you for investigating this 🔥

PR #12646, opened 2025-09-08
MQE: report query stats for remotely executed queries

charleskorn:
#### What this PR does

This PR builds on https://github.com/grafana/mimir/pull/12551.

This PR adds reporting of query stats for remotely executed queries, bringing the stats reported in line with those reported when remote execution is not used.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/querier/dispatcher.go
```diff
@@ -32,12 +33,14 @@ import (
...
 	"github.com/grafana/mimir/pkg/mimirpb"
 	"github.com/grafana/mimir/pkg/querier/querierpb"
+	querier_stats "github.com/grafana/mimir/pkg/querier/stats"
 	"github.com/grafana/mimir/pkg/streamingpromql"
 	"github.com/grafana/mimir/pkg/streamingpromql/types"
 	"github.com/grafana/mimir/pkg/util/limiter"
 )
 
 var evaluateQueryRequestMessageName = proto.MessageName(&querierpb.EvaluateQueryRequest{})
+var timeNow = time.Now
```
56quarters: This looks like it's only used from within `Dispatcher` and the `evaluationObserver` created by the `Dispatcher`. Maybe make it a member there and avoid the global variable cleanup logic.

```go
dispatcher = NewDispatcher(/* ... */)
dispatcher.timeNow = func() { return someNow }
```

pkg/streamingpromql/optimize/plan/remoteexec/operators.go
```diff
@@ -20,7 +21,21 @@ func finalize(ctx context.Context, resp RemoteExecutionResponse, annos *annotati
...
+				queryStats.IncrementSamplesAtTimestamp(step.Timestamp, step.Value)
+			}
+
+			// IncrementSamplesAtTimestamp updates TotalSamples, so there's nothing more to do.
+		} else {
+			queryStats.TotalSamples += int64(remoteStats.SamplesProcessed)
+		}
+	}
+
+	if localStats := stats.FromContext(ctx); localStats != nil {
```
56quarters, line 35: I don't quite understand what's going on here. We're combining stats from the local MQE engine and remote MQE engine... but then also updating querier `stats.Stats` for this request but only some fields? (I assume that if we updated all fields we'd be double counting somehow)

pkg/querier/dispatcher.go
charleskorn: Good idea, changed in 73a35a4

pkg/streamingpromql/optimize/plan/remoteexec/operators.go
charleskorn: I've tried to clarify this in e0ff751 - does that make it clearer?

charleskorn: Given the approval, I'm going to merge this, but let me know if you have any further feedback.

PR #12656, opened 2025-09-08
[compactor] Log sizes of downloaded and uploaded blocks

pr00se:
#### What this PR does

When troubleshooting e.g. compactor disk space issues it is useful to know which compactions are filling up the disk. Rather than add metrics, log some additional information in existing INFO logs.

Fixes #n/a

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/compactor/bucket_compactor.go
+++ b/pkg/compactor/bucket_compactor.go
@@ -473,7 +473,8 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
 			objstore.WithUploadConcurrency(c.maxPerBlockUploadConcurrency),
 		}
 
-		if err := block.Upload(ctx, jobLogger, c.bkt, bdir, nil, opts...); err != nil {
+		uploadedMeta, err := block.Upload(ctx, jobLogger, c.bkt, bdir, nil, opts...)
+		if err != nil {
 			var uploadErr *block.UploadError
 			if errors.As(err, &uploadErr) {
 				c.metrics.blockUploadsFailed.WithLabelValues(uploadErr.FileType()).Inc()
@@ -491,28 +492,12 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
 		sampleCount := uint64(0)
 		compactionLevel := 0
 
-		// Upload doesn't write the file stats to the local meta file, so we need to calculate the block size manually here.
-		files, err := block.GatherFileStats(bdir)
-		if err == nil {
-			for _, f := range files {
-				if f.SizeBytes > 0 {
-					blockSize += f.SizeBytes
-				}
-			}
-			// Use atomic to avoid race conditions in concurrent execution
+		if uploadedMeta != nil {
+			seriesCount = uploadedMeta.Stats.NumSeries
+			sampleCount = uploadedMeta.Stats.NumSamples
+			compactionLevel = uploadedMeta.Compaction.Level
+			blockSize = uploadedMeta.BlockBytes()
 			totalUploadedSize.Add(blockSize)
-		} else {
-			level.Warn(jobLogger).Log("msg", "failed to gather local file stats after upload", "block", blockToUpload.ulid, "err", err)
-		}
-
-		// Read info from local meta file
-		blockMeta, err := block.ReadMetaFromDir(bdir)
-		if err == nil {
-			seriesCount = blockMeta.Stats.NumSeries
-			sampleCount = blockMeta.Stats.NumSamples
-			compactionLevel = blockMeta.Compaction.Level
-		} else {
-			level.Warn(jobLogger).Log("msg", "failed to read local block metadata after upload", "block", blockToUpload.ulid, "err", err)
 		}
 
 		level.Info(jobLogger).Log(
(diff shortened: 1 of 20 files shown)


pkg/compactor/bucket_compactor.go
```diff
@@ -481,15 +485,54 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
 		c.metrics.blockUploadsDuration.WithLabelValues(jobType).Observe(elapsed.Seconds())
-		level.Info(jobLogger).Log("msg", "uploaded block", "result_block", blockToUpload.ulid, "duration", elapsed, "duration_ms", elapsed.Milliseconds(), "external_labels", labels.FromMap(blockToUpload.labels))
+
+		blockSize := int64(0)
+		seriesCount := uint64(0)
+		sampleCount := uint64(0)
+		compactionLevel := 0
+
+		// Upload doesn't write the file stats to the local meta file, so we need to calculate the block size manually here.
+		files, err := block.GatherFileStats(bdir)
```
dmwilson-grafana: Could we remove this block of code and just get `totalUploadedSize` from a call to `blockMeta.BlockBytes()` following L509?

pkg/compactor/bucket_compactor.go
```diff
@@ -481,15 +485,54 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
+				}
+			}
+			// Use atomic to avoid race conditions in concurrent execution
+			totalUploadedSize.Add(blockSize)
+		} else {
+			level.Warn(jobLogger).Log("msg", "failed to gather local file stats after upload", "block", blockToUpload.ulid, "err", err)
+		}
+
+		// Read info from local meta file
+		blockMeta, err := block.ReadMetaFromDir(bdir)
```
dmwilson-grafana: IIRC, `block.ReadMetaFromDir` can be somewhat expensive. Is it worth it to collect these stats after L414 to avoid the extra read?

pkg/compactor/bucket_compactor.go
pr00se: `BlockBytes()` will return `0` because `Upload()` doesn't write the results of `GatherFileStats()` back to the meta file on disk.

pkg/compactor/bucket_compactor.go
pr00se: We could, but i'm not wondering if we could kill two birds with on stone and just change `Upload()` to return the updated meta in addition to the error. It's a bit more work to change the function signature, but then we could just get everything out of that instead. What do you think?

dmwilson-grafana: What if instead of returning the updated meta we just took the pointer to meta as an argument and then use it after it's been updated? I think we can avoid changes to the signature of block.Upload and avoid duplicate calls by just doing the following:

```diff
-               if err := block.Upload(ctx, jobLogger, c.bkt, bdir, nil, opts...); err != nil {
+               var blockMeta block.Meta
+               if err := block.Upload(ctx, jobLogger, c.bkt, bdir, &blockMeta, opts...); err != nil {
```

and then in `pkg/storage/tsdb/` something like the following before `GatherFileStats`

```diff
--- a/pkg/storage/tsdb/block/block.go
+++ b/pkg/storage/tsdb/block/block.go
@@ -131,6 +131,14 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blockDi                                                                                                                                                       
+       } else if meta.ULID == (ulid.ULID{}) {  // there's probably a better check...
+               readMeta, err := ReadMetaFromDir(blockDir)
+               if err != nil {
+                       return errors.Wrap(err, "read meta")
+               }
+               *meta = *readMeta
+       }
```

tacole02 (approved):
Changelog looks good.

pkg/compactor/bucket_compactor.go
```diff
@@ -481,15 +486,38 @@ func (c *BucketCompactor) runCompactionJob(ctx context.Context, job *Job) (shoul
...
 		c.metrics.blockUploadsDuration.WithLabelValues(jobType).Observe(elapsed.Seconds())
-		level.Info(jobLogger).Log("msg", "uploaded block", "result_block", blockToUpload.ulid, "duration", elapsed, "duration_ms", elapsed.Milliseconds(), "external_labels", labels.FromMap(blockToUpload.labels))
+
+		blockSize := int64(0)
+		seriesCount := uint64(0)
+		sampleCount := uint64(0)
+		compactionLevel := 0
+
+		if uploadedMeta != nil {
+			seriesCount = uploadedMeta.Stats.NumSeries
```
dmwilson-grafana, line 496: This can panic if `Stats` is nil. May want to check that before accessing `uploadedMeta.Stats.*`.

pkg/storage/tsdb/block/block.go
```diff
@@ -188,9 +189,9 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blockDi
 		// and even though cleanUp will not see it yet, meta.json may appear in the bucket later.
 		// (Eg. S3 is known to behave this way when it returns 503 "SlowDown" error).
 		// If meta.json is not uploaded, this will produce partial blocks, but such blocks will be cleaned later.
-		return &UploadError{err, FileTypeMeta}
+		return nil, &UploadError{err, FileTypeMeta}
 	}
-	return nil
+	return meta, nil
```
dmwilson-grafana, line 194: nit: In the past, we have seen meta files as large as 1MB because `sources` can be thousands of ULID long. I don't think this will be a major issue, but we may want to watch GC.

pkg/compactor/bucket_compactor.go
pr00se: `Stats` isn't a [pointer](https://github.com/prometheus/prometheus/blob/main/tsdb/block.go#L179-L180), so it can't be `nil`.

pkg/compactor/bucket_compactor.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/thanos-io/objstore"
 	"github.com/thanos-io/objstore/providers/filesystem"
+	"go.uber.org/atomic"
```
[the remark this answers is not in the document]
pr00se: These are in order already

PR #12659, opened 2025-09-08
Improve memory consumption tracker accuracy for Labels

lamida:
Update: The PR is moved to https://github.com/grafana/mimir/pull/12984.

Must be merged only after rebasing https://github.com/grafana/mimir/pull/12878

Tracking memory consumption for Labels. 

* Memory consumption for data from ingester is in query.go
* Memory consumption for data from store-gateway is in block_store_queryable.go

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/series/memory_tracking_series_set.go
+++ b/pkg/storage/series/memory_tracking_series_set.go
@@ -13,7 +13,8 @@ import (
 type MemoryTrackingSeriesSet struct {
 	inner                    storage.SeriesSet
 	memoryConsumptionTracker *limiter.MemoryConsumptionTracker
-	memoryDecreased          bool
+	// This flag prevents multiple decrements for the same series position when At() is called multiple times.
+	memoryDecreased bool
 }
 
 func NewMemoryTrackingSeriesSet(inner storage.SeriesSet, memoryConsumptionTracker *limiter.MemoryConsumptionTracker) storage.SeriesSet {
@@ -24,7 +25,6 @@ func NewMemoryTrackingSeriesSet(inner storage.SeriesSet, memoryConsumptionTracke
 }
 
 func (m *MemoryTrackingSeriesSet) Next() bool {
-	// This flag is needed because some SeriesSet implementation might reuse internal pointer of currentSeries.
 	m.memoryDecreased = false
 	return m.inner.Next()
 }


pkg/distributor/query.go
```diff
@@ -380,6 +380,30 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
+					memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+				}
+			}
+
+			for _, ls := range res.streamingSeries.Series {
+				memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+			}
+		}
+
+	}()
```
charleskorn: Why is this needed? We should only call `DecreaseMemoryConsumptionForLabels` once the labels are no longer held in memory.

pkg/querier/distributor_queryable.go
```diff
@@ -122,17 +123,28 @@ func (q *distributorQuerier) Select(ctx context.Context, _ bool, sp *storage.Sel
...
+	memoryTracker := limiter.MemoryTrackerFromContextWithFallback(ctx)
+	set := q.streamingSelect(ctx, minT, maxT, matchers, memoryTracker)
+	for set.Next() {
+		memoryTracker.DecreaseMemoryConsumptionForLabels(set.At().Labels())
+	}
+	return set
 }
 
-func (q *distributorQuerier) streamingSelect(ctx context.Context, minT, maxT int64, matchers []*labels.Matcher) storage.SeriesSet {
+func (q *distributorQuerier) streamingSelect(ctx context.Context, minT, maxT int64, matchers []*labels.Matcher, memoryTracker *limiter.MemoryConsumptionTracker) storage.SeriesSet {
```
charleskorn: Won't the changes below double-count the memory consumption from labels already accounted for in `QueryStream`?

pkg/querier/distributor_queryable.go
```diff
@@ -122,17 +123,28 @@ func (q *distributorQuerier) Select(ctx context.Context, _ bool, sp *storage.Sel
 		return series.LabelsToSeriesSet(ms)
 	}
 
-	return q.streamingSelect(ctx, minT, maxT, matchers)
+	memoryTracker := limiter.MemoryTrackerFromContextWithFallback(ctx)
+	set := q.streamingSelect(ctx, minT, maxT, matchers, memoryTracker)
+	for set.Next() {
+		memoryTracker.DecreaseMemoryConsumptionForLabels(set.At().Labels())
+	}
+	return set
```
charleskorn: Why is this needed? And won't this result in returning an exhausted `SeriesSet`?

pkg/querier/blocks_store_queryable.go
```diff
@@ -919,6 +919,11 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
 			mtx.Lock()
 			if len(mySeries) > 0 {
 				seriesSets = append(seriesSets, &blockQuerierSeriesSet{series: mySeries})
+
+				for _, s := range mySeries {
+					ls := mimirpb.FromLabelAdaptersToLabels(s.Labels)
+					memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+				}
```
charleskorn: Why is this needed?

pkg/querier/blocks_store_queryable.go
```diff
@@ -930,6 +935,10 @@ func (q *blocksStoreQuerier) fetchSeriesFromStores(ctx context.Context, sp *stor
 					remoteAddress: c.RemoteAddress(),
 				})
 				streamReaders = append(streamReaders, streamReader)
+
+				for _, ls := range myStreamingSeriesLabels {
+					memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+				}
```
charleskorn, line 963: Why is this needed?

pkg/distributor/query.go
lamida: Should be fixed in the latest commit.

pkg/querier/distributor_queryable.go
lamida: Yes I was still not clear on the streaming chain in the beginning. Should be fixed in the latest commit.

pkg/querier/distributor_queryable.go
lamida: Should be fixed in the latest commit.

pkg/querier/blocks_store_queryable.go
lamida: Should be fixed in the latest commit.

pkg/querier/blocks_store_queryable.go
lamida: Should be fixed in the latest commit.

pkg/distributor/query.go
```diff
@@ -394,16 +394,36 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
 // * If the response has StreamingSeries, a slice is returned with the label sets of each series.
 // A bool is also returned to indicate whether the end of the stream has been reached.
-func (r *ingesterQueryResult) receiveResponse(stream ingester_client.Ingester_QueryStreamClient, queryLimiter *limiter.QueryLimiter) ([]labels.Labels, bool, error) {
+func (r *ingesterQueryResult) receiveResponse(stream ingester_client.Ingester_QueryStreamClient, queryLimiter *limiter.QueryLimiter, memoryTracker *limiter.MemoryConsumptionTracker) ([]labels.Labels, bool, error) {
 	resp, err := stream.Recv()
 	if err != nil {
 		return nil, false, err
 	}
-	defer resp.FreeBuffer()
+	defer func() {
```
lamida: I think we can optimise this because decreasing and increasing memory involve the same labelset.

pkg/querier/blocks_store_queryable.go
```diff
@@ -977,15 +977,33 @@ func (q *blocksStoreQuerier) receiveMessage(c BlocksStoreClient, stream storegat
 
 		return mySeries, myWarnings, myQueriedBlocks, myStreamingSeriesLabels, indexBytesFetched, false, false, err
 	}
-	defer resp.FreeBuffer()
+	defer func() {
```
lamida: I think we can optimise this because decreasing and increasing memory involve the same labelset.

lamida:
@charleskorn this PR is still not complete. Pending unit test and more testing to ensure the correctness of memory tracker placement. But from a bit manual testing I did, it seems it track the labels correctly. I will update the PR again once I am back from PTO. Please leave some comment if you see anything significantly wrong in this PR's current form.

pkg/distributor/query.go
```diff
@@ -394,16 +394,36 @@ func (d *Distributor) queryIngesterStream(ctx context.Context, replicationSets [
...
+			memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+		}
+		for _, series := range resp.Chunkseries {
+			ls := mimirpb.FromLabelAdaptersToLabels(series.Labels)
+			memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+		}
+		for _, series := range resp.StreamingSeries {
+			ls := mimirpb.FromLabelAdaptersToLabels(series.Labels)
+			memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+		}
```
charleskorn: Why do we decrease the memory consumption here? Won't we still be holding the labels in memory when this method returns?

pkg/querier/blocks_store_queryable.go
```diff
@@ -977,15 +977,33 @@ func (q *blocksStoreQuerier) receiveMessage(c BlocksStoreClient, stream storegat
...
+		if ss := resp.GetSeries(); ss != nil {
+			ls := mimirpb.FromLabelAdaptersToLabels(ss.Labels)
+			memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+		}
+		if sss := resp.GetStreamingSeries(); sss != nil {
+			for _, ss := range sss.Series {
+				ls := mimirpb.FromLabelAdaptersToLabels(ss.Labels)
+				memoryTracker.DecreaseMemoryConsumptionForLabels(ls)
+			}
+		}
```
charleskorn: Same as above: why do we decrease the memory consumption here? Won't we still be holding the labels in memory when this method returns?

lamida:
Thanks @charleskorn . I only catch it on how SeriesSet iterator passed around during the streaming process. I have been trying making the change there now and ensuing the correctness. Will try to have the thing fixed today and will ask another review by then.

lamida:
@charleskorn I am adding `memoryTrackingSeriesSet` which will call `DecreaseMemoryConsumptionForLabels` when `At()` is called. `At()` call somewhat an indication the SeriesSet just been advanced or going to be advanced. 

Some integration tests are still failing. I will fix it. But in the meantime do you think this approach looks OK or also still wrong?

pkg/storage/lazyquery/lazyquery.go
```diff
@@ -47,6 +49,10 @@ func NewLazyQuerier(next storage.Querier) storage.Querier {
 
 // Select implements Storage.Querier
 func (l LazyQuerier) Select(ctx context.Context, selectSorted bool, params *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
+	// Remote read call this directly hence it is better to initiate a memory tracker here.
+	memoryTracker := limiter.MemoryTrackerFromContextWithFallback(ctx)
+	ctx = limiter.AddMemoryTrackerToContext(ctx, memoryTracker)
```
charleskorn, line 54: Why do we need to add a memory consumption tracker here? Can't anything that needs the memory consumption tracker call `limiter.MemoryTrackerFromContextWithFallback(ctx)` itself?

pkg/streamingpromql/benchmarks/benchmarks.go
```diff
@@ -60,6 +62,10 @@ func (c BenchCase) Run(ctx context.Context, t testing.TB, start, end time.Time,
 		return nil, nil
 	}
 
+	// Initiate memoryConsumptionTracker in context so that downstream read path, including query via prometheus engine will see the same memoryConsumptionTracker
+	memoryConsumptionTracker := limiter.MemoryTrackerFromContextWithFallback(ctx)
+	ctx = limiter.AddMemoryTrackerToContext(ctx, memoryConsumptionTracker)
+
```
charleskorn, line 68: Same as above - why is this needed?

lamida: `limiter.MemoryTrackerFromContextWithFallback(ctx)` ([code])(https://github.com/grafana/mimir/blob/fbd01eeb12e357913ab1f1dc6b8343fd49c06a58/pkg/util/limiter/memory_consumption.go#L64-L71) will initiate memory tracker if it can't find in context. But we still need a step to add it to the context otherwise the call to `IncreaseMemoryConsumptionForLabels` and `DecreaseMemoryConsumptionForLabels` will use different instance of memoryConsumptionTracker hence resulting incorrect accounting.

In the normal request path, `AddMemoryTrackerToContext` only happen in `evaluator.Evaluate` ([code](https://github.com/grafana/mimir/blob/a5dfebbd85245275ec9e7977022fd82e361bfd31/pkg/streamingpromql/evaluator.go#L66)), which is before `query.Exec`, therefore in this test I added to context here.

pkg/storage/lazyquery/lazyquery.go
lamida: The [comment](https://github.com/grafana/mimir/pull/12659#discussion_r2386732190) applies also for this. The integration test in `querier_remote_read_test.go` fail without this.

pkg/streamingpromql/benchmarks/benchmarks.go
charleskorn: `Query.Exec()` calls `Evaluator.Evaluate()` - which code needs the memory consumption tracker from the context before `Evaluator.Evaluate()` runs?

pkg/storage/lazyquery/lazyquery.go
charleskorn: This makes sense for remote read calls.

I wonder if we should rejig this slightly to avoid hard-to-find bugs: change `limiter.MemoryTrackerFromContextWithFallback` to return a memory consumption tracker or fail if there isn't one in the context, and then introduce a new method that checks if the context already has a memory consumption tracker, and if it doesn't, then it creates the memory consumption tracker, adds it to the context and returns the new context.

lamida: I think I can try to add. Should we do it in the different PR or just add it here? I will try to have integration test with pedantic mode and check for more correctness of tracking in this PR today.

pkg/streamingpromql/benchmarks/benchmarks.go
lamida: I think I looked that incorrectly. Yes Exec call Eval not other way around in mqe path. The panic actually will be triggered on promengine path in this benchmark. Hence I needed to created one here to make sure a memory tracker existed in the context before query go to the promengine path. Do you have any suggestion how should we handle this?

lamida: Actually after saying this, I noticed that in promengine path, there is no place where the memory consumption is decreased after getting increased when receivingResponse in ingester. Assuming MQE will always be the default engine, I will have two options:
1. Disable memory tracking for prom engine
2. Put the place of memory consumption decrease for label in prom engine path

pkg/storage/lazyquery/lazyquery.go
charleskorn: A separate PR makes sense to me.

pkg/streamingpromql/benchmarks/benchmarks.go
charleskorn: > I noticed that in promengine path, there is no place where the memory consumption is decreased after getting increased when receivingResponse in ingester. 

I think disabling memory consumption tracking for Prometheus' engine is the simpler solution. In this case, keeping `MemoryTrackerFromContextWithFallback` might be the easiest way to do this - then the `Queryable` implementations can use that and will automatically get an unlimited tracker when Prometheus' engine is in use.

lamida: I want to correct slight inaccuracy in this statement.
> Actually after saying this, I noticed that in promengine path, there is no place where the memory consumption is decreased after getting increased when receivingResponse in ingester. Assuming MQE will always be the default engine, I will have two options:

The prom engine get panic, because we never add memoryConsumptionTracker to context for query going through promq.Engine. Hence when calling decrease and increase it will call different instance of memoryConsumptionTracker and will cause panic for decrease where memoryConsumptionTracker goes to 0.

pkg/storage/lazyquery/lazyquery.go
lamida: I created [this PR](https://github.com/grafana/mimir/pull/12878) to enforce having explicityly added memory consumption tracker in context instead of having fallback.

pkg/streamingpromql/benchmarks/benchmarks.go
lamida: In this same [PR](https://github.com/grafana/mimir/pull/12878) I also skip memory consumption tracking if query is coming via promql.Engine.

pkg/util/limiter/memory_consumption.go
```diff
@@ -20,6 +21,43 @@ const (
 	memoryConsumptionTracker contextKey = 0
 )
 
+// memoryTrackingSeriesSet is storage.SeriesSet that tracks the wrapped SeriesSet memory consumption.
+type memoryTrackingSeriesSet struct {
```
charleskorn: This feels out of place here - I think this is better suited to `pkg/storage` rather than this package.

pkg/util/limiter/memory_consumption.go
```diff
@@ -20,6 +21,43 @@ const (
...
+	m.memoryDecreased = false
+	return m.SeriesSet.Next()
+}
+
+func (m *memoryTrackingSeriesSet) At() storage.Series {
+	at := m.SeriesSet.At()
+
+	// Only decrease memory once per series position
+	if m.memoryConsumptionTracker != nil && !m.memoryDecreased {
+		if m.currentSeries != at {
```
charleskorn: Why do we need both `m.memoryDecreased` and `m.currentSeries`? Isn't `m.currentSeries` enough?

pkg/util/limiter/memory_consumption.go
```diff
@@ -20,6 +21,43 @@ const (
...
+			// New series at this position
+			m.currentSeries = at
+			m.memoryDecreased = true
+			defer m.memoryConsumptionTracker.DecreaseMemoryConsumptionForLabels(at.Labels())
```
charleskorn: Why does this need to be deferred?

charleskorn: Could you please add a comment here explaining that we expect the caller to call `IncreaseMemoryConsumption` if they will retain the labels?

pkg/util/limiter/memory_consumption.go
charleskorn: It would be good to see some tests for this type.

pkg/util/limiter/memory_consumption.go
```diff
@@ -20,6 +21,43 @@ const (
 	memoryConsumptionTracker contextKey = 0
 )
 
+// memoryTrackingSeriesSet is storage.SeriesSet that tracks the wrapped SeriesSet memory consumption.
+type memoryTrackingSeriesSet struct {
+	storage.SeriesSet
```
charleskorn: I think it'd be better to explicitly implement the `storage.SeriesSet` interface in case other methods are added in the future, rather than embedding the value like this - this will force us to stop and think about if we need to do anything to support these new methods correctly.

pkg/querier/distributor_queryable_test.go
```diff
@@ -903,7 +937,18 @@ func (m *mockDistributor) QueryExemplars(ctx context.Context, from, to model.Tim
...
+	response := args.Get(0).(client.CombinedQueryStreamResponse)
+	// Simulate how memoryConsumptionTracker track memory utilisation as we process response from ingester
+	for _, ts := range response.Timeseries {
+		m.memoryConsumptionTracker.IncreaseMemoryConsumptionForLabels(mimirpb.FromLabelAdaptersToLabels(ts.Labels)) //nolint:errcheck
+	}
+	for _, cs := range response.Chunkseries {
+		m.memoryConsumptionTracker.IncreaseMemoryConsumptionForLabels(mimirpb.FromLabelAdaptersToLabels(cs.Labels)) //nolint:errcheck
+	}
+	for _, ss := range response.StreamingSeries {
+		m.memoryConsumptionTracker.IncreaseMemoryConsumptionForLabels(ss.Labels) //nolint:errcheck
```
charleskorn, line 949: Should we check the errors here? Might lead to hard-to-debug problems later if we don't.

pkg/util/limiter/memory_consumption.go
lamida: Moved to `pkg/storage/series` in 0642442125d48cda7c0272d8a6d55050f95f21b2

lamida: Added in 479d6f7cfaca033b72e6eb68f3200a6a0c6bccf

pkg/util/limiter/memory_consumption.go
lamida: Implemented in 5aa047e69d373a34fb00159532d378bafa584ceb

pkg/util/limiter/memory_consumption.go
lamida: After I checked again we can do only with `memoryDecreased` check because we can't rely on `m.currentSeries != at` because some `SeriesSet` implementation reused the returned object from At() method. 

Removed in 2ed97cbba18f6f5e21b0b55711945ca695875493

pkg/querier/distributor_queryable_test.go
lamida: Actually it is just test code which we don't expect the err to happen. Do we still need them nonetheless?

pkg/util/limiter/memory_consumption.go
lamida: Yeah it isn't needed. Removed in 43058a7a296da5274ef48126d868a0f3f468f3e3

lamida: Added in 9ee34e5278059464bf9d50ea12a405566b8febf3

lamida:
@charleskorn I have addressed your last feedbacks. Ideally you should only review this again after I rebase https://github.com/grafana/mimir/pull/12878 once that one is approved.

lamida:
Closing this and continue the PR in https://github.com/grafana/mimir/pull/12984

PR #12660, opened 2025-09-08
continuous-test: Equate NaN in all possible sub-fields in Timeseries

alexweav:
#### What this PR does

Generally we're using a copy of gogoproto's message comparison to assert equality in objects. Gogoproto just uses standard Go rules for comparing floats, meaning NaNs aren't equal, thus any object containing a NaN field always fails the comparison.

Previously we added logic to equate NaNs in samples. I had hoped this was "good enough" given this tool's somewhat temporary nature, but in one test cell we found a native histogram with a NaN sum being frequently reported. Best not concern ourselves with it going forward. This PR expands equating NaNs to all floating point fields in exemplars and histograms.

This is an artifact of how we are comparing timeseries in the test. Previously we tested different approaches to equate timeseries, in general most other comparison methods tended to use reflection and were far too slow, or choked on various pointer comparisons and needed significant hand-holding. The generated proto comparision was the fastest and most consistently accurate, despite the issue we're correcting here.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/continuoustest/ingest_storage_record.go
+++ b/pkg/continuoustest/ingest_storage_record.go
@@ -478,7 +478,7 @@ metaLoop:
 	return result
 }
 
-// TimeseriesEqual is a copy of mimirpb.TimeSeries.Equal that calls SampleEqual and ExemplarEqual instead.
+// TimeseriesEqual is a copy of mimirpb.TimeSeries.Equal that calls SampleEqual, ExemplarEqual, and HistogramEqual instead.
 func TimeseriesEqual(this *mimirpb.TimeSeries, that interface{}) bool {
 	if that == nil {
 		return this == nil


pkg/continuoustest/ingest_storage_record.go
```diff
@@ -559,11 +559,165 @@ func SampleEqual(this *mimirpb.Sample, that interface{}) bool {
...
-	if math.IsNaN(this.Value) && math.IsNaN(that1.Value) {
-		return true
+	if !floatEqualsEquateNaN(this.Value, that1.Value) {
+		return false
+	}
+	return true
+}
+
+// ExemplarEqual is a copy of mimirpb.Exemplar.Equal but equates NaN values.
+func ExemplarEqual(this *mimirpb.Exemplar, that interface{}) bool {
```
alexweav, line 569: These are all copy pasted from generated proto comparison code, aside from the line where we call
`!floatEqualsEquateNaN(x, y)` instead of `x != y`. No need to deeply review.

pkg/continuoustest/ingest_storage_record_test.go
```diff
@@ -370,3 +371,302 @@ func TestDropExactDuplicatesPreservesOrder(t *testing.T) {
 	assert.Equal(t, "metric3", result[2].MetricFamilyName)
 	assert.Equal(t, 3, len(result))
 }
+
+func TestExemplarEqual(t *testing.T) {
+	t.Run("equal", func(t *testing.T) {
```
alexweav, line 376: Not intending to test every branch since the code is mostly generated by gogoproto. We only deeply test the manual NaN branches + smoke test the rest.

PR #12661, opened 2025-09-09
Add a `handler` label to the uncompressed payload size histogram

seizethedave:
#### What this PR does

This adds a `handler` label to the existing histogram metric `cortex_distributor_uncompressed_request_body_size_bytes_sum`. It is currently shared by both the OTLP handler and the regular push handler, which makes it hard to understand which is producing the large payloads.

#### Checklist

- [x] Tests updated.
- (n/a) Documentation added.


dmwilson-grafana:
LGTM. I think It would be ideal if we could have either separate tracking for each handler or unified tracking for all metrics w. a `"handler"` label. I think (provided we update some E2E tests) either of the below could work, but it doesn't need to be on this PR.

```go
// by handler
type PushMetrics struct {
	influxRequestCounter       *prometheus.CounterVec
	influxUncompressedBodySize *prometheus.HistogramVec
	otlpRequestCounter         *prometheus.CounterVec
	otlpUncompressedBodySize   *prometheus.HistogramVec
	pushRequestCounter         *prometheus.CounterVec
	pushUncompressedBodySize   *prometheus.HistogramVec
}

// unified
type PushMetrics struct {
	requestCounter       *prometheus.CounterVec
	uncompressedBodySize *prometheus.HistogramVec
}
```

dmwilson-grafana (approved):

seizethedave:
@dmwilson-grafana That would be a good end-state. A more involved change, probably, due to dashboard dependencies and the like.

PR #12663, opened 2025-09-09
Active series matchers optimizations fix

pr00se:
#### What this PR does

This is a resubmission of https://github.com/grafana/mimir/pull/12184, fixing a bug in the original implementation.

---
**Original PR description:**

Optimizes the performance of `activeseries.Matchers.Matches()` and adds real-life benchmarks.

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/ingester/activeseries/model
cpu: Apple M3 Pro
                                          │  main.mem   │            optimized.mem            │
MatchesSeriesRealTrackers/match_series-12   80.20m ± 2%   22.68m ± 3%  -71.73% (p=0.000 n=10)

                                          │   main.mem   │            optimized.mem             │
                                          │     B/op     │     B/op      vs base                │
MatchesSeriesRealTrackers/match_series-12   18.47Ki ± 0%   16.39Ki ± 0%  -11.25% (p=0.000 n=10)

                                          │  main.mem   │          optimized.mem          │
                                          │  allocs/op  │  allocs/op   vs base            │
MatchesSeriesRealTrackers/match_series-12   2.000k ± 0%   2.000k ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

When creating the Matchers, we'll now split them in three groups:
- First, separate trackers that match on a single label
- Second, separate trackers that have an exact value matcher: either it's an equality or it's a set-matcher (`foo|bar`).
- The rest are called unoptimized matchers.

If we have matchers in the first two groups, we iterate the labels of the series and check for each label whether there's something matching: this way we avoid the lookup of the label value for each one of those matchers.

---

The original implementation had a bug where matchers of the form `foo=~"(bar|baz)"` and `foo!~"(bar|baz)"` were treated as equivalent, resulting in incorrect tracker counts.

This is addressed in 8ed480383731dbb81e6838b98ee5d2e460db52db

Fixes #

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


tacole02 (approved):
Changelog LGTM

narqo (approved):
I mainly looked at the diff between #12184 and this one: the changes looks good to me 🔥

PR #12666, opened 2025-09-09
OTLP push middleware

zeitlinger:
#### What this PR does

Adds an OTLP middleware to be able to intercept OTLP requests

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/distributor/distributor.go
```diff
@@ -264,6 +264,9 @@ type Config struct {
 	// These functions will only receive samples that don't get dropped by HA deduplication.
 	PushWrappers []PushWrapper `yaml:"-"`
 
+	// OtlpPushMiddlewares are wrappers that are called when an OTLP push request is received.
+	OtlpPushMiddlewares []OtlpPushMiddleware `yaml:"-"`
```
aknuds1: [Nit] Idiomatic Go naming for consistency :)
```suggestion
	// OTLPPushMiddlewares are wrappers that are called when an OTLP push request is received.
	OTLPPushMiddlewares []OTLPPushMiddleware `yaml:"-"`
```

tools/trafficdump/model.go
```diff
@@ -62,7 +62,7 @@ type pushRequest struct {
 	Error string `json:"error,omitempty"`
 }
 
-type otlpPushRequest struct {
+type OTLPPushRequest struct {
```
aknuds1: Can you drop this change?
```suggestion
type otlpPushRequest struct {
```

tools/trafficdump/parser.go
```diff
@@ -265,8 +265,8 @@ const (
 	jsonContentType = "application/json"
 )
 
-func (rp *parser) decodeOTLPRequest(req *http.Request, body []byte) (*otlpPushRequest, bool) {
-	res := &otlpPushRequest{}
+func (rp *parser) decodeOTLPRequest(req *http.Request, body []byte) (*OTLPPushRequest, bool) {
+	res := &OTLPPushRequest{}
```
aknuds1: Can you drop this change?
```suggestion
func (rp *parser) decodeOTLPRequest(req *http.Request, body []byte) (*otlpPushRequest, bool) {
	res := &otlpPushRequest{}
```

tools/trafficdump/model.go
zeitlinger: just noticed that you have this: https://github.com/grafana/mimir/blob/0c8a9c4952b4973116a102850c6fda4fe813ed7c/pkg/distributor/otel.go#L416

so `OtlpPush..` would be consistent with what you have already

aknuds1: @zeitlinger The code base could be more consistent, I agree :) If we keep diverging it will only get worse though. There are also examples to the opposite.

PR #12674, opened 2025-09-10
Remove remote execution debugging HTTP endpoint in queriers

charleskorn:
#### What this PR does

This PR removes the debugging endpoint introduced to queriers in https://github.com/grafana/mimir/pull/12302, as this is no longer necessary with https://github.com/grafana/mimir/pull/12551 in place.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.


pkg/querier/dispatcher.go
```diff
@@ -305,11 +232,9 @@ func (w *queryResponseWriter) WriteError(ctx context.Context, typ mimirpb.QueryE
 }
 
 func (w *queryResponseWriter) write(ctx context.Context, resp *frontendv2pb.QueryResultStreamRequest) error {
-	if w.querierResponseMessageBytes != nil {
```
charleskorn, line 308: Note to reviewers: this check is no longer required as all requests are Protobuf requests and therefore these observers will be populated.

PR #12671, opened 2025-09-09
index lookup: reduce cost of tracing

dimitarvdimitrov:
#### What this PR does

instrumenting for traces creates a lot of allocations - about 30% of the existing query path allocations (without index lookup planning)

Screenshot 2025-09-05 at 14 04 35 (1)

This PR makes a few changes
* replaces the usage of `DebugLog` with `LogFields`: this avoid boxing by the go compiler and allocating the individual floats to the heap: instead they're on the stack in the `otlog.Field` structs
* replaces the slice of `[]any` with direct calls to `LogFields` - this again avoid putting the slice on the heap
  * this was easier to do if we put every plan in its own event 
* only prints matchers if the span is sampled by using a `log.Lazy`

### Benchmark

TLDR

```
                │  before.txt   │              after.txt     │
                │    sec/op     │    sec/op      vs base     │
geomean              40.42µ         25.37µ         -37.24%

                │  before.txt   │              after.txt     │
                │     B/op      │     B/op      vs base      │
geomean              6.038Ki        1.453Ki       -75.93%

                │ before.txt    │             after.txt      │
                │  allocs/op    │ allocs/op   vs base        │
geomean              137.9          27.63       -79.96%

```

<details><summary>BenchmarkCostBasedPlannerPlanIndexLookup</summary>
<p>

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/ingester/lookupplan
cpu: Apple M1 Pro
                                                                                                                        │  before.txt   │              after.txt               │
                                                                                                                        │    sec/op     │    sec/op      vs base               │
CostBasedPlannerPlanIndexLookup/simple_existing_metric_name-10                                                            3037.5n ± 22%   889.3n ±   1%  -70.72% (p=0.002 n=6)
CostBasedPlannerPlanIndexLookup/simple_existing_label_name-10                                                             3568.0n ± 44%   796.0n ±   2%  -77.69% (p=0.002 n=6)
CostBasedPlannerPlanIndexLookup/simple_existing_label_name#01-10                                                          108.74µ ±  9%   95.06µ ±   1%  -12.58% (p=0.002 n=6)
CostBasedPlannerPlanIndexLookup/simple_low_cardinality_label_name-10                                                       22.82µ ±  7%   18.69µ ±   1%  -18.08% (p=0.002 n=6)
[...]

PR #12675, opened 2025-09-10
MQE: include details of query planning in traces

charleskorn:
#### What this PR does

This PR improves the traces emitted during the query planning process. 

Specifically, the original expression, optimised expression, original plan and optimised plan will be added as events on the query planning span.

I've opted not to add a changelog entry given this is not a user-facing change.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/streamingpromql/planning.go
+++ b/pkg/streamingpromql/planning.go
@@ -11,7 +11,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/grafana/dskit/tracing"
+	"github.com/go-kit/log"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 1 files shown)


pkg/streamingpromql/planning.go
```diff
@@ -170,6 +177,8 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 		return nil, err
 	}
 
+	spanLogger.DebugLog("msg", "AST optimisation passes complete", "expression", expr)
```
lamida: Nits. Shouldn't this be "completed" to be consistent with other spanloggers too??
```suggestion
	spanLogger.DebugLog("msg", "AST optimisation passes completed", "expression", expr)
```

lamida (approved):
One nit comment but other than that LGTM.

56quarters:
> it's a bit different for the QF and the MQE, but consider logging this only when the trace is sampled. In the ingester we wasted a lot of allocations logging plans for traces which aren't sampled (see #12671)

Span logging [has an optimization](https://github.com/grafana/dskit/blob/6b6da59b5cc4c5a7242f37fbf7ace0a62bea774b/spanlogger/spanlogger.go#L169-L171) to only log messages when the trace is sampled. Was the overhead from generating string representations of the logging arguments?

56quarters (approved):
LGTM

charleskorn:
> Was the overhead from generating string representations of the logging arguments?

I've deliberately passed the expression or plan as-is (without calling `String()`) so that they're only converted to strings if the trace is sampled.

dimitarvdimitrov:
> Was the overhead from generating string representations of the logging arguments?

in the ingester case - yes

PR #12676, opened 2025-09-10
Refactor query sharding logic in preparation for MQE implementation

charleskorn:
#### What this PR does

This PR refactors some parts of the query sharding logic in preparation for the MQE implementation of query sharding.

MQE's implementation of query sharding will re-use most of the existing logic. It will run as an AST optimisation pass that wraps sharded legs in a special function (likely named something like `concat` or `sharded`), rather than embedding queries in a specially-constructed vector selector.

To make this easier to implement, and reduce the size of the PR that implements it, in this PR, I'm doing some refactoring of the existing logic.

I've tried to break each change into its own commit for easier review.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/frontend/querymiddleware/astmapper/sharding_test.go
```diff
@@ -544,29 +544,29 @@ func concatShards(shards int, queryTemplate string) string {
 	queries := make([]EmbeddedQuery, shards)
 	for shard := range queries {
 		queryStr := strings.ReplaceAll(queryTemplate, "x_of_y", sharding.FormatShardIDLabelValue(uint64(shard), uint64(shards)))
-		queries[shard] = NewEmbeddedQuery(queryStr, nil)
+		expr, err := parser.ParseExpr(queryStr)
+		if err != nil {
+			panic(err)
```
56quarters: WDYT about adding a `T` param here and doing `require.NoError(t, err)` to make errors more readable?

pkg/frontend/querymiddleware/astmapper/subtree_folder.go
```diff
@@ -41,24 +43,13 @@ func (f *subtreeFolder) MapExpr(ctx context.Context, expr parser.Expr) (mapped p
...
-	case *parser.VectorSelector:
-		if n.Name == EmbeddedQueriesMetricName {
-			return true, nil
-		}
-	}
-	return false, nil
-}
-
-// hasEmbeddedQueries returns whether the expr is a vector selector.
+// isVectorSelector returns whether the expr is a vector selector.
```
56quarters, line 52: Can you add a comment explaining why this returns `(bool, error)` even though it can't ever return an error?

pkg/frontend/querymiddleware/astmapper/sharding_test.go
charleskorn: Good idea, changed in fee94f2

pkg/frontend/querymiddleware/astmapper/subtree_folder.go
charleskorn: Good idea, added in 89e7d35

PR #12652, opened 2025-09-08
Update mimir-prometheus to c2f03d6d14d2 for the OTLP endpoint rewrite

krajorama:
#### What this PR does

This PR updates mimir-prometheus to [`c2f03d6d14d2`.](https://github.com/grafana/mimir-prometheus/commit/c2f03d6d14d20b6f2f26cff2b8ce05c1ce3d7007)

Brings in https://github.com/prometheus/prometheus/pull/16951.

Changelog: OTLP: native support for OpenTelemetry metric start time to Prometheus metric created timestamp conversion, instead of converting to QuietZeroNaNs introduced in #10238. The configuration parameter `-distributor.otel-start-time-quiet-zero` is therefore deprecated and will be removed. Now supports start time for exponential histograms. This is a major rewrite of the endpoint in upstream Prometheus and Mimir. #12652

Implementation does away with the generated otlp translation code as we don't need to rewrite it for `mimirpb` anymore. The storage logic is behind an interface.

#### Which issue(s) this PR fixes or relates to

(none, internal)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [x] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

Succeeds #12152

PR #12679, opened 2025-09-11
Remove references to make generate-otlp target

dimitarvdimitrov:
## Summary

Remove references to the `make generate-otlp` target from the GitHub workflow.

## Context

PR #12652 removed the `generate-otlp` make target, but the 
`update-vendored-mimir-prometheus.yml` workflow still referenced it, 
causing automation failures when the workflow runs.

### DIFF
--- a/.github/workflows/update-vendored-mimir-prometheus.yml
+++ b/.github/workflows/update-vendored-mimir-prometheus.yml
@@ -143,7 +143,6 @@ jobs:
           
           echo "Dependencies updated successfully"
 
-
       - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


.github/workflows/update-vendored-mimir-prometheus.yml
```diff
@@ -143,8 +143,6 @@ jobs:
           
           echo "Dependencies updated successfully"
 
-      - name: Regenerate OTLP files
-        run: make generate-otlp
```
dimitarvdimitrov:
```suggestion
```

PR #12681, opened 2025-09-11
Ingester: Introduce metric cortex_ingester_owned_target_info_series

aknuds1:
#### What this PR does

Introduce the `cortex_ingester_owned_target_info_series` metric, for counting of owned `target_info` series per tenant. The motivation is to make it possible to determine the amount of churn before relabeling of `target_info` series.

I don't think the changes to owned series recomputation should have any significant perf impact, especially considering this is only called every 15 seconds.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 * [FEATURE] Query-frontend and querier: Add experimental support for performing query planning in query-frontends and distributing portions of the plan to queriers for execution. #12302 #12551 #12665 #12687 #12745 #12757
 * [FEATURE] Alertmanager: add Microsoft Teams V2 as a supported integration. #12680
 * [FEATURE] Distributor: Add experimental flag `-validation.label-value-length-over-limit-strategy` to configure how to handle label values over the length limit. #12627
+* [FEATURE] Ingester: Introduce metric `cortex_ingester_owned_target_info_series` for counting the number of owned `target_info` series. #12681
 * [ENHANCEMENT] Query-frontend: CLI flag `-query-frontend.enabled-promql-experimental-functions` and its associated YAML configuration is now stable. #12368
 * [ENHANCEMENT] Query-scheduler/query-frontend: Add native histogram definitions to `cortex_query_{scheduler|frontend}_queue_duration_seconds`. #12288
 * [ENHANCEMENT] Querier: Add native histogram definition to `cortex_bucket_index_load_duration_seconds`. #12094
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 * [FEATURE] Query-frontend and querier: Add experimental support for performing query planning in query-frontends and distributing portions of the plan to queriers for execution. #12302 #12551 #12665 #12687 #12745 #12757
 * [FEATURE] Alertmanager: add Microsoft Teams V2 as a supported integration. #12680
 * [FEATURE] Distributor: Add experimental flag `-validation.label-value-length-over-limit-strategy` to configure how to handle label values over the length limit. #12627
-* [FEATURE] Ingester: Introduce metric `cortex_ingester_owned_target_info_series` for counting the number of owned `target_info` series. #12681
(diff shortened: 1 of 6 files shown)


pkg/ingester/user_tsdb.go
```diff
@@ -334,6 +336,9 @@ func (u *userTSDB) PostCreation(metric labels.Labels) {
 	// but we will recompute owned series when ingester joins the ring.)
 	u.ownedStateMtx.Lock()
 	u.ownedState.ownedSeriesCount++
+	if metric.Get(model.MetricNameLabel) == "target_info" {
```
juliusmh, line 339: We can merge this with the `metricName` extraction in line 344.

pkg/ingester/user_tsdb.go
```diff
@@ -618,25 +633,34 @@ func (u *userTSDB) updateTokenRanges(newTokenRanges []uint32) bool {
...
 	idx := u.Head().MustIndex()
 	defer idx.Close()
 
+	lb := labels.NewScratchBuilder(128)
 	u.Head().ForEachSecondaryHash(func(refs []chunks.HeadSeriesRef, secondaryHashes []uint32) {
 		for i, sh := range secondaryHashes {
 			if u.ownedTokenRanges.IncludesKey(sh) {
 				count++
+				lb.Reset()
+				if err := idx.Series(storage.SeriesRef(refs[i]), &lb, nil); err != nil {
```
juliusmh, line 654: The `IndexReader` also provides `LabelValueFor`. I couldn't find any other uses of it inside Mimir, but it should be a bit more efficient because we don't alloc and it returns early. E.g.:

```suggestion
idx.LabelValueFor(context.Background(), storage.SeriesRef(refs[i]), "__name__")
```

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 * [FEATURE] Query-frontend and querier: Add experimental support for performing query planning in query-frontends and distributing portions of the plan to queriers for execution. #12302 #12551 #12665 #12687 #12745 #12757
 * [FEATURE] Alertmanager: add Microsoft Teams V2 as a supported integration. #12680
 * [FEATURE] Distributor: Add experimental flag `-validation.label-value-length-over-limit-strategy` to configure how to handle label values over the length limit. #12627
+* [FEATURE] Ingester: Introduce metric `cortex_ingester_owned_target_info_series` for counting the number of owned `target_info` series. #12681
```
juliusmh, line 40: [nit] 

```suggestion
* [FEATURE] Ingester: Introduce metric `cortex_ingester_owned_target_info_series` for counting the number of owned `target_info` series by tenant. #12681
```

pkg/ingester/user_tsdb.go
aknuds1: Interesting, I didn't think of `LabelValueFor`.

aknuds1: Done.

pkg/ingester/user_tsdb.go
aknuds1: Done, thanks.

pkg/ingester/user_tsdb.go
```diff
@@ -330,15 +332,22 @@ func (u *userTSDB) getSeriesCountAndMinLocalLimit() (int, int) {
 func (u *userTSDB) PostCreation(metric labels.Labels) {
 	u.instanceSeriesCount.Inc()
 
+	metricName, err := extract.MetricNameFromLabels(metric)
+	if err != nil {
```
juliusmh, line 336: This is a noop (metricName is already empty). You can instead check `err != nil` in line 350.

juliusmh (approved):
Nice! Just a non-blocking comment.

pkg/ingester/user_tsdb.go
aknuds1: I thought you might comment on that :) Explicitly setting `metricName` to the empty string is to avoid assuming what gets returned when `err != nil`. It's better to be robust, and I can't imagine there's any performance penalty.

PR #12683, opened 2025-09-11
blockbuilder: pre-initialize cortex_blockbuilder_scheduler_job_gap_detected

narqo:
#### What this PR does

Following https://github.com/grafana/mimir/pull/11867

In the mentioned PR I suggested adding the `partition` label to the `cortex_blockbuilder_scheduler_job_gap_detected` metric. It has turned out this creates more problems that benefits. I.e. because the counter is initialized on the first increment, on scheduler's restart, this creates a gap in the series. This isn't a problem typically. But for the kind of counter that increments rarely — like the gap detector — this creates a possibility for an alert to miss the increment (when the counter starts with 1).

This PR updates the metrics

1. remove the partition label from the series
2. pre-initialize the possible "offset_type" series with 0

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/scheduler/metrics.go
+++ b/pkg/blockbuilder/scheduler/metrics.go
@@ -76,9 +76,9 @@ func newSchedulerMetrics(reg prometheus.Registerer) schedulerMetrics {
 		}, []string{"offset_type"}),
 	}
 
-	// Make sure the gap detection counters starts with zero to avoid gaps in their series.
-	scm.jobGapDetected.WithLabelValues(offsetNamePlanned).Add(0)
-	scm.jobGapDetected.WithLabelValues(offsetNameCommitted).Add(0)
+	// Make sure the gap detection counters are pre-initialized. This avoids misleading blanks in the series on restart.
+	scm.jobGapDetected.WithLabelValues(offsetNamePlanned)
+	scm.jobGapDetected.WithLabelValues(offsetNameCommitted)
 
 	return scm
 }


pkg/blockbuilder/scheduler/metrics.go
```diff
@@ -73,6 +73,12 @@ func newSchedulerMetrics(reg prometheus.Registerer) schedulerMetrics {
 		jobGapDetected: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "cortex_blockbuilder_scheduler_job_gap_detected",
 			Help: "The number of times an unexpected gap was detected between jobs.",
-		}, []string{"offset_type", "partition"}),
+		}, []string{"offset_type"}),
 	}
+
+	// Make sure the gap detection counters starts with zero to avoid gaps in their series.
+	scm.jobGapDetected.WithLabelValues(offsetNamePlanned).Add(0)
```
seizethedave: Hmm, just trying to convince myself this would do the thing. My read of [counter.Add(0)](https://github.com/grafana/mimir/blob/1f526adbe16df897f8449205ce3de102fa1359ec/vendor/github.com/prometheus/client_golang/prometheus/counter.go#L132-L135) seems like it would be a no-op. Am I wrong?

seizethedave: Ah, it's `WithLabelValues` that seems to be initializing the metric. As a nit you could drop the Add(0) which I think is just misleading.

narqo: > As a nit you could drop the Add(0) which I think is just misleading.

Good call. I looked at how we pre-initialize metrics in different mimir components, and we do both styles. I've dropped `Add(0)` — we do [similar in store-gateway](https://github.com/grafana/mimir/blob/d5aeaa4b53a2578f7dc44efc3cc9009f5397a9e2/pkg/storegateway/gateway.go#L148-L151)

PR #12692, opened 2025-09-12
Revert "index planning: Disable planning for blocks older than 3 hours"

dimitarvdimitrov:
Reverts grafana/mimir#12608

after https://github.com/grafana/mimir/pull/12607 we have stats for each block, so we can remove this patch


dimitarvdimitrov:
i'll take this in  a separate PR along with using NoopPlanner for smaller blocks

PR #12694, opened 2025-09-12
perf(rw2): use mimirpb.TimeseriesFromPool()

krajorama:
#### What this PR does

Instead of allocating from heap.

Related to #12652

#### Checklist

- [x] Tests updated. With pooling enabled you can get mimirpb.Timeseries from the pool where fields are empty , not nil - randomly.
- N/A Documentation added.
- N/A  `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.

### DIFF
--- a/pkg/distributor/otlpappender/mimir_appender.go
+++ b/pkg/distributor/otlpappender/mimir_appender.go
@@ -142,11 +142,9 @@ func (c *MimirAppender) processLabelsAndMetadata(ls labels.Labels) (hash uint64,
 }
 
 func (c *MimirAppender) createNewSeries(idx *labelsIdx, collisionIdx int, hash uint64, ls labels.Labels, ct int64) {
-	// TODO(krajorama): consider using mimirpb.TimeseriesFromPool
-	ts := &mimirpb.TimeSeries{
-		Labels:           mimirpb.FromLabelsToLabelAdapters(ls),
-		CreatedTimestamp: ct,
-	}
+	ts := mimirpb.TimeseriesFromPool()
+	ts.Labels = mimirpb.FromLabelsToLabelAdapters(ls)
+	ts.CreatedTimestamp = ct
 	c.series = append(c.series, mimirpb.PreallocTimeseries{TimeSeries: ts})
 	idx.idx = len(c.series) - 1
 
--- a/pkg/mimirpb/testutil/timeseries.go
+++ b/pkg/mimirpb/testutil/timeseries.go
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package testutil
+
+import "github.com/grafana/mimir/pkg/mimirpb"
+
+// RemoveEmptyObjectFromSeries is a test utility to replace some empty fields
+// inside the given series with `nil` for easy comparition.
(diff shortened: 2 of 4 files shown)


pkg/mimirpb/testutil/timeseries.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package testutil
+
+import "github.com/grafana/mimir/pkg/mimirpb"
+
+// RemoveEmptyObjectFromSeries is a test utility to replace some empty fields
+// inside the given series with `nil` for easy comparition.
```
tcard:
```suggestion
// inside the given series with `nil` for easy comparison.
```

pkg/distributor/otlpappender/mimir_appender.go
```diff
@@ -142,11 +142,9 @@ func (c *MimirAppender) processLabelsAndMetadata(ls labels.Labels) (hash uint64,
 }
 
 func (c *MimirAppender) createNewSeries(idx *labelsIdx, collisionIdx int, hash uint64, ls labels.Labels, ct int64) {
-	// TODO(krajorama): consider using mimirpb.TimeseriesFromPool
-	ts := &mimirpb.TimeSeries{
-		Labels:           mimirpb.FromLabelsToLabelAdapters(ls),
-		CreatedTimestamp: ct,
-	}
+	ts := mimirpb.TimeseriesFromPool()
```
tcard, line 145: Do we put this back in the pool somewhere?

pkg/distributor/otlpappender/mimir_appender.go
```diff
@@ -142,11 +142,9 @@ func (c *MimirAppender) processLabelsAndMetadata(ls labels.Labels) (hash uint64,
...
+	ts.Labels = mimirpb.FromLabelsToLabelAdapters(ls)
```
tcard, line 146: Note that `mimirpb.FromLabelsToLabelAdapters` clones `ls` into a new allocation, so we're reusing relatively little by calling `mimirpb.TimeseriesFromPool`.

The big advantage of `TimeseriesFromPool` is that it returns a `TimeSeries` with a `Labels` already pointing at a preallocated `Labels` array of some capacity. Ideally we would take advantage of that too, but that probably requires a deeper refactor.

pkg/distributor/otlpappender/mimir_appender.go
krajorama: Yes: https://github.com/grafana/mimir/blob/a5b92a35492dab5d35394458bb4ea7383fd447d0/pkg/distributor/otel.go#L116

(I've put printout in it and ran integration tests and it's being called actually.)

This call has been there already so we've been inflating the pool as far as I can tell.

pkg/distributor/otlpappender/mimir_appender.go
krajorama: ack, that can be next step, thanks!

tcard (approved):

Issue #12604, opened 2025-09-04
index planning: collect per-block statistics


dimitarvdimitrov:
In cases when the head block has stopped receiving series we end up using empty stats to plan lookups for older blocks. One solution to this is to maintain per-block statistics. This should also allow us to have more accurate plans because stats will only include series which are definitely in the block.

This isn't that hard because the block index reader presents the same interface as the mempostings - list of label names, list of label values for each label name, list of series IDs for every label value

**TODO**

- [x] https://github.com/grafana/mimir/pull/12608
- [x] https://github.com/grafana/mimir-prometheus/pull/973
- [x] https://github.com/grafana/mimir/pull/12603
- [x] https://github.com/grafana/mimir/pull/12607
- [x] https://github.com/grafana/mimir-prometheus/pull/981
- [x] https://github.com/grafana/mimir/pull/12752
- [x] https://github.com/grafana/mimir/pull/12696

Issue #12432, opened 2025-08-18
index planning: refine statistics generation after initial phase of testing


dimitarvdimitrov:
> **From very preliminary testing in a dev cluster**

### Reduce the size of sketches/avoid calculating sketches algoether
 
sketches are using about 1.18% of memory - it’s not insignificant, but also not a big deal for now.

We can probably reduce the epsilon we pass to `NewCountMinSketch` for labels with overall low number of values.

chencs:
In this commit (https://github.com/grafana/mimir-prometheus/pull/962/commits/9405bb1f6d3f4805b567c630ccdd9d3d7d53b553), we figured out that setting the count-min sketch to an epsilon of `0.01` is not enough to differentiate high-cardinality values from low-cardinality values for a single label name. This setting is now `0.05`, which means sketches will use more memory. We are going to get this running in a higher-volume environment to get a better idea of how much of a problem this is before pursuing optimizations.

dimitarvdimitrov:
putting this on hold because #13160 #13127. is a good approximation of what we need. we need some experiments and see how it fares in production clusters

PR #12696, opened 2025-09-12
index planning: use planning for all sizeable blocks

dimitarvdimitrov:
We had a patch (#12608) to prevent using stale stats. That wasn't effective and still left the possibility of calculating stats on an empty head, the head receiving series and then using the empty stats on a non-empty head. This results in very inefficient plans. Now that we have #12604, we can remove this logic.

This PR makes the logic for when we use planning a bit more testable with the `lookupplan.PlannerFactory` and removed the changes from #12608

related to #12432 and #12604

### DIFF
--- a/pkg/ingester/lookupplan/planner_factory.go
+++ b/pkg/ingester/lookupplan/planner_factory.go
@@ -9,6 +9,8 @@ import (
 	"github.com/prometheus/prometheus/tsdb/index"
 )
 
+const minSeriesPerBlockForQueryPlanning = 10_000
+
 // PlannerFactory provides index lookup planners for TSDB blocks.
 type PlannerFactory struct {
 	metrics        Metrics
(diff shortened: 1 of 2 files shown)


pkg/ingester/lookupplan/planner_factory.go
```diff
@@ -0,0 +1,48 @@
...
+// For very small blocks (< 10,000 series), it returns NoopPlanner to avoid planning overhead.
+// For larger blocks, it generates statistics and returns a CostBasedPlanner.
+// If statistics generation fails, it falls back to NoopPlanner.
+func (p *PlannerFactory) CreatePlanner(meta tsdb.BlockMeta, reader tsdb.IndexReader) index.LookupPlanner {
+	logger := log.With(p.logger, "block", meta.ULID.String(), "block_series", meta.Stats.NumSeries)
+	const statsThreshold = 10_000
+
+	if meta.Stats.NumSeries < statsThreshold {
+		// For very small blocks, the planning overhead is likely not worth it.
+		// This also prevents problems when we've gathereed stats on an empty head block,
```
chencs:
```suggestion
		// This also prevents problems when we've gathered stats on an empty head block,
```

pkg/ingester/lookupplan/planner_factory.go
```diff
@@ -0,0 +1,48 @@
...
+		// but then the ingester statrs receiving series for that tenant.
```
chencs:
```suggestion
		// but then the ingester starts receiving series for that tenant.
```

pkg/ingester/lookupplan/planner_factory.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+// CreatePlanner returns an appropriate index.LookupPlanner for the given block metadata and reader.
+// For very small blocks (< 10,000 series), it returns NoopPlanner to avoid planning overhead.
+// For larger blocks, it generates statistics and returns a CostBasedPlanner.
+// If statistics generation fails, it falls back to NoopPlanner.
+func (p *PlannerFactory) CreatePlanner(meta tsdb.BlockMeta, reader tsdb.IndexReader) index.LookupPlanner {
+	logger := log.With(p.logger, "block", meta.ULID.String(), "block_series", meta.Stats.NumSeries)
+	const statsThreshold = 10_000
```
chencs: It would be nice (if we need to tune this) to pull this out to the top-level, and then have the boundary tests (e.g., "medium 5000-series block," "boundary case just below threshold (9999 series)", at the boundary, etc.) set those series with arithmetic based on the constant instead of hardcoded.

PR #12700, opened 2025-09-12
continuous-test: Improve adherence to max-records-per-run inside record correctness test

alexweav:
#### What this PR does

Normally, we let the client choose the fetch size, and fetch until we exceed max-records-per-run. That means we follow the limit, plus or minus the size of one fetch. Fetches are bounded in size in terms of bytes.

When there are lots of small records (as there tends to be in big cells), the client might do a single fetch that contains a huge amount of records, so "+/- the size of one fetch" actually can be an enormous amount of variance. In such cells we sometimes see run sizes vastly exceeding the limit:
```
msg="run complete" size=113749
```
(with limit of 80k)

When there's a lot of extra records, even though they're small, each one gets processed to RW2 and back individually, so the memory usage is volatile.

This PR hard-limits the records returned from polling to the limit, so we adhere to the limit precisely instead of by +/- one fetch.

n/a

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/continuoustest/ingest_storage_record.go
+++ b/pkg/continuoustest/ingest_storage_record.go
@@ -245,7 +245,7 @@ func (t *IngestStorageRecordTest) Run(ctx context.Context, now time.Time) error 
 		if ctx.Err() != nil {
 			return err
 		}
-		fetches := t.client.PollFetches(ctx)
+		fetches := t.client.PollRecords(ctx, t.cfg.MaxRecordsPerRun)
 		if errs := fetches.Errors(); len(errs) > 0 {
 			level.Error(t.logger).Log("msg", "fetch errors", "err", fetches.Err())
 			break


pkg/continuoustest/ingest_storage_record.go
```diff
@@ -245,7 +245,7 @@ func (t *IngestStorageRecordTest) Run(ctx context.Context, now time.Time) error
 		if ctx.Err() != nil {
 			return err
 		}
-		fetches := t.client.PollFetches(ctx)
+		fetches := t.client.PollRecords(ctx, t.cfg.MaxRecordsPerRun)
```
alexweav, line 248: Very convenient, kgo has this `PollRecords()` version where you can hard limit the number of records fetched from the internal buffer

alexweav:
> we observed that most of goroutines just seated idle for 300-500ms.

Thank you for the review! Makes total sense. In original testing for this there were 10 workers, and that's been reduced to 4 already and showed some decent improvement.

Previously this ran into issues where the processing would rarely overtake the 5-minute run interval, hence the reason for the parallelism to counter it. However, there were also other fixes that improved this problem, like stripping out reflection when doing comparisons, so it may not be relevant anymore. I'll look into it if we continue to see memory spiking too high.

> Consider checking the [storage/ingest.PusherConsumer](https://github.com/grafana/mimir/blob/bf16abb72a2995aedcb0cf6719dd397e4379083d/pkg/storage/ingest/pusher.go#L65C9-L65C23)

I had hoped to not need the complexity - the goal here was for this tool to be "temporary-ish" and thus have the simplest possible kgo consumer. But, writes are a very high workload indeed and I don't want to reduce the record counts to the point where we don't have confidence in the test anymore. Let's see how far we get with the above two, to figure out if we need to take on the complexity

alexweav:
Backporting this fix to r360 so that we can get back to testing faster. It's tiny/trivial.

PR #12701, opened 2025-09-13
Block-builder-scheduler: Clean up initial probing and warn when we don't probe far enough.

seizethedave:
#### What this PR does

If the scheduler is started with a `max-scan-age` that is smaller than the actual age of the commit record, we will skip data. This PR adds a warning log when this happens to make this situation more apparent.

And it polishes up the commentary around startup probing a little more.

#### Checklist

- (n/a) Tests updated.
- (n/a) Documentation added.
- (n/a) `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.


narqo (approved):
Nice 🔥

PR #12693, opened 2025-09-12
storegateway: BucketChunkReader to validate chunk data length

narqo:
#### What this PR does

The PR adds a validation to the `BucketChunkReader` to make sure it doesn't panic, if the parsed chunk length exceed the allowed buffer length. See #12691 for the details.

#### Which issue(s) this PR fixes or relates to

Fixes #12691

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12729, opened 2025-09-16
storegateway: safeguard chunk parser

narqo:
#### What this PR does

As @dimitarvdimitrov brought in https://github.com/grafana/mimir/pull/12693#discussion_r2352625595 the `tsdb.EstimatedMaxChunkSize` is only an estimation, and we shouldn't use it as the parser's cap. The PR updates the code to cap it with a more relaxed value.

Follows https://github.com/grafana/mimir/pull/12693

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket_chunk_reader.go
+++ b/pkg/storegateway/bucket_chunk_reader.go
@@ -31,7 +31,7 @@ import (
 
 // maxChunkDataLen is a safeguard to protect the chunk data parsing from over allocating in case of a broken chunk.
 // Refer to https://github.com/grafana/mimir/issues/12691 for the background.
-// 512MB would a ridiculously large chunk. But to keep things relaxed, we need a much larger value than tsdb.EstimatedMaxChunkSize.
+// 512MB would be a ridiculously large chunk. But, to keep things relaxed, we use a much larger value than tsdb.EstimatedMaxChunkSize.
 const maxChunkDataLen = 512 * 1024 * 1024
 
 type bucketChunkReader struct {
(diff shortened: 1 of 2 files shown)


pkg/storegateway/bucket_chunk_reader.go
```diff
@@ -148,9 +153,9 @@ func (r *bucketChunkReader) loadChunks(ctx context.Context, res []seriesChunks,
 
 		// Validate chunk data length to prevent unreasonably large memory allocations or panics from a corrupted data.
 		// Note, that on a 64-bit system Go should allow to allocate up to 1<<47 bytes; a larger len causes "makeslice: len out of range".
```
dimitarvdimitrov:
```suggestion
		// Note, that on a 64-bit system Go should allow to allocate up to ~140TB; a larger len causes "makeslice: len out of range".
```

dimitarvdimitrov (approved):

pkg/storegateway/bucket_chunk_reader.go
```diff
@@ -29,6 +29,11 @@ import (
 	"github.com/grafana/mimir/pkg/util/pool"
 )
 
+// maxChunkDataLen is a safeguard to protect the chunk data parsing from over allocating in case of a broken chunk.
+// Refer to https://github.com/grafana/mimir/issues/12691 for the background.
+// 512MB would a ridiculously large chunk. But to keep things relaxed, we need a much larger value than tsdb.EstimatedMaxChunkSize.
```
narqo:
```suggestion
// 512MB would be a ridiculously large chunk. But, to keep things relaxed, we use a much larger value than tsdb.EstimatedMaxChunkSize.
```

PR #12734, opened 2025-09-17
Include more error information in query traces

charleskorn:
#### What this PR does

This PR improves the information logged in query traces when an error occurs:
* MQE will now log whether the query succeeded or failed, and if it failed, what the error was.
* Query-frontends will log the error returned by queriers, if any.

I've opted not to add a changelog entry given this only affects tracing.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/limits.go
+++ b/pkg/frontend/querymiddleware/limits.go
@@ -288,9 +288,14 @@ type httpQueryRequestRoundTripperHandler struct {
 	codec  Codec
 }
 
-func (rth httpQueryRequestRoundTripperHandler) Do(ctx context.Context, r MetricsQueryRequest) (Response, error) {
+func (rth httpQueryRequestRoundTripperHandler) Do(ctx context.Context, r MetricsQueryRequest) (resp Response, err error) {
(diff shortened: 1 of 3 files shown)


pkg/frontend/querymiddleware/limits.go
```diff
@@ -288,9 +288,14 @@ type httpQueryRequestRoundTripperHandler struct {
 	codec  Codec
 }
 
-func (rth httpQueryRequestRoundTripperHandler) Do(ctx context.Context, r MetricsQueryRequest) (Response, error) {
+func (rth httpQueryRequestRoundTripperHandler) Do(ctx context.Context, r MetricsQueryRequest) (resp Response, err error) {
```
monoxane, line 291: Are we using named returns to ensure any return is implicitly instantiated as empty here or is it just for the convenience of having names?

IIRC named return values always result in an additional memory allocation and value copy instead of just passing the object directly which could have performance implications, particularly with high RPS handlers.

charleskorn: This is so we can examine the returned error in the deferred function below.

PR #12737, opened 2025-09-17
kafkatool: print write requests as JSON

juliusmh:
#### What this PR does

Adds option to print WriteRequests as JSON when debugging Kafka dumps.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/tools/kafkatool/dump.go
+++ b/tools/kafkatool/dump.go
@@ -28,6 +28,7 @@ type DumpCommand struct {
 	inOutFile         *os.File
 	getKafkaClient    func() *kgo.Client
 	printer           Printer
+	printFormat       string
 }
 
 // Register is used to register the command to a parent command.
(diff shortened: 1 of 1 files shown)


tools/kafkatool/dump.go
```diff
@@ -36,16 +37,21 @@ func (c *DumpCommand) Register(app *kingpin.Application, getKafkaClient func() *
 	c.printer = printer
 
 	cmd := app.Command("dump", "Dump Kafka topic contents")
-	cmd.Flag("topic", "Kafka topic to dump").Required().StringVar(&c.topic)
```
tcard, line 39: So were those just in the wrong command?

tools/kafkatool/dump.go
```diff
@@ -36,16 +37,21 @@ func (c *DumpCommand) Register(app *kingpin.Application, getKafkaClient func() *
...
 	cmd.Flag("file", "File to read from or write to.").Required().OpenFileVar(&c.inOutFile, os.O_RDWR|os.O_CREATE, 0600)
 
-	cmd.Command("import", "Import records from a file into a Kafka topic").Action(c.doImport)
-	cmd.Command("export", "Export records from a Kafka topic into a file").Action(c.doExport)
-	cmd.Command("print", "Print the write requests inside records dumped using this tool").Action(c.doPrint)
+	importCmd := cmd.Command("import", "Import records from a file into a Kafka topic").Action(c.doImport)
+	importCmd.Flag("topic", "Kafka topic to import into.").Required().StringVar(&c.topic)
+	importCmd.Flag("partition", "Kafka partition to import into").Required().IntVar(&c.partition)
+
+	exportCmd := cmd.Command("export", "Export records from a Kafka topic into a file").Action(c.doExport)
```
tcard, line 47: I understand this wasn't introduced by your PR, but it's unclear to have both `dump` and `export`...

tcard (approved):

tools/kafkatool/dump.go
juliusmh: So the final CLI experience becomes:

```shell
./kafkatool dump export <args>
./kafkatool dump print <args>
```

tools/kafkatool/dump.go
juliusmh: They were all in all subcommands. But not all of them need all the flags. E.g. just inspecting/printing a dump doesn't need partition and topic.

tools/kafkatool/dump.go
tcard: Ah! I totally missed `dump` was a parent flag.

Issue #12736, opened 2025-09-17
Bug: RW2 receivers must return 415 upon bad encoding or bad content type


roidelapluie:
### What is the bug?

The [Remote Write 2 specification](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/) mandates that RW2 receivers returns 415 upon bad encoding or bad content type:

> Unsupported Request Content
> Receivers MUST return [415 HTTP Unsupported Media Type](https://www.rfc-editor.org/rfc/rfc9110.html#name-415-unsupported-media-type) status code if they don't support a given content type or encoding provided by Senders.

### How to reproduce it?

1. in mimir/development/mimir-monolithic-mode: `./compose-up.sh --profile prometheusRW2`

2.

```
curl "http://127.0.0.1:8101/api/v1/push"   -H "Content-Type: application/json"   -H "Content-Encoding: snappy"   --data-binary '{}' -v
*   Trying 127.0.0.1:8101...
* Connected to 127.0.0.1 (127.0.0.1) port 8101
* using HTTP/1.x
> POST /api/v1/push HTTP/1.1
> Host: 127.0.0.1:8101
> User-Agent: curl/8.14.1
> Accept: */*
> Content-Type: application/json
> Content-Encoding: snappy
> Content-Length: 2
>
* upload completely sent off: 2 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Wed, 17 Sep 2025 09:27:10 GMT
< Content-Length: 41
<
decompress snappy: snappy: corrupt input
* Connection #0 to host 127.0.0.1 left intact
```

3. 

```
curl "http://127.0.0.1:8101/api/v1/push" -H "Content-Type: application/x-protobuf;proto=io.prometheus.write.v2.Request"   -H "Content-Encoding: superzip"   --data-binary '{}' -v
*   Trying 127.0.0.1:8101...
* Connected to 127.0.0.1 (127.0.0.1) port 8101
* using HTTP/1.x
> POST /api/v1/push HTTP/1.1
> Host: 127.0.0.1:8101
> User-Agent: curl/8.14.1
> Accept: */*
> Content-Type: application/x-protobuf;proto=io.prometheus.write.v2.Request
> Content-Encoding: superzip
> Content-Length: 2
>
* upload completely sent off: 2 bytes
< HTTP/1.1 400 Bad Request
[...]

PR #12739, opened 2025-09-17
distributor: validate content-encoding header

juliusmh:
#### What this PR does

This PR adds validation for the `Content-Encoding` header.
If receiving a value other than `""` or `"snappy"` we return 415.
The `content-type` header validation was also adjusted to return 415.

According to the spec, RW2 receivers should also return 415 for invalid content types. Since Mimir's
`handler` implements both versions (and switches based on `Content-Type` header) it must return 415 for both versions.

Relates to https://github.com/grafana/mimir/issues/12736

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/push.go
+++ b/pkg/distributor/push.go
@@ -166,20 +166,14 @@ func handler(
 
 		isRW2, err := isRemoteWrite2(r)
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			http.Error(w, err.Error(), http.StatusUnsupportedMediaType)
 		}
 
 		// https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/#unsupported-request-content:
 		// Receivers MUST return 415 HTTP Unsupported Media Type status code if they
-		// don't support a given content type or encoding provided by Senders. Since this handler implements
-		// RW1 and RW2 (based on content-type header) we can only handle the case of invalid encoding when we are
-		// acting as a remote-write 2.0 receiver.
+		// don't support a given content type or encoding provided by Senders.
 		if !isValidContentEncoding(r) {
-			code := http.StatusBadRequest
-			if isRW2 {
-				code = http.StatusUnsupportedMediaType
-			}
-			http.Error(w, "invalid content encoding", code)
+			http.Error(w, "invalid content encoding", http.StatusUnsupportedMediaType)
 		}
 
 		supplier := func() (*mimirpb.WriteRequest, func(), error) {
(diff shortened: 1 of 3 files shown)


pkg/distributor/push.go
```diff
@@ -164,6 +164,20 @@ func handler(
...
+		// Receivers MUST return 415 HTTP Unsupported Media Type status code if they
+		// don't support a given content type or encoding provided by Senders. Since this handler implements
+		// RW1 and RW2 (based on content-type header) we can only handle the case of invalid encoding when we are
+		// acting as a remote-write 2.0 receiver.
+		if !isValidContentEncoding(r) {
+			code := http.StatusBadRequest
+			if isRW2 {
+				code = http.StatusUnsupportedMediaType
+			}
+			http.Error(w, "invalid content encoding", code)
```
juliusmh: Should we return early here? We don't in line 165, but I'm wondering if there is a specific reason?

pkg/distributor/push.go
```diff
@@ -297,6 +305,18 @@ func isRemoteWrite2(r *http.Request) (bool, error) {
 	return false, nil
 }
 
+// isValidContentEncoding validates the content-encoding header is either empty
+// (in which case Mimir assumes the "snappy") or set to a supported value.
```
bboreham: Nit: don’t need “the”

pkg/distributor/push.go
```diff
@@ -162,8 +162,16 @@ func handler(
 
 		isRW2, err := isRemoteWrite2(r)
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			http.Error(w, err.Error(), http.StatusUnsupportedMediaType)
```
bboreham, line 169: Here you seem to be assuming that all errors from `isRemoteWrite2` are content-type errors. 
Suggest this should be written into the docstring, or have the function return the error type.

bboreham (commented):
Broadly ok, thanks. 
I will note there is a long history of being surprised at what fails after changing an error code. Though I can’t predict anything specific in this case.

pkg/distributor/push.go
juliusmh: The `isRemoteWrite2` function is permissive. This PR doesn't change that, it just returns 415 for the cases where we would already return 400.

| Content-Type Header | RW Version | Err |
|---------------------|------------|-----|
| ""                  | 1          | nil |
| unknown                  | 1          | nil |
| application/x-protobuf                  | 1          | nil |
| application/x-protobuf==                  |           | Err invalid parts format |
| application/x-protobuf;proto=X                  |           | Err invalid proto message || application/x-protobuf;proto=prometheus.WriteRequest                  |    1       | nil |
| application/x-protobuf;proto=io.prometheus.write.v2.Request                 |    2       | nil |

tacole02 (approved):
Changelog looks good.

dimitarvdimitrov:

PR #12741, opened 2025-09-17
fix(distributor): NHCB bucket count validation wrong

krajorama:
#### What this PR does

If the bucket limit is set for native histograms then no NHCB passes validation due to a coding error.

Enabled bucket limit in test config to reproduce and fixed bug.

Related to RW2 compliance work. https://github.com/prometheus/prometheus/issues/16944

#### Checklist

- [x] Tests updated.
- N/A Documentation added.


krajorama:
> Good catch.

credit goes to @roidelapluie

PR #12745, opened 2025-09-18
Remote execution: add support for strong consistency, chunk info logging and other features that rely on HTTP headers

charleskorn:
#### What this PR does

This PR adds support for strong consistency to the remote execution path introduced in https://github.com/grafana/mimir/pull/12551 and https://github.com/grafana/mimir/pull/12302.

I've introduced the concept of `Injector`s, which are responsible for adding headers / metadata to a request from query-frontends to queriers, and the concept of `Propagator`s, which are responsible for extracting information from headers / metadata and adding them to the `context.Context`.

This model also allows us to support other features in GEM that relied on propagating information in HTTP headers, so I've reverted #12699.

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/scheduler/schedulerpb/scheduler.proto
```diff
@@ -61,7 +61,16 @@ message SchedulerToQuerier {
 
 message ProtobufRequest {
   google.protobuf.Any payload = 1;
-  map<string, string> traceHeaders = 2;
+
+  // Metadata is metadata orthogonal to the request itself, analogous to HTTP headers or gRPC metadata.
+  // We can't use gRPC metadata in this context because this message is sent over long-lived streams, rather
```
56quarters, line 66: This is the part I was missing, I assumed that gRPC must have _something_ like this. Can you include something like this in `util/propagation` to explain why the package needs to exist?

pkg/scheduler/schedulerpb/custom.go
```diff
@@ -2,8 +2,55 @@
...
+	metadataMap := make(map[string][]string, len(m))
+
+	for _, item := range m {
+		metadataMap[item.Key] = item.Value
+	}
+
+	return metadataMap
+}
+
+func MetadataMapToSlice(m map[string][]string) []MetadataItem {
```
56quarters: personal preference but I'd expect this to be named like this based on what it does:
```suggestion
func MapToMetadataSlice(m map[string][]string) []MetadataItem {
```

pkg/scheduler/schedulerpb/custom.go
```diff
@@ -2,8 +2,55 @@
...
+		metadataSlice = append(metadataSlice, MetadataItem{
+			Key:   k,
+			Value: v,
+		})
+	}
+
+	return metadataSlice
+}
+
+type MetadataMapCarrier map[string][]string
```
56quarters: So there's a `Carrier` interface in `pkg/propagation` but this type isn't an implementation of it, it's for an OTEL `Carrier` interface. Can you name this something different so it's a bit less confusing?

pkg/querier/worker/worker.go
```diff
@@ -81,7 +81,7 @@ type RequestHandler interface {
 }
 
 type ProtobufRequestHandler interface {
-	HandleProtobuf(context.Context, *types.Any, frontendv2pb.QueryResultStream)
+	HandleProtobuf(context.Context, *types.Any, map[string][]string, frontendv2pb.QueryResultStream)
```
56quarters: Would it make more sense to define this as accepting a `propagation.Carrier`?

pkg/frontend/querymiddleware/limits.go
```diff
@@ -324,6 +327,12 @@ func (rth *engineQueryRequestRoundTripperHandler) Do(ctx context.Context, r Metr
 	spanLogger, ctx := spanlogger.New(ctx, rth.logger, tracer, "engineQueryRequestRoundTripperHandler.Do")
 	defer spanLogger.Finish()
 
+	headers := map[string][]string{}
+	if err := rth.codec.AddHeadersForMetricQueryRequest(ctx, r, propagation.MapCarrier(headers)); err != nil {
```
56quarters, line 341: Feels a little weird to include the entire `Codec` to propagate some headers but I guess this is temporary until remote execution is used for all queries.

pkg/util/propagation/propagation.go
```diff
@@ -0,0 +1,123 @@
...
+
+type HttpHeaderCarrier http.Header
+
+func (h HttpHeaderCarrier) Get(name string) string {
+	// We don't need to canonicalize name here, because Get does that for us.
+	return http.Header(h).Get(name)
+}
+
+func (h HttpHeaderCarrier) GetAll(name string) []string {
+	return h[textproto.CanonicalMIMEHeaderKey(name)]
```
56quarters: I think you can also use `http.Header(h).Values(name)` here

pkg/scheduler/schedulerpb/scheduler.proto
charleskorn: Yep, added in d5e81ae

pkg/scheduler/schedulerpb/custom.go
charleskorn: Good idea, changed in 9d0cc53

pkg/scheduler/schedulerpb/custom.go
charleskorn: Good idea, changed in d218ae8

pkg/querier/worker/worker.go
charleskorn: Yes it would, changed in 25dd5af

pkg/util/propagation/propagation.go
charleskorn: Thank you - changed in 9552fe1

PR #12750, opened 2025-09-18
Do not wrap vector-scalar comparison in DeduplicateAndMerge

Konstantinov-Innokentii:
#### What this PR does

Vector-scalar comparison without bool modifier doesn't drop names, so it will not produce duplicates. Therefore, DeduplicateAndMerge is not needed.

#### Checklist

- [x] Tests updated.
- ~[ ] Documentation added.~
- ~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.~

### DIFF
--- a/pkg/streamingpromql/planning.go
+++ b/pkg/streamingpromql/planning.go
@@ -409,6 +409,15 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
 			(lhsType == parser.ValueTypeScalar && rhsType == parser.ValueTypeVector)
 
 		if isVectorScalar {
+			// Comparison vector-scalar operations without bool modifier don't drop the __name__ label.
+			// So don't need to wrap in DeduplicateAndMerge.
+			isComparison := expr.Op == parser.EQLC || expr.Op == parser.NEQ ||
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/planning.go
```diff
@@ -409,6 +409,15 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
 			(lhsType == parser.ValueTypeScalar && rhsType == parser.ValueTypeVector)
 
 		if isVectorScalar {
+			// Comparison vector-scalar operations without bool modifier don't drop the __name__ label.
+			// So don't need to wrap in DeduplicateAndMerge.
+			isComparison := expr.Op == parser.EQLC || expr.Op == parser.NEQ ||
+				expr.Op == parser.LTE || expr.Op == parser.LSS ||
+				expr.Op == parser.GTE || expr.Op == parser.GTR
+			if isComparison && !expr.ReturnBool {
```
charleskorn: Prometheus already has a method that does this for us:

```suggestion
			if expr.Op.IsComparisonOperator() && !expr.ReturnBool {
```

PR #12751, opened 2025-09-18
Add support to configure max eventual consistency delay on a per-query basis

pracucci:
#### What this PR does

In this PR I'm adding the experimental support to configure the max eventual consistency delay on a per-query basis. The max delay is currently propagated via a new internal HTTP header named `X-Read-Consistency-Max-Delay`, but in a follow up PR I will try to reuse the existing `X-Read-Consistency` header to both specify the level (as of today) and the max delay (new).

The max delay is intended to be used internally, and in particular by ruler. When a rule is evaluated with eventual consistency, we may want to specify a max delay to ensure it doesn't get executed on very stale data. I initially though to set the delay equal to the configured `ruler_evaluation_delay_duration`, but in this PR I'm introducing a new experimental configuration option called `ruler_evaluation_consistency_max_delay`. The new config option give us the ability to experiment with the delay, before deciding whether just setting the delay to `ruler_evaluation_delay_duration` is a good idea or not.

When the ingester receives a request with `eventual` consistency level and max delay configured, the ingester checks if the max delay is honored. If it's not honored, the ingester immediately fail the query. I opted for this solution, rather than having the ingester waiting until max delay is honored, because the idea is to set max delay to a fairly high value (e.g. 1 minute) and so waiting extra time may not give much benefit, but would slow down other enqueued queries not having max delay enforcement (the possible slowdown is due to  saturating querier workers, that would get occupied just by waiting for max delay to be enforced).

The max delay is not enforced when the consistency level of a query is `strong`, because strong consistency already implies an up to date state.

I've also manually tested this PR in a local dev env, injecting a fake high latency when consuming from Kafka and looks working as expected.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/codec.go
+++ b/pkg/frontend/querymiddleware/codec.go
@@ -54,9 +54,9 @@ var (
 	allFormats        = []string{formatJSON, formatProtobuf}
 
 	// List of HTTP headers to propagate when a Prometheus request is encoded into a HTTP request.
-	// api.ReadConsistencyHeader is propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
+	// Read consistency headers are propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
 	codecPropagateHeadersMetrics = []string{compat.ForceFallbackHeaderName, chunkinfologger.ChunkInfoLoggingHeader, api.ReadConsistencyOffsetsHeader, querier.FilterQueryablesHeader}
-	// api.ReadConsistencyHeader is propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
+	// Read consistency headers are propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
 	codecPropagateHeadersLabels = []string{api.ReadConsistencyOffsetsHeader, querier.FilterQueryablesHeader}
 )
 
--- a/pkg/frontend/querymiddleware/codec.go
+++ b/pkg/frontend/querymiddleware/codec.go
@@ -54,9 +54,9 @@ var (
 	allFormats        = []string{formatJSON, formatProtobuf}
 
 	// List of HTTP headers to propagate when a Prometheus request is encoded into a HTTP request.
-	// Read consistency headers are propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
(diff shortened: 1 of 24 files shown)


pkg/frontend/querymiddleware/codec.go
```diff
@@ -54,9 +54,9 @@ var (
 	allFormats        = []string{formatJSON, formatProtobuf}
 
 	// List of HTTP headers to propagate when a Prometheus request is encoded into a HTTP request.
-	// api.ReadConsistencyHeader is propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
+	// Read consistency headers are propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
 	codecPropagateHeadersMetrics = []string{compat.ForceFallbackHeaderName, chunkinfologger.ChunkInfoLoggingHeader, api.ReadConsistencyOffsetsHeader, querier.FilterQueryablesHeader}
-	// api.ReadConsistencyHeader is propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
+	// Read consistency headers are propagated as HTTP header -> Request.Context -> Request.Header, so there's no need to explicitly propagate it here.
```
narqo: (nit) I find the updated phrasing slightly confusing because both lists include `api.ReadConsistencyOffsetsHeader`. From the code reader's perspective, it's not clear if this header isn't in the "Read consistency headers" group, or the actual values in the lists have diverged from the comment.

pkg/ruler/compat.go
```diff
@@ -392,7 +393,7 @@ func DefaultTenantManagerFactory(
 		}
 
 		// Wrap the query function with our custom logic.
-		wrappedQueryFunc := WrapQueryFuncWithReadConsistency(queryFunc, logger)
+		wrappedQueryFunc := WrapQueryFuncWithReadConsistency(queryFunc, overrides, userID, logger)
```
narqo, line 396: Don't forget that GEM has its own TenanManagerFactory, thus merging the change in `WrapQueryFuncWithReadConsistency` will break it, I guess 🙈

pkg/frontend/querymiddleware/codec.go
pracucci: You're right. Fixed in https://github.com/grafana/mimir/pull/12751/commits/d8197a1edb4ad9dfb3d521f883848ae636252642

pkg/ruler/compat.go
pracucci: You're absolutely right. I will take care of GEM as soon as this PR is merged.

PR #12752, opened 2025-09-18
inex planning: Add statistics generation for head block

### LINKED ISSUE #12604: index planning: collect per-block statistics

dimitarvdimitrov:
## Summary

This PR implements a background statistics service that pre-computes query planning statistics for head blocks. The idea is to move all statistics collection and creation out of prometheus and  inside mimir. This is the last part of https://github.com/grafana/mimir/issues/12604.

This includes the changes from https://github.com/grafana/mimir-prometheus/pull/981

## Necessary changes

* the metrics we register are no longer in the prometheus TSDB; instead we instantiate a per-tenant `lookupplan.Metrics` instance for every tenant

final piece of #12604, so this closes #12604

### DIFF
--- a/pkg/ingester/lookupplan/planner_test.go
+++ b/pkg/ingester/lookupplan/planner_test.go
@@ -49,7 +49,7 @@ func TestCostBasedPlannerPlanIndexLookup(t *testing.T) {
 	testCases := data.ParseTestCases(t)
 
 	stats := newHighCardinalityMockStatistics()
-	metrics := NewMetrics(nil)
+	metrics := NewMetrics(nil).ForUser("test-user")
 	planner := NewCostBasedPlanner(metrics, stats)
 
 	const writeOutNewResults = false
@@ -110,7 +110,7 @@ func BenchmarkCostBasedPlannerPlanIndexLookup(b *testing.B) {
 	testCases := data.ParseTestCases(b)
 
 	stats := newHighCardinalityMockStatistics()
-	metrics := NewMetrics(nil)
+	metrics := NewMetrics(nil).ForUser("test-user")
 	planner := NewCostBasedPlanner(metrics, stats)
 
 	for _, tc := range testCases {
@@ -143,7 +143,7 @@ func matchersStrings(ms []*labels.Matcher) []string {
 func TestCostBasedPlannerTooManyMatchers(t *testing.T) {
 	ctx := context.Background()
 	stats := newMockStatistics()
-	metrics := NewMetrics(nil)
+	metrics := NewMetrics(nil).ForUser("test-user")
 	planner := NewCostBasedPlanner(metrics, stats)
 
 	// Create more than 10 matchers to trigger the limit
@@ -183,7 +183,7 @@ func (p *basicLookupPlan) ScanMatchers() []*labels.Matcher {
 func TestCostBasedPlannerPreservesAllMatchers(t *testing.T) {
 	ctx := context.Background()
 	stats := newHighCardinalityMockStatistics()
-	metrics := NewMetrics(nil)
+	metrics := NewMetrics(nil).ForUser("test-user")
 	planner := NewCostBasedPlanner(metrics, stats)
 
 	t.Run("mixed_index_and_scan_matchers", func(t *testing.T) {
@@ -252,7 +252,7 @@ func TestCostBasedPlannerPreservesAllMatchers(t *testing.T) {
 func TestCostBasedPlannerPrefersIndexMatchersOverCheapestPlan(t *testing.T) {
 	ctx := context.Background()
 	stats := newSingleValueStatistics()
-	metrics := NewMetrics(nil)
+	metrics := NewMetrics(nil).ForUser("test-user")
 	planner := NewCostBasedPlanner(metrics, stats)
 
 	matchers := []*labels.Matcher{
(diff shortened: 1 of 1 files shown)


pkg/ingester/planner_provider.go
```diff
@@ -0,0 +1,62 @@
...
+// storePlanner stores a planner for the given block ULID.
+func (s *plannerProvider) storePlanner(blockULID ulid.ULID, planner index.LookupPlanner) {
+	s.plannersMtx.Lock()
+	defer s.plannersMtx.Unlock()
+
+	s.planners[blockULID] = planner
+}
+
+// generateAndStorePlanner generates a planner for the given block metadata and index reader, and stores it in the cache.
+func (s *plannerProvider) generateAndStorePlanner(blockMeta tsdb.BlockMeta, indexReader tsdb.IndexReader) {
```
chencs, line 60: Where are we caching planners for non-head blocks? I only see this used for head stats generation.

pkg/ingester/planner_provider_test.go
```diff
@@ -0,0 +1,72 @@
...
+type MockPlannerFactory struct {
+	mock.Mock
+}
+
+func (m *MockPlannerFactory) CreatePlanner(meta tsdb.BlockMeta, reader tsdb.IndexReader) index.LookupPlanner {
+	args := m.Called(meta, reader)
+	return args.Get(0).(index.LookupPlanner)
+}
+
+func TestPlannerProvider_getPlanner_DoesNotCachePlanners(t *testing.T) {
```
chencs, line 26: Do you strongly prefer mocking here? I think these tests could probably be simpler to read if we just check the planner map for a cached planner.

pkg/ingester/planner_provider.go
dimitarvdimitrov: yeaahhhh you're right. we're not caching them. the reason for that is because if we cache them we'd also need to remove them from the cache eventually because blocks get old out of retention.

at the same time i decided to go with a map for the blocks because i couldn't find an easy way to compare the ulid of the head block with the ulid we get when generating statistics - the ulids are private and are just implementation details.

pkg/ingester/planner_provider_test.go
dimitarvdimitrov: i'd rather not rely on internal implementation details. it makes the tests brittle. i can do something like a simpler mock without using `mock.Mock` if that'd be simpler?

chencs: Maybe. Mainly, it was just quite easy to miss the `Once()` and `Twice()`. Reading more closely helped, maybe just moving mock return stuff closer to the `getPlanner` calls is enough.

Also, is there a reason to mix `require` and `assert` here?

dimitarvdimitrov: > Maybe. Mainly, it was just quite easy to miss the Once() and Twice(). Reading more closely helped, maybe just moving mock return stuff closer to the getPlanner calls is enough.


ok i'll see how can i can make this simpler. claude went with the mock even though we're not that heavy users ourselves. i'll see what i can do

dimitarvdimitrov: > Also, is there a reason to mix require and assert here?


they do different things. `require` aborts the test if the assertion fails. `assert` continues with the rest of the test. `assert` is helpful because you see all assertions that failed, not just the first one - it helps debugging when you know if everything is broken or just one assertion.

pkg/ingester/planner_provider.go
chencs: I'd update this comment to note that the generated planner isn't cached: 

https://github.com/grafana/mimir/blob/cd19ed716d68a36d76d902479ebec284b744769c/pkg/ingester/user_tsdb.go#L172

And maybe add a bit to the `plannerProvider` description to mention that we don't cache block stats (I know it already mentions that it's just for the head blocks, but I forgot while jumping around, so I fully believe someone else will later)

dimitarvdimitrov: done in https://github.com/grafana/mimir/pull/12752/commits/36a0b86cad92a54c9fb24a80ca78f8510e2f9116

pkg/ingester/planner_provider_test.go
dimitarvdimitrov: simplified the mocks in https://github.com/grafana/mimir/pull/12752/commits/c5e0555213c8061214e98b9ff87d39cf6c86dca7

dimitarvdimitrov:
i think i addressed your comments, PTAL @chencs

PR #12761, opened 2025-09-19
Parameterised Microservice Dev Config

monoxane:
#### What this PR does
This PR reworks the local development docker-compose files to allow arbitrarily scaling or disabling all of the Mimir, something I needed previously to validate how some changes affected metrics and the dashboards.

Unfortunately it is practically impossible to generate Nginx configuration files with Jsonnet, so it's been replaced with Envoy, this does have the benefit of letting us gather metrics and traces end to end.

This is the first time I've really gone deep with Jsonnet so please be nit-picky on what can be done better/cleaner/faster. I've noticed that theres usually several ways to do anything so I've mostly just stuck to how things were in the repo originally, but that may not extend very well to multi-file outputs with imported libs.

#### Checklist

- [n/a] Tests updated.
- [ ] Documentation will be added.


monoxane:
As a new starter I was not aware of the processes and availability of the internal dev clusters, and until now was under the impression that the docker-compose was used for basically everything before release. Evidently this isn't the case so we'll just call this a bit of sunk cost and close it out. Nonetheless this was still a good experience for me to learn jsonnet, so it's not all wasted effort...

I'll get more info regarding the access and processes for dev cluster use and write some internal docs for other new people to follow in the future.

PR #12668, opened 2025-09-09
CI: Notify MQE maintainers when pkg/streamingpromql tests fail on vendor PRs

dimitarvdimitrov:
## Summary

Automatically notify MQE maintainers when `pkg/streamingpromql` tests fail on mimir-prometheus vendoring PRs by posting a GitHub comment with @mentions.

These tests usually fail because there are tests in `pkg/streamingpromql/testdata/upstream/*.test` which need updating. Usually there's also the need to update some other unit tests or fix functionality. This is hard to automate and needs human judgement.

This PR tags @charleskorn, @56quarters, @lamida, @zenador and @Konstantinov-Innokentii  when `pkg/stremingpromql` fails unit tests during vendoring PRs.

## Changes

- **Modified `.github/workflows/scripts/run-unit-tests-group.sh`**:
  - Capture test output using `tee` while preserving real-time display
  - Extract all failed package names and store in `FAILED_PACKAGES` environment variable
  - Maintain original exit code behavior

- **Updated `.github/workflows/test-build-deploy.yml`**:
  - Added notification step that runs on test failure for PRs with `vendored-mimir-prometheus-update` label
  - Checks if `pkg/streamingpromql` is in the list of failed packages
  - Posts PR comment with context about test group, failed packages, and links to logs

PR #12763, opened 2025-09-19
test-build-deploy: Install GitHub CLI in test workflow step

dimitarvdimitrov:
The test workflow has a step that uses 'gh pr comment' but the GitHub CLI
is not available in the build container. This adds an install step to
install gh CLI before it's used in the streamingpromql failure notification.
Screenshot 2025-09-19 at 15 14 26

follow-up of #12668


armandgrillet:
I'm surprised by this PR, `ubuntu-latest` is used for the step and contains the GitHub CLI according to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli.

PR #12766, opened 2025-09-19
jsonnet: Add timeout validation for querier and query-frontend

dimitarvdimitrov:
## Summary

Adds comprehensive timeout validation for querier and query-frontend components that accounts for both explicitly configured values and default flag values, preventing misconfigurations that could cause request failures.

### Enhanced parseDuration Function
- Added support for milliseconds (`500ms`)
- Added support for combined formats (`4m30s`, `1h30m`)
- Fixed edge case where `500ms` was incorrectly parsed as combined format

### Default Values Integration
- Added `tools/extract-flag-defaults.jq` to extract flag defaults from config descriptor
- Extended `reference-help` Makefile target to generate `operations/mimir/mimir-flags-defaults.json`
- Added `getFlagDefault()` and `getFlagDefaultSeconds()` utility functions in common.libsonnet
- Automatic nanosecond-to-seconds conversion for duration defaults

### Validation Added
- **querier**: `querier.timeout` must be ≤ `server.http-write-timeout` (uses defaults when flags not set)
- **query-frontend**: `querier.timeout` must be ≤ `server.http-write-timeout` (uses defaults when flags not set)
- Clear error messages showing whether using explicit values or defaults

## Implementation Details

The validation now works in all scenarios:
- Both timeouts explicitly configured
- One timeout explicit, one using default
- Both timeouts using defaults (120s each)

Error messages clearly indicate the source:
- `querier.timeout (150s) must be less than or equal to server.http-write-timeout (default: 120s)`


CHANGELOG.md
```diff
@@ -96,6 +96,7 @@
 * [CHANGE] Ingester: Disable shipping of blocks on the third zone (zone-c) when using `ingest_storage_ingester_zones: 3` on ingest storage #12743 #12744
 * [CHANGE] Distributor: Increase `server.grpc-max-concurrent-streams` from 100 to 1000. #12742
 * [CHANGE] Ruler Query Frontend: Increase `server.grpc-max-concurrent-streams` from 100 to 300. #12742
+* [ENHANCEMENT] Add timeout validation for querier and query-frontend. Enhanced `parseDuration` to support milliseconds and combined formats (e.g., "4m30s").
```
```suggestion
* [ENHANCEMENT] Add timeout validation for querier and query-frontend. Enhanced `parseDuration` to support milliseconds and combined formats (e.g., "4m30s"). #12766
```

operations/mimir/querier.libsonnet
```diff
@@ -35,6 +35,28 @@
...
+  local validateQuerierTimeouts() =
+    local q_timeout = if 'querier.timeout' in $.querier_args then
+      $.util.parseDuration($.querier_args['querier.timeout'])
+    else null;
+
+    local q_write_timeout = if 'server.http-write-timeout' in $.querier_args then
+      $.util.parseDuration($.querier_args['server.http-write-timeout'])
+    else null;
+
+    assert q_timeout == null || q_write_timeout == null || q_timeout <= q_write_timeout :
```
alexweav, line 50: Does it make sense to add the defaults for these into the jsonnet, so the assertion works when one of the args is unprovided?

dimitarvdimitrov: really good point. i think i have a solution. I don't know if it's not overengineered, but i think it will at least prevent bugs https://github.com/grafana/mimir/pull/12766/commits/c69bf2b76cdbc56466e6015eb539314c220db31c

PR #12772, opened 2025-09-19
Block-builder-scheduler: safe handling of multiple concurrent jobs in a partition

seizethedave:
#### What this PR does

Add a buffer of active completions where a job is only garbage collected and used to advance the commit when both:
1. it is complete
2. all known jobs with a lower offset have both been completed and garbage collected.

This is the remaining safety mechanism required for us to relax the `max_jobs_per_partition` setting and consume jobs in parallel within a partition. 

There are two features of the scheduler that make use of this:

1. In normal operation, where the scheduler is creating new jobs, assigning them, and workers are completing them.
1. At startup: in `finalizeObservations` we process observed job updates and form an idea of planned and completed jobs. Any such observations whose start and end offsets form a contiguous offset range starting from the stored commit will be added (in ascending order) to the plannedJobs structure. Further, any such jobs which are _completed_ in `finalizeObservation` will have `partitionState.completeJob` called on them, which contains the safe offset advancement/GCing mentioned previously.

After we've amassed a certain level of confidence, the "pending" job system and the job creation policy things can all be deleted, as they only existed as a safety measure until we had this safe book-keeping implemented.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/blockbuilder/scheduler/scheduler.go
```diff
@@ -995,17 +1045,17 @@ const (
 // expected to be monotonically increasing and contiguous. Advance will not
 // allow backwards movement. If a gap is detected, a warning is logged and a
 // metric is incremented.
-func (o *advancingOffset) advance(key jobKey, spec schedulerpb.JobSpec) {
+func (o *advancingOffset) advance(jobID string, spec schedulerpb.JobSpec) {
```
seizethedave, line 1052: [note: a jobKey contains an epoch, which relates to the current assignment. I've replaced a number of jobKeys in this file with the ID, since they're agnostic to the assignment, and I was often passing `jobKey{id, 0}` anyway.]

pkg/blockbuilder/scheduler/scheduler_test.go
```diff
@@ -554,18 +554,19 @@ func TestOffsetMovement(t *testing.T) {
...
 	require.NoError(t, e)
+	ps.addPlannedJob("ingest/1/5524", spec)
 	key, _, err := sched.jobs.assign("w0")
 	require.NoError(t, err)
 
 	require.NoError(t, sched.updateJob(key, "w0", false, spec))
 	sched.requireOffset(t, "ingest", 1, 5000, "ingest/1 is in progress, so we should not move the offset")
 	require.NoError(t, sched.updateJob(key, "w0", true, spec))
-	sched.requireOffset(t, "ingest", 1, 6000, "ingest/1 is in progress, so we should be advanced")
+	sched.requireOffset(t, "ingest", 1, 6000, "ingest/1 is complete, so we should advance")
```
seizethedave, line 564: a typo.

narqo (approved):
Looks good to me 🔥

CHANGELOG.md
```diff
@@ -45,7 +45,7 @@
...
 * [ENHANCEMENT] MQE: Add support for applying common subexpression elimination to range vector expressions in instant queries. #12236
 * [ENHANCEMENT] Ingester: Improve the performance of active series custom trackers matchers. #12184
-* [ENHANCEMENT] Ingester: Add postings cache sharing and invalidation. Sharing and head cache invalidation can be enabled via `-blocks-storage.tsdb.shared-postings-for-matchers-cache` and `-blocks-storage.tsdb.head-postings-for-matchers-cache-invalidation` respectively, and the number of metric versions per cache can be configured via -`blocks-storage.tsdb.head-postings-for-matchers-cache-versions`. #12333
+* [ENHANCEMENT] Ingester: Add postings cache sharing and invalidation. Sharing and head cache invalidation can be enabled via `-blocks-storage.tsdb.shared-postings-for-matchers-cache` and `-blocks-storage.tsdb.head-postings-for-matchers-cache-invalidation` respectively, and the number of metric versions per cache can be configured via `-blocks-storage.tsdb.head-postings-for-matchers-cache-versions`. #12333
```
```suggestion
* [ENHANCEMENT] Ingester: Add postings cache sharing and invalidation. You can enable sharing and head cache invalidation via `-blocks-storage.tsdb.shared-postings-for-matchers-cache` and `-blocks-storage.tsdb.head-postings-for-matchers-cache-invalidation` respectively, and you can configure the number of metric versions per cache via `-blocks-storage.tsdb.head-postings-for-matchers-cache-versions`. #12333
```

tacole02 (commented):

PR #12783, opened 2025-09-23
Fix bug with pruning toggles not working in query planner

zenador:
#### What this PR does

Pruning query toggles was broken with https://github.com/grafana/mimir/pull/12303 when it moved to AST optimization passes, as the query planner preprocess step wraps some parts of the expression in StepInvariantExpressions which we did not account for in the code and unit tests working on it in isolation instead of as part of the query plan.

This fixes that bug and adds preprocessing to the unit tests for some AST optimization passes. Will do testing with more of the query plan in another [PR](https://github.com/grafana/mimir/pull/12784) as that requires bigger changes to the planning code.

Follow up to https://github.com/grafana/mimir/pull/12303

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,7 @@
 * [BUGFIX] Ruler: Fix ruler remotequerier request body consumption on retries. #12514
 * [BUGFIX] Block-builder: Fix a bug where a consumption error can cause a job to stay assigned to a worker for the remainder of its lifetime. #12522
 * [BUGFIX] Querier: Fix possible panic when evaluating a nested subquery where the parent has no steps. #12524
-* [BUGFIX] Querier: Fix bug with pruning toggles AST optimization pass not working in query planner. #12783
+* [BUGFIX] Querier: Fix bug where the pruning toggles AST optimization pass doesn't work in the query planner. #12783
 * [BUGFIX] Ingester: Fix a bug where prepare-instance-ring-downscale endpoint would return an error while compacting and not read-only. #12548
 * [BUGFIX] Block-builder: Fix a bug where lease renewals would cease during graceful shutdown, leading to an elevated rate of job reassignments. #12643
 * [BUGFIX] OTLP: Return HTTP OK for partially rejected requests, e.g. due to OOO exemplars. #12579
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
 * [BUGFIX] Ruler: Fix ruler remotequerier request body consumption on retries. #12514
 * [BUGFIX] Block-builder: Fix a bug where a consumption error can cause a job to stay assigned to a worker for the remainder of its lifetime. #12522
 * [BUGFIX] Querier: Fix possible panic when evaluating a nested subquery where the parent has no steps. #12524
+* [BUGFIX] Querier: Fix bug with pruning toggles AST optimization pass not working in query planner. #12783
 * [BUGFIX] Ingester: Fix a bug where prepare-instance-ring-downscale endpoint would return an error while compacting and not read-only. #12548
 * [BUGFIX] Block-builder: Fix a bug where lease renewals would cease during graceful shutdown, leading to an elevated rate of job reassignments. #12643
 * [BUGFIX] OTLP: Return HTTP OK for partially rejected requests, e.g. due to OOO exemplars. #12579
--- a/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
+++ b/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
@@ -37,6 +37,7 @@ var testCasesReorderHistogramAggregation = map[string]string{
 	"histogram_sum(rate(foo[2m]))":              "histogram_sum(rate(foo[2m]))",
 	`3 + (((histogram_sum(sum(foo)))))`:         `3 + (((sum(histogram_sum(foo)))))`,
 	`vector(3) + (((histogram_sum(sum(foo)))))`: `vector(3) + (((sum(histogram_sum(foo)))))`,
+	`histogram_sum(sum(foo @ 1234 * bar))`:      `sum(histogram_sum(foo @ 1234 * bar))`,
 
 	// Unsupported aggregations
 	`histogram_sum(max(foo))`:     `histogram_sum(max(foo))`,
--- a/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
+++ b/pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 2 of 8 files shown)


zenador:
@charleskorn Only the change in the 1st commit is actually required for the fix. Accounted for Step Invariant Expressions in more parts of the code in the 2nd commit, but they are not actually necessary right now, should we still go ahead with the changes in the 2nd commit?

Also since this wasn't actually running, when we roll it out should we disable it first, and then enable it after a day or so of the new version so we can see the effect?

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -17,8 +17,11 @@ import (
 
 	"github.com/grafana/mimir/pkg/streamingpromql"
 	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
 )
 
+var instantQueryTimeRange = types.NewInstantQueryTimeRange(timestamp.Time(1000))
```
charleskorn: [nit] What about something like `dummyTimeRange` to make it clear this isn't intended to be used?

pkg/streamingpromql/optimize/ast/reorder_histogram_aggregation_test.go
```diff
@@ -64,6 +66,9 @@ func TestReorderHistogramAggregation(t *testing.T) {
 
 			inputExpr, err := parser.ParseExpr(input)
 			require.NoError(t, err)
+			inputExpr, err = promql.PreprocessExpr(inputExpr, timestamp.Time(instantQueryTimeRange.StartT), timestamp.Time(instantQueryTimeRange.EndT), 0)
```
charleskorn: It might be good to pull this out to a helper method in `util_test.go` so we don't have to repeat it everywhere (and to make the intention clearer).

charleskorn:
> @charleskorn Only the change in the 1st commit is actually required for the fix. Accounted for Step Invariant Expressions in more parts of the code in the 2nd commit, but they are not actually necessary right now, should we still go ahead with the changes in the 2nd commit?

I think it can't hurt - could you add test cases that exercise these changes?

> Also since this wasn't actually running, when we roll it out should we disable it first, and then enable it after a day or so of the new version so we can see the effect?

Given this functionality was previously running, I think it's fine as-is.

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -12,13 +12,21 @@ import (
...
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/promql/promqltest"
 	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/mimir/pkg/streamingpromql"
 	"github.com/grafana/mimir/pkg/streamingpromql/testutils"
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
 )
 
+var dummyTimeRange = types.NewInstantQueryTimeRange(timestamp.Time(1000))
```
charleskorn: Given this is now only used in `preprocessQuery`, could you please move it into there?

zenador:
> I think it can't hurt - could you add test cases that exercise these changes?

Tried but not really sure how to, sorry. The `StepInvariantExpr` only wraps `vector(X)` (couldn't find any other examples), so it's not really relevant to reordering histogram aggregations and propagating matchers as those work on vector selectors. As for pruning toggles, the existing tests fail if this check isn't added to `isConst`, but given the current logic of how the wrapping works in Prometheus' `preprocessExpr`, it will never wrap the individual sides in binary expressions like `vector(0) = 0`, so `isConstVector` and `isNumber` won't see any `StepInvariantExpr` here. That's why I said these changes aren't really necessary, but it's just there to be thorough (and in case the `preprocessExpr` logic or something related changes in future).

> Given this functionality was previously running, I think it's fine as-is.

I was thinking more as a way to measure the impact, but it's fine either way!

charleskorn:
> Tried but not really sure how to, sorry. The `StepInvariantExpr` only wraps `vector(X)` (couldn't find any other examples), so it's not really relevant to reordering histogram aggregations and propagating matchers as those work on vector selectors.

For reordering histogram aggregations: expressions like `histogram_sum(sum(foo @ start()))`, `histogram_sum(sum(foo @ 1234))` and `histogram_sum(sum(foo @ 1234)) * bar` should have some part of the expression wrapped in a `StepInvariantExpr`.

Similar thing for propagating matchers: adding `@ ...` to one or both sides would make it step-invariant.

charleskorn:
> > Given this functionality was previously running, I think it's fine as-is.
> 
> I was thinking more as a way to measure the impact, but it's fine either way!

As sorry, I misunderstood your intention - if you want to do that to make the impact easier to see, that sounds good to me.

CHANGELOG.md
```diff
@@ -72,6 +72,7 @@
 * [BUGFIX] Ruler: Fix ruler remotequerier request body consumption on retries. #12514
 * [BUGFIX] Block-builder: Fix a bug where a consumption error can cause a job to stay assigned to a worker for the remainder of its lifetime. #12522
 * [BUGFIX] Querier: Fix possible panic when evaluating a nested subquery where the parent has no steps. #12524
+* [BUGFIX] Querier: Fix bug with pruning toggles AST optimization pass not working in query planner. #12783
```
```suggestion
* [BUGFIX] Querier: Fix bug where the pruning toggles AST optimization pass doesn't work in the query planner. #12783
```

tacole02: I think this reads a bit easier, but let me know if the meaning isn't as intended.

zenador:
Propagating matchers already has this test case:
```
	`up{foo="bar"} @ 1 + down @ 2`:                                             `up{foo="bar"} @ 1 + down{foo="bar"} @ 2`,
```
Tested that by setting breakpoints and debugging, somehow `StepInvariantExpr`s don't seem to be there at all even after preprocessing, there are just regular vector selectors. I also tried adding a similar test case with `start()` but it was the same.

Also tested with the histogram aggregation test cases you mentioned and it was the same.

charleskorn:
> Propagating matchers already has this test case:
> 
> ```
> 	`up{foo="bar"} @ 1 + down @ 2`:                                             `up{foo="bar"} @ 1 + down{foo="bar"} @ 2`,
> ```
> 
> Tested that by setting breakpoints and debugging, somehow `StepInvariantExpr`s don't seem to be there at all even after preprocessing, there are just regular vector selectors. I also tried adding a similar test case with `start()` but it was the same.
> 
> Also tested with the histogram aggregation test cases you mentioned and it was the same.

Something's going wrong somewhere then - that should definitely be getting wrapped in a `StepInvariantExpr`.

zenador:
Turns out the step invariant expressions were wrapped around the whole expression for those cases, and there weren't any inner ones to be affected. Added test cases where the step invariant expressions were inside, which only pass with the new additions.

charleskorn (approved):

PR #12784, opened 2025-09-23
Test pruning toggles AST optimization pass with query planner

zenador:
#### What this PR does

Make our unit testing of the pruning toggles more thorough by testing it as part of the query planner.

Just pruning toggles for now as the other new ones are applied as frontend middlewares.

Follow up to https://github.com/grafana/mimir/pull/12783

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

Only updates tests, not user facing.

### DIFF
--- a/pkg/streamingpromql/optimize/ast/util_test.go
+++ b/pkg/streamingpromql/optimize/ast/util_test.go
@@ -33,16 +33,6 @@ func runASTOptimizationPass(t *testing.T, ctx context.Context, input string, cre
 	reg := prometheus.NewPedanticRegistry()
 	opts.CommonOpts.Reg = reg
 	optimizer := createOptimizerFunc(opts.CommonOpts.Reg)
-	outputExpr := runASTOptimizationPassInner(t, ctx, input, optimizer, opts)
-	return reg, outputExpr
-}
-
-func runASTOptimizationPassWithoutMetrics(t *testing.T, ctx context.Context, input string, optimizer optimize.ASTOptimizationPass) parser.Expr {
-	opts := streamingpromql.NewTestEngineOpts()
-	return runASTOptimizationPassInner(t, ctx, input, optimizer, opts)
-}
-
-func runASTOptimizationPassInner(t *testing.T, ctx context.Context, input string, optimizer optimize.ASTOptimizationPass, opts streamingpromql.EngineOpts) parser.Expr {
 	dummyTimeRange := types.NewInstantQueryTimeRange(timestamp.Time(1000))
 	planner, err := streamingpromql.NewQueryPlannerWithoutOptimizationPasses(opts)
 	require.NoError(t, err)
@@ -50,6 +40,13 @@ func runASTOptimizationPassInner(t *testing.T, ctx context.Context, input string
 	observer := streamingpromql.NoopPlanningObserver{}
 	outputExpr, err := planner.ParseAndApplyASTOptimizationPasses(ctx, input, dummyTimeRange, observer)
 	require.NoError(t, err)
+	return reg, outputExpr
+}
+
+func runASTOptimizationPassWithoutMetrics(t *testing.T, ctx context.Context, input string, optimizer optimize.ASTOptimizationPass) parser.Expr {
+	_, outputExpr := runASTOptimizationPass(t, ctx, input, func(prometheus.Registerer) optimize.ASTOptimizationPass {
+		return optimizer
+	})
 	return outputExpr
 }
 


pkg/streamingpromql/planning.go
```diff
@@ -129,20 +129,7 @@ type PlanningObserver interface {
...
-	queryID, err := p.activeQueryTracker.InsertWithDetails(ctx, qs, "planning", timeRange)
-	if err != nil {
-		return nil, err
-	}
-
-	defer p.activeQueryTracker.Delete(queryID)
-
-	spanLogger.DebugLog("msg", "starting planning", "expression", qs)
-
+func (p *QueryPlanner) RunPrePlanningStages(ctx context.Context, qs string, timeRange types.QueryTimeRange, observer PlanningObserver) (parser.Expr, error) {
```
charleskorn: Perhaps something like this would be clearer?

```suggestion
func (p *QueryPlanner) ParseAndApplyASTOptimizationPasses(ctx context.Context, qs string, timeRange types.QueryTimeRange, observer PlanningObserver) (parser.Expr, error) {
```

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -97,6 +98,29 @@ func checkPruneTogglesMetrics(t *testing.T, g prometheus.Gatherer, expectedTotal
 	require.NoError(t, testutil.GatherAndCompare(g, strings.NewReader(expectedMetrics), metricNameTotal, metricNameChanged))
 }
 
+func TestPruneTogglesWithQueryPlan(t *testing.T) {
```
charleskorn: Rather than introducing a new unit test for this, what if we changed the existing unit tests to follow this structure?

(and not just this optimisation pass, but all of them)

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
```diff
@@ -97,6 +98,29 @@ func checkPruneTogglesMetrics(t *testing.T, g prometheus.Gatherer, expectedTotal
...
+	opts := streamingpromql.NewTestEngineOpts()
+	opts.CommonOpts.Reg = reg
+	planner, err := streamingpromql.NewQueryPlannerWithoutOptimizationPasses(opts)
+	require.NoError(t, err)
+	o := ast.NewPruneToggles(opts.CommonOpts.Reg)
+	planner.RegisterASTOptimizationPass(o)
+	observer := streamingpromql.NoopPlanningObserver{}
+
+	for input, expected := range testCasesPruneToggles {
+		t.Run("input", func(t *testing.T) {
```
charleskorn:
```suggestion
		t.Run(input, func(t *testing.T) {
```

pkg/streamingpromql/optimize/ast/prune_toggles_test.go
zenador: Didn't make this change for reordering histograms and propagating matchers as those are currently frontend middlewares, not optimization passes, so can't be added to the query plan.

pkg/streamingpromql/planning.go
```diff
@@ -129,20 +129,7 @@ type PlanningObserver interface {
...
-	queryID, err := p.activeQueryTracker.InsertWithDetails(ctx, qs, "planning", timeRange)
-	if err != nil {
-		return nil, err
-	}
-
-	defer p.activeQueryTracker.Delete(queryID)
-
-	spanLogger.DebugLog("msg", "starting planning", "expression", qs)
-
+func (p *QueryPlanner) ParseAndApplyASTOptimizationPasses(ctx context.Context, qs string, timeRange types.QueryTimeRange, observer PlanningObserver) (parser.Expr, error) {
```
charleskorn, line 135: Might be good to add a comment here explaining that this method primarily exists for testing AST optimisation passes.

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -26,6 +28,21 @@ func preprocessQuery(t *testing.T, expr parser.Expr) (parser.Expr, error) {
 	return promql.PreprocessExpr(expr, timestamp.Time(dummyTimeRange.StartT), timestamp.Time(dummyTimeRange.EndT), 0)
 }
 
+func getOutputFromASTOptimizationPassWithQueryPlan(t *testing.T, ctx context.Context, input string, createOptimizerFunc func(prometheus.Registerer) optimize.ASTOptimizationPass) (*prometheus.Registry, parser.Expr) {
```
charleskorn: Might be nice to introduce a helper method that takes a optimisation pass as a parameter that calls this new method that can be used by the tests that don't care about metrics.

eg. the collapse constants tests can use the extra method, while the prune toggles tests can use `getOutputFromASTOptimizationPassWithQueryPlan`

charleskorn: [nit] Perhaps `runASTOptimizationPass` might be a clearer name?

pkg/streamingpromql/planning.go
```diff
@@ -129,6 +129,9 @@ type PlanningObserver interface {
 	OnAllPlanningStagesComplete(finalPlan *planning.QueryPlan) error
 }
 
+// ParseAndApplyASTOptimizationPasses runs the AST optimization passes on the input string and outputs
+// an expression and any error encountered. This is separated into its own method to allow testing of
+// the AST optimization passes in the query planning.
```
```suggestion
// an expression and any error encountered. This is separated into its own method to allow testing of
// AST optimization passes.
```

pkg/streamingpromql/optimize/ast/util_test.go
```diff
@@ -28,19 +28,29 @@ func preprocessQuery(t *testing.T, expr parser.Expr) (parser.Expr, error) {
 	return promql.PreprocessExpr(expr, timestamp.Time(dummyTimeRange.StartT), timestamp.Time(dummyTimeRange.EndT), 0)
 }
 
-func getOutputFromASTOptimizationPassWithQueryPlan(t *testing.T, ctx context.Context, input string, createOptimizerFunc func(prometheus.Registerer) optimize.ASTOptimizationPass) (*prometheus.Registry, parser.Expr) {
-	dummyTimeRange := types.NewInstantQueryTimeRange(timestamp.Time(1000))
-	reg := prometheus.NewPedanticRegistry()
+func runASTOptimizationPass(t *testing.T, ctx context.Context, input string, createOptimizerFunc func(prometheus.Registerer) optimize.ASTOptimizationPass) (*prometheus.Registry, parser.Expr) {
```
charleskorn, line 31: I think we can simplify this further: `runASTOptimizationPassWithoutMetrics` can call `runASTOptimizationPass` passing a `createOptimizerFunc` value that just returns the optimization pass provided to `runASTOptimizationPassWithoutMetrics`. Then there's no need for `runASTOptimizationPassInner`.

zenador: Ah okay, I thought we were trying to avoid creating and passing the registry unnecessarily, but I guess we just wanted a simpler signature? Updated.

charleskorn: In tests, creating the unnecessary registry isn't a big issue.

Issue #12673, opened 2025-09-09
Remove anything other than chunks streaming between queriers and ingesters / store-gateways


56quarters:
_copied from internal issue_

### What should we deliver?

Queriers currently support loading chunks from ingesters and store-gateways by both streaming chunks during query evaluation and by reading all chunks into memory upfront. 

Streaming chunks from both sources has been the default for quite some time, and supporting both streaming and non-streaming cases adds complexity to both queriers, ingesters and store-gateways.

However, we can't just remove the non-streaming case, as that will break backwards compatibility with older queriers that don't support streaming chunks. This would make upgrades more difficult. (Queriers already gracefully fallback to the non-streaming case if ingesters or store-gateways don't support streaming chunks.)

### How will we do it (roughly)?

I propose we do this before we release Mimir 3.0.

Specifically, we'd need to: 
* Remove non-streaming code from queriers, ingesters and store-gateways
* Add clear advice to the release notes on how to upgrade from older versions without issue (ie. upgrade all components to the last version where both cases are supported, ensure chunks streaming is enabled in queriers, then upgrade to 3.0)

### Size (development only)

Small (<= 1 month dev)

### What are the documentation dependencies?

We'll need to include clear upgrade instructions in the release notes.

### Enablement or training?

(none)

### Proposer?

@charleskorn

56quarters:
Another important thing to note: we use the non-streaming code path in [store-gateways](https://github.com/grafana/mimir/blob/c9c7de150a1c98f732daf6a83afa9d67187d964f/pkg/storegateway/bucket.go#L558-L561) when serving requests for the Prometheus `/api/v1/series` API. In this case, the `Series` RPC call to store-gateways sets the `SkipChunks` flag in the request which uses the non-streaming path to fetch only series labels an no chunks. I don't think there's anything preventing us from using the streaming path for this besides needing to implement it.

56quarters:
This is complete as of https://github.com/grafana/mimir/pull/13026

PR #12790, opened 2025-09-23
refactor: Require non-zero size for querier streaming buffers

56quarters:
#### What this PR does

This change requires a non-zero value for the size of buffers used when streaming results between queriers and ingesters or store-gateways. This has the effect of requiring streaming to always be enabled. Enabling streaming has been the default since Mimir 2.14.

Part of #12673

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


56quarters:
Note for reviewers: actual removal of non-streaming code is more involved (since it is used in some internal request paths) and will likely be done in follow-up PRs after Mimir 3.0.

PR #12799, opened 2025-09-24
distributor: Report correct size in "message too big" error

tcard:
#### What this PR does

In the Prometheus push endpoint, when the write request _after_ compression exceeds the limit, the HTTP request's content length (this _before_ compression) would be reported, so you could see an error like this:

```
the incoming push request has been rejected because its message size of 16245 bytes is larger than the allowed limit of 100000 bytes
```

With this change:

* Compressed and uncompressed sizes are now handled explicitly.
* The error message always has either `(compressed)` or `(uncompressed)` to make it always clear which one it's talking about.
* The implementation of the several "message too large" error types have been made uniform.

—

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/util/http.go
```diff
@@ -179,11 +179,17 @@ func ParseProtoReader(ctx context.Context, reader io.Reader, expectedSize, maxSi
 }
 
 type MsgSizeTooLargeErr struct {
-	Actual, Limit int
+	Compressed, Actual, Limit int
```
56quarters, line 182: What do you think about adding a constructor for this that prevents both `Compressed` and `Actual` from being set at the same time to prevent misuse?

pkg/distributor/push.go
```diff
@@ -115,25 +115,29 @@ func Handler(
...
-	msgSizeDesc := fmt.Sprintf(" of %d bytes", e.actual)
-	if e.actual < 0 {
-		msgSizeDesc = ""
+	msgSizeDesc := ""
+	if e.actual > 0 {
+		msgSizeDesc = fmt.Sprintf(" of %d bytes (decompressed)", e.actual)
+	} else if e.compressed > 0 {
+		msgSizeDesc = fmt.Sprintf(" of %d bytes (compressed)", e.compressed)
 	}
 	return globalerror.DistributorMaxOTLPRequestSize.MessageWithPerInstanceLimitConfig(fmt.Sprintf("the incoming OTLP request has been rejected because its message size%s is larger than the allowed limit of %d bytes", msgSizeDesc, e.limit), maxOTLPRequestSizeFlag)
```
56quarters, line 142: Is it possible to not have either the compressed or actual size?

pkg/distributor/otel_test.go
```diff
@@ -1164,9 +1164,9 @@ func TestHandlerOTLPPush(t *testing.T) {
 			},
 			responseCode:          http.StatusRequestEntityTooLarge,
 			responseContentType:   pbContentType,
-			responseContentLength: 292,
-			errMessage:            "the incoming OTLP request has been rejected because its message size of 89 bytes is larger",
-			expectedLogs:          []string{`level=warn user=test msg="detected an error while ingesting OTLP metrics request (the request may have been partially ingested)" httpCode=413 err="rpc error: code = Code(413) desc = the incoming OTLP request has been rejected because its message size of 89 bytes is larger than the allowed limit of 30 bytes (err-mimir-distributor-max-otlp-request-size). To adjust the related limit, configure -distributor.max-otlp-request-size, or contact your service administrator." insight=true`},
+			responseContentLength: 307,
```
56quarters, line 1167: I see that this isn't new but this seems very fragile. Where does this number come from? Can we verify it some other way?

pkg/distributor/otel_test.go
```diff
@@ -1229,9 +1229,9 @@ func TestHandlerOTLPPush(t *testing.T) {
...
+			responseContentLength: 308,
+			errMessage:            "the incoming OTLP request has been rejected because its message size of 106 bytes (decompressed) is larger",
+			expectedLogs:          []string{`level=warn user=test msg="detected an error while ingesting OTLP metrics request (the request may have been partially ingested)" httpCode=413 err="rpc error: code = Code(413) desc = the incoming OTLP request has been rejected because its message size of 106 bytes (decompressed) is larger than the allowed limit of 30 bytes (err-mimir-distributor-max-otlp-request-size). To adjust the related limit, configure -distributor.max-otlp-request-size, or contact your service administrator." insight=true`},
```
56quarters: Personal preference but I really don't like testing for extremely specific log messages like this. It makes the tests fragile and is testing for implementation specific side-effects instead of the response for the method on question.

tcard: I had the same thought (also about https://github.com/grafana/mimir/pull/12799#discussion_r2375830990) while updating those tests, but I wanted to keep the PR focused.

I'll open a separate PR relaxing those tests.

pkg/util/http.go
tcard: Done at [8250a2f](https://github.com/grafana/mimir/pull/12799/commits/8250a2ff96462165185346a1c0d6c6706b425f79).

pkg/distributor/push.go
tcard: Yes, because we only decompress requests until reaching the limit, so we don't exactly know by how much a compressed request is over the limit after decompression.

I added a specific constructor for this case at [8250a2f](https://github.com/grafana/mimir/pull/12799/commits/8250a2ff96462165185346a1c0d6c6706b425f79).

56quarters (approved):

CHANGELOG.md
```diff
@@ -88,6 +88,7 @@
 * [BUGFIX] Ingester, Block-builder: silently ignore duplicate sample if it's due to zero sample from created timestamp. Created timestamp equal to the timestamp of the first sample of series is a common case if created timestamp comes from OTLP where start time equal to timestamp of the first sample simply means unknown start time. #12726
 * [BUGFIX] Distributor: Fix error when native histograms bucket limit is set then no NHCB passes validation. #12741
 * [BUGFIX] Ingester: Fix continous reload of active series counters when cost-attribution labels are above the max cardinality. #12822
+* [BUGFIX] Distributor: Report correct size in `err-mimir-distributor-max-write-message-size` error. #12799
```
```suggestion
* [BUGFIX] Distributor: Report the correct size in the `err-mimir-distributor-max-write-message-size` error. #12799
```

tacole02 (approved):

PR #12792, opened 2025-09-23
mimirpb: Refactor RW2 metadata handling out of generated code

alexweav:
#### What this PR does

The RW2 unmarshal uses a `map[string]*orderAwareMetricMetadata` that's injected everywhere, used to track and deduplicate metadata, and also store them as we convert back to RW1.

This PR is a refactor that puts this behind a `metadataSet` interface. We then introduce a `dedupingMetadataSet` implementation that just wraps the very same map. In the future we want to introduce alternate metadata handling strategies that don't deduplicate, this makes it easy.

Just converting it to a `map[string][]*orderAwareMetricMetadata` is too unwieldy and allocations inside the inner slices are difficult to control, this approach is more minimal. It also means we can move a little custom logic out of `mimir.pb.go` and into regular Go files, reducing the divergence of our edits to the generated code.

This abstraction doesn't cause any perf difference or extra allocations.
<details>

<summary>benchmark results</summary>

Run with:
```
go test ./pkg/mimirpb/... -run ^$ -bench BenchmarkUnMarshal -benchtime 5s -benchmem
```

```
goos: linux
goarch: amd64
pkg: github.com/grafana/mimir/pkg/mimirpb
cpu: AMD Ryzen 7 PRO 5850U with Radeon Graphics
                                               │  before.txt  │              after.txt               │
                                               │    sec/op    │    sec/op     vs base                │
UnMarshal/Marshal/RW1-16                         39.04m ± ∞ ¹   38.45m ± ∞ ¹       ~ (p=1.000 n=1) ²
UnMarshal/Marshal/RW2-16                         21.19m ± ∞ ¹   21.08m ± ∞ ¹       ~ (p=1.000 n=1) ²
UnMarshal/Unmarshal/RW1/skipExemplars=true-16    24.75m ± ∞ ¹   23.54m ± ∞ ¹       ~ (p=1.000 n=1) ²
UnMarshal/Unmarshal/RW2/skipExemplars=true-16    26.15m ± ∞ ¹   26.80m ± ∞ ¹       ~ (p=1.000 n=1) ²
UnMarshal/Unmarshal/RW1/skipExemplars=false-16   27.58m ± ∞ ¹   26.16m ± ∞ ¹       ~ (p=1.000 n=1) ²
UnMarshal/Unmarshal/RW2/skipExemplars=false-16   28.39m ± ∞ ¹   29.15m ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                          27.36m         27.03m        -1.20%
[...]

PR #12804, opened 2025-09-24
ingest-storage: Do not deduplicate metadata by family name when unmarshalling

alexweav:
#### What this PR does

This PR builds on the refactor in https://github.com/grafana/mimir/pull/12792.
By default, RW2 deserialization deduplicates metric metadata by family name. All metadata beyond the first instance for a given family are discarded. This PR adds a new deserialization option, called `SkipDeduplicateMetadata`, plus a new `metadataSet` implementation that avoids the deduplication and preserves all given metadata and ordering.

This is done so that we can preserve metadata to the letter, even if it conflicts, ensuring we [adhere to prometheus behavior](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metadata) particularly in RW1.0.

RW1 requests don't have the duplication that natural RW2 requests tend to have. Natural RW2 requests will already be deduplicated in the distributor before they reach the ingest-storage layer, so this change does not add undue duplication for clients sending RW2.0 into Mimir.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/mimirpb/timeseries.go
+++ b/pkg/mimirpb/timeseries.go
@@ -76,6 +76,11 @@ type PreallocWriteRequest struct {
 	// SkipNormalizeMetadataMetricName skips normalization of metric name in metadata on unmarshal. E.g., don't remove `_count` suffixes from histograms.
 	// Has no effect on marshalled or existing structs; must be set prior to Unmarshal calls.
 	SkipNormalizeMetadataMetricName bool
+	// SkipDeduplicateMetadata skips deduplication of RW2 metadata by metric family name.
+	// Normally this is done because RW2 requests to repeat metadata as it's embedded in timeseries.
+	// Some applications, like RW1->RW2 translation, might choose to disable it.
+	// Has no effect on marshalled or existing strucs; must be set prior to Unmarshal calls.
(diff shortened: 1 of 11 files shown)


pkg/mimirpb/timeseries.go
```diff
@@ -76,6 +76,11 @@ type PreallocWriteRequest struct {
 	// SkipNormalizeMetadataMetricName skips normalization of metric name in metadata on unmarshal. E.g., don't remove `_count` suffixes from histograms.
 	// Has no effect on marshalled or existing structs; must be set prior to Unmarshal calls.
 	SkipNormalizeMetadataMetricName bool
+	// SkipDeduplicateMetadata skips deduplication of RW2 metadata by metric family name.
+	// Normally this is done because RW2 requests to repeat metadata as it's embedded in timeseries.
+	// Some applications, like RW1->RW2 translation, might choose to disable it.
+	// Has no effect on marshalled or existing strucs; must be set prior to Unmarshal calls.
```
pracucci: [nit]
```suggestion
	// Has no effect on marshalled or existing structs; must be set prior to Unmarshal calls.
```

Issue #11918, opened 2025-06-30
index planning: Add metrics to validate impact


dimitarvdimitrov:
Implement the necessary metrics to help validate the impact of index lookup planning changes. Focus on items identified in the [Instrumenting ingester queries](https://github.com/grafana/mimir/issues/11916#instrumenting-ingester-queries) section, ensuring that the metrics will provide actionable insights.

A few useful metrics to have
- the portion of request time  that the ingester spends on evaluating regular expressions
- as close as possible on the volume of postrings, series & chunks

### Open PRs as part of this issue (updated)

- [x] https://github.com/prometheus/prometheus/pull/16853
- [x] https://github.com/grafana/mimir-prometheus/pull/935 (change to mimir-prometheus, so we aren't blocked on the upstream reviews)
- [x] https://github.com/grafana/mimir/pull/12042
- [ ] ~~allow exporting native histograms in dskit metrics registries~~ irrelevant after https://github.com/grafana/mimir/pull/12752 where we keep metrics in our own scope
  - [ ] https://github.com/grafana/dskit/pull/735
  - [ ] https://github.com/grafana/dskit/pull/736
- [ ] how many series we fetch from the index just using the index matchers and then how many we end up selecting out of those to return to the query

dimitarvdimitrov:
### Regex matching

we got early access to profiles-to-metrics and can track the percentage of CPU spent on regular expressions.

I'm also looking into adding metrics for the time spent running regular expression per query. I started with an approach where we do `time.Now()` and `time.Since()` on every `Matcher.Matches`, this appears very expensive, so I'm exploring other options. (code for now is [on this branch and needs some small changes in (mimir-)prometheus](https://github.com/grafana/mimir/compare/main...dimitar%2Fr349%2Fwith-regex-time-tracking?body=))

### Cardinality estimations

I looked at the metrics we want to add, and the one that stands out for planning is the Q error—the ratio between estimated and actual cardinality. Adding it directly to Prometheus feels a bit forced since there’s no real estimation happening, and it would result in every Prometheus emitting this metric for no clear reason.

If we *did* add it there, it would be somewhat invasive but doable. We'd need to modify block and head queries to include metrics and adjust how we construct those queries. It’s a bit messy and might break downstream Prometheus projects.

[...]

PR #12807, opened 2025-09-24
index planning: Add query cardinality metrics

dimitarvdimitrov:
## Summary

Add infrastructure to track estimated vs actual cardinalities in query processing to improve observability of lookup planning accuracy. This enables monitoring how well the cost-based planner estimates match actual query execution.

There are 4 key metrics: 
* estimated/actual selected postings
* estimated/actual final series

We track the ratio between actual/estimated + ratio between actual postings/actual final series

## Changes

- Add `lookupplan.QueryStats` tracker with atomic operations for thread-safe cardinality tracking
- Add 3 native histogram metrics for cardinality ratio analysis. Measurements there are observed with exemplars.
- Implement stats-tracking querier wrapper to measure actual cardinalities
- Integrate PostingsCloner to capture actual selected postings count
- Populate estimated fields in CostBasedPlanner from plan `intersectionSize()` & `cardinality()`

this is the final piece of https://github.com/grafana/mimir/issues/11918; closes https://github.com/grafana/mimir/issues/11918

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -2353,7 +2353,7 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
 		return err
 	}
 
-	i.metrics.queriedSeries.Observe(float64(numSeries))
+	i.metrics.queriedSeries.WithLabelValues("returned").Observe(float64(numSeries))
 	i.metrics.queriedSamples.Observe(float64(numSamples))
 	spanlog.DebugLog("series", numSeries, "samples", numSamples)
 	return nil
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -2353,7 +2353,7 @@ func (i *Ingester) QueryStream(req *client.QueryRequest, stream client.Ingester_
 		return err
 	}
 
-	i.metrics.queriedSeries.WithLabelValues("returned").Observe(float64(numSeries))
+	i.metrics.queriedSeries.WithLabelValues("send").Observe(float64(numSeries))
 	i.metrics.queriedSamples.Observe(float64(numSamples))
 	spanlog.DebugLog("series", numSeries, "samples", numSamples)
 	return nil
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -2956,10 +2956,10 @@ func (i *Ingester) createBlockChunkQuerier(userID string, b tsdb.BlockReader, mi
 		return nil, err
 	}
 
-	statsQuerier := newStatsTrackingChunkQuerier(defaultQuerier, i.metrics, i.lookupPlanMetrics.ForUser(userID))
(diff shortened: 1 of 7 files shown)


pkg/ingester/ingester.go
```diff
@@ -2948,24 +2948,28 @@ func (i *Ingester) createTSDB(userID string, walReplayConcurrency int) (*userTSD
...
-// createBlockChunkQuerier creates a BlockChunkQuerier that optionally wraps the default querier with mirroring
+// createBlockChunkQuerier creates a BlockChunkQuerier that wraps the default querier with stats tracking
+// and optionally with mirroring for comparison.
 func (i *Ingester) createBlockChunkQuerier(userID string, b tsdb.BlockReader, mint, maxt int64) (storage.ChunkQuerier, error) {
 	defaultQuerier, err := tsdb.NewBlockChunkQuerier(b, mint, maxt)
 	if err != nil {
 		return nil, err
 	}
 
+	statsQuerier := newStatsTrackingChunkQuerier(defaultQuerier, i.metrics, i.lookupPlanMetrics.ForUser(userID))
```
chencs: Maybe rename to `lookupStatsQuerier` or something like that? "Stats" as a concept is getting overloaded with this project, what with block statistics vs these query statistics (not to mention general "query stats")

pkg/ingester/lookupplan/query_stats.go
```diff
@@ -0,0 +1,134 @@
...
+func QueryStatsFromContext(ctx context.Context) *QueryStats {
+	o := ctx.Value(queryStatsCtxKey)
+	if o == nil {
+		return nil
+	}
+	return o.(*QueryStats)
+}
+
+// IsQueryStatsEnabled returns whether query stats tracking is enabled in the context.
+func IsQueryStatsEnabled(ctx context.Context) bool {
```
chencs: Seems like this is only used for testing (we just check for non-nil `QueryStatsFromContext` outside of test files); do we need it/should it be made private and moved to a test file?

dimitarvdimitrov: 👍 i just got rid of it. the tests were working fine without it

pkg/ingester/lookupplan/query_stats.go
```diff
@@ -0,0 +1,127 @@
...
+
+// ContextWithQueryStats returns a context with empty QueryStats.
+func ContextWithQueryStats(ctx context.Context, stats *QueryStats) context.Context {
+	return context.WithValue(ctx, queryStatsCtxKey, stats)
+}
+
+// QueryStatsFromContext gets the QueryStats out of the Context. Returns nil if stats have not
+// been initialised in the context. Note that QueryStats methods are safe to call with
+// a nil receiver.
+func QueryStatsFromContext(ctx context.Context) *QueryStats {
```
chencs, line 26: Does this need to be public?

dimitarvdimitrov: the method to add to the context is public and used from outside the package. it felt a bit weird to not expose this one too.

PR #12808, opened 2025-09-25
MQE: add version to query plans, and enforce that queriers only evaluate plans they support

charleskorn:
#### What this PR does

This PR adds a version number to query plans, and modifies the behaviour of `EncodedQueryPlan.ToDecodedPlan` to ensure that queriers only evaluate query plans they support.

#12302 and #12551

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/planning/plan.go
```diff
@@ -261,8 +271,12 @@ func NodeTypeName(n Node) string {
 // ToDecodedPlan converts this encoded plan to its decoded form.
 // It returns references to the specified nodeIndices.
 func (p *EncodedQueryPlan) ToDecodedPlan(nodeIndices ...int64) (*QueryPlan, []Node, error) {
+	if p.Version > MaximumSupportedQueryPlanVersion {
+		return nil, nil, apierror.Newf(apierror.TypeBadData, "query plan has version %v, but the maximum supported query plan version is %v", p.Version, MaximumSupportedQueryPlanVersion)
```
56quarters, line 275: Why the use of `apierror` here? It seems odd to use an error specific to the Prometheus HTTP API.

56quarters (approved):
LGTM

charleskorn:
Given the approval, I'm going to merge this so we can start testing it. We can continue discussing the error type and make any changes in a follow-up PR.

pkg/streamingpromql/planning/plan.go
charleskorn: Maybe I've been using `apierror` wrong this whole time - I view it as a way to hint what kind of error it is, so that the API endpoint logic doesn't need to know about every kind of error and how to translate that to an error code / HTTP status code.

Is there something else you had in mind instead?

PR #12787, opened 2025-09-23
Improve query sharding performance and preserve expression position information

charleskorn:
#### What this PR does

This PR improves the performance of rewriting queries for query sharding, and modifies its behaviour to preserve expression position information.

The key change is that `cloneExpr` no longer stringifies and then parses the expression to clone it. Instead, it traverses the AST and creates new `parser.Expr` instances for each node.

This ensures that expression position information is preserved. This is important when annotations are later emitted during evaluation, as these annotations can now show correct expression positions.

Query sharding latency for shardable expressions improves by anywhere from 9% to 45% in our benchmarks, primarily due to a reduction in allocations:

<details><summary>Benchmark results</summary>
<p>

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/mimir/pkg/frontend/querymiddleware
cpu: Apple M1 Pro
                                                                                                                                                                                                     │  before.txt   │              after.txt              │
                                                                                                                                                                                                     │    sec/op     │    sec/op     vs base               │
QueryShardingRewriting/sum_by_(foo)_(histogram_quantile(0.9,_rate(http_request_duration_seconds_bucket[10m])))-10                                                                                       6.782µ ± 35%   6.547µ ± 47%        ~ (p=0.699 n=6)
QueryShardingRewriting/sum_by_(foo,bar)_(min_over_time(bar1{baz="blip"}[1m]))-10                                                                                                                       15.088µ ±  3%   9.995µ ±  1%  -33.75% (p=0.002 n=6)
QueryShardingRewriting/sum(rate(bar1[1m]))_or_rate(bar2[1m])-10                                                                                                                                         6.365µ ± 33%   6.405µ ± 20%        ~ (p=0.589 n=6)
QueryShardingRewriting/sum(rate(bar1[1m]))_or_sum(rate(bar2[1m]))-10                                                                                                                                    22.64µ ±  3%   15.53µ ±  4%  -31.40% (p=0.002 n=6)
QueryShardingRewriting/histogram_quantile(0.5,_sum(rate(cortex_cache_value_size_bytes_bucket[5m]))_by_(le))-10                                                                                          14.70µ ±  4%   10.68µ ±  0%  -27.37% (p=0.002 n=6)
QueryShardingRewriting/sum(_______count(_________count(___________bar1_________)__by_(drive,instance)_______)__by_(instance)_____)-10                                                                   9.560µ ±  3%   8.699µ ±  3%   -9.01% (p=0.002 n=6)
[...]

PR #12812, opened 2025-09-25
Use MQE and upstream test cases to further exercise `cloneExpr`

charleskorn:
#### What this PR does

@colega suggested fuzz testing `cloneExpr` in https://github.com/grafana/mimir/pull/12787#discussion_r2371969778.

Generating valid PromQL expressions through fuzzing is difficult and didn't produce a lot of useful results.

However, we do have a large corpus of PromQL expressions that are used to test MQE - both those we've written ourselves as well as those from upstream, so in this PR I'm exercising `cloneExpr` with these expressions. This should help catch any newly-added features, as these are very likely to be exercised with these test scripts.

This uncovered one case that was not being handled properly in `cloneExpr`, as well as some deficiencies in `requireNoSharedPointers`.

#12787

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/astmapper.go
+++ b/pkg/frontend/querymiddleware/astmapper/astmapper.go
@@ -265,6 +265,9 @@ func cloneExpr(expr parser.Expr) (parser.Expr, error) {
 			Expr: expr,
 		}, nil
 
+	case *parser.DurationExpr:
+		return cloneDurationExpr(e)
+
 	default:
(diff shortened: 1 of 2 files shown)


colega:
> Generating valid PromQL expressions through fuzzing is difficult and didn't produce a lot of useful results.

Interesting, that has not been my experience fuzzing PromQL so far, and it had identified _plenty_ of bugs in promql-related features I've touched in the past.

Issue #12691, opened 2025-09-12
store-gateway panics with "makeslice: len out of range"


narqo:
### What is the bug?

We see occasional panics in store-gateway's `bucketChunkReader`

```
panic: runtime error: makeslice: len out of range
goroutine 147915604 [running]:
github.com/grafana/mimir/pkg/util/pool.(*SlabPool[...]).Get(0x22b3b10?, 0x12884816f240c?)
	/__w/backend-enterprise/backend-enterprise/vendor/github.com/grafana/mimir/pkg/util/pool/pool.go:67 +0x4c
github.com/grafana/mimir/pkg/util/pool.(*SafeSlabPool[...]).Get(0x41337d6500?, 0x4314c48?)
	/__w/backend-enterprise/backend-enterprise/vendor/github.com/grafana/mimir/pkg/util/pool/pool.go:123 +0xc4
github.com/grafana/mimir/pkg/storegateway.(*bucketChunkReader).loadChunks(0x40130728a0, {0x435ec30, 0x4046e08500}, {0x414fed8000, 0x242, 0x408c426f50?}, 0x4, {0x2774196, 0x2c1d5df, {0x0, ...}}, ...)
	/__w/backend-enterprise/backend-enterprise/vendor/github.com/grafana/mimir/pkg/storegateway/bucket_chunk_reader.go:150 +0x4fc
github.com/grafana/mimir/pkg/storegateway.(*bucketChunkReader).load.func3()
	/__w/backend-enterprise/backend-enterprise/vendor/github.com/grafana/mimir/pkg/storegateway/bucket_chunk_reader.go:100 +0x64
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/__w/backend-enterprise/backend-enterprise/vendor/golang.org/x/sync/errgroup/errgroup.go:93 +0x54
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 147915603
	/__w/backend-enterprise/backend-enterprise/vendor/golang.org/x/sync/errgroup/errgroup.go:78 +0x94
```

The relevant piece of code is here https://github.com/grafana/mimir/blob/db608fb47e884044689a8958aae5a0f2dbf0ff15/pkg/storegateway/bucket_chunk_reader.go#L148-L150

### How to reproduce it?

n/a

### What did you think would happen?

n/a

### What was your environment?

Kubernetes, jsonnet, mimir weekly r359

narqo:
[...]

PR #12814, opened 2025-09-25
store-gw: Second attempt at safe guarding chunk parser

gotjosh:

#### What this PR does

This is a third attempt at fixing https://github.com/grafana/mimir/issues/12691 and works on top of https://github.com/grafana/mimir/pull/12729.

We add a check to ensure chunkDataLen can be safely converted from uint64 to int
before conversion. The previous validation only checked against maxChunkDataLen
but didn’t prevent overflow when chunkDataLen + ChunkEncodingSize exceeded
math.MaxInt.

Fixes https://github.com/grafana/mimir/issues/12691

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storegateway/bucket_chunk_reader.go
+++ b/pkg/storegateway/bucket_chunk_reader.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io"
+	"math"
 	"slices"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


pkg/storegateway/bucket_chunk_reader.go
```diff
@@ -158,6 +159,10 @@ func (r *bucketChunkReader) loadChunks(ctx context.Context, res []seriesChunks,
 			return fmt.Errorf("chunk seq %d: parsed data length %d exceeds expected maximum chunk size %d", seq, chunkDataLen, maxChunkDataLen)
 		}
 
+		if chunkDataLen > uint64(math.MaxInt-chunks.ChunkEncodingSize) {
```
dimitarvdimitrov, line 162: `maxChunkDataLen` is 512MB and we know that `chunkDataLen` is less than that. how can it be larger than `2.3058430092E18` bytes (2305 petabytes)

gotjosh: 🤦 ignore me

PR #12818, opened 2025-09-26
store-gateway: use streaming for skipchunks requests

56quarters:
#### What this PR does

WIP

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/storegateway/bucket_store_server_test.go
```diff
@@ -172,7 +167,7 @@ func (s *storeTestServer) Series(ctx context.Context, req *storepb.SeriesRequest
 		}
 	}
 
-	if req.StreamingChunksBatchSize > 0 && !req.SkipChunks {
+	if !req.SkipChunks {
```
charleskorn: [nit] Rather than having `if !req.SkipChunks { A } else { B }`, what if we flipped this around to avoid the negation (ie. `if req.SkipChunks { B } else { A }`)?

pkg/storegateway/bucket.go
```diff
@@ -593,21 +597,20 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storegatewaypb.Stor
...
-	var readers *bucketChunkReaders
-	if !req.SkipChunks {
-		readers = newChunkReaders(chunkReaders)
-	}
+	defer func() {
+		for _, r := range indexReaders {
+			runutil.CloseWithLogOnErr(s.logger, r, "close block index reader")
+		}
+	}()
+	defer func() {
```
charleskorn: [nit] Any reason to have these as separate `defer`s?

pkg/storegateway/bucket.go
```diff
@@ -617,92 +620,76 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storegatewaypb.Stor
...
+		seriesLoadStart      = time.Now()
+		chunksLimiter        = s.chunksLimiterFactory(s.metrics.queriesDropped.WithLabelValues("chunks"))
+		seriesLimiter        = s.seriesLimiterFactory(s.metrics.queriesDropped.WithLabelValues("series"))
 	)
-	for _, b := range blocks {
-		resHints.AddQueriedBlock(b.meta.ULID)
 
-		if b.meta.Compaction.Level == 1 && b.meta.Thanos.Source == block.ReceiveSource && !b.queried.Load() {
-			level.Debug(s.logger).Log("msg", "queried non-compacted block", "blockId", b.meta.ULID, "ooo", b.meta.Compaction.FromOutOfOrder())
-		}
```
charleskorn, line 629: It looks like this logging has been dropped - do we still need it?

pkg/storegateway/bucket.go
```diff
@@ -617,92 +620,76 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storegatewaypb.Stor
...
+	if streamingSeriesCount == 0 {
+		// There are no series to send chunks for.
+		return nil
 	}
 
-	// We create the limiter twice in the case of streaming so that we don't double count the series
-	// and hit the limit prematurely.
-	chunksLimiter := s.chunksLimiterFactory(s.metrics.queriesDropped.WithLabelValues("chunks"))
-	seriesLimiter := s.seriesLimiterFactory(s.metrics.queriesDropped.WithLabelValues("series"))
+	if !req.SkipChunks {
```
charleskorn: Given we don't do anything other than return after this `if` block, what if we invert the condition and return early if `req.SkipChunks` is true?

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
...
 * [CHANGE] Ingester: Remove deprecated `-ingester.stream-chunks-when-using-blocks` CLI flag and `ingester_stream_chunks_when_using_blocks` runtime configuration option. #12615
-* [CHANGE] Querier: Require non-zero values for `-querier.streaming-chunks-per-ingester-buffer-size` and `-querier.streaming-chunks-per-store-gateway-buffer-size` CLI flags and corresponding YAML configurations. This is part of a change that makes streaming required between queriers and ingesters and store-gateways. Streaming has been the default since Mimir 2.14. #12790
+* [CHANGE] Querier: Require non-zero values for `-querier.streaming-chunks-per-ingester-buffer-size` and `-querier.streaming-chunks-per-store-gateway-buffer-size` CLI flags and corresponding YAML configurations. This is part of a change that makes streaming required between queriers and ingesters and store-gateways. Streaming has been the default since Mimir 2.14. #12790 #12818
```
```suggestion
* [CHANGE] Querier: Require non-zero values for `-querier.streaming-chunks-per-ingester-buffer-size` and `-querier.streaming-chunks-per-store-gateway-buffer-size` CLI flags and corresponding YAML configurations. This is part of a change that makes streaming required between queriers, ingesters, and store-gateways. Streaming has been the default since Mimir 2.14. #12790 #12818
```

tacole02 (approved):

pkg/storegateway/bucket.go
56quarters: This was added when @jhalterman and I were trying to reduce the time range queried from ingesters. We've since reverted that experiment so I'm removing the logging here. We can add it back when we attempt to reduce the range queried by ingesters again.

pkg/storegateway/bucket_store_server_test.go
56quarters: Fixed in abfbb24b0109fe62f7f88b8d456fa02347d998bb

pkg/storegateway/bucket.go
56quarters: Fixed in abfbb24b0109fe62f7f88b8d456fa02347d998bb

pkg/storegateway/bucket.go
56quarters: Fixed in abfbb24b0109fe62f7f88b8d456fa02347d998bb

PR #12821, opened 2025-09-26
costattribution.ActiveSeriesTracker: remove duplicate code

colega:
#### What this PR does

This piece of code duplicates the previous lines, and is never executed because the previous branches have their return statement.

No changelog or anything needed, this is a noop.

### DIFF
--- a/pkg/costattribution/active_tracker.go
+++ b/pkg/costattribution/active_tracker.go
@@ -119,27 +119,6 @@ func (at *ActiveSeriesTracker) Increment(lbls labels.Labels, now time.Time, nati
 		}
 		return
 	}
-
-	c, ok = at.observed[string(buf.Bytes())]
-	if ok {
-		c.activeSeries.Inc()
-		if nativeHistogramBucketNum >= 0 {
-			c.nativeHistograms.Inc()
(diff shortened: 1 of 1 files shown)


pkg/costattribution/active_tracker.go
```diff
@@ -110,27 +110,6 @@ func (at *ActiveSeriesTracker) Increment(lbls labels.Labels, now time.Time, nati
...
 
-	if !at.overflowSince.IsZero() {
-		at.observedMtx.RUnlock()
-		at.overflowCounter.activeSeries.Inc()
-		if nativeHistogramBucketNum >= 0 {
-			at.overflowCounter.nativeHistograms.Inc()
-			at.overflowCounter.nativeHistogramBuckets.Add(int64(nativeHistogramBucketNum))
-		}
-		return
-	}
```
colega, line 121: The git diff looks a little bit weird, the lines 102-121 are the ones that are staying, and the lines 123-142 are the ones removes, which are exactly the same ones.

tcard:
I think the extra `at.observedMtx.Lock()` may make this a bit more than a no-op. Do we actually need a write-lock? At a glance, it doesn't look like we do.

colega:
Discussed offline: there's a write at the end.

PR #12823, opened 2025-09-26
docs(nativehistograms): add NHCB

krajorama:
#### What this PR does

Document using native histograms with custom buckets with Mimir.

(Not including Grafana Alloy until I test it.)

Fixes N/A

#### Checklist

- N/A Tests updated.
- [x] Documentation added.

### DIFF
--- a/docs/sources/mimir/send/native-histograms/_custom_buckets.md
+++ b/docs/sources/mimir/send/native-histograms/_custom_buckets.md
@@ -14,7 +14,6 @@ title: Send native histograms with custom buckets to Grafana Mimir
 weight: 2
 ---
 
-
 # Send native histograms with custom buckets to Grafana Mimir
 
 Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with custom buckets, also known as NHCBs, are a sample type in the Prometheus ecosystem that makes it possible to store classic Prometheus [histograms](https://prometheus.io/docs/concepts/metric_types/#histogram) as NHCBs.
--- a/docs/sources/mimir/send/native-histograms/_exponential_buckets.md
+++ b/docs/sources/mimir/send/native-histograms/_exponential_buckets.md
@@ -12,7 +12,6 @@ title: Send native histograms with exponential buckets to Grafana Mimir
 weight: 1
 ---
 
-
 # Send native histograms with exponential buckets to Grafana Mimir
 
 Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with exponential buckets, also known officially as native histograms with standard schemas, is a sample type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.


docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Send native histograms with custom buckets to Grafana Mimir
+
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with custom buckets, also known as NHCB, is a data type in the Prometheus ecosystem that makes it possible to store classic Prometheus [histograms](https://prometheus.io/docs/concepts/metric_types/#histogram) as native histograms with custom buckets.
+
+Native histograms with custom buckets are different from classic Prometheus histograms in a number of ways:
+
+- An instance of a native histogram metric only requires a single time series, because the buckets, sum of observations, and the count of observations are stored in a single data type called `native histogram` rather than in separate time series using the `float` data type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, and `<metric>_count` series. There is only `<metric>` time series.
```
beorn7: Grammar nit:

```suggestion
- An instance of a native histogram metric only requires a single time series, because the buckets, sum of observations, and the count of observations are stored in a single data type called `native histogram` rather than in separate time series using the `float` data type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, and `<metric>_count` series. There is only a `<metric>` time series.
```

Another thought: Maybe use "sample type" instead of "data type".

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+- Querying native histograms via the Prometheus query language (PromQL) uses a different syntax. For details, refer to [Visualize native histograms](https://grafana.com/docs/mimir/<MIMIR_VERSION>/visualize/native-histograms/) and [functions](https://prometheus.io/docs/prometheus/latest/querying/functions/).
+
+For an introduction to native histograms in general, watch the [Native Histograms in Prometheus](https://www.youtube.com/watch?v=AcmABV6NCYk) presentation. For a short introduction to NHCB, watch [Native Histograms With Custom Buckets](https://www.youtube.com/watch?v=2v9DOGq2Mos).
+
+For more information on native histogram flavors and how to query them using PromQL, refer to the [Prometheus Native Histograms specification](https://prometheus.io/docs/specs/native_histograms/).
```
beorn7: Maybe hint that this is a very long and comprehensive document, only read it if you want to know all the details. Otherwise, a reader might start to skim through this and will be very overwhelmed…

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Advantages and disadvantages
+
+There are advantages and disadvantages of using native histograms with custom buckets compared to the classic Prometheus histograms.
+
+### Advantages
+
+- Lower storage costs. Native histograms with custom buckets use a single series instead of multiple and apply a sparse representation that aims to avoid storing buckets that are empty.
+- Storage and query of native histograms with custom buckets are atomic, meaning that either all or none of a native histograms with custom buckets sample is stored or retrieved. This is different from classic histograms that are split into multiple independent time series that might be individually lost or delayed, resulting in corrupted or inconsistent data.
```
beorn7:
```suggestion
- Storage and query of native histograms with custom buckets are atomic, meaning that either all or none of a native histogram with custom buckets sample is stored or retrieved. This is different from classic histograms, which are split into multiple independent time series that might be individually lost or delayed, resulting in corrupted or inconsistent data.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Instrumentation
+
+No change is required in instrumentation.
+
+Currently scrape processes like Prometheus or Grafana Alloy convert classic histograms during scrape to native histograms with custom buckets as there is no support for native histograms with custom buckets in exposition formats.
+
+## Scrape and send native histograms with Prometheus
+
+Use latest Prometheus version.
```
beorn7:
```suggestion
Use the latest Prometheus version.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. To enable scraping native histograms from the application, you need to enable the native histograms feature via a feature flag on the command line:
```
beorn7, line 57: Maybe now that we have laid out our plan, you could already say that this is needed up to version 3.7 (and maybe even mention the scrape config potion for 3.8 and later).

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+   In certain situations, the Protobuf parsing changes the number formatting of
+   the `le` labels of conventional histograms and the `quantile` labels of
+   summaries. Typically, this happens if the scraped target is instrumented with
+   [client_golang](https://github.com/prometheus/client_golang), provided that
+   [promhttp.HandlerOpts.EnableOpenMetrics](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#HandlerOpts)
+   is set to `false`. In such cases, integer label values are represented
+   as `quantile="1"` or `le="2"` omitting the zero fractional.
+   However, the Protobuf parsing changes the representation to always include a fractional (following the OpenMetrics
+   specification), so the examples above become `quantile="1.0"` and `le="2.0"` after
+   ingestion into Prometheus. This changes the identity of the metric from what was originally ingested.
```
beorn7: Do we still need this? Prometheus v3+ normalizes these label values upon ingestion.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,196 @@
...
+
+   For more information, refer to [le and quantile label values](https://prometheus.io/docs/prometheus/latest/migration/#le-and-quantile-label-values) in the Prometheus documentation.
```
beorn7: Which explains what I said above. We only need this if people use v2.x. But these instructions say "Use latest version", so I would argue we can get rid of this whole section.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+menuTitle: Native histograms standard buckets
+title: Send native histograms with exponential buckets to Grafana Mimir
+weight: 1
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Send native histograms with exponential buckets to Grafana Mimir
+
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with exponential buckets, also known officially as native histograms with standard schema, is a data type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
```
beorn7:
```suggestion
Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with exponential buckets, also known officially as native histograms with standard schemas, is a data type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+
+- Exponential bucket boundaries are calculated by a formula that depends on the scale (resolution) of the native histogram, and are not user defined. The calculation produces exponentially increasing bucket boundaries. For details, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
+- Exponential bucket boundaries might change (widen) dynamically if the observations result in too many buckets. For details, refer to [Limit the number of buckets](#limit-the-number-of-buckets).
+- Exponential bucket counters only count observations inside the bucket boundaries, whereas the classic histogram buckets only have an upper bound called `le` and count all observations in the bucket and all lower buckets (cumulative).
+- An instance of a native histogram metric only requires a single time series, because the buckets, sum of observations, and the count of observations are stored in a single data type called `native histogram` rather than in separate time series using the `float` data type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, and `<metric>_count` series. There is only `<metric>` time series.
```
beorn7: As above, maybe use "sample type".

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+- Querying native histograms via the Prometheus query language (PromQL) uses a different syntax. For details, refer to [Visualize native histograms](https://grafana.com/docs/mimir/<MIMIR_VERSION>/visualize/native-histograms/) and [functions](https://prometheus.io/docs/prometheus/latest/querying/functions/).
+
+For an introduction to native histograms, watch the [Native Histograms in Prometheus](https://www.youtube.com/watch?v=AcmABV6NCYk) presentation.
+
+For more information on native histogram flavors and how to query them using PromQL, refer to the [Prometheus Native Histograms specification](https://prometheus.io/docs/specs/native_histograms/).
```
beorn7: As above, maybe mention that this is very comprehensive reference documentation.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+
+There are advantages and disadvantages of using native histograms with exponential buckets compared to the classic Prometheus histograms. For more information and a real example, refer to the [Prometheus Native Histograms in Production](https://www.youtube.com/watch?v=TgINvIK9SYc&t=127s) video.
+
+### Advantages
+
+- Simpler instrumentation: you do not need to think about bucket boundaries because they are created automatically.
+- Better resolution in practice: custom bucket layouts are usually not high resolution.
+- Native histograms with exponential buckets are compatible with each other: they have an automatic layout, which makes them easy to combine.
+  {{< admonition type="note" >}}
+  The operation might scale down an operand to lower resolution to match the other operand.
```
beorn7:
```suggestion
  The operation might scale down an operand to a lower resolution to match the other operand.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+  {{< /admonition >}}
+
+### Disadvantages
+
+- Observations might be distributed in a way that is not a good fit for the exponential bucket schema, such as sound pressure measured in decibels, which are already logarithmic.
+- If converting from an externally represented histogram with specific bucket boundaries, there is generally no precise match with the exponential bucket boundaries of the native histogram, and in which case you need to use interpolation.
```
beorn7:
```suggestion
- If converting from an externally represented histogram with specific bucket boundaries, there is generally no precise match with the exponential bucket boundaries of the native histogram, in which case you need to use interpolation.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+
+The value of `NativeHistogramMaxBucketNumber`/`nativeMaxNumberOfBuckets` limits the number of exponential buckets produced by the observations. This can be especially useful if the receiver side is limiting the number of buckets that can be sent. For more information about the bucket limit refer to [Limit the number of buckets](#limit-the-number-of-buckets).
+
+The duration in `NativeHistogramMinResetDuration`/`nativeResetDuration` will prohibit automatic counter resets inside that period. Counter resets are related to the bucket limit, for more information refer to [Limit the number of buckets](#limit-the-number-of-buckets).
+
+## Scrape and send native histograms with Prometheus
+
+Use Prometheus version 2.47 or later.
+
+1. To enable scraping native histograms from the application, you need to enable the native histograms feature via a feature flag on the command line:
```
beorn7, line 107: As above, maybe mention v3.8 changes.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+   [client_golang](https://github.com/prometheus/client_golang), provided that
+   [promhttp.HandlerOpts.EnableOpenMetrics](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#HandlerOpts)
+   is set to `false`. In such cases, integer label values are represented
+   as `quantile="1"` or `le="2"` omitting the zero fractional.
+   However, the Protobuf parsing changes the representation to always include a fractional (following the OpenMetrics
+   specification), so the examples above become `quantile="1.0"` and `le="2.0"` after
+   ingestion into Prometheus. This changes the identity of the metric from what was originally ingested.
+
+   For more information, refer to [le and quantile label values](https://prometheus.io/docs/prometheus/latest/migration/#le-and-quantile-label-values) in the Prometheus documentation.
+   {{< /admonition >}}
```
beorn7: Let's say above that you must use at least Prom v3 and then let's get rid of this whole section.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,379 @@
...
+   [client_golang](https://github.com/prometheus/client_golang), provided that
+   [promhttp.HandlerOpts.EnableOpenMetrics](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#HandlerOpts)
+   is set to `false`. In such cases, integer label values are represented
+   as `quantile="1"` or `le="2"` omitting the zero fractional.
+   However, the Protobuf parsing changes the representation to always include a fractional (following the OpenMetrics
+   specification), so the examples above become `quantile="1.0"` and `le="2.0"` after
+   ingestion into Prometheus. This changes the identity of the metric from what was originally ingested.
+
+   For more information, refer to [Feature Flags Native Histograms](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) in the Prometheus documentation.
+   {{< /admonition >}}
```
beorn7: Hasn't Alloy followed Prom v3 in this regard?

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: We'll just have to update this doc when 3.8 comes around, because we never document forward.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
krajorama: Same as above, we don't document forward.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: I let Claude come up with something, ptal

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
krajorama: I let Claude come up with something, ptal

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: removed

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
krajorama: The documentation only mentions it in an unrelased 1.11 version: https://grafana.com/docs/alloy/latest/release-notes/#v111-unreleased . So not yet apparently.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
beorn7: OK, if that's the way… :shrug:

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
beorn7: Oooof, I hope they plan this very soon. The number formatting has haunted us for way too long.

Maybe you should still say that this behavior is only relevant for alloy and only up to v1.10 and that users of Prom v3 always see the "new" formatting?

docs/sources/mimir/send/native-histograms/_index.md
```diff
@@ -15,360 +15,11 @@ weight: 100
 
 # Send native histograms to Mimir
 
-Prometheus native histograms is a data type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) is a data type in the Prometheus ecosystem that offers two use cases:
 
-Native histograms are different from classic Prometheus histograms in a number of ways:
+1. Make it possible to produce, store, and query a high-resolution histogram of observations. Use native histograms with exponential buckets in your instrumentation for this purpose.
```
tacole02:
```suggestion
1. Produce, store, and query a high-resolution histogram of observations. Use native histograms with exponential buckets in your instrumentation for this purpose.
```

docs/sources/mimir/send/native-histograms/_index.md
```diff
@@ -15,360 +15,11 @@ weight: 100
...
+1. Make it possible to store classic histograms in a more resilient and cost effective way, without changes to instrumentation. Use native histograms with custom buckets for this purpose.
```
tacole02:
```suggestion
1. Store classic histograms in a more resilient and cost effective way, without changes to instrumentation. Use native histograms with custom buckets for this purpose.
```

docs/sources/mimir/send/native-histograms/_index.md
```diff
@@ -15,360 +15,11 @@ weight: 100
 
 # Send native histograms to Mimir
 
-Prometheus native histograms is a data type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) is a data type in the Prometheus ecosystem that offers two use cases:
```
tacole02:
```suggestion
Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) is a data type in the Prometheus ecosystem that enables the following use cases:
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+   For example, use this scrape protocol setting to avoid scraping native histograms with exponential schema:
+
+   ```yaml
+   scrape_configs:
+     - job_name: myapp
+       scrape_protocols:
+         [OpenMetricsText1.0.0, OpenMetricsText0.0.1, PrometheusText0.0.4]
+   ```
+
+1. To enable converting classic histograms into NHCB, you need to set `convert_histograms_to_nhcb` to `true` in your scrape jobs. This setting will have no effect for histograms that already have a native histogram defined, such as with [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets).
```
zenador: Isn't it `convert_classic_histograms_to_nhcb`?

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
+---
+description: Learn how to store classic histograms more efficiently in Grafana Mimir.
```
tacole02, line 2:
```suggestion
description: Learn how to store classic histograms in Grafana Mimir.
```

docs/sources/mimir/send/native-histograms/_index.md
tacole02:
```suggestion
Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) are a data type in the Prometheus ecosystem that offers two use cases:
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+keywords:
+  - send metrics
+  - native histogram
+  - custom buckets
+  - NHCB
+  - prometheus
+  - grafana alloy
+  - instrumentation
+  - TCO
+menuTitle: Native histograms custom buckets
```
tacole02:
```suggestion
menuTitle: Native histograms with custom buckets
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+title: Send native histograms with custom buckets to Grafana Mimir
+weight: 2
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Send native histograms with custom buckets to Grafana Mimir
+
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with custom buckets, also known as NHCB, is a sample type in the Prometheus ecosystem that makes it possible to store classic Prometheus [histograms](https://prometheus.io/docs/concepts/metric_types/#histogram) as NHCB.
```
tacole02:
```suggestion
Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with custom buckets, also known as NHCB, are a sample type in the Prometheus ecosystem that makes it possible to store classic Prometheus [histograms](https://prometheus.io/docs/concepts/metric_types/#histogram) as NHCB.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+NHCB are different from classic Prometheus histograms in a number of ways:
+
+- An instance of a native histogram metric only requires a single time series, because the buckets, sum of observations, and the count of observations are stored in a single sample type called `native histogram` rather than in separate time series using the `float` sample type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, and `<metric>_count` series. There is only a `<metric>` time series.
```
tacole02:
```suggestion
- An instance of a native histogram metric only requires a single time series. This is because the buckets, sum of observations, and the count of observations are stored in a single sample type called `native histogram` rather than in separate time series using the `float` sample type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, or `<metric>_count` series. There is only a `<metric>` time series.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+- Querying native histograms via the Prometheus query language (PromQL) uses a different syntax. For details, refer to [Visualize native histograms](https://grafana.com/docs/mimir/<MIMIR_VERSION>/visualize/native-histograms/) and [functions](https://prometheus.io/docs/prometheus/latest/querying/functions/).
+
+For an introduction to native histograms in general, watch the [Native Histograms in Prometheus](https://www.youtube.com/watch?v=AcmABV6NCYk) presentation. For a short introduction to NHCB, watch [Native Histograms With Custom Buckets](https://www.youtube.com/watch?v=2v9DOGq2Mos).
+
+This document provides a practical implementation guide for using NHCB in Grafana Mimir, but it's not as comprehensive as the official [specification](https://prometheus.io/docs/specs/native_histograms/) in the Prometheus documentation.
```
tacole02:
```suggestion
This document provides a practical implementation guide for using NHCB in Grafana Mimir. For comprehensive guidance, refer to the official [specification](https://prometheus.io/docs/specs/native_histograms/) in the Prometheus documentation.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Advantages and disadvantages
+
+There are advantages and disadvantages of using NHCB compared to the classic Prometheus histograms.
```
tacole02:
```suggestion
There are advantages and disadvantages of using NHCB compared to classic Prometheus histograms.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Advantages
+
+- Lower storage costs. NHCB use a single series instead of multiple and apply a sparse representation that aims to avoid storing buckets that are empty.
```
tacole02:
```suggestion
- Lower storage costs. NHCB use a single series and applies a sparse representation that avoids storing empty buckets.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+- Storage and query of NHCB are atomic, meaning that either all or none of a NHCB sample is stored or retrieved. This is different from classic histograms, which are split into multiple independent time series that might be individually lost or delayed, resulting in corrupted or inconsistent data.
+- It is possible to migrate to NHCB without modifying the instrumentation.
+- Easy to migrate to [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) later as queries and visualizations would be already migrated.
```
tacole02:
```suggestion
- It's easy to migrate to [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets), as queries and visualizations are already migrated.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Advantages and disadvantages
+
+There are advantages and disadvantages of using NHCB compared to the classic Prometheus histograms.
+
+### Advantages
+
+- Lower storage costs. NHCB use a single series instead of multiple and apply a sparse representation that aims to avoid storing buckets that are empty.
+- Storage and query of NHCB are atomic, meaning that either all or none of a NHCB sample is stored or retrieved. This is different from classic histograms, which are split into multiple independent time series that might be individually lost or delayed, resulting in corrupted or inconsistent data.
+- It is possible to migrate to NHCB without modifying the instrumentation.
```
tacole02:
```suggestion
- It's possible to migrate to NHCB without modifying the instrumentation.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+- Easy to migrate to [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) later as queries and visualizations would be already migrated.
+
+### Disadvantages
+
+- Queries and visualizations need to be adopted to enable the advantages.
```
tacole02:
```suggestion
- You need to adopt certaim queries and visualizations to enable the advantages.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+- Depends on experimental Prometheus Remote-Write 2.0.
```
tacole02:
```suggestion
- NHCB depends on experimental Prometheus Remote-Write 2.0.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+- Some overhead to the scrape process. This depends on the scrape protocol (Protobuf is the best) and the ratio of classic histograms to other kind of metrics.
```
tacole02:
```suggestion
- There's some overhead to the scrape process. This depends on the scrape protocol and the ratio of classic histograms to other kind of metrics.
```

tacole02: I think calling out Protobuf here feels out of place.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Instrumentation
+
+No change is required in instrumentation.
+
+Currently scrape processes like Prometheus or Grafana Alloy convert classic histograms during scrape to NHCB as there is no support for NHCB in exposition formats.
```
tacole02:
```suggestion
Currently scrape processes like Prometheus or Grafana Alloy convert classic histograms during scrape to NHCB, as there is no support for NHCB in exposition formats.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+   This flag makes Prometheus detect and scrape [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) over the `PrometheusProto` scrape protocol and ignore the classic histogram version of metrics that have native histograms defined as well.
+
+   {{< admonition type="note" >}}
+   <!-- Issue: https://github.com/prometheus/prometheus/issues/11265 -->
+
+   Native histograms don't have a textual presentation on the application's `/metrics` endpoint. Therefore, Prometheus negotiates a Protobuf protocol transfer in these cases.
+   {{< /admonition >}}
+
+1. Optionally, to disable native histograms with exponential buckets and only use NHCB, you need to set scrape protocols that don't carry native histograms with exponential buckets. This is not recommended as it does not save on migration time or effort, but looses the advantages of exponential buckets and has higher overhead than using `PrometheusProto`.
```
tacole02:
```suggestion
1. Optionally, to disable native histograms with exponential buckets and only use NHCB, you need to set scrape protocols that don't carry native histograms with exponential buckets. This is not a best practice, as it does not save on migration time or effort, yet loses the advantages of exponential buckets and has higher overhead than using `PrometheusProto`.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+   For example, use this scrape protocol setting to avoid scraping native histograms with exponential schema:
```
tacole02:
```suggestion
   For example, use this scrape protocol setting to avoid scraping native histograms with the exponential schema:
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
tacole02:
```suggestion
1. To enable converting classic histograms into NHCB, you need to set `convert_histograms_to_nhcb` to `true` in your scrape jobs. This setting has no effect for histograms that already have a native histogram defined, such as with [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets).
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+   ```yaml
+   scrape_configs:
+     - job_name: myapp
+       convert_histograms_to_nhcb: true
+   ```
+
+1. To keep scraping the classic histogram version of native histogram metrics, you need to set `always_scrape_classic_histograms` to `true` in your scrape jobs.
+
+   {{< admonition type="note" >}}
+   In Prometheus versions earlier than 3.0, the `always_scrape_classic_histograms` setting is called `scrape_classic_histograms`. Use the setting name that corresponds to the version of Prometheus you're using.
```
tacole02: Instead of "earlier than 3.0", it more clear to say "versions x.x and earlier".

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+   For example, to get both classic and native histograms, use the following configuration:
+
+   ```yaml
+   scrape_configs:
+     - job_name: myapp
+       convert_histograms_to_nhcb: true
+       always_scrape_classic_histograms: true
+   ```
+
+1. To be able to send native histograms to a Prometheus remote-write 2.0 compatible receiver, for example Grafana Cloud Metrics, Mimir, etc, set `send_native_histograms` to `true` and `protobuf_message` to `io.prometheus.write.v2.Request` in the remote-write configuration, for example:
```
tacole02:
```suggestion
1. To be able to send native histograms to a Prometheus remote-write 2.0 compatible receiver, for example, Grafana Cloud Metrics or Mimir, set `send_native_histograms` to `true` and `protobuf_message` to `io.prometheus.write.v2.Request` in the remote-write configuration. For example:
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+   ```yaml
+   remote_write:
+     - url: http://.../api/prom/push
+       send_native_histograms: true
+       protobuf_message: "io.prometheus.write.v2.Request"
+   ```
+
+   {{< admonition type="note" >}}
+   Prometheus remote-write 2.0 is experimental, for more information see the [Prometheus Remote-Write 2.0 specification](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/).
```
tacole02:
```suggestion
   Prometheus remote-write 2.0 is experimental. For more information, refer to the [Prometheus Remote-Write 2.0 specification](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/).
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+   {{< /admonition >}}
+
+   {{< admonition type="note" >}}
+   Prometheus remote-write 2.0 doesn't fully support [metadata](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metadata) currently.
```
tacole02:
```suggestion
   Prometheus remote-write 2.0 doesn't fully support [metadata](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metadata).
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+To ease the migration process, you can keep the custom bucket definition of a classic histogram and add native histogram buckets at the same time.
+
+1. Let Prometheus scrape both classic and NHCB.
+1. Send native histograms to remote write, along with the existing classic histograms.
+1. Modify dashboards to use the native histogram metrics. Refer to [Visualize native histograms](https://grafana.com/docs/mimir/<MIMIR_VERSION>/visualize/native-histograms/) for more information.
+
+   Use one of the following strategies to update dashboards.
+
+   - (Recommended) Add new dashboards with the new native histogram queries. This solution requires looking at different dashboards for data before and after the migration, until data before the migration is removed due to passing its retention time. You can publish the new dashboard when sufficient time has passed to serve users with the new data.
```
tacole02:
```suggestion
   - (Recommended) Add new dashboards with the new native histogram queries. This solution requires looking at different dashboards for data before and after the migration until data before the migration is removed due to passing its retention time. You can publish the new dashboard when sufficient time has passed to serve users with the new data.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+     ```
+
+     Where `classic_query` is the original query and `native_query` is the same query using native histogram query syntax.
+
+     {{< admonition type="warning" >}}
+     Using the PromQL operator `or` can lead to unexpected results. For example, if a query uses a range of seven days, such as `sum(rate(http_request_duration_seconds[7d]))`, then this query returns a value as soon as there are two native histograms samples present before the end time specified in the query. In this case, the seven day rate is calculated from a couple of minutes, rather than seven days, worth of data. This results in an inaccuracy in the graph around the time you started scraping native histograms.
+     {{< /admonition >}}
+
+1. Start adding _new_ recording rules and alerts to use native histograms. Do not remove the old recording rules and alerts at this time.
+1. It is important to keep scraping both classic and native histograms for at least the period of the longest range in your recording rules and alerts, plus one day. This is the minimum amount of time, but it's recommended to keep scraping both sample types until the new rules and alerts can be verified.
```
tacole02:
```suggestion
1. It is important to keep scraping both classic and native histograms for at least the period of the longest range in your recording rules and alerts plus one day. This is the minimum amount of time, but it's recommended to keep scraping both sample types until the new rules and alerts can be verified.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+   For example, if you have an alert that calculates the rate of requests, such as `sum(rate(http_request_duration_seconds[7d]))`, this query looks at the data from the last seven days plus the Prometheus [lookback period](https://prometheus.io/docs/prometheus/latest/querying/basics/#staleness). When you start sending native histograms, the data isn't there for the entire seven days, and therefore, the results might be unreliable for alerting.
+
+1. After configuring native histogram collection, choose one of the following ways to stop collecting classic histograms.
```
tacole02:
```suggestion
1. After configuring the collection of native histograms, choose one of the following ways to stop collecting classic histograms.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+
+   - Drop the classic histogram series with [Prometheus relabeling](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) or [Grafana Alloy prometheus.relabel](https://grafana.com/docs/alloy/<ALLOY_VERSION>/reference/components/prometheus/prometheus.relabel) at the time of scraping.
+   - Stop scraping the classic histogram version of metrics. This option applies to all metrics of a scrape target.
+
+1. Clean up recording rules and alerts by deleting the classic histogram version of the rule or alert.
+
+## Limit the number of buckets
+
+There is no way to automatically reduce the resolution or merge custom buckets like with [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets#limit-the-number-buckets). Set bucket limits greater than the largest expected bucket count to avoid having NHCB rejected.
```
tacole02:
```suggestion
There is no way to automatically reduce the resolution or merge custom buckets like there is with [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets#limit-the-number-buckets). Set bucket limits greater than the largest expected bucket count to avoid having NHCB rejected.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
tacole02: This question applies through this entire topic: should NHCB be plural, as in NHSBs?
Prometheus native histograms with custom buckets seems plural.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
+
+# Send native histograms with exponential buckets to Grafana Mimir
+
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with exponential buckets, also known officially as native histograms with standard schemas, is a sample type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
+
+Native histograms with exponential buckets are different from classic Prometheus histograms in a number of ways:
+
+- Exponential bucket boundaries are calculated by a formula that depends on the scale (resolution) of the native histogram, and are not user defined. The calculation produces exponentially increasing bucket boundaries. For details, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```
tacole02:
```suggestion
- Exponential bucket boundaries are calculated by a formula that depends on the scale, or resolution of the native histogram, and are not user-defined. The calculation produces exponentially increasing bucket boundaries. For details, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+- Exponential bucket boundaries might change (widen) dynamically if the observations result in too many buckets. For details, refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```
tacole02:
```suggestion
- Exponential bucket boundaries might change dynamically if the observations result in too many buckets. For details, refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+- Exponential bucket counters only count observations inside the bucket boundaries, whereas the classic histogram buckets only have an upper bound called `le` and count all observations in the bucket and all lower buckets (cumulative).
```
tacole02:
```suggestion
- Exponential bucket counters only count observations inside the bucket boundaries, whereas the classic histogram buckets only have an upper bound called `le` and count all observations in the bucket and all lower buckets cumulatively.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+- An instance of a native histogram metric only requires a single time series, because the buckets, sum of observations, and the count of observations are stored in a single sample type called `native histogram` rather than in separate time series using the `float` sample type. Thus, there are no `<metric>_bucket`, `<metric>_sum`, and `<metric>_count` series. There is only a `<metric>` time series.
+- Querying native histograms via the Prometheus query language (PromQL) uses a different syntax. For details, refer to [Visualize native histograms](https://grafana.com/docs/mimir/<MIMIR_VERSION>/visualize/native-histograms/) and [functions](https://prometheus.io/docs/prometheus/latest/querying/functions/).
+
+For an introduction to native histograms, watch the [Native Histograms in Prometheus](https://www.youtube.com/watch?v=AcmABV6NCYk) presentation.
+
+This document provides a practical implementation guide for using native histograms with exponential schemas in Grafana Mimir, but it's not as comprehensive as the official [specification](https://prometheus.io/docs/specs/native_histograms/) in the Prometheus documentation.
```
tacole02:
```suggestion
This document provides a practical implementation guide for using native histograms with exponential schemas in Grafana Mimir. For comprehensive guidance, refer to the official [specification](https://prometheus.io/docs/specs/native_histograms/) in the Prometheus documentation.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Advantages and disadvantages
+
+There are advantages and disadvantages of using native histograms with exponential buckets compared to the classic Prometheus histograms. For more information and a real example, refer to the [Prometheus Native Histograms in Production](https://www.youtube.com/watch?v=TgINvIK9SYc&t=127s) video.
```
tacole02:
```suggestion
There are advantages and disadvantages of using native histograms with exponential buckets compared to the classic Prometheus histograms. For more information, refer to the [Prometheus Native Histograms in Production](https://www.youtube.com/watch?v=TgINvIK9SYc&t=127s) video.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Advantages
+
+- Simpler instrumentation: you do not need to think about bucket boundaries because they are created automatically.
+- Better resolution in practice: custom bucket layouts are usually not high resolution.
+- Native histograms with exponential buckets are compatible with each other: they have an automatic layout, which makes them easy to combine.
```
tacole02:
```suggestion
- Simpler instrumentation: You don't need to think about bucket boundaries because they are created automatically.
- Better resolution in practice: Custom bucket layouts are usually not high resolution.
- Increased compatibility: Native histograms with exponential buckets are compatible with each other. They have an automatic layout, which makes them easy to combine.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+  {{< admonition type="note" >}}
+  The operation might scale down an operand to a lower resolution to match the other operand.
```
tacole02:
```suggestion
  This operation might scale down an operand to a lower resolution to match the other operand.
```

tacole02: I'm not sure this note is in the right place here- this seems like just a general section to call out advantages and not to go into specific details.

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+The preceding problems are mitigated by high resolution, which native histograms with exponential buckets can provide at a much lower resource cost compared to classic histograms.
+
+## Instrument application with Prometheus client libraries
+
+The following examples have some reasonable defaults to define a new native histogram with exponential buckets metric. The examples use the [Go client library](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Histogram) version 1.16 and the [Java client library](https://prometheus.github.io/client_java/api/io/prometheus/metrics/core/metrics/Histogram.Builder.html) 1.0.
+
+{{< admonition type="note" >}}
+Native histogram options can be added to existing classic histograms to get both the classic and native histogram at the same time. Refer to [Migrate from classic histograms](#migrate-from-classic-histograms).
```
tacole02:
```suggestion
You can add native histogram options to existing classic histograms to get both the classic and native histogram at the same time. Refer to [Migrate from classic histograms](#migrate-from-classic-histograms).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+{{< /admonition >}}
+
+{{< admonition type="note" >}}
+Native histograms with exponential schema take precedence over [native histograms with custom buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_custom_buckets), which means that if a histogram is scraped as native histograms with exponential bucket, then no native histograms with custom buckets can be generated for the same histogram.
```
tacole02:
```suggestion
Native histograms with exponential schema take precedence over [native histograms with custom buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_custom_buckets), which means that if a histogram is scraped as a native histogram with exponential buckets, then no native histograms with custom buckets can be generated for the same histogram.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+     .nativeOnly()
+     .nativeInitialSchema(3)
+     .nativeMaxNumberOfBuckets(100)
+     .nativeResetDuration(1, TimeUnit.HOURS)
+     .register();
+```
+
+{{< /code >}}
+
+In Go, the `NativeHistogramBucketFactor` option sets an upper limit of the relative growth from one exponential bucket to the next. The value 1.1 means that an exponential bucket is at most 10% wider than the next smaller exponential bucket. The currently supported values range from `1.0027` or 0.27% up to 65536 or 655%. For more detailed explanation, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```
tacole02:
```suggestion
In Go, the `NativeHistogramBucketFactor` option sets an upper limit of the relative growth from one exponential bucket to the next. The value 1.1 means that an exponential bucket is at most 10% wider than the next smaller exponential bucket. The currently supported values range from `1.0027` or 0.27% up to 65536 or 655%. For a more detailed explanation, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+..., (76.1, 83], (83, 91], (91, 99], ...
+
+..., (512, 558], (558, 608], (608, 663], ...
+
+In Java `.nativeInitialSchema` using schema value `3` results in the same exponential bucket boundaries. For more information about the schema supported in Java, consult the documentation for [nativeInitialSchema](<https://prometheus.github.io/client_java/api/io/prometheus/metrics/core/metrics/Histogram.Builder.html#nativeInitialSchema(int)>).
+
+The value of `NativeHistogramMaxBucketNumber`/`nativeMaxNumberOfBuckets` limits the number of exponential buckets produced by the observations. This can be especially useful if the receiver side is limiting the number of buckets that can be sent. For more information about the bucket limit refer to [Limit the number of buckets](#limit-the-number-of-buckets).
+
+The duration in `NativeHistogramMinResetDuration`/`nativeResetDuration` will prohibit automatic counter resets inside that period. Counter resets are related to the bucket limit, for more information refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```
tacole02:
```suggestion
The duration in `NativeHistogramMinResetDuration`/`nativeResetDuration` prohibits automatic counter resets inside that period. Counter resets are related to the bucket limit, for more information refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+   For example, to get both classic and native histograms, use the following configuration:
+
+   ```yaml
+   scrape_configs:
+     - job_name: myapp
+       always_scrape_classic_histograms: true
+   ```
+
+1. To be able to send native histograms to a Prometheus remote-write compatible receiver, for example Grafana Cloud Metrics, Grafana Mimir, etc, set `send_native_histograms` to `true` in the remote-write configuration, for example:
```
tacole02:
```suggestion
1. To send native histograms to a Prometheus remote-write compatible receiver, for example Grafana Cloud Metrics or Grafana Mimir, set `send_native_histograms` to `true` in the remote-write configuration. For example:
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+1. To scrape native histograms, set the `scrape_protocols` argument in the `prometheus.scrape` component to specify `PrometheusProto` as the first protocol to negotiate.
+
+   ```
+    scrape_protocols = ["PrometheusProto", "OpenMetricsText1.0.0", "OpenMetricsText0.0.1", "PrometheusText0.0.4"]
+   ```
+
+1. To scrape classic histograms in addition to native histograms, set `always_scrape_classic_histograms` to `true`.
+
+   {{< admonition type="note" >}}
+   In Prometheus versions earlier than 3.0, the `always_scrape_classic_histograms` setting is called `scrape_classic_histograms`. Use the setting name that corresponds to the version of Prometheus you're using.
```
tacole02: Same comment about the clarity of versions. We should say "In Prometheus versions x.x and earlier . . . "

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+   ```
+
+## Migrate from classic histograms
+
+To ease the migration process, you can keep the custom bucket definition of a classic histogram and add native histogram exponential buckets at the same time.
+
+1. Add the native histogram exponential buckets definition to an existing histogram in the instrumentation.
+1. If the existing histogram doesn't have buckets defined, add the default buckets to keep the classic histogram.
+
+   Code examples with both classic and native histogram defined for the same metric:
```
tacole02:
```suggestion
   The following code examples have both classic and native histograms defined for the same metric:
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+      .nativeInitialSchema(3)
+      .nativeMaxNumberOfBuckets(100)
+      .nativeResetDuration(1, TimeUnit.HOURS)
+      .register();
+   ```
+
+   {{< /code >}}
+
+1. Let Prometheus or Grafana Alloy scrape both classic and native histograms with exponential buckets for metrics that have both defined.
+1. Send native histograms to remote write, along with the existing classic histograms.
```
tacole02:
```suggestion
1. Send native histograms to remote write along with the existing classic histograms.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+     <native_query> or <classic_query>
+     ```
+
+     Where `classic_query` is the original query and `native_query` is the same query using native histogram query syntax.
+
+     {{< admonition type="warning" >}}
+     Using the PromQL operator `or` can lead to unexpected results. For example, if a query uses a range of seven days, such as `sum(rate(http_request_duration_seconds[7d]))`, then this query returns a value as soon as there are two native histograms samples present before the end time specified in the query. In this case, the seven day rate is calculated from a couple of minutes, rather than seven days, worth of data. This results in an inaccuracy in the graph around the time you started scraping native histograms.
+     {{< /admonition >}}
+
+1. Start adding _new_ recording rules and alerts to use native histograms. Do not remove the old recording rules and alerts at this time.
```
tacole02:
```suggestion
1. Start adding new recording rules and alerts to use native histograms. Do not remove the existing recording rules and alerts at this time.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+     <native_query> or <classic_query>
+     ```
+
+     Where `classic_query` is the original query and `native_query` is the same query using native histogram query syntax.
+
+     {{< admonition type="warning" >}}
+     Using the PromQL operator `or` can lead to unexpected results. For example, if a query uses a range of seven days, such as `sum(rate(http_request_duration_seconds[7d]))`, then this query returns a value as soon as there are two native histograms samples present before the end time specified in the query. In this case, the seven day rate is calculated from a couple of minutes, rather than seven days, worth of data. This results in an inaccuracy in the graph around the time you started scraping native histograms.
+     {{< /admonition >}}
+
+1. Start adding _new_ recording rules and alerts to use native histograms. Do not remove the old recording rules and alerts at this time.
```
tacole02:
```suggestion
1. Start adding new recording rules and alerts to use native histograms. Do not remove the existing recording rules and alerts at this time.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+Native histograms that have a higher bucket count than the limit are converted to a native histogram with a lower resolution by merging buckets to reduce the number of buckets. In some rare cases, if the buckets are too widely spread out, merging them isn't possible and the native histogram is rejected.
+
+### Limit the number of buckets in application instrumentation
+
+The instrumentation libraries of Prometheus have automation to keep the number of exponential buckets down, provided that the maximum bucket number option is used, such as `NativeHistogramMaxBucketNumber` in Go.
+
+After the set maximum is exceeded, the following strategy is enacted:
+
+1. First, if the last reset (or the creation) of the histogram is at least the minimum reset duration ago, then the whole histogram is reset to its initial state (including classic buckets). This only works if the minimum reset duration was set (`NativeHistogramMinResetDuration` in Go).
```
tacole02:
```suggestion
1. First, if the last reset, or the creation, of the histogram is at least the minimum reset duration, then the whole histogram is reset to its initial state, including classic buckets. This only works if the minimum reset duration was set (`NativeHistogramMinResetDuration` in Go).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+1. If less time has passed, or if the minimum reset duration is zero, no reset is performed. Instead, the zero threshold is increased sufficiently to reduce the number of exponential buckets to or below the maximum bucket number, but not to more than the maximum zero threshold (`NativeHistogramMaxZeroThreshold` in Go). Thus, if the threshold is at or above the maximum threshold already nothing happens at this step.
```
tacole02:
```suggestion
1. If less time has passed, or if the minimum reset duration is zero, no reset is performed. Instead, the zero threshold is increased sufficiently to reduce the number of exponential buckets to or below the maximum bucket number, but not to more than the maximum zero threshold (`NativeHistogramMaxZeroThreshold` in Go). Thus, if the threshold is at or above the maximum threshold already, nothing happens at this step.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+1. After that, if the number of exponential buckets still exceeds maximum bucket number, the resolution of the histogram is reduced by doubling the width of all the exponential buckets (up to a growth factor between one bucket to the next of 2^(2^4) = 65536, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation)).
```
tacole02:
```suggestion
1. After that, if the number of exponential buckets still exceeds the maximum bucket number, the resolution of the histogram is reduced by doubling the width of all the exponential buckets up to a growth factor between one bucket to the next of 2^(2^4) = 65536. Refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation)).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,364 @@
...
+
+1. Any increased zero threshold or reduced resolution is reset back to their original values once the minimum reset duration has passed (since the last reset or the creation of the histogram).
```
tacole02:
```suggestion
1. Any increased zero threshold or reduced resolution is reset back to their original values once the minimum reset duration has passed since the last reset or the creation of the histogram.
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: We need some kind of qualifier here as people know how to store classic histograms as classic histograms already and this document is about storing classic histograms as a kind of native histograms.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: yes, thank you!

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
krajorama: that's fair, I'll remove

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
krajorama: 1.11 is out , so I'm going to double check this and hopefully just remove this text.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: Beorn's spec uses NHCBs for the type of histogram and "an NHCB" when talking about a sample, so I'll go ahead the read through again with this in mind.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: Maybe
```suggestion
description: Learn how to store classic histograms as native histograms in Grafana Mimir.
```

tacole02: Thanks for clarifying this. In this case, I think what you have currently, ` more efficiently` makes sense.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+  - prometheus
+  - grafana alloy
+  - instrumentation
+  - TCO
+menuTitle: Native histograms with custom buckets
+title: Send native histograms with custom buckets to Grafana Mimir
+weight: 2
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
```
tacole02: You can remove this note. I have https://github.com/grafana/mimir/pull/12992 open to remove this from the rest.

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+## Advantages and disadvantages
+
+There are advantages and disadvantages of using NHCBs compared to classic Prometheus histograms.
+
+### Advantages
+
+- Lower storage costs. NHCBs use a single series and applies a sparse representation that avoids storing empty buckets.
+- Storage and query of NHCBs are atomic, meaning that either all or none of a NHCB sample is stored or retrieved. This is different from classic histograms, which are split into multiple independent time series that might be individually lost or delayed, resulting in corrupted or inconsistent data.
+- It's possible to migrate to NHCBs without modifying the instrumentation.
+- It's easy to migrate to [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) once NHCBs are in use, as queries and visualizations are already migrated.
```
tacole02:
```suggestion
- It's easy to migrate to [native histograms with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) once NHCBs are in use, as queries and visualizations are already migrated.
```

tacole02: We try to avoid saying something is "easy", since some people might not find it easy. Can we say something else here, like "efficient" or "streamlined"?

docs/sources/mimir/send/native-histograms/_custom_buckets.md
```diff
@@ -0,0 +1,183 @@
...
+   Prometheus remote-write 2.0 is experimental. For more information, refer to the [Prometheus Remote-Write 2.0 specification](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/).
+   {{< /admonition >}}
+
+   {{< admonition type="note" >}}
+   Prometheus remote-write 2.0 doesn't fully support [metadata](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metadata).
+   {{< /admonition >}}
+
+## Migrate from classic histograms
+
+To ease the migration process, you can keep scraping classic histograms and NHCBS at the same time.
```
tacole02:
```suggestion
To ease the migration process, you can keep scraping classic histograms and NHCBs at the same time.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+  - prometheus
+  - grafana alloy
+  - instrumentation
+  - high resolution
+menuTitle: Native histograms standard buckets
+title: Send native histograms with exponential buckets to Grafana Mimir
+weight: 1
+---
+
+<!-- Note: This topic is mounted in the GEM documentation. Ensure that all updates are also applicable to GEM. -->
```
tacole02:
```suggestion
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+
+# Send native histograms with exponential buckets to Grafana Mimir
+
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) with exponential buckets, also known officially as native histograms with standard schemas, is a sample type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
+
+Native histograms with exponential buckets are different from classic Prometheus histograms in a number of ways:
+
+- Exponential bucket boundaries are calculated by a formula that depends on the scale, or resolution of the native histogram, and are not user-defined. The calculation produces exponentially increasing bucket boundaries. For details, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```
tacole02:
```suggestion
- Exponential bucket boundaries are calculated by a formula that depends on the scale, or resolution, of the native histogram and are not user-defined. The calculation produces exponentially increasing bucket boundaries. For details, refer to [Exponential bucket boundary calculation](#exponential-bucket-boundary-calculation).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+
+Some of the resulting exponential buckets for factor `1.1` rounded to two decimal places are:
+
+..., (0.84, 0.92], (0.92, 1], (1, 1.09], (1.09, 1.19], (1.19, 1.30], ...
+
+..., (76.1, 83], (83, 91], (91, 99], ...
+
+..., (512, 558], (558, 608], (608, 663], ...
+
+In Java `.nativeInitialSchema` using schema value `3` results in the same exponential bucket boundaries. For more information about the schema supported in Java, consult the documentation for [nativeInitialSchema](<https://prometheus.github.io/client_java/api/io/prometheus/metrics/core/metrics/Histogram.Builder.html#nativeInitialSchema(int)>).
```
tacole02:
```suggestion
In Java, `.nativeInitialSchema` using a schema value of `3` results in the same exponential bucket boundaries. For more information about the schema supported in Java, consult the documentation for [nativeInitialSchema](<https://prometheus.github.io/client_java/api/io/prometheus/metrics/core/metrics/Histogram.Builder.html#nativeInitialSchema(int)>).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+
+The value of `NativeHistogramMaxBucketNumber`/`nativeMaxNumberOfBuckets` limits the number of exponential buckets produced by the observations. This can be especially useful if the receiver side is limiting the number of buckets that can be sent. For more information about the bucket limit refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```
tacole02:
```suggestion
The value of `NativeHistogramMaxBucketNumber`/`nativeMaxNumberOfBuckets` limits the number of exponential buckets produced by the observations. This can be especially useful if the receiver side is limiting the number of buckets that can be sent. For more information about the bucket limit, refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+
+The duration in `NativeHistogramMinResetDuration`/`nativeResetDuration` prohibits automatic counter resets inside that period. Counter resets are related to the bucket limit, for more information refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```
tacole02:
```suggestion
The duration in `NativeHistogramMinResetDuration`/`nativeResetDuration` prohibits automatic counter resets inside that period. Counter resets are related to the bucket limit. For more information, refer to [Limit the number of buckets](#limit-the-number-of-buckets).
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+
+Use Prometheus version 3.00 or later.
+
+1. To enable scraping native histograms from the application, you need to enable the native histograms feature via a feature flag on the command line:
+
+   ```bash
+   prometheus --enable-feature=native-histograms
+   ```
+
+   This flag makes Prometheus detect and scrape [native histogram with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) over the `PrometheusProto` scrape protocol and ignore the classic histogram version of metrics that have native histograms defined as well. Classic histograms without native histogram definitions are not affected.
```
tacole02:
```suggestion
   This flag makes Prometheus detect and scrape [native histograms with exponential buckets](https://grafana.com/docs/mimir/<MIMIR_VERSION>/send/native-histograms/_exponential_buckets) over the `PrometheusProto` scrape protocol and ignore the classic histogram version of metrics that have native histograms defined as well. Classic histograms without native histogram definitions are not affected.
```

docs/sources/mimir/send/native-histograms/_exponential_buckets.md
```diff
@@ -0,0 +1,353 @@
...
+Use [Grafana Alloy](https://grafana.com/docs/alloy/<ALLOY_VERSION>) version 1.11 or later.
+
+1. To scrape native histograms, set the `scrape_native_histograms` argument in the `prometheus.scrape` component to `true`.
+
+   ```
+   scrape_native_histograms = true
+   ```
+
+   {{< admonition type="warning" >}}
+   In Grafana Alloy versions 1.10 and earlier setting `scrape_native_histograms` to `true` was not required.
```
tacole02:
```suggestion
   In Grafana Alloy versions 1.10 and earlier, setting `scrape_native_histograms` to `true` isn't required.
```

docs/sources/mimir/send/native-histograms/_index.md
```diff
@@ -15,360 +15,11 @@ weight: 100
 
 # Send native histograms to Mimir
 
-Prometheus native histograms is a data type in the Prometheus ecosystem that makes it possible to produce, store, and query a high-resolution [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) of observations.
+Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) are a data type in the Prometheus ecosystem that enables the following use cases:
```
tacole02:
```suggestion
Prometheus [native histograms](https://prometheus.io/docs/specs/native_histograms/) are a data type in the Prometheus ecosystem that enable the following use cases:
```

docs/sources/mimir/send/native-histograms/_custom_buckets.md
krajorama: trivial ?

krajorama: I've reworded to 
"- A migration to [native histograms with exponential buckets](...) takes fewer steps once NHCBs are in use, as queries and visualizations are already migrated."

tacole02: Perfect!

Issue #11920, opened 2025-06-30
index planning: Refine cost models after initial phase of testing


dimitarvdimitrov:
Validate and refine the cost models for index lookup planning. Focus on assumptions around regex matching, posting intersections, and count-min sketch accuracy/size trade-offs.

Test and analyze current cost models, enhance them where assumptions do not hold, and document findings.

Enable index lookup planning in single ingester zones, deploying cell-by-cell with close monitoring. Start deployment with a few small cells and then move to larger single-tenant cells as the target.

Progressive deployment strategy, collecting metrics and feedback from initial cells before scaling to larger ones.

dimitarvdimitrov:
> **NB:** the results here are based on a version of the code which had bugs


> * Use square root selectivity reduction instead of assuming independence

I gave this a try with this small diff https://github.com/grafana/mimir/compare/main...dimitar/lookupplan/sqrt-intersection-size

I tested it against real data and real queries from multiple production Mimir clusters. Unfortunately, it comes with some serious tradeoffs in some cases.




[ded-17-benchstat.txt](https://github.com/user-attachments/files/24059136/ded-17-benchstat.txt): Some queries are 99% faster, others 1000% slower. geomean says overall 100% slower queries, **even though** memory consumption is 30-34% lower in the geomean

The rest don't see much of a difference apart from some outlier queries:
[ded-14-benchstat.txt](https://github.com/user-attachments/files/24059140/ded-14-benchstat.txt)
[ops-03-benchstat.txt](https://github.com/user-attachments/files/24059138/ops-03-benchstat.txt)
[prod-01-benchstat.txt](https://github.com/user-attachments/files/24059139/prod-01-benchstat.txt)
[prod-10-benchstat.txt](https://github.com/user-attachments/files/24059135/prod-10-benchstat.txt)
[prod-36-benchstat.txt](https://github.com/user-attachments/files/24059137/prod-36-benchstat.txt)

dimitarvdimitrov:
> * Use square root selectivity reduction instead of assuming independence

code: https://github.com/grafana/mimir/compare/main...dimitar/lookupplan/sqrt-intersection-size

I tested it against real data and real queries from multiple production Mimir clusters. Unfortunately, it comes with some serious tradeoffs in some cases.

It's hard to tell whether this change will make a positive difference or not. Some benchmarks are mixed and I couldn't figure out why the negative effect did happen. 

[...]

PR #12830, opened 2025-09-26
lookup planning: adjust plan's intersection cost

dimitarvdimitrov:
### Problem

We've seen that ingesters can end up doing more work when this optimization is enabled. This is due to fetching significantly more series and checking them for sharding. This is not something our cost model accounted for.

### What this PR does

this PR introduces the cost of retrieving a single series from the index. The cost of doing this depends a lot on whether the block is an in-memory or an on-disk block. In-memory blocks have much more efficient sharding code. For now we're sticking with a single cost for all blocks, but we can change that later.

### why 10?

I added a benchmark to help me come up with the new number. I'm comparing the baseline of 2ns on my machine with the 2ns that it takes to do a string comparison (see https://github.com/grafana/mimir-prometheus/pull/990). This means our cost should be around 10-40. 10 is 3 orders of magnitude higher than what we had before (0.01) already, so I was cautious not to move this too much

<details><summary>Cost of sharded vs non-sharded postings iteration</summary>
<p>

false: not sharded
true: sharded

```
benchstat -col=/sharded -row=/size -filter='(/sharded:false OR /reuseCache:true)' 
        │    false     │                   true                   │
        │    sec/op    │    sec/op      vs base                   │
128       326.1n ± ∞ ¹   2937.0n ± ∞ ¹          ~ (p=0.100 n=3) ²
128K      301.5µ ± ∞ ¹   4943.5µ ± ∞ ¹          ~ (p=0.100 n=3) ²
1M        2.435m ± ∞ ¹   83.056m ± ∞ ¹          ~ (p=0.100 n=3) ²
geomean   62.10µ          1.064m        +1614.11%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

        │    false     │                   true                   │
        │ sec/posting  │  sec/posting   vs base                   │
128       2.548n ± ∞ ¹   22.940n ± ∞ ¹          ~ (p=0.100 n=3) ²
128K      2.300n ± ∞ ¹   37.720n ± ∞ ¹          ~ (p=0.100 n=3) ²
1M        2.322n ± ∞ ¹   79.210n ± ∞ ¹          ~ (p=0.100 n=3) ²
geomean   2.387n          40.92n        +1614.16%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

</p>
</details> 

### Potential next steps
1. detect whether the block is in-memory or on-disk and adjust our plans.
2. adjust the cost based on whether we have query sharding enabled or not; hashing series is expensive

1\. will probably give higher returns

related to https://github.com/grafana/mimir/issues/11920


chencs (approved):
LGTM; for the future, I'm wondering if it would make sense to make `estimatedRetrievedPostingCost` and `estimatedRetrievedSeriesCost` configurable so that tweaking them as we figure it out doesn't require rolling out a new image.

PR #12831, opened 2025-09-27
Incorporate matcher deduplication to lookup planning

chencs:
#### What this PR does
This PR creates a new `MatcherReducerPlanner`, which can be used to deduplicate matchers and remove "nonselective matchers" if more selective matchers exist for the same label name.

Nonselective matchers are conservatively, an explicit set of matchers which would select all values of a label name:
* `foo=~".*"`
* `foo!=""`
* `foo!~""`

Conversely, selective matchers are any matcher not in the above set.

If no selective matchers exist for a label name, one nonselective matcher will be preserved.

Relates to https://github.com/grafana/mimir/issues/11920

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/lookupplan/matcher_reducer_planner.go
+++ b/pkg/ingester/lookupplan/matcher_reducer_planner.go
@@ -1,0 +1,108 @@
+package lookupplan
+
+import (
+	"context"
+	"slices"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/tsdb/index"
+)
+
+// MatcherReducerPlanner deduplicates matchers from the input plan,
+// and removes matchers that select for all non-empty values if a more selective matcher for the same label name already exists.
+// It returns the input index and scan matchers in the input order.
+// If matchers are duplicated across the index and scan matchers, they will be returned as index-only matchers.
+type MatcherReducerPlanner struct{}
+
+// concreteLookupPlan implements LookupPlan by storing pre-computed index and scan matchers.
+type concreteLookupPlan struct {
+	indexMatchers []*labels.Matcher
+	scanMatchers  []*labels.Matcher
+}
+
+func (p concreteLookupPlan) ScanMatchers() []*labels.Matcher {
+	return p.scanMatchers
+}
+
+func (p concreteLookupPlan) IndexMatchers() []*labels.Matcher {
+	return p.indexMatchers
+}
+
+func (p MatcherReducerPlanner) PlanIndexLookup(ctx context.Context, inPlan index.LookupPlan, minT, maxT int64) (index.LookupPlan, error) {
+	if planningDisabled(ctx) {
+		return inPlan, nil
+	}
+
+	// For the purpose of deduping, we'll treat all matchers the same for now
+	matchers := slices.Concat(inPlan.IndexMatchers(), inPlan.ScanMatchers())
+
+	matchers = filterDuplicateLabelNameNonemptyMatchers(matchers)
+	dedupedMatchers := make(map[labels.Matcher][]bool, len(matchers))
+	for _, m := range matchers {
+		dedupedMatchers[*m] = []bool{true}
+	}
+	outIndexMatchers := make([]*labels.Matcher, 0)
+	outScanMatchers := make([]*labels.Matcher, 0)
+	for _, m := range inPlan.IndexMatchers() {
+		// We only want to add the matcher if it hasn't already been added to an output slice
+		if _, ok := dedupedMatchers[*m]; ok && len(dedupedMatchers[*m]) < 2 {
+			outIndexMatchers = append(outIndexMatchers, m)
+			dedupedMatchers[*m] = append(dedupedMatchers[*m], true)
+		}
+	}
+	for _, m := range inPlan.ScanMatchers() {
+		if _, ok := dedupedMatchers[*m]; ok && len(dedupedMatchers[*m]) < 2 {
+			outScanMatchers = append(outScanMatchers, m)
+			dedupedMatchers[*m] = append(dedupedMatchers[*m], true)
+		}
+	}
+	return &concreteLookupPlan{
+		indexMatchers: outIndexMatchers,
+		scanMatchers:  outScanMatchers,
+	}, nil
+}
+
+// filterDuplicateLabelNameNonemptyMatchers returns a slice of the input matchers,
+// excluding any matchers with duplicate label names where one of said matchers matches all nonempty values.
+// For example, an input of namespace="foo", namespace!="" should return only namespace="foo".
+// If multiple matchers all match any nonempty value, only one matcher for that label name will be returned.
+// Note that the input order is not maintained.
+func filterDuplicateLabelNameNonemptyMatchers(ms []*labels.Matcher) []*labels.Matcher {
+	matchersByName := make(map[string][]*labels.Matcher)
+	for _, m := range ms {
+		if _, ok := matchersByName[m.Name]; !ok {
+			matchersByName[m.Name] = make([]*labels.Matcher, 0, 1)
+		}
+		matchersByName[m.Name] = append(matchersByName[m.Name], m)
+	}
+	matcherNamesReturned := make(map[string]bool, len(matchersByName))
+	filteredMatchers := make([]*labels.Matcher, 0)
+
+	for name, matchers := range matchersByName {
+		if len(matchers) > 1 {
+			for i, matcherForName := range matchers {
+				// If this is the last matcher for the label name,
+				// and we haven't added any matchers for this label name to the result set,
+				// add this matcher to the result set.
+				if (i == len(matchers)-1) && !matcherNamesReturned[name] {
+					filteredMatchers = append(filteredMatchers, matcherForName)
+					matcherNamesReturned[name] = true
+					continue
+				}
+				// Otherwise, check for nonempty matchers and skip them
+				if matcherForName.Type == labels.MatchRegexp && matcherForName.Value == ".*" {
+					continue
+				}
+				if (matcherForName.Type == labels.MatchNotRegexp || matcherForName.Type == labels.MatchNotEqual) && matcherForName.Value == "" {
+					continue
+				}
+				filteredMatchers = append(filteredMatchers, matcherForName)
+				matcherNamesReturned[name] = true
+			}
+		} else {
+			filteredMatchers = append(filteredMatchers, matchers...)
+			matcherNamesReturned[name] = true
+		}
+	}
+	return filteredMatchers
+}
--- a/pkg/ingester/lookupplan/matcher_reducer_planner.go
+++ b/pkg/ingester/lookupplan/matcher_reducer_planner.go
@@ -75,34 +75,21 @@ func filterDuplicateLabelNameNonemptyMatchers(ms []*labels.Matcher) []*labels.Ma
 		}
 		matchersByName[m.Name] = append(matchersByName[m.Name], m)
 	}
-	matcherNamesReturned := make(map[string]bool, len(matchersByName))
 	filteredMatchers := make([]*labels.Matcher, 0)
 
-	for name, matchers := range matchersByName {
-		if len(matchers) > 1 {
-			for i, matcherForName := range matchers {
-				// If this is the last matcher for the label name,
-				// and we haven't added any matchers for this label name to the result set,
-				// add this matcher to the result set.
-				if (i == len(matchers)-1) && !matcherNamesReturned[name] {
-					filteredMatchers = append(filteredMatchers, matcherForName)
-					matcherNamesReturned[name] = true
-					continue
-				}
-				// Otherwise, check for nonempty matchers and skip them
-				if matcherForName.Type == labels.MatchRegexp && matcherForName.Value == ".*" {
-					continue
-				}
-				if (matcherForName.Type == labels.MatchNotRegexp || matcherForName.Type == labels.MatchNotEqual) && matcherForName.Value == "" {
-					continue
-				}
-				filteredMatchers = append(filteredMatchers, matcherForName)
-				matcherNamesReturned[name] = true
+	for _, matchers := range matchersByName {
+		matchersToAddForName := make([]*labels.Matcher, 0)
+		for i, m := range matchers {
+			if i == len(matchers)-1 && len(matchersToAddForName) == 0 {
+				matchersToAddForName = append(matchersToAddForName, m)
+			} else if !(
+			// If the matcher matches all non-empty values, we skip this block
+			(m.Type == labels.MatchRegexp && m.Value == ".*") ||
+				((m.Type == labels.MatchNotRegexp || m.Type == labels.MatchNotEqual) && m.Value == "")) {
+				matchersToAddForName = append(matchersToAddForName, m)
 			}
-		} else {
-			filteredMatchers = append(filteredMatchers, matchers...)
-			matcherNamesReturned[name] = true
 		}
+		filteredMatchers = append(filteredMatchers, matchersToAddForName...)
 	}
 	return filteredMatchers
 }
--- a/pkg/ingester/lookupplan/matcher_reducer_planner_test.go
+++ b/pkg/ingester/lookupplan/matcher_reducer_planner_test.go
@@ -1,0 +1,150 @@
+package lookupplan
+
+import (
+	"context"
+	"testing"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMatcherReducerPlanner_PlanIndexLookup(t *testing.T) {
+	ctxPlanningDisabled := ContextWithDisabledPlanning(context.Background())
+
+	equalMatcher, _ := labels.NewMatcher(labels.MatchEqual, "foo", "bar")
+	selectiveRegexMatcher, _ := labels.NewMatcher(labels.MatchRegexp, "foo", ".*bar.*")
+	nonSelectiveRegexMatcher, _ := labels.NewMatcher(labels.MatchRegexp, "foo", ".*")
+	nonSelectiveNotRegexMatcher, _ := labels.NewMatcher(labels.MatchNotRegexp, "foo", "")
+	nonSelectiveNotEqualsMatcher, _ := labels.NewMatcher(labels.MatchNotEqual, "foo", "")
+
+	otherLabelNonSelectiveNotRegexMatcher, _ := labels.NewMatcher(labels.MatchNotRegexp, "baz", "")
+	otherLabelSelectiveMatcher, _ := labels.NewMatcher(labels.MatchNotEqual, "baz", "bananas")
+
+	tests := []struct {
+		name         string
+		ctx          context.Context
+		inPlan       concreteLookupPlan
+		expectedPlan concreteLookupPlan
+	}{
+		{
+			name: "planning disabled should not alter plan",
+			ctx:  ctxPlanningDisabled,
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher, nonSelectiveRegexMatcher, nonSelectiveNotEqualsMatcher},
+				scanMatchers:  []*labels.Matcher{equalMatcher, nonSelectiveNotRegexMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher, nonSelectiveRegexMatcher, nonSelectiveNotEqualsMatcher},
+				scanMatchers:  []*labels.Matcher{equalMatcher, nonSelectiveNotRegexMatcher},
+			},
+		},
+		{
+			name: "planning should not alter plan with multiple selective matchers for label name",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher},
+				scanMatchers:  []*labels.Matcher{selectiveRegexMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher},
+				scanMatchers:  []*labels.Matcher{selectiveRegexMatcher},
+			},
+		},
+		{
+			name: "deduplicate index matchers",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher, equalMatcher},
+				scanMatchers:  []*labels.Matcher{},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher},
+				scanMatchers:  []*labels.Matcher{},
+			},
+		},
+		{
+			name: "deduplicate scan matchers",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{},
+				scanMatchers:  []*labels.Matcher{selectiveRegexMatcher, selectiveRegexMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{},
+				scanMatchers:  []*labels.Matcher{selectiveRegexMatcher},
+			},
+		},
+		{
+			name: "duplicate matchers across index and scan should be deduplicated to only index matcher",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher},
+				scanMatchers:  []*labels.Matcher{equalMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalMatcher},
+				scanMatchers:  []*labels.Matcher{},
+			},
+		},
+		{
+			name: "remove all non-selective matchers if a selective matcher exists",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{nonSelectiveRegexMatcher},
+				scanMatchers:  []*labels.Matcher{equalMatcher, nonSelectiveNotEqualsMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{},
+				scanMatchers:  []*labels.Matcher{equalMatcher},
+			},
+		},
+		{
+			name: "keep one non-selective matcher if no selective matcher exists",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{nonSelectiveRegexMatcher},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher, nonSelectiveNotRegexMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher},
+			},
+		},
+		{
+			name: "keep at least one matcher for each label name",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{nonSelectiveRegexMatcher, otherLabelNonSelectiveNotRegexMatcher},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher, nonSelectiveNotRegexMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{otherLabelNonSelectiveNotRegexMatcher},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher},
+			},
+		},
+		{
+			name: "keep at least one matcher for each label name, keep selective matchers over non-selective matchers",
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{nonSelectiveRegexMatcher, otherLabelNonSelectiveNotRegexMatcher},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher, nonSelectiveNotRegexMatcher, otherLabelSelectiveMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{},
+				scanMatchers:  []*labels.Matcher{nonSelectiveNotRegexMatcher, otherLabelSelectiveMatcher},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			planner := MatcherReducerPlanner{}
+			outPlan, err := planner.PlanIndexLookup(tt.ctx, tt.inPlan, 0, 0)
+			require.NoError(t, err)
+			assert.Equal(t, tt.expectedPlan.IndexMatchers(), outPlan.IndexMatchers())
+			assert.Equal(t, tt.expectedPlan.ScanMatchers(), outPlan.ScanMatchers())
+
+		})
+	}
+}
--- a/pkg/ingester/lookupplan/planner_factory.go
+++ b/pkg/ingester/lookupplan/planner_factory.go
@@ -45,5 +45,6 @@ func (p *PlannerFactory) CreatePlanner(meta tsdb.BlockMeta, reader tsdb.IndexRea
 		level.Warn(logger).Log("msg", "failed to generate statistics; queries for this block won't use query planning", "err", err)
 		return NoopPlanner{}
 	}
-	return NewCostBasedPlanner(p.metrics, stats)
+
+	return index.ChainLookupPlanners(MatcherReducerPlanner{}, NewCostBasedPlanner(p.metrics, stats))
 }
--- a/pkg/ingester/lookupplan/planner_factory_test.go
+++ b/pkg/ingester/lookupplan/planner_factory_test.go
@@ -131,14 +131,10 @@ func TestPlannerFactory_CreatePlanner(t *testing.T) {
 			expectedPlanner: reflect.TypeOf(NoopPlanner{}),
 		},
 		{
-			name: "large block with working IndexReader returns CostBasedPlanner",
-			blockMeta: tsdb.BlockMeta{
-				Stats: tsdb.BlockStats{
-					NumSeries: 15000,
-				},
-			},
+			name:            "large block with working IndexReader returns CostBasedPlanner",
+			blockMeta:       tsdb.BlockMeta{Stats: tsdb.BlockStats{NumSeries: 15000}},
 			indexReader:     createMockIndexReaderWithSeries(15000),
-			expectedPlanner: reflect.TypeOf(&CostBasedPlanner{}),
+			expectedPlanner: reflect.TypeOf(index.ChainLookupPlanners(MatcherReducerPlanner{}, CostBasedPlanner{})),
 		},
 		{
 			name: "exactly at threshold with working IndexReader returns CostBasedPlanner",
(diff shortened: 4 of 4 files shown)


pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: nit: can we preallocate to the size of `ms`. it's more expensive to resize and copy as the slice grows than to allocate a pointer or two more
```suggestion
	filteredMatchers := make([]*labels.Matcher, 0, len(ms))
```

dimitarvdimitrov, line 98:
```suggestion
	matchersByName := make(map[string][]*labels.Matcher, len(ms))
```

tiny nitpick

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -80,16 +80,24 @@ func filterDuplicateLabelNameNonemptyMatchers(ms []*labels.Matcher) []*labels.Ma
 	for _, matchers := range matchersByName {
 		matchersToAddForName := make([]*labels.Matcher, 0)
 		for i, m := range matchers {
-			if i == len(matchers)-1 && len(matchersToAddForName) == 0 {
-				matchersToAddForName = append(matchersToAddForName, m)
-			} else if !(
-			// If the matcher matches all non-empty values, we skip this block
-			(m.Type == labels.MatchRegexp && m.Value == ".*") ||
-				((m.Type == labels.MatchNotRegexp || m.Type == labels.MatchNotEqual) && m.Value == "")) {
+			// If we're on the last element and haven't kept any matchers for this label name yet, keep this one.
```
dimitarvdimitrov: nitpick, but can you explain why, the comment right now kind of only says what the `if` statement already does. maybe just add a comment on the whole loop and explain what/why we're doing with that loop

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,103 @@
...
+		matchersByName[m.Name] = append(matchersByName[m.Name], m)
+	}
+	filteredMatchers := make([]*labels.Matcher, 0)
+
+	for _, matchers := range matchersByName {
+		matchersToAddForName := make([]*labels.Matcher, 0)
+		for i, m := range matchers {
+			// If we're on the last element and haven't kept any matchers for this label name yet, keep this one.
+			// Otherwise, only keep selective matchers.
+			if (i == len(matchers)-1 && len(matchersToAddForName) == 0) || !isNonSelectiveMatcher(m) {
```
dimitarvdimitrov: i think this has a bug. if you put in a test case with your `otherLabelSelectiveMatcher` and `nonSelectiveNotEqualsMatcher` for your in matchers, the output will drop the `matcher!=""`. But that's a wrong query result because we want all results with a value for `matcher`. I think we have to do something like keep track of which label names have a "positive" matcher which doesn't also match `""` and only for them drop matchers which exclude empty values (`isNonSelectiveMatcher`).

dimitarvdimitrov (commented):
i think i found a bug; otherwise the code so far looks good

i don't know how much more work it will be but what's your take on adding some "correctness" testing for this planner? like having a set of, say, 10 hardcoded series in the test and running the in matchers against them to see which series match and then running the out matchers and assert the same set of series is selected

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: > if you put in a test case with your otherLabelSelectiveMatcher and nonSelectiveNotEqualsMatcher for your in matchers, the output will drop the matcher!="". 

Hm, I wasn't able to repro this behavior from the top level (i.e., calling `PlanIndexLookup`). In this code block, we're only looking at matchers for a single label name so I'm not sure how such a bug would manifest here.

chencs: I did push up a test as you described in 5d13ea313ae764f81f409a705ff7ddec60d6236a, but I'm not convinced it adds more coverage (and it does pass)

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Done in 1ff753412f9d56d4558edf2414081b4bbe121606

chencs:
> what's your take on adding some "correctness" testing for this planner? like having a set of, say, 10 hardcoded series in the test and running the in matchers against them to see which series match and then running the out matchers and assert the same set of series is selected

I think this could be useful if we decide to expand the definition of "nonselective matchers," but for the current, more restrictive definition, asserting that sets of series are selected is covered by asserting on the index and scan matchers, right?

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Should be fixed, mainly by generalizing set reduction.

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	matchers := slices.Concat(inPlan.IndexMatchers(), inPlan.ScanMatchers())
+	matchers = setReduceMatchers(matchers)
+
+	allowedOutMatchers := make(map[labels.Matcher]bool, len(matchers))
+	for _, m := range matchers {
+		allowedOutMatchers[*m] = false
+	}
+
+	// Rebuild the index and scan matchers in the input order, less the filtered/deduplicated matchers
+	droppedMatchers := make([]*labels.Matcher, 0)
```
dimitarvdimitrov: can we just pass nil instead of doing an allocation?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+		if matcher.Matches(m.Value) {
+			return true
+		}
+	}
+	return false
+}
+
+// filterRegexMatchers returns a subset of regex matchers which would actually reduce the result set size for either positive or negative regex matchers.
+// If a regex matcher matches any equals matchers, it is dropped, since the equals matcher will match a strict subset of values that the regex matcher would.
+func filterRegexMatchers(mf map[labels.MatchType][]*labels.Matcher, regexType labels.MatchType) []*labels.Matcher {
```
dimitarvdimitrov, line 159: let's give this some more descriptive name like `collapseOverlappingMatchers` or `dedupeOverlappingMatchers` or `removeSubsetMatchers`

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	return &concreteLookupPlan{
+		indexMatchers: outIndexMatchers,
+		scanMatchers:  outScanMatchers,
+	}, nil
+}
+
+// buildOutMatchers takes a slice of index or scan matchers, and returns the same slice in order,
+// less any duplicate matchers. allowedOutMatchers contains the set of matchers which can be returned.
+// If dedupedMatchers[*m] is false, the matcher has not been added to any result set.
+// droppedMatchers is used for logging purposes to record which matchers have been removed from the plan.
```
dimitarvdimitrov: i'd move this comment to whatever uses droppedMatchers. this function doesn't need to document how its arguments are used before/after being passed to it

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+func buildOutMatchers(inMatchers []*labels.Matcher, allowedOutMatchers map[labels.Matcher]bool, droppedMatchers []*labels.Matcher) ([]*labels.Matcher, map[labels.Matcher]bool, []*labels.Matcher) {
+	outMatchers := make([]*labels.Matcher, 0, len(inMatchers))
+	for _, m := range inMatchers {
+		// dedupedMatchers is used to both keep track of all unique matchers (evidenced by existence in the map),
+		// and whether the matcher has already been added to a set of output matchers (evidenced by the value in the map).
+		// We only want to add the matcher if it hasn't already been added to an output slice.
+		if _, ok := allowedOutMatchers[*m]; ok && !allowedOutMatchers[*m] {
```
dimitarvdimitrov:
```suggestion
		if !allowedOutMatchers[*m] {
```

this does the same

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/grafana/mimir/pkg/util"
+)
+
+// MatcherReducerPlanner deduplicates matchers from the input plan,
+// and removes matchers that select for all non-empty values if a more selective matcher for the same label name already exists.
+// It returns the input index and scan matchers in the input order.
+// If matchers are duplicated across the index and scan matchers, they will be returned as index-only matchers.
```
dimitarvdimitrov, line 20: do you think all of this will be easier to grok if we just concatenate the input scan and index matchers?

in our use case this planner will always get 0 scan matchers, so why build for other cases? you might even return an error if there are any scan matchers to keep your tests simpler too

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+		indexMatchers: outIndexMatchers,
+		scanMatchers:  outScanMatchers,
+	}, nil
+}
+
+// buildOutMatchers takes a slice of index or scan matchers, and returns the same slice in order,
+// less any duplicate matchers. allowedOutMatchers contains the set of matchers which can be returned.
+// If dedupedMatchers[*m] is false, the matcher has not been added to any result set.
+// droppedMatchers is used for logging purposes to record which matchers have been removed from the plan.
+func buildOutMatchers(inMatchers []*labels.Matcher, allowedOutMatchers map[labels.Matcher]bool, droppedMatchers []*labels.Matcher) ([]*labels.Matcher, map[labels.Matcher]bool, []*labels.Matcher) {
```
dimitarvdimitrov: nit to rename `allowedOutMatchers` to `alreadyIncludedOutMatchers`. right now the function reads like "if the matcher is not allowed, then include it in out matchers, otherwise drop it"

dimitarvdimitrov: hmm nit on naming: doesn't `deduplciateMatchers` describe better what this function does?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	matchers = setReduceMatchers(matchers)
+
+	allowedOutMatchers := make(map[labels.Matcher]bool, len(matchers))
+	for _, m := range matchers {
+		allowedOutMatchers[*m] = false
+	}
+
+	// Rebuild the index and scan matchers in the input order, less the filtered/deduplicated matchers
+	droppedMatchers := make([]*labels.Matcher, 0)
+	outIndexMatchers, allowedOutMatchers, droppedMatchers := buildOutMatchers(inPlan.IndexMatchers(), allowedOutMatchers, droppedMatchers)
```
dimitarvdimitrov: would we save some work if we first deduplicate matchers and then do the `setReduce`? does it matter?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+			outMatchers = append(outMatchers, m)
+			allowedOutMatchers[*m] = true
+		} else {
+			droppedMatchers = append(droppedMatchers, m)
+		}
+	}
+	return outMatchers, allowedOutMatchers, droppedMatchers
+}
+
+func setReduceMatchers(ms []*labels.Matcher) []*labels.Matcher {
```
dimitarvdimitrov, line 96: some godoc would be helpful here. i think what you have in the description is pretty good

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: oh, we still do `dedupeMatchers` in `setReduceMatchers`. I don't understand why we do it here again? is it just so we dedup them across index & scan?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+		matchersByType := groupMatchersByType(matchers)
+		outMatchers = append(outMatchers, matchersByType[labels.MatchEqual]...)
+		// If we have more than one unique equals matcher, we can just return those;
+		// we know we'll only return an empty set
+		if len(matchersByType[labels.MatchEqual]) > 1 {
+			continue
+		}
+		outMatchers = append(outMatchers, filterRegexMatchers(matchersByType, labels.MatchRegexp)...)
+		outMatchers = append(outMatchers, filterNotEqualsMatchers(matchersByType)...)
+		outMatchers = append(outMatchers, filterRegexMatchers(matchersByType, labels.MatchNotRegexp)...)
```
dimitarvdimitrov, line 117: would be nice if we can include the matchers we filter out in the trace. dropped matchers currently includes only the deduplicated matchers across index&scan

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	// Group matchers by their label names so we can evaluate each label name independently
+	matchersByName := make(map[string][]*labels.Matcher)
+	for _, m := range ms {
+		if _, ok := matchersByName[m.Name]; !ok {
+			matchersByName[m.Name] = make([]*labels.Matcher, 0, 1)
+		}
+		// Always drop wildcard matchers
+		if m.Type == labels.MatchRegexp && m.Value == ".*" {
+			continue
+		}
```
dimitarvdimitrov: nit: this is a bit tucked away and mixed with other logic. WDYT about pulling it out in a function of its own which takes `[]*labels.Matcher` and returns `[]*labels.Matcher`?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+func filterRegexMatchers(mf map[labels.MatchType][]*labels.Matcher, regexType labels.MatchType) []*labels.Matcher {
+	var matchers []*labels.Matcher
+	switch regexType {
+	case labels.MatchRegexp:
+		matchers = mf[labels.MatchRegexp]
+	case labels.MatchNotRegexp:
+		matchers = mf[labels.MatchNotRegexp]
+	default:
+		// Do not filter anything if passed a non-regex type.
+		return mf[regexType]
```
dimitarvdimitrov: nitpick but i would panic instead of doing the unexpected thing silently

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	outMatchers := make([]*labels.Matcher, 0, len(matchers))
+	for _, m := range matchers {
+		// If a regex matcher matches any equals matcher, that equals matcher is a subset of the regex,
+		// and we should not add it to outMatchers.
```
dimitarvdimitrov:
```suggestion
		// and we should not add the regex matcher to outMatchers.
```

as a non-native speaker the `it` was confusing

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
dimitarvdimitrov, line 1: have you thought about adding the correctness unit tests too?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+		if !matcherMatchesAnyValues(m, mf[labels.MatchEqual]) {
+			outMatchers = append(outMatchers, m)
+		}
+	}
```
dimitarvdimitrov, line 181: won't this work for equals matchers too and allow to deduplicate them? 

for example with `{foo="x", foo="y", foo="x"}` if `m` is `foo="x"` and then we can deduplicate the other `foo="x"`, but keep `foo="y"`. Need to make sure not to compare a matcher with itself but that should be a simple `==` check on the pointer.

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+	return outMatchers
+}
+
+// filterNotEqualsMatchers returns a subset of not-equals matchers which should actually reduce the result set size.
+// This subset is determined by comparing each not-equals matchers against equals and not-regex matchers.
+// A not-equals matcher is dropped if:
+// - there exists any not-regex matcher which would also remove the not-equals matcher value from the result set, OR
+// - there exists any equals matcher for a value that is matched by the not-equals matcher
```
dimitarvdimitrov: is this saying that if we have `{foo!="x", foo="x"}` then we'd drop `foo!="x"` and end up with `{foo="x"}`?  i think that's a wrong transformation

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,202 @@
...
+func filterNotEqualsMatchers(mf map[labels.MatchType][]*labels.Matcher) []*labels.Matcher {
+	outMatchers := make([]*labels.Matcher, 0, len(mf[labels.MatchNotEqual]))
+	// If the value of a not-equals matcher is matched by the inverse of any not-regex matcher,
+	// that not-equals matcher can be dropped, because it will select a subset of the regex.
+	// Using the example of x!="a" and x!~"a.*",
+	// x!~"a.*" will exclude more values than x!="a".
+	for _, m := range mf[labels.MatchNotEqual] {
+		matchesNoRegex := true
```
dimitarvdimitrov: nit: matchesNoRegex -> isSubsetOfNotRegexMatcher or noRegexContainsM

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: an example or two in the function godoc would help

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: also an example or two would be helpful

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
```diff
@@ -0,0 +1,203 @@
...
+			ctx:  context.Background(),
+			inPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalsMatcher},
+				scanMatchers:  []*labels.Matcher{supersetRegexpMatcher},
+			},
+			expectedPlan: concreteLookupPlan{
+				indexMatchers: []*labels.Matcher{equalsMatcher},
+				scanMatchers:  []*labels.Matcher{},
+			},
+		},
```
dimitarvdimitrov: are these test cases easy to read for you? i find it hard to have to go back to the top to read what `supersetRegexpMatcher` is. have you thought about using `parseMatcher` and having the matcher literals in each test case?

```
				indexMatchers: []*labels.Matcher{parseMatcher(`foo="bar"`},
```


or just having a slice of strings as indexMatchers and scanMatchers and converting them to labels.Matchers in the loop of the test?

dimitarvdimitrov (commented):
nice one! I think this reads more easily that the previous version. i left a few comments, a lot of nitpicks though. 

* i think there's a bug in turning `{foo!="x", foo="x"}` into just `{foo="x"}`
* you might be able to simplify this to only work with a single slice of matchers - we shouldn't need to support both index and scan matchers. 
* some notes on the tests too, would be nice to get better tests before we can merge this, so bugs like the one i found don't creep in

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Ah, no, if we have `{fool != "x", foo = "x"}`, checking `foo != "x"` against `x` (the value of `foo = "x"`) will return false.

If we have `{foo != "x", foo = "y"}`, `foo != "x"` _would_ match `y`, so we would drop `foo != "x"`.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: I'm not sure I understand. Are you suggesting to modify the input plan so we only return index matchers?

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Mm, I erred on the side of trying to match the input plan more exactly rather than being loose about it, so `buildOutMatchers` deduplicates matchers between the index and scan matcher slices while keeping the remaining matchers in the input order. I think that needs to be done last. 

We do deduplicate the whole set of (index + scan) matchers in `setReduce`, but return everything as one slice, and not guaranteed to match the input order.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: > is it just so we dedup them across index & scan?

Yeah, I don't know that we can do that part earlier. I think almost everything else in `setReduce` would be fine/behave the same with duplicate matchers _except_ the early exit on multiple unique `=` matchers, so we could instead only dedupe equals matchers in `setReduce`

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: I can pull it into `filterRegexMatchers`

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: `allowedOutMatchers[*m]` will be false if `ok` is false, but a matcher is only an "allowed" matcher if it's in the map (`ok` is true) and it hasn't already been added (`allowedOutMatchers[*m]` is false)

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: I think it'd be cheaper to just dedupe the equals matchers? We can limit deduping to just that set, since, as you pointed out, deduping across index + scan matchers also functionally dedupes matchers generally.

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
chencs: That's a good idea! Checking expected regex matchers will be a little more complicated (or like, it'll be a weaker equality check), but each test will be more readable.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Technically all of the `filter<type>Matchers` functions are collapsing overlapping matchers and removing subset matchers; do you find the naming unclear as is?

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: > dropped matchers currently includes only the deduplicated matchers across index&scan

This was a good callout; this shouldn't be true (i.e., if there are multiple of the same matcher, I think we should see it multiple times in `droppedMatchers`, since we add to `droppedMatchers` above every time we skip appending to `outMatchers` in `buildOutMatchers`.

However, when I was checking on this, I realized that because I reuse pointers in tests, `allowedOutMatchers` is actually correctly removing regex matchers that wouldn't actually necessarily have the same `FastRegexMatcher` pointer in real life.

I updated `allowedOutMatchers` to use the stringified matcher as the key instead of a value in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
chencs: Redid the tests in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Renamed to `noRegexMatchesValue` in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Added examples in 648266ca20d16683e2ff4ca6e93002032b8bceda, as well as unit tests to explicitly illustrate the concern you described.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Switched to deduping only equals matchers in 648266ca20d16683e2ff4ca6e93002032b8bceda, and now just rely on the final scan to build output index + scan matchers to dedupe others.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Removed (mostly) redundant deduplication in `setReduce` in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Added in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Used the named map value instead in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Done in 648266ca20d16683e2ff4ca6e93002032b8bceda

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: pff i don't know. i guess i wanted the set-language to stand out more. the examples may be enough

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: we can do that or we can return an error if the input plan has any scan matchers. Either way it's not a code path we'd hit ~~often~~ ever. I was thinking about this planner sitting in front of the existing planner, so this planner would be getting plans only with index matchers. `outScanMatchers` will always be empty is what i'm saying

pkg/ingester/lookupplan/matcher_reducer_planner.go
dimitarvdimitrov: i think trying to keep the order makes this a bit brittle. there's nothing in matcher's order which changes meaning. if a place needs a particular order, then it should enforce that order before doing any operations. i wouldn't rely on one planner sorting the matchers one way before something else in a different part of the code using that sort order

i think this become moot if we only work with a single slice of matchers https://github.com/grafana/mimir/pull/12831#discussion_r2402243753

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,227 @@
...
+	case labels.MatchNotRegexp:
+		matchers = mf[labels.MatchNotRegexp]
+	default:
+		// Do not filter anything if passed a non-regex type.
+		panic("cannot filter unknown type regex matchers")
+	}
+	outMatchers := make([]*labels.Matcher, 0, len(matchers))
+	for _, m := range matchers {
+		// Always drop wildcard matchers
+		if m.Value == ".*" {
```
dimitarvdimitrov: we shouldn't be dropping those for not-regexp matchers? For example `{__name__!~".*"}` matches no series, so if we eliminate `!~".*"`, then we're changing the meaning of the query

can you also add some tests like we talked about - generating some sample series and running a query with and without this planner to check for correctness

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,227 @@
...
+
+// filterNotEqualsMatchers returns a subset of not-equals matchers which should actually reduce the result set size.
+// This subset is determined by comparing each not-equals matchers against equals and not-regex matchers.
+// A not-equals matcher is dropped on any of these conditions:
+//   - there are any not-regex matchers which would also remove the not-equals matcher value from the result set
+//   - any equals matcher for a value that is matched by the not-equals matcher
+//
+// Examples:
+//   - for the matcher set {foo!="bar", foo="b"}, foo!="bar" does match the value "b", so foo!="bar" is dropped.
+//   - for the matcher set {foo!="bar", foo="bar"}, foo!="bar" matches the value "bar", so foo!="bar" is not dropped.
```
dimitarvdimitrov: both sentences say that `foo!="bar"` matches the other value. i think one of them should say that foobar doesn't match the value?

pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -0,0 +1,227 @@
...
+func (p concreteLookupPlan) ScanMatchers() []*labels.Matcher {
+	return p.scanMatchers
+}
+
+func (p concreteLookupPlan) IndexMatchers() []*labels.Matcher {
+	return p.indexMatchers
+}
+
+func (p MatcherReducerPlanner) PlanIndexLookup(ctx context.Context, inPlan index.LookupPlan, minT, maxT int64) (index.LookupPlan, error) {
+	span, _, traceSampled := tracing.SpanFromContext(ctx)
```
dimitarvdimitrov: nit to move this close to where it's used

chencs: Moved in 1811cbc245cf36b9309d841f1339b1b40be1f1ea

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Fixed in 1811cbc245cf36b9309d841f1339b1b40be1f1ea

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: Fixed in 1811cbc245cf36b9309d841f1339b1b40be1f1ea; working on tests next.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: In 1811cbc245cf36b9309d841f1339b1b40be1f1ea, we now skip planning (return the inPlan) if it contains scan matchers.

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
chencs: Done in 059fb1fa76544030a7862f030cc62d9e7d62dc3d. 

I had to rebase to resolve CHANGELOG conflicts, so other commit links might be broken now, but this is the latest.

pkg/ingester/lookupplan/matcher_reducer_planner.go
chencs: I tossed the ordering requirement in 82d97e18d5025715ee7373894617a7841bc86ecf.

We still do iterate over input matchers once to drop matchers, because it was going to get (imo) annoying to read if I moved the logic to build/track dropped matchers to inside of `setReduce`.

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
```diff
@@ -147,17 +142,222 @@ func TestMatcherReducerPlanner_PlanIndexLookup(t *testing.T) {
...
+		{
+			name:     "regex matcher and intersecting equals matcher",
+			matchers: []string{`__name__=~".*requests.*"`, `__name__="http_requests_total"`},
+		},
+		{
+			name:     "regex matcher and non-intersecting equals matcher",
+			matchers: []string{`__name__=~".*requests.*"`, `__name__="cpu_usage_percent"`},
+		},
+		{
+			name:     "regex matcher and non-subtractive not-equals matcher",
```
dimitarvdimitrov:
```suggestion
			name:     "regex matcher and subtractive, non-overlapping, not-equals matcher",
```

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
```diff
@@ -147,17 +142,222 @@ func TestMatcherReducerPlanner_PlanIndexLookup(t *testing.T) {
...
+		{
+			name:     "wildcard matcher and not-equals matcher for different label names",
+			matchers: []string{`instance=~".*"`, `__name__!=""`},
+		},
+		{
+			name:     "regex matcher and intersecting equals matcher",
+			matchers: []string{`__name__=~".*requests.*"`, `__name__="http_requests_total"`},
+		},
+		{
+			name:     "regex matcher and non-intersecting equals matcher",
```
dimitarvdimitrov:
```suggestion
			name:     "regex matcher and non-overlapping equals matcher",
```

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
```diff
@@ -147,17 +142,222 @@ func TestMatcherReducerPlanner_PlanIndexLookup(t *testing.T) {
...
+			matchers: []string{`__name__!="http_requests_total"`, `__name__="http_requests_total"`},
+		},
+		{
+			name:     "not-equals matcher and intersecting not-equals matcher",
+			matchers: []string{`__name__!=""`, `__name__!="cache_requests_total"`},
+		},
+		{
+			name:     "not-equals matcher and non-intersecting not-equals matcher for different label names",
+			matchers: []string{`backend!=""`, `__name__!="cache_requests_total"`},
+		},
```
dimitarvdimitrov:
```suggestion
			name:     "not-equals matcher and overlapping equals matcher",
			matchers: []string{`__name__!=""`, `__name__="http_requests_total"`},
		},
		{
			name:     "not-equals matcher and non-overlapping equals matcher",
			matchers: []string{`__name__!="http_requests_total"`, `__name__="http_requests_total"`},
		},
		{
			name:     "not-equals matcher and overlapping not-equals matcher",
			matchers: []string{`__name__!=""`, `__name__!="cache_requests_total"`},
		},
		{
			name:     "not-equals matcher and non-overlapping not-equals matcher for different label names",
			matchers: []string{`backend!=""`, `__name__!="cache_requests_total"`},
		},
```

pkg/ingester/lookupplan/matcher_reducer_planner_test.go
```diff
@@ -147,17 +142,222 @@ func TestMatcherReducerPlanner_PlanIndexLookup(t *testing.T) {
...
+			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*usage.*"`},
+		},
+		{
+			name:     "regex matcher and non-intersecting not-regex matcher",
+			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*"`},
+		},
+		{
+			name:     "regex matcher and partially intersecting not-regex matcher",
+			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*cache.*"`},
+		},
```
dimitarvdimitrov:
```suggestion
			name:     "regex matcher and overlapping regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__=~".*total.*"`},
		},
		{
			name:     "regex matcher and non-overlapping regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__=~".*percent.*"`},
		},
		{
			name:     "regex matcher and partially overlapping regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__=~".*http.*"`},
		},
		{
			name:     "regex matcher and overlapping not-regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*usage.*"`},
		},
		{
			name:     "regex matcher and non-overlapping not-regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*"`},
		},
		{
			name:     "regex matcher and partially overlapping not-regex matcher",
			matchers: []string{`__name__=~".*requests.*"`, `__name__!~".*cache.*"`},
		},
```

PR #12832, opened 2025-09-28
ruler: add ruler_max_rule_evaluation_results per-tenant limit

narqo:
#### What this PR does

This PR adds a new per-tenant limit, `ruler_max_rule_evaluation_results`. The limit controls the maximum number of alerts an alerting rule or series a recording rule can produce for the group.

The underlying implementation relies on prometheus to automatically enforce these limits during evaluation (ref to the related [docs in Prometheus](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#limiting-alerts-and-series)).

The user-visible behaviour of the limit is similar to what Grafana has for its managed alerts and ruler. Ref changes https://github.com/grafana/grafana/pull/89468

Relates to https://github.com/grafana/pir-actions/issues/400

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -5841,7 +5841,7 @@
           "kind": "field",
           "name": "ruler_max_rule_evaluation_results",
           "required": false,
-          "desc": "Maximum number of alerts an alerting rule or series a recording rule can produce for rule groups. 0 is no limit.",
+          "desc": "Maximum number of alerts or series one alerting rule or one recording rule respectively can produce. 0 is no limit.",
           "fieldValue": null,
           "fieldDefaultValue": 0,
           "fieldFlag": "ruler.max-rule-evaluation-results",
--- a/docs/sources/mimir/configure/configuration-parameters/index.md
+++ b/docs/sources/mimir/configure/configuration-parameters/index.md
@@ -4162,8 +4162,8 @@ ruler_alertmanager_client_config:
 # CLI flag: -ruler.min-rule-evaluation-interval
 [ruler_min_rule_evaluation_interval: <duration> | default = 0s]
 
-# (experimental) Maximum number of alerts an alerting rule or series a recording
-# rule can produce for rule groups. 0 is no limit.
+# (experimental) Maximum number of alerts or series one alerting rule or one
+# recording rule respectively can produce. 0 is no limit.
 # CLI flag: -ruler.max-rule-evaluation-results
 [ruler_max_rule_evaluation_results: <int> | default = 0]
 
--- a/cmd/mimir/help-all.txt.tmpl
+++ b/cmd/mimir/help-all.txt.tmpl
@@ -2882,7 +2882,7 @@ Usage of ./cmd/mimir/mimir:
   -ruler.max-notification-batch-size int
     	Maximum number of notifications to send to Alertmanager in one request. (default 256)
   -ruler.max-rule-evaluation-results int
-    	[experimental] Maximum number of alerts an alerting rule or series a recording rule can produce for rule groups. 0 is no limit.
+    	[experimental] Maximum number of alerts or series one alerting rule or one recording rule respectively can produce. 0 is no limit.
   -ruler.max-rule-groups-per-tenant int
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,7 @@
 * [ENHANCEMENT] Compactor: Log sizes of downloaded and uploaded blocks. #12656
 * [ENHANCEMENT] Ingester: Add `cortex_ingest_storage_reader_receive_and_consume_delay_seconds` metric tracking the time between when a write request is received in the distributor and its content is ingested in ingesters, when the ingest storage is enabled. #12751
 * [ENHANCEMENT] Ruler: Add `ruler_evaluation_consistency_max_delay` per-tenant configuration option support, to specify the maximum tolerated ingestion delay for eventually consistent rule evaluations. This feature is used only when ingest storage is enabled. By default, no maximum delay is enforced. #12751
+* [ENHANCEMENT] Ruler: Add `ruler_max_rule_evaluation_results` per-tenant configuration option to limit the maximum number of alerts an alerting rule or series a recording rule can produce for the group. By default, no maximum delay is enforced. #12832
```
```suggestion
* [ENHANCEMENT] Ruler: Add the `ruler_max_rule_evaluation_results` per-tenant configuration option to limit the maximum number of alerts an alerting rule or series a recording rule can produce for the group. By default, no maximum delay is enforced. #12832
```

cmd/mimir/config-descriptor.json
```diff
@@ -6067,6 +6067,17 @@
           "fieldType": "duration",
           "fieldCategory": "experimental"
         },
+        {
+          "kind": "field",
+          "name": "ruler_max_rule_evaluation_results",
+          "required": false,
+          "desc": "Maximum number of alerts an alerting rule or series a recording rule can produce for rule groups. 0 is no limit.",
```
```suggestion
          "desc": "Maximum number of alerts an alerting rule or series a recording rule can produce for rule groups. Set to 0 for no limit.",
```

docs/sources/mimir/configure/configuration-parameters/index.md
```diff
@@ -5140,6 +5140,11 @@ ruler_alertmanager_client_config:
 # CLI flag: -ruler.min-rule-evaluation-interval
 [ruler_min_rule_evaluation_interval: <duration> | default = 0s]
 
+# (experimental) Maximum number of alerts an alerting rule or series a recording
```
56quarters: I don't quite understand the "for rule groups" wording. Does that mean that this limit is enforced for all results generated by all rules within a single rule group? Or enforced for each rule within a rule group?

pkg/ruler/rulespb/rules.proto
```diff
@@ -50,6 +50,8 @@ message RuleGroupDesc {
     (gogoproto.nullable) = false,
     (gogoproto.customtype) = "github.com/grafana/mimir/pkg/mimirpb.LabelAdapter"
   ];
+
+  int32 limit = 15;
```
56quarters, line 55: Can you add a comment here explain this and why we're setting it on the rule group even though it comes from per-tenant limits.

56quarters, line 55: This seems like it should be a `uint32` since a negative limit doesn't make sense.

docs/sources/mimir/configure/configuration-parameters/index.md
narqo: In Prometheus, the limit is set for all rules in a group. But it enforces the number of results per-rule evaluation. I updated the wording to be more explicit for how it works

pkg/ruler/rulespb/rules.proto
narqo: From Prometheus point, this not a new functionality. Mimir just didn't support it. In this PR we enforce this limit per-user, in the same way we enforce min-evaluation interval for the `interval` field.

narqo: I agree, but it's `int` in Prometheus, so I'd keep it compatible ([doc](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#rule_group))

PR #12835, opened 2025-09-29
Add support for query sharding inside MQE

charleskorn:
#### What this PR does

This PR implements support for query sharding inside MQE as an AST optimisation pass.

There are some missing features that will be added in subsequent PRs:

* we need to buffer query responses for sharded legs in query-frontends to avoid deadlocking if multiple query-frontends are evaluating shardable queries at the same time and the total number of legs exceeds the number of available querier workers: https://github.com/grafana/mimir/pull/12885
* we need to limit the number of sharded legs for a single query request evaluated at once, similar to the behaviour of the `limitedParallelismRoundTripper`: https://github.com/grafana/mimir/pull/12883
* we need to block user requests that use the `__sharded_concat__()`, as this may otherwise lead to unpredictable behaviour: https://github.com/grafana/mimir/pull/12886

:warning: The integration tests in this PR will only reliably pass with the fix from https://github.com/grafana/mimir/pull/12856. 

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/planning/analysis/analysis.go
charleskorn, line 1: Note to reviewers: the majority of this file and `analysis_test.go` have been moved from `planning.go`/`planning_test.go` to a) better separate concerns into different packages and b) resolve an import cycle for the `querymiddleware` package needed so we can support the `Sharding-Control` header when performing query analysis

cmd/mimir/config-descriptor.json
```diff
@@ -7859,6 +7859,17 @@
           "fieldFlag": "query-frontend.parallelize-shardable-queries",
           "fieldType": "boolean"
         },
+        {
+          "kind": "field",
+          "name": "use_mimir_query_engine_for_sharding",
+          "required": false,
+          "desc": "True to enable performing query sharding inside the Mimir query engine (MQE). Has no effect if sharding is disabled. Requires remote execution and MQE to be enabled.",
```
```suggestion
          "desc": "Set to true to enable performing query sharding inside the Mimir query engine (MQE). This setting has no effect if sharding is disabled. You need to enable remote execution and MQE.",
```

tacole02 (approved):
Docs look good! I left a few minor suggestions. Thank you!

cmd/mimir/config-descriptor.json
charleskorn: Resolved in 40992b4

pkg/streamingpromql/optimize/ast/sharding/concat.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	combined, err := types.SeriesMetadataSlicePool.Get(seriesCount, c.MemoryConsumptionTracker)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, result := range results {
+		combined = append(combined, result...)
+		clear(result) // We want to retain the SeriesMetadata elements, but we need to return the slice to the pool.
```
56quarters: Is this because we sometimes mangle values returned to the pool? If so, can you mention it's because of mangling in the comment.

56quarters (approved):
Nicely done. Way smaller than I thought it would be.

pkg/streamingpromql/optimize/ast/sharding/concat.go
charleskorn: It's to avoid reducing the memory consumption estimate based on the labels in `result`, as we're retaining them in `combined`. I've clarified this in 3566945.

PR #12844, opened 2025-09-29
distributor: Warn on truncated or dropped label value

tcard:
#### What this PR does

Adds warning log when a label value gets truncated or dropped.

Followup of #12627

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/validate_test.go
+++ b/pkg/distributor/validate_test.go
@@ -132,6 +132,7 @@ func TestValidateLabels(t *testing.T) {
 	manager, err := costattribution.NewManager(5*time.Second, 10*time.Second, log.NewNopLogger(), overrides(limits), reg, careg)
 	require.NoError(t, err)
 
+	var logged logRecorder
 	ds, ingesters, _, _ := prepare(t, prepConfig{
 		numIngesters:       2,
 		happyIngesters:     2,
@@ -140,6 +141,7 @@ func TestValidateLabels(t *testing.T) {
 		overrides:          overrides,
 		reg:                reg,
 		costAttributionMgr: manager,
+		logger:             log.NewLogfmtLogger(&logged),
 	})
 	d := ds[0]
 
@@ -177,6 +179,7 @@ func TestValidateLabels(t *testing.T) {
 		customUserID             string
 		wantErr                  func(model.ValidationScheme) error
 		wantLabels               map[model.LabelName]model.LabelValue
+		wantLog                  []string
 	}{
 		{
 			name:                     "missing metric name",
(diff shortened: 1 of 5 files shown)


pkg/distributor/validate_test.go
```diff
@@ -600,6 +613,20 @@ func TestValidateLabels(t *testing.T) {
...
+								found := false
+								for _, line := range logged.Writes() {
+									if stringContainsAll(line, c.wantLog) {
+										found = true
+										break
+									}
+								}
+								if !found {
+									t.Log(logged.Writes())
+									require.Fail(t, "Expected logs to contain line with parts", "parts: %q", c.wantLog)
```
aknuds1, line 627: `require.Fail` doesn't do string formatting, right. I also don't think `%q` is the right formatting verb for a list.
```suggestion
require.Fail(t, "Expected logs to contain line with parts", "parts:", c.wantLog)
```

pkg/distributor/validate_test.go
```diff
@@ -1232,3 +1259,30 @@ func TestHashLabelValueInto(t *testing.T) {
...
+func (b *logRecorder) Write(buf []byte) (int, error) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	b.writes = append(b.writes, string(buf))
+	return len(buf), nil
+}
+
+func (b *logRecorder) Writes() []string {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
```
aknuds1, line 1277: [Nit]
```suggestion
	b.mtx.RLock()
	defer b.mtx.RUnlock()
```

pkg/distributor/validate_test.go
tcard: As is, the failure looks like this:

```
                	Error:      	Expected logs to contain line with parts
                	Test:       	TestValidateLabels/label_value_too_long_gets_dropped/legacy
                	Messages:   	parts: ["badLabelValueToDrop" "label value was replaced by its hash value" "insight=true"]
```

Your suggestion doesn't quite work:

```
                	Error:      	Expected logs to contain line with parts
                	Test:       	TestValidateLabels/label_value_too_long_gets_dropped/legacy
                	Messages:   	parts:%!(EXTRA []string=[badLabelValueToDrop label value was replaced by its hashoo value insight=true])
```

pkg/distributor/validate_test.go
tcard: In this case, there is not much point in read-locking. This lock has many writers per reader, so it's unlikely readers will try to get the lock concurrently.

PR #12848, opened 2025-09-29
MQE: Move out NameDrop functionality from DeduplicateAndMerge

Konstantinov-Innokentii:
#### What this PR does
This PR introduces DropName operator, responsible for name label drop when DelayedNameRemoval is enabled.

**Why?**
It's a prerequisite to support **DelayedNameRemoval** in the **EliminateDeduplicateAndMergeOptimizationPass**
With the introduction of **DelayedNameRemoval**, **DeduplicateAndMerge** became responsible for dropping the name label. However, it conflicts with the idea of skipping DeduplicateAndMerge ([PR](https://github.com/grafana/mimir/pull/12528)). There are cases when we might skip DeduplicateAndMerge, but still need to drop a name label, so we can't do it. So, solution is to split them.

**Example:**
For the expression `rate(foo[5m])` default plan is
```
- DeduplicateAndMerge
    - FunctionCall: rate(...)
        - MatrixSelector: {__name__="foo"}[5m0s]
```
selector `foo` guarantees that all series will be unique, so we can skip DeduplicateAndMerge. However, with DelayedNameRemoval enabled we can't eliminate it, because it's now dropping name.

After the change plan is:
```
- DeduplicateAndMerge
    - DropName
        - FunctionCall: rate(...)
           - MatrixSelector: {__name__="foo"}[5m0s]
```
And therefore DeduplicateAndMerge could be removed.

#### Checklist

- [x] Tests updated.
- ~[ ] Documentation added.~
- ~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`~, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.

### DIFF
--- a/pkg/streamingpromql/planning_test.go
+++ b/pkg/streamingpromql/planning_test.go
@@ -1116,7 +1116,7 @@ func TestPlanCreationEncodingAndDecoding(t *testing.T) {
 						NodeType:       planning.NODE_TYPE_DROP_NAME,
 						Details:        marshalDetails(&core.DropNameDetails{}),
 						Type:           "DropName",
-						Description:    "drop metric names",
+						Description:    "",
 						Children:       []int64{0},
 						ChildrenLabels: []string{""},
 					},


pkg/streamingpromql/planning.go
```diff
@@ -190,21 +190,10 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 		}
 
 		if p.enableDelayedNameRemoval {
```
charleskorn, line 211: The introduction of `NameDrop` is a breaking change, and our first breaking change since we merged https://github.com/grafana/mimir/pull/12808.

So if delayed name removal is enabled and we're going to add a `NameDrop` node, we should bump the query plan version to at least v1, if it's not already higher. 

And we should set `MaximumSupportedQueryPlanVersion` to 1 in this PR as well.

pkg/streamingpromql/operators/name_drop.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+
+	"github.com/prometheus/prometheus/promql/parser/posrange"
+
+	"github.com/grafana/mimir/pkg/streamingpromql/types"
+	"github.com/grafana/mimir/pkg/util/limiter"
+)
+
+type NameDrop struct {
```
charleskorn: [nit] `DropName` sounds more natural to me than `NameDrop` (similar to how `DeduplicateAndMerge` starts with a verb`

pkg/streamingpromql/planning/plan.go
```diff
@@ -22,10 +22,14 @@ import (
 	"github.com/grafana/mimir/pkg/util/limiter"
 )
 
-var MaximumSupportedQueryPlanVersion = int64(0)
+var MaximumSupportedQueryPlanVersion = int64(1)
 
 const QueryPlanVersionZero = int64(0)
 
+// Thins version introduces:
```
tcp13equals2: typo? `This`

pkg/streamingpromql/planning/plan.go
```diff
@@ -22,10 +22,14 @@ import (
...
+// 1. DropName node
+const QueryPlanVersionOne = int64(1)
```
tcp13equals2: I see a conflict on our future - https://github.com/grafana/mimir/pull/12931 ... we both have this :)

pkg/streamingpromql/planning/core/drop_name.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (n *DropName) EquivalentTo(other planning.Node) bool {
+	otherDropName, ok := other.(*DropName)
+
+	return ok && n.Inner.EquivalentTo(otherDropName.Inner)
+}
+
+func (n *DropName) Describe() string {
+	return "drop metric names"
```
charleskorn: There's no need for a description if the node name is sufficient to describe what it does.

pkg/streamingpromql/planning/plan.go
```diff
@@ -22,10 +22,14 @@ import (
 	"github.com/grafana/mimir/pkg/util/limiter"
 )
 
-var MaximumSupportedQueryPlanVersion = int64(0)
+var MaximumSupportedQueryPlanVersion = int64(1)
```
charleskorn:
```suggestion
var MaximumSupportedQueryPlanVersion = QueryPlanVersionOne
```

pkg/streamingpromql/planning/plan.go
charleskorn: Spelling out the version number (`One`, `Two` etc.) is going to get annoying when we get to higher-numbered versions (`OneHundredAndTwentyThree`). Perhaps we should just use digits?


```suggestion
const QueryPlanVersion0 = int64(0)

// This version introduces:
// 1. DropName node
const QueryPlanVersion1 = int64(1)
```

pkg/streamingpromql/planning/plan.proto
```diff
@@ -64,7 +64,8 @@ enum NodeType {
 
   NODE_TYPE_DUPLICATE = 10;
   NODE_TYPE_DEDUPLICATE_AND_MERGE = 11;
+  NODE_TYPE_DROP_NAME = 13;
 
   NODE_TYPE_REMOTE_EXEC = 12;
-  NODE_TYPE_TEST = 13;
+  NODE_TYPE_TEST = 14;
```
charleskorn: We can't renumber existing enum entries, as that's a backwards incompatible change. I would leave this as-is, and make `NODE_TYPE_DROP_NAME` have value 14.

Konstantinov-Innokentii: Yeah, fix already on the way :)

pkg/streamingpromql/planning/plan.go
Konstantinov-Innokentii: If we will merge them both this week, I think we can use same version?

Konstantinov-Innokentii: Makes sense

pkg/streamingpromql/planning/plan.go
Konstantinov-Innokentii: done

pkg/streamingpromql/planning/core/drop_name.go
Konstantinov-Innokentii: done

pkg/streamingpromql/planning/plan.go
Konstantinov-Innokentii: done

pkg/streamingpromql/operators/name_drop.go
Konstantinov-Innokentii: done

pkg/streamingpromql/planning.go
Konstantinov-Innokentii: done

pkg/streamingpromql/planning/plan.go
charleskorn: It's probably simplest to have different versions for each.

charleskorn (approved):

PR #12856, opened 2025-09-30
Remote execution: fix two cancellation-related issues

charleskorn:
#### What this PR does

This PR fixes two issues in the query-frontend:

##### Unnecessary request cancellation messages are sometimes sent to the query-scheduler

This PR fixes a race condition where queries evaluated with remote execution could sometimes result in a cancellation message being sent to the query-scheduler after the response had been completely read by query-frontends. 

This issue had no user-visible impact, but it causes confusing messages in request traces, and in some cases can cause unnecessary shuffling of queriers if shuffle sharding is enabled. (This happens because cancelling the request causes the query-scheduler to close the querier/query-scheduler stream, and if all streams for a single querier are closed before any are re-established, then the querier is removed from all shuffle shards and so reassignment of queriers to tenants occurs.)

Relevant commits: all commits except 84b2bd4

##### `ProtobufResponseStream.Next()` sometimes returns a `nil` message if the context passed to `DoProtobufRequest` is cancelled after at least some of the response is received from the querier

This PR fixes an issue where `Next` might sometimes return a `nil` message if the overall request context is cancelled after the query-frontend has started to receive the response from the querier and the caller has consumed all messages sent to the query-frontend. This can result in errors like those https://github.com/grafana/mimir/pull/12870 tries to mitigate.

Relevant commits: 84b2bd4

https://github.com/grafana/mimir/pull/12551

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/frontend/v2/frontend.go
```diff
@@ -399,25 +402,37 @@ func (f *Frontend) DoProtobufRequest(ctx context.Context, req proto.Message, min
...
+			// If we've already received some of the response, wait a short time for it to be completely read
+			// by receiveResultForProtobufRequest.
+			// This mitigates a race condition where the consumer reads the last message in the stream and
+			// calls Close before protobufResponseDone is closed by receiveResultForProtobufRequest.
+			// This would cause us to send an unnecessary cancellation request to the scheduler, which would
+			// then break the connection between the querier and the scheduler. If this happened to all of a
+			// querier's workers at the same time, it caused shuffle shard assignment churn.
+			case <-time.After(20 * time.Millisecond):
+				// Give up, fall through to below.
 			}
```
charleskorn: Note to reviewers: I don't love this, open to other more robust ideas.

charleskorn: Actually, I've had a better idea of how to handle this, I'll try this tomorrow.

pkg/frontend/v2/frontend.go
```diff
@@ -399,25 +402,37 @@ func (f *Frontend) DoProtobufRequest(ctx context.Context, req proto.Message, min
 		}
 
 		freq.spanLogger.DebugLog("msg", "request enqueued successfully, waiting for response")
+		<-ctx.Done() // The context will be cancelled if the request is cancelled by DoProtobufRequest's caller, if the response has been completely read, or if the stream is closed by the caller.
```
56quarters: It took me a while to understand how this worked. Is it possible to expand this comment a little bit?

56quarters:
> This PR mitigates a race condition where queries evaluated with remote execution could sometimes result in a cancellation message being sent to the query-scheduler after the response had been completely read by queriers.

I'm not clear on which components are being described here. A cancellation message being sent from query-frontends to query-schedulers after the response(?) had been read by queriers? Which response are queriers reading?

pkg/frontend/v2/frontend_test.go
```diff
@@ -228,6 +226,50 @@ func TestFrontend_Protobuf_HappyPath(t *testing.T) {
 	require.Nil(t, msg)
 }
 
+func TestFrontend_Protobuf_ShouldNotCancelRequestAfterSuccess(t *testing.T) {
```
56quarters, line 243: Can you add a comment here explaining what exactly this test is doing?

56quarters (commented):
Changes seem fine if a little hard to understand at first pass (not new, this whole flow is complicated). I'll take another pass after you make the change you alluded to.

pkg/frontend/v2/frontend_test.go
charleskorn: How's what I've added in bb6e838?

charleskorn:
> > This PR mitigates a race condition where queries evaluated with remote execution could sometimes result in a cancellation message being sent to the query-scheduler after the response had been completely read by queriers.
> 
> I'm not clear on which components are being described here. A cancellation message being sent from query-frontends to query-schedulers after the response(?) had been read by queriers? Which response are queriers reading?

Sorry, this had a typo in it, it should be:

This PR mitigates a race condition where queries evaluated with remote execution could sometimes result in a cancellation message being sent to the query-scheduler after the response had been completely read by ~~queriers~~ **query-frontends**.

pkg/frontend/v2/frontend.go
charleskorn: How's this after my most recent changes?

pkg/frontend/v2/frontend_test.go
```diff
@@ -979,6 +979,66 @@ func TestFrontend_Protobuf_ReadingCancelledRequest(t *testing.T) {
 	require.Nil(t, msg)
 }
 
+func TestFrontend_Protobuf_ReadingCancelledRequestAfterResponseReceivedFromQuerier(t *testing.T) {
```
56quarters, line 982: This is a surprisingly straightforward test for now subtle this issue is :partying_face:

pkg/frontend/v2/frontend.go
```diff
@@ -357,28 +358,32 @@ func (f *Frontend) RoundTripGRPC(ctx context.Context, httpRequest *httpgrpc.HTTP
...
 // minT and maxT should be the start and end time of the queried data.
 // These timestamps should consider the lookback delta (ie. are not necessarily the time range provided in the query request).
-func (f *Frontend) DoProtobufRequest(ctx context.Context, req proto.Message, minT, maxT time.Time) (*ProtobufResponseStream, error) {
-	logger, ctx := spanlogger.New(ctx, f.log, tracer, "frontend.DoProtobufRequest")
+func (f *Frontend) DoProtobufRequest(requestContext context.Context, req proto.Message, minT, maxT time.Time) (*ProtobufResponseStream, error) {
+	logger, requestContext := spanlogger.New(requestContext, f.log, tracer, "frontend.DoProtobufRequest")
 	logger.SetTag("request.type", proto.MessageName(req))
 
-	freq, ctx, cancel, err := f.createNewRequest(ctx)
+	freq, streamContext, cancel, err := f.createNewRequest(requestContext)
```
56quarters:
```suggestion
	freq, streamContext, cancelStream, err := f.createNewRequest(requestContext)
```

56quarters (approved):
This LGTM. Differentiating between `requestContext`, `streamContext` and the context given to `ProtobufResponseStream.Next()` really helped me understand what's going on here. Given the complexity here maybe @francoposa or @chencs could take a look? If they can't that's fine too.

francoposa:
@56quarters afk today but we never actually dealt with this end of the code for the scheduler project. Might be someone else that worked in code recently that would have a better opinion.

pkg/frontend/v2/frontend.go
charleskorn: Good idea, changed in 5417fe6

charleskorn:
> Given the complexity here maybe @francoposa or @chencs could take a look? If they can't that's fine too.

Given your approval and the fact this is behind a feature flag, and the queue of other PRs waiting on this one, I'm going to merge this so I can start testing it in some test environments, but I'd love to hear any further feedback after I merge.

PR #12859, opened 2025-09-30
Split deploy preview to reduce confusion about targeted versions

jdbaldry:
The current deploy preview workflow tries to accurately mimic the relationship between the Helm chart documentation and the Mimir documentation.

This causes confusion because most PRs just modify the Mimir docs and the deploy preview mounts those docs at a specific version rather than "next" where they're actually published.

I've split the workflow into two with triggers for each directory of documentation.


armandgrillet (approved):
LGTM

jdbaldry:
Don't have merge permissions so feel free to include this when you're happy :)

.github/workflows/deploy-pr-preview-helm-charts.yml
narqo, line 34: Do you still want the changes in Helm chart's docs to build mimir's docs? I also see that the new workflow below uses the `/next` version. Shouldn't we use the same for the chart here?

narqo (commented):
How will these two workflows work with the "Preview deployed" comment, that the upstream's `deploy-preview.yml` workflow produces? Won't they override each other?

.github/workflows/deploy-pr-preview-helm-charts.yml
jdbaldry: IIRC, the reason for building both together is because the Helm chart docs link extensively to the Mimir docs.

I do think `next` could be less confusing as a mount destination for the Helm charts docs in the deploy preview. I'll make that change

jdbaldry:
> How will these two workflows work with the "Preview deployed" comment, that the upstream's `deploy-preview.yml` workflow produces? Won't they override each other?

Good spot. I've updated the workflow to support separate comments and workflow runs using the `title` input (previously unused) as a distinguishing element and updated the titles of these two workflows so they're distinct.

PR #12861, opened 2025-09-30
parquet converter: new parquet-converter.min-data-age flag

npazosmendez:
Even more knobs for conditional conversions.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -22739,7 +22739,7 @@
           "kind": "field",
           "name": "min_data_age",
           "required": false,
-          "desc": "Minimum age of data in blocks to convert. Only convert blocks containing data older than this duration from now. Set to 0 to disable age filtering.",
+          "desc": "Minimum age of data in blocks to convert. Only convert blocks containing data older than this duration from now, based on their MinTime. Set to 0 to disable age filtering.",
           "fieldValue": null,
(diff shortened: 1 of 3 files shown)


pkg/parquetconverter/parquet_converter.go
```diff
@@ -459,6 +461,15 @@ func (c *ParquetConverter) discoverAndEnqueueBlocks(ctx context.Context) {
 				}
 			}
 
+			if c.Cfg.MinDataAge > 0 {
+				blockMinTime := time.UnixMilli(m.MinTime)
+				cutoffTime := time.Now().Add(-c.Cfg.MinDataAge)
+				if blockMinTime.After(cutoffTime) {
```
seizethedave, line 467: Is this backwards? We want to keep the ones that are after the cutoff?

npazosmendez: We want to keep the ones _before_ that cutoff (i.e. older than the specified duration) - if it's _after_, the data is too new for our criteria. 

These are always confusing to reason about though, so maybe it's time we add some tests :grin:

seizethedave: ah yeah. the old "smaller time has bigger difference from now."

npazosmendez: I commit to adding tests in a followup PR

PR #12877, opened 2025-10-01
parquet converter: test and refactor block conversion skip logic

npazosmendez:
Refactors the skipping logic out to make it testable. Ideally we'd test it end to end but that requires a more complex setup, for a code that will probably be discarded soon. For now this is enough.

Additionally, I'm removing the `parquet-converter.max-queue-size` flag. As it is now the flag does not fully limit the queue size, as the check is only done at the top of the discovery routine. Because we were leaving the default value of 5, this resulted in no blocks being added before the worker got to (roughly) the end of the queue, which skewed the real queue size for monitoring.  Also, with the current logic, it doesn't help with its original purpose of distributing load, probably because some things changed since it was added.

### DIFF
--- a/cmd/mimir/config-descriptor.json
+++ b/cmd/mimir/config-descriptor.json
@@ -22813,17 +22813,6 @@
           "fieldCategory": "advanced"
         },
         {
-          "kind": "field",
-          "name": "max_queue_size",
-          "required": false,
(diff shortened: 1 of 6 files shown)


tacole02 (approved):
Docs look good! Thank you!

seizethedave (approved):
Looks good to me. ™️

PR #12878, opened 2025-10-01
Ensure valid MemoryConsumptionTracker in context

lamida:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/util/limiter/memory_consumption.go
```diff
@@ -20,13 +20,12 @@ const (
...
-// does not enforce any limits is returned.
-func MemoryTrackerFromContextWithFallback(ctx context.Context) *MemoryConsumptionTracker {
-	tracker, ok := ctx.Value(memoryConsumptionTracker).(*MemoryConsumptionTracker)
+// MustMemoryTrackerFromContext returns a MemoryConsumptionOperation that has been added to this
+// context. If there is no MemoryConsumptionOperation in this context, this will panic.
+func MustMemoryTrackerFromContext(ctx context.Context) MemoryConsumptionOperation {
+	tracker, ok := ctx.Value(memoryConsumptionTracker).(MemoryConsumptionOperation)
 	if !ok {
-		return NewMemoryConsumptionTracker(ctx, 0, nil, "")
+		panic("memory tracker not found in context")
```
56quarters: Is there anything we could add here to help debugging?

pkg/util/limiter/noop_memory_consumption_promql_engine.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (n noopMemoryConsumptionTracker) DecreaseMemoryConsumptionForLabels(_ labels.Labels) {}
+
+// NoopMemoryTrackerPromqlEngine wrap promql.Engine so that this can
+// pass noopMemoryConsumptionTracker which will do nothing when processing query.
+//
+// We will have MemoryConsumptionTracker when processing response from ingester and store gateway,
+// but we only want to track memory consumption for query coming from streamingpromql.Engine.
+type NoopMemoryTrackerPromqlEngine struct {
```
56quarters: This shouldn't live in this package. Since this will be used via interface everywhere it can live in `pkg/mimir` or elsewhere but it shouldn't be in `limiter`.

56quarters (commented):
~I'm a little worried about introducing a source of panics into the codebase. Do we have instances of memory tracking being wrong / missing with the current way it works?~ NM, needed to read the PR description more closely

integration/e2emimir/client.go
charleskorn, line 1: Why are the changes in this file needed? `c.querierClient` makes HTTP calls to the API, so the presence (or lack of) a memory consumption tracker should make no difference.

pkg/distributor/query_ingest_storage_test.go
```diff
@@ -20,12 +20,15 @@ import (
...
 	"github.com/grafana/mimir/pkg/querier/stats"
+	"github.com/grafana/mimir/pkg/util/limiter"
 )
 
 func TestDistributor_QueryStream_ShouldSupportIngestStorage(t *testing.T) {
 	const tenantID = "user"
 
 	ctx := user.InjectOrgID(context.Background(), tenantID)
+	memoryTracker := limiter.NewMemoryConsumptionTracker(ctx, 0, nil, "")
+	ctx = limiter.AddMemoryTrackerToContext(ctx, memoryTracker)
```
charleskorn: [nit] Might be useful to introduce a `limiter.ContextWithNewMemoryConsumptionTracker()` helper method that does `NewMemoryConsumptionTracker` + `AddMemoryTrackerToContext` to make it simpler to use in tests like this

pkg/util/limiter/memory_consumption.go
charleskorn: What if this returned an error rather than panicking?

pkg/util/limiter/memory_consumption.go
```diff
@@ -111,6 +110,13 @@ func (s MemoryConsumptionSource) String() string {
 	}
 }
 
+type MemoryConsumptionOperation interface {
```
charleskorn: Why is this interface needed?

pkg/util/limiter/noop_memory_consumption_promql_engine.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/storage"
+)
+
+type noopMemoryConsumptionTracker struct{}
+
+func NewNoopMemoryConsumptionTracker() MemoryConsumptionOperation {
```
charleskorn: It's not clear to me why we need this - can't we create a `MemoryConsumptionTracker` with no limit instead?

pkg/streamingpromql/benchmarks/comparison_test.go
```diff
@@ -105,6 +106,8 @@ func TestBothEnginesReturnSameResultsForBenchmarkQueries(t *testing.T) {
 	require.NoError(t, err)
 
 	ctx := user.InjectOrgID(context.Background(), UserID)
+	memoryTracker := limiter.NewMemoryConsumptionTracker(ctx, 0, nil, "")
+	ctx = limiter.AddMemoryTrackerToContext(ctx, memoryTracker)
```
charleskorn: This is only needed for Prometheus' engine - should we perhaps only pass it when calling `c.Run` with Prometheus' engine?

We should also create this new every time inside `t.Run` below, to avoid cross-contamination between test cases.

pkg/util/limiter/memory_consumption.go
lamida: Initially I thought we need another implementation of noopMemoryConsumptionTracker that does nothing, hence I added this interface. But now I think we can just use "unlimited" memory consumption tracker, hence we can remove this interface.

pkg/util/limiter/noop_memory_consumption_promql_engine.go
lamida: I was thinking to make the thing more efficient for not tracking at all. But for now, I follow your suggestion to just use "unlimited" MemoryConsumptionTracker.

pkg/streamingpromql/benchmarks/comparison_test.go
lamida: Make sense. Updated in 59f98f0ae0ed9f4fd5c2b7d49e3ae6ae0e99bd28

pkg/distributor/query_ingest_storage_test.go
lamida: Updated in the recent commits

pkg/util/limiter/noop_memory_consumption_promql_engine.go
lamida: This is now renamed to `UnlimitedMemoryTrackerPromqlEngine` in file `memory_consumption_promql_engine_compat.go` and it is moved to pkg/mimir as suggested.

pkg/util/limiter/memory_consumption.go
lamida: If it we return err, what kind of error handling that we can do? We can initiate a new memory tracker or unlimited memory tracker in handling the error. But that makes the inconsistent memory tracking might happen due to multiple instances of memory tracker across the request call chain which is the exact issue we want to prevent in this PR. Or is there any other error handling option?

pkg/util/limiter/noop_memory_consumption_promql_engine.go
lamida: I moved this back to pkg/util/limiter. When I put in pkg/mimir, I got an import cycle. I think the limiter package can be just the good place for now since it just initiate memory tracker in context for query via promql.Engine. 

Also the consumer of `NewUnlimitedMemoryTrackerPromqlEngine` are only in pkg/mimir and pkg/querier.

pkg/util/limiter/memory_consumption.go
charleskorn: If we return an error, then the caller should give up and return the error, as it would for any other kind of error. The request will still fail, but at least the whole process won't crash.

pkg/streamingpromql/benchmarks/comparison_test.go
```diff
@@ -111,7 +112,7 @@ func TestBothEnginesReturnSameResultsForBenchmarkQueries(t *testing.T) {
 			start := time.Unix(int64((NumIntervals-c.Steps)*intervalSeconds), 0)
 			end := time.Unix(int64(NumIntervals*intervalSeconds), 0)
 
-			prometheusResult, prometheusClose := c.Run(ctx, t, start, end, interval, prometheusEngine, q)
+			prometheusResult, prometheusClose := c.Run(limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(ctx), t, start, end, interval, prometheusEngine, q)
```
charleskorn: Could we use a `UnlimitedMemoryTrackerPromqlEngine` instance here (and in all other tests that create an instance of Prometheus' engine in `pkg/streamingpromql`)? Then we don't need to do this.

pkg/util/limiter/memory_consumption.go
lamida: Ok I think that make sense. Updated in 3b55d7afc035a7fa4d97b2165b985da6b799d1d1.

pkg/streamingpromql/benchmarks/comparison_test.go
lamida: I meant to use `UnlimitedMemoryTrackerPromqlEngine` for engine but instead used `ContextWithNewUnlimitedMemoryConsumptionTracker` in context. 

Fixed in 711fbdde234d26fed1de9fe94589ebd1105bdadc

For other test, if they don't touch both ingester/store-gateway response layer and streamingpromql operator (which only applies in the labels memory tracking case), I think we can just leave the prometheusEngine used in non-breaking test as it is.

integration/e2emimir/client.go
lamida: You are right. I didn't fully understood before on how context won't be passed by default from http client to server, unless if use something like `NewRequestWithContext`. 

I removed the change in this file.

pkg/streamingpromql/benchmarks/comparison_test.go
lamida: Updated. But I will not wrap other occurrences of `promql.Engine` with `UnlimitedMemoryTrackerPromqlEngine`, unless if somewhere during test execution it needs access to MemoryConsumptionTracker.

pkg/util/limiter/memory_consumption.go
lamida: As suggested in another comment, we replace panic with error return.

pkg/querier/blocks_store_queryable.go
```diff
@@ -467,7 +467,11 @@ func (q *blocksStoreQuerier) selectSorted(ctx context.Context, sp *storage.Selec
 	}
 
 	queryF := func(clients map[BlocksStoreClient][]ulid.ULID, minT, maxT int64) ([]ulid.ULID, error) {
-		seriesSets, queriedBlocks, warnings, streamReaders, chunkEstimator, err := q.fetchSeriesFromStores(ctx, sp, clients, minT, maxT, tenantID, convertedMatchers)
+		memoryTracker, err := limiter.MemoryConsumptionTrackerFromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
```
charleskorn: Should we move this outside `queryF`? Given we use the same context for all `queryF` calls, there's no need to run this for every `queryF` invocation.

pkg/util/limiter/memory_consumption.go
```diff
@@ -20,16 +21,17 @@ const (
 	memoryConsumptionTracker contextKey = 0
 )
 
-// MemoryTrackerFromContextWithFallback returns a MemoryConsumptionTracker that has been added to this
-// context. If there is no MemoryConsumptionTracker in this context, a new no-op tracker that
-// does not enforce any limits is returned.
-func MemoryTrackerFromContextWithFallback(ctx context.Context) *MemoryConsumptionTracker {
+var ErrNoMemoryConsumptionTrackerInContext = errors.New("no memory consumption tracker in context")
```
charleskorn: Is there a reason why this is exported?
```suggestion
var errNoMemoryConsumptionTrackerInContext = errors.New("no memory consumption tracker in context")
```

pkg/util/limiter/memory_consumption_test.go
```diff
@@ -19,12 +19,12 @@ import (
...
-func TestFromContextWithFallback(t *testing.T) {
+func TestMemoryConsumptionTrackerFromContext(t *testing.T) {
 	t.Run("does not exist", func(t *testing.T) {
 		ctx := context.Background()
-		tracker := MemoryTrackerFromContextWithFallback(ctx)
-		require.NotNil(t, tracker)
-		require.Equal(t, uint64(0), tracker.CurrentEstimatedMemoryConsumptionBytes())
+		tracker, err := MemoryConsumptionTrackerFromContext(ctx)
+		require.Nil(t, tracker)
+		require.ErrorAs(t, err, &ErrNoMemoryConsumptionTrackerInContext)
```
charleskorn: There's no need for `As` here, an `Is` check is sufficient and slightly clearer:

```suggestion
		require.ErrorIs(t, err, ErrNoMemoryConsumptionTrackerInContext)
```

pkg/streamingpromql/benchmarks/comparison_test.go
```diff
@@ -111,7 +112,7 @@ func TestBothEnginesReturnSameResultsForBenchmarkQueries(t *testing.T) {
 			start := time.Unix(int64((NumIntervals-c.Steps)*intervalSeconds), 0)
 			end := time.Unix(int64(NumIntervals*intervalSeconds), 0)
 
-			prometheusResult, prometheusClose := c.Run(ctx, t, start, end, interval, prometheusEngine, q)
+			prometheusResult, prometheusClose := c.Run(ctx, t, start, end, interval, limiter.NewUnlimitedMemoryTrackerPromqlEngine(prometheusEngine), q)
```
charleskorn: Why do we need to wrap `prometheusEngine` here? Can't we wrap `prometheusEngine` when it is created on line 100 above?

pkg/util/limiter/memory_consumption_promql_engine_compat.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// UnlimitedMemoryTrackingQuery wraps promql.Query so that this can
+// pass an unlimited MemoryConsumptionTracker, specifically in the Exec call.
+type UnlimitedMemoryTrackingQuery struct {
+	promql.Query
+}
+
+func (u *UnlimitedMemoryTrackingQuery) Exec(ctx context.Context) *promql.Result {
+	ctx = ContextWithNewUnlimitedMemoryConsumptionTracker(ctx)
```
charleskorn: Rather than wrapping just the `Exec` call like this, I think it'd be safer to create the memory consumption tracker in `NewInstantQuery` or `NewRangeQuery` below and store it in `UnlimitedMemoryTrackingQuery`. Then in `NewInstantQuery`, `NewRangeQuery` and `Exec`, we can wrap the passed context with the memory consumption tracker.

This ensures that creating the query also gets the same memory consumption tracker.

pkg/util/limiter/memory_consumption_promql_engine_compat.go
```diff
@@ -0,0 +1,51 @@
...
+	"time"
+
+	"github.com/prometheus/prometheus/promql"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// UnlimitedMemoryTrackingQuery wraps promql.Query so that this can
+// pass an unlimited MemoryConsumptionTracker, specifically in the Exec call.
+type UnlimitedMemoryTrackingQuery struct {
+	promql.Query
```
charleskorn: Rather than embedding the `Query` instance like this, I'd prefer we store it as a field and explicitly implement the other `Query` interface methods. This way we will be forced to consider if we need to do anything else if this interface is expanded in the future.

pkg/util/limiter/memory_consumption_promql_engine_compat.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx = ContextWithNewUnlimitedMemoryConsumptionTracker(ctx)
+	return u.Query.Exec(ctx)
+}
+
+func NewUnlimitedMemoryTrackingQuery(query promql.Query) *UnlimitedMemoryTrackingQuery {
+	return &UnlimitedMemoryTrackingQuery{query}
+}
+
+// UnlimitedMemoryTrackerPromqlEngine wraps promql.Engine so that it can have promql.Query that can pass
+// MemoryConsumptionTracker during the Exec call.
```
charleskorn: This comment doesn't really explain why we want to do this - perhaps something like this is clearer?

```suggestion
// UnlimitedMemoryTrackerPromqlEngine wraps promql.Engine so that queries always run
// with a MemoryConsumptionTracker in their context.
// 
// The provided MemoryConsumptionTracker will enforce no limit on query memory consumption,
// but is still required as some querier components track memory consumption even
// when running in Prometheus' engine (eg. the ingester and store-gateway Queryable
// implementations).
```

pkg/util/limiter/memory_consumption_promql_engine_compat.go
```diff
@@ -0,0 +1,51 @@
...
+type UnlimitedMemoryTrackerPromqlEngine struct {
```
charleskorn: [nit] "Promql" isn't one word:

```suggestion
type UnlimitedMemoryTrackerPromQLEngine struct {
```

pkg/ruler/compat.go
```diff
@@ -408,11 +409,15 @@ func DefaultTenantManagerFactory(
 			appendeable = NewNoopAppendable()
 		}
 
+		// Add memory tracker to context for ruler queries (both QueryFunc and direct Queryable.Select calls)
+		ruleManagerCtx := user.InjectOrgID(ctx, userID)
+		ruleManagerCtx = limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(ruleManagerCtx)
```
charleskorn: I don't think this is the right place for this, as this means that all rule evaluations will share the same memory consumption tracker. Perhaps rulers should be using a `UnlimitedMemoryTrackerPromqlEngine` instance if they're using Promtheus' query engine? (The engine instance seems to come from `querier.New()` in `pkg/querier/querier.go`, which is already wrapping the engine instance in `UnlimitedMemoryTrackerPromqlEngine`, so perhaps this change here is not needed at all?)

pkg/api/handlers.go
```diff
@@ -286,6 +287,15 @@ func NewQuerierHandler(
...
 	api.Register(promRouter)
 
+	// Helper function to wrap handlers with memory tracker middleware for API endpoints
+	// that access the queryable directly (like /series) without going through PromQL engine.
+	withMemoryTracker := func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(r.Context())
+			handler.ServeHTTP(w, r.WithContext(ctx))
+		})
+	}
```
charleskorn: Could this be done as a middleware instead? We already follow the middleware pattern for a couple of other concerns in `initQuerier()`, so I'm wondering if we can do something similar.

pkg/querier/blocks_store_queryable.go
lamida: That's true. Previously I was trying to think how to return the error in selectSorted method. But apparently we can just return err wrapped with `storage.ErrSeriesSet`. With the same context, we can just retrieve the tracker once, outside of queryF invocation as you suggested. Updated in 07980bde9a793adc71e08e01210c4dc5c7b72858.

pkg/ruler/compat.go
lamida: > as this means that all rule evaluations will share the same memory consumption tracker

Yes, we shouldn't have single memory tracker for all rule evaluations. I am not familiar yet with ruler evaluation codes to put this correctly. But some ruler integration is failing without this. Let me see the code again to make sure putting it in the right place.

pkg/streamingpromql/benchmarks/comparison_test.go
lamida: Updated in 561aea60c3b43c7d01ff9071a9f55cc0be454301

pkg/util/limiter/memory_consumption.go
lamida: No reason. I was just thinking it might be needed outside of the package but it isn't. Hence we can make it private 9488d6850ebdf5b280527b92db6fb259901a867f.

pkg/util/limiter/memory_consumption_promql_engine_compat.go
lamida: Valid point. 52db07c81f5beb6725becbbe33e366d8f15ea968

pkg/util/limiter/memory_consumption_promql_engine_compat.go
lamida: Updated in f3598835619a18b10cfa508c6ce221f62598d982

pkg/api/handlers.go
lamida: Refactored in acd02309f85c3b71b9d2caf1b56607df7af7e72c. Let me know if this is what you you are envisioning

pkg/util/limiter/memory_consumption_promql_engine_compat.go
lamida: Should be updated in https://github.com/grafana/mimir/commit/52db07c81f5beb6725becbbe33e366d8f15ea968

pkg/util/limiter/memory_consumption_test.go
lamida: Resolved in ce6d2e26b7733aa99417d5d616a9140763eb6eff

pkg/api/handlers.go
```diff
@@ -305,7 +306,7 @@ func NewQuerierHandler(
 	router.Path(path.Join(promPrefix, "/query_exemplars")).Methods("GET", "POST").Handler(exemplarsQueryStats.Wrap(promRouter))
 	router.Path(path.Join(promPrefix, "/labels")).Methods("GET", "POST").Handler(labelsQueryStats.Wrap(promRouter))
 	router.Path(path.Join(promPrefix, "/label/{name}/values")).Methods("GET").Handler(labelsQueryStats.Wrap(promRouter))
-	router.Path(path.Join(promPrefix, "/series")).Methods("GET", "POST", "DELETE").Handler(seriesQueryStats.Wrap(promRouter))
+	router.Path(path.Join(promPrefix, "/series")).Methods("GET", "POST", "DELETE").Handler(limiter.MemoryTrackerMiddleware{}.Wrap(seriesQueryStats.Wrap(promRouter)))
```
charleskorn: Is there a reason why we need to do this here rather than in `pkg/frontend/querymiddleware/roundtrip.go`'s `newQueryTripperware`?

If not, I'd suggest moving this to that file and moving the middleware to the `querymiddleware` package as well.

lamida: The reason, it is simpler :D. I tried to move it to roundtrip. Do we have to convert the middleware to roundtripper? I can't make it work by trying this
```
func (m memoryTrackerRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
	ctx := limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(r.Context())
	r = r.WithContext(ctx)
	return m.next.RoundTrip(r)
}
```

The memory tracker seems not propagated when reaching the Queryable. 🤔 

And passing it in the roundtripper

```
			series = newRetryRoundTripper(memoryTrackerRoundTripper{series}, log, cfg.MaxRetries, retryMetrics)
```

lamida: I will keep trying to fix that.

lamida: Ok after I checked again, I think the reason why we need the middleware wrapper just in `pkg/api/handlers.go` is because `newQueryTripperware` in `pkg/frontend/querymiddleware/roundtrip.go` is called only for query frontend. As the call reaches querier, we need a way to initiate memoryConsumptionTracker again. Hence I believe we must keep the handler as it is.

pkg/querier/memory_tracker_queryable.go
```diff
@@ -0,0 +1,109 @@
...
+
+// memoryTrackerQuerier wraps a storage.Querier and ensures all Select calls have an unlimited
+// MemoryConsumptionTracker in their context.
+type memoryTrackerQuerier struct {
+	inner storage.Querier
+}
+
+func (q *memoryTrackerQuerier) Select(ctx context.Context, sortSeries bool, hints *storage.SelectHints, matchers ...*labels.Matcher) storage.SeriesSet {
+	if _, err := limiter.MemoryConsumptionTrackerFromContext(ctx); err != nil {
+		ctx = limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(ctx)
```
charleskorn: Could you please add a comment here explaining why we don't unconditionally add an unlimited memory consumption tracker?

pkg/querier/memory_tracker_queryable.go
```diff
@@ -0,0 +1,109 @@
...
+// memoryTrackerSampleAndChunkQueryable wraps a storage.SampleAndChunkQueryable and ensures all queries have an unlimited
+// MemoryConsumptionTracker in their context.
+type memoryTrackerSampleAndChunkQueryable struct {
+	inner storage.SampleAndChunkQueryable
+}
+
+// NewMemoryTrackerSampleAndChunkQueryable wraps a storage.Queryable to inject an unlimited MemoryConsumptionTracker
+// into the context for all queriers.
+func NewMemoryTrackerSampleAndChunkQueryable(inner storage.Queryable) storage.Queryable {
+	// Try to cast to SampleAndChunkQueryable to preserve the interface
```
charleskorn: Why do we need to support both types? Can we just support one?

pkg/util/limiter/memory_tracker_middleware.go
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package limiter
+
+import (
+	"net/http"
+)
+
+// MemoryTrackerMiddleware adds an unlimited MemoryConsumptionTracker to the request context.
+type MemoryTrackerMiddleware struct{}
```
charleskorn: [nit] I'd make it super clear that this will not enforce a memory consumption limit:

```suggestion
type UnlimitedMemoryTrackerMiddleware struct{}
```

pkg/util/limiter/unlimited_memory_consumption_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// The provided MemoryConsumptionTracker will enforce no limit on query memory consumption,
+// but is still required as some querier components track memory consumption even
+// when running in Prometheus' engine (eg. the ingester and store-gateway Queryable
+// implementations).
+type UnlimitedMemoryTrackerPromQLEngine struct {
+	inner *promql.Engine
+}
+
+func NewUnlimitedMemoryTrackerPromqlEngine(inner *promql.Engine) UnlimitedMemoryTrackerPromQLEngine {
```
charleskorn: [nit] To match the type name:

```suggestion
func NewUnlimitedMemoryTrackerPromQLEngine(inner *promql.Engine) UnlimitedMemoryTrackerPromQLEngine {
```

pkg/mimir/modules.go
```diff
@@ -1050,6 +1051,10 @@ func (t *Mimir) initRuler() (serv services.Service, err error) {
 			return nil, fmt.Errorf("could not create queryable for ruler: %w", err)
 		}
 
+		// Wrap the queryable to ensure all queries have a memory tracker in the context.
+		// This is needed for ruler operations like alert state restoration that may bypass
+		// the query engine's NewInstantQuery/NewRangeQuery methods.
```
charleskorn: Could you please expand the comment here to explain that the `Queryable` won't inject a new memory consumption tracker if there's already one on the context?

pkg/api/handlers.go
```diff
@@ -297,15 +298,18 @@ func NewQuerierHandler(
 	cardinalityQueryStats := usagestats.NewRequestsMiddleware("querier_cardinality_query_requests")
 	formattingQueryStats := usagestats.NewRequestsMiddleware("querier_formatting_requests")
 
+	// Add memory consumption tracker in middleware for endpoints like /read and /series that need memory consumption tracker to be available in context.
```
charleskorn: Could you please expand the comment here to explain why we need a memory consumption tracker, and why we can't do this somewhere else?

pkg/querier/memory_tracker_queryable.go
lamida: Yes actually for ruler use case, it needs only Queryable. Removed the SampleAndChunkQueryable implementation in 2782a5c3a7ac6d10c0237ab5de824d563f62355f

pkg/querier/memory_tracker_queryable.go
lamida: Actually I think we don't need this conditional injection. We started this tracker from the place that we know has no memory tracker, from initRuler.

Removed the conditional injection in 38f02e1e1048b9d4cd5feadf26cf82cd5a486a1a

Let me know if you oversees that the conditional injection is actually necessary.

pkg/mimir/modules.go
lamida: In this [comment](https://github.com/grafana/mimir/pull/12878#discussion_r2410257618) I mentioned that maybe we don't need conditional memory tracking injection and just create a new memory tracking when running `Select`.  We know for ruler case, the queryable will be using this wrapper `unlimitedMemoryTrackerQueryable` which will return `unlimitedMemoryTrackerQuerier` which will inject memoryConsumptionTracker everytime a new query is executed by ruler, which is when running `Select`.

Do you think I missed something important here?

pkg/querier/memory_tracker_queryable.go
charleskorn: If the ruler is running without a query-frontend (ie. is evaluating queries inside the ruler process, rather than delegating to a query-frontend), then won't it use this `Queryable` / `Querier` with the MQE instance evaluating queries?

pkg/mimir/modules.go
charleskorn: I think we need to address https://github.com/grafana/mimir/pull/12878#discussion_r2412137491 first before we can resolve this.

pkg/querier/memory_tracker_queryable.go
lamida: Let me try to clarify what I was trying to say

> Actually I think we don't need this conditional injection.

What I mean  we don't need this if check
```
	if _, err := limiter.MemoryConsumptionTrackerFromContext(ctx); err != nil {
```

But we still have to add 
```
ctx = limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(ctx)
return q.inner.Select(ctx, sortSeries, hints, matchers...)
```

The `if` check is not needed, because `unlimitedMemoryTrackerQuerier` is used only for ruler in [internal operation mode](https://grafana.com/docs/mimir/latest/references/architecture/components/ruler/#internal) and not in other places. For ruler with [remote operation mode](https://grafana.com/docs/mimir/latest/references/architecture/components/ruler/#remote), the query-frontend and then the querier will handle the part it initiate the unlimitedMemoryTracker.

Please let me know if the clarification still missing point you are trying to ask.

charleskorn: If the ruler is running in internal operation mode, doesn't it still use a query engine (which could be MQE) to evaluate queries?

lamida: Ok I got your point now. There are many indirection there which is not immediately obvious to me. 

Interestingly, for internal mode, actually we might not need adding the tracker here at all. Because in internal operation mode, ruler will get the memory tracker from either `UnlimitedMemoryTrackingQuery.Exec` for promql or from `Query.Exec` for MQE. 

Now, I am trying to find what code path from ruler that don't have the tracker initialized in context because if we just remove `ctx = limiter.ContextWithNewUnlimitedMemoryConsumptionTracker(ctx)`, at least `TestRuler_RestoreWithLongForPeriod` is failing. Let me check this.

charleskorn: Restoration of alert state (which is what that test is testing) doesn't use a query engine as far as I know. Perhaps we can modify it to use a context with an unlimited memory consumption tracker, and then we don't need this `Querier` implementation at all?

lamida: I made the change in 08c2bde92f0fe2edf35eea75fe6d5c14eb4fa4d7 should cover the points in this discussion.

* We still need `unlimitedMemoryTrackerQueryable`.
* But we will use specifically for ruler state restoration, hence the wrapping now is moved to ruler/compat.go and memoryTracker injection will happen for this specific use case only.
* Moving the wrapping there is better because now we really don't need the `	if _, err := limiter.MemoryConsumptionTrackerFromContext(ctx); err != nil` check anymore.
* Other ruler operations, will use `UnlimitedMemoryTrackerPromQLEngine` to inject unlimited memory tracker in the context.

lamida:
@charleskorn I have updated the PR description with the current state of the implementation. Please help to review again.

pkg/ruler/compat.go
```diff
@@ -400,6 +400,10 @@ func DefaultTenantManagerFactory(
 
 		// Wrap the queryable with our custom logic.
```
charleskorn, line 401: Took me a second to realise this queryable is not used for rule evaluation because `wrappedQueryFunc` uses a different queryable (and therefore the unlimited memory consumption limit doesn't apply to rule evaluation, only alert state restoration) - could you please add a comment here to explain that this is the intended behaviour?

lamida: I added more comments several lines above dbd0f840d36c6432b45b5924797f04c784182bce to give more details for both. Let me know if that is enough.

pkg/ruler/compat.go
```diff
@@ -400,6 +406,10 @@ func DefaultTenantManagerFactory(
...
 		wrappedQueryable := WrapQueryableWithReadConsistency(queryable, logger)
+		// Wrap the queryable to ensure all queries have a memory tracker in the context.
+		// This is needed for ruler operations, specifically alert state restoration that is not using
+		// the query engine's NewInstantQuery/NewRangeQuery methods from UnlimitedMemoryTrackerPromQLEngine.
```
charleskorn: I think the comment above and this comment can be summarised as this:

```suggestion
		// Wrap the queryable to ensure all queries have a memory tracker in the context.
		// This is needed for ruler operations, specifically alert state restoration that does not use a query engine.
		// (Operations that use a query engine use QueryFunc, and it's expected that
		// the QueryFunc either uses MQE, or uses a Prometheus engine instance wrapped
		// in UnlimitedMemoryTrackerPromQLEngine to provide a memory consumption tracker.)
```

pkg/ruler/compat.go
charleskorn: Added a suggestion below.

pkg/ruler/compat.go
lamida: Committed the suggestion and removed the above comment here d32102c991e449a92413c36739e9ae948621cc1e

PR #12880, opened 2025-10-01
bucketcache: ignore context cancellation in cachedGetRange

narqo:
#### What this PR does

This PR is a temporal hack, while we investigate the details outlined in https://github.com/grafana/mimir/issues/12691#issuecomment-3356081978

My current lead is that there is a rare race condition, that got exacerbated, after the gomemcache `GetMulti` started to respect the cancelled context. I'm still trying to proof that with a test. But, meanwhile, I suggest we revert the behaviour to one before https://github.com/grafana/dskit/pull/737 

Note, we can't easily revert the changes in dskit at this moment, because there are already changes that depend on the new interface in the upstream dependencies, that GEM uses.

Relates to https://github.com/grafana/mimir/issues/12691

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


narqo:
Close in favour of a proper fix https://github.com/grafana/gomemcache/pull/28

PR #12883, opened 2025-10-02
Limit parallelism of sharded queries when sharding is running inside MQE

charleskorn:
#### What this PR does

This PR fixes the issue where the `max_query_parallelism` / `-querier.max-query-parallelism` option is not respected if sharding is running inside MQE.

⚠️ This PR is dependent on https://github.com/grafana/mimir/pull/12835 and https://github.com/grafana/mimir/pull/12885.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/limits.go
+++ b/pkg/frontend/querymiddleware/limits.go
@@ -211,14 +211,18 @@ func (l limitsMiddleware) Do(ctx context.Context, r MetricsQueryRequest) (Respon
 
 type limitedParallelismRoundTripper struct {
 	downstream MetricsQueryHandler
-	limits     Limits
+	limits     LimitedParallelismLimits
 
 	codec      Codec
 	middleware MetricsQueryMiddleware
 }
 
+type LimitedParallelismLimits interface {
+	MaxQueryParallelism(userID string) int
+}
+
 // NewLimitedParallelismRoundTripper creates a new roundtripper that enforces MaxQueryParallelism to the `next` roundtripper across `middlewares`.
-func NewLimitedParallelismRoundTripper(next MetricsQueryHandler, codec Codec, limits Limits, middlewares ...MetricsQueryMiddleware) http.RoundTripper {
+func NewLimitedParallelismRoundTripper(next MetricsQueryHandler, codec Codec, limits LimitedParallelismLimits, middlewares ...MetricsQueryMiddleware) http.RoundTripper {
 	return limitedParallelismRoundTripper{
 		downstream: next,
 		codec:      codec,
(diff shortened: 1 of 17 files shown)


pkg/frontend/querymiddleware/model_extra.go
```diff
@@ -35,8 +35,8 @@ var (
 	}.Froze()
 )
 
-// newEmptyPrometheusResponse returns an empty successful Prometheus query range response.
-func newEmptyPrometheusResponse() *PrometheusResponse {
+// NewEmptyPrometheusResponse returns an empty successful Prometheus query range response.
+func NewEmptyPrometheusResponse() *PrometheusResponse {
```
lamida, line 39: Why do we export this function? The usage seems still all within the package?

pkg/frontend/querymiddleware/limits.go
```diff
@@ -242,25 +248,25 @@ func (rt limitedParallelismRoundTripper) RoundTrip(r *http.Request) (*http.Respo
...
-			if err := sem.Acquire(ctx, 1); err != nil {
-				// Without this change, using WithTimeoutCause has no effect when calling Do on
-				// limitedParallelismRoundTripper, since that would need to return the cause as error,
-				// which is the normal behaviour except that semaphore does not do that.
-				if errors.Is(err, ctx.Err()) {
-					err = context.Cause(ctx)
+			// If remote execution is enabled, then don't apply the parallelism limit here, as we'll enforce it in Frontend.DoProtobufRequest instead.
+			// If we enforce the limit here, then we'll count requests twice: once here, and again in Frontend.DoProtobufRequest.
+			// We need to enforce the limit in DoProtobufRequest because the requests we see here are before sharding is applied if sharding inside MQE is enabled.
+			if !rt.remoteExecutionEnabled {
```
lamida: Nits. Can we have the comment to explain the thing consistent with the line being commented? The comment start with "If remote execution is enabled, then don't apply..." while the line being commented is `if !rt.remoteExecutionEnabled`.

```suggestion
			// Apply parallelism limit here if remoteExecution is not enabled.
			// If remoteExecution is enabled, enforce the limit in DoProtobufRequest because the requests we see here are before sharding is applied if sharding inside MQE is enabled.
			// // If we enforce the limit here, then we'll count requests twice: once here, and again in Frontend.DoProtobufRequest.
			if !rt.remoteExecutionEnabled {
```

pkg/frontend/querymiddleware/model_extra.go
charleskorn: It's now used in `pkg/frontend/v2`'s tests, in `TestMaxQueryParallelismWithRemoteExecution`.

pkg/frontend/querymiddleware/limits.go
charleskorn: Good idea, clarified in ff5b26f.

PR #12884, opened 2025-10-02
Misc compactor scheduler renaming

jhalterman:
#### What this PR does

This PR makes a few small changes:

- Adds a mockable clock everywhere time.Now() is used
- Documents and orders member variables to indicate fields that are guarded
- Makes the Remove methods more DRY
- Short circuits the ExpireLeases loop
- Fixed a *Job type assertion

And renames a few things:

- Renames job k/v to be more descriptive
- Uses more general variable names for job storage
- Tries to improve naming of return vars
- Renames the job tracker receiver vars

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -238,7 +238,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.25.3 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sts v1.33.19 // indirect
 	github.com/aws/smithy-go v1.22.5 // indirect
-	github.com/benbjohnson/clock v1.3.5 // indirect
+	github.com/benbjohnson/clock v1.3.5
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bits-and-blooms/bitset v1.20.0 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect


pkg/compactor/scheduler/job_tracker.go
```diff
@@ -39,157 +43,154 @@ func (j *Job[K, V]) IsLeased() bool {
...
-	leased     *list.List          // leased jobs
-	elementMap map[K]*list.Element // all tracked jobs will be in this map, element is in one and only one of q or leased
-	maxLeases  int
-	mtx        *sync.Mutex
+	clock     clock.Clock
+	maxLeases int
+	mtx       *sync.Mutex
+
+	// Guarded by mtx
+	pending *list.List
```
andyasp, line 63: Oleg told me about wearing mutexes as a hat to indicate this rather than a comment and I've really liked it: https://dmitri.shuralyov.com/idiomatic-go/entries/2

pkg/compactor/scheduler/job_tracker.go
```diff
@@ -39,157 +43,154 @@ func (j *Job[K, V]) IsLeased() bool {
...
 		next = e.Next() // get the next element now since it can't be done after removal
 
 		j := e.Value.(*Job[K, V])
 		if now.Sub(j.lastRenewalTime) > leaseDuration {
-			if jq.endLease(e, j) {
+			if jt.endLease(e, j) {
 				revivedAny = true
 			}
+		} else {
+			// No more expirable jobs
```
andyasp, line 163: I don't think this `else` is correct. A full iteration is required currently because lease heartbeats can be received (or not received) at different times for each job. That means a job's position in the active list doesn't indicate if it can be expired.

Since newly leased items are pushed to the back we could change renewals to move the element to the back of the active list and then this would work (assuming a monotonic clock).

pkg/compactor/scheduler/job_tracker.go
```diff
@@ -7,30 +7,34 @@ import (
...
-const INFINITE_LEASES = 0
+const InfiniteLeases = 0
 
 // Job is the wrapper type within the JobTracker.
 // K identifies the job. If another job is offered with K the conflict must be decided.
 // V is the contained type
 type Job[K comparable, V any] struct {
-	k K
-	v V
+	id    K
```
andyasp: I find `id` more confusing than `k` here, but I don't feel strongly about it.

pkg/compactor/scheduler/rotator.go
```diff
@@ -18,24 +18,24 @@ const outsideRotation int = -1
...
-	tenantStateMap map[string]*TenantRotationState
-	rotation       []string
-	mtx            *sync.RWMutex
-
+	planTracker          *JobTracker[string, struct{}]
+	leaseDuration        time.Duration
+	leaseCheckInterval   time.Duration
+	maxLeases            int
 	rotationIndexCounter *atomic.Int32 // only increments, overflow is okay
+	mtx                  *sync.RWMutex
```
andyasp: Same mutex hat comment (I did the reverse before apparently).

pkg/compactor/scheduler/job_tracker.go
jhalterman: I like that too

pkg/compactor/scheduler/job_tracker.go
jhalterman: Removed K as we discussed and made id a string

pkg/compactor/scheduler/job_tracker.go
jhalterman: Good catch. I updated the RenewLease method to move a renewed item to the back of the active list

pkg/compactor/scheduler/job_tracker.go
```diff
@@ -7,243 +7,243 @@ import (
...
 		return false
 	}
 
-	j := e.Value.(*Job[K, V])
+	j := e.Value.(*Job[V])
 	if j.IsLeased() && j.epoch == epoch {
-		j.lastRenewalTime = time.Now()
+		j.lastRenewalTime = jt.clock.Now()
+		jt.active.Remove(e)
+		jt.allJobs[id] = jt.active.PushBack(j)
```
andyasp:
```suggestion
		jt.active.MoveToBack(e)
```

PR #12885, opened 2025-10-02
Remote execution: buffer responses from queriers in query-frontends if needed

charleskorn:
#### What this PR does

This PR adds support for buffering responses from queriers in query-frontends if the query contains sharded expressions and this sharding is being performed inside MQE. This avoids a possible deadlock scenario:

When concatenating results from multiple sharded legs, we need the series metadata for all sharded legs before we can start streaming series data back to the caller. If we don't eagerly load responses from queriers, then in a query with (for example) 8 sharded legs, we might end up deadlocked waiting for some or all of the last legs because these requests can't start either due to the query-frontend's per-query max parallelism limit or the number of available querier workers. If we eagerly load responses, then exhausting the responses that have started will eventually make space for the outstanding requests to complete.

⚠️ This PR is dependent on https://github.com/grafana/mimir/pull/12835.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/v2/frontend_test.go
+++ b/pkg/frontend/v2/frontend_test.go
@@ -369,10 +369,11 @@ func TestFrontend_Protobuf_ResponseClosedBeforeResponseReceived(t *testing.T) {
 		}
 
 		go func() {
+			resp := <-respChannel
+
 			// Close the stream returned by DoProtobufRequest once we're confident the goroutine in DoProtobufRequest has observed that the request has been enqueued
 			// and is waiting for streamContext to be cancelled.
-			// This ensures that closing the channel doesn't trigger the code path that calls writeEnqueueError().
-			resp := <-respChannel
+			// This ensures that closing the stream doesn't trigger the code path that calls writeEnqueueError().
 			time.Sleep(10 * time.Millisecond)
 			resp.Close()
 		}()


pkg/streamingpromql/optimize/plan/remoteexec/node.proto
```diff
@@ -12,5 +12,13 @@ option (gogoproto.messagename_all) = true;
 option (gogoproto.unmarshaler_all) = true;
 
 message RemoteExecutionDetails {
-  // No fields yet.
+  // If the query has been sharded, we need to eagerly load and buffer the response from queriers in query-frontends
+  // to avoid a possible deadlock:
```
lamida: Nits. Do you intend to put period or colon?
```suggestion
  // to avoid a possible deadlock.
```

pkg/frontend/v2/remoteexec.go
```diff
@@ -385,3 +393,154 @@ func accountForHPointMemoryConsumption(d []promql.HPoint, memoryConsumptionTrack
...
+	e.inner.Close() // This is expected to release any pending Next() call in startBuffering(), so we don't need to do anything to terminate it here.
+}
+
+type responseStreamBuffer struct {
+	msgs       []bufferedMessage
+	startIndex int
+	length     int
+}
+
+type bufferedMessage struct {
```
lamida, line 502: Nits. Any reason why we name this with Message suffix? Since we meant to buffer ResultStream, maybe just name this bufferedResult?

pkg/frontend/v2/remoteexec.go
```diff
@@ -385,3 +393,154 @@ func accountForHPointMemoryConsumption(d []promql.HPoint, memoryConsumptionTrack
...
+	payload *frontendv2pb.QueryResultStreamRequest
+	err     error
+}
+
+func (b *responseStreamBuffer) Push(msg bufferedMessage) {
+	if b.length == cap(b.msgs) {
+		newCap := max(len(b.msgs)*2, 1)
```
lamida, line 509: What is the typical size of the buffered result? Would keep doubling as the buffer full will always be optimal? Also why don't we just initiate capacity to 1, instead of having max setting it for the first call when buffer is empty?

pkg/frontend/v2/remoteexec.go
```diff
@@ -385,3 +393,154 @@ func accountForHPointMemoryConsumption(d []promql.HPoint, memoryConsumptionTrack
...
+	}
+
+	return msg
+}
+
+func (b *responseStreamBuffer) Any() bool {
+	return b.length > 0
+}
+
+var responseMessageSlicePool = pool.NewBucketedPool(131072, func(size int) []bufferedMessage {
```
lamida: Should we set 131072 to a constant with an explanation why we set to that size instead of passing as something like magic number here.

pkg/frontend/v2/remoteexec.go
```diff
@@ -385,3 +393,154 @@ func accountForHPointMemoryConsumption(d []promql.HPoint, memoryConsumptionTrack
...
+}
+
+func newEagerLoadingResponseStream(ctx context.Context, inner responseStream) *eagerLoadingResponseStream {
+	stream := &eagerLoadingResponseStream{
+		inner:  inner,
+		mtx:    &sync.Mutex{},
+		buffer: &responseStreamBuffer{},
+	}
+
+	go stream.startBuffering(ctx)
```
lamida, line 413: I am still trying to wrap my head around this but, will there be any situation when this goroutine won't be terminated even after we call Close? For example we call Close before stream started (`responseStarted` in `ProtobufResponseStream` still blocks), hence Close goes to `s.cancelStream(errStreamClosed)`. In that case won't `ProtobufResponseStream.Next` (subsequently this goroutine) will be blocked?

zenador (commented):
LGTM, though I don't fully understand everything, trying to get into the habit of reviewing more PRs and learning more about how the streaming works.

pkg/frontend/v2/remoteexec.go
zenador: Or maybe just a comment will do, 128kb seems to be somewhat commonly used in Mimir anyway.

pkg/frontend/v2/remoteexec.go
```diff
@@ -385,3 +393,154 @@ func accountForHPointMemoryConsumption(d []promql.HPoint, memoryConsumptionTrack
...
+		defer e.mtx.Unlock()
+
+		if msg.err != nil {
+			e.bufferedErr = msg.err
+		}
+
+		return msg.payload, msg.err
+	}
+
+	newDataAvailable := make(chan struct{}) // We can't store this directly in e.newDataAvailable because e.newDataAvailable might be cleared by the time we wait on it below.
```
zenador, line 472: (non-blocking) Don't really understand this part, just making a note so I can ask about it next time.

pkg/streamingpromql/optimize/plan/remoteexec/node.proto
charleskorn: The colon is intentional, the next sentence explains the deadlock scenario. I'll reformat this to make it clearer.

pkg/frontend/v2/remoteexec.go
charleskorn: To me, the result is the whole response, whereas the response is made up of many messages. I don't hold this opinion strongly though.

pkg/frontend/v2/remoteexec.go
charleskorn: >What is the typical size of the buffered result?

This is very hard to say, it depends both on the rate at which messages are being consumed and the type of query being performed. If the messages are being consumed as fast as they are produced, then this could stay with just a single element, but if the entire result is buffered before any of it is read, then there will be `2 + numSeries` messages in the buffer for an instant vector result.

>Would keep doubling as the buffer full will always be optimal?

This aligns with the size of the slices pooled by `pool.BucketedPool`, which are always powers of 2. 

>Also why don't we just initiate capacity to 1, instead of having max setting it for the first call when buffer is empty?

We definitely could, but doing it this way removes the need for a constructor method - a new instance is ready for use.

pkg/frontend/v2/remoteexec.go
charleskorn: Good point - I've opted to reuse the value of `types.MaxExpectedSeriesPerResult` in 426e5c9 as the vast majority of these slices will be used for instant vector results, which will have `2 + numSeries` messages.

pkg/frontend/v2/remoteexec.go
charleskorn: Yep, good catch, I think you're right. I'll put together a unit test to confirm if this is an issue and come up with a fix.

pkg/frontend/v2/remoteexec.go
charleskorn: Could you elaborate? Are you referring to why we don't store this in `e.newDataAvailable`?

pkg/frontend/v2/remoteexec.go
charleskorn: Fixed in 9838373.

pkg/frontend/v2/remoteexec.go
lamida: I think just leave it as it is.

pkg/frontend/v2/frontend_test.go
```diff
@@ -359,6 +359,56 @@ func TestFrontend_Protobuf_ResponseClosedBeforeStreamExhausted(t *testing.T) {
...
+		if msg.Type != schedulerpb.ENQUEUE {
+			return &schedulerpb.SchedulerToFrontend{Status: schedulerpb.ERROR, Error: fmt.Sprintf("unexpected message type %v sent to scheduler", msg.Type)}
+		}
+
+		go func() {
+			// Close the stream returned by DoProtobufRequest once we're confident the goroutine in DoProtobufRequest has observed that the request has been enqueued
+			// and is waiting for streamContext to be cancelled.
+			// This ensures that closing the channel doesn't trigger the code path that calls writeEnqueueError().
+			resp := <-respChannel
+			time.Sleep(10 * time.Millisecond)
```
lamida, line 377: I wish we can do this without time.Sleep, but I believe that will make the code harder to read 🤔 . Hopefully this won't make the test flaky in CI.

lamida (approved):
LGTM. Thanks for addressing the feedbacks.

pkg/frontend/v2/frontend_test.go
charleskorn: Now that https://github.com/grafana/mimir/pull/12917 has been merged, this might be a good candidate for using `synctest`, let me give that a go.

charleskorn: Unfortunately, because we're using network I/O in this test, we can't easily use `synctest`, as being blocked on I/O doesn't count as durably blocked.

PR #12886, opened 2025-10-02
Block queries containing functions not intended to be used by users

charleskorn:
#### What this PR does

This PR adds a query-frontend middleware that can be used to block internal functions, such as the `__sharded_concat__` function used by MQE's sharding implementation.

⚠️ This PR is dependent on https://github.com/grafana/mimir/pull/12835.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/block_internal_functions.go
+++ b/pkg/frontend/querymiddleware/block_internal_functions.go
@@ -34,25 +34,30 @@ func (b *blockInternalFunctionsMiddleware) Do(ctx context.Context, request Metri
 		return nil, err
 	}
 
-	containsInternalFunction, err := astmapper.AnyNode(expr, b.isInternalFunctionCall)
+	forbiddenFunctionName := ""
+	containsInternalFunction, err := astmapper.AnyNode(expr, func(node parser.Node) (bool, error) {
+		call, isCall := node.(*parser.Call)
+		if !isCall {
+			return false, nil
+		}
+
+		if b.functionsToBlock.Contains(call.Func.Name) {
+			forbiddenFunctionName = call.Func.Name
+			return true, nil
+		}
+
+		return false, nil
+	})
+
 	if err != nil {
 		return nil, err
 	}
 
 	if containsInternalFunction {
-		return nil, apierror.Newf(apierror.TypeBadData, "expression contains an internal function not permitted in queries")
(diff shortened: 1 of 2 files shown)


pkg/streamingpromql/optimize/ast/sharding/concat.go
```diff
@@ -19,6 +20,7 @@ import (
 
 func init() {
 	parser.Functions[ConcatFunction.Name] = ConcatFunction
+	querymiddleware.InternalFunctionNames[ConcatFunction.Name] = struct{}{}
```
56quarters: Can we make the registry of internal functions its own dedicated type of object with methods to register functions?

pkg/frontend/querymiddleware/block_internal_functions.go
```diff
@@ -0,0 +1,60 @@
...
+var InternalFunctionNames = map[string]struct{}{
+	// Internal functions are registered by the packages that implement them.
+}
+
+type blockInternalFunctionsMiddleware struct {
+	next   MetricsQueryHandler
+	logger log.Logger
+}
+
+func newBlockInternalFunctionsMiddleware(logger log.Logger) MetricsQueryMiddleware {
```
56quarters: Instead of referencing the global from this middleware, could we pass a reference to it when the middleware is created?

charleskorn:
> I'd prefer minimizing global state when possible.

Makes sense to me, how's this looking now @56quarters?

pkg/frontend/querymiddleware/block_internal_functions.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, err
+	}
+
+	containsInternalFunction, err := astmapper.AnyNode(expr, b.isInternalFunctionCall)
+	if err != nil {
+		return nil, err
+	}
+
+	if containsInternalFunction {
+		return nil, apierror.Newf(apierror.TypeBadData, "expression contains an internal function not permitted in queries")
```
lamida: Nits. Can we output the function name in this error? It is only `sharding.ConcatFunction.Name` now but it will be helpful to investigate if more functions are added.

lamida (approved):
Add small nits comment. But otherwise LGTM.

56quarters (approved):
LGTM, thanks!

pkg/frontend/querymiddleware/block_internal_functions.go
charleskorn: Good idea, added in 2d66496.

PR #12891, opened 2025-10-02
continuous-test: Fix false positive in continuous tester metadata assertion

alexweav:
#### What this PR does

If a request contains multiple metadata that are exact duplicates of each other, the request has the same meaning if we deduplicate them.

In the translation logic we generally try to do this, but opportunistically. We drop duplicates when it's computationally convenient to do so, but we aren't guaranteed to drop dupes on every single request. In RW2.0 duplicates are cheap given the strings are already symbolized, so it's a best-effort approach.

When asserting metadata, the correctness tester adjusts the V1 metadata for dupes but forgot to adjust the V2 metadata. So, for a request that didn't fall in the best effort category, we're comparing deduplicated against duplicates being present, meaning the `cmp.Equal` fails. The requests still mean the same thing, but we flag it when we shouldn't.

To avoid the false positive we simply adjust for duplicates on both sides before comparing.
no-changelog because this is an experimental test.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/continuoustest/ingest_storage_record.go
+++ b/pkg/continuoustest/ingest_storage_record.go
@@ -410,8 +410,8 @@ func (t *IngestStorageRecordTest) testRec(rec *kgo.Record, report batchReport) b
 		sortMetadata := cmpopts.SortSlices(func(m1, m2 *mimirpb.MetricMetadata) bool {
 			return m1.MetricFamilyName < m2.MetricFamilyName
 		})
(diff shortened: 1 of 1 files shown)


alexweav:
I'm going to backport this to r362 so we can advance the testing a release earlier without needing to worry about results being clouded by it. It worked great in dev.

PR #12894, opened 2025-10-03
store-gateway: create block directory before storing sparse index header

dimitarvdimitrov:
there's a race between those two functions. `ensureIndexHeaderOnDisk` creates a directory and `tryDownloadSparseHeader` needs that directory. Sometimes `tryDownloadSparseHeader` fails because it runs before `ensureIndexHeaderOnDisk`

https://github.com/grafana/mimir/blob/c8df29104106ce842c92cb38eeb81f972783607f/pkg/storage/indexheader/lazy_binary_reader.go#L141-L149

example log line
```
ts=2025-10-03T12:11:26.370447959Z caller=lazy_binary_reader.go:187 level=info user=10809 msg="could not store sparse index-header on disk; will reconstruct when the block is queried" err="open /data/tsdb/10809/01K6M9CCNX2WG0ZDKWPW59YSFM/sparse-index-header.tmp: no such file or directory"
```

### DIFF
--- a/pkg/storage/indexheader/lazy_binary_reader.go
+++ b/pkg/storage/indexheader/lazy_binary_reader.go
@@ -182,6 +182,17 @@ func tryDownloadSparseHeader(ctx context.Context, logger log.Logger, bkt objstor
 		level.Info(logger).Log("msg", "could not download sparse index-header from bucket; will reconstruct when the block is queried", "err", err)
 		return
 	}
+
+	dir := filepath.Dir(sparseHeaderPath)
+	if df, err := os.Open(dir); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+			level.Info(logger).Log("msg", "could not create directory for sparse index-header; will reconstruct when the block is queried", "err", err)
+			return
+		}
+	} else {
+		_ = df.Close()
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/storage/indexheader/lazy_binary_reader.go
```diff
@@ -182,6 +182,17 @@ func tryDownloadSparseHeader(ctx context.Context, logger log.Logger, bkt objstor
 		level.Info(logger).Log("msg", "could not download sparse index-header from bucket; will reconstruct when the block is queried", "err", err)
 		return
 	}
+
+	dir := filepath.Dir(sparseHeaderPath)
+	if df, err := os.Open(dir); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+			level.Info(logger).Log("msg", "could not create directory for sparse index-header; will reconstruct when the block is queried", "err", err)
```
narqo: (nit)

```suggestion
			level.Warn(logger).Log("msg", "could not create directory for sparse index-header; will reconstruct when the block is queried", "err", err)
```

dimitarvdimitrov:
> The change is ok with me, overall. I wonder, wouldn't it be more future-proof if the `NewLazyBinaryReader` constructor guaranteed the existence of the directory?

the reason i went with doing it for the sparse reader is because WriteBinary was already checking for that directory. But we can move this in a single place - see https://github.com/grafana/mimir/pull/12894/commits/874e7ceb2d8d3da95d66edf02e9aafb087a77865

PR #12896, opened 2025-10-03
index planning: add configurable cost

dimitarvdimitrov:
Since our iteration speed is a bit limited, it would be helpful to tune the cost of the planner without having to recompile it. This PR extracts some threshold and costs as configration options.

```
-blocks-storage.tsdb.index-lookup-planning.min-series-per-block-for-query-planning
-blocks-storage.tsdb.index-lookup-planning.retrieved-posting-cost
-blocks-storage.tsdb.index-lookup-planning.retrieved-posting-list-cost
-blocks-storage.tsdb.index-lookup-planning.retrieved-series-cost
```

For consistency, I also renamed two existing configuration flags.

```
# PREVIOUSLY
-blocks-storage.tsdb.index-lookup-planning-comparison-portion
-blocks-storage.tsdb.index-lookup-planning-enabled

# NOW
-blocks-storage.tsdb.index-lookup-planning.comparison-portion
-blocks-storage.tsdb.index-lookup-planning.enabled
```

This PR is based on #12830 because of a config option added there


pkg/ingester/lookupplan/plan.go
```diff
@@ -114,7 +108,13 @@ func (p plan) intersectionCost() float64 {
 		iteratedPostings += pred.cardinality
 	}
 
-	return float64(iteratedPostings) * estimatedRetrievedSeriesCost
+	return float64(iteratedPostings) * p.config.RetrievedPostingCost
```
chencs, line 111: Was the switch from estimated retrieved **series** cost to estimated retrieved **posting** cost intentional?

dimitarvdimitrov: So this is the diff that your comment is on 



Image


i'm not sure how github is calculating these diffs 😕 This is what I see for this PR 



Image

and this is what the previous PR did (#12830)

Image


In both cases we stick to postings & series without changing them

dimitarvdimitrov:
i rebased on top of main and dropped the commits from #12830

pkg/storage/tsdb/config.go
```diff
@@ -356,9 +361,10 @@ func (cfg *TSDBConfig) RegisterFlags(f *flag.FlagSet) {
...
 	f.BoolVar(&cfg.TimelyHeadCompaction, "blocks-storage.tsdb.timely-head-compaction-enabled", false, "Allows head compaction to happen when the min block range can no longer be appended, without requiring 1.5x the chunk range worth of data in the head.")
 	f.BoolVar(&cfg.BiggerOutOfOrderBlocksForOldSamples, "blocks-storage.tsdb.bigger-out-of-order-blocks-for-old-samples", false, "When enabled, ingester produces 24h blocks for out-of-order data that is before the current day, instead of the usual 2h blocks.")
-	f.BoolVar(&cfg.IndexLookupPlanningEnabled, "blocks-storage.tsdb.index-lookup-planning-enabled", false, "Controls the collection of statistics and whether to defer some vector selector matchers to sequential scans. This leads to better performance.")
+	f.BoolVar(&cfg.IndexLookupPlanning.Enabled, "blocks-storage.tsdb.index-lookup-planning.enabled", false, "Controls the collection of statistics and whether to defer some vector selector matchers to sequential scans. This leads to better performance.")
```
chencs, line 365:
```suggestion
	f.BoolVar(&cfg.IndexLookupPlanning.Enabled, "blocks-storage.tsdb.index-lookup-planning_enabled", false, "Controls the collection of statistics and whether to defer some vector selector matchers to sequential scans. This leads to better performance.")
```

dimitarvdimitrov: this is a CLI flag and we don't use underscores for them

i went from a dash to a dot because now all index lookup planning parameters are under the same block - `blocks-storage.tsdb.index-lookup-planning`

dimitarvdimitrov: but this made me notice that `blocks-storage.tsdb.head-statistics-collection-frequency` had wrong name and wasn't put in the index-lookup block. i fixed that in https://github.com/grafana/mimir/pull/12896/commits/4568bdb37e44c93fb678b5027bd8c18a6ebe7349

PR #12897, opened 2025-10-03
ingester: always use streaming response for queries

56quarters:
#### What this PR does

This changes ingesters to always stream series and chunks instead of sending all series and chunks in a single response. Streaming has been the default for some time and the ability to disable it was removed in a previous commit.

Part of #12673

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


tcp13equals2 (approved):

pkg/ingester/ingester_test.go
```diff
@@ -6062,11 +6035,14 @@ func TestIngester_QueryStream(t *testing.T) {
 						return receivedSeries, err
 					}
 
-					actualTimeseries = append(actualTimeseries, resp.Timeseries...)
-					actualChunkseries = append(actualChunkseries, resp.Chunkseries...)
+					if resp.StreamingSeries != nil {
```
charleskorn: Why do we need to check for `nil` here and below? `append` and `len` can both handle `nil` slices.

charleskorn (approved):
LGTM modulo possible simplification for tests:

pkg/ingester/ingester_test.go
56quarters: Fixed in fc532fda0556bc86a40658c95d18d120e73a67c9

PR #12898, opened 2025-10-03
Block-builder: fail readiness checks when idle.

seizethedave:
#### What this PR does
Make block-builder workers fail readiness checks when not actively working on a job.
When the block-builder worker pool is sized by the horizontal pod autoscaler, this will cause Kubernetes to select **idle** (i.e., non-`Ready`) pods for termination upon scaledown ([ref](https://medium.com/@AlexanderObregon/how-kubernetes-hpa-decides-which-pod-to-terminate-when-scaling-down-6675ebbdf56f#5cc0)). Without this, Kubernetes would choose ~randomly, which will often select pods that are actively working a job, which may not actually complete within the graceful termination period. We'd prefer that scaledowns terminate idle pods.

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/blockbuilder/blockbuilder.go
+++ b/pkg/blockbuilder/blockbuilder.go
@@ -213,6 +213,7 @@ func (b *BlockBuilder) running(ctx context.Context) error {
 			continue
 		}
 
+		// Once we've gotten a job, we attempt to complete it even if the context is cancelled.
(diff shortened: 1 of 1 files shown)


seizethedave:
Closing for now, as failed readiness checks can affect rollouts.

PR #12911, opened 2025-10-06
Fix issue where step-invariant expressions aren't sharded correctly if preprocessing has been applied

charleskorn:
#### What this PR does

This PR is a follow up to https://github.com/grafana/mimir/pull/12900 and fixes an issue where expressions containing step-invariant expressions aren't sharded properly if preprocessing has been applied to the query and the processed `parser.Expr` is passed to `shardSummer`.

This includes expressions like `sum(foo @ 123)`, as well as `sum(foo) > (2*2)`.

This doesn't affect the middleware-based sharding implementation, as the processed `parser.Expr` isn't passed to `shardSummer`, but it does affect [the MQE implementation](https://github.com/grafana/mimir/pull/12835).

I'd suggest reviewing each commit separately, fedf5e1784ef73d2d8fd7d514dc295f29e12b1e8 is a large diff but this is just refactoring to make the following commit smaller.

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


charleskorn:
Putting this back to a draft as it will need to change after https://github.com/grafana/mimir/pull/12921 is merged.

PR #12916, opened 2025-10-06
Alertmanager: Reuse tenant registries

santihernandezc:
### Problem

When creating an Alertmanager for a tenant, we [create a registry](https://github.com/grafana/mimir/blob/e9c08a31003c14dcaf9feb6890c40b7ffffdca9c/pkg/alertmanager/multitenant.go#L1099) for them.

When deleting a tenant, we [soft-delete their registry](https://github.com/grafana/mimir/blob/e9c08a31003c14dcaf9feb6890c40b7ffffdca9c/pkg/alertmanager/alertmanager_metrics.go#L301-L305). We use soft deletes because some of these metrics are aggregated across tenants before being exposed, and certain metric types decreasing (counters, histograms, summaries) would be interpreted as restarts.

The problem is that we're [gathering](https://github.com/grafana/dskit/blob/d6a51ec2b8c9331eb59e29629661d94324c00a72/metrics/tenant_registries.go#L728-L732) and [keeping](https://github.com/grafana/dskit/blob/d6a51ec2b8c9331eb59e29629661d94324c00a72/metrics/tenant_registries.go#L695-L697) these metrics around indefinitely. The memory footprint will go up with each tenant that comes and goes.

### Solution

This PR makes it possible to reuse tenant registries when creating a new Alertmanager.

### Testing

I used the following settings to make the multi-tenant Alertmanager start per-tenant Alertmanagers on request and shut them down after a grace period of 30 seconds:
- `multitenancy_enabled: true`
- `alertmanager.strict_initialization: true`
- `alertmanager.grafana_alertmanager_idle_grace_period: 30s`

I wrote a script to simulate 1000 tenants sending requests, going quiet for 40 seconds, and repeating. This causes the multi-tenant AM to start and stop Alertmanagers on a loop.

Without the fix, `process_resident_memory_bytes` goes up on each _off -> on_ cycle. After the fix, memory usage is stable throughout cycles.

before_after

### Note

It would be a good idea to eventually change how we soft-delete registries in `dskit`.

### DIFF
--- a/pkg/alertmanager/alertmanager.go
+++ b/pkg/alertmanager/alertmanager.go
@@ -202,8 +202,8 @@ type Replicator interface {
 
 type option func(*Alertmanager) *Alertmanager
 
-// withRegister overrides the 'registry' field in the Alertmanager struct.
-func withRegister(reg prometheus.Registerer) option {
+// withRegisterOverride overrides the 'registry' field in the Alertmanager struct.
+func withRegisterOverride(reg prometheus.Registerer) option {
 	return func(am *Alertmanager) *Alertmanager {
 		am.registry = reg
 		return am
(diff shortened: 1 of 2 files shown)


santihernandezc:
Sub-components in the Alertmanager use the same registry to register metrics. If we pass a no-op registerer, metric updates will also be no-ops.
We need to fix this in `dskit`.

PR #12917, opened 2025-10-06
Bump Go to 1.25

narqo:
#### What this PR does

I propose we update the allowed Go version to 1.25. This version provides [`testing/synctest`][1] that was can start adopting to make our tests less flaky.

[1]: https://pkg.go.dev/testing/synctest

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/grafana/mimir
 
-go 1.24.2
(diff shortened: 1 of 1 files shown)


alexweav (approved):
Build image already using it. Nice!
Probably will need a corresponding enterprise change (enterprise build image still on 1.24)

PR #12677, opened 2025-09-11
Extract query sharding logic for MQE, and don't clone expression while rewriting query

charleskorn:
#### What this PR does

This PR builds upon https://github.com/grafana/mimir/pull/12676.

It makes a couple of notable changes:

* Most of the query sharding logic has been extracted to `querySharder`, in preparation for use by MQE
* If rewriting the query as a sharded query fails, query evaluation now fails, whereas before we would attempt to evaluate the original expression. My rationale for this is:
  * This means it's possible to modify the expression in-place rather than clone it, which brings a noticeable performance benefit (see below).
  * We would previously attempt to evaluate the original expression if it took longer than 10s to rewrite the query for sharding. However, the only instance of this happening that I can find in all of our production clusters at Grafana Labs is for a ridiculously complex query that would likely fail anyway.
  * Apart from the single timeout above, looking across all of our production clusters at Grafana Labs, I haven't found an instance of query sharding failing in the past seven days.
* We no longer clone the expression before rewriting it, or determining if the expression is shardable. This drastically reduces the amount of garbage produced during query sharding.

Benchmark results comparing f5b777d to 2b739d6 show a large reduction in terms of both number and size of allocations for sharded queries, especially for those containing many nested binary expressions. Most reasonable queries containing a binary expression see a 20-40% reduction, and one extreme case sees a 99% reduction. Note that this benchmark does not include the parsing and stringifying done in the query sharding middleware. (I'll include these results in [a comment below](https://github.com/grafana/mimir/pull/12677#issuecomment-3278410805), GitHub doesn't like how long the output is.)

#### Which issue(s) this PR fixes or relates to

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.
- [n/a] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [n/a] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12921, opened 2025-10-07
Fix more issues where some binary expressions aren't sharded

charleskorn:
#### What this PR does

This PR is a follow up to https://github.com/grafana/mimir/pull/12900. It fixes a number of other similar cases that affect other expressions on one side of binary expressions. These were all caused by `canShardAllVectorSelectors` incorrectly returning false. 

I don't believe any of these would have caused incorrect query results, but instead would have caused sharding not to be applied when it should have been.

There is an opportunity to remove the "inspect only" mode introduced in #12677 and instead make use of `willShardAllVectorSelectors` when calculating the number of shardable legs, but I will do this in a follow up PR.

I haven't added a changelog entry for the same reason I did not add one in #12900.

#12900 and #12677

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/astmapper/sharding.go
+++ b/pkg/frontend/querymiddleware/astmapper/sharding.go
@@ -63,7 +63,7 @@ type shardSummer struct {
 
 	shardLabeller ShardLabeller
 
-	canShardAllVectorSelectorsCache map[parser.Expr]bool
+	willShardAllSelectorsCache map[parser.Expr]bool
 }
 
 func newShardSummer(shards int, inspectOnly bool, squasher Squasher, logger log.Logger, stats *MapperStats, shardLabeller ShardLabeller) *shardSummer {
@@ -81,7 +81,7 @@ func newShardSummer(shards int, inspectOnly bool, squasher Squasher, logger log.
 
 		shardLabeller: shardLabeller,
 
-		canShardAllVectorSelectorsCache: make(map[parser.Expr]bool),
+		willShardAllSelectorsCache: make(map[parser.Expr]bool),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -175,47 +168,84 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
+		// If we can't shard this expression, we might still be able to shard the inner expression.
+		// eg. in avg(count(test)), we can't shard the avg(), but we can shard the count().
+		return summer.willShardAllSelectors(expr.Expr)
 
-	if binop, ok := expr.(*parser.BinaryExpr); ok {
-		canShardLHS, err := summer.canShardAllVectorSelectors(binop.LHS)
-		if err != nil {
-			return false, err
+	case *parser.BinaryExpr:
+		if CanParallelize(expr, summer.logger) && summer.isShardableBinOp(expr) {
```
56quarters: nit: `isShardableBinOp()` looks cheaper than `CanParallelize`, can we move it first?

pkg/frontend/querymiddleware/astmapper/sharding_test.go
```diff
@@ -542,14 +573,43 @@ func TestShardSummer(t *testing.T) {
...
+				willShardAllSelectors, err := summer.willShardAllSelectors(expr)
+				require.NoError(t, err)
+				hasVectorSelectors, err := AnyNode(expr, isVectorSelector)
+				require.NoError(t, err)
 
-		t.Run(tt.in+" with 'inspect only' set", func(t *testing.T) {
-			// The query should be unchanged, but the number of sharded queries should still be reported.
-			runTest(t, tt.in, tt.in, 1, true, tt.expectedShardableQueries)
+				if hasVectorSelectors {
+					require.Equal(t, tt.expectedShardableQueries > 0, willShardAllSelectors)
```
56quarters: Can you add a message here since the default failure message of "true is not equal to false" isn't very helpful?

56quarters (approved):
Changes LGTM with some minor suggestions.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: Good idea, changed in 6bca454

pkg/frontend/querymiddleware/astmapper/sharding_test.go
charleskorn: Good idea, added in 2ac504d

PR #12928, opened 2025-10-07
[r362] Fix more issues where some binary expressions aren't sharded

charleskorn:
Manually backport https://github.com/grafana/mimir/commit/626e9a1815a503d6a6957100596554ab5f666aa7 from https://github.com/grafana/mimir/pull/12921


charleskorn:
I'm ignoring the Zizmor failure: the issue is fixed in `main`.

PR #12929, opened 2025-10-07
chore: clean up unused non-streaming code for querying ingesters

56quarters:
#### What this PR does

This change removes code for handling non-streaming query responses from ingesters. Streaming has been the default since Mimir 2.11. The ability to disable streaming was removed in #12790 and the code to send non-streaming responses was removed in #12897.

Part of #12673

Related #12897

#### Note to reviewers

Hide whitespace changes, there's a lot of them because of removing the non-streaming variants from unit tests. There are likely further cleanup and simplifications that could be done but this PR is too big already.

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.


pkg/querier/distributor_queryable_test.go
```diff
@@ -200,117 +170,6 @@ func TestDistributorQuerier_Select(t *testing.T) {
 	}
 }
 
-func TestDistributorQuerier_Select_MixedChunkseriesTimeseriesAndStreamingResults(t *testing.T) {
```
56quarters, line 203: Removed because chunkseries and timeseries are gone so there's no merging to test.

pkg/querier/distributor_queryable_test.go
```diff
@@ -415,45 +274,63 @@ func TestDistributorQuerier_Select_MixedFloatAndIntegerHistograms(t *testing.T)
...
 
-	mergedSamplesS1S2 := []interface{}{
-		genTestHistogram(1000, 1),
-		genTestHistogram(2000, 2),
-		genTestFloatHistogram(2500, 25),
-		genTestHistogram(3000, 3),
-		genTestFloatHistogram(4000, 4),
-		genTestHistogram(5000, 5),
-		genTestHistogram(5500, 55),
+	expectMergedSamplesS1S2 := []interface{}{
```
56quarters, line 275: We need to include counter reset hints in the expected samples now because results are all created from chunks and thus _have_ counter reset hints IIUC.

pkg/querier/distributor_queryable_test.go
```diff
@@ -510,51 +384,67 @@ func TestDistributorQuerier_Select_MixedHistogramsAndFloatSamples(t *testing.T)
...
 		mimirpb.Sample{Value: 2.5, TimestampMs: 2500},
 		mimirpb.Sample{Value: 3, TimestampMs: 3000},
 		mimirpb.Sample{Value: 4, TimestampMs: 4000},
 		mimirpb.Sample{Value: 5, TimestampMs: 5000},
 		// {Value: 5.5, TimestampMs: 5500},  masked by histograms
-		genTestFloatHistogram(5500, 55),
-		genTestFloatHistogram(6000, 60),
-		genTestFloatHistogram(7000, 70),
-		genTestFloatHistogram(8000, 80),
+		genExpectedFloatHistogram(5500, 55, mimirpb.Histogram_UNKNOWN),
```
56quarters, line 393: We need to include counter reset hints in the expected samples now because results are all created from chunks and thus have counter reset hints IIUC.

pkg/querier/querier_test.go
```diff
@@ -430,84 +470,6 @@ func TestBatchMergeChunks(t *testing.T) {
 	require.ElementsMatch(t, m[0].Histograms, m[1].Histograms)
 }
 
-func BenchmarkQueryExecute(b *testing.B) {
```
56quarters, line 433: Removed because streaming responses are stateful which makes them hard to mock. This benchmark is also testing too much code for my preference.

pkg/ingester/client/ingester.proto
```diff
@@ -99,21 +99,14 @@ message ActiveSeriesRequest {
...
-// chunks or timeseries. Only one of these fields will be populated.
-//
-// If streaming chunks is enabled with prefer_streaming_chunks: QueryStreamResponse contains either:
+// QueryStreamResponse contains either:
 // - a batch of series labels (sorted in series order), with is_end_of_series_stream=true for the last batch or
 // - a batch of chunks (sorted in series order).
 //
 // Only one of these two options will be populated.
 message QueryStreamResponse {
-  repeated TimeSeriesChunk chunkseries = 1 [(gogoproto.nullable) = false];
```
charleskorn, line 115: (might be something for a follow-up PR) Looks like we can remove `TimeSeriesChunk` from this file as well.

56quarters: Yes! The process for updating the `.expdiff` file was terrible so I decided to to leave updating the protos anymore for another PR :melting_face:

PR #12931, opened 2025-10-08
Step invariant expression handling

tcp13equals2:

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.


pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,133 @@
...
+// To optimise for this, a single value is determined and used to populate the vector for each step of the given series.
+type StepInvariantExpression struct {
+	*StepInvariantExpressionDetails
+	Inner planning.Node `json:"-"`
+}
+
+func (s *StepInvariantExpression) Describe() string {
+	builder := &strings.Builder{}
+	builder.WriteString("[step invariant]")
+	return builder.String()
```
lamida: Why do we need string builder here? Can't we just return the string directly?

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,133 @@
...
+	case *FunctionCall:
+		opf, err = MaterializeFunctionCall(inner, materializer, adjustedTimeRange, params)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	if opf == nil {
+		return nil, fmt.Errorf("unable to materialize step invariant expression - not materializer found for inner type, inner=%v", reflect.TypeOf(s.Inner))
```
lamida: Small typo I guess. Do you mean to say `no` or `not`? Added small suggestion too to just remove dash there.

```suggestion
		return nil, fmt.Errorf("unable to materialize step invariant expression because of no materializer found for inner type, inner=%v", reflect.TypeOf(s.Inner))
```

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	switch op := op.(type) {
+	case types.InstantVectorOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantInstantVectorOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case types.ScalarOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantScalarOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	}
+
+	return opf, fmt.Errorf("unable to materialize step invariant expression - unhandled operator, operator=%v", reflect.TypeOf(op))
```
lamida: nits
```suggestion
	return opf, fmt.Errorf("unable to materialize step invariant expression because of unexpected operator, operator=%v", reflect.TypeOf(op))
```

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1420,4 +1420,24 @@ eval instant at 60s ts_of_first_over_time(metric_histogram{type="mix_first_float
...
+  {__name__="metric", label="bar", group="b"} 3 3 3 3 3 3 3
+
+eval range from 0 to 60s step 10s sum(metric{label="bar"} @ start())
+  {} 8 8 8 8 8 8 8
+
+eval range from 0 to 60s step 10s metric{label="foo"} / ignoring(label) sum(metric{label="bar"} @ end())
+  {} 2 2 2 2 2 2 2
+
+eval range from 0 to 60s step 10s metric{label="foo"} / ignoring(label) sum(metric{label="bar"} @ end())
+  {} 2 2 2 2 2 2 2
```
lamida: How does these two tests differ?

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1420,4 +1420,24 @@ eval instant at 60s ts_of_first_over_time(metric_histogram{type="mix_first_float
...
-clear
\ No newline at end of file
+clear
+
+load 10s
+  metric{label="foo"} 16x10
+  metric{label="bar", group="a"} 5 4x9
+  metric{label="bar", group="b"} 3 4x9
+
+eval range from 0 to 60s step 10s metric{label="bar"} @ start()
```
lamida, line 1434: I am not sure if that is needed. But do we have to add test for literal? Just like 
```
eval range from 0 to 60s step 10s 55
```
Not sure if such expression goes to the the StepInvariantOperator in this PR.

Maybe also a test with function call like 
```
eval range from 0 to 60s step 10s timestamp(metric @ 1)
```
since we are adding thing in TimestampOperatorFactory above around this.

pkg/streamingpromql/testdata/ours/functions.test
```diff
@@ -1420,4 +1420,24 @@ eval instant at 60s ts_of_first_over_time(metric_histogram{type="mix_first_float
...
   {type="mix_first_missing"} 10
 
-clear
\ No newline at end of file
+clear
+
+load 10s
+  metric{label="foo"} 16x10
+  metric{label="bar", group="a"} 5 4x9
+  metric{label="bar", group="b"} 3 4x9
```
lamida, line 1429: Should we also add test cases with increasing sample values instead of constant to detect possibility of off by one mistake? I.e.  something like

```
metric{label="bar", group="c"} 0+1x10
```

and then add corresponding eval to check.

Also should we add histogram test here?

lamida (dismissed):
In general the changes looks cover the requirement. However I still left some non major comments to consider.

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -468,9 +479,7 @@ func TimestampFunctionOperatorFactory(args []types.Operator, _ labels.Labels, op
 	}
 
 	f := Timestamp
-	_, isSelector := args[0].(*selectors.InstantVectorSelector)
-
-	if isSelector {
+	if isOrWrapsInstantVectorSelector(args[0]) {
```
charleskorn: Could you please add a test for this case?

Also, why is this special case needed? If the expression is something like `timestamp(metric @ 123)`, then I'd expect the whole expression (ie. including the `timestamp()` call) to be step-invariant, so there should never be an inner step-invariant expression.

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
```diff
@@ -0,0 +1,114 @@
...
+	return &StepInvariantInstantVectorOperator{
+		inner:                    op,
+		originalTimeRange:        originalTimeRange,
+		memoryConsumptionTracker: memoryConsumptionTracker,
+	}
+}
+
+func (s *StepInvariantInstantVectorOperator) Inner() types.InstantVectorOperator {
+	return s.inner
+}
```
charleskorn: Why is this needed?

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
```diff
@@ -0,0 +1,114 @@
...
+		return types.InstantVectorSeriesData{}, err
+	}
+
+	if s.originalTimeRange.IsInstant || s.originalTimeRange.StepCount <= 1 {
+		return data, nil
+	}
+
+	// The inner query should be pinned to a single point in time with a single step. We don't expect to have multiple values per series here
+	if len(data.Floats) > 1 || len(data.Histograms) > 1 {
+		return types.InstantVectorSeriesData{}, fmt.Errorf("expected a single value float or histogram series. floats=%d, histograms=%d", len(data.Floats), len(data.Histograms))
```
charleskorn: [nit]

```suggestion
		return types.InstantVectorSeriesData{}, fmt.Errorf("expected a single value float or histogram series, but got %d floats and %d histograms", len(data.Floats), len(data.Histograms))
```

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
```diff
@@ -0,0 +1,114 @@
...
+
+		floats = append(floats, promql.FPoint{T: data.Floats[0].T, F: data.Floats[0].F})
+
+		// Fill the expected steps with the same point.
+		for ts := s.originalTimeRange.StartT + s.originalTimeRange.IntervalMilliseconds; ts <= s.originalTimeRange.EndT; ts += s.originalTimeRange.IntervalMilliseconds {
+			floats = append(floats, promql.FPoint{
+				T: ts,
+				F: data.Floats[0].F,
+			})
+		}
```
charleskorn: Couldn't we simplify this to this?

```suggestion
		// Fill the expected steps with the same point.
		for ts := s.originalTimeRange.StartT; ts <= s.originalTimeRange.EndT; ts += s.originalTimeRange.IntervalMilliseconds {
			floats = append(floats, promql.FPoint{
				T: ts,
				F: data.Floats[0].F,
			})
		}
```

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
```diff
@@ -0,0 +1,114 @@
...
+		histograms, err := types.HPointSlicePool.Get(s.originalTimeRange.StepCount, s.memoryConsumptionTracker)
+		if err != nil {
+			return types.InstantVectorSeriesData{}, err
+		}
+
+		h := data.Histograms[0].H
+
+		// Note that we create a copy of the histogram for each step as we can not re-use the same *FloatHistogram in the slice from the pool.
+		// A copy also ensures that when the original slice is returned to the pool the histograms we continue to reference are not mangled.
+		histograms = append(histograms, promql.HPoint{T: data.Histograms[0].T, H: h.Copy()})
```
charleskorn: [nit] We could save one `Copy` by reusing `data.Histograms[0].H` for the first point, and then set `data.Histograms[0].H` to `nil` just before the `types.HPointSlicePool.Put` call to ensure it's not mutated when something else gets the slice.

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
```diff
@@ -0,0 +1,114 @@
...
+	return s.inner.SeriesMetadata(ctx, matchers)
+}
+
+func (s *StepInvariantInstantVectorOperator) NextSeries(ctx context.Context) (types.InstantVectorSeriesData, error) {
+	data, err := s.inner.NextSeries(ctx)
+	if err != nil {
+		return types.InstantVectorSeriesData{}, err
+	}
+
+	if s.originalTimeRange.IsInstant || s.originalTimeRange.StepCount <= 1 {
```
charleskorn, line 56: If the query is an instant query, should we elide the step invariant operator altogether? (same feedback applies for scalar operator below)

pkg/streamingpromql/operators/step_invariant_scalar_operator.go
```diff
@@ -0,0 +1,81 @@
...
+		return types.ScalarData{}, err
+	}
+
+	if s.originalTimeRange.IsInstant || s.originalTimeRange.StepCount <= 1 {
+		return data, nil
+	}
+
+	// The inner query should be pinned to a single point in time with a single step. We don't expect to have multiple values per series here
+	if len(data.Samples) > 1 {
+		return types.ScalarData{}, fmt.Errorf("expected a single value series. samples=%d", len(data.Samples))
```
charleskorn, line 58: [nit]

```suggestion
		return types.ScalarData{}, fmt.Errorf("expected a single value series, but got %d samples", len(data.Samples))
```

pkg/streamingpromql/operators/step_invariant_scalar_operator.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		floats = append(floats, data.Samples[0])
+
+		for ts := s.originalTimeRange.StartT + s.originalTimeRange.IntervalMilliseconds; ts <= s.originalTimeRange.EndT; ts += s.originalTimeRange.IntervalMilliseconds {
+			floats = append(floats, promql.FPoint{
+				T: ts,
+				F: data.Samples[0].F,
+			})
+		}
```
charleskorn, line 74: [nit] Couldn't we do this?

```suggestion
		for ts := s.originalTimeRange.StartT; ts <= s.originalTimeRange.EndT; ts += s.originalTimeRange.IntervalMilliseconds {
			floats = append(floats, promql.FPoint{
				T: ts,
				F: data.Samples[0].F,
			})
		}
```

pkg/streamingpromql/planning/core/step_invariant_expression.go
charleskorn: We could just return an empty string here, the type of the node is sufficient to describe it (similar to how we don't have a description for `Duplicate` nodes)

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,133 @@
...
+		opf, err = MaterializeAggregateExpression(inner, materializer, adjustedTimeRange, params)
+	case *VectorSelector:
+		opf, err = MaterializeVectorSelector(inner, materializer, adjustedTimeRange, params)
+	case *NumberLiteral:
+		opf, err = MaterializeNumberLiteral(inner, materializer, adjustedTimeRange, params)
+	case *DeduplicateAndMerge:
+		opf, err = MaterializeDeduplicateAndMerge(inner, materializer, adjustedTimeRange, params)
+	case *FunctionCall:
+		opf, err = MaterializeFunctionCall(inner, materializer, adjustedTimeRange, params)
+	}
```
charleskorn: Why is this needed? Why not call `materializer.ConvertNodeToOperator(s.Inner, adjustedTimeRange)`?

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,133 @@
...
+func MaterializeStepInvariantExpression(s *StepInvariantExpression, materializer *planning.Materializer, timeRange types.QueryTimeRange, params *planning.OperatorParameters) (planning.OperatorFactory, error) {
+
+	// note that we set the StartT == EndT with a single step count.
+	adjustedTimeRange := types.QueryTimeRange{
+		StartT:               timeRange.StartT,
+		EndT:                 timeRange.StartT,
+		IntervalMilliseconds: timeRange.IntervalMilliseconds,
+		StepCount:            1,
+		IsInstant:            timeRange.IsInstant,
+	}
```
charleskorn:
```suggestion
	adjustedTimeRange := types.NewInstantQueryTimeRange(timestamp.Time(timeRange.StartT))
```

pkg/streamingpromql/planning.go
```diff
@@ -462,10 +478,11 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
...
 			f.AbsentLabels = mimirpb.FromLabelsToLabelAdapters(functions.CreateLabelsForAbsentFunction(expr.Args[0]))
 		case functions.FUNCTION_TIMESTAMP:
-			vs, isVectorSelector := args[0].(*core.VectorSelector)
-			if isVectorSelector {
+			// Set the ReturnSampleTimestamps on the underlying VectorSelector.
+			// Check for both a raw VectorSelector or when it is nested within a StepInvariantExpression
+			p.applyIfIsOrWrapsVectorSelector(args[0], func(vs *core.VectorSelector) {
 				vs.ReturnSampleTimestamps = true
-			}
+			})
```
charleskorn: Same as above, I'm surprised we need to do this, I wouldn't expect the argument to `timestamp()` to be a step-invariant expression as `timestamp()` itself would be step-invariant if the inner expression is step-invariant.

pkg/streamingpromql/planning.go
```diff
@@ -554,14 +571,46 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
...
 	default:
 		return nil, fmt.Errorf("unknown expression type: %T", expr)
 	}
 }
 
+// wrapInStepInvariantExpr returns true if this expression can be wrapped in a StepInvariantExpression.
+// The StepInvariantExpression wrapping is not required when the expression result will be a matrix, as it will have its own unique timestamps which do not depend on the step.
+func (p *QueryPlanner) wrapInStepInvariantExpr(expr parser.Expr) bool {
+	// We do not duplicate results for range selectors since result is a matrix
+	// with their unique timestamps which does not depend on the step.
```
charleskorn: We don't need to duplicate the underlying samples for range vector selectors and subqueries, but we could still save a bit of work from not unnecessarily recreating the ring buffer views at each step - so I think it'd still be worthwhile to add support for step-invariant range vector expressions.

This would also have the added benefit of removing the need for this special logic here.

pkg/streamingpromql/operators/functions/factories.go
tcp13equals2: The prometheus `promql.engine.go` explicitly excludes the `timestamp` call from being wrapped in a step invariant - regardless of it's argument. 

https://github.com/prometheus/prometheus/blob/main/promql/engine.go#L3945

`case *parser.Call:
		_, ok := AtModifierUnsafeFunctions[n.Func.Name]
		isStepInvariant := !ok
		...
`

AtModifierUnsafeFunctions includes the `timestamp` function. 

Tests have been added for various combinations of `timestamp( ... @ ..)`

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
tcp13equals2: The `functions/factories.go#TimestampFunctionOperatorFactory()` needs to consider if the inner operator is an InstantVectorSelector.

Would you prefer that I make `inner` exportable without needing this accessor func or I could add a helper func to the `StepInvariantInstantVectorOperator` to test if it wraps a certain type?

pkg/streamingpromql/operators/functions/factories.go
tcp13equals2: https://github.com/prometheus/prometheus/pull/17313

pkg/streamingpromql/operators/step_invariant_instant_vector_operator.go
tcp13equals2: ha - I had that originally! Will revert.

pkg/streamingpromql/operators/functions/factories.go
```diff
@@ -468,9 +468,7 @@ func TimestampFunctionOperatorFactory(args []types.Operator, _ labels.Labels, op
 	}
 
 	f := Timestamp
-	_, isSelector := args[0].(*selectors.InstantVectorSelector)
-
-	if isSelector {
+	if _, ok := args[0].(*selectors.InstantVectorSelector); ok {
```
charleskorn: [nit] Could you please revert this change?

pkg/streamingpromql/planning.go
```diff
@@ -462,9 +463,28 @@ func (p *QueryPlanner) nodeFromExpr(expr parser.Expr) (planning.Node, error) {
 		case functions.FUNCTION_ABSENT, functions.FUNCTION_ABSENT_OVER_TIME:
 			f.AbsentLabels = mimirpb.FromLabelsToLabelAdapters(functions.CreateLabelsForAbsentFunction(expr.Args[0]))
 		case functions.FUNCTION_TIMESTAMP:
-			vs, isVectorSelector := args[0].(*core.VectorSelector)
-			if isVectorSelector {
-				vs.ReturnSampleTimestamps = true
+
+			// A special case to re-order when we have a timestamp(StepInvariantExpression(VectorSelector)).
+			// The StepInvariantExpression is moved up to encase the entire function call.
+			// Note that the DeduplicateAndMerge still wraps the function call as the timestamp function returns true under functionNeedsDeduplication().
```
charleskorn, line 504: Could you please add a note here linking to your Prometheus PR so we know when we can remove this?

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,96 @@
...
+
+	switch op := op.(type) {
+	case types.InstantVectorOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantInstantVectorOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case types.ScalarOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantScalarOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	}
+
+	return planning.NewNoOpOperatorFactory(), fmt.Errorf("unable to materialize step invariant expression because of unhandled operator, operator=%v", reflect.TypeOf(op))
```
charleskorn:
```suggestion
	return nil, fmt.Errorf("unable to materialize step invariant expression because of unhandled operator, operator=%v", reflect.TypeOf(op))
```

pkg/streamingpromql/planning/operators.go
```diff
@@ -30,6 +30,10 @@ func NewSingleUseOperatorFactory(operator types.Operator) OperatorFactory {
 	}
 }
 
+func NewNoOpOperatorFactory() OperatorFactory {
```
charleskorn: This is not necessary, see suggestion in `step_invariant_expression.go` above.

pkg/streamingpromql/planning_test.go
```diff
@@ -118,6 +118,14 @@ func TestPlanCreationEncodingAndDecoding(t *testing.T) {
 						Type:        "VectorSelector",
 						Description: `{__name__="some_metric"} @ 0 (1970-01-01T00:00:00Z)`,
 					},
+					{
+						NodeType:       planning.NODE_TYPE_STEP_INVARIANT_EXPRESSION,
+						Details:        marshalDetails(&core.StepInvariantExpressionDetails{}),
+						ChildrenLabels: []string{""},
+						Children:       []int64{0},
+						Type:           "StepInvariantExpression",
+						Description:    ``,
```
charleskorn: [nit] There's no need to specify an empty `Description` like this given that's the default value. 

(same applies in a few places below)

tcp13equals2:
@charleskorn Note that I set this in `plan.go`

```
var MaximumSupportedQueryPlanVersion = QueryPlanVersionOne

const QueryPlanVersionZero = int64(0)
const QueryPlanVersionOne = int64(1)
```

Is this what you were thinking for bumping the supported version?

charleskorn:
> @charleskorn Note that I set this in `plan.go`
> 
> ```
> var MaximumSupportedQueryPlanVersion = QueryPlanVersionOne
> 
> const QueryPlanVersionZero = int64(0)
> const QueryPlanVersionOne = int64(1)
> ```
> 
> Is this what you were thinking for bumping the supported version?

Yep, looks good to me. Might be good to add a comment explaining what was added in v1.

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,100 @@
...
+	*StepInvariantExpressionDetails
+	Inner planning.Node `json:"-"`
+}
+
+func (s *StepInvariantExpression) Describe() string {
+	return ""
+}
+
+func (s *StepInvariantExpression) ChildrenTimeRange(timeRange types.QueryTimeRange) types.QueryTimeRange {
+	return timeRange
```
charleskorn: We should return `types.NewInstantQueryTimeRange(timestamp.Time(timeRange.StartT))` here and then use this method in `MaterializeStepInvariantExpression` below.

pkg/streamingpromql/planning.go
charleskorn: Are you planning to implement this in this PR? Or in a follow-up PR?

tcp13equals2:
@charleskorn Comment is added and the adjustment to the ChildrenTimeRange made ...

tacole02 (commented):
Changelog LGTM

charleskorn:
> @charleskorn Comment is added and the adjustment to the ChildrenTimeRange made ...

Sorry, I was waiting for a response to https://github.com/grafana/mimir/pull/12931#discussion_r2415197760 - are you planning to cover that in this PR?

pkg/streamingpromql/planning.go
tcp13equals2: hmmm - weird. I had all that pulled out. I must have reverted some files when I merged in the node version stuff. Will fix.

tcp13equals2:
> > @charleskorn Comment is added and the adjustment to the ChildrenTimeRange made ...
> 
> Sorry, I was waiting for a response to [#12931 (comment)](https://github.com/grafana/mimir/pull/12931#discussion_r2415197760) - are you planning to cover that in this PR?

This has been updated.

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	switch op := op.(type) {
+	case types.InstantVectorOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantInstantVectorOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case types.ScalarOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantScalarOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case *operators.Subquery:
+		return planning.NewSingleUseOperatorFactory(op), nil
```
charleskorn: Any reason not to support all range vector operators here?

Could you also add a note explaining that we don't need a special operator here because either:
* the root of the query is this node, in which case the query must be an instant query (because range queries can't return range vectors) and so we don't need to do anything further to support this
* the root of the query is something else, in which case the range vector operator must be wrapped in a function like `max_over_time`, `min_over_time` etc. that will also be step-invariant - and so we'll use the instant vector step invariant operator for that

Also, could you please add some tests to `TestRangeVectorSelectors` and `TestSubqueries` to ensure this works as expected? eg. `foo[3m] @ 100` and  `foo[3m:1m] @ 100`

tcp13equals2: I've attempted to add this comment in my own words. Let me know if I have captured it correctly.

I've added tests to both places, noting that `TestRangeVectorSelectors` already had a `selector with @ modifier`.

tcp13equals2:
@charleskorn This PR is updated with the latest merge from main. Note that I left the step invariant plan version to be 1, so both Innokentii's changes and this change are listed under v1.

pkg/streamingpromql/planning/core/step_invariant_expression.go
```diff
@@ -0,0 +1,116 @@
...
+	case types.InstantVectorOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantInstantVectorOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case types.ScalarOperator:
+		return planning.NewSingleUseOperatorFactory(operators.NewStepInvariantScalarOperator(op, timeRange, params.MemoryConsumptionTracker)), nil
+	case types.RangeVectorOperator:
+		// Notes on range vector handling:
+		//
+		// If the query was wrapped in a step-invariant expression and this branch is reached,
+		// the inner expression must be a subquery that returns a range vector at a fixed evaluation
+		// timestamp — for example: metric[3m] @ 100.
```
charleskorn: It could be a range vector selector as well:

```suggestion
		// the inner expression must be a subquery or range vector selector that returns a range vector at a fixed evaluation
		// timestamp — for example: metric[3m:1m] @ 100 or metric[3m] @ 100.
```

PR #12932, opened 2025-10-08
Build postings cache factories per ingester

jhalterman:
#### What this PR does

Updates the ingester to build postings cache factories once per ingester, for head and compacted blocks respectively, rather than once per userTSDB. This is the intended goal of the shared postings for matchers cache - where caches are multi-tenant.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -342,6 +342,10 @@ type Ingester struct {
 	usersMetadataMtx sync.RWMutex
 	usersMetadata    map[string]*userMetricsMetadata
 
+	// For producing postings caches
+	headPostingsForMatchersCacheFactory  tsdb.PostingsForMatchersCacheFactory
+	blockPostingsForMatchersCacheFactory tsdb.PostingsForMatchersCacheFactory
+
 	// Rate of pushed samples. Used to limit global samples push rate.
 	ingestionRate             *util_math.EwmaRate
 	inflightPushRequests      atomic.Int64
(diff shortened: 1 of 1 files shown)


jhalterman:
Good questions. 

> what's the reason for doing this? the description is a bit br

I updated the description - the PR makes the caches multi-tenant, which was the intended end goal for shared caches.

> wouldn't different tenants then invalidate each other's caches when metric name invalidation is enabled?

Yes, that's possible, but it's ok. We can decrease hash collisions which cause premature invalidations by increasing the size of the versions slice, but if/when this happens, it's fine - we just lose a bit of efficiency when postings are recomputed.

> also there's no protection against two tenants getting the same ulid for a block, so we might end up mixing cache contents between tenants

When building a key for a shared/multi-tenant cache, we use the [tenantID as the keyFunc](https://github.com/grafana/mimir/blob/main/pkg/ingester/ingester.go#L2760), which provides additional key information when building keys for shared caches. When shared/multi-tenant caching is enabled, the cache ensures that a [keyFunc be provided](https://github.com/grafana/mimir/blob/main/vendor/github.com/prometheus/prometheus/tsdb/postings_for_matchers_cache.go#L288-L291) and uses its result [when building a shared cache key](https://github.com/grafana/mimir/blob/main/vendor/github.com/prometheus/prometheus/tsdb/postings_for_matchers_cache.go#L729).

PR #12933, opened 2025-10-08
Simplify sharding logic by removing "inspect only" mode

charleskorn:
#### What this PR does

This PR builds upon #12677, #12900 and #12921 to simplify some of the logic introduced in #12677.

#12677 removed all instances of cloning the expression during sharding, but this required us to introduce an "inspect only" mode to allow `QuerySharder` to determine the correct number of shards without mutating the expression.

This PR replaces this by building on the `willShardAllSelectors` method introduced in the PRs above, as this method can calculate the number of sharded selectors that will be returned in an expression, which removes the need for "inspect only" mode.

#12677, #12900 and #12921

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


Konstantinov-Innokentii:
In the spirit of discussion we had yesterday on the MQE weekly - Why do we need to calculate the number of sharded selectors that will be returned in an expression?

charleskorn:
> In the spirit of discussion we had yesterday on the MQE weekly - Why do we need to calculate the number of sharded selectors that will be returned in an expression?

Good question: we use this information to determine how to shard the query while respecting the maximum number of shards per query and other constraints.

https://github.com/grafana/mimir/blob/e6a02ddfa11db733fb974e792e8f41b97e66141f/pkg/frontend/querymiddleware/querysharding.go#L382 is the place where this happens in the existing code.

For example: let's say the maximum number of shards per query is 8. If there's only one sharded selector (eg. the expression is `sum(metric_1)`), then we can shard that selector into 8 shards. But if there are four sharded selectors (eg. the expression is `sum(metric_1) + sum(metric_2) - sum(metric_3) - sum(metric_4)`), then we can only shard each selector into 2 shards.

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -168,83 +166,146 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
 
 	case *parser.SubqueryExpr:
 		// If we've reached this point, then the subquery is not shardable.
 		// If the subquery was shardable, we would have returned true when examining the outer function call.
-		return false, nil
+		return UnshardedExpression(), nil
 
 	default:
-		return false, fmt.Errorf("willShardAllSelectors: unexpected expression type %T", expr)
+		return UnshardedExpression(), fmt.Errorf("unexpected expression type %T in ShardingAnalyzer.Analyze", expr)
```
Konstantinov-Innokentii, line 332: Should we panic if we receive unknown expression type?

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -44,44 +44,42 @@ func (lbl *queryShardLabeller) GetParams(_ int) map[string]string {
...
 }
 
-func NewShardSummerWithLabeller(shards int, inspectOnly bool, squasher Squasher, logger log.Logger, stats *MapperStats, labeller ShardLabeller) ASTMapper {
-	summer := newShardSummer(shards, inspectOnly, squasher, logger, stats, labeller)
+func NewShardSummerWithLabeller(shards int, squasher Squasher, logger log.Logger, stats *MapperStats, labeller ShardLabeller) ASTMapper {
+	summer := newShardSummer(shards, squasher, logger, stats, labeller)
 	return NewASTExprMapper(summer)
 }
 
 type shardSummer struct {
```
Konstantinov-Innokentii, line 56: Nit, and not directly related to the PR:

The comment for NewQueryShardSummer says:
`// NewQueryShardSummer instantiates an ASTMapper which will fan out sum queries by shard.`
That made me think it should only handle sum(...) queries.
but it also doing stuff with avg and other aggregations. Am I missing something?

Konstantinov-Innokentii: Actually, I realized that's how sharded queries are constructed back to one - it sums different shard queries.

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -168,83 +166,146 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
+		}
+		if !lhs.WillShardAllSelectors {
+			return UnshardedExpression(), nil
 		}
-		if !willLHS {
-			return false, nil
+
+		rhs, err := analyzer.Analyze(expr.RHS)
+		if !rhs.WillShardAllSelectors {
+			return UnshardedExpression(), nil
```
lamida, line 292: Should we do err check first like lhs check line 258 above?

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -168,83 +166,146 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
-// willShardAllSelectors returns true if all selectors in expr will be sharded.
+type ShardingAnalyzer struct {
+	logger        log.Logger
+	analysisCache map[parser.Expr]ShardingAnalysisResult
+}
+
+type ShardingAnalysisResult struct {
+	// WillShardAllSelectors is true if all selectors in the expression will be sharded,
+	// or if there are no selectors in the expression.
+	WillShardAllSelectors bool
```
lamida, line 177: This is only used internally. Do we need this to be exported?

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -168,83 +166,146 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
+
+	// The number of sharded selectors in the final expression.
+	// - If the original query was sum(foo), this would be 1.
+	// - If the original query was sum(foo) + bar, this would be 0, as the bar selector won't be sharded and so the sum is not sharded.
+	// - If the original query was sum(foo) + sum(bar), this would be 2.
+	// - If the original query was avg(foo), this would be 2, as avg(foo) is rewritten as sum(foo) / count(foo).
+	//
+	// If WillShardAllSelectors is false, then this will be 0.
+	ShardedSelectors int
```
lamida, line 186: To enforce this invariant (`If WillShardAllSelectors is false, then this will be 0.`), shouldn't we un-export this field and add an accessor method for read access?

pkg/frontend/querymiddleware/astmapper/sharding.go
```diff
@@ -168,83 +166,146 @@ func (summer *shardSummer) MapExpr(ctx context.Context, expr parser.Expr) (mappe
...
-		return summer.willShardAllSelectors(expr.Expr)
+		return analyzer.Analyze(expr.Expr)
 
 	case *parser.SubqueryExpr:
 		// If we've reached this point, then the subquery is not shardable.
 		// If the subquery was shardable, we would have returned true when examining the outer function call.
-		return false, nil
+		return UnshardedExpression(), nil
 
 	default:
```
zenador, line 331: Should we consider `StepInvariantExpr` too?

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: I'd prefer not to panic: if we panic, then the query-frontend will crash and all inflight queries will fail. Whereas if we return an error, then just the affected queries will fail.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: Yep, I don't love the name either - I think we should rename it to something like `ShardingMapper`, but I didn't want to change too much in this PR.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: Yes we should, good catch. I'm surprised the linter didn't complain about this.

charleskorn: Ah, it's because we return the error below anyway. I still think we should make this clearer though.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: This is enforced by the constructor methods below (`UnshardedExpression` and `ShardedExpression`). Even if we unexport the field, then there's nothing stopping something in this package from mutating the field or accessing it directly.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: For the classic (non-MQE) sharding path, we don't need to, as the sharding middleware doesn't work with preprocessed expressions and so the expression won't contain `StepInvariantExpr` nodes.

When this is running inside MQE, we do. I plan to address this in an updated version of https://github.com/grafana/mimir/pull/12911, as it requires a large refactoring to `sharding_test.go` and I'd like to keep this PR small.

pkg/frontend/querymiddleware/astmapper/sharding.go
charleskorn: We don't strictly need it to be exported, no, but I don't see much harm in exporting it.

PR #12938, opened 2025-10-08
jsonnet: Allow minReadySeconds to be set for cache STS

### LINKED ISSUE #12925: Idea: set `minReadySeconds` on Memcached pods

Memcached is sometimes restarted during rollouts (version updates, exporter updates, config changes). It restarts very quickly which has the effect of completely clearing the cache at once. This results in slow requests and increased load during rollouts.

There was a previous attempt to slow down rollouts [using liveness or readiness probes](https://github.com/grafana/mimir/pull/11658) but this resulted in pods being removed from the set of live pods which breaks caching (due to how jumphash works). Instead we can [set `minReadySeconds`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#min-ready-seconds) which requires pods to stay up and ready after updates for some number of seconds before moving on the to next pod. This will slow down Memcached rollouts which is the goal.

_No response_

_No response_

Small (<= 1 month dev)

_No response_

### Proposer?

@pracucci @56quarters

56quarters:
#### What this PR does

Allow minReadySeconds to be set on each cache STS to slow rollouts down by requiring that they are "ready" for at least 60 seconds before moving on to the next pod.

Fixes #12925

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,8 +114,9 @@
 * [CHANGE] Ingester: Disable shipping of blocks on the third zone (zone-c) when using `ingest_storage_ingester_zones: 3` on ingest storage #12743 #12744
 * [CHANGE] Distributor: Increase `server.grpc-max-concurrent-streams` from 100 to 1000. #12742
 * [CHANGE] Ruler Query Frontend: Increase `server.grpc-max-concurrent-streams` from 100 to 300. #12742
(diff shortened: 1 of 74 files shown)


narqo (approved):
🔥

PR #12949, opened 2025-10-08
docs: mark only query-frontend has having Prometheus HTTP API

### LINKED ISSUE #11880: Require query-frontend component
This is a config and breaking change. Reasoning about Mimir is easier if all queries go through query-frontend.

56quarters:
#### What this PR does

This change marks only the query-frontend as having Prometheus HTTP API endpoints instead of the querier. HTTP APIs will be removed from queriers in a future change leaving them with only a gRPC API. This shouldn't affect any users since the query-frontend has been the default way to run Mimir since its inception.

Fixes #11880

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.

### DIFF
--- a/docs/sources/mimir/references/http-api/_index.md
+++ b/docs/sources/mimir/references/http-api/_index.md
@@ -538,9 +538,9 @@ GET /ingester/tsdb/{tenant}
 
 Displays a web page with details about tenant's open TSDB on given ingester.
 
-## Querier / Query-frontend
+## Query-frontend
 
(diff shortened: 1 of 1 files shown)


docs/sources/mimir/references/http-api/_index.md
charleskorn, line 1: There's a "Querier / Query-frontend" heading later in this page that describes each of the query endpoints - should we change that heading to just "query-frontend"?

56quarters: Fixed in 94f25f408ec3815fdf3e4efcd81c1349bee58a07

PR #12951, opened 2025-10-08
Add option to exclude workloads from `MimirRolloutStuck` alert

charleskorn:
#### What this PR does

This PR adds a configuration option to exclude particular StatefulSets or Deployments from the `MimirRolloutStuck` alert. 

This is useful to reduce false-positives for workloads that are expected to be unavailable at times (eg. experimental workloads or those running on spot nodes).

The default configuration preserves the existing behaviour of considering all workloads.

(none)

#### Checklist

- [n/a] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/operations/mimir-mixin/config.libsonnet
+++ b/operations/mimir-mixin/config.libsonnet
@@ -227,6 +227,7 @@
     // The alertname is used to create a hyperlink to the runbooks. Currenlty we only have a single set of runbooks, so different products (e.g. GEM) should still use the same runbooks.
     alert_product: $._config.product,
 
+    // The Deployment or StatefulSet names (eg. 'querier' or 'ingester-zone-a') to exclude from the MimirRolloutStuck alert.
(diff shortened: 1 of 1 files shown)


operations/mimir-mixin/config.libsonnet
```diff
@@ -227,6 +227,9 @@
     // The alertname is used to create a hyperlink to the runbooks. Currenlty we only have a single set of runbooks, so different products (e.g. GEM) should still use the same runbooks.
     alert_product: $._config.product,
 
+    rollout_stuck_alert_ignore_deployments: [],
```
tcp13equals2, line 231: [nit] perhaps a comment here with some examples would be helpful to those less familiar with the rollout group names

charleskorn: Good idea, added in eb5e567

PR #12930, opened 2025-10-07
feat(parquet converter): parallel shard writes

francoposa:
#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #12957, opened 2025-10-09
feat(parquet store gateway): support querying multiple shards per block

npazosmendez:
Support for querying blocks that have multiple parquet files ("shards"). We need this after adding https://github.com/grafana/mimir/pull/12930

### DIFF
--- a/pkg/storegateway/parquet_bucket_store_test.go
+++ b/pkg/storegateway/parquet_bucket_store_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/grafana/dskit/gate"
 	"github.com/grafana/dskit/grpcutil"
 	"github.com/grafana/dskit/services"
+	"github.com/prometheus-community/parquet-common/convert"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/promslog"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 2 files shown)


pkg/parquetconverter/conversion_mark.go
```diff
@@ -25,37 +25,43 @@ const (
...
 	}
 
-	mark := ConversionMark{}
+	mark := ConversionMark{
+		// Default to 1 shard if not set, for retrocompatibility
+		ShardCount: 1,
+	}
 	err = json.Unmarshal(metaContent, &mark)
-	return &mark, err
+	return &mark, err == nil, err
```
jesusvazquez: Why do you need err == nil as a return argument? 

If its an error the execution is already finishing below:
```
	mark, ok, err = ReadConversionMark(ctx, meta.ULID, uBucket, logger)
	if err != nil {
		level.Error(logger).Log("msg", "failed to read conversion mark, skipping", "err", err, "block", meta.ULID.String())
		return
	}
```

pkg/storage/parquet/block/lazy_reader.go
```diff
@@ -447,7 +448,6 @@ func NewLazyReaderLocalLabelsBucketChunks(
 	earlyValidation bool,
 	logger log.Logger,
 ) (*LazyReaderLocalLabelsBucketChunks, error) {
-	shardIdx := FirstShardIndex
```
jesusvazquez: I assume that in these lines below, if there is an error the file name would surface. Just concerned about missing out which shard index failed was passed as input but nothing serious in case the wrapped error already has the information.
```
	if err != nil && !os.IsNotExist(err) {
		return nil, errors.Wrap(err, "read parquet labels file from disk")
	} else {
```

npazosmendez: I assume file-specific errors will probably have the file name yeah, but nevertheless the parent caller still enriches the error string with the shard index, to know which shard caused the error: https://github.com/grafana/mimir/blob/975ea19d1fefce9070f6c17c00052408efcf9d05/pkg/storegateway/parquet_bucket_store.go#L1105

pkg/parquetconverter/conversion_mark.go
npazosmendez: Agree that it's not really necessary with how we currently use the function, I just thought it was easy enough to stick to the contract of "`ok` is true iff you can use the returned value", even if the non-nil error is already telling you

jesusvazquez: ah i see thats the reason 👍

jesusvazquez (approved):

PR #12960, opened 2025-10-10
[r362] Fix issue where aggregation parameters aren't sharded even if the rest of the expression is shardable

charleskorn:
Backport https://github.com/grafana/mimir/commit/63087dada1ead7d1ad3ae17e2072d4a0c53a880e from https://github.com/grafana/mimir/pull/12958


charleskorn:
> Not sure about the `zizmor` error?

It's an issue that has been fixed on `main`, but not backported to this branch.

PR #12961, opened 2025-10-10
Add min plan version to MQE planning nodes

tcp13equals2:
#### What this PR does

This PR introduces a MinimumRequiredPlanVersion version number to each MQE query planning node.

Once a query plan has been built, the largest MinimumRequiredPlanVersion found across all the nodes is used as the query plan's version.

Note that this version is encoded in the query plan when sent to remote queriers, there by allowing queriers to reject plans which they do not support.

#### Checklist

- [ x] Tests updated.
- [ x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.

### DIFF
--- a/pkg/streamingpromql/planning.go
+++ b/pkg/streamingpromql/planning.go
@@ -244,7 +244,7 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 	if err := plan.DeterminePlanVersion(); err != nil {
 		return nil, err
 	}
-	
+
 	if err := observer.OnAllPlanningStagesComplete(plan); err != nil {
 		return nil, err
 	}


pkg/streamingpromql/planning/plan.go
```diff
@@ -23,6 +24,8 @@ import (
 
 var MaximumSupportedQueryPlanVersion = int64(0)
 
+var DefaultNodeMinimumRequiredQueryPlanVersion = int64(0)
```
Konstantinov-Innokentii: nit: WDYT about calling it QueryPlanVersionZero?

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 * [ENHANCEMENT] Distributor: Add new optional config flag `distributor.ha-tracker.failover-sample-timeout` for HA tracker as an additional failover timeout check based on sample time instead of server time. #12331
 * [ENHANCEMENT] Distributor: Add reactive concurrency limiters to protect push operations from overload. #12923
 * [ENHANCEMENT] Ingester: Add experimental matcher set reduction to cost-based lookup planning. #12831
+* [ENHANCEMENT] MQE: Add support for determining a query plan version based on versioned planning nodes. #12961
```
charleskorn: This isn't a user-facing change, so I would lump this under the same changelog entry as https://github.com/grafana/mimir/pull/12808

pkg/streamingpromql/planning/plan.go
charleskorn: +1

pkg/streamingpromql/planning_test.go
```diff
@@ -1164,6 +1164,7 @@ func TestPlanCreationEncodingAndDecoding(t *testing.T) {
 func TestPlanVersioning(t *testing.T) {
 	originalMaximumPlanVersion := planning.MaximumSupportedQueryPlanVersion
 	planning.MaximumSupportedQueryPlanVersion = 9001
+	planning.DefaultNodeMinimumRequiredQueryPlanVersion = 9000
```
charleskorn: I don't think we should change this - instead, I would change the plan below to include a `testNode` with `minimumRequiredPlanVersion` set to 9000.

tcp13equals2: ok - but won't we run into serialization issues with the testNode going through the encode and decode?

charleskorn: Encoding should encode whatever version of the plan it is given. Decoding won't fail because the version of the query plan will be less than `MaximumSupportedQueryPlanVersion`.

pkg/streamingpromql/planning/plan.go
charleskorn: Is there a reason this isn't a constant?

pkg/streamingpromql/planning/plan.go
```diff
@@ -181,6 +187,25 @@ func (p *QueryPlan) ToEncodedPlan(includeDescriptions bool, includeDetails bool)
...
+	if p.Root == nil {
+		return errors.New("query plan version can not be determined without a root node")
+	}
+	p.Version = p.maxMinimumRequiredPlanVersion(p.Root)
+	return nil
+}
+
+func (p *QueryPlan) maxMinimumRequiredPlanVersion(node Node) int64 {
+	maxVersion := node.MinimumRequiredPlanVersion()
+	if node.Children() != nil {
```
charleskorn: There's no need to check for `nil` here: iterating over a `nil` slice is safe.

pkg/streamingpromql/planning.go
```diff
@@ -241,6 +241,12 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 		}
 	}
 
+	if err := plan.DeterminePlanVersion(); err != nil {
+		return nil, err
+	}
+
+	spanLogger.DebugLog("msg", "plan version determined", "plan", plan)
+
```
charleskorn: Rather than logging this here, I would add the plan version as another attribute to the `DebugLog` call a few lines down. (`plan` will be stringified and the string doesn't include the plan version.)

pkg/streamingpromql/planning/plan.pb.go
charleskorn, line 1: You'll need to make these changes in `plan.proto`, as this file is generated from the `.proto` file.

pkg/streamingpromql/planning_test.go
```diff
@@ -1613,3 +1618,68 @@ func TestFunctionNeedsDeduplicationHandlesAllKnownFunctions(t *testing.T) {
 		})
 	}
 }
+
+// testNode is a node for use with TestPlanVersioning.
+// It uses the NumberLiteralDetails to encode an arbitrary minimumRequiredPlanVersion
+// Note that most of the Node interface functions return dummy values, and it does not support children.
+type testNode struct {
```
charleskorn: [nit] Maybe a name like `versioningTestNode` will make the intention clearer here  (ie. that it's not a node suitable for all kinds of tests)

PR #12968, opened 2025-10-10
Add support for specializing keep identifying OTel resource attributes setting

aknuds1:
#### What this PR does

Add support for specializing the _keep identifying OTel resource attributes_ per-tenant parameter, in the same fashion as it works for specializing the OTel resource attributes promotion per-tenant parameter. This is to allow for other strategies than runtime overrides of this particular configuration parameter.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/distributor/otel.go
+++ b/pkg/distributor/otel.go
@@ -214,6 +214,12 @@ func newOTLPParser(
 	discardedDueToOtelParseError *prometheus.CounterVec,
 	OTLPPushMiddlewares []OTLPPushMiddleware,
 ) parserFunc {
+	if resourceAttributePromotionConfig == nil {
+		resourceAttributePromotionConfig = limits
+	}
+	if keepIdentifyingOTelResourceAttributesConfig == nil {
+		keepIdentifyingOTelResourceAttributesConfig = limits
+	}
 	return func(ctx context.Context, r *http.Request, maxRecvMsgSize int, buffers *util.RequestBuffers, req *mimirpb.PreallocWriteRequest, logger log.Logger) error {
 		contentType := r.Header.Get("Content-Type")
 		contentEncoding := r.Header.Get("Content-Encoding")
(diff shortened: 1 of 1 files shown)


pkg/distributor/otel_test.go
```diff
@@ -1901,5 +1922,5 @@ type fakeResourceAttributePromotionConfig struct {
 }
 
 func (c fakeResourceAttributePromotionConfig) PromoteOTelResourceAttributes(string) []string {
-	return []string{"resource.attr"}
+	return c.promote
```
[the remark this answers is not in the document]
aknuds1: Wrong.

pkg/distributor/distributor.go
```diff
@@ -221,6 +221,12 @@ type OTelResourceAttributePromotionConfig interface {
 	PromoteOTelResourceAttributes(id string) []string
 }
 
+// KeepIdentifyingOTelResourceAttributesConfig contains methods for configuring keeping of identifying OTel resource attributes.
+type KeepIdentifyingOTelResourceAttributesConfig interface {
+	// OTelKeepIdentifyingResourceAttributes returns whether to keep identifying OTel resource attributes.
+	OTelKeepIdentifyingResourceAttributes(id string) bool
```
juliusmh: [nit] 
```suggestion
	OTelKeepIdentifyingResourceAttributes(tenantId string) bool
```

pkg/distributor/otel.go
```diff
@@ -328,8 +334,11 @@ func newOTLPParser(
 		if resourceAttributePromotionConfig == nil {
 			resourceAttributePromotionConfig = limits
 		}
+		if keepIdentifyingOTelResourceAttributesConfig == nil {
```
juliusmh: This initialization can be moved to line 217 (before `return func`) to avoid checking on each request. Unrelated, but the check above for the promotion config could be moved as well.

aknuds1:
> Could you explain why we have to pass keepIdentifyingOTelResourceAttributesConfig to newOTLPParser if we can take it from limits anyways?

@juliusmh You can't take it from `limits` because `keepIdentifyingOTelResourceAttributesConfig` is an extension point, which _can_ be the same as `limits` or it can be specialization (or plug-in if you will). It's the same reason as why `OTLPHandler` takes `resourceAttributePromotionConfig`.

pkg/distributor/otel.go
aknuds1: Thanks. Done, PTAL.

PR #12969, opened 2025-10-10
Query Middleware: Clone expressions more efficiently

bboreham:
#### What this PR does

In many places Query Middleware copies the data structure for a PromQL expression, so that it can be modified in private. 
Instead of doing this copy by rendering to a string then parsing the string, use the `CloneExpr` function introduced in #12787.

#### Checklist

- [x] Tests updated.
- NA Documentation added.

---

> [!NOTE]
> Switch middlewares to clone and transform PromQL using `CloneExpr` on parsed ASTs, adding `GetParsedQuery()` and using `WithExpr()` instead of reparsing strings.
> 
> - **AST/Cloning**:
>   - Export `astmapper.CloneExpr` (was `cloneExpr`) and update all internal calls.
> - **Request API**:
>   - Add `MetricsQueryRequest.GetParsedQuery()`; implement in `PrometheusRangeQueryRequest`, `PrometheusInstantQueryRequest`, and remote-read request.
>   - Prefer `WithExpr(parser.Expr)` over `WithQuery(string)` where rewriting/mapping occurs.
> - **Middlewares updated to use parsed AST + CloneExpr**:
>   - Durations, Experimental Functions, Query Sharding, Subquery Spin-off, Stats, Block Internal Functions now clone from `GetParsedQuery()`.
>   - Rewrite middleware now rewrites `parser.Expr` and returns an expr; uses `WithExpr()`.
>   - Split-and-Cache: clone parsed expr; `evaluateAtModifierFunction` now mutates AST directly; `splitQueryByInterval` uses `WithExpr()`.
> - **Tests**:
>   - Update tests to use `CloneExpr`; adjust expectations for `@` timestamp formatting where needed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d306fb6b8622f4b89c9dcd154319b905b313d386. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pkg/frontend/querymiddleware/astmapper/astmapper.go
```diff
@@ -65,19 +65,19 @@ func NewMultiMapper(xs ...ASTMapper) *MultiMapper {
 	return m
 }
 
-// cloneExpr is a helper function to clone an expression.
-func cloneExpr(expr parser.Expr) (parser.Expr, error) {
+// CloneExpr is a helper function to clone an expression.
+func CloneExpr(expr parser.Expr) (parser.Expr, error) {
```
bboreham, line 69: Should this move to a more neutral / utility package?  I couldn't see an obvious place.

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -670,7 +671,12 @@ func doRequests(ctx context.Context, downstream MetricsQueryHandler, reqs []Metr
 func splitQueryByInterval(req MetricsQueryRequest, interval time.Duration) ([]MetricsQueryRequest, error) {
 	// Replace @ modifier function to their respective constant values in the query.
 	// This way subqueries will be evaluated at the same time as the parent query.
-	query, err := evaluateAtModifierFunction(req.GetQuery(), req.GetStart(), req.GetEnd())
+	query, err := astmapper.CloneExpr(req.GetParsedQuery())
+	if err != nil {
+		return nil, err
+	}
+	evaluateAtModifierFunction(query, req.GetStart(), req.GetEnd())
+	splitReq, err := req.WithExpr(query)
```
bboreham, line 679: Pulled this out of the loop, since `splitReq` is cloned again by `WithStartEnd`.

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -1827,11 +1826,11 @@ func TestSplitQueryByInterval(t *testing.T) {
 	queryFooExpr, _ := parser.ParseExpr(queryFoo)
 	queryFooAtStart := "foo @ start()"
 	queryFooAtStartExpr, _ := parser.ParseExpr(queryFooAtStart)
-	queryFooAtZero := "foo @ 0.000"
+	queryFooAtZero := "foo @ 0.00000"
```
bboreham, line 1829: This is necessary to avoid errors on mismatched positions.  Interesting.

pkg/frontend/querymiddleware/split_and_cache_test.go
```diff
@@ -2047,21 +2045,17 @@ func Test_evaluateAtModifier(t *testing.T) {
...
-			out, err := evaluateAtModifierFunction(tt.in, start, end)
-			if tt.err != nil {
-				require.Equal(t, tt.err, err)
-				return
-			}
+			expectedExpr, err := parser.ParseExpr(tt.expected)
 			require.NoError(t, err)
-			require.Equal(t, expectedExpr.String(), out)
+			evaluateAtModifierFunction(expr, start, end)
+			require.Equal(t, expectedExpr.String(), expr.String()) // Character positions are not identical between exprs.
```
bboreham, line 2058: I couldn't figure out what made the positions so different.

pkg/frontend/querymiddleware/block_internal_functions.go
```diff
@@ -29,7 +29,7 @@ func newBlockInternalFunctionsMiddleware(functionsToBlock FunctionNamesSet, logg
 }
 
 func (b *blockInternalFunctionsMiddleware) Do(ctx context.Context, request MetricsQueryRequest) (Response, error) {
-	expr, err := parser.ParseExpr(request.GetQuery())
+	expr, err := astmapper.CloneExpr(request.GetParsedQuery())
```
charleskorn, line 32: This method doesn't mutate `expr`, so we could just use the expression as-is without cloning:

```suggestion
	expr := request.GetParsedQuery()
```

pkg/frontend/querymiddleware/durations.go
```diff
@@ -55,7 +56,7 @@ func (d *durationsMiddleware) rewriteIfNeeded(ctx context.Context, req MetricsQu
 	defer spanLog.Finish()
 
 	origQuery := req.GetQuery()
-	expr, err := parser.ParseExpr(origQuery)
+	expr, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn, line 59: Do we need to clone the expression here? Either we successfully rewrite the expression, and then use that, or something goes wrong, and we abort the request.

pkg/frontend/querymiddleware/astmapper/astmapper.go
charleskorn: I can't see a better place for it.

pkg/frontend/querymiddleware/experimental_functions.go
```diff
@@ -58,7 +59,7 @@ func (m *experimentalFunctionsMiddleware) Do(ctx context.Context, req MetricsQue
 		return m.next.Do(ctx, req)
 	}
 
-	expr, err := parser.ParseExpr(req.GetQuery())
+	expr, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn, line 62: I don't think we need to clone the expression here, we don't mutate it below.

pkg/frontend/querymiddleware/codec.go
```diff
@@ -100,6 +100,8 @@ type MetricsQueryRequest interface {
 	GetStep() int64
 	// GetQuery returns the query of the request.
 	GetQuery() string
+	// GetParsedQuery returns the query, parsed into an AST.
+	GetParsedQuery() parser.Expr
```
charleskorn, line 104: Should we add a note here that if the returned expression is mutated in a way that affects the time range selected, the caller must call `WithExpr` to ensure that the request's `minT` and `maxT` are recalculated correctly?

pkg/frontend/querymiddleware/querysharding.go
```diff
@@ -99,7 +99,7 @@ func (s *querySharding) Do(ctx context.Context, r MetricsQueryRequest) (Response
 		return s.next.Do(ctx, r)
 	}
 
-	queryExpr, err := parser.ParseExpr(r.GetQuery())
+	queryExpr, err := astmapper.CloneExpr(r.GetParsedQuery())
```
charleskorn, line 102: Cloning the expression here should not be necessary - either rewriting the expression for sharding succeeds, sharding is not applicable for the expression (and so it remains unchanged) or we return an error.

pkg/frontend/querymiddleware/remote_read.go
```diff
@@ -282,6 +282,16 @@ func (r *remoteReadQueryRequest) GetQuery() string {
 	return r.promQuery
 }
 
+func (r *remoteReadQueryRequest) GetParsedQuery() parser.Expr {
```
charleskorn, line 285: Why do we need to be able to convert a remote read request into a PromQL expression?

pkg/frontend/querymiddleware/rewrite.go
```diff
@@ -81,29 +82,28 @@ func (m *rewriteMiddleware) Do(ctx context.Context, r MetricsQueryRequest) (Resp
...
 	}
 
 	return m.next.Do(ctx, updatedReq)
 }
 
-func (m *rewriteMiddleware) rewriteQuery(ctx context.Context, query string) (string, bool, error) {
-	// Parse the query.
-	expr, err := parser.ParseExpr(query)
+func (m *rewriteMiddleware) rewriteQuery(ctx context.Context, expr parser.Expr) (parser.Expr, bool, error) {
+	rewrittenQuery, err := astmapper.CloneExpr(expr)
```
charleskorn, line 94: Do we need to clone the expression here? Either we successfully rewrite the query and return the mutated expression, or something goes wrong and we abort.

pkg/frontend/querymiddleware/stats.go
```diff
@@ -71,7 +72,7 @@ func (s queryStatsMiddleware) Do(ctx context.Context, req MetricsQueryRequest) (
 }
 
 func (s queryStatsMiddleware) trackRegexpMatchers(req MetricsQueryRequest) {
-	expr, err := parser.ParseExpr(req.GetQuery())
+	expr, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn, line 75: Do we need to clone the expression here? Looks like we're just inspecting it below.

pkg/frontend/querymiddleware/stats.go
```diff
@@ -137,7 +138,7 @@ func (s queryStatsMiddleware) populateQueryDetails(ctx context.Context, req Metr
 func (s queryStatsMiddleware) findMinMaxTime(ctx context.Context, req MetricsQueryRequest) (int64, int64, bool) {
 	switch r := req.(type) {
 	case *PrometheusRangeQueryRequest, *PrometheusInstantQueryRequest:
-		expr, err := parser.ParseExpr(r.GetQuery())
+		expr, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn: Do we need to clone the expression here? Looks like we're not mutating it below.

pkg/frontend/querymiddleware/spin_off_subqueries.go
```diff
@@ -141,7 +140,7 @@ func (s *spinOffSubqueriesMiddleware) Do(ctx context.Context, req MetricsQueryRe
 	defer cancel()
 	mapper := astmapper.NewSubquerySpinOffMapper(s.defaultStepFunc, spanLog, mapperStats)
 
-	expr, err := parser.ParseExpr(req.GetQuery())
+	expr, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn, line 143: Probably not something for this PR, but I wonder if we can avoid cloning the expression by rejigging the spinoff logic so we either a) leave the expression as-is, b) rewrite it for spinoff or c) return an error, similar to what's been done for the sharding logic.

pkg/frontend/querymiddleware/split_and_cache.go
```diff
@@ -670,7 +671,12 @@ func doRequests(ctx context.Context, downstream MetricsQueryHandler, reqs []Metr
 func splitQueryByInterval(req MetricsQueryRequest, interval time.Duration) ([]MetricsQueryRequest, error) {
 	// Replace @ modifier function to their respective constant values in the query.
 	// This way subqueries will be evaluated at the same time as the parent query.
-	query, err := evaluateAtModifierFunction(req.GetQuery(), req.GetStart(), req.GetEnd())
+	query, err := astmapper.CloneExpr(req.GetParsedQuery())
```
charleskorn, line 674: Do we need to clone the expression here? Even if `evaluateAtModifierFunction` mutates `req`'s expression, `req` is never used again, so I don't think cloning is necessary.

bboreham:
Note I had a previous attempt at not cloning expressions - #11437 - which was reverted after panics were seen.
I would prefer to take the small win from more efficient cloning, then come back to avoid cloning where possible, with more care.

pkg/frontend/querymiddleware/codec.go
bboreham: I would not expect the returned expression to be mutated in-place.
Where do you think this happens?

charleskorn: Mutating the expression in place allows us to avoid cloning it if we expect that the mutation will succeed, or we'll abandon the request. 

For example, in query sharding, we either leave the expression as-is (because it's not shardable), rewrite it for sharding, or return an error.

charleskorn:
> Note I had a previous attempt at not cloning expressions - #11437 - which was reverted after panics were seen. I would prefer to take the small win from more efficient cloning, then come back to avoid cloning where possible, with more care.

Makes sense to me.

charleskorn (approved):
Approving to unblock, but I'd like to see a clarification for https://github.com/grafana/mimir/pull/12969#discussion_r2424921042 before this is merged.

pkg/frontend/querymiddleware/remote_read.go
bboreham: The trivial answer is that this is implemented to conform to `MetricsQueryRequest` interface.
Looking at `remoteReadRoundTripper`, it is running the request through middlewares, so seems intentional.

charleskorn: This seems weird, but not something we're going to fix in this PR - I'd expect an interface that represents the common subset of remote read requests and query requests, and then a separate interface with the parts that are specific to query requests (eg. an expression)

PR #12972, opened 2025-10-10
store-gateway: fix POST /prepare-shutdown without synced tenants

dimitarvdimitrov:
there was a problem where the store-gateway wouldn't accept the /prepare-shutdown call because it couldn't persist the marker on disk. In this particular case it could persist the marker because there weren't any blocks synced for any tenant yet.

This is the place which failed https://github.com/grafana/mimir/blob/77c8cab8dffbe71557b6591f73d077e8c1da5c1c/pkg/storegateway/gateway_prepare_shutdown_http.go#L33-L73

The fix is to create the sync dir during startup


tacole02 (commented):
Changelog LGTM

PR #12974, opened 2025-10-10
Docs: migrate to ingest storage

tacole02:
#### What this PR does
This pull request adds a new documentation page that provides a step-by-step guide for migrating a Grafana Mimir cluster from the classic architecture to the ingest storage architecture with no downtime. 

* Added a new file, `migrate-ingest-storage.md`, that outlines the process for migrating from classic to ingest storage architecture, including prerequisites, detailed migration steps, and operational recommendations.
* The guide covers configuring parallel clusters, duplicating writes, handling compactor and ruler configurations, switching read and write clients, and decommissioning the old cluster.
* Included links to relevant Helm and Jsonnet setup documentation, as well as references to background material on Mimir architectures.
* Added admonitions to highlight important notes and tips regarding cost and alert deduplication during migration.

Fixes [#<issue number>](https://github.com/grafana/mimir-squad/issues/3130)

#### Checklist

- [ ] Tests updated.
- [X] Documentation added.


tacole02:
@dimitarvdimitrov here's the draft for the migration topic. Please have a look and let me know your feedback!

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. Configure write clients to duplicate writes.
+
+   Configure all write clients, such as Prometheus servers and OpenTelemetry Collectors, to send metrics to both clusters simultaneously.
+
+   This ensures that the new cluster receives all recent samples while it accesses historical blocks from object storage.
+
+   Keep read clients, such as Grafana, configured to use the old cluster during this step.
+
+1. Wait for the old cluster’s ingesters to flush data.
```
56quarters: This isn't exactly correct. Ingesters in the old cluster upload new blocks constantly and create new blocks on disk every 2h. What we're waiting for is for both clusters to have a complete view of the data: ingesters are responsible for the last 12h or 13h of data so we need to double write for at least that length of time.

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,115 @@
...
+
+   Allow enough time for all ingesters in the old cluster to flush their in-memory data to object storage. This usually takes about 13 hours. During this time, both clusters are writing and both are healthy.
```
56quarters: Because the data between clusters is being duplicated, it's not clear to me if we need to disable block upload from ingesters in the new cluster until we're ready for the cut-over. @dimitarvdimitrov do you know?

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,115 @@
...
+1. Scale down compactors in the old cluster
+
+   Reduce the number of compactor replicas in the old cluster to zero. For example:
+
+   ```yaml
+   compactor:
+     replicas: 0
+   ```
+
+   Run this step immediately before scaling up the new compactors, ideally within 30 minutes.
```
56quarters: This is a good call-out but I think we need to put more emphasis on this: if the bucket index isn't updated by the compactor in the new cluster within 1h of the compactors in the old cluster being stopped reads will start to fail.

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
dimitarvdimitrov: i don't think it's a requirement. compactors will compact the blocks together. what we might mention is that they should scale out compactors for the duration of the migration. Like add a step before creating the new cluster which is to add 2x compactors to the old cluster to account for the added volume of blocks from the new cluster. After the migration is done, they can scale back down the number of compactors in the new cluster

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
dimitarvdimitrov: good point. perhaps we can also say 15 minutes instead of 30. Or "ideally within 15-30 minutes"?

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Migrate your Grafana Mimir cluster to ingest storage
+
+Follow these steps to migrate your Grafana Mimir cluster from classic architecture to ingest storage architecture.
+
+1. Set up a new Grafana Mimir cluster with ingest storage.
+
+   Deploy a new Grafana Mimir cluster that uses ingest storage. Configure the new cluster to use the same object storage bucket as the old cluster.
+
+   To prevent conflicts, disable the compactor, as shown in the following code sample:
```
dimitarvdimitrov:
```suggestion
   To prevent conflicts, disable the compactor, as shown in the following Helm values file sample:
```

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,115 @@
...
+   compactor:
+     replicas: 0
+   ```
+
+   This configuration lets the new cluster use existing blocks from the old cluster while avoiding block overwrites.
+
+   Optionally, to avoid duplicate alerts, disable alert sending from the ruler. For example:
+
+   ```yaml
+   limits:
```
dimitarvdimitrov: this example isn't a helm example. I think we should either stick with helm examples or non-helm examples. If we mix them, it might be confusing (there's no non-helm example for scaling down compactors, so we just won't show anything)

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
tacole02: I adjusted the wording and added a few new steps for scaling up and down the number of compactors. Please have a look! @dimitarvdimitrov

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
tacole02: Please have another look!

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
tacole02: I think that's a good point. Is there a helm example we can replace this one with?

tacole02:
@dimitarvdimitrov @56quarters thank you for your feedback. I've updated the docs. Could you please take another look?

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
dimitarvdimitrov: that should be it!

```yaml
mimir:
  structuredConfig:
    limits:
      ruler_alerting_rules_evaluation_enabled: false

# or 

ruler:
  extraArgs:
    ruler.alerting-rules-evaluation-enabled: false
```

docs/sources/mimir/set-up/migrate/migrate-ingest-storage.md
```diff
@@ -0,0 +1,137 @@
...
+
+   Keep read clients, such as Grafana, configured to use the old cluster during this step.
+
+   After completing this step, open the Mimir overview dashboards for both clusters and verify that all components are healthy. Check that ingestion and querying are succeeding and that there are no gaps or errors.
+
+1. Wait until both clusters have a complete view of the data.
+
+   Keep write duplication running long enough for both clusters to have a complete view of recent data.
+   
+   Ingesters in the old cluster continuously upload new blocks to object storage and roll new blocks on disk every 2 hours. Because ingesters typically hold about 12–13 hours of recent data in memory and on local disk, continue double writing for at least that duration. This ensures that the new cluster receives all data written during the overlap period and that queries return a complete dataset.
```
56quarters:
```suggestion
   Ingesters in the old cluster continuously upload new blocks to object storage and create new blocks on disk every 2 hours. Because ingesters typically hold about 12–13 hours of recent data in memory and on local disk, continue double writing for at least that duration. This ensures that the new cluster receives all data written during the overlap period and that queries return a complete dataset.
```

PR #12984, opened 2025-10-13
Improve memory consumption tracker accuracy for Labels

lamida:
Original work was in https://github.com/grafana/mimir/pull/12659. Recreated this PR because there are too many rebase needed.

Tracking memory consumption for Labels.

Memory consumption for data from ingester is in query.go
Memory consumption for data from store-gateway is in block_store_queryable.go

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/querier_test.go
+++ b/pkg/querier/querier_test.go
@@ -1311,48 +1311,6 @@ func (m *errDistributor) ActiveNativeHistogramMetrics(context.Context, []*labels
 	return nil, errDistributorError
 }
 
-type emptyDistributor struct{}
-
-func (d *emptyDistributor) LabelNamesAndValues(_ context.Context, _ []*labels.Matcher, _ cardinality.CountMethod) (*client.LabelNamesAndValuesResponse, error) {
-	return nil, errors.New("method is not implemented")
-}
-
-func (d *emptyDistributor) QueryStream(context.Context, *stats.QueryMetrics, model.Time, model.Time, ...*labels.Matcher) (client.CombinedQueryStreamResponse, error) {
-	return client.CombinedQueryStreamResponse{}, nil
-}
-
-func (d *emptyDistributor) QueryExemplars(context.Context, model.Time, model.Time, ...[]*labels.Matcher) (*client.ExemplarQueryResponse, error) {
-	return nil, nil
-}
-
-func (d *emptyDistributor) LabelValuesForLabelName(context.Context, model.Time, model.Time, model.LabelName, *storage.LabelHints, ...*labels.Matcher) ([]string, error) {
-	return nil, nil
-}
-
-func (d *emptyDistributor) LabelNames(context.Context, model.Time, model.Time, *storage.LabelHints, ...*labels.Matcher) ([]string, error) {
-	return nil, nil
-}
-
-func (d *emptyDistributor) MetricsForLabelMatchers(context.Context, model.Time, model.Time, *storage.SelectHints, ...*labels.Matcher) ([]labels.Labels, error) {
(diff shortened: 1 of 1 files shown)


pkg/querier/blocks_store_queryable.go
```diff
@@ -1015,6 +1019,12 @@ func (q *blocksStoreQuerier) receiveMessage(c BlocksStoreClient, stream storegat
...
 		mySeries = append(mySeries, s)
 
+		ls := mimirpb.FromLabelAdaptersToLabels(s.Labels)
+
+		if err := memoryTracker.IncreaseMemoryConsumptionForLabels(ls); err != nil {
+			return mySeries, myWarnings, myQueriedBlocks, myStreamingSeriesLabels, indexBytesFetched, false, false, err
+		}
+
 		// Add series fingerprint to query limiter; will return error if we are over the limit
 		if err := queryLimiter.AddSeries(mimirpb.FromLabelAdaptersToLabels(s.Labels)); err != nil {
```
charleskorn: Should we use `ls` here as well?
```suggestion
		if err := queryLimiter.AddSeries(ls); err != nil {
```

pkg/storage/series/memory_tracking_series_set.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (m *MemoryTrackingSeriesSet) At() storage.Series {
+	at := m.inner.At()
+
+	// Only decrease memory once per each single Next() call and multiple At() call
+	if !m.memoryDecreased {
+		// New series at this position
+		m.memoryDecreased = true
+		// The caller should IncreaseMemoryConsumption for labels if they want to retain the labels.
```
charleskorn, line 39: Could you please add a note here that in the future, we should avoid calling `DecreaseMemoryConsumptionForLabels` and make it the caller's responsibility?

pkg/storage/series/memory_tracking_series_set_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.Equal(t, series1, series)
+	memoryAfterThirdAt := tracker.CurrentEstimatedMemoryConsumptionBytes()
+	require.Equal(t, memoryAfterSecondAt, memoryAfterThirdAt, "multiple At() calls should not decrease memory multiple times")
+
+	// Second series
+	require.True(t, memTrackingSet.Next())
+	series = memTrackingSet.At()
+	require.Equal(t, series2, series)
+	memoryAfterSecondSeries := tracker.CurrentEstimatedMemoryConsumptionBytes()
+	require.Less(t, memoryAfterSecondSeries, memoryAfterThirdAt)
```
charleskorn: Shouldn't the memory consumption estimate be 0 at this point? (ie. can't we make this assertion stronger?)

pkg/querier/blocks_store_queryable.go
lamida: Good catch. Fixed in 5f9e03f66a213af8e5eef7914b57a77561da007b

pkg/storage/series/memory_tracking_series_set.go
lamida: Added in 07866eae629ca54a3a62241c5c5b6c7ee3b6abc6

pkg/storage/series/memory_tracking_series_set_test.go
lamida: Fixed in f6ee5d9743a3b26500e257719f0739deea6d387b

Issue #12010, opened 2025-07-07
Proposal: Mark etcd deprecated and offer migrations for HA tracker


dimitarvdimitrov:

56quarters:
Is this proposal only for HA tracking, not the ring?

dimitarvdimitrov:
there are already [some docs on using memberlist for the hash rings](https://grafana.com/docs/mimir/latest/configure/configure-hash-rings/#memberlist-default). The HA tracker was the only remaining component being blocked on a consistent KV store.

related to https://github.com/grafana/mimir/issues/1597

cc @NickAnge @aldernero

armandgrillet:
@dimitarvdimitrov  We need a clearer description of what needs to be done for this, we discussed this just now and aren't sure. This is one of the last Mimir 3.0 items.

dimitarvdimitrov:
added some details @armandgrillet

NickAnge:
For the `add a migration doc` we can use the steps we followed  ([Memberlist - Ha Tracker](https://docs.google.com/presentation/d/1mnn2STrmR_8ZYWYj5oIEFhq23BC-CgVaMnF-Dayx7u0/edit?slide=id.g3686ee06364_0_44#slide=id.g3686ee06364_0_44))

1. Use [multi kvStore](https://github.com/grafana/dskit/blob/main/kv/multi.go): 
   - Primary : Current one 
   - Secondary: Memberlist

2. Switch Memberlist to Primary kv Store

3.  Replace multiKv store with Memberlist

narqo:
There is a PR that addresses a small bit of the related docs https://github.com/grafana/mimir/pull/12857

56quarters:
Summarizing some internal discussion:

* We will leave support for using Consul or Etcd for the HA tracker in the Mimir code base.
* Use of Consul or Etcd for the HA tracker will _not_ be deprecated.
* However, memberlist will become the default for jsonnet and the helm chart.
* Jsonnet for using Consul or Etcd will be removed.

PR #12998, opened 2025-10-14
distributor: make memberlist default for HA tracker

dimitarvdimitrov:
Memberlist support for the HA tracker was added in Mimir 2.17. This PR updates the code documentation to indicate that memberlist is the recommended backend and that consul/etcd are deprecated as of Mimir 3.0.

Part of #12010.


pkg/distributor/ha_tracker.go
```diff
@@ -142,7 +142,9 @@ type HATrackerConfig struct {
 	// This is a potentially high cardinality metric, so it is disabled by default.
 	EnableElectedReplicaMetric bool `yaml:"enable_elected_replica_metric"`
 
-	KVStore kv.Config `yaml:"kvstore" doc:"description=Backend storage to use for the ring. Supported values are: consul, etcd, inmemory, memberlist, multi. Note that etcd is deprecated."`
+	// KVStore is the backend storage for the HA tracker. Memberlist is the recommended backend.
+	// Consul and etcd are deprecated for the HA tracker as of Mimir 3.0 and will be removed in a future release.
```
tacole02: It seems contradictory to state that `consul` and `etc` are supported values and then that they're deprecated in the next sentence. What if we said something like:

"Supported values: consul, etcd, inmemory, memberlist, multi. Memberlist is recommended. Consul and etcd are still available for backward compatibility but are considered deprecated for this purpose."

dimitarvdimitrov: yeah that makes sense. thanks for the callout! 

```suggestion
	// Supported values: consul, etcd, inmemory, memberlist, multi. Memberlist is recommended. Consul and etcd are still available for backward compatibility but are considered deprecated for this purpose.
```

docs/sources/mimir/configure/configuration-parameters/index.md
narqo: (nit) Why do we still set `consul` as default, even though it was deprecated?

dimitarvdimitrov: oooohhhh, that's something we should change too i think

dimitarvdimitrov: done in https://github.com/grafana/mimir/pull/12998/commits/117e6097ab81bff2b6199799be7f976ec4f7c291

narqo (approved):
It's even better 🔥

tacole02 (commented):
Great work!

PR #12923, opened 2025-10-07
[distributor]: add support for reactive limiters

duricanikolic:
#### What this PR does

 This PR introduces a reactive limiter feature for the distributor that automatically adjusts the maximum number of inflight requests based on observed response times and system performance. This helps prevent overload and maintain optimal throughput.

This is an experimental feature and can be configured via the following experimental configuration options:
- `-distributor.reactive-limiter.enabled`: Enables or disables reactive limiters.
- `-distributor.reactive-limiter.min-limit`: Specifies the minimum inflight requests limit.
- `-distributor.reactive-limiter.max-limit`: Specifies the maximum inflight requests limit.
- `-distributor.reactive-limiter.initial-limit`: Specifies the initial inflight requests limit.

Additionally, this PR adds the following metrics:
- `cortex_distributor_reactive_limiter_inflight_limit`: A gauge showing the current inflight request limit calculated by the reactive limiter.
- `cortex_distributor_reactive_limiter_inflight_requests`: A gauge showing the current number of inflight requests being tracked by the reactive limiter.
- `cortex_distributor_reactive_limiter_queued_requests`: A gauge showing the current number of queued requests waiting to be processed by the reactive limiter.

#### Which issue(s) this PR fixes or relates to

Fixes #<issue number>

#### Checklist

- [x] Tests updated.
- [x] Documentation added.
- [x] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.
- [x] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features.

PR #13003, opened 2025-10-14
[distributor]: add support for reactive limiters for in HTTP calls

duricanikolic:
#### What this PR does

This PR extends the reactive limiter feature introduced in https://github.com/grafana/mimir/pull/12923 to support HTTP calls in addition to gRPC calls. It ensures that reactive limiters properly protect push operations across both transport protocols.

Additionally, this PR changes the gRPC status code for errors with cause `mimirpb.ERROR_CAUSE_INSTANCE_LIMIT` from `codes.Internal` to `codes.Unavailable`. This better reflects the transient nature of reactive limiter rejections. It also aligns with standard gRPC semantics where `codes.Unavailable` indicates temporary service unavailability.

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


pkg/distributor/distributor.go
```diff
@@ -1810,6 +1845,13 @@ func (d *Distributor) limitsMiddleware(next PushFunc) PushFunc {
 			return middleware.DoNotLogError{Err: err}
 		}
 
+		defer func() {
+			rs.pushErr = retErr
+		}()
```
jhalterman: I don't think this matters in practice, but it's probably better to have this defer statement after the `acquireReactiveLimiterPermit` call below. That way we only set `rs.pushErr` is we acquired a permit.

duricanikolic: I have moved the defer statement after the call to `acquireReactiveLimiterPermit`. Thank you

PR #13008, opened 2025-10-14
helm: remove GEM license mount on all components

dimitarvdimitrov:
based on #13007 this removes the license volume on all components


narqo (approved):
🔥

PR #13009, opened 2025-10-14
helm: remove usage of enterprise values

dimitarvdimitrov:
based on #13008 this removes any usage of enterprise-related values from any template in the chart

the idea is to show that in reality very little changes in our rendered manifests for non-enterprise charts (see offline tests). The two offline tests that change have enterprise enabled


tacole02 (commented):
Changelog LGTM

PR #13010, opened 2025-10-14
helm: remove references to GEM from values files

dimitarvdimitrov:
based on #13009 this removes any references to GEM and GEM components from the values files in the chart. Probably the biggest change here is that we're removing the buckets from the minio setup config

### DIFF
--- a/operations/helm/charts/mimir-distributed/CHANGELOG.md
+++ b/operations/helm/charts/mimir-distributed/CHANGELOG.md
@@ -29,7 +29,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## main / unreleased
 
-* [CHANGE] Remove all remaining GEM (enterprise) references from test configurations, reference files, and values.yaml. #13005 #13006 #13007 #13008 #13009
+* [CHANGE] Remove all remaining GEM (enterprise) references from test configurations, reference files, and values.yaml. #13005 #13006 #13007 #13008 #13009 #13010
 * [ENHANCEMENT] Added `trafficDistribution` configuration option to distributor, gateway, querier, query-frontend, query-scheduler, ingester, and store-gateway services to enable same-zone traffic routing. #12243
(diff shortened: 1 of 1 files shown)


operations/compare-helm-with-jsonnet/helm/00-base/values.yaml
```diff
@@ -20,7 +20,6 @@ serviceAccount:
 # The Jsonnet bundle doesn't provide any gateway or ingress.
 gateway:
     enabled: false
-    enabledNonEnterprise: false
```
narqo, line 23: 👍

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -266,53 +223,13 @@ mimir:
       cache_results: true
       query_sharding_target_series_per_shard: 2500
       {{- end }}
-      {{- if not .Values.federation_frontend.disableOtherComponents }}
       scheduler_address: {{ template "mimir.fullname" . }}-query-scheduler-headless.{{ .Release.Namespace }}.svc:{{ include "mimir.serverGrpcListenPort" . }}
-      {{- end }}
-      {{- if .Values.enterprise.enabled }}
-      log_query_request_headers: X-Access-Policy-Name,X-Token-Name
-      {{- end }}
```
narqo, line 274: LOL, I still remember adding these 🙃

operations/helm/charts/mimir-distributed/values.yaml
```diff
@@ -3350,11 +3204,10 @@ gateway:
     annotations: {}
     # -- Labels for the Service
     labels: {}
-    # -- DEPRECATED Legacy compatibility port the GEM gateway service listens on, set to 'null' to disable
+    # -- DEPRECATED Legacy compatibility port, set to 'null' to disable
     legacyPort: 8080
```
narqo, line 3208: (nit) we could remove the `legacyPort` from the gateway's service.

We can do that separately, later.

PR #13014, opened 2025-10-14
Base tooling images on distroless/static-debian12

pr00se:
#### What this PR does

We moved Mimir to distroless a while ago, but the tooling images (`metaconvert`, `mimirtool`, etc.) are still based on `alpine`. This bases them on `distroless/static-debian12` like the main `mimir` image.

Fixes #<issue number>

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/cmd/metaconvert/Dockerfile
+++ b/cmd/metaconvert/Dockerfile
@@ -3,9 +3,12 @@
 # Provenance-includes-license: Apache-2.0
 # Provenance-includes-copyright: The Cortex Authors.
 
-FROM       alpine:3.22.2
-ARG        EXTRA_PACKAGES
-RUN        apk add --no-cache ca-certificates tzdata $EXTRA_PACKAGES
+ARG        BASEIMG=gcr.io/distroless/static-debian12
+# TODO(rwwiv): Remove once CA bundle is updated upstream.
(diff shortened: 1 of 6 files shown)


cmd/metaconvert/Dockerfile
```diff
@@ -3,16 +3,21 @@
 # Provenance-includes-license: Apache-2.0
 # Provenance-includes-copyright: The Cortex Authors.
 
-FROM       alpine:3.22.2
-ARG        EXTRA_PACKAGES
-RUN        apk add --no-cache ca-certificates tzdata $EXTRA_PACKAGES
+ARG        BASEIMG=gcr.io/distroless/static-debian12
+# TODO(rwwiv): Remove once CA bundle is updated upstream.
```
56quarters, line 7: This has since been resolved since `gcr.io/distroless/static-debian13` is available. Understandable if you don't want to do that as part of this PR.

tacole02 (commented):
Changelog LGTM

cmd/metaconvert/Dockerfile
pr00se: i'd rather just upgrade everything at once, since we probably want to upgrade the mimir image too

PR #13015, opened 2025-10-15
EliminateDeduplicateAndMerge: support for DelatedNameRemoval

Konstantinov-Innokentii:
#### What this PR does
PR introduce support of DelayedNameRemoval to the EliminateDeduplicateAndMerge optimization.

When DelayedNameRemoval is enabled, the name label is kept until the end of query execution. This means that we need to keep DeduplicateAndMerge closest to the root, not the selector, unlike when DelayedNameRemoval is disabled.

#### Checklist

- [x] Tests updated.
- ~[ ] Documentation added.~
- ~[ ] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`. If changelog entry is not needed, please add the `changelog-not-needed` label to the PR.~

### DIFF
--- a/pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
+++ b/pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
@@ -4,7 +4,6 @@ package plan
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/prometheus/prometheus/model/labels"
 
@@ -46,7 +45,7 @@ func NewEliminateDeduplicateAndMergeOptimizationPass(enableDelayedNameRemoval bo
 }
 
 func (e *EliminateDeduplicateAndMergeOptimizationPass) Name() string {
-	return fmt.Sprintf("Eliminate DeduplicateAndMerge (enableDelayedNameRemoval: %t)", e.enableDelayedNameRemoval)
+	return "Eliminate DeduplicateAndMerge"
 }
 
 func (e *EliminateDeduplicateAndMergeOptimizationPass) Apply(ctx context.Context, plan *planning.QueryPlan, _ planning.QueryPlanVersion) (*planning.QueryPlan, error) {


pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
```diff
@@ -78,49 +86,71 @@ func (e *EliminateDeduplicateAndMergeOptimizationPass) collectNodesToRemove(node
...
-			*nodesToRemove = nil
+		switch len(*nodes) {
+		case 0:
+		case 1:
+			(*nodes)[0].keep = true
+		default:
+			if e.enableDelayedNameRemoval {
+				// Also keep the root DeduplicateAndMerge to handle final deduplication, if delayed name removal is enabled.
+				// Name is dropped at the very end of query execution and label_replace or label_join might mess with it even if selector guarantees unique series.
+				// TODO - I'm not really sure about keeping the root one, corresponding test case is "function call enclosing label_replace, selector with exact name matcher".
```
Konstantinov-Innokentii: More context. The case I'm unsure about is this expression:
`abs(label_replace(rate(foo[5m])`
Plan looks like this
```
//
//                          ┌──────────────┐
//                          │    Dedup?     │  ◄─────── Skip this dedup? 
//                          └──────┬───────┘          
//                                 │                 
//                          ┌──────▼───────┐          
//                          │   DropName   │          
//                          └──────┬───────┘         
//                                 │                 
//                          ┌──────▼───────┐         
//                          │     abs      │         
//                          └──────┬───────┘         
//                                 │                 
//                          ┌──────▼───────┐         
//                          │    Dedup     │ ◄─────── Deduplicates possible duplicates after label_replace mess with labels.
//                          └──────┬───────┘
//                                 │
//                          ┌──────▼────────┐
//                          │ label_replace │ 
//                          └──────┬────────┘ 
//                                 │           
//                          ┌──────▼───────┐   
//                          │     ̶D̶e̶d̶u̶p̶ ̶    │ ◄─────── This Dedup is skipped, selector has exact name matcher
//                          └──────┬───────┘
//                                 │
//                          ┌──────▼───────┐
//                          │    rate()    │
//                          └──────┬───────┘
//                                 │
//                          ┌──────▼───────┐
//                          │   selector   │ ─┐ foo{region="singapore"} 1 2 3
//                          └──────────────┘  └ foo{region="melbourne"} 4 5 6
```

What I'm trying to figure out is that if label_replace/label_join can somehow convert
```
foo{region="singapore"} 1 2 3
foo{region="melbourne"} 4 5 6
```
to 
```
foo{region="singapore"} 1 2 3
bar{region="singapore"} 4 5 6
```
Which requires to keep Dedup node after  a name is dropped.

My understading is that it's **NOT** possible within one label_replace/label_join call.
And if there will be two operations:
**Initial:**
```
foo{region="singapore"} 1 2 3
foo{region="melbourne"} 4 5 6
```
label_replace A:
```
foo{region="singapore"} 1 2 3
foo{region="singapore"} 4 5 6
```
label_replace B:
```
foo{region="singapore"} 1 2 3
bar{region="singapore"} 4 5 6
```
Then DedupNode after label_replace A will detect duplicates and throw an error.
Looking for counterarguments, but the PR currently implements a "conservative" version, which keeps extra Deduplication.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
```diff
@@ -78,49 +86,69 @@ func (e *EliminateDeduplicateAndMergeOptimizationPass) collectNodesToRemove(node
...
-		if len(*nodesToRemove) >= 1 {
-			*nodesToRemove = (*nodesToRemove)[:len(*nodesToRemove)-1]
+		if e.enableDelayedNameRemoval {
+			// With Delayed Name Removal name is dropped at the very end of query execution.
+			// Keep the DeduplicateAndMerge closest to root to handle final deduplication.
+			if len(*nodes) > 0 {
+				(*nodes)[0].keep = true
+			}
+		} else {
+			// Without delayed name removal name is dropped immediately, so there should be no duplicates after..
```
charleskorn:
```suggestion
			// Without delayed name removal name is dropped immediately, so there should be no duplicates after.
```

pkg/streamingpromql/planning.go
```diff
@@ -63,13 +63,7 @@ func NewQueryPlanner(opts EngineOpts) (*QueryPlanner, error) {
 	// This optimization pass is registered before CSE to keep the query plan as a simple tree structure.
 	// After CSE, the query plan may no longer be a tree due to multiple paths culminating in the same Duplicate node,
 	// which would make the elimination logic more complex.
-	if opts.EnableEliminateDeduplicateAndMerge {
```
charleskorn, line 68: Shouldn't we still check for `opts.EnableEliminateDeduplicateAndMerge` here?

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -27,116 +27,221 @@ import (
...
 func TestEliminateDeduplicateAndMergeOptimizationPassPlan(t *testing.T) {
 	testCases := map[string]struct {
-		expr            string
-		expectedPlan    string
-		nodesEliminated int // Number of DeduplicateAndMerge nodes expected to be eliminated
+		expr                               string
+		expectedPlanWithoutDelayedName     string
+		expectedPlanWithDelayedNameRemoval string
+		nodesEliminatedWithoutDelayed      int
+		nodesEliminatedWithDelayed         int
```
charleskorn: [nit] To keep things consistent with `expectedPlanWithDelayedNameRemoval`:

```suggestion
		expectedPlanWithoutDelayedNameRemoval    string
		expectedPlanWithDelayedNameRemoval       string
		nodesEliminatedWithoutDelayedNameRemoval int
		nodesEliminatedWithDelayedNameRemoval    int
```

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -173,44 +304,152 @@ func TestEliminateDeduplicateAndMergeOptimizationPassPlan(t *testing.T) {
...
-			nodesEliminated: 0,
+			// should keep:
+			// - DeduplicateAndMerge around label_replace - to deduplicate after label_replace might modify labels and therefore potentially introduce duplicates.
+			// - DeduplicateAndMerge at the root - to deduplicate after __name__ is dropped, because selector without exact name matcher might produce non-unique series.
+			// drops:
+			// - DeduplicateAndMerge wrapping rate is dropped, because despite selector has NO exact name matcher, name is not dropped immediately and therefore duplicates won't appear.
+			expectedPlanWithDelayedNameRemoval: `
+					- DeduplicateAndMerge
+						- DropName
+							- FunctionCall: label_replace(...)
```
charleskorn: The comment here seems to be different to the expected plan.

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge_test.go
```diff
@@ -238,23 +501,106 @@ func TestEliminateDeduplicateAndMergeOptimizationPassPlan(t *testing.T) {
...
+				- DeduplicateAndMerge
+					- DropName
+						- BinaryExpression: LHS or RHS
+							- LHS: VectorSelector: {__name__="foo"}
+							- RHS: VectorSelector: {__name__="bar"}
+				`,
+			nodesEliminatedWithoutDelayed: 0,
+			nodesEliminatedWithDelayed:    0,
 		},
 		"or operator with functions - should keep DeduplicateAndMerge": {
```
charleskorn, line 601: Might be worth adding a note here that we could remove the `DeduplicateAndMerge` nodes on each side, but we currently don't for simplicity.

pkg/streamingpromql/planning.go
Konstantinov-Innokentii: thanks for catch

Konstantinov-Innokentii:
@charleskorn I adressed all the comments, do you mind to take a second look?

pkg/streamingpromql/optimize/plan/eliminate_deduplicate_and_merge.go
```diff
@@ -33,20 +36,24 @@ type dedupNodeInfo struct {
...
-	return &EliminateDeduplicateAndMergeOptimizationPass{}
+func NewEliminateDeduplicateAndMergeOptimizationPass(enableDelayedNameRemoval bool) *EliminateDeduplicateAndMergeOptimizationPass {
+	return &EliminateDeduplicateAndMergeOptimizationPass{
+		enableDelayedNameRemoval: enableDelayedNameRemoval,
+	}
 }
 
 func (e *EliminateDeduplicateAndMergeOptimizationPass) Name() string {
-	return "Eliminate DeduplicateAndMerge"
+	return fmt.Sprintf("Eliminate DeduplicateAndMerge (enableDelayedNameRemoval: %t)", e.enableDelayedNameRemoval)
```
charleskorn: We generally don't put configuration information like this in the optimisation pass `Name`, as this is used for metrics.

Konstantinov-Innokentii: Fixed, thanks

charleskorn (approved):

PR #13017, opened 2025-10-15
Remote execution: querier backwards compatibility support

charleskorn:
#### What this PR does

This PR changes the behaviour of query-frontends to only generate query plans that the running set of queriers support.

This ensures that query-frontends don't start generating plans that queriers don't support during rollouts until all queriers have upgraded.

Queriers now join a ring and report the maximum supported query plan version as part of their instance description, and query-frontends then pass this to the query planner. 

Optimisations passes can then choose to skip optimisations that introduce functionality that is not supported by the running set of queriers.

Not included in this PR, but will be added separately: alerts to warn when the cluster is in an inconsistent state (eg. query-frontends are reporting a maximum supported version that does not match what queriers support).

Don't be put off by the large number of files changed: most of these are the regenerated Helm or Jsonnet test files, or files that needed to be updated due to the changed signature of `AddIngester` from the dskit PR.

:warning: This PR depends on https://github.com/grafana/dskit/pull/757.

https://github.com/grafana/mimir/pull/12808, https://github.com/grafana/mimir/pull/12961

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.


pkg/streamingpromql/planning/plan.proto
```diff
@@ -20,7 +20,7 @@ message EncodedQueryPlan {
 
   bool enableDelayedNameRemoval = 5;
 
-  int64 version = 6;
+  uint64 version = 6;
```
charleskorn, line 23: Note to reviewers: this change is considered safe given the only values we set this to at present are 0 or 1.

Docs: https://protobuf.dev/programming-guides/proto3/#conditionally-safe-changes

pkg/streamingpromql/planning.go
```diff
@@ -245,6 +259,12 @@ func (p *QueryPlanner) NewQueryPlan(ctx context.Context, qs string, timeRange ty
 		return nil, err
 	}
 
+	if plan.Version > maximumSupportedQueryPlanVersion {
+		return nil, fmt.Errorf("maximum supported query plan version is %d, but generated plan version is %d - this is a bug", maximumSupportedQueryPlanVersion, plan.Version)
+	}
```
tcp13equals2, line 264: Perhaps emitting a metric here would be interesting to track if we do ever reach this code path?

pkg/querier/querier_ring.go
```diff
@@ -0,0 +1,177 @@
...
+		HeartbeatPeriod:                 cfg.Common.HeartbeatPeriod,
+		HeartbeatTimeout:                cfg.Common.HeartbeatTimeout,
+		TokensObservePeriod:             0,
+		NumTokens:                       ringNumTokens,
+		KeepInstanceInTheRingOnShutdown: false,
+		HideTokensInStatusPage:          true,
+		ShowVersionsInStatusPage:        true,
+		ComponentNames:                  querierRingComponentNames,
+		Versions: ring.InstanceVersions{
+			MaximumSupportedQueryPlanVersion: uint64(planning.MaximumSupportedQueryPlanVersion),
```
lamida: I believe we don't need this type conversion.

```suggestion
			MaximumSupportedQueryPlanVersion: planning.MaximumSupportedQueryPlanVersion,
```

pkg/querier/querier_ring.go
```diff
@@ -0,0 +1,177 @@
...
+		Help: "The maximum supported query plan version this process was compiled to support.",
+	}).Set(float64(planning.MaximumSupportedQueryPlanVersion))
+
+	return provider
+}
+
+var queryPlanVersioningOp = ring.NewOp([]ring.InstanceState{ring.ACTIVE, ring.PENDING, ring.JOINING, ring.LEAVING}, nil)
+var errQuerierHasNoSupportedQueryPlanVersion = fmt.Errorf("one or more queriers in the ring is not reporting a supported query plan version")
+
+func (r *RingQueryPlanVersionProvider) GetMaximumSupportedQueryPlanVersion(ctx context.Context) (uint64, error) {
```
lamida: Do we need `ctx` at all here? It is currently not used.

pkg/streamingpromql/planning_test.go
```diff
@@ -1352,7 +1382,7 @@ func BenchmarkPlanEncodingAndDecoding(b *testing.B) {
 	}
 
 	opts := NewTestEngineOpts()
-	planner, err := NewQueryPlanner(opts)
+	planner, err := NewQueryPlanner(opts, NewMaximumSupportedVersionQueryPlanVersionProvider())
```
lamida, line 1625: I see this constructor call with `NewMaximumSupportedVersionQueryPlanVersionProvider` all over the tests, shouldn't we create something like `NewQueryPlannerWithNewQueryPlanVersionProvider` constructor? 

I also want to suggest something similar with `NewQueryPlannerWithoutOptimizationPasses` -> `NewQueryPlannerWithoutOptimizationPassesWithNewQueryPlanVersionProvider` but maybe it becomes too mouthful, unless we shorten the constructor name in some way.

pkg/mimir/mimir.go
```diff
@@ -1019,6 +1021,14 @@ func (t *Mimir) Run() error {
 		t.API.RegisterIngesterRing(t.Ingester.RingHandler())
 	}
 
+	// Similar to above, register the querier ring handler if it exists,
```
lamida: Nits, I don't think the "Similar to above" give any relevant information to what we are going to do below here. Also when saying "above" that might be get outdated as the code get reordered or refactored in the future. Hence I suggest to just remove the reference.

```suggestion
	// Register the querier ring handler if it exists,
```

pkg/mimir/mimir_test.go
```diff
@@ -166,10 +166,18 @@ func TestMimir(t *testing.T) {
...
+					KVStore: kv.Config{
+						Store: "memberlist",
+					},
+					InstanceAddr: "test:8080",
+				},
+			},
 		},
 		Frontend: frontend.CombinedFrontendConfig{
-			QueryEngine: "prometheus",
+			QueryEngine: "mimir",
```
lamida, line 180: Why do we change this? It seem the test doesn't check this at all.

pkg/mimir/modules.go
```diff
@@ -984,8 +1016,19 @@ func (t *Mimir) initQueryFrontendQueryPlanner() (services.Service, error) {
 	reg := prometheus.WrapRegistererWith(prometheus.Labels{"component": "query-frontend"}, t.Registerer)
 	_, mqeOpts := engine.NewPromQLEngineOptions(t.Cfg.Querier.EngineConfig, t.ActivityTracker, util_log.Logger, reg)
 
+	var versionProvider streamingpromql.QueryPlanVersionProvider
+
+	if t.Cfg.Frontend.QueryMiddleware.EnableRemoteExecution {
+		versionProvider = querier.NewRingQueryPlanVersionProvider(t.QuerierRing, t.Registerer)
+	} else {
+		// If we're not using remote execution, then the query plans we generate in the query-frontend will
```
lamida: Nits

```suggestion
		// If remote execution is not enabled, then the query plans we generate in the query-frontend will
```

pkg/streamingpromql/planning/plan.go
```diff
@@ -41,7 +41,7 @@ type QueryPlan struct {
 	//
 	// Queriers use this to ensure they do not attempt to execute a query plan that contains features they
 	// cannot safely or correctly execute (eg. new nodes or new meaning for existing node details).
-	Version int64
+	Version uint64
```
lamida, line 45: Not necessary to be applied in this PR, but should we have a type alias for this?
```
  // QueryPlanVersion represents the version of a query plan's structure and supported features.
  type QueryPlanVersion uint64
```

That should be helpful for type safety because we use query plan version everywhere like 

*  Apply(ctx context.Context, plan *planning.QueryPlan, maximumSupportedQueryPlanVersion uint64)
* GetMaximumSupportedQueryPlanVersion(ctx context.Context) (uint64, error
* etc

It will be safer and also clearer when reading code if the version is specific domain concept and hence we can prevent accidentally passing raw uint64 not intended for this. It is also easier to navigate codes for future maintainer.

Furthermore we can provide method for convenience like

```
  // For convenience
  func (v QueryPlanVersion) String() string {
      return strconv.FormatUint(uint64(v), 10)
  }
```
and even more.

There might be performance overhead but I believe will be minimum.

lamida (dismissed):
My main comment is to consider using type alias for Query plan version. Although that can be applied as separate PR. Other than that I left mostly minor nits comments.

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
...
+  -querier.ring.instance-enable-ipv6
+    	Enable using a IPv6 instance address. (default false)
+  -querier.ring.instance-id string
+    	Instance ID to register in the ring. (default "<hostname>")
+  -querier.ring.instance-interface-names string
+    	List of network interface names to look up when finding the instance IP address. (default [<private network interfaces>])
+  -querier.ring.instance-port int
+    	Port to advertise in the ring (defaults to -server.grpc-listen-port).
+  -querier.ring.multi.mirror-enabled
+    	Mirror writes to secondary store.
```
tacole02, line 2197:
```suggestion
    	Mirror writes to the secondary store.
```

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
...
+  -querier.ring.multi.mirror-timeout duration
+    	Timeout for storing value to secondary store. (default 2s)
```
tacole02, line 2199:
```suggestion
    	Timeout for storing a value to the secondary store. (default 2s)
```

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
...
+  -querier.ring.etcd.tls-min-version string
+    	Override the default minimum TLS version. Allowed values: VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13
+  -querier.ring.etcd.tls-server-name string
+    	Override the expected name on the server certificate.
+  -querier.ring.etcd.username string
+    	Etcd username.
+  -querier.ring.heartbeat-period duration
+    	Period at which to heartbeat to the ring. 0 = disabled. (default 15s)
+  -querier.ring.heartbeat-timeout duration
+    	The heartbeat timeout after which queriers are considered unhealthy within the ring. 0 = never (timeout disabled). (default 1m0s)
```
tacole02, line 2185:
```suggestion
    	Heartbeat timeout after which Mimir marks a querier unhealthy in the ring. Set to 0 to disable. (default 1m0s)
```

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
...
+  -querier.ring.instance-addr string
+    	IP address to advertise in the ring. Default is auto-detected.
+  -querier.ring.instance-enable-ipv6
+    	Enable using a IPv6 instance address. (default false)
+  -querier.ring.instance-id string
+    	Instance ID to register in the ring. (default "<hostname>")
+  -querier.ring.instance-interface-names string
+    	List of network interface names to look up when finding the instance IP address. (default [<private network interfaces>])
+  -querier.ring.instance-port int
+    	Port to advertise in the ring (defaults to -server.grpc-listen-port).
```
tacole02, line 2195:
```suggestion
    	Port to advertise in the ring (defaults to `-server.grpc-listen-port`).
```

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
...
+  -querier.ring.etcd.username string
+    	Etcd username.
+  -querier.ring.heartbeat-period duration
+    	Period at which to heartbeat to the ring. 0 = disabled. (default 15s)
+  -querier.ring.heartbeat-timeout duration
+    	The heartbeat timeout after which queriers are considered unhealthy within the ring. 0 = never (timeout disabled). (default 1m0s)
+  -querier.ring.instance-addr string
+    	IP address to advertise in the ring. Default is auto-detected.
+  -querier.ring.instance-enable-ipv6
+    	Enable using a IPv6 instance address. (default false)
```
tacole02, line 2189:
```suggestion
    	Enable using an IPv6 instance address.
```

cmd/mimir/help-all.txt.tmpl
```diff
@@ -2135,6 +2135,74 @@ Usage of ./cmd/mimir/mimir:
     	The time after which a metric should be queried from storage and not just ingesters. 0 means all queries are sent to store. If this option is enabled, the time range of the query sent to the store-gateway will be manipulated to ensure the query end is not more recent than 'now - query-store-after'. (default 12h0m0s)
   -querier.response-streaming-enabled
     	[experimental] Enables streaming of responses from querier to query-frontend for response types that support it (currently only `active_series` responses do).
+  -querier.ring.auto-forget-unhealthy-periods int
+    	Number of consecutive timeout periods an unhealthy instance in the ring is automatically removed after. Set to 0 to disable auto-forget. (default 10)
```
tacole02, line 2141:
```suggestion
    	Number of consecutive timeout periods after which Mimir automatically removes an unhealthy instance from the ring. Set to 0 to disable auto-forget. (default 10)
```

tacole02 (commented):
Docs look good! I left a few minor suggestions. Thank you!

pkg/streamingpromql/planning_test.go
charleskorn: > I see this constructor call with `NewMaximumSupportedVersionQueryPlanVersionProvider` all over the tests, shouldn't we create something like `NewQueryPlannerWithNewQueryPlanVersionProvider` constructor?

I'd prefer to keep things in the tests explicit. If this becomes too annoying in the future, we could introduce a helper method for tests like `NewQueryPlannerForTests` or something like that.

pkg/mimir/mimir_test.go
charleskorn: This test is intended to run with the default Mimir config. You are correct that the test doesn't check this at all, but I changed this to the current default value as part of "leaving the campsite cleaner than I found it" 🙂

pkg/streamingpromql/planning/plan.go
charleskorn: Good idea, but yep, I'd like to keep this for a future PR to keep this PR as small as possible, it's already quite large.

pkg/querier/querier_ring.go
charleskorn: Good catch, removed in c0efe33.

pkg/querier/querier_ring.go
charleskorn: I've made this clearer in c0efe33.

pkg/mimir/mimir.go
charleskorn: Updated in 0cb0d3e.

pkg/mimir/modules.go
charleskorn: Addressed in f0adfa4.

charleskorn:
> Docs look good! I left a few minor suggestions. Thank you!

All of the suggestions make sense to me, but are coming from a shared place that affects more than just the new parameters added in this PR. I'd prefer to address these in a separate PR, as any changes will affect all instances of these parameters.

pkg/streamingpromql/planning.go
charleskorn: If we hit this code path, it'll be visible in the request error metrics emitted by queriers, as well as available in logs that we can derive metrics from.

dimitarvdimitrov:

lamida (approved):

pkg/streamingpromql/planning/plan.go
charleskorn: Done in https://github.com/grafana/mimir/pull/13119.

charleskorn:
> > Docs look good! I left a few minor suggestions. Thank you!
> 
> All of the suggestions make sense to me, but are coming from a shared place that affects more than just the new parameters added in this PR. I'd prefer to address these in a separate PR, as any changes will affect all instances of these parameters.

These should be covered by https://github.com/grafana/dskit/pull/761 and https://github.com/grafana/mimir/pull/13125.

PR #13024, opened 2025-10-15
Fix bug in matcher reducer planner which would remove all wildcard matchers

chencs:
#### What this PR does
This fixes an edge case where, if all matchers are wildcard matchers (e.g., `foo=~".*"`), they can all be excluded from the result set, which can produce inconsistent query results. 

Relates to https://github.com/grafana/mimir/pull/12831

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/lookupplan/matcher_reducer_planner.go
+++ b/pkg/ingester/lookupplan/matcher_reducer_planner.go
@@ -72,7 +72,13 @@ func buildOutMatchers(inMatchers []*labels.Matcher, allowedOutMatchers []*labels
 	for _, m := range allowedOutMatchers {
 		allowedInResultSet[m.String()] = false
 	}
-	for _, m := range dedupedMatchers {
+	// If we have reached the last deduped input matcher and are still not returning any matchers,
+	// we should return at least one matcher. This can happen if all input matchers are wildcard matchers.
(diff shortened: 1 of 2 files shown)


pkg/ingester/lookupplan/matcher_reducer_planner.go
```diff
@@ -72,7 +72,13 @@ func buildOutMatchers(inMatchers []*labels.Matcher, allowedOutMatchers []*labels
 	for _, m := range allowedOutMatchers {
 		allowedInResultSet[m.String()] = false
 	}
-	for _, m := range dedupedMatchers {
+	// If we have reached the last deduped input matcher and are still not returning any matchers,
+	// we should return at least one matcher. This can happen if all input matchers are wildcard matchers.
+	for i, m := range dedupedMatchers {
+		if i == len(dedupedMatchers)-1 && len(outMatchers) == 0 {
+			outMatchers = append(outMatchers, m)
```
dimitarvdimitrov, line 79: in this case would `dropped` be wrong?

chencs: Nope, we only add things to `dropped` if they're removed during deduping or we've already iterated over the matcher.

PR #13026, opened 2025-10-16
chore: remove non-streaming code from block store queryable

56quarters:
#### What this PR does

Since #12818, store-gateways no longer send non-streaming series in any responses. Remove code for handling non-streaming series from the queryable used with store-gateways.

Part of #12673

#### Note to reviewers

Most of the line count here is that I needed to make a lot of changes to `block_store_queryable_test.go`. The way it previously worked is that test fixtures were defined in terms of expected non-streaming responses. For streaming tests, the responses were converted to the expected streaming form (series, hints, chunks split over multiple responses). I changed this now that we don't use non-streaming responses to be a little more intuitive. 

#### Checklist

- [X] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/blocks_store_queryable_test.go
+++ b/pkg/querier/blocks_store_queryable_test.go
@@ -1651,19 +1651,16 @@ func TestBlocksStoreQuerier_Select(t *testing.T) {
 		t.Run(testName, func(t *testing.T) {
 			reg := prometheus.NewPedanticRegistry()
 
-			// Count the number of series to check the stats later.
-			// We also make a copy of the testData.storeSetResponses where relevant so that
-			// we can run the streaming and non-streaming case in any order.
-			var storeSetResponses []interface{}
+			// Count the number of series and chunks to check the stats later.
 			seriesCount, chunksCount := 0, 0
 			for _, res := range testData.storeSetResponses {
 				m, ok := res.(map[BlocksStoreClient][]ulid.ULID)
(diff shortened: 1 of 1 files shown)


dimitarvdimitrov:

pkg/querier/blocks_store_queryable_test.go
```diff
@@ -1701,146 +1649,139 @@ func TestBlocksStoreQuerier_Select(t *testing.T) {
...
-					for _, res := range testData.storeSetResponses {
-						m, ok := res.(map[BlocksStoreClient][]ulid.ULID)
-						if !ok {
-							storeSetResponses = append(storeSetResponses, res)
-							continue
+			reg := prometheus.NewPedanticRegistry()
+
+			// Count the number of series to check the stats later.
+			// We also make a copy of the testData.storeSetResponses where relevant so that
+			// we can run the streaming and non-streaming case in any order.
```
charleskorn: Is this still relevant? Do we still need to make a copy of the data here?

charleskorn (approved):
LGTM modulo possible simplification of one test:

pkg/querier/blocks_store_queryable_test.go
56quarters: Tests still pass and the race detector doesn't complain. Fixed in 7830b8190296dd6c3617937d7e6836feb08adad8

PR #13027, opened 2025-10-16
Fix two more sharding bugs

charleskorn:
#### What this PR does

This PR fixes two issues in sharding.

##### Queries containing a function call where some arguments are shardable and some are not fail

A query like `clamp_max(sum(foo), scalar(bar))` would fail with `shardedQuerier is expecting embedded queries but didn't find any`.

This is another case similar to https://github.com/grafana/mimir/pull/12958. Prior to https://github.com/grafana/mimir/pull/12785, this wouldn't be an issue: the remaining `scalar(bar)` would have been rewritten into an embedded query by `subtreeFolder`, so it would still have been evaluated without sharding. However, with https://github.com/grafana/mimir/pull/12785, `subtreeFolder `has been removed and so `scalar(bar)` would have been left as-is, resulting in the query failing with `shardedQuerier is expected embedded queries but didn't find any`. 

Unlike #12958, where the fix was to apply sharding to the aggregated parameter, here the fix is to not shard the expression at all. A future improvement would be to shard the shardable part of the expression and retrieve the scalar arguments separately, but that's out of scope for this fix.

##### Unshardable queries fail if the `query_sharding_max_sharded_queries` limit is 0

Even with the fix above, queries like `scalar(bar)` would also fail with `shardedQuerier is expecting embedded queries but didn't find any` if the `query_sharding_max_sharded_queries` limit is 0.

This is because we weren't checking if the query was shardable if the limit was 0, and so assumed the query should be sharded into the default number of shards (`query_sharding_total_shards`).

Again, the removal of `subtreeFolder` in https://github.com/grafana/mimir/pull/12785 uncovered the incorrect logic here: previously, the unshardable expression would have been rewritten into an embedded query by `subtreeFolder`, but with that removed, the query was not being rewritten.

https://github.com/grafana/mimir/pull/12785

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/querymiddleware/querysharding.go
+++ b/pkg/frontend/querymiddleware/querysharding.go
@@ -271,6 +271,7 @@ func NewQuerySharder(
 func (s *QuerySharder) Shard(ctx context.Context, tenantIDs []string, expr parser.Expr, requestedShardCount int, seriesCount *EstimatedSeriesCount, totalQueries int32) (parser.Expr, error) {
 	log := spanlogger.FromContext(ctx, s.logger)
 	log.DebugLog("msg", "attempting query sharding", "query", expr)
+	s.metrics.shardingAttempts.Inc()
 
 	totalShards, err := s.getShardsForQuery(ctx, tenantIDs, expr, requestedShardCount, seriesCount, totalQueries, log)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lamida (approved):
Still have my eyes not trained enough in this area too. But I think the change LGTM.

PR #13035, opened 2025-10-16
index planning: fix planner creation

dimitarvdimitrov:
#### What this PR does

we were skipping creating planners for on-disk blocks. This happened because planners are created for each block upon opening the prometheus TSDB. The callback to create a planner needed to have the userDB created and saved in the Ingester. The userDB wasn't saved because it was still being created.

This PR moves the planer creation to the userDB instead of the ingester.

this was introduced in #12607 

#### Checklist

- [ ] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/ingester/ingester.go
+++ b/pkg/ingester/ingester.go
@@ -2630,27 +2630,6 @@ func (i *Ingester) getTSDB(userID string) *userTSDB {
 	return db
 }
 
-// getIndexLookupPlannerFunc returns the appropriate index lookup planner function based on configuration.
-// When index lookup planning is enabled, it first checks if a pre-computed planner exists in the repository.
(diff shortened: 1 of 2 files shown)


chencs:
This is probably tricky, but I'm wondering if there's a way that we can test for this behavior.

chencs (approved):
Approving for now to unblock, just have that one question outstanding about whether we can write a test for this.

dimitarvdimitrov:
i couldn't come up with anything. i expect that being able to run benchmarks as well correctness tests from #13041 and its follow-up PR will tell us if the planners are wired correctly. Making perf improvements should effectively be tied to posting a benchmark result on the PR, so I think we will be covering for this ourselves

PR #13038, opened 2025-10-16
[compactor]: add health-check to ensure read/write access to data volume

aarogoss:
#### What this PR does
This PR adds a new health check to the compactor which tests the /data volume for read/write access.  We implement the `CheckReady` function for the `MultitenantCompactor`; this function runs these three operations:
* write an empty file to the `/data` partition
* stat the file
* read the file

If any of these three operations fails, the check fails and the compactor won't start.  If they succeed, we will attempt to remove the file, and log a warning if that operation fails.

The `mimir.go` module's `Mimir.readyHandler` is also updated so this function is called when starting up the compactor.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.


dimitarvdimitrov:

PR #13040, opened 2025-10-16
compactor: allow skipping jobs with blocks' maxTime in the future

narqo:
#### What this PR does

When the compactor runs against blocks produced from the block-builder, we observed an elevated and unbalanced workload. We believe this's due to the fact that the block-builder produces blocks during the whole period of an hour. This pattern is very different from how ingesters produce blocks, at (almost) exactly the beginning of every second hour.

In this PR I'm updating how the compactor skips the "too recent" blocks. The theory is that this will allow making its workload more smooth (will need to verify that against a dev cell, first).

The feature is controlled by the experimental `-compactor.first-level-compaction-skip-future-max-time` (off by default).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,7 +88,7 @@
 * [ENHANCEMENT] Querier: prefer querying ingesters and store-gateways in a specific zone when `-querier.prefer-availability-zone` is configured. Added the following metrics tracking the data transfer between the querier and ingesters / store-gateways respectively: #13045
   * `cortex_ingester_client_transferred_bytes_total{ingester_zone="..."}`
   * `cortex_storegateway_client_transferred_bytes_total{store_gateway_zone="..."}`
-* [ENHANCEMENT] Compactor: Add experimental `-compactor.first-level-compaction-skip-future-max-time` flag to skip first-level compactio if any source block has a MaxTime more recent than the wait period threshold. #13040
+* [ENHANCEMENT] Compactor: Add experimental `-compactor.first-level-compaction-skip-future-max-time` flag to skip first-level compaction if any source block has a MaxTime more recent than the wait period threshold. #13040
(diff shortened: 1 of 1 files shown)


pkg/compactor/job.go
```diff
@@ -174,6 +175,10 @@ func jobWaitPeriodElapsed(ctx context.Context, job *Job, waitPeriod time.Duratio
 			continue
 		}
 
+		if skipFutureMaxTime && timestamp.Time(meta.MaxTime).After(threshold) {
+			return false, meta, nil
+		}
+
```
narqo, line 181: Note for the reviewers: this is the seed of the changes in the PR. The rest are mechanical changes to support the new feature flag.

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@
 * [ENHANCEMENT] Update Docker base images from `alpine:3.22.1` to `alpine:3.22.2`. #12991
 * [ENHANCEMENT] Ruler: Add the `ruler_max_rule_evaluation_results` per-tenant configuration option to limit the maximum number of alerts an alerting rule or series a recording rule can produce for the group. By default, no limit is enforced. #12832
 * [ENHANCEMENT] Jsonnet: Changed the default KV store for the HA tracker from etcd to memberlist. Etcd and Consul are now deprecated for HA tracker usage but remain supported for backward compatibility. #13000
+* [ENHANCEMENT] Compactor: Add experimental `-compactor.first-level-compaction-skip-future-max-time` flag to skip first-level compactio if any source block has a MaxTime more recent than the wait period threshold. #13040
```
andyasp:
```suggestion
* [ENHANCEMENT] Compactor: Add experimental `-compactor.first-level-compaction-skip-future-max-time` flag to skip first-level compaction if any source block has a MaxTime more recent than the wait period threshold. #13040
```

PR #13041, opened 2025-10-16
index planning: add local benchmarks against TSDB blocks and "query stats" queries

dimitarvdimitrov:
Adds benchmarks for index lookup planning using real PromQL queries from production logs against actual ingester data.

The existing benchmarks use synthetic queries and data. Running against production queries helps identify performance characteristics and edge cases we'd otherwise miss.

Next steps for a follow-up PR: add comparison of query results between running with and without index lookup optimizations

Major changes:
- `pkg/util/bench`: New query loader that parses PromQL queries from Loki logs, extracts vector selectors, and supports tenant/query ID filtering and sampling
- `pkg/ingester/lookupplan/benchmarks`: Benchmark suite that loads ingester data and executes queries via `QueryStream`
- `pkg/streamingpromql/benchmarks`: Made ingester setup configurable via options pattern; exposed some helper functions for setting up an ingester

related to https://github.com/grafana/mimir/issues/11920

### DIFF
--- a/pkg/util/bench/query_loader.go
+++ b/pkg/util/bench/query_loader.go
@@ -318,20 +318,13 @@ func (q *Query) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-// parseTime parses a time string that can be either a Unix timestamp (float) or RFC3339Nano format.
+// parseTime parses a time string in RFC3339Nano format.
 func parseTime(str string) (time.Time, error) {
-	t, err := strconv.ParseFloat(str, 64)
+	timestamp, err := time.Parse(time.RFC3339Nano, str)
 	if err != nil {
-		timestamp, err := time.Parse(time.RFC3339Nano, str)
-		if err != nil {
-			return time.Time{}, err
-		}
-		return timestamp, nil
+		return time.Time{}, err
 	}
-
-	s, ns := math.Modf(t)
-	ns = math.Round(ns*1000) / 1000
-	return time.Unix(int64(s), int64(ns*float64(time.Second))).UTC(), nil
+	return timestamp, nil
 }
 
 // parseDuration parses a duration string that can be either a float (seconds) or Prometheus duration format.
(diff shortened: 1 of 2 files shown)


pkg/util/bench/query_loader.go
```diff
@@ -0,0 +1,416 @@
...
+			return fmt.Errorf("failed to parse step: %w", err)
+		}
+		q.Step = step
+	}
+
+	return nil
+}
+
+// parseTime parses a time string that can be either a Unix timestamp (float) or RFC3339Nano format.
+func parseTime(str string) (time.Time, error) {
```
chencs, line 322: Is all the logic to parse from a float necessary. It looks like we always format as RFC3339Nano ([code](https://github.com/grafana/mimir/blob/bdb104564726b8b0e919e25d636aad7f39b1aea2/pkg/frontend/transport/handler.go#L459)).

pkg/util/bench/query_loader.go
```diff
@@ -0,0 +1,416 @@
...
+		return timestamp, nil
+	}
+
+	s, ns := math.Modf(t)
+	ns = math.Round(ns*1000) / 1000
+	return time.Unix(int64(s), int64(ns*float64(time.Second))).UTC(), nil
+}
+
+// parseDuration parses a duration string that can be either a float (seconds) or Prometheus duration format.
+func parseDuration(s string) (time.Duration, error) {
```
chencs, line 331: Same as [my comment on parseTime](https://github.com/grafana/mimir/pull/13041/files#r2441384305) -- we always format step duration as an int ([code](https://github.com/grafana/mimir/blob/bdb104564726b8b0e919e25d636aad7f39b1aea2/pkg/frontend/transport/handler.go#L466-L468))

pkg/util/bench/query_loader.go
```diff
@@ -0,0 +1,416 @@
...
+//	  '{namespace="mimir", name="query-frontend"} |= "query stats" | logfmt | path=~".*/query(_range)?"' \
+//	  > logs.json
+//
+// Each log line should have the following structure:
+//
+//	{
+//	  "labels": {
+//	    "param_query": "sum(rate(metric[5m]))",
+//	    "param_start": "1697452800",
+//	    "param_end": "1697456400",
```
chencs, line 51: Same as [other comment](https://github.com/grafana/mimir/pull/13041/files#r2441384305) -- I think we always log these as RFC3339Nano

dimitarvdimitrov:
thanks for catching this @chencs , i copied the code from the query-step-alignment-analysis https://github.com/grafana/mimir/blob/2a2095b4594ed168d339f3d9b34b42ac9aa48e0b/tools/query-step-alignment-analysis/main.go#L28-L41

and that was from the time when we would just out the literal parameter from the request without parsing it (and it could be a float)

chencs (approved):
This is neat! Thanks for setting it up

PR #13045, opened 2025-10-17
Add store-gateway support for -querier.prefer-availability-zone

pracucci:
#### What this PR does

We currently have `-querier.prefer-availability-zone` that allows to specify the preferred zone when querying ingesters. In this PR I'm extending it to store-gateways too. The CLI flag was experimental and hidden. In this PR I'm exposing it in the doc, but it's still experimental.

In this PR I've also added 2 new metrics, tracking the data transfer between queriers and ingesters / store-gateways respectively:

- `cortex_ingester_client_transferred_bytes_total{ingester_zone="..."}`
- `cortex_storegateway_client_transferred_bytes_total{store_gateway_zone="..."}`

These two metrics should make it easily to monitor the cross-AZ data transfer that happens during fallbacks.

N/A

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/querier/store_gateway_client_test.go
+++ b/pkg/querier/store_gateway_client_test.go
@@ -67,12 +67,17 @@ func Test_newStoreGatewayClientFactory(t *testing.T) {
 	// we can't predict the exact time it took, we need to workaround it.
 	metrics, err := reg.Gather()
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


pracucci:
Tested in a dev cell, looks working as expected. The metrics also match with the container networking metrics.

PR #13058, opened 2025-10-20
Remote execution: fix issue where labels can be corrupted

### LINKED ISSUE #12982: Flaky TestMaxQueryParallelismWithRemoteExecution
## `TestMaxQueryParallelismWithRemoteExecution`

`pkg/frontend/v2/remoteexec_test.go`

### About This Issue
This issue tracks a flaky test that has been detected failing inconsistently. Each week our analysis tool runs and detects this test as flaky, it will add a comment below with recent failure data and who might be able to help.

### 🔍 How to investigate
1. **Run it locally** to see if you can reproduce: `go test -count=10000 -run TestMaxQueryParallelismWithRemoteExecution ./...`
2. **Check the failure logs** in the comments below - they might show a pattern
3. **Look for timing issues** - race conditions, timeouts, or external dependencies
4. **Review recent changes** - check commits that modified this test or the code under test recently

### 🎯 Next steps
- **Can you fix it?** Great! That would help the whole team
- **Need help?** Ask someone who has worked on this area recently
- **Can't fix it right now?** Consider adding logs and metrics so that next time it's easier to debug.
- **Obsolete test?** Maybe it's time to remove it or skip it with `t.Skip()`

_This test has been identified as flaky by [go-flaky-tests](https://github.com/grafana/shared-workflows/tree/main/actions/go-flaky-tests)._

charleskorn:
#### What this PR does

This PR fixes an issue where queries evaluated with remote execution can have corrupted labels.

This occurred when the underlying buffer for the gRPC message containing the labels was reused before the labels could be converted to a `labels.Labels` instance by `mimirpb.FromLabelAdaptersToLabels`.

This PR also fixes an issue where `TestMaxQueryParallelismWithRemoteExecution` can flake (resolves #12982), as it was consistently flaking with the change to use a real gRPC stream in the tests.

I would recommend reviewing each commit separately.

(none)

#### Checklist

- [x] Tests updated.
- [n/a] Documentation added.

### DIFF
--- a/pkg/frontend/v2/frontend_test.go
+++ b/pkg/frontend/v2/frontend_test.go
@@ -16,6 +16,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -80,7 +81,7 @@ func setupFrontendWithConcurrencyAndServerOptions(t testing.TB, reg prometheus.R
 
 	schedulerOpts := opts
 	frontendOpts := append(
-		opts,
+		slices.Clone(opts),
 		grpc.StreamInterceptor(middleware.StreamServerUserHeaderInterceptor),
 		grpc.UnaryInterceptor(middleware.ServerUserHeaderInterceptor),
 	)


dimitarvdimitrov:

pkg/frontend/v2/frontend_test.go
```diff
@@ -64,22 +66,35 @@ func setupFrontend(t testing.TB, reg prometheus.Registerer, schedulerReplyFunc f
...
+	frontendListener, err := net.Listen("tcp", "localhost:0")
 	require.NoError(t, err)
 
-	server := grpc.NewServer(opts...)
+	schedulerListener, err := net.Listen("tcp", "localhost:0")
+	require.NoError(t, err)
+
+	schedulerOpts := opts
+	frontendOpts := append(
+		opts,
```
56quarters: Won't this use the memory from `opts` if it has enough space for the additional interceptors? It's not clear to me if this will end up modifying `schedulerOpts`.

pkg/frontend/v2/frontend_test.go
```diff
@@ -1313,6 +1315,79 @@ func TestFrontend_Protobuf_ResponseWithUnexpectedUserID(t *testing.T) {
...
+
+				resp.Close()
+			}
+		})
+	}
+
+	wg.Wait()
+}
+
+func generateConcurrencyTestResponse(idx uint64) *frontendv2pb.QueryResultStreamRequest {
```
56quarters, line 1396: There's a surprising amount of things going on in this method. Can you add some comments that explain why we're generating each of these labels for a particular response?

pkg/frontend/v2/remoteexec.go
```diff
@@ -123,11 +123,14 @@ type scalarExecutionResponse struct {
...
 
 func (r *scalarExecutionResponse) GetValues(ctx context.Context) (types.ScalarData, error) {
-	resp, err := readNextEvaluateQueryResponse(ctx, r.stream)
+	resp, releaseMessage, err := readNextEvaluateQueryResponse(ctx, r.stream)
 	if err != nil {
 		return types.ScalarData{}, err
 	}
 
+	// Nothing in the message retains a reference to the underlying buffer, so we can release it immediately.
+	releaseMessage()
```
56quarters: I don't really like that we're asserting that we can immediately free the buffer associated with this message. It seems like it would be more correct to always `defer releaseMessage()` so that we aren't encoding how the query response message works here.

56quarters (approved):
LGTM with some non-blocking comments.

pkg/frontend/v2/remoteexec.go
charleskorn: Good point, fixed in 03cecbc.

pkg/frontend/v2/frontend_test.go
charleskorn: I've added some explanation in 34ae032 - how's that?

pkg/frontend/v2/frontend_test.go
charleskorn: I think this is safe, but I'll add a clone here to make it very clear what the intention is.

charleskorn: Done in 8df0e7f.

PR #13067, opened 2025-10-20
[compactor]: add health-check to ensure read/write access to data volume 

aarogoss:
This PR adds a new health check to the compactor which tests the /data volume for read/write access.  We implement the `CheckReady` function for the `MultitenantCompactor`; this function runs these three operations:
* write an empty file to the `/data` partition
* stat the file
* read the file

If any of these three operations fails, the check fails and the compactor won't start.  If they succeed, we will attempt to remove the file, and log a warning if that operation fails.

There is also a check before running these operations which tests for the `-compactor.data-dir` config value readiness:
* if the value is empty, log a warning and return nil
* stat the value: if it doesn't exist yet, lag a warning and return nil
* if stat fails w/a different error, log an error msg and return the error

The `mimir.go` module's `Mimir.readyHandler` is also updated so this function is called when starting up the compactor.

N/A

- [ ] Tests updated.
- [ ] Documentation added.
- [ ] [`about-versioning.md`](https://github.com/grafana/mimir/blob/main/docs/sources/mimir/configure/about-versioning.md) updated with experimental features


dimitarvdimitrov (commented):
we already check the directory on startup 

https://github.com/grafana/mimir/blob/d7dde5ab5c47ed75e6aa28f8c5732d93d9b5e1eb/pkg/mimir/sanity_check.go#L62-L64


can you share a bit about the problem you're solving with duplicating this in the readiness check?

aarogoss:
> we already check the directory on startup
> 
> https://github.com/grafana/mimir/blob/d7dde5ab5c47ed75e6aa28f8c5732d93d9b5e1eb/pkg/mimir/sanity_check.go#L62-L64
> 
> can you share a bit about the problem you're solving with duplicating this in the readiness check?

Hey Dimitar, yes, this came out of a situation where the `-compactor.data-dir` filesystem was corrupted and the compactor was no longer able to write to it.  From what I've been able to find, we have a sanity-check which tests the volume, but I don't believe this is run periodically at runtime.  If we do already have this check running regularly, then I missed it and this is redundant and can absolutely be closed.

dimitarvdimitrov:
what's the idea behind failing the readiness check? the compactor doesn't serve any requests, so taking the pod out of the service would do much

aarogoss:
> what's the idea behind failing the readiness check? the compactor doesn't serve any requests, so taking the pod out of the service would do much

Upon further discussion, the intention here should be to implement the `-compactor.data-dir` tests using a liveness check rather than a readiness check.  I'll take this back to the drawing board so I can better understand how that particular check would be implemented.

tacole02 (commented):
Changelog LGTM

PR #13095, opened 2025-10-22
[release-2.17] Base tooling images on distroless/static-debian12

pr00se:
Backport https://github.com/grafana/mimir/pull/13014 to release-2.17

### DIFF
--- a/cmd/mimirtool/Dockerfile
+++ b/cmd/mimirtool/Dockerfile
@@ -1,8 +1,11 @@
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: AGPL-3.0-only
 
-FROM       alpine:3.22.2
-ARG        EXTRA_PACKAGES
-RUN        apk add --no-cache ca-certificates tzdata $EXTRA_PACKAGES
+ARG        BASEIMG=gcr.io/distroless/static-debian12
+# TODO(rwwiv): Remove once CA bundle is updated upstream.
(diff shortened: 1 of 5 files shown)


cmd/mimirtool/Dockerfile
```diff
@@ -1,13 +1,23 @@
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: AGPL-3.0-only
```
charleskorn, line 1: Is there a reason why this has changed?

narqo:
In https://github.com/grafana/mimir/pull/13091 @aknuds1 works on pinning the distroless images to a specific sha. Maybe we could do the same here?

aknuds1:
> Maybe we could do the same here?

Thanks for mentioning it @narqo, I would recommend it! I'm backporting that PR to v2.17 BTW. I don't think it can hurt.

aknuds1:
My [backport](https://github.com/grafana/mimir/pull/13106) of #13091. It should be complementary to this PR.

cmd/mimirtool/Dockerfile
pr00se: I changed this in the original PR to match the other files without the `Provenance` tags (e.g. copyblocks)

mattmendick: mimirtool can't be Apache-2.0 right? It uses packages from Mimir which are AGPL?

PR #13096, opened 2025-10-22
fix: flaky `TestWriter_WriteSync` due to explicit context cancellation 

### LINKED ISSUE #12703: Flaky TestWriter_WriteSync
## `TestWriter_WriteSync`

`pkg/storage/ingest/writer_test.go`

### About This Issue
This issue tracks a flaky test that has been detected failing inconsistently. Each week our analysis tool runs and detects this test as flaky, it will add a comment below with recent failure data and who might be able to help.

### 🔍 How to investigate
1. **Run it locally** to see if you can reproduce: `go test -count=10000 -run TestWriter_WriteSync ./...`
2. **Check the failure logs** in the comments below - they might show a pattern
3. **Look for timing issues** - race conditions, timeouts, or external dependencies
4. **Review recent changes** - check commits that modified this test or the code under test recently

### 🎯 Next steps
- **Can you fix it?** Great! That would help the whole team
- **Need help?** Ask someone who has worked on this area recently
- **Can't fix it right now?** Consider adding logs and metrics so that next time it's easier to debug.
- **Obsolete test?** Maybe it's time to remove it or skip it with `t.Skip()`

_This test has been identified as flaky by [go-flaky-tests](https://github.com/grafana/shared-workflows/tree/main/actions/go-flaky-tests)._

rMaxiQp:
#### What this PR does

Current `TestWriter_WriteSync`'s subtest "should interrupt the WriteSync() on context cancelled but other concurrent requests should not fail" fires two gorountines after the first request's completion, and cancel the first gorountine's context after there are 3 records buffered. However, the buffered record doesn't imply the write has completed, and the first gorountine's context can be cancelled before it expires (due to concurrency)

To address this issue,  we can remove the `cancelSecondRequest()` to avoid the context cancellation.

Fixes #12703

#### Checklist

- [x] Tests updated.
- [ ] Documentation added.

### DIFF
--- a/pkg/storage/ingest/writer_test.go
+++ b/pkg/storage/ingest/writer_test.go
@@ -327,12 +327,9 @@ func TestWriter_WriteSync(t *testing.T) {
 		// Once the 1st Produce request is received by the server but still processing (there's a 1s sleep),
 		// issue two more requests. One with a short context timeout (expected to expire before the next Produce
 		// request will be sent) and one with no timeout.
-		var cancelSecondRequest context.CancelFunc
 
 		runAsyncAfter(&wg, firstRequestReceived, func() {
-			var secondRequestCtx context.Context
-
-			secondRequestCtx, cancelSecondRequest = context.WithTimeout(ctx, 10*time.Millisecond)
+			secondRequestCtx, cancelSecondRequest := context.WithTimeout(ctx, 10*time.Millisecond)
 			t.Cleanup(cancelSecondRequest)
 
 			assert.ErrorIs(t, writer.WriteSync(secondRequestCtx, partitionID, tenantID, &mimirpb.WriteRequest{Timeseries: series2, Metadata: nil, Source: mimirpb.API}), context.DeadlineExceeded)


pkg/storage/ingest/writer_test.go
```diff
@@ -352,9 +352,6 @@ func TestWriter_WriteSync(t *testing.T) {
 			return client.BufferedProduceRecords() == 3
 		}, time.Second, 10*time.Millisecond)
 
-		// Cancel the 2nd request context.
-		cancelSecondRequest()
```
56quarters, line 356: Since we aren't calling the cancel function from outside the closure, the declaration above could be removed and local to the closure for the second request. e.g.

```go
runAsyncAfter(&wg, firstRequestReceived, func() 
	secondRequestCtx, cancelSecondRequest := context.WithTimeout(ctx, 10*time.Millisecond)
	t.Cleanup(cancelSecondRequest)

	assert.ErrorIs(t, writer.WriteSync(secondRequestCtx, partitionID, tenantID, &mimirpb.WriteRequest{Timeseries: series2, Metadata: nil, Source: mimirpb.API}), context.DeadlineExceeded)
})
```

rMaxiQp: Hi @56quarters , thank you for the review! 

I've address your suggestion, and will wait for @pracucci's review to confirm this change is reasonable.

